PR #862, opened 2019-11-26
MIT license for pixellink and changelog

benhoff:


nmanovic:
@benhoff , thanks for your time and help!

PR #867, opened 2019-11-27
keep the file order provided in the task data api

godlikejay:
There are some cases that the file order is significant in the annocation process. 

The previous implementation uses `dict` to store intermediate data, and `dict.keys()` to get the list back.  So that the file order is lost as Python prior to 3.6 does not guarantee the return order from `dict.keys()`.

As of Python 3.6, for the CPython implementation of Python, dictionaries [maintain insertion order](https://docs.python.org/3.6/whatsnew/3.6.html#new-dict-implementation) by default. But for compatibility, this pull request uses `OrderedDict` to keep the insertion order for both Python 3.6+ and prior.

### DIFF
--- a/cvat/apps/engine/media_extractors.py
+++ b/cvat/apps/engine/media_extractors.py
@@ -39,7 +39,7 @@ class ImageListExtractor(MediaExtractor):
         if not source_path:
             raise Exception('No image found')
         super().__init__(
-            source_path=sorted(source_path),
+            source_path=source_path,
             dest_path=dest_path,
             image_quality=image_quality,
             step=1,
--- a/cvat/apps/engine/task.py
+++ b/cvat/apps/engine/task.py
@@ -13,6 +13,7 @@ from ast import literal_eval
 from urllib import error as urlerror
 from urllib import parse as urlparse
 from urllib import request as urlrequest
+from collections import OrderedDict
 
 from cvat.apps.engine.media_extractors import get_mime, MEDIA_TYPES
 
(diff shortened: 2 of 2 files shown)


cvat/apps/engine/media_extractors.py
```diff
@@ -39,7 +39,7 @@ def __init__(self, source_path, dest_path, image_quality, step=1, start=0, stop=
         if not source_path:
             raise Exception('No image found')
         super().__init__(
-            source_path=sorted(source_path),
+            source_path=source_path,
```
nmanovic: It is a very significant change. The primary problem that we can read the list of images from a directory (client directory, server share directory). OS doesn't guarantee the order. Thus the `sorted` operation is the way to make the process predictable. When you want to generate a pre-annotation you have to know the order of images. Using the technic (`sorted` operation) you can "repeat" the order locally in your script.

Thus the solution cannot be accepted. Need to find a more general one.

godlikejay: Yes, OS doesn't guarantee the order of the directory contents. But for the directory case, I think it is already uniformed in [DirectoryExtractor](https://github.com/opencv/cvat/blob/develop/cvat/apps/engine/media_extractors.py#L135).

nmanovic: But if you send a list of files from a browser (file dialog) you cannot control its order. You even cannot know the order. Right?

I believe the only way to solve the problem is to use filename (or path) as ID instead of indexes on the server. If you can prepare such patch we will accept it.

For now I will close the PR because it will give us some problems. I understand that from the command line you can specify the order of files. But from UI it isn't possible.

PR #869, opened 2019-11-27
The tfrecord file starts at id 0, which is incompatible with tensorflow.

benhoff:
It has to start at 1, 0 is reserved in Tensorflow.

Ref: https://github.com/opencv/cvat/issues/866

### DIFF
--- a/cvat/apps/annotation/tfrecord.py
+++ b/cvat/apps/annotation/tfrecord.py
@@ -98,7 +98,7 @@ def dump(file_object, annotations):
         with codecs.open(os.path.join(out_dir, labelmap_file), 'w', encoding='utf8') as f:
             for label, idx in label_ids.items():
                 f.write(u'item {\n')
-                f.write(u'\tid: {}\n'.format(idx))
+                f.write(u'\tid: {}\n'.format(idx+1))
                 f.write(u"\tname: '{}'\n".format(label))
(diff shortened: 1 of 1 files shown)


cvat/apps/annotation/tfrecord.py
```diff
@@ -98,7 +98,7 @@ def create_tf_example(img_id, img_size, image_name, boxes, label_ids):
         with codecs.open(os.path.join(out_dir, labelmap_file), 'w', encoding='utf8') as f:
             for label, idx in label_ids.items():
                 f.write(u'item {\n')
-                f.write(u'\tid: {}\n'.format(idx))
+                f.write(u'\tid: {}\n'.format(idx+1))
```
nmanovic: @benhoff , I see that label_ids is used below. Should be fix label_ids itself or adjust only the line? Did you reproduce the issue and the change really fix it?

@zhiltsov-max , could you please look and advise?

zhiltsov-max: It looks like it should be done above, where `label_ids = OrderedDict` is declared.

benhoff: I was looking at this: https://github.com/onepanelio/cvat/commit/674cdceed76cf5258346dd4db0f795c3cd95bd84

Didn't pull it down to check.

PR #876, opened 2019-11-28
[Datumaro] Disable lazy image caching by default

zhiltsov-max:
Temporarily fixes server OOM/swap DOS for large dataset export.

### DIFF
--- a/datumaro/tests/test_images.py
+++ b/datumaro/tests/test_images.py
@@ -1,0 +1,33 @@
+import numpy as np
+import os.path as osp
+from PIL import Image
+import timeit
+
+from unittest import TestCase
+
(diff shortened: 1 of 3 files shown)


datumaro/tests/test_images.py
```diff
@@ -0,0 +1,33 @@
...
+from PIL import Image
+import timeit
+
+from unittest import TestCase
+
+from datumaro.util.test_utils import TestDir
+from datumaro.util.image import lazy_image
+
+
+class LazyImageTest(TestCase):
```
nmanovic: I will recommend modify the test in accordance with the hint: https://stackoverflow.com/questions/5445080/compare-if-two-variables-reference-the-same-object-in-python

PR #885, opened 2019-12-02
VOC converter: Use depth from CVAT XML if available

a-andre:


nmanovic:
@a-andre , thanks for the PR. I will merge it but for your information the script is deprecated. Use UI or REST API or Datumaro framework to convert annotations from CVAT to VOC.

PR #890, opened 2019-12-03
added in OpenVINO road semantic segmentation

benhoff:


utils/open_model_zoo/Transportation/semantic-segmentation-adas/interp.py
```diff
@@ -0,0 +1,27 @@
...
+
+for frame_results in detections:
+    frame_height = frame_results['frame_height']
+    frame_width = frame_results['frame_width']
+    frame_number = frame_results['frame_id']
+    detection = frame_results['detections']
+    detection = detection[0, 0, :, :]
+    width, height = detection.shape
+
+    for i in range(20):
```
nmanovic: @benhoff , does 0 corresponds to background? Should we add it into label_map? Should we have here `range(21)`?

nmanovic:
@benhoff , it is a great contribution as usual. Thanks for all your time. Let's clarify a couple of questions and merge.

utils/open_model_zoo/Transportation/semantic-segmentation-adas/interp.py
benhoff: Off by one strikes! Good catch.

benhoff:
I'm not sure about the background thing. Let me take a look.

benhoff:
@nmanovic , I couldn't find any documentation about a background instance.

However, the demo in the `open_model_zoo` seems to suggest there is a 20th index.

See here: https://github.com/opencv/open_model_zoo/blob/e372d4173e50741a6828cda415d55c37320f89cd/demos/python_demos/segmentation_demo/segmentation_demo.py#L140

Doing some testing on an image I have locally seems to suggest that the first index is a road. The other labels seem to line up well with the mappings as they are.

nmanovic:
@benhoff , thanks!

PR #892, opened 2019-12-03
React  & Antd: Dashboard migration

bsekachev:
image

image

image


nmanovic:
Problems:
- [ ] need to configure our proxy to work without ports (internal)
- [ ] fix open task in annotation view
- [ ] fix preview (sometimes they are not shown)
- [ ] add links on notifications about tasks
- [ ] remove error messages from console (to avoid confusion)
- [ ] do we really need to open download link in a separate tab? Please investigate
- [ ] REST API with port doesn’t work in UI. Why? Please investigate (internal)
- [ ] fix page navigation
- [ ] page_size query parameter doesn’t work in dashboard (pass it to REST API).

Ideas (submit issues if it is necessary):
- [ ] Configure number of tasks per page
- [ ] Specify ordering of tasks (by creating time, name, owner, …)

PR #906, opened 2019-12-05
[Datumaro] Add YOLO converter

### LINKED ISSUE #891: Add respective frames to YOLO dump
**Description**
I don't know if this feature already exists, but to train a darknet YOLOv3 model, I need to also get the frames and `obj.data` generated. The model gives an `obj.names` but there is no `obj.data` present and I would like to generate this for training a detector. 

Can I try making a change for the same?

**Proposed change**
```python
## Add the image file somewhere here?
with ZipFile(file_object, "w") as output_zip:
        for frame_annotation in annotations.group_by_frame():
            image_name = frame_annotation.name
            annotation_name = "{}.txt".format(os.path.splitext(os.path.basename(image_name))[0])
            width = frame_annotation.width
            height = frame_annotation.height

            # TODO : Add this line
'''
            output_zip.write(image_name,{}.jpg".format(os.path.splitext(os.path.basename(image_name))[0]), 
compression=Optional)
'''

            yolo_annotation = ""
            for shape in frame_annotation.labeled_shapes:
                if shape.type != "rectangle":
                    continue

                label = shape.label
                yolo_bb = convert_to_yolo((width, height), shape.points)
                yolo_bb = " ".join("{:.6f}".format(p) for p in yolo_bb)
                yolo_annotation += "{} {}\n".format(labels_ids[label], yolo_bb)
     
            output_zip.writestr(annotation_name, yolo_annotation)
        output_zip.writestr("obj.names", "\n".join(l[0] for l in sorted(labels_ids.items(), key=lambda x:x[1])))
```

zhiltsov-max:
Closes #891

Added:
- YOLO support in Datumaro

### DIFF
--- a/cvat/apps/dataset_manager/task.py
+++ b/cvat/apps/dataset_manager/task.py
@@ -379,7 +379,12 @@ EXPORT_FORMATS = [
         'name': 'MS COCO',
         'tag': 'coco',
         'is_default': False,
-    }
+    },
+    {
+        'name': 'YOLO',
+        'tag': 'yolo',
+        'is_default': False,
+    },
 ]
 
 def get_export_formats():
--- a/datumaro/datumaro/components/converters/__init__.py
+++ b/datumaro/datumaro/components/converters/__init__.py
@@ -23,6 +23,8 @@ from datumaro.components.converters.voc import (
     VocSegmentationConverter,
 )
 
+from datumaro.components.converters.yolo import YoloV3Converter
+
 
 items = [
     ('datumaro', DatumaroConverter),
@@ -40,4 +42,6 @@ items = [
     ('voc_segm', VocSegmentationConverter),
     ('voc_action', VocActionConverter),
     ('voc_layout', VocLayoutConverter),
+
+    ('yolo_v3', YoloV3Converter),
 ]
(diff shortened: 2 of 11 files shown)


nmanovic:
@zhiltsov-max ,

image

1. Content of obj.data doesn't correspond to the layout of the downloaded directory. Is it OK?

```
classes = 2
train = data/train.txt
names = data/obj.names
backup = backup/
```

2. Also obj_train_data contains empty txt files. Is it OK?

3. Should obj_train_data to be renamed?

4. What is `backup = backup/`?

HagegeR:
@nmanovic:  

1. the content of obj.data is consistent with the way yolo works. Typically the file tells the darknet framework where are all the files that it need in order to train.

2. empty text files are used for images with no annotation, I hope that was the case here

4. `backup = backup/` is the place where all the weights being trained and/or checkpoints in the training process are.

can we get some options as well? 

such as: split of the dataset (train/valid/test), attribute filtering (occluded), class filtering (in case the weights are train for specific task that doesn't need the whole dataset)

nmanovic:
[task_road2-2019_12_09_11_15_03-datumaro_project.zip](https://github.com/opencv/cvat/files/3939445/task_road2-2019_12_09_11_15_03-datumaro_project.zip)

Issue #896, opened 2019-12-04
'Annotation has failed' for YOLOv3 object detection in CVAT.


PareshKamble:
Hi!

I want to use YOLOv3 for automatic annotation using CVAT tool. I understand that I need .bin, .xml, mapping.json and interp.py files. 

I followed the steps given [here](https://docs.openvinotoolkit.org/latest/_docs_MO_DG_prepare_model_convert_model_tf_specific_Convert_YOLO_From_Tensorflow.html#yolov3-overview) for generation of .pb file and IR files subsequently. For the above said task, I used official [yolov3](https://pjreddie.com/media/files/yolov3.weights) weights and custom attributes from the corresponding [cfg](https://github.com/pjreddie/darknet/tree/master/cfg) file. 
I successfully generated the IR files.

Later, I used the mapping.json and interp.py files provided [here](https://github.com/opencv/cvat/tree/develop/utils/open_model_zoo/yolov3). Now, I had the 4 required files for auto annotation at CVAT.

Finally, I provided these 4 files as YOLOv3 through the **model manager** option in CVAT GUI for object detection. When I press the **_Run Auto Annotation_** button, the process completes with **_Annotation has Failed_** message.

Is there something I am missing? 
Is there any method to know what might have gone wrong?
During the process of IR files generation, I kept the yolov3.json file as it is since the custom_attributes were the same in the provided file and that in yolov3.cfg [here](https://github.com/pjreddie/darknet/tree/master/cfg)

Any help would be highly appreciated!

nmanovic:
@benhoff ,

Do you think you can answer on the question?

benhoff:
@PareshKamble , can you attach the logs? `$ docker cvat log` is the command I believe.

Is there a traceback that is provided when you submit the 4 files? If not, what version of cvat are you using?

PareshKamble:
Thanks @benhoff for the reply. The command is correct but the log is too big to be attached. Do you want a particular portion of the log to be attached or the complete log? 

No traceback is provided when I submit the 4 files. Interestingly, everything works smooth unless I run the auto annotation. How to know the version of cvat? However, I cloned cvat on 23 November 2019. Hope it does help? 

I get the below shown message after running the auto annotation. First the message as shown in the left image then the right image.
[...]

PR #912, opened 2019-12-07
Yolov3 interpretration script fix for 'Annotation failed' issue (#896)

himalayanZephyr:
- Enforcing checks on bbox coordinates to not go out of boundary
- Changing the mapping.json to have index starting from 0 instead of 1 (this was detected after noticing that for 'Person' label, bicycles were being detected instead during auto annotation)


benhoff:
@nmanovic, it looks good to me too. This will fix the issues from the logs I received.

As a side note, I have yet to stand up a functioning cvat environment after leaving my last job. Hopefully I'll get some time next week.

Issue #913, opened 2019-12-07
User Experience Issues with Auto Annotation Run Model Script


benhoff:
If an auto annotation script fails for someone, it is currently difficult to debug with a casual user.

Ref: https://github.com/opencv/cvat/issues/896

Ref: There's a conversation in gitter where Boris walked through with someone how to copy relevant files from their local machine to docker so they could use the `run_model.py` script.

This is a current deficiency with the auto annotation `run_model.py` script. The script is setup to locally debug, but it would be better to be a more general user-facing debugging script.

Need to implement a second interface where users can identify:

1. The relevant model
2. The relevant task

and get a traceback that they can share for easier debugging.

This would solve the issue of explaining to users how to copy files from their local machines to docker. I also had a user (with a functioning CVAT docker installation), try to run `run_model.py` script but did not correctly activate OpenVINO's `setupvars.sh` on their local machine. This lead to importing OpenVINO errors that were unrelated to the problem that the user was trying to solve (poor user experience).

In order to solve this:

Lines 21-24 of the auto annotation script will need to be changed so that the `py, json, xml, and bin` files are no longer required. https://github.com/opencv/cvat/blob/32027ce884c0584015874c4ae99ba7f53ffb46c0/utils/auto_annotation/run_model.py#L21

Instead the script would enforce either the passing of the aforementioned 4 variables, or a "model name"/"task reference" number. 

Would need to explore what implementation (Docker based, REST based, etc) would make the most sense.

benhoff:
Alternatively, could setup Auto Annotation's `Load Model` UI, so that it could optionally check a task to test the model with. This would catch significantly more errors presented back to the user on upload (as it is currently testing with an all black image).

PR #931, opened 2019-12-11
added in label visualization to auto annotation runner

benhoff:
This should answer https://github.com/opencv/cvat/issues/914

PR #934, opened 2019-12-11
Add in ability to run Auto Annotation model using model name and task id

benhoff:
This needs to be tested extensively and I'm not a spot to do it currently.

This should resolve https://github.com/opencv/cvat/issues/913

@nmanovic, do you want me to merge these two pull requests? I didn't think about it until now.

### DIFF
--- a/utils/auto_annotation/README.md
+++ b/utils/auto_annotation/README.md
@@ -4,13 +4,29 @@ A small command line program to test and run AutoAnnotation Scripts.
 
 ## Instructions
 
-Change in to the root of the project directory and run
+There are two modes to run this script in. If you already have a model uploaded into the server, and you're having 
+issues with running it in production, you can pass in the model name and a task id that you want to test against.
 
 ```shell
-$ python cvat/utils/auto_annotation/run_model.py --py /path/to/python/interp.py \
-                                                 --xml /path/to/xml/file.xml \
-                                                 --bin /path/to/bin/file.bin \
-                                                 --json /path/to/json/mapping/mapping.json
+# Note that this module can be found in cvat/utils/auto_annotation/run_model.py
+$ python /path/to/run_model.py --model-name mymodel --task-id 4
+```
+
+If you're running in docker, this can be useful way to debug your model.
+
+``` shell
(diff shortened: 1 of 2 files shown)


nmanovic:
> This needs to be tested extensively and I'm not a spot to do it currently.
> 
> This should resolve #913
> 
> This is dependent on #931.
> 
> @nmanovic, do you want me to merge these two pull requests? I didn't think about it until now.

@benhoff ,

If your PRs are ready please remove WIP and I will start the review.

nmanovic:
@benhoff , thanks for the contribution. I really don't think that we have to merge the PR. Let me explain reasons behind that:
1. We are going to change layout of the task directory soon. Thus the script will not work after that. It tries to rely on something which it should not (how CVAT keeps files internally).
1. I don't think any problems for the user to run the original script inside the docker container. Why do we need an extra helper here? I don't see any reason for that.

benhoff:
@nmanovic 

It's difficult to debug a script with a user. See here: https://github.com/opencv/cvat/issues/896

Right now, because the "test case" for a submit of files into CVAT is an all black image, so it's easy to submit non-working code up into the server.

Once the code has been uploaded to the server, it's hard to get a debug script out of the server with the low level information of a traceback that `run_model.py` provides.

In the case of https://github.com/opencv/cvat/issues/896, I knew there was an issue with the code they had loaded in the server, but I couldn't figure out how to get the users to provide a traceback back to me. 

CVAT logs are too large to submit up to Github.

One user tried to run the `run_model.py` code outside of the server, but because they didn't have OpenVINO correctly installed (they forgot to activate the script to add it to their path). Debugging over email is difficult, and I didn't know what they environment was like, except I did know that they had it installed successfully in docker.

Hopefully that explains the reasoning for your second point. Is there a good way to address your first point?

nmanovic:
@benhoff , if you think that it is useful I will be glad to merge the PR. But https://github.com/opencv/cvat/pull/787 is going to break the new feature. I hope you will help us to fix that in the future. Thanks again for all your time which you spend on the project!

PR #935, opened 2019-12-11
Fix new tracker edit

### LINKED ISSUE #918: Edit Issue tracker
When change “Issue Tracking”, text “not specified” remains in the editing field. It would be more convenient if field for editing became empty (if link to issue tracking not specified).

Edit issue tracker

mmarkelov:
Fix #918 

As a workaround:

Clicking on "Edit" button will clear textarea.

1. If validation was fallen it will reset text to placeholder
2. Otherwise it will update text to input

I suppose it will be better to select all of text by clicking, but I suppose it will be not so easy to do with Antd Text implementation

### DIFF
--- a/cvat-ui/src/components/task-page/details.tsx
+++ b/cvat-ui/src/components/task-page/details.tsx
@@ -22,6 +22,7 @@ import patterns from '../../utils/validation-patterns';
 import { getReposData, syncRepos } from '../../utils/git-utils';
 
 const core = getCore();
+const PLACEHOLDER = 'Not specified';
 
 interface Props {
     previewImage: string;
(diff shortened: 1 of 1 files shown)


bsekachev:
@mmarkelov 

Hi, thanks for PR.
Unfortunately this solution doesn't work properly in any case. 
For example when I started editing and press Esc button, the edit mode closing and the field became empty.

mmarkelov:
@bsekachev I'll take a look on this case

mmarkelov:
@bsekachev unfortunately there is no method to catch end of editing of editable text right now. I suppose I can add this functionality to antd library, and fix this case

bsekachev:
@mmarkelov 

It looks like too hard way to fix such a minor issue.
I'll think how we can get around the problem (for example with combination of two elements like <Text /> and <Input />)

bsekachev:
@mmarkelov 

Let's close this PR till we have found a robust solution

PR #939, opened 2019-12-12
add in ability to upload multiple PDFs at a time

benhoff:


cvat/apps/engine/media_extractors.py
```diff
@@ -89,17 +89,19 @@ def __init__(self, source_path, dest_path, image_quality, step=1, start=0, stop=
...
+                output = os.path.join(self._temp_directory, str(count) + '.jpg')
+                count += 1
+                self._dimensions.append(page.size)
+                page.save(output, 'JPEG')
 
         self._length = len(os.listdir(self._temp_directory))
 
     def _get_imagepath(self, k):
-        img_path = os.path.join(self._temp_directory, self._basename + str(k) + '.jpg')
+        img_path = os.path.join(self._temp_directory, str(k) + '.jpg')
```
nmanovic: @benhoff ,

How to map annotations of files to original's pdf files?

nmanovic:
@benhoff , thanks for the pull request. Great contribution as always. Let's resolve some questions.

cvat/apps/engine/media_extractors.py
benhoff: I don't think it's possible with how this code base is setup.

Maybe track the information in the JPG metadata? The extractor could theoretically pull that information out on the backside.

nmanovic: @benhoff , previously you have _basename. Thus it was possible to match a jpg with corresponding pdf+page. Also I treat a PDF file as a video file which can contain multiple pages (frames). If somebody needs to annotate multiple PDF files it is necessary to create multiple tasks.

I will vote to be consistent. If we want to allow multiple pdfs per task we need to do the same for video files.

benhoff: @nmanovic , I agree that it would be nice, I'm just not sure how to program it. If you look at the `__iter__` method in the class, that is the problem. By previously having the `_basename` be consistent, and knowing how many files are in the directory, it was possible to easily iterate over all the files.

Is it possible to nest subdirectories? That would be one way to keep the files cleanly separated.

nmanovic: @benhoff , let's fix only problem with the docker file and process only one pdf for now. I like the solution with one pdf. Multiple PDFs are inconsistent with archives and video files.

benhoff:
@nmanovic 

Removed the python edits  but left the dockerfile.

PR #941, opened 2019-12-12
Improve UX with creating new shape by shortkey

mmarkelov:
This PR contains improvement for creating shape with shortkey.

Current behavior:
You click **N** key for new shape creation, but you got two red helper lines only if **mousemove** will trigger, which is is a little confusing.

As workaround:
I'm listening **mousemove** event and draw this lines when creation of new shape is triggered by shortkey

### DIFF
--- a/cvat/apps/engine/static/engine/js/shapeCreator.js
+++ b/cvat/apps/engine/static/engine/js/shapeCreator.js
@@ -409,7 +409,7 @@ class ShapeCreatorView {
         }.bind(this));
     }
 
-    _create(usingShortkey) {
+    _create() {
         let sizeUI = null;
         switch(this._type) {
         case 'box':
@@ -445,7 +445,7 @@ class ShapeCreatorView {
                     this._controller.finish(box, this._type);
                 }
 
-                this._controller.switchCreateMode(true, usingShortkey);
+                this._controller.switchCreateMode(true);
             }.bind(this)).on('drawupdate', (e) => {
                 sizeUI = drawBoxSize.call(sizeUI, this._frameContent, this._frameText, e.target.getBBox());
             }).on('drawcancel', () => {
@@ -548,7 +548,7 @@ class ShapeCreatorView {
 
             this._createButton.text("Stop Creation");
             document.oncontextmenu = () => false;
-            this._create(model.usingShortkey);
+            this._create();
         }
         else {
             this._removeAim();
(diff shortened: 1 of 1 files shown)


cvat/apps/engine/static/engine/js/shapeCreator.js
```diff
@@ -517,34 +533,27 @@ class ShapeCreatorView {
...
-                        });
-                    }
-                });
             }
 
             this._createButton.text("Stop Creation");
             document.oncontextmenu = () => false;
-            this._create();
+            this._create(model._usingShortkey);
+            if (!model._usingShortkey) {
```
bsekachev: Why do you need this block of code?
There is exactly the same code a couple of lines above?

cvat/apps/engine/static/engine/js/shapeCreator.js
```diff
@@ -517,34 +533,27 @@ class ShapeCreatorView {
...
-                            x2: this._aimCoord.x,
-                        });
-                    }
-                });
             }
 
             this._createButton.text("Stop Creation");
             document.oncontextmenu = () => false;
-            this._create();
+            this._create(model._usingShortkey);
```
bsekachev: Please, write a getter for the field

cvat/apps/engine/static/engine/js/shapeCreator.js
```diff
@@ -146,15 +147,15 @@ class ShapeCreatorController {
...
+                this.switchCreateMode(false, true);
             }.bind(this));
 
             Mousetrap.bind(shortkeys["switch_draw_mode"].value, switchDrawHandler.bind(this), 'keydown');
         }
     }
 
-    switchCreateMode(force) {
-        this._model.switchCreateMode(force);
+    switchCreateMode(force, usingShortkey) {
```
bsekachev: What about default value? This method is called with a single arguments in many places

cvat/apps/engine/static/engine/js/shapeCreator.js
```diff
@@ -425,7 +441,7 @@ class ShapeCreatorView {
                     this._controller.finish(box, this._type);
                 }
 
-                this._controller.switchCreateMode(true);
+                this._controller.switchCreateMode(true, usingShortkey);
```
bsekachev: Always false is suitable here, I assume. 
And you do not need an additional argument :)

bsekachev:
@mmarkelov 

Hi again, 

Nice PR. Only minor comments.
But I wouldn't spend time improving legacy JS code.
It will be fully deleted soon :)

mmarkelov:
@bsekachev done. And I hope that it will be rewrite to React ecosystem?)

bsekachev:
@mmarkelov 
Will be :)

PR #944, opened 2019-12-12
[WIP] Add about cvat decription and version number

LiSa20120:
Added section for version number 

before clicking 
before

after clicking
after

### DIFF
--- a/cvat-ui/src/components/header/header.tsx
+++ b/cvat-ui/src/components/header/header.tsx
@@ -7,13 +7,16 @@ import {
     Layout,
     Icon,
     Button,
+    Popover,
     Menu,
 } from 'antd';
 
 import Text from 'antd/lib/typography/Text';
 
 import getCore from '../../core';
-
+import axios from 'axios';
+import { render } from 'react-dom';
+import ReactDOM from 'react-dom';
 const core = getCore();
 const serverHost = core.config.backendAPI.slice(0, -7);
 
@@ -31,6 +34,8 @@ type Props = HeaderContainerProps & RouteComponentProps;
 
 const cvatLogo = (): JSX.Element => <img alt='' src='/assets/cvat-logo.svg' />;
 const userLogo = (): JSX.Element => <img alt='' src='/assets/icon-account.svg' />;
+const text = React.createElement('div', {id : 'about'});
+const content = React.createElement('div', {id : 'version'});
 
 function HeaderContainer(props: Props): JSX.Element {
     const {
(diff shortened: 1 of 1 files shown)


cvat-ui/src/components/header/header.tsx
```diff
@@ -62,6 +67,34 @@ function HeaderContainer(props: Props): JSX.Element {
...
+                <Popover placement="bottom" title={content} content={text} trigger="click" overlayStyle={{width : "50vw"}}>
+                <Button
+                    className='cvat-header-button'
+                    type='link'
+                    value='tasks'
+                    onClick={
+                        (): void => {
+                        const t = localStorage.getItem('token');
+                        const token ='Token '+ t.replace(/['"]+/g, '')
+                        const data = axios.get(`${serverHost}/api/v1/server/about`,{ headers: {Authorization : token}}).then(
```
bsekachev: @LiSa20120 

We shouldn't make server requests here. 
There are two problems:
1. To be consistent and do not do extra requests, we should use ``redux`` here (create reducer [for example ServerInfoReducer], add it to CVAT storage, add some actions for the reducer) and provide some props from the storage to the component. Look for example how we get users (usersReducer, usersActions) in such a scheme. 
2. You do not need to do request manually with axios and add an authorization token. Just use cvat-core to get this data:
```js
const core = getCore();
core.server.about()
```

cvat-ui/src/components/header/header.tsx
```diff
@@ -62,6 +67,34 @@ function HeaderContainer(props: Props): JSX.Element {
                 >
                     Tasks
                 </Button>
+                <Popover placement="bottom" title={content} content={text} trigger="click" overlayStyle={{width : "50vw"}}>
```
bsekachev: I agree with you about it should be a switchable window, but this component doesn't look pretty.
I believe a Modal window would be great here.
You know, a tool name, an image, description, a version, the github link, etc.

cvat-ui/src/components/header/header.tsx
```diff
@@ -62,6 +67,34 @@ function HeaderContainer(props: Props): JSX.Element {
...
+                            const version = 'cvat-version : ' + val.version;
+                            ReactDOM.render(version,document.getElementById('version'))
+                            ReactDOM.render(val.description,document.getElementById('about'))
+                            
+                        });
+                        
+                    }
+                    }
+                >
+                    About
```
bsekachev: I believe the best place to put this button is menu in header (near with logout button).

bsekachev:
@LiSa20120 

Hi, good work. 
Nevertheless we have some comments about the solution.

bsekachev:
@LiSa20120 

A couple additional comments. 
Container cannot be built because of the error: ``Module not found: Error: Can't resolve 'axios' in '/tmp/cvat-ui/src/components/header'``
And please, look at issues which have been found by Codacy

nmanovic:
@LiSa20120 , please look at the PR from Boris: https://github.com/opencv/cvat/pull/963. It has `about` button. Just need to implement it. If you are not going to work on the PR please close it.

image

LiSa20120:
@bsekachev Thank you. 
I would like to work on it.

nmanovic:
Hi @LiSa20120 , I will close the PR for now. Please don't hesitate to reopen it when you are ready.

cvat-ui/src/components/header/header.tsx
LiSa20120: Hi, @bsekachev  
What image do you want to include here? 
As of now, I have the tool name, version number and description (in the same order). Also, I believe adding a github link would make it redundant as it already appears on the header. Kindly let me know about these. Thanks :)

bsekachev: Hi, @LiSa20120 

To be honest, I don't know what image can we use here. Probably it's better to skip this moment for now. 

I have some ideas about content of the dialog window:
- The name of the tool
- Tool description
- Server version (can get via core.server.about()) 
- Core version (can get via core.client.version)
- Release notes ("what's new") (just a link to the section of the ```CHANGELOG.md``` file)
- License information (a link to ```LICENSE``` file in the repository)
- "Need help?" (a link to the gitter chat)
- Maybe link to [Forum on Intel Developer Zone](https://software.intel.com/en-us/forums/intel-distribution-of-openvino-toolkit)

@nmanovic, what do you think?

LiSa20120: about
@bsekachev Have a look at it.

LiSa20120: @bsekachev How are you converting md files to html?

LiSa20120: @bsekachev How are you converting the md files to html? (this is to include CHANGELOG.md)
Should we install [react-markdown](https://www.npmjs.com/package/markdown-react-js)?

PR #952, opened 2019-12-13
[WIP] OpenCV.js plugin

dkurt:
Hi!

In this PR you may find OpenCV's `equalizeHist`. Note: default image quality is 70% (Jpeg compression) which produces artifacts. Current way to apply it correctly is to set 100% image quality at the new task creation stage.

origin | equalizeHist over 70% | equalizeHist over 100%
----------|--------------|------------
 |  | 

OpenCV.js: [opencv.js.zip](https://github.com/opencv/cvat/files/3960855/opencv.js.zip)

TODO:
- [ ] Keep OpenCV transformations in case of other changes: rotation, scroll

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -163,6 +163,9 @@ RUN sed -r "s/^(.*)#.*$/\1/g" ${HOME}/datumaro/requirements.txt | xargs -n 1 -L 
 RUN patch --binary -p1 < ${HOME}/cvat/apps/engine/static/engine/js/3rdparty.patch
 RUN chown -R ${USER}:${USER} .
 
+# Download OpenCV.js
+RUN wget https://docs.opencv.org/master/opencv.js -O ${HOME}/cvat/apps/engine/static/engine/js/3rdparty/opencv.js
+
 # RUN all commands below as 'django' user
 USER ${USER}
 
--- a/cvat/apps/engine/static/engine/js/annotationUI.js
+++ b/cvat/apps/engine/static/engine/js/annotationUI.js
@@ -176,34 +176,7 @@ function setupFrameFilters() {
     });
 
     equalize.on('click', () => {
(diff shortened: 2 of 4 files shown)


nmanovic:
@dkurt , I cannot see opencv.js in the PR.

dkurt:
@nmanovic, Yes, there is no OpenCV.js there because if there will new new methods it's required to upload it again that means all the previous versions of OpenCV.js will still live in the git history.

Is there an option to download it during `docker-compose build` ?

dkurt:
@nmanovic, Uploaded opencv.js as npm package: https://www.npmjs.com/package/opencv_js_cvat

Probably this is not the best solution, but to download it I used https://unpkg.com/

Dockerfile
```diff
@@ -163,6 +163,9 @@ RUN sed -r "s/^(.*)#.*$/\1/g" ${HOME}/datumaro/requirements.txt | xargs -n 1 -L
 RUN patch --binary -p1 < ${HOME}/cvat/apps/engine/static/engine/js/3rdparty.patch
 RUN chown -R ${USER}:${USER} .
 
+# Download OpenCV.js
+RUN wget https://docs.opencv.org/master/opencv.js -O ${HOME}/cvat/apps/engine/static/engine/js/3rdparty/opencv.js
```
nmanovic: It is not a complete solution. For development locally we don't use docker. Thus it will be necessary to download/update the library manually. It is painful.

dkurt: Can you please share a docker free approach to use CVAT so I can try it locally?

nmanovic: https://github.com/opencv/cvat/blob/develop/CONTRIBUTING.md

nmanovic:
@dkurt , I will close the PR. It looks like nobody is going to finish it.

PR #971, opened 2019-12-17
CSS Enhancement

bsekachev:
- Adding vendor prefixes automatically
- Polyfills for css with ``postcss-preset-env``
- .scss instead of .css
- Reading antd .less styles instead of compiled .css (allow to change theme, for example. In theory less bundle size)


cvat-ui/webpack.config.js
```diff
@@ -39,9 +39,7 @@ module.exports = {
                     }]],
                     presets: [
                         ['@babel/preset-env', {
-                            targets: {
-                                chrome: 58,
-                            },
+                            targets: '> 3%', // https://github.com/browserslist/browserslist
```
nmanovic: Probably it is better to have the same strategy across all tools (e.g. > 2%)

Issue #937, opened 2019-12-11
using datumaro to convert voc annotations to cvat


verasativa:
I'm trying to figure out how datumaro works, and eventually how to integrate to our dvc workflow. But I'm starting with a simple task: convert voc annotations to cvat format.
I'm creating a datumaro dataset with:
`pythonw -m datumaro create -d data/datumaro_dataset`
then adding the one source with:
`pythonw -m datumaro source import -f voc -n v1 -p data/datumaro_dataset/ dir data/Workers/version_1/Annotations`
then from the `data/datumaro_dataset`dir:
`pythonw -m datumaro source export -n datumaro_dataset -d ../export -f cvat`
but I'm getting a:
`KeyError: 'Key "datumaro_dataset" does not exist` or a:
`FileExistsError: [Errno 17] File exists: '/Users/vera/repositories/workers-detector/data/export'`

Questions:
What will be the proper way to do this conversion?
I'm interested on contributing to datumaro; what will be the best communication channel with developers? (no really lucky on gitter)

cc: @zhiltsov-max

zhiltsov-max:
@verasativa, currently, CVAT format support is not implemented in Datumaro, we are going to do it soon. Once it is supported, to export a dataset you would do:
```
datum project import -d data/datumaro_dataset -f voc data/Workers/version_1/Annotations
datum project export -p data/datumaro_dataset -f cvat ../export
```

Your variant with `source`s is also OK, but it should look like this:
```
pythonw -m datumaro create -d data/datumaro_dataset
pythonw -m datumaro source import -f voc -n v1 -p data/datumaro_dataset/ dir data/Workers/version_1/Annotations
pythonw -m datumaro source export -n v1 -d ../export -f cvat # note v1 here as source name
```
`KeyError` is for non-existing project source which was named `v1`, not `datumaro_dataset`.

Regarding `FileExistsError`: currently we avoid overwriting existing dirs to prevent data loss. We're open for your suggestions on design.

PR #974, opened 2019-12-18
[Datumaro] CVAT format import

zhiltsov-max:
Added:
- CVAT extractor
- Additional logs in CLI
- Missing `yolo` importer to CLI

Fixes:
- Regularized importer and extractor logic
- Refactored format tests
- Refactored importers, `Datumaro` and `COCO` extractor
- Fixed the `project import` issue with `--overwrite` option, in which if the target dir had project data it would be imported into the new project. Resolved by cleaning of the existing dir.

Related: #937

### DIFF
--- a/datumaro/datumaro/components/formats/cvat.py
+++ b/datumaro/datumaro/components/formats/cvat.py
@@ -1,0 +1,10 @@
+
+# Copyright (C) 2019 Intel Corporation
+#
+# SPDX-License-Identifier: MIT
+
+class CvatPath:
+    IMAGES_DIR = 'images'
+    ANNOTATIONS_DIR = 'annotations'
+
+    IMAGE_EXT = '.jpg'
--- a/datumaro/tests/test_cvat_format.py
+++ b/datumaro/tests/test_cvat_format.py
@@ -101,7 +101,7 @@ class CvatExtractorTest(TestCase):
     def test_can_load(self):
         class TestExtractor(Extractor):
             def __iter__(self):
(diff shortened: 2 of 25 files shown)


datumaro/tests/test_cvat_format.py
```diff
@@ -0,0 +1,148 @@
...
+class CvatExtractorTest(TestCase):
+    @staticmethod
+    def generate_dummy_cvat(path):
+        images_dir = osp.join(path, Cvat.CvatPath.IMAGES_DIR)
+        anno_dir = osp.join(path, Cvat.CvatPath.ANNOTATIONS_DIR)
+
+        os.makedirs(images_dir)
+        os.makedirs(anno_dir)
+
+        root_elem = ET.Element('annotations')
```
nmanovic: @zhiltsov-max , I will vote for a readable form and parse xml from a string. Why does it necessary to construct cvat xml this way?

datumaro/datumaro/components/formats/cvat.py
```diff
@@ -0,0 +1,10 @@
+
+# Copyright (C) 2019 Intel Corporation
+#
+# SPDX-License-Identifier: MIT
+
+class CvatPath:
+    IMAGES_DIR = 'images'
+    ANNOTATIONS_DIR = 'annotations'
+
+    IMAGE_EXT = '.jpg'
```
nmanovic: Does it mean that we cannot work with a dataset with png images?

PR #976, opened 2019-12-18
Change-submit-button-style

mmarkelov:
Small change to improve UI/UX, I suppose **Submit** button mustn't be in danger color

### DIFF
--- a/cvat-ui/src/components/create-task-page/create-task-content.tsx
+++ b/cvat-ui/src/components/create-task-page/create-task-content.tsx
@@ -242,7 +242,7 @@ export default class CreateTaskContent extends React.PureComponent<Props, State>
                     <Button
                         loading={loading}
                         disabled={loading}
-                        type="primary"
+                        type='primary'
                         onClick={this.handleSubmitClick}
                     >
                         Submit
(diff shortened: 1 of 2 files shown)


cvat-ui/src/components/create-task-page/create-task-content.tsx
```diff
@@ -242,7 +242,6 @@ export default class CreateTaskContent extends React.PureComponent<Props, State>
                     <Button
                         loading={loading}
                         disabled={loading}
-                        type='danger'
```
bsekachev: What about setting it as a 'primary'?

And we have the same situation in upload a model form. So, to be consistent we should change button type there too.

cvat-ui/src/components/create-task-page/create-task-content.tsx
```diff
@@ -242,6 +242,7 @@ export default class CreateTaskContent extends React.PureComponent<Props, State>
                     <Button
                         loading={loading}
                         disabled={loading}
+                        type="primary"
```
bsekachev: Please, use single quotes
There will be linter error otherwise

PR #984, opened 2019-12-19
Improve some array loop methods

mmarkelov:
There are some places with some unnecessary looping throw arrays, so I think it can be reduced

### DIFF
--- a/cvat-ui/src/actions/models-actions.ts
+++ b/cvat-ui/src/actions/models-actions.ts
@@ -485,9 +485,10 @@ ThunkAction<Promise<void>, {}, {}, AnyAction> {
                 );
 
                 parse(response.run)
-                    .filter((inferenceMeta: InferenceMeta): boolean => inferenceMeta.active)
                     .forEach((inferenceMeta: InferenceMeta): void => {
-                        subscribe('auto_annotation/check', inferenceMeta, dispatch);
+                        if (inferenceMeta.active) {
+                            subscribe('auto_annotation/check', inferenceMeta, dispatch);
+                        }
                     });
             }
 
@@ -503,9 +504,10 @@ ThunkAction<Promise<void>, {}, {}, AnyAction> {
                 );
 
                 parse(response)
-                    .filter((inferenceMeta: InferenceMeta): boolean => inferenceMeta.active)
                     .forEach((inferenceMeta: InferenceMeta): void => {
-                        subscribe('tensorflow/annotation/check/task', inferenceMeta, dispatch);
+                        if (inferenceMeta.active) {
+                            subscribe('tensorflow/annotation/check/task', inferenceMeta, dispatch);
+                        }
                     });
             }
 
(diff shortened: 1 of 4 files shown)


cvat-ui/src/actions/models-actions.ts
```diff
@@ -485,9 +485,10 @@ ThunkAction<Promise<void>, {}, {}, AnyAction> {
                 );
 
                 parse(response.run)
-                    .filter((inferenceMeta: InferenceMeta): boolean => inferenceMeta.active)
                     .forEach((inferenceMeta: InferenceMeta): void => {
-                        subscribe('auto_annotation/check', inferenceMeta, dispatch);
+                        if (inferenceMeta.active) {
```
bsekachev: I consider the previous version more 'functional programming style'.
Don't think you so?

cvat-ui/src/components/models-page/models-page.tsx
```diff
@@ -43,21 +56,20 @@ export default function ModelsPageComponent(props: Props): JSX.Element {
...
-                && <BuiltModelsList models={integratedModels} />
-            }
             { !!uploadedModels.length && (
-                <UploadedModelsList
-                    registeredUsers={registeredUsers}
-                    models={uploadedModels}
-                    deleteModel={deleteModel}
-                />
+                <>
+                    <BuiltModelsList models={integratedModels} />
```
bsekachev: We do not need to render this component if ``integratedModels.length === 0``

cvat-ui/src/components/models-page/models-page.tsx
```diff
@@ -43,21 +56,20 @@ export default function ModelsPageComponent(props: Props): JSX.Element {
         );
     }
 
-    const uploadedModels = models.filter((model): boolean => model.id !== null);
-    const integratedModels = models.filter((model): boolean => model.id === null);
+    const [uploadedModels, integratedModels] = getModels(models);
```
bsekachev: I would prefer two lines instead of extra function

cvat-ui/src/components/labels-editor/labels-editor.tsx
```diff
@@ -80,26 +95,17 @@ export default class LabelsEditor
...
-        const savedLabels = [];
-
-        for (const label of labels) {
-            if (label.id >= 0) {
-                savedLabels.push(label);
-            } else {
-                unsavedLabels.push(label);
-            }
-        }
+        const [savedLabels, unsavedLabels] = getFilteredLabels(labels)
```
bsekachev: Why not? 

```js
const savedLabels = labels.filter((label) => label.id >= 0)
const unsavedLabels = labels.filter((label) => label.id < 0)
```

cvat-ui/src/components/models-page/models-page.tsx
mmarkelov: It will render if only **integratedModels.length** will be grater than 0

cvat-ui/src/components/labels-editor/labels-editor.tsx
mmarkelov: Cause you loop through **labels** two times instead of one. This is just about performance

cvat-ui/src/components/models-page/models-page.tsx
bsekachev: But condition is ``uploadedModels.length``, not ``integratedModels.length``

cvat-ui/src/components/labels-editor/labels-editor.tsx
bsekachev: @mmarkelov 

I believe in this place performance doesn't matter just because it isn't a bottleneck.

cvat-ui/src/actions/models-actions.ts
mmarkelov: That was all about performance. But if you are sure that it is OK, it can be closed.

PR #985, opened 2019-12-19
UI Enhancements 

bsekachev:
* A little bit improved annotation view
* Added two pages (task not found, job not found) in case of bad URL parameters or access reject
* base.scss now contains only vars (a lot of extra classes were defined before)
* A bug fixed with browser 100% CPU load when go to a task or a job using a text value instead of a number value in URL

### DIFF
--- a/cvat-ui/src/containers/task-page/task-page.tsx
+++ b/cvat-ui/src/containers/task-page/task-page.tsx
@@ -14,23 +14,29 @@ import {
 type Props = RouteComponentProps<{id: string}>;
 
 interface StateToProps {
-    task: Task | null;
+    task: Task | null | undefined;
     fetching: boolean;
     deleteActivity: boolean | null;
     installedGit: boolean;
 }
 
 interface DispatchToProps {
-    onFetchTask: (tid: number) => void;
+    getTask: () => void;
 }
 
 function mapStateToProps(state: CombinedState, own: Props): StateToProps {
     const { plugins } = state.plugins;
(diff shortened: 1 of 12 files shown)


cvat-ui/src/containers/task-page/task-page.tsx
```diff
@@ -14,23 +14,29 @@ import {
...
-    const { deletes } = state.tasks.activities;
+    const { tasks } = state;
+    const { gettingQuery } = tasks;
+    const { deletes } = tasks.activities;
+
     const id = +own.match.params.id;
 
-    const filtered = state.tasks.current.filter((task) => task.instance.id === id);
-    const task = filtered[0] || null;
+    const filteredTasks = state.tasks.current
```
mmarkelov: it can be just **tasks.current**

cvat-ui/src/containers/task-page/task-page.tsx
```diff
@@ -14,23 +14,29 @@ import {
...
     installedGit: boolean;
 }
 
 interface DispatchToProps {
-    onFetchTask: (tid: number) => void;
+    getTask: () => void;
 }
 
 function mapStateToProps(state: CombinedState, own: Props): StateToProps {
     const { plugins } = state.plugins;
```
mmarkelov: `state.plugins.plugins` get me confused

bsekachev: Yep, a little bit

PR #997, opened 2019-12-23
Automatic bordering feature during drawing/editing

bsekachev:


nmanovic:
Let's fix the following minor issues in the next PR:
- Change its name: "Common border"
- It should be visible only when drawing points (polygon, polyline, points)
- Don't reset it by default
- Let's add the common border only if three points are added in the row. If the user clicks outside the common border I believe we need to add a regular point.

Issue #949, opened 2019-12-13
Issue with hanging web application in browser when annotations save is called.


vskubriev:
Specifically, with the current task, we had a problem - the interface freezes when saving a frame (frames through autosave). The server is powerful enough and its la is very low. Therefore, I believe that the problem is precisely in the client part of the markup executable in the browser. Apparently, the interface freezes at the time of autosave or manual saving by `Ctrl+S`. The entire task contains 3.5 thousand frames. Frames can contain up to 80 objects of type polyline. The entire task almost completely marked.

It can also be noted that:

1. Previously (when marking up images in this task just started): the open menu button and autosave function did not cause web ui hanging.
2. Also, another task with a small fraction of similarly marked images works without problems. 
3. First pre-main page with LOADING shows a long time compared to other tasks.wed 
4. Saving in another tasks with bounding boxes and smaller frames is running without hanging.

How can I help myself in solving/analyze this problem?

vskubriev:
![image](https://user-images.githubusercontent.com/4639077/70801020-bff94e80-1dbe-11ea-9718-f4747691971e.png)

vskubriev:
![image](https://user-images.githubusercontent.com/4639077/70801057-dbfcf000-1dbe-11ea-92d1-398dbee0e6f0.png)

vskubriev:
![image](https://user-images.githubusercontent.com/4639077/70801478-13b86780-1dc0-11ea-9058-f42f2fa760af.png)

vskubriev:
When is beta 1.0 roughly planned?

nmanovic:
@vskubriev , we will look in the problem. Could you please provide some parameters of your task? How many annotations do you have and its type? Please post label specification?

nmanovic:
@vskubriev , we will try to fix the issue till next release but cannot promise.

nmanovic:
@vskubriev , you are welcome to contribute!

vskubriev:
> @vskubriev , we will look in the problem. Could you please provide some parameters of your task? How many annotations do you have and its type? Please post label specification?

Task is about 3.5k images with 1920*1200 pixels. Most of images are marked up.

What do you mean by "its type" - type of objects is polylines.
What do you mean by "label specification" - Where I can find it?

vskubriev:
Latest info from one of out markupers:

When he switched to marking up earlier frames (almost at the very beginning), saving became much faster in the same folder.

bsekachev:
@vskubriev 

[...]

PR #1000, opened 2019-12-24
Couple of fixes in "save annotations"

bsekachev:
Related with #949

### DIFF
--- a/cvat-core/src/annotations-saver.js
+++ b/cvat-core/src/annotations-saver.js
@@ -102,12 +102,18 @@
                 },
             };
 
+            const keys = ['id', 'label_id', 'group', 'frame',
+                'occluded', 'z_order', 'points', 'type', 'shapes',
+                'attributes', 'value', 'spec_id', 'outside'];
+
             // Find created and updated objects
             for (const type of Object.keys(exported)) {
                 for (const object of exported[type]) {
                     if (object.id in this.initialObjects[type]) {
-                        const exportedHash = JSON.stringify(object);
-                        const initialHash = JSON.stringify(this.initialObjects[type][object.id]);
+                        const exportedHash = JSON.stringify(object, keys);
+                        const initialHash = JSON.stringify(
+                            this.initialObjects[type][object.id], keys,
+                        );
                         if (exportedHash !== initialHash) {
                             splitted.updated[type].push(object);
                         }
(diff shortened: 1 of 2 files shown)


cvat-core/src/annotations-saver.js
```diff
@@ -102,12 +102,18 @@
                 },
             };
 
+            const keys = ['id', 'label_id', 'group', 'frame',
```
nmanovic: Let's create a function to covert an object into "hash". Thus you don't need to repeat keys in several files.

cvat/apps/engine/static/engine/js/annotationSaver.js
```diff
@@ -241,20 +245,33 @@ class AnnotationSaverModel extends Listener {
...
             } else {
                 const [created, updated, deleted] = this._split(exported);
                 this.notify('saveCreated');
                 const savedCreated = await this._create(created);
                 this._updateCreatedObjects(created, savedCreated, mapping);
                 this._version = savedCreated.version;
+
                 for (const type of Object.keys(this._initialObjects)) {
                     for (const object of savedCreated[type]) {
+                        if (object.shapes) {
```
nmanovic: Does it make sense to create a special function here and use it 3 times?

cvat/apps/engine/static/engine/js/annotationSaver.js
```diff
@@ -241,20 +245,33 @@ class AnnotationSaverModel extends Listener {
                 this._shapeCollection.flush = false;
                 this._version = savedObjects.version;
                 this._resetState();
+
                 for (const type of Object.keys(this._initialObjects)) {
                     for (const object of savedObjects[type]) {
+                        if (object.shapes) {
```
nmanovic: Could you please add a comment why it is important to delete shape.id here?

cvat-core/src/annotations-saver.js
bsekachev: @nmanovic 
The first file from legacy UI,
The second from cvat-core.

They don't related anyhow.

cvat/apps/engine/static/engine/js/annotationSaver.js
bsekachev: Probably, it does

nmanovic: I missed that it is from legacy UI. Let's keep it as is for now.

PR #1001, opened 2019-12-24
Change Modal submit button okType

mmarkelov:


mmarkelov:
This was not fixed in #976

PR #1008, opened 2019-12-25
[Datumaro] Instance polygon-mask conversions in COCO format

zhiltsov-max:
Added:
- Polygons to mask conversions and vise-versa in COCO format
- Covered polygons or masks can be cropped to only visible part

Test:
```
datum project export -f coco -d test_dir -- -h

--segmentation-mode {guess,polygons,mask} 
    Save mode for instance segmentation (default: guess): 
   - 'guess': guess the mode for each instance, use 'is_crowd' attribute as hint; 
   - 'polygons': save polygons, merge and convert masks, prefer polygons; 
   - 'mask': save masks, merge and convert polygons, prefer masks; 

--crop-covered        
    Crop covered segments so that background objects' segmentation was more accurate (default: False)
```

The algorithm of instance detection in image annotations:
1. Find relevant annotations - labeled masks (which are considered instance masks), bboxes and polygons
1. Group annotations by `group` attribute, call these groups as "instances"
1. Find group bbox - it will be instance bbox
1. Find largest element in the group by area, call it "leader object", i.e. the instance we are trying to annotate
1. Merge masks or polygons in the group into one segment
1. (optional) Crop background segments


nmanovic:
@zhiltsov-max , could you please resolve conflicts?

zhiltsov-max:
@nmanovic, done

Issue #1011, opened 2019-12-27
Remote access to CVAT, unable to connect remotely via LAN or internet domain (via <HOST-IP>:port or <FQDN>:port). Blocked by CORS policy: No 'Access-Control-Allow-Origin, Network Error and net::ERR CONNECTION REFUSED errors.


pvalls:
I deployed CVAT on an AWS instance and I struggled to access it remotely. 

The errors I encountered along the way were:
> Could not check authorization on the server. Error: Network Error.

> blocked by CORS policy: No 'Access-Control-Allow-Origin'

> net::ERR_CONNECTION_REFUSED

Nevertheless, finally I managed to get it working. Following I report the solutions to this errors so it can help others and to prepare a PR to contribute.

Gathering of related issues: <br>
* [Issue 930](https://github.com/opencv/cvat/issues/930): Unable to remote connect #930

* [Issue 907](https://github.com/opencv/cvat/issues/907): Remote access  issue #907

* [Issue 165](https://github.com/opencv/cvat/issues/165): How can I connect from other computers which are NOT on the LAN? #165 
 
* [Issue 102](https://github.com/opencv/cvat/issues/102):Connection from another computer gives Bad Request(400) #102
  
* [Issue 4](https://github.com/opencv/cvat/issues/4): Running on AWS EC2 #4
 
* [Issue 128](https://github.com/opencv/cvat/issues/128): Access CVAT remotely via \<HOST_IP>:port or \<FQDN>:port? #128

### **Solution:**
The following changes did the trick for me:

1. Add following line to _react_nginx.conf_: 
`add_header Access-Control-Allow-Origin "*";`

A *react_nginx.conf* example would be:

	
	server {
	    root /usr/share/nginx/html;
	    # Any route that doesn't have a file extension (e.g. /devices)
	    
	    location / {
	        try_files $uri $uri/ /index.html;
	        add_header Access-Control-Allow-Origin "*";
	    }
	}
	

2. **To *docker-compose.override.yml*, add:** <br>
 `ALLOWED_HOSTS: '*'` <br>
 `UI_HOST: mysite.com`<br>
` REACT_APP_API_HOST: mysite.com`
 
 A *docker-compose.override.yml* example would be:

	
	version: "2.3"
	
	services:
	  cvat:
	    environment:
	      ALLOWED_HOSTS: '*'
	      UI_HOST: mysite.com
	    ports:
	      - "80:8080"

	  cvat_ui:
	    build:
              args:
[...]

PR #1012, opened 2019-12-27
fix remote access issues (#1011)

pvalls:
fix for the following remote access errors:

> Could not check authorization on the server. Error: Network Error.

> blocked by CORS policy: No 'Access-Control-Allow-Origin'

> net::ERR_CONNECTION_REFUSED

when UI_URL is in the form \<FQDN>:80 by adding new variable `UI_URL_WITHOUT_PORT` to `CORS_ORIGIN_WHITELIST`

The reason for this is that, in my edge case,  when accessing a CVAT AWS instance, after adding `ALLOWED_HOSTS: '*'`,  `UI_HOST: mysite.com`, ` REACT_APP_API_HOST: mysite.com`
to my *docker-compose.override.yml* and adding `add_header Access-Control-Allow-Origin "*";` to _react_nginx.conf_, the errors were still there and the proposed change to `CORS_ORIGIN_WHITELIST` proved to be necessary.

I believe that in the `cvat/settings/base.py` file,  adding the new variable `UI_URL_WITHOUT_PORT` to CORS_ORIGIN_WHITELIST proved to solve the problem because when the port number is :80 and the `UI_URL` is not in the LAN (\<HOST-IP>:port), but instead it is a Fully Qualified Domain Name (\<FQDN>:port), the port 80 is redundant (UI_URL == http://mydomain.com:80) and the errors arise.

This could be changed to a more elegant or pythonic solution. I also wonder, should this PR include the step of adding the line `add_header Access-Control-Allow-Origin "*"` ; to _react_nginx.conf_ ? I am not sure if it is necessary to do this.


cvat/settings/base.py
```diff
@@ -192,10 +192,15 @@ def generate_ssh_keys():
...
+UI_URL_WITHOUT_PORT = UI_URL
+
+
 if len(UI_URL):
     UI_URL += ':{}'.format(UI_PORT)
 
-CORS_ORIGIN_WHITELIST = [UI_URL]
+# Add UI_URL_WITHOUT_PORT to avoid CORS Policy, ERR_CONNECTION_REFUSED
+# errors when UI_URL is <FQDN>:80
+CORS_ORIGIN_WHITELIST = [UI_URL, UI_URL_WITHOUT_PORT]
```
nmanovic: Do you think we should have `if UI_PORT == 80` here?

nmanovic:
@pvalls , it seems something wrong with your PR. It contains many changes which you didn't do.

cvat/settings/base.py
pvalls: Yes, it would be cleaner.  Now I'm deploying again with the if statement and testing the issue to assert that the problem is still fixed. Otherwise, it doesn't hurt to always have the UI_URL_WITHOUT_PORT variable in the CORS_ORIGIN_WHITELIST.

pvalls:
> @pvalls , it seems something wrong with your PR. It contains many changes which you didn't do.

My fault, I did a git pull from upstream develop. It is back to normal now, only my change appears.

cvat/settings/base.py
pvalls: I believe that the UI_URL should be added to the CORS_ORIGIN_WHITELIST with and without the port number regardless of which port it is, because the application port can be 8080 or any other even when connecting via FQDN. I'd say this PR fix is suitable to be accepted

nmanovic:
@pvalls , one more time thanks for the PR. We will fix the issue in another way. It is better to provide a proxy server (in a docker container) to solve the CORS issue. Thus users will communicate only with the proxy server and the proxy will redirect requests to server, ui, db, Redis, ...

I will close the PR for now. Once again thanks for your time.

PR #1039, opened 2020-01-10
Added Support for Multi-GPU Inference and TF-Models in Model-Manager

san999p:
We, at Onepanel, have implemented few new features for CVAT to make it more easy and faster to annotate images. 

- This PR adds two features as described below:

1) Currently, you can only run inference on single GPU when you are running Run-TF-Annotation.
   Generally, this is achieved by running inference through Tensorflow-Serving. But we were
   not using that so we have tried to achieve the same by using threading and distributing 
   images across multiple gpus. We tried to allocate GPUs manually using device placement, but it 
   didn't work somehow. Please review these changes.

2) As of now, you can use the default Tensorflow model only for Run-TF-Annotation. We can
   allow users to use any custom model by allowing them to upload model file and classes file.
   We can achieve the same by adding UI components to Model-Manager. This update will allow
   users to upload tensorflow models through model-manager. They can select whether its
   openvino model or tensorflow model. Once they click on Run-TF-Annotation, the UI will show
   all available models with TF-Annotation tag just like Run OpenVINO Model. These changes are live 
  on Onepanel, you can check it out here: https://c.onepanel.io/onepanel-demo/projects/cvat-public- demo

We have also added a new feature on Onepanel which allow users to train model on annotated data.
This can drastically reduce the annotation time as you can annotate some images, train a model on it,
and use it to pre-annotate the rest of the images. This feature is live on Onepanel, you can take a look. Once you annotate images, you will find a Create Annotation Model button on the dashboard, where you can select the model and other parameters. 
It uses some features of Onepanel (i.e Jobs) so we can't directly push it here. Let us know
if you are interested in adding it here.

CC @aleksandrmelnikov


nmanovic:
@san999p , @aleksandrmelnikov ,

thanks for the PR. Will the patch work with the latest develop branch (ReactJS + Antd)?

san999p:
@nmanovic I believe it should. We have been working on slightly older version of Develop branch. However, I think it will work. Can you give it a try and let us know? Thanks

nmanovic:
@san999p , did you try to build it with the latest develop branch? Our team really likes contributions but we expect that if somebody submits a PR it works at least.

Please attach a couple of screenshots with the new API. I believe the patch should not work just because it uses functionality in dashboard which doesn't exist anymore. We use cvat-core component now to communicate with the server.

aleksandrmelnikov:
Ahhh okay, thanks @nmanovic . Savan and I will do some testing and get back to you.

nmanovic:
@san999p , @aleksandrmelnikov , are you going to change the PR?

san999p:
@nmanovic Onepanel runs on slightly older version of CVAT. We will be updating cvat in near future. But since I don't have experience with React, it will take some time. 
Another option is, if someone can help me with the front-end/UI, then I can adjust backend/python code according to the newer version. I think there are very few UI changes in this PR, so it should be quick.

san999p:
@nmanovic We will start working on this new UI sometime in next week. So, it will take few days before we can send a new PR. I am closing this PR and send a new one with the updates. Thanks

PR #1046, opened 2020-01-13
Removed VIM package (it isn't necessary)

nmanovic:


bsekachev:
But sometimes it can be useful

nmanovic:
It is useful only for development. We will be able to do `apt-get update && apt-get install vim` if it is necessary.

PR #1051, opened 2020-01-14
Added ability to match many model labels to one task labels

### LINKED ISSUE #1049: Suggestion: tools
I am using the CVAT tool, it's great.
It would be great if you can improve two things:
1- Auto annotation: map model label to different Openvino labels
It means that we might have different type of categories of the same objects but the model recognize it as one object.

2- After creating a task it's impossible to remove some labels from it. I need to create another task with a new defined labels.

Thanks

bsekachev:
- Also fixed small bug with a share browsing

Resolved #1049

### DIFF
--- a/cvat-ui/src/components/model-runner-modal/model-runner-modal.tsx
+++ b/cvat-ui/src/components/model-runner-modal/model-runner-modal.tsx
@@ -121,7 +121,7 @@ export default class ModelRunnerModalComponent extends React.PureComponent<Props
             if (!selectedModelInstance.primary) {
                 if (!selectedModelInstance.labels.length) {
                     notification.warning({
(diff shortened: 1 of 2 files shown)


test4fest:
Thanks for openenigs the this task. I am a little bit confused. should the title be match one model labels to many task labels?

bsekachev:
@test4fest 

You can reckon do many to one mapping only in one direction: multiple model labels -> one task label.

Example:
Model labels: ``car``, ``bike``.
Task labels: ``vehicle`` 
Mapping: ``car -> vehicle``, ``bike -> vehicle``

But if a model can recognize, for example ``vehicle``, you cannot match it as both ``car`` and ``bike`` simultaneously. Only if to create N copies of the same object artificially.

PR #1053, opened 2020-01-14
[Datumaro] Dataset annotations filter

### LINKED ISSUE #994: It dumps occluded objects when data is generated in YOLO format
HI,

We have used this tool for a project and got the scariest output ever with YOLO Framework.

Later we checked the generated dataset text files. When we set some object as occluded, this should be in the YOLO annotation files, but the tool does it somehow. At this moment I am not allowed to share the proprietary data and the annotations. But you can reproduce the issue with only 2 or 3 frames easily.

Reproduction Path:
1. Annotate a frame
2. Set the object as occluded
3. Dump in YOLO Format

Later I downloaded in another format and curated the dataset with some scripts and considered the IsOccluded field to avoid dumping the object in the annotation file.

However, overall this tool is quite amazing and have better prospect from other candidates as it directly comes from the OpenCV ecosystem. So we can rely on it. 

I hope my observation will help you make this tool better and we can rely on it without any doubts. 

I appreciate your good work. Please keep it up.

Meer

zhiltsov-max:
Added:
- Dataset annotations filter in CLI: `datumaro project extract -a`

To test:
```
  -e FILTER, --filter FILTER 
                        XML XPath filter expression for dataset items.
                        Examples: 
                        - extract images with width < height:
                        '/item[image/width < image/height]'; 
                        - extract images with large-area bboxes: 
                        '/item[annotation/type="bbox" and annotation/area>2000]'
                        - filter out irrelevant annotations from items: 
                        '/item/annotation[label = "person"]'
  -a, --filter-annotations 
                        Filter annotations instead of dataset items (default: False)
  --remove-empty          
                         Remove an item if there are no annotations left after filtration
  --dry-run         Print XML representations to be filtered and exit

datum project extract -a -e '/item/annotation[label = "person"]' --remove-empty -d temp
```

In general, to filter annotations make an XPath that returns `annotation` XML elements. Check `dry-run` option to obtain XML representations of dataset items.

Resolves #994

### DIFF
--- a/datumaro/datumaro/cli/project/__init__.py
+++ b/datumaro/datumaro/cli/project/__init__.py
@@ -11,6 +11,7 @@ import shutil
 
 from datumaro.components.project import Project
 from datumaro.components.comparator import Comparator
+from datumaro.components.dataset_filter import DatasetItemEncoder
 from .diff import DiffVisualizer
 from ..util.project import make_project_path, load_project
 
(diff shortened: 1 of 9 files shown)


nmanovic:
@zhiltsov-max , just a stupid suggestion to use instead of --filter-annotations the following option: --filter-mode
* `--filter-mode datum` - by default, filter dataset items
* `--filter-mode annotation` - filter annotations only
* `--filter-mode datum+annotation` - filter both annotations and datum

or something like that.

PR #1058, opened 2020-01-15
React UI: Player updates

bsekachev:
- Moving, zooming, additional canvas handler (zoom handler to select a ROI) 
- Shape activating & changing
- Saving of annotations on the server
- Drawing for shapes/tracks
- Minor fixes in CVAT-Canvas & CVAT-Core
- Absolute import in typescript UI instead of a lot of `../../../../../` (also plugins for eslint & webpack)
- Merge, group, split

So, right side panel is ready except of tags creation
Screenshot from 2020-01-16 17-03-05


cvat-canvas/src/typescript/canvasModel.ts
```diff
@@ -250,15 +258,41 @@ export class CanvasModelImpl extends MasterImpl implements CanvasModel {
...
         ));
 
-        this.notify(UpdateReasons.FIT_CANVAS);
-        this.notify(UpdateReasons.OBJECTS);
+        this.notify(UpdateReasons.FITTED_CANVAS);
+        this.notify(UpdateReasons.OBJECTS_UPDATED);
+    }
+
+    public dragCanvas(enable: boolean): void {
+        if (enable && this.data.mode !== Mode.IDLE) {
```
mmarkelov: What about moving this condition in separate function? Like so:
```javascript
public checkIfCanvasIsBusy(enable: boolean): void {
        if (enable && this.data.mode !== Mode.IDLE) {
            throw Error(`Canvas is busy. Action: ${this.data.mode}`);
        }
}
```

cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/draw-shape-popover-content.tsx
```diff
@@ -0,0 +1,167 @@
...
+        } = this.state;
+
+        const {
+            shapeType,
+            labels,
+            onDrawStart,
+            canvasInstance,
+        } = this.props;
+
+        let minimumPoints = 0;
```
mmarkelov: Maybe me move this condition to function out of render method? Like so:
```javascript
public getMinimumPoints(shapeType): number {
        let minimumPoints = 0;
        if (shapeType === ShapeType.POINTS) {
            minimumPoints = 1;
        } else if (shapeType === ShapeType.POLYLINE) {
            minimumPoints = 2;
        } else if (shapeType === ShapeType. POLYGON) {
            minimumPoints = 3;
        }
        return minimumPoints;
}

```

cvat-ui/src/reducers/annotation-reducer.ts
```diff
@@ -26,13 +39,19 @@ export default (state = defaultState, action: AnyAction): AnnotationState => {
             };
         }
         case AnnotationActionTypes.GET_JOB_SUCCESS: {
+            const { jobInstance } = action.payload;
```
mmarkelov: Maybe add other props here?
```javascript
  const { jobInstance, frame, frameData,  annotations } = action.payload;
```

bsekachev: Yes, great idea

cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/draw-shape-popover-content.tsx
```diff
@@ -0,0 +1,167 @@
...
+                                    step={1}
+                                />
+                            </Col>
+                        </Row>
+                    )
+                }
+                <Row type='flex' justify='space-around'>
+                    <Col span={12}>
+                        <Button
+                            onClick={(): void => {
```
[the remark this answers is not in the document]
bsekachev: Yes, we need some refactoring here

cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/draw-shape-popover-content.tsx
bsekachev: Yes

cvat-canvas/src/typescript/canvasModel.ts
bsekachev: I am not sure it makes sense. 
Yes, this condition is repeated twice, but other methods of ``cvat-canvas`` also check if canvas is busy, but using other conditions.

Issue #33, opened 2018-08-17
Remove all annotations inside a range of frames


Alexander-Petrikov:
This is very useful option to remove all options from one frame to another.
I want to reannotate part of video. It is not good idea to search for keyframes and turn off them or
delete lines in xml file and upload annotation.

nmanovic:
@Alexander-Petrikov ,

Do I understand correctly that you want to remove annotations from frame N till frame N+K? Are you trying to annotate an existing dataset? I just want to understand your annotation flow.

Alexander-Petrikov:
Yes, I want to remove annotation on video from frame N till frame N+K, very K is big.
I have annotated video. In part of video bounding boxes have too big size and I want to reannotate this part (not all). I don't want to search for keyframes.

gudipudiramanakumar:
Hey, I've submitted pull request #3617  to this issue. Can someone have a look at it

nmanovic:
@gudipudiramanakumar , thanks for your contribution.

PR #1061, opened 2020-01-16
Remove annotations

Priya4607:
Issue #33  
Included a button **Remove Annotation** (remove all annotations in a single frame or a range of frames) which prompts for user input, a user can enter a frame number `number` or a range of frames by prescribing `startvalue,endvalue` and press okay. In case the user enters three numbers each followed by a comma, it displays an alert message. This removes all annotations inside the range that users specifies. 
I did not alter the functionality of original `Remove Annotation` button just changed the name to `Remove all Annotations`.
Tasks:
-[x] clears history of removed annotations
-[x] removes annotations for a range of frames / a single frame
-[  ] save work

Issues: I need some clarity on `shapes` under class `shapeCollectionModel`  in `cvat/apps/engine/static/engine/js/shapeCollection.js`. I understand that boxmodels are stored as an array in `shapes`. I'm not sure I understand how to reference them with the frames if that's possible  so that particular boxmodel  can be removed when **Remove Annotation** is invoked. Whenever I try to save the changes I am not able to see the changes, as it retains the original work. How should I proceed from here? Could you please shed some light on this issue?

### DIFF
--- a/cvat/apps/engine/static/engine/js/annotationUI.js
+++ b/cvat/apps/engine/static/engine/js/annotationUI.js
@@ -458,7 +458,28 @@ function setupMenu(job, task, shapeCollectionModel,
                     shapeCollectionModel.empty();
                 });
         }
+    
+    
     });
+
+    $('#removeAnnotationButton1').on('click', () => {
+        if (!window.cvat.mode) {
+            hide();
+            let some= prompt('Enter range of frames: ', 'eg: 2,10/eg: 3');
+            let  some1 = some.split(",").map(Number);
+            console.log(some1);
+            shapeCollectionModel.empty1(some1);
+            historyModel.empty1(some1);
+
+            
+            
+        }
+    
+    
+    });
+
+
+
 
     // JS function cancelFullScreen don't work after pressing
     // and it is famous problem.
--- a/cvat/apps/engine/static/engine/js/history.js
+++ b/cvat/apps/engine/static/engine/js/history.js
@@ -56,7 +56,12 @@ class HistoryModel extends Listener {
         }
 
         this.notify();
+        console.log(this._redo_stack);
+        console.log(this._undo_stack);
+        
     }
+
+    
 
     redo() {
         let frame = window.cvat.player.frames.current;
@@ -87,7 +92,15 @@ class HistoryModel extends Listener {
         }
 
         this.notify();
+        console.log(this._undo_stack);
+        console.log(this._redo_stack);
+        
+
+
+
     }
+
+    
 
     addAction(name, undo, redo, frame) {
         if (this._locked) return;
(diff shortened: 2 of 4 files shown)


nmanovic:
@Priya4607 , thanks for the PR. We appreciate your time but the PR cannot be accepted. Next version of UI is coming (https://github.com/opencv/cvat/tree/develop/cvat-ui). It will be completely rewritten (ReactJS + Antd). It is already available in develop branch but it isn't enabled by default. If you can move the feature into the new UI we will be  glad to accept it.

Priya4607:
@nmanovic Any suggestions on how should I proceed further if I were to move up the features to a new UI. Would be glad if you could help me with the files I should be dealing with to solve issue #33   .

nmanovic:
@Priya4607 , in general there a couple of similar issues. In one case you want to delete objects in a range of frames. In another case you want to delete objects by a condition (e.g. with a specific label or attribute). Thus it is better to delete by a filter and a range of frames it is only one filter criteria. But it can be a starting point.

You need to add specific code into cvat-core to remove annotations: https://github.com/opencv/cvat/tree/develop/cvat-core/

After that you need to add a button and a dialog into cvat-ui and call cvat-core to actually remove annotations: https://github.com/opencv/cvat/tree/develop/cvat-ui

cvat-core is a js client library and acts as a proxy between cvat-ui and cvat-server. cvat-ui implements the new UI.

Priya4607:
@nmanovic thanks for the response, I am not able to view the new UI, I did update my local by `git pull upstream develop` and did the following:
`cd cvat`
`cd cvat-core && npm install`
`cd ../cvat-ui && npm install`
`cd  ../cvat-canvas && npm install`
`cd  ../cvat-ui `
`npm start`
inside the virtual environment,
`python manage.py migrate`
`python manage.py runserver 0:7000`

On opening http://localhost:3000/ and navigating to open the job of a task, I don't see the UI changes in the Annotation page, it still looks the same.  
Annotation UI

I had a look at the mock up of annotation page. Is the new UI being run on a different port? How should I be able to see the new UI? 

And regarding the ` filter`, is it already available in the new UI?

nmanovic:
@Priya4607 , 

- `filter` for annotations is available now. 
- New UI is not visible by default. You should specify URL explicitly `/tasks/<task id>/jobs/<job id>`

Priya4607:
@nmanovic ,
As you said `filter` for annotation is available now, but I don't see `filter`(that allows one to select objects by condition) for annotations in the new UI. Could you provide some clarity, as in where it is available exactly?
As you pointed out it is better to use filter to delete objects, I was thinking about using filter to specify criteria that selects labels/attributes/all objects in a range of frame instead of all frames. And then adding a `delete` button beside the `filter`, which prompts for user confirmation. I'm not sure about its feasibility, please present your views whenever possible.

nmanovic:
@Priya4607 , sorry for misleading. I said "filter for annotations is available now". A common filter should be implemented.

Issue #915, opened 2019-12-08
CVAT does not work when annotating PDFs


philippschw:
Hi, 

I am trying to annotate pdf documents instead of images with cvat and noticed a number of problems that I am not able to resolve alone. I am using the develop branch, because on the master branch the Docker image of cvat does not build successfully.

1. I am only able to upload a single pdf document (with many pages) but not several pdf documents. The error Code explains that I can only upload a single pdf but it would be helpful to understand the rational for this:
```ValueError: Only one video, archive, pdf or many image, directory can be used simultaneously,             but 0 image(s), 0 video(s), 0 archive(s), 2 pdf(s), 0 directory(s) found.```

2. The conversion from pdf to image with pdf2image is not working, because poppler is missing from the Dockerfile. I fixed it by adding it to the Dockerfile:
```
# Install poppler for working with pdfs
RUN apt-get update && apt install -y poppler-utils
```

3. After annotating a few items, I attempted to dump the annotation and no matter which format I use it fails, here is the error message. Note, dumping annotated png images works perfectly, seems to be a problem specific to pdfs.
```
2019-12-07 23:45:12,475 DEBG 'rqworker_default_1' stderr output:
23:45:12 default: cvat.apps.engine.annotation.dump_task_data('5', <SimpleLazyObject: <User: admin>>, '/home/django/data/5/5_IDP.admin.2019_12_07_23_45_12.zip', <AnnotationDumper: AnnotationDumper object (YOLO ZIP 1.0)>, 'http', 'localhost:8080') (admin@/api/v1/tasks/5/annotations/YOLO ZIP 1.0/5_IDP)

2019-12-07 23:45:12,574 DEBG 'rqworker_default_1' stderr output:
23:45:12 cvat.apps.engine.utils.InterpreterError: ValueError at line 308: '.upload' is not in list
Traceback (most recent call last):
  File "/home/django/cvat/apps/engine/utils.py", line 45, in execute_python_code
    exec(source_code, global_vars, local_vars)
  File "<string>", line 1, in <module>
  File "<string>", line 104, in dump
  File "/home/django/cvat/apps/annotation/annotation.py", line 325, in group_by_frame
    _get_frame(annotations, shape).labeled_shapes.append(self._export_labeled_shape(shape))
  File "/home/django/cvat/apps/annotation/annotation.py", line 308, in _get_frame
    rpath = os.path.sep.join(rpath[rpath.index(".upload")+1:])
ValueError: '.upload' is not in list

[...]

PR #1083, opened 2020-01-21
fix multiple pdfs bug

tanvi15:
fix with reference to issue #915


tanvi15:
i committed the wrong version. sorry

PR #1088, opened 2020-01-23
Add feature to upload and annotate multiple pdfs

tanvi15:
fix to issue #915


azhavoro:
Hi @tanvi15, thanks for the contribution. The similar PR https://github.com/opencv/cvat/pull/939 has been submitted before. At the moment we don't want to change the current CVAT limitations for PDFs. In the future we plan to resolve this issue by another way. Thanks.

benhoff:
With the concurrence of the CVAT devs, I would recommend tracking the PDF metadata in either a new metadata file, or in the JPEG exif data.

You could maintain the current flat data structure, and reconstruct the file names on export.

https://stackoverflow.com/questions/52729428/how-to-write-custom-metadata-into-jpeg-with-python

Tradeoffs would probably be: 

- more code with creating a new metadata format
- External dependency with exif data

PR #1094, opened 2020-01-27
Resolved import error caused by pip 20.0

Priya4607:
I faced the same ImportError as raised in the https://gitter.im/opencv-cvat/public by @praffulkmr, while installing TF Object Detection API: Auto Annotation.
Pip_error
No such error is found and able to build the docker image successfully, if the pip upgrade is set to version 20.0.1 in the Dockerfile.


nmanovic:
@Priya4607 , thanks for the fix. Do you know if pip has an open issue regarding the problem?

azhavoro:
Related pip issue: https://github.com/pypa/pip/issues/7217

Priya4607:
@nmanovic,
Pip issue: pypa/pip#7620
Related PR: pypa/pip#7621

PR #1098, opened 2020-01-27
Update base.py

LukeAI:
Modification necessary for using CVAT from remote machines when accessing with FQDNs
See https://github.com/opencv/cvat/issues/1011#issue-542817055

This might just be a hacky workaround, compensating for a bug in proper parsing of the URL

"I believe the reason for this is that sometimes if the port number is :80 and the URL is not in the LAN (<HOST-IP>:port), but instead it is a Fully Qualified Domain Name (<FQDN>:port), the port 80 is redundant (mydomain.com:80) and the errors arise."


LukeAI:
I have tested, found it to work fine and crearted a new PR https://github.com/opencv/cvat/pull/1099

Issue #1063, opened 2020-01-16
extreme clicking method


test4fest:
If it's applicable , I would like to ask you to support **[extreme clicking method](http://openaccess.thecvf.com/content_ICCV_2017/papers/Papadopoulos_Extreme_Clicking_for_ICCV_2017_paper.pdf)** for future versions.

Thanks

nmanovic:
@test4fest , it should not very difficult to implement. It can be done in UI only. Just need a new virtual shape like "box by 4 points". Internally it can be mapped on a usual bounding box but the drawing method should be handled in another way.

Do you think it is something you can contribute?

azhavoro:
Implemented #1111

PR #1111, opened 2020-01-30
Add extreme clicking feature to draw box by 4 points

Jooong:
Hi, 
This is the [extreme clicking](https://arxiv.org/pdf/1708.02750.pdf) feature requested in #1063.  A virtual shape **box by 4 points** is added, which is actually built as **box** shape after drawing is finished. The drawing method is similar to that of **dextr_segmentation**, except that it automatically finishes drawing after the fourth click. I also added explanation in [user guide](https://github.com/Jooong/cvat/blob/extreme_clicking/cvat/apps/documentation/user_guide.md#annotation-with-box-by-4-points).

Sample demo:
gif016


cvat/apps/engine/static/engine/js/shapeCreator.js
```diff
@@ -125,7 +125,7 @@ class ShapeCreatorModel extends Listener {
     }
 
     set defaultType(type) {
-        if (!['box', 'points', 'polygon', 'polyline'].includes(type)) {
+        if (!['box', 'box_by_4_points', 'points', 'polygon', 'polyline'].includes(type)) {
```
bsekachev: There is incompatibilities with this code: https://github.com/opencv/cvat/blob/e062c2309ecc648afc2e7261bdf41224a7508dcf/cvat/apps/dextr_segmentation/static/dextr_segmentation/js/enginePlugin.js#L57

Jooong: Fixed this here: https://github.com/opencv/cvat/pull/1111/commits/211efd883a5678e3742bf04cf52e2319003eed42#diff-2abd9452793a719d412491d1bfa0d501R57

cvat/apps/engine/static/engine/js/shapeCreator.js
```diff
@@ -511,6 +511,67 @@ class ShapeCreatorView {
...
+                            point.y = Math.clamp(point.y, 0, frameHeight);
+
+                            // update bounding box
+                            box.xtl = Math.min(point.x, box.xtl);
+                            box.ytl = Math.min(point.y, box.ytl);
+                            box.xbr = Math.max(point.x, box.xbr);
+                            box.ybr = Math.max(point.y, box.ybr);
+                        }
+
+                        if ((box.ybr - box.ytl) * (box.xbr - box.xtl) >= AREA_TRESHOLD) {
```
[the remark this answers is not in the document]
Jooong: Fixed this here: https://github.com/opencv/cvat/pull/1111/commits/211efd883a5678e3742bf04cf52e2319003eed42#diff-cba90afd35f0130c00e84a703eba92a1R559

Jooong:
@azhavoro 
Thanks for your review. Regarding your question, the new drawing mode should also work for interpolation tracks. This problem is fixed in [211efd](https://github.com/opencv/cvat/pull/1111/commits/211efd883a5678e3742bf04cf52e2319003eed42#diff-cba90afd35f0130c00e84a703eba92a1R255), and other commented issues are also fixed. Regarding the implementation for the new UI, what would be good idea between making commit in this PR, or to make a new PR after this?

azhavoro:
> @azhavoro
> Thanks for your review. Regarding your question, the new drawing mode should also work for interpolation tracks. This problem is fixed in [211efd](https://github.com/opencv/cvat/pull/1111/commits/211efd883a5678e3742bf04cf52e2319003eed42#diff-cba90afd35f0130c00e84a703eba92a1R255), and other commented issues are also fixed. Regarding the implementation for the new UI, what would be better idea between making commit in this PR, or to make a new PR after this?
Thanks, I think it's better to make a new PR for the new UI.

azhavoro:
@Jooong 
I tried to add a interpolation bounding box by 4 points, but it was created as an annotation object...
image

Jooong:
> @Jooong
> I tried to add a interpolation bounding box by 4 points, but it was created as an annotation object...
> image

@azhavoro  Sorry for the bug. I forgot to fix [here](https://github.com/opencv/cvat/commit/046885fb7ea6da1bd7ec59f39ec19024d8102160#diff-cba90afd35f0130c00e84a703eba92a1R57). I amended the last commit. It works fine now.

gif018

PR #1127, opened 2020-02-06
Add extreme clicking method in cvat-canvas and cvat-ui

Jooong:
Hi, 
This is the reimplementation of [extreme clicking](https://arxiv.org/pdf/1708.02750.pdf) (#1111) in new UI with React and Ant Design. `Drawing Method` option is added in rectangle drawing popover. The conventional bbox drawing method (`By 2 Points`) is selected by default. Extreme clicking method is available with `By 4 Points`. Drawing finishes right after the fourth click, as it does in #1111. Here are sample images for this feature.
aa
track

### DIFF
--- a/cvat-canvas/src/typescript/canvasModel.ts
+++ b/cvat-canvas/src/typescript/canvasModel.ts
@@ -40,7 +40,7 @@ export interface ActiveElement {
 export interface DrawData {
     enabled: boolean;
     shapeType?: string;
-    boxDrawingType?: string;
+    rectDrawingMethod?: string;
     numberOfPoints?: number;
     initialState?: any;
     crosshair?: boolean;
--- a/cvat-canvas/src/typescript/canvasModel.ts
+++ b/cvat-canvas/src/typescript/canvasModel.ts
@@ -40,6 +40,7 @@ export interface ActiveElement {
 export interface DrawData {
     enabled: boolean;
     shapeType?: string;
+    boxDrawingType?: string;
     numberOfPoints?: number;
     initialState?: any;
     crosshair?: boolean;
--- a/cvat-canvas/src/typescript/drawHandler.ts
+++ b/cvat-canvas/src/typescript/drawHandler.ts
@@ -203,40 +203,39 @@ export class DrawHandlerImpl implements DrawHandler {
 
     private drawBoxBy4Points(): void {
         let numberOfPoints = 0;
-        this.drawInstance = (this.canvas as any).polygon().draw({
-            snapToGrid: 0.1,
-        }).addClass('cvat_canvas_shape_drawing').style({
-            'stroke-width': 0,
-            opacity: 0,
-        }).on('drawstart', () => {
-            // init numberOfPoints as one on drawstart
-            numberOfPoints = 1;
-        }).on('drawpoint', (e: CustomEvent) => {
-            // increase numberOfPoints by one on drawpoint
-            numberOfPoints += 1;
+        this.drawInstance = (this.canvas as any).polygon()
+            .addClass('cvat_canvas_shape_drawing').attr({
+                'stroke-width': 0,
+                opacity: 0,
+            }).on('drawstart', () => {
+                // init numberOfPoints as one on drawstart
+                numberOfPoints = 1;
+            }).on('drawpoint', (e: CustomEvent) => {
+                // increase numberOfPoints by one on drawpoint
+                numberOfPoints += 1;
 
-            // finish if numberOfPoints are exactly four
-            if (numberOfPoints === 4) {
-                const bbox = (e.target as SVGPolylineElement).getBBox();
-                const [xtl, ytl, xbr, ybr] = this.getFinalRectCoordinates(bbox);
+                // finish if numberOfPoints are exactly four
+                if (numberOfPoints === 4) {
+                    const bbox = (e.target as SVGPolylineElement).getBBox();
+                    const [xtl, ytl, xbr, ybr] = this.getFinalRectCoordinates(bbox);
 
-                if ((xbr - xtl) * (ybr - ytl) >= consts.AREA_THRESHOLD) {
-                    this.onDrawDone({
-                        shapeType: this.drawData.shapeType,
-                        points: [xtl, ytl, xbr, ybr],
-                    });
-                } else {
-                    this.onDrawDone(null);
+                    if ((xbr - xtl) * (ybr - ytl) >= consts.AREA_THRESHOLD) {
+                        this.onDrawDone({
+                            shapeType: this.drawData.shapeType,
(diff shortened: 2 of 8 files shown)


azhavoro:
@Jooong Hi,
cool, many thanks for the contribution!!! I'll try the patch tomorrow.

cvat-ui/src/containers/annotation-page/standard-workspace/controls-side-bar/draw-shape-popover.tsx
```diff
@@ -67,6 +70,7 @@ function mapStateToProps(state: CombinedState, own: OwnProps): StateToProps {
 type Props = StateToProps & DispatchToProps;
 
 interface State {
```
bsekachev: Let's add this value also to global redux state.
We need it in order to in the future be able to start drawing using only one shortcut, with all previous settings.

cvat-canvas/src/typescript/canvasModel.ts
```diff
@@ -40,6 +40,7 @@ export interface ActiveElement {
 export interface DrawData {
     enabled: boolean;
     shapeType?: string;
+    boxDrawingType?: string;
```
bsekachev: How about renaming it everywhere to ``rectDrawingMethod``?

cvat-canvas/src/typescript/drawHandler.ts
```diff
@@ -491,9 +531,15 @@ export class DrawHandlerImpl implements DrawHandler {
                 }
             }
         } else if (this.drawData.shapeType === 'rectangle') {
-            this.drawBox();
-            // Draw instance was initialized after drawBox();
-            this.shapeSizeElement = displayShapeSize(this.canvas, this.text);
+            if (this.drawData.boxDrawingType === 'by_two_points') {
```
bsekachev: Actually boxDrawingType declared as optional. So, it would be better to redesign this condition like:
```
this.drawData.boxDrawingType === 'by_four_points' =>> draw with four points
else =>> draw with two points
```

cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/draw-shape-popover.tsx
```diff
@@ -71,7 +77,39 @@ function DrawShapePopoverComponent(props: Props): JSX.Element {
...
+                            <Col>
+                                <Text className='cvat-text-color'> Drawing method </Text>
+                            </Col>
+                        </Row>
+                        <Row type='flex' justify='space-around'>
+                            <Col>
+                                <Radio.Group
+                                    style={{ display: 'flex' }}
+                                    defaultValue={BoxDrawingType.BY_TWO_POINTS}
+                                    onChange={(event: RadioChangeEvent): void => {
```
bsekachev: Maybe shouldn't we create one more lambda here?
Why do not just pass ``onChangeBoxDrawingType(event:  RadioChangeEvent): void`` to this component from the container

cvat-canvas/src/typescript/canvasModel.ts
bsekachev: Please, update the interface in ``cvat-canvas/README.md``

azhavoro:
Couple of comments from me:
1. One issue here: 
   Scenario: If I try to draw a new box by 4 point, but want to cancel the process and click the `rectangle` button again. In this case the button remains pressed and I cannot start draw a new shape.
   Peek 2020-02-07 14-38
1. Could you please fix the codacy issues?

cvat-ui/src/containers/annotation-page/standard-workspace/controls-side-bar/draw-shape-popover.tsx
Jooong: I see. I added activeRectDrawingMethod in [cvat-ui/src/reducers/annotation-reducer.ts](https://github.com/opencv/cvat/pull/1127/files#diff-385a09313d2054c79c235452ff60070bR359)

cvat-canvas/src/typescript/canvasModel.ts
Jooong: > How about renaming it everywhere to `rectDrawingMethod`?

Cool. This seems to be more straightforward. Renamed it everywhere.

> Please, update the interface in `cvat-canvas/README.md`

Updated [cvat-canvas/README.md](https://github.com/opencv/cvat/pull/1127/files#diff-80bfdbe0c80ffbc37c31dcaadfa68348R43)

Jooong: BTW, I rebased from the current develop branch to avoid merge-conflict. `Outdated` tag is attached in this thread because of it.

cvat-canvas/src/typescript/drawHandler.ts
Jooong: Thanks, fixed it [here](https://github.com/opencv/cvat/pull/1127/commits/c400fb04d52c1baa4867753bcad6e5da360bd4a0#diff-1286785df8910c7d3f05a91f3a96e17dR582)

cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/draw-shape-popover.tsx
Jooong: Thought it would be better to keep the parameter as simple data type as in other named call back functions. But directly passing the event seems to be more readable. Fixed [here](https://github.com/opencv/cvat/pull/1127/commits/865c0aa3ec910d4ed441766208329fdbea9561ce#diff-789be1c3834d95b25049c8d995578c19R92).

Jooong:
> Couple of comments from me:
> 
> 1. One issue here:
>    Scenario: If I try to draw a new box by 4 point, but want to cancel the process and click the `rectangle` button again. In this case the button remains pressed and I cannot start draw a new shape.
>    Peek 2020-02-07 14-38
> 2. Could you please fix the codacy issues?

1. I fixed cancel condition [here](https://github.com/opencv/cvat/pull/1127/commits/865c0aa3ec910d4ed441766208329fdbea9561ce#diff-1286785df8910c7d3f05a91f3a96e17dR163). Thanks for your review.

2. I also fixed codacy issue.

azhavoro:
@Jooong Hi, unfortunately, adding bbox by 4 points doesn't work for me
Peek 2020-02-11 11-09

Jooong:
> @Jooong Hi, unfortunately, adding bbox by 4 points doesn't work for me
> Peek 2020-02-11 11-09

Sorry, there was a bug after rebasing from develop branch. This is now fixed. Thanks again for your review.
dd

nmanovic:
@Jooong , thanks for the great contribution. I hope to see more PRs from you in the future!

PR #1132, opened 2020-02-07
Docker compose with webpack

nglazov:
I've added dev mode for webpack in docker-compose
Command to run app: `docker-compose -f docker-compose.yml -f docker-compose.dev.yml up`

There is problem with node-sass: it should be compiled for a specific OS. 
It was solved so: 
1. docker installs npm modules in tmp files
1. docker-comopose mount host folders to container (cvat-ui, cvat-core, cvat-canvas)
1. in case when there is no node_modules folder it will be copied from container to local machine (once at first run)

### DIFF
--- a/Dockerfile.dev.ui
+++ b/Dockerfile.dev.ui
@@ -1,7 +1,5 @@
 FROM node:10-alpine
 
-RUN apk update && apk add bash
-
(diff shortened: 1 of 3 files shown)


nmanovic:
@nglazov , as discussed a debugging session can be started using `npm start` command and it is more convenient for now.

PR #1136, opened 2020-02-11
Typed reducers

nglazov:
I suggest to put actions and reducers in modules folder, it will be more comfortable to work with it (when you are changing reducer you are changing actions and when they are in one folder it's more fast)

Also i've added typed actions in reducer and when you are writing reducers case you know which payload is there.

I moved one component inside of container, it allow to decrease loc (you don't need to define typings twice) and used typeof and ResolveThunks to avoid extra typings


cvat-ui/src/containers/register-page/register-page.tsx
```diff
@@ -1,38 +1,80 @@
...
+    onRegister: typeof registerAsync;
 }
 
-function mapStateToProps(state: CombinedState): StateToProps {
-    return {
-        fetching: state.auth.fetching,
+type Props = StateToProps & ResolveThunks<DispatchToProps> & RouteComponentProps;
+
+function RegisterPageContainer(props: Props) {
+    const sizes = {
```
mmarkelov: It's better to move out of function

cvat-ui/src/containers/register-page/register-page.tsx
```diff
@@ -1,38 +1,80 @@
...
+    } = props;
+
+    return (
+        <Row type='flex' justify='center' align='middle'>
+            <Col {...sizes}>
+                <Title level={2}> Create an account </Title>
+                <RegisterForm
+                    fetching={fetching}
+                    onSubmit={(registerData: RegisterData): void => {
+                        onRegister(
```
mmarkelov: I suppose using object here will be better

nglazov: It was just moved code without any changes

cvat-ui/src/modules/auth/authReducer.ts
```diff
@@ -53,7 +52,7 @@ export default (state = defaultState, action: AnyAction): AuthState => {
             return {
                 ...state,
                 fetching: true,
-                user: action.payload.user,
+                user: undefined,
```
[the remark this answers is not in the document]
nglazov: I've used value by types: i had empty object and user of {} is undefined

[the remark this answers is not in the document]
bsekachev: It was a bug

cvat-ui/package.json
```diff
@@ -14,12 +18,15 @@
     "@babel/preset-env": "^7.6.0",
     "@babel/preset-react": "^7.0.0",
     "@babel/preset-typescript": "^7.6.0",
-    "@typescript-eslint/eslint-plugin": "^2.10.0",
+    "@typescript-eslint/eslint-plugin": "^2.19.2",
+    "@typescript-eslint/experimental-utils": "^2.19.2",
```
bsekachev: I can't see where you use this package

bsekachev:
@nglazov 

There is a merge conflict

cvat-ui/package.json
nglazov: Yes, i think it was added, when i tried to run eslint local...
It looks like "babel": "^6.23.0" may be removed to, is not it?

bsekachev: @nglazov

Yep, it looks like babel is also not needed.

cvat-ui/src/utils/redux.ts
```diff
@@ -0,0 +1,18 @@
...
+export interface ActionWithPayload<T, P> extends Action<T> {
+    payload: P;
+}
+
+export function createAction<T extends string>(type: T): Action<T>;
+export function createAction<T extends string, P>(type: T, payload: P): ActionWithPayload<T, P>;
+export function createAction<T extends string, P>(type: T, payload?: P) {
+    return typeof payload === 'undefined' ? { type } : { type, payload };
+}
+
```
bsekachev: I am not very get on well with typescript, and I am trying to sort it out. 

They are overloads, aren't they?
I [read](https://howtodoinjava.com/typescript/function-overloading/), that all overloads must have the same return type

bsekachev: Why do not extend type redux::Action by adding a payload property that has a generic type?

Something like that:
```ts
interface Action<T extends string, P>  implements  ReduxAction<T> {
    type: T,
    payload: P,
}

export function createAction<T extends string>(type: T): Action<T, undefined>;
export function createAction<T extends string, P>(type: T, payload: P): Action<T, P>;
export function createAction<T extends string, P>(type: T, payload?: P) {
    return typeof payload === 'undefined' ? { type } : { type, payload };
}

```

bsekachev: It's more question, than offer.

bsekachev: Oh, actually you do almost the same, but nevertheless return different types.

cvat-ui/src/cvat-store.ts
```diff
@@ -21,9 +22,10 @@ const middlewares = [
 let store: Store | null = null;
 
 export default function createCVATStore(createRootReducer: () => Reducer): void {
+    // todo enable devtools only in dev mode
```
bsekachev: So, for now does it work in production?
Do you happen to know how it affects performance?

cvat-ui/src/utils/redux.ts
bsekachev: Okay, I will investigate about

cvat-ui/src/cvat-store.ts
nglazov: It's easier to remove todo:)

cvat-ui/package.json
nglazov: Fixed

PR #1140, opened 2020-02-12
[Datumaro] Introduce image info

zhiltsov-max:
Added:
- `Image` class, which has `path` and `size` as metainfo
- `ImageDir` converter, which saves images to a directory
- `IdFromImageName` transform, which allows to merge CVAT and COCO datasets with all others (allows to match items)
- Added additional image metainfo file support for YOLO format, which allows to load such a dataset when no images present
- Updated CVAT/Datumaro bindings that it was possible to export and *import* tasks
- `Inference` and `DatasetFilter` are now `Transforms`

Fixed:
- Transforms requiring image info now can be chained

### DIFF
--- a/cvat/apps/dataset_manager/bindings.py
+++ b/cvat/apps/dataset_manager/bindings.py
@@ -14,7 +14,7 @@ from cvat.apps.engine.annotation import TaskAnnotation
 from cvat.apps.engine.models import Task, ShapeType, AttributeType
 
 import datumaro.components.extractor as datumaro
-from datumaro.util.image import lazy_image
+from datumaro.util.image import Image
 
 
 class CvatImagesDirExtractor(datumaro.Extractor):
@@ -29,8 +29,7 @@ class CvatImagesDirExtractor(datumaro.Extractor):
                 path = osp.join(dirpath, name)
                 if self._is_image(path):
                     item_id = Task.get_image_frame(path)
-                    item = datumaro.DatasetItem(
-                        id=item_id, image=lazy_image(path))
+                    item = datumaro.DatasetItem(id=item_id, image=path)
                     items.append((item.id, item))
 
         items = sorted(items, key=lambda e: int(e[0]))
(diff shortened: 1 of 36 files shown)


nmanovic:
@zhiltsov-max , after previous merged PRs I got merge conflicts in the PR. Could you please update the PR?

cvat/apps/dataset_manager/bindings.py
```diff
@@ -183,8 +160,64 @@ def convert_attrs(label, cvat_attrs):
...
+        cvat_annotations = TaskAnnotation(db_task.id, user)
+        with transaction.atomic():
+            cvat_annotations.init_from_db()
+        cvat_annotations = Annotation(cvat_annotations.ir_data, db_task)
+        super().__init__(url, cvat_annotations)
+
+
+def match_frame(item, cvat_task_anno):
+    frame_number = None
+    if frame_number is None:
```
nmanovic: @zhiltsov-max , probably we don't need the 'if'.

datumaro/datumaro/util/image.py
```diff
@@ -123,38 +124,100 @@ def decode_image(image_bytes):
...
-        return hash((id(self), self.path, self.loader))
\ No newline at end of file
+        return hash((id(self), self.path, self.loader))
+
+class Image:
+    def __init__(self, data=None, path=None, loader=None, cache=None,
+            size=None):
+        assert size is None or len(size) == 2
+        if size is not None:
+            assert len(size) == 2 and 0 < size[0] and 0 < size[1], size
```
nmanovic: len(size) == 2 already checked above.

datumaro/datumaro/util/image.py
```diff
@@ -123,38 +124,100 @@ def decode_image(image_bytes):
...
+            size = tuple(size)
+        else:
+            size = None
```
nmanovic: size is None here already.

PR #1142, opened 2020-02-14
Add editorconfig

nglazov:
https://editorconfig.org/

It's a file to unify multi editors/IDE work

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,17 @@
+# EditorConfig helps developers define and maintain consistent
+# coding styles between different editors and IDEs
+# editorconfig.org
+
+root = true
+
+[*]
+
+# Change these settings to your own preference
+indent_style = space
+indent_size = 2
+
+# We recommend you to keep these unchanged
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,7 +8,7 @@ root = true
 
 # Change these settings to your own preference
 indent_style = space
-indent_size = 2
+indent_size = 4
 
 # We recommend you to keep these unchanged
 end_of_line = lf


.editorconfig
```diff
@@ -0,0 +1,17 @@
...
+# coding styles between different editors and IDEs
+# editorconfig.org
+
+root = true
+
+[*]
+
+# Change these settings to your own preference
+indent_style = space
+indent_size = 2
```
nmanovic: I'm not sure that indent_size = 2 is correct in our case.

nglazov: It may be configured per file type. 
I've got an error with idents in json-file and thought that such ident is valid for all file types.

.editorconfig
```diff
@@ -0,0 +1,17 @@
...
+
+# We recommend you to keep these unchanged
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
```
bsekachev: It would be nice to mention in ``CONTRIBUTING.md`` that a user needs to install EditorConfig plugin, because in another case this file doesn't make sense. 

Also, as Nikita said, there is a problem with ``indent_size``. Now most of our files (except of several config files) have indent 4. So, I believe we should use the value 4 as a default value.

nglazov: I'm not sure about plugins part in CONTRIBUTING.md, because it should not depend on user's editor.

indent_size = 4 for [*] or for some file mask like [*.ts{,x}]?

bsekachev: @nglazov

Yes, it should not depend on user's editor. But if a user knows nothing about the project has ``.editorconfig``, the file is useless. Because by default (without a plugin) IDEs don't use the file, right? (at least such behavior in my vscode)

Idea is to say a user: "Hey, there is a file to setup your IDE. Just need to install a plugin. For example in vscode you can do it in extensions manager"

I assume ``indent_size = 4 for [*]`` in this PR. Next we can add additional lines or fix config files for consistent spacing.

nglazov: I've checked few repos ([react](https://reactjs.org/docs/how-to-contribute.html), [redux](https://github.com/reduxjs/redux/blob/master/CONTRIBUTING.md), [angular](https://github.com/angular/angular/blob/master/CONTRIBUTING.md)) and there is no word about editorconfig in CONTRIBUTING.

nmanovic: @nglazov , I agree with Boris. It seems that some code editors (like VS code) will not use the file by default: see https://editorconfig.org. Better to add a paragraph about the new feature.

nmanovic:
@nglazov , @bsekachev , let's accept the PR as is for now.

PR #1147, opened 2020-02-17
[Datumaro] Add offline subset remapping and bbox conversion

zhiltsov-max:
Added:
- `MapSubsets` transform, which allows to rename subsets
    `datum project transform -t map_subsets -- -s 'a:b' -s 'train:val' -s ':test'`
- `ShapesToBoxes` transform, which converts masks, points, polygons and lines to polygons


nmanovic:

PR #1149, opened 2020-02-17
[Datumaro] Pretty output folder names

zhiltsov-max:
Added:
- CLI commands now generate output folder names using operation parameters such as output format name and transform name

Fixed:
- Failing `project info` command


nmanovic:

PR #1151, opened 2020-02-17
Replace YOLO format support in CVAT with Datumaro

zhiltsov-max:
Replace YOLO import and export with Datumaro. It is to be noted that replacement is not `drop-in`, i.e. file format is slightly changed, so now YOLO dataset is required.

Before:
```
annotations.zip/
- obj.names
- image_name1.txt
- image_name2.txt
...
```

After:
```
# the format spec: https://github.com/AlexeyAB/darknet#how-to-train-to-detect-your-custom-objects

archive.zip/
├── obj.data 
├── obj.names # the same
├── obj_<subset>_data
│   ├── image_name1.txt # the same
│   └── image_name2.txt # the same
└── train.txt # list of subset image paths - mandatory

# train.txt
data/obj_train_data/image_name1.jpg
data/obj_train_data/image_name2.jpg

# obj.data
classes = 1
train = data/train.txt
names = data/obj.names
backup = backup/
```

Depends on #1140

### DIFF
--- a/cvat/apps/annotation/yolo.py
+++ b/cvat/apps/annotation/yolo.py
@@ -24,98 +24,46 @@ format_spec = {
 
 def load(file_object, annotations):
     from pyunpack import Archive
-    import os
+    import os.path as osp
     from tempfile import TemporaryDirectory
     from glob import glob
-
-    def convert_from_yolo(img_size, box):
-        # convertation formulas are based on https://github.com/pjreddie/darknet/blob/master/scripts/voc_label.py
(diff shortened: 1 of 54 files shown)


nmanovic:

cvat/apps/annotation/yolo.py
```diff
@@ -24,98 +24,46 @@
 
 def load(file_object, annotations):
```
nmanovic: @zhiltsov-max , I don't think that the file should exist at all. The behaviour should be similar to 'dataset export' functionality. But let's go step by step.

PR #1156, opened 2020-02-18
[Datumaro] Add masks to tfrecord format

zhiltsov-max:
Added:
- Support for instance masks in tfrecord format, export flag `--save-masks`
- Little refactoring for similar COCO export code


nmanovic:

PR #1157, opened 2020-02-18
Replace tfrecord format support in CVAT with Datumaro

zhiltsov-max:
No regressions, identical files.

Depends on #1140


nmanovic:

PR #1162, opened 2020-02-19
Fixed dump for cases when special URL characters in task name

### LINKED ISSUE #1161: Can't dump tasks with special characters in name (like: #)

bsekachev:
Resolved #1161

### DIFF
--- a/cvat-core/src/server-proxy.js
+++ b/cvat-core/src/server-proxy.js
@@ -559,7 +559,9 @@
             async function dumpAnnotations(id, name, format) {
                 const { backendAPI } = config;
                 const filename = name.replace(/\//g, '_');
-                let url = `${backendAPI}/tasks/${id}/annotations/${filename}?format=${format}`;
+                const baseURL = `${backendAPI}/tasks/${id}/annotations/${encodeURIComponent(filename)}`;
+                let query = `format=${encodeURIComponent(format)}`;
+                let url = `${baseURL}?${query}`;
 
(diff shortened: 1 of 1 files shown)


cvat-core/src/server-proxy.js
```diff
@@ -569,7 +571,8 @@
                             if (response.status === 202) {
                                 setTimeout(request, 3000);
                             } else {
-                                url = `${url}&action=download`;
+                                query = `${query}&action=download`;
```
nglazov: Suggest to use some (library)[https://www.npmjs.com/package/qs], to work with query as object

nmanovic:
@bsekachev , can we add a unit test?

cvat-core/src/server-proxy.js
bsekachev: @nglazov 

Looks like overhead, to add library for such a minor case
What is more, similar functionality for working with query string is already part of native JS: https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams

bsekachev:
@nmanovic 

Yes, but probably it will be in next PRs, just because we are short of time to implement different features that affect UI till next release.

cvat-core/src/server-proxy.js
nglazov: Yes, it's good solution too
I mean, that better to work with objects instead of strings

bsekachev: It makes sense

PR #1164, opened 2020-02-19
React-UI: settings

### LINKED ISSUE #490: Settings page
This page is responsible to task settings managing.

Features list:
- [x] Workspace settings
- [x] Player settings

Designs: https://projects.invisionapp.com/share/AXQ5CDD3QTC#/screens/341984218

ActiveChooN:
* Image filters: brightness, contrast, saturation
* Auto saving
* Frame auto fit
* Player speed
* Leave confirmation for unsaved changes

Resolved #490


cvat-ui/src/components/annotation-page/standard-workspace/canvas-wrapper.tsx
```diff
@@ -343,6 +349,12 @@ export default class CanvasWrapperComponent extends React.PureComponent<Props> {
         }
         canvasInstance.grid(gridSize, gridSize);
 
+        // Filters
+        const backgroundElement = window.document.getElementById('cvat_canvas_background');
+        if (backgroundElement) {
+            backgroundElement.style.filter = `brightness(${brightnessLevel/50}) contrast(${contrastLevel/50}) saturate(${saturationLevel/50})`;
```
bsekachev: Let's setup such border values for:
Brightness: [50 - 200], default 100
Contrast: [50 - 200], default 100
Saturate: [0 - 300], default 100

cvat-ui/src/containers/annotation-page/top-bar/top-bar.tsx
```diff
@@ -107,30 +132,80 @@ function mapDispatchToProps(dispatch: any): DispatchToProps {
...
             playing,
             canvasIsReady,
-            onChangeFrame,
             onSwitchPlay,
         } = this.props;
 
+        
         if (playing && canvasIsReady) {
             if (frameNumber < jobInstance.stopFrame) {
+                const delay: number = frameChangeTime ? Math.max(0,~~(1000/frameSpeed) -  new Date().getTime() + frameChangeTime) : 0;
```
bsekachev: Let's use Math.floor because it acts more obvious than double NOT bit wise operator

cvat-ui/src/containers/annotation-page/top-bar/top-bar.tsx
```diff
@@ -107,30 +132,80 @@ function mapDispatchToProps(dispatch: any): DispatchToProps {
...
                 setTimeout(() => {
                     const { playing: stillPlaying } = this.props;
                     if (stillPlaying) {
-                        onChangeFrame(frameNumber + 1);
+                        this.onChangeFrame(frameNumber + 1, new Date().getTime());
```
bsekachev: If we trying to calculate delay depends on updating time I would suggest simpler method:

``changeFrameAsync`` gets time at the beginning and at the end of the function. It computes difference. 
Finally it pushes difference to store via the CHANGE_FRAME_SUCCESS action.
The store uses this value to update ``delay`` value (using ``frameSpeed`` as a base value).
The top bar container gets ``delay`` value from the state and just uses it here. 

It would allow us remove extra argument from ``onChangeFrame`` and incapsulate all logic related with computing time in the one place.

cvat-ui/src/containers/annotation-page/top-bar/top-bar.tsx
```diff
@@ -107,30 +132,80 @@ function mapDispatchToProps(dispatch: any): DispatchToProps {
...
 
+    componentDidMount(): void {
+        const {
+            autoSave, 
+            autoSaveInterval,
+            saving,
+            unsaved,
+         } = this.props;
+
+        this.autoSaveInterval = window.setInterval((autoSave: boolean, saving: boolean): void => {
```
bsekachev: Just get these values from closure. It isn't needed to pass them to setInterval.

cvat-ui/src/reducers/annotation-reducer.ts
```diff
@@ -353,6 +358,7 @@ export default (state = defaultState, action: AnyAction): AnnotationState => {
                 annotations: {
                     ...state.annotations,
                     activatedStateID: null,
+                    unsaved: true,
```
bsekachev: There are a lot of ways to change annotations. Do not try to compute whether annotations was updated or not here. 
``cvat-core`` has a method hasUnsavedChanges(), please, use it instead. 

Like: ``jobInstance.annotations.hasUnsavedChanges()`` // true or false

cvat-ui/src/containers/annotation-page/top-bar/top-bar.tsx
```diff
@@ -107,30 +132,80 @@ function mapDispatchToProps(dispatch: any): DispatchToProps {
...
+            canvasIsReady,
+            onChangeFrame,
+            resetZoom,
+        } = this.props;
+
+        onChangeFrame(newFrame, time);
+
+        if (canvasIsReady && resetZoom) {
+            canvasInstance.fit();
+        }
```
bsekachev: It will be executed immediately after ``onChangeFrame()`` function (which is async). So, frame isn't expected to be changed yet. 

To make fit() I would recommend you to use ``canvas.setup`` event (see canvas-wrapper component, initialSetup method)

cvat-ui/src/components/task-page/job-list.tsx
```diff
@@ -34,12 +36,18 @@ export default function JobListComponent(props: Props): JSX.Element {
...
     const columns = [{
         title: 'Job',
         dataIndex: 'job',
         key: 'job',
-        render: (id: number): JSX.Element => (<a href={`${baseURL}/?id=${id}`}>{ `Job #${id}` }</a>),
+        render: (id: number): JSX.Element => (
+            <div>
+                <Link to={`/tasks/${taskId}/jobs/${id}`}>{`Job #${id}`}</Link>
+                {' | '}
+                <a href={`${baseURL}/?id=${id}`}>Old</a>
```
bsekachev: Old => Legacy UI

cvat-ui/src/containers/annotation-page/top-bar/top-bar.tsx
ActiveChooN: > `changeFrameAsync` gets time at the beginning and at the end of the function. It computes difference.

Such approach can't satisfy applicable timing as i can imagine, because there are some computation out of the `changeFrameAsync` and delay can't be correctly computed. But we can still get time one time somewhere in `changeFrameAsync` and put in a state and compute delay using previous one.

bsekachev:
@ActiveChooN 

Let me know when I can review the PR

cvat-ui/src/actions/annotation-actions.ts
```diff
@@ -27,25 +27,28 @@ function getStore(): Store<CombinedState> {
...
         frame,
+        showAllInterpolationTracks,
     };
 }
 
 export enum AnnotationActionTypes {
     GET_JOB = 'GET_JOB',
     GET_JOB_SUCCESS = 'GET_JOB_SUCCESS',
     GET_JOB_FAILED = 'GET_JOB_FAILED',
+    REMOVE_JOB = 'REMOVE_JOB',
```
bsekachev: Let's rename to CLOSE_JOB

bsekachev:
Issue with changing a job. Steps to reproduce. 

1) Open tasks page: ``http://localhost:3000/tasks``
2) Open a task, open a job
3) Press browser "Back" button twice
4) Open another task, open another job

Result: Previous job is opened.

bsekachev:
@ActiveChooN 
Probably we should skip some number of frames in ``fast`` and ``fastest`` speed. 
At least such approach in our current implementation. 
For example when it is ``fast``, play with step 2, when it is ``fastest``, play with step 4.

cvat-ui/src/components/annotation-page/top-bar/top-bar.tsx
```diff
@@ -20,6 +20,7 @@ interface Props {
     frameNumber: number;
     startFrame: number;
     stopFrame: number;
+    annotations: any;
```
bsekachev: Unused

cvat-ui/src/components/annotation-page/top-bar/top-bar.tsx
```diff
@@ -48,6 +49,7 @@ function AnnotationTopBarComponent(props: Props): JSX.Element {
         frameNumber,
         startFrame,
         stopFrame,
+        annotations,
```
bsekachev: Unused

cvat-ui/src/containers/annotation-page/top-bar/top-bar.tsx
```diff
@@ -105,30 +130,84 @@ function mapDispatchToProps(dispatch: any): DispatchToProps {
...
+            resetZoom,
+        } = this.props;
+
+        onChangeFrame(newFrame);
+
+        if (resetZoom) {
+            canvasInstance.html().addEventListener('canvas.setup', () => {
+                canvasInstance.fit();
+            }, { once: true });
+        }
```
bsekachev: I would prefer do it in a ``canvas-wrapper`` component. 
In this case we have one extra relation (top bar container knows about canvas).

cvat-ui/src/components/task-page/job-list.tsx
```diff
@@ -24,22 +27,45 @@ const baseURL = core.config.backendAPI.slice(0, -7);
...
         key: 'job',
-        render: (id: number): JSX.Element => (<a href={`${baseURL}/?id=${id}`}>{ `Job #${id}` }</a>),
+        render: (id: number): JSX.Element => (
+            <div>
+                <Button
+                    type='link'
+                    onClick={(): void => {
+                        if (currentJobId !== id) {
+                            onJobremove();
+                        }
```
bsekachev: Probably it would better if implement it in ``getJobAsync`` function.
Just dispatch one more action if another task is already opened. 
It would look much easier.

cvat-ui/src/containers/annotation-page/top-bar/top-bar.tsx
```diff
@@ -344,6 +443,7 @@ class AnnotationTopBarContainer extends React.PureComponent<Props> {
             jobInstance: {
                 startFrame,
                 stopFrame,
+                annotations,
```
bsekachev: Unused

cvat-ui/src/containers/annotation-page/top-bar/top-bar.tsx
```diff
@@ -370,6 +470,7 @@ class AnnotationTopBarContainer extends React.PureComponent<Props> {
                 startFrame={startFrame}
                 stopFrame={stopFrame}
                 frameNumber={frameNumber}
+                annotations={annotations}
```
bsekachev: Unused

cvat-ui/src/containers/annotation-page/top-bar/top-bar.tsx
```diff
@@ -336,6 +407,34 @@ class AnnotationTopBarContainer extends React.PureComponent<Props> {
...
 
+    private checkUnsavedChanges(): void {
+        const {
+            jobInstance,
+            history,
+        } = this.props;
+
+        jobInstance.annotations.hasUnsavedChanges().then((unsaved: boolean) => {
+            if (unsaved) {
+                window.addEventListener('beforeunload', AnnotationTopBarContainer.beforeUnloadCallback);
```
bsekachev: In this case can we have a lot of copies of event listeners for ``beforeunload`` to be setup?

cvat-ui/src/components/task-page/job-list.tsx
bsekachev: One more option which looks good is to use ReactRouter::history::listen to close a previous job.

cvat-ui/src/containers/annotation-page/top-bar/top-bar.tsx
ActiveChooN: In case callback functions are the same object answer is no. But we can register few callbacks for `beforeunload` with different functions.

cvat-ui/src/containers/annotation-page/top-bar/top-bar.tsx
```diff
@@ -15,18 +18,23 @@ import {
...
     frameStep: number;
+    frameSpeed: FrameSpeed;
+    frameDelay: number;
     playing: boolean;
     saving: boolean;
     canvasIsReady: boolean;
     savingStatuses: string[];
     undoAction?: string;
     redoAction?: string;
+    resetZoom: boolean;
```
bsekachev: ``resetZoom`` is now unused here, isn't it?

cvat-ui/src/containers/annotation-page/top-bar/top-bar.tsx
bsekachev: Got it

cvat-ui/src/actions/annotation-actions.ts
```diff
@@ -109,14 +113,16 @@ export enum AnnotationActionTypes {
     CHANGE_ANNOTATIONS_FILTERS = 'CHANGE_ANNOTATIONS_FILTERS',
     FETCH_ANNOTATIONS_SUCCESS = 'FETCH_ANNOTATIONS_SUCCESS',
     FETCH_ANNOTATIONS_FAILED = 'FETCH_ANNOTATIONS_FAILED',
+    ROTATE_FRAME = 'ROTATE_FRAME',
```
bsekachev: Do you actually need this action in this PR?

bsekachev:
One problem was found
When I open a job  (video with 72 frames), set the fastest playing speed and press `play`:
Screenshot from 2020-02-26 10-23-58

cvat-ui/src/components/annotation-page/annotation-page.tsx
```diff
@@ -17,14 +19,24 @@ interface Props {
...
         getJob,
+        match: {
+            params,
+        },
     } = props;
 
-    if (job === null) {
+    const jid = +params.jid;
+
+    if (job === null || (job !== undefined && job.id !== jid)) {
```
bsekachev: Let's implement this logic in function ``mapStateToProps`` for ``annotations-page`` component. 
There already included router props like ``tid`` and ``jid``.
We need only add a simple condition:
```ts
return {
   ... 
   job: !job || jid === job.id ?  job : null,
}
```

cvat-ui/src/components/annotation-page/annotation-page.tsx
```diff
@@ -17,14 +19,24 @@ interface Props {
     getJob(): void;
 }
 
-export default function AnnotationPageComponent(props: Props): JSX.Element {
+type RoutePros = RouteComponentProps<{
+    tid: string;
```
bsekachev: You do not use ``tid``, don't you?

cvat-ui/src/containers/annotation-page/top-bar/top-bar.tsx
```diff
@@ -105,30 +125,78 @@ function mapDispatchToProps(dispatch: any): DispatchToProps {
...
+                        onChangeFrame(frameNumber + 1 + framesSkiped);
                     }
-                });
+                }, frameDelay);
             } else {
                 onSwitchPlay(false);
             }
         }
+
+        this.checkUnsavedChanges();
```
bsekachev: I have several thoughts here.

First of all the method ``componentDidUpdate()`` here is called after each change of annotations because of props ``undoAction`` and ``redoAction`` are changed. It isn't very obvious, is it? If in the future we can remove them from here (forgot about this fact), and it won't work. And even now there is examples of scenario when it doesn't work:

1) Open a job
2) Move an object
3) Save the job
4) Move an object 
5) Go to task link

Result: You do not get any confirmation messages. 

It is because in both cases ``undoAction`` is a string "Changed points" and the second time component isn't going to update because it sees that there aren't updated props. 

One more concern is that ``checkUnsavedChanges()`` can be extremely slow (especially when there are a lot of annotations), but in this implementation it is going to be called after any small change. UI in this case is slow and has a lot of lags.

cvat-ui/src/containers/annotation-page/annotation-page.tsx
```diff
@@ -74,7 +76,7 @@ function mapDispatchToProps(dispatch: any, own: OwnProps): DispatchToProps {
     };
 }
 
-function AnnotationPageContainer(props: StateToProps & DispatchToProps): JSX.Element {
+function AnnotationPageContainer(props: StateToProps & DispatchToProps & OwnProps): JSX.Element {
```
bsekachev: OwnProps is not needed here

bsekachev:
@ActiveChooN 
Let's resolve conflicts and merge it

PR #1168, opened 2020-02-20
doc: fix description of attribute CVAT XML format

artemisart:
I realized the format description was completely wrong after dumping a CVAT XML file.

### DIFF
--- a/cvat/apps/documentation/xml_format.md
+++ b/cvat/apps/documentation/xml_format.md
@@ -25,7 +25,15 @@ There are two different formats for annotation and interpolation modes at the mo
         <label>
           <name>String: name of the label (e.g. car, person)</name>
           <attributes>
-            <attribute>String: attributes for the label (e.g. @select=quality:good,bad)</attribute>
+            <attribute>
+              <name>String: attribute name</name>
+              <mutable>Boolean: mutable (allow different values between frames)</mutable>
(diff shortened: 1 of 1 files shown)


nmanovic:
@artemisart , thanks for the PR. But it isn't complete. We have two formats (for images and for videos). Both have the same problem.

Also I dumped a file and see the following error:

```
     <labels>
        <label>
          <name>car</name>
          <attributes>
            <attribute>
              <name>model</name>
              <mutable>False</mutable>
              <input_type>select</input_type>
              <default_value>a,b,c,d</default_value>
              <values>a,b,c,d</values>
            </attribute>
          </attributes>
        </label>
```

The default value is wrong. It should be 'a'. Values are separated by ','. Could you please help us to fix mentioned problems as well?

artemisart:
When I dump my files I get:
```
<label>
          <name>Name</name>
          <attributes>
            <attribute>
              <name>ok</name>
              <mutable>False</mutable>
              <input_type>radio</input_type>
              <default_value>unchecked</default_value>
              <values>unchecked
yes
no</values>
            </attribute>
          </attributes>
        </label>
```
Are you sure you don't have just one value 'a,b,c,d' ?That would be logical with the default value you get. Or maybe there is a difference between versions ?

Additionally for numbers I think values describe min,max,step instead of every possible value but I cannot find any doc on this.

PR #1177, opened 2020-02-21
Az/cvat proxy

azhavoro:


cvat_proxy/conf.d/cvat.conf.template
```diff
@@ -0,0 +1,37 @@
+server {
```
bsekachev: Let's rename the directory ``cvat_proxy`` to ``cvat-proxy`` just to be consistent

nmanovic: Let's merge as is but it should be consistent. Agree.

PR #1194, opened 2020-02-25
React UI: Changing color for a shape

bsekachev:
Screenshot from 2020-02-25 17-07-45


nmanovic:
@ActiveChooN , could you please review and test the patch?

PR #1199, opened 2020-02-26
Update base.txt

pziewiec:
required gitdb in version 0.6.4


nmanovic:
@pziewiec , what is the problem? We don't use GitDB directly. Why do we need the change?

pziewiec:
docker-compose build doesnt work now

nmanovic:
@pziewiec , a similar problem was fixed recently: https://github.com/opencv/cvat/issues/1152. Is it a new issue?

pziewiec:
You right - the same problem

PR #1207, opened 2020-02-27
[Datumaro] Move project directory

zhiltsov-max:
- Project layout is now `any_dir/.datumaro/config.yaml`, i.e. corresponds to the documented one
- In the case of errors during commands, generated directories are removed

### DIFF
--- a/datumaro/datumaro/cli/contexts/project/__init__.py
+++ b/datumaro/datumaro/cli/contexts/project/__init__.py
@@ -10,7 +10,8 @@ import os
 import os.path as osp
 import shutil
 
-from datumaro.components.project import Project, Environment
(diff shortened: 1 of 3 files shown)


datumaro/datumaro/cli/contexts/project/__init__.py
```diff
@@ -138,19 +142,23 @@ def build_import_parser(parser_ctor=argparse.ArgumentParser):
 
 def import_command(args):
     project_dir = osp.abspath(args.dst_dir)
```
nmanovic: Looks like code duplication. It seems that `import` command should be implemented using `create` command with appropriate arguments.

PR #1213, opened 2020-02-28
[Datumaro] Add random split transform

zhiltsov-max:
Added random split transform.

```bash
datum project transform -p test_project -t random_split -- --subset train:0.6 --subset test:0.4
```

### DIFF
--- a/datumaro/datumaro/plugins/transforms.py
+++ b/datumaro/datumaro/plugins/transforms.py
@@ -5,6 +5,7 @@
 
 import logging as log
 import os.path as osp
+import random
 
 import pycocotools.mask as mask_utils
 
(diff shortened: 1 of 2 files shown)


datumaro/tests/test_transforms.py
```diff
@@ -320,3 +320,40 @@ def __iter__(self):
...
+        self.assertEqual(3, len(actual.get_subset('test')))
+
+    def test_random_split_gives_error_on_non1_ratios(self):
+        class SrcExtractor(Extractor):
+            def __iter__(self):
+                return iter([DatasetItem(id=1)])
+
+        has_error = False
+        try:
+            transforms.RandomSplit(SrcExtractor(), splits=[
```
nmanovic: Probably you can use `self.assertRaises` here

datumaro/datumaro/plugins/transforms.py
```diff
@@ -295,6 +296,66 @@ def transform_item(self, item):
...
+            help="Subsets in the form of: '<subset>:<ratio>' (repeatable)")
+        parser.add_argument('--seed', type=int, help="Random seed")
+        return parser
+
+    def __init__(self, extractor, splits, seed=None):
+        super().__init__(extractor)
+
+        total_ratio = sum((s[1] for s in splits), 0)
+        if not total_ratio == 1:
+            raise Exception(
```
nmanovic: It is better to throw `ValueError` exception.

datumaro/datumaro/plugins/transforms.py
```diff
@@ -295,6 +296,66 @@ def transform_item(self, item):
...
+        parser.add_argument('-s', '--subset', action='append',
+            type=cls._split_arg, dest='splits',
+            help="Subsets in the form of: '<subset>:<ratio>' (repeatable)")
+        parser.add_argument('--seed', type=int, help="Random seed")
+        return parser
+
+    def __init__(self, extractor, splits, seed=None):
+        super().__init__(extractor)
+
+        total_ratio = sum((s[1] for s in splits), 0)
```
nmanovic: Can you have problems here with fp precision? https://docs.python.org/3/library/math.html#math.isclose

PR #1220, opened 2020-03-01
Adding singularity compose deployment

vsoch:
This pull request will add work to deploy the containers with Singularity Compose, including container recipes, and documentation. I developed this a few weeks ago (just squashed and merged onto my development branch) to help deploy the library on Nero. Per request of @cleemesser I'm opening a pull request here.

Note that there is a recent bug with [gitdb2](https://github.com/gitpython-developers/GitPython/issues/983#issuecomment-586952758) that would impact any sort of build. It seems to involve version conflicts between gitdb2 and gitpython, and my first suggestion would be to try the versions in the comment linked.

Signed-off-by: Vanessa Sochat <vsochat@stanford.edu>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 # Project Specific
+/volumes/
 /data/
 /models/
 /share/
@@ -13,6 +14,11 @@
 !/ssh/README.md
 node_modules
 /Mask_RCNN/
+
+# Singularity Containers
+*.sif
+etc.hosts
+resolv.conf
 
 # Ignore temporary files
 docker-compose.override.yml
--- a/Singularity
+++ b/Singularity
@@ -1,0 +1,178 @@
+Bootstrap: docker
+From: ubuntu:16.04
+
+# This is the Singularity container to build the cvat image.
+# It would be more ideal to have a shared docker base to start from
+# but this doesn't currently exist.
+
+%environment
+# These were previously Docker build args
+export CUDA_SUPPORT="no"
+export WITH_DEXTR="no"
+export http_proxy=
+export https_proxy=
+export no_proxy=
+export socks_proxy=
+export TF_ANNOTATION="no"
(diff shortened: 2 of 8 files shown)


vsoch:
A quick comment about the line length - you run into trouble with github addresses for a specific file and commit, for which the link alone is > 120 characters. The only way to satisfy the linter in these cases is to completely remove of the link, and of course with it you remove the reference for the reader to learn something about a design choice, etc. In this case I removed a link to the Singularity (sylabs) code base that showed why instance names can't have dashes. This is easy enough to state, but if a user wants to find the code it would be a bit of a search.

Looks like tests are passing now! :)

nmanovic:
@vsoch , thanks for the pull request. It is great to see the contribution. But the problem here is our team cannot support several ways to run the app. It is really huge overhead for us. We are going to redesign the container system (create more containers instead of running multiple processes inside cvat container). Thus somebody should update the code. Also need to test the build with different components, ...

vsoch:
I've tested locally - the request came to me to use this application on the Nero cluster, for which Singularity would be required. The current PR work here will work for those versions, and it's designed to not create redundancy in the assets. If you want to throw away the work you can do that too, but I don't see why it wouldn't work to maintain with the present version, and then ask for help when you want to update or make changes.

vsoch:
If you like, we can continue working on my fork (researchapps) and then try to keep up to date with your master. Would that be more appropriate? I apologize that I assumed that the request came from your group (I just asked and it didn't).

nmanovic:
> If you like, we can continue working on my fork (researchapps) and then try to keep up to date with your master. Would that be more appropriate? I apologize that I assumed that the request came from your group (I just asked and it didn't).

@vsoch, let's do that for now. If you can contribute a PR into CVAT with reference on your branch with Singularity support it will be awesome.

vsoch:
Sure thing, will do!

PR #1225, opened 2020-03-02
React UI: Filters history

ActiveChooN:
You can choose any of last ten filters
image

### DIFF
--- a/cvat-ui/src/components/annotation-page/standard-workspace/objects-side-bar/objects-list-header.tsx
+++ b/cvat-ui/src/components/annotation-page/standard-workspace/objects-side-bar/objects-list-header.tsx
@@ -63,6 +63,7 @@ interface Props {
     statesCollapsed: boolean;
     statesOrdering: StatesOrdering;
     annotationsFilters: string[];
+    annotationsFiltersHistory: string[];
     changeStatesOrdering(value: StatesOrdering): void;
     changeAnnotationsFilters(value: SelectValue): void;
     lockAllStates(): void;
@@ -76,6 +77,7 @@ interface Props {
 function ObjectListHeader(props: Props): JSX.Element {
     const {
         annotationsFilters,
+        annotationsFiltersHistory,
         statesHidden,
         statesLocked,
         statesCollapsed,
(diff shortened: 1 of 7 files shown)


cvat-ui/src/reducers/annotation-reducer.ts
```diff
@@ -61,6 +61,7 @@ const defaultState: AnnotationState = {
         collapsed: {},
         states: [],
         filters: [],
+        filtersHistory: [],
```
bsekachev: In old UI we use local storage to save filters between sessions.
Let's do the same here.

cvat-ui/src/components/annotation-page/standard-workspace/objects-side-bar/objects-list-header.tsx
```diff
@@ -105,9 +107,12 @@ function ObjectListHeader(props: Props): JSX.Element {
                                 <span style={{ marginLeft: 5 }}>Annotations filter</span>
                             </>
                         )}
-                        dropdownStyle={{ display: 'none' }}
                         onChange={changeAnnotationsFilters}
-                    />
+                    >
+                        {annotationsFiltersHistory.map((element: string): JSX.Element => (
+                            <Select.Option key={element} value={element}>{element}</Select.Option>
```
bsekachev: Let's add comment here.
Something like: ``it is guaranteed, that each filter is unique``
because keys must be unique

cvat-ui/src/reducers/annotation-reducer.ts
```diff
@@ -947,10 +948,19 @@ export default (state = defaultState, action: AnyAction): AnnotationState => {
         }
         case AnnotationActionTypes.CHANGE_ANNOTATIONS_FILTERS: {
             const { filters } = action.payload;
+            const { filtersHistory, filters: oldFilters } = state.annotations;
```
bsekachev: Actually why do you need oldFilters here?
All old filters already should be in filtersHistory (after previous ``CHANGE_ANNOTATIONS_FILTERS`` action). Am I wrong?

ActiveChooN: @bsekachev if there are more than 10 active filters for example.

cvat-ui/src/reducers/annotation-reducer.ts
ActiveChooN: @bsekachev, maybe it can be useful to have one storage for user data like settings, history etc, can it?

bsekachev: @ActiveChooN 

Don't think about settings storage now. It was not a feature before and we should not hasten about it. Need to discuss it IRL before implementation

cvat-ui/src/reducers/annotation-reducer.ts
ActiveChooN: @bsekachev, I guess it's not so good
image

PR #1241, opened 2020-03-06
React UI: tags support

ActiveChooN:
### Added tags support to new UI 
Now you can add tags to annotation with button on the sidebar
image

And edit it with objects panel like other objects
image
TODO:
- [x] Fix filters
- [x] Fix `Ctrl+C` `N` shortcuts
- [x] Add color changing
- [x] Fix objects hiding
- [ ]  ~~add tags drawing to cvat-canvas~~


bsekachev:
Hi, some found issues:
1) Probably we should not display buttons "to background", "to foreground" at all for tags.
2) Filter doesn't work
3) Changing color for a tag doesn't work
4) N shortcut to create new tag doesn't work
5) Ctrl+C, Ctrl+B shortcuts don't work
6) Hide/show all button doesn't work correctly (try to create tag and hide/show all)
7) Left/Right arrow buttons (search with a filter) don't work correctly if there is a tag on a found frame (I mean next or previous)

ActiveChooN:
> Changing color for a tag doesn't work

Tags are not displayed on canvas so maybe no color changer is needed

ActiveChooN:
@bsekachev fixed issues, can you make review again?

bsekachev:
Sure. 
A couple of issues with copy:
1) Ctrl + V works only once after Ctrl + C. Is it expected?
2) It seems to be attributes aren't copied from source tag.

cvat-canvas/src/typescript/canvasView.ts
```diff
@@ -399,8 +399,11 @@ export class CanvasViewImpl implements CanvasView, Listener {
                 this.svgTexts[state.clientID].remove();
             }
 
-            this.svgShapes[state.clientID].off('click.canvas');
-            this.svgShapes[state.clientID].remove();
+            const shape = this.svgShapes[state.clientID];
```
bsekachev: Here and in other places in this file you check if ``points`` or ``shape`` exist before using it, right?
But why do you need it if canvas doesn't know about tags? (it doesn't draw them anyhow)

cvat-core/src/annotations-collection.js
```diff
@@ -810,11 +810,13 @@
                         'The object has not been saved yet. Call annotations.put([state]) before',
                     );
                 }
-
-                const distance = object.constructor.distance(state.points, x, y);
-                if (distance !== null && (minimumDistance === null || distance < minimumDistance)) {
-                    minimumDistance = distance;
-                    minimumState = state;
+                if (!(object instanceof Tag)) {
```
bsekachev: I would suggest to modify a condition above instead:

``
if (state.outside || state.hidden || state.objectType === ObjectType.TRACK) continue;
``

cvat-core/src/annotations-filter.js
```diff
@@ -165,18 +165,21 @@ class AnnotationsFilter {
             let xbr = Number.MIN_SAFE_INTEGER;
             let ytl = Number.MAX_SAFE_INTEGER;
             let ybr = Number.MIN_SAFE_INTEGER;
+            let [width, height] = [null, null];
+
+            if (state.objectType !== 'tag') {
```
bsekachev: Please, use ``ObjectType.TAG`` instead  (import it from enums)

cvat-ui/src/reducers/annotation-reducer.ts
```diff
@@ -437,6 +437,30 @@ export default (state = defaultState, action: AnyAction): AnnotationState => {
...
+                    ...state.canvas,
+                    activeControl,
+                },
+                drawing: {
+                    ...defaultState.drawing,
+                    activeLabelID: labelID,
+                    activeObjectType: objectType,
+                },
+            };
+        }
```
bsekachev: Why do not use ``AnnotationActionTypes.DRAW_SHAPE``? Actually it does the same as far as I understood.
We can rename it for better understanding (for example: ``REMEMBER_CREATE_PARAMETERS``)

cvat-ui/src/actions/annotation-actions.ts
```diff
@@ -1151,14 +1152,59 @@ export function searchAnnotationsAsync(
...
+        dispatch({
+            type: AnnotationActionTypes.ADD_TAG,
+            payload: {
+                labelID,
+                objectType: ObjectType.TAG,
+                activeControl: ActiveControl.CURSOR,
+            },
+        });
+
+        canvasInstance.cancel();
```
bsekachev: Probably it would be better to call ``canvasInstance.cancel()`` before dispatch
And also you do not need ```activeControl: ActiveControl.CURSOR``` because ``canvasInstance.cancel()`` does it.

cvat-ui/src/containers/annotation-page/standard-workspace/controls-side-bar/setup-tag-popover.tsx
```diff
@@ -0,0 +1,112 @@
...
+    };
+
+    private onSetup(): void {
+        const {
+            frame,
+            onAddTag,
+        } = this.props;
+
+        const { selectedLabelID } = this.state;
+
```
bsekachev:
```
const { labels, frame } = this.props;
const labelIndex = labels.map((label) => label.id).indexOf(selectedLabelID))
// call rememberCreateObject action
tag = new ObjectState({objectType: ObjectType.TAG, label: labels[labelIndex], frame })
// call createAnnotationsAsync with a tag
```

The key idea is probably we do not need additional actions at all

cvat-ui/src/components/annotation-page/standard-workspace/objects-side-bar/styles.scss
```diff
@@ -121,6 +121,7 @@
 .cvat-objects-sidebar-state-item-color {
     width: 7px;
     opacity: 1;
+    background-color: $default-object-colorpicker-item-background-color;
```
bsekachev: Do you actually need it?

vinorth05:
Hello, can you give an estimation of when the tags would be merged? One week? One month? more?

Thank you,
Best regards

cvat-ui/src/actions/annotation-actions.ts
```diff
@@ -138,6 +138,7 @@ export enum AnnotationActionTypes {
     SWITCH_Z_LAYER = 'SWITCH_Z_LAYER',
     ADD_Z_LAYER = 'ADD_Z_LAYER',
     SEARCH_ANNOTATIONS_FAILED = 'SEARCH_ANNOTATIONS_FAILED',
+    ADD_TAG = 'ADD_TAG',
```
bsekachev: extra line

ActiveChooN:
@vinorth05, thank you for your interest in CVAT project. I guess this PR will be finished and merged in few days

PR #1269, opened 2020-03-15
OpenVino 2020

benhoff:
I have a lack of ability to test this. Specifically I didn't test to make sure that OpenVINO 2019 still works with this, if the DEXTR model works, or if tf_annotation works.

I did load and test that this works with the auto_annotation model however. Just didn't want to leave the rest of it broken.


benhoff:
Should fix https://github.com/opencv/cvat/issues/1179

azhavoro:
@benhoff Hi, many thanks for the contribution! I'll test the patch with v2019.

benhoff:
@nmanovic, changelog updated.

PR #1007, opened 2019-12-25
Data streaming using chunks

azhavoro:
Known issues in this PR:

- [x] There are no get frame REST API methods.

- [x] Crash browser tab on 4k video after several navigation events (Possible solution: change the size of the decoded frames buffer dynamically).

- [x] Some number of  fast clicks (>10-20) on the navigation bar leads to the corresponding number of chunk request. Decoding begins after all requests have finished. Possible solution: implement the same logic as for cvat-data (process only the last request).

- [x] Build instruction for avc.wasm and Decoder.js as first step and dockerized build as final solution.

- [x] There are no access to `original quality` chunks and frames.

- [ ] Need to check the creation task time for both case (images and video) and compare with current version.

- [ ] Estimate network traffic for new and old version on the same data.

- [ ] Add small-size preview for the navigation slider.

- [x] Video playback is not smooth - player freezes and skips some frames after that. Partially fixed. Need to implement a buffering on the player side to make playback more smoothly.

- [x] Player issue: now player is requiring frames every 1/25s. It may leads to the following case: the frame which requested later will be received by the player earlier. (If decoder works slow on the large frame resolution).

- [x] Decoder potentially may has performance problems and decoding speed may be not enough to play the video smoothly, especially on the high resolution videos. Solution: use downscaled images on the backend and upscale on the frontend. Implemented https://github.com/ygnn123/cvat/pull/7/files

- [ ] Don't create decoder worker threads for each chunk, it leads to extra server requests, etc. (Checked: h264 decoder works fine).

- [x]  Access to the original data for old tasks issue (data migration)

- [x] Cleanup code, fix codacy issues.

- [x] auto_segmentation, reid and datumaro are broken 

- [x] compressed and original data contains incorrect data

PR #1274, opened 2020-03-16
[Datumaro] Fix frame matching in video annotations import

zhiltsov-max:
Add extra frame matching way for videos: frames can be named `frame_XXXXXX[.ext]`. This should be revised when #1007 is merged.
Fixed YOLO image id reading for provided image meta list.


zhiltsov-max:
@nmanovic, done.

PR #1276, opened 2020-03-16
Use source label map for voc export

### LINKED ISSUE #1275: VOC export of task can lead to error
How to reproduce:
1. Create a task with attached configuration of labels

[task_labels.txt](https://github.com/opencv/cvat/files/4338691/task_labels.txt)

2. Dump empty annotations as COCO or VOC. It will lead to different errors (e.g. FileNotFoundError, AssertionError)

zhiltsov-max:
Fixes #1275 

VOC export now does not use official label map by default, but takes one from the source task to avoid primary class and class part clashing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 -
 
 ### Changed
-- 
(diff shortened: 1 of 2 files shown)


zhiltsov-max:
@nmanovic, done.

PR #1288, opened 2020-03-18
React UI: Added logging

bsekachev:
Implemented almost all methods from the previous version except of:
- ``jumpFrame`` (i believe we do not need it because we have changeFrame)
- ``sendException`` (will be done in another request because we need add react error boundaries first)
- ``pressShortcut`` (since react-hotkeys doesn't allow to get combination. So, I don't see right now a right way to implement it) 

Added ``uploadAnnotations`` log that contains the same data like ``loadJob`` and ``sendTaskInfo``.


cvat-ui/src/actions/annotation-actions.ts
```diff
@@ -88,6 +89,23 @@ function computeZRange(states: any[]): number[] {
...
 }
 
+async function jobInfoGenerator(job: any): Promise<Record<string, number>> {
+    const { total } = await job.annotations.statistics();
+    return {
+        'frame count': job.stopFrame - job.startFrame + 1,
+        'track count': total.rectangle.shape + total.rectangle.track
+            + total.polygon.shape + total.polygon.track
+            + total.polyline.shape + total.polyline.track
+            + total.points.shape + total.points.track,
```
bsekachev: @azhavoro 
Could you please look at this code?
I did it like in the previous version, but I am not sure it's right
I tell about ``track count``

PR #1289, opened 2020-03-18
[Datumaro] MOT format

zhiltsov-max:
Added support for MOT detection det and gt formats for video sequences.

### DIFF
--- a/datumaro/datumaro/util/__init__.py
+++ b/datumaro/datumaro/util/__init__.py
@@ -4,6 +4,7 @@
 # SPDX-License-Identifier: MIT
 
 import os
+import os.path as osp
 
 
 def find(iterable, pred=lambda x: True, default=None):
@@ -18,3 +19,27 @@ def dir_items(path, ext, truncate_ext=False):
                 f = f[:ext_pos]
             items.append(f)
     return items
+
+def split_path(path):
+    path = osp.normpath(path)
+    parts = []
+
+    while True:
+        path, part = osp.split(path)
+        if part:
+            parts.append(part)
+        else:
+            if path:
+                parts.append(path)
+            break
+    parts.reverse()
+
+    return parts
+
+def cast(value, type_conv, default=None):
+    if value is None:
+        return default
+    try:
+        return type_conv(value)
+    except Exception:
+        return default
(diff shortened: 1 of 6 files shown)


datumaro/datumaro/util/__init__.py
```diff
@@ -17,4 +18,28 @@ def dir_items(path, ext, truncate_ext=False):
             if truncate_ext:
                 f = f[:ext_pos]
             items.append(f)
-    return items
\ No newline at end of file
+    return items
+
+def split_path(path):
```
nmanovic: I prefer something like code below:

```
def split_path(p):
    head, tail = os.path.split(p)
    return split_path(head) + [tail] if tail else [head]
```

datumaro/tests/test_mot_format.py
```diff
@@ -0,0 +1,146 @@
...
+
+        class DstExtractor(Extractor):
+            def __iter__(self):
+                return iter([
+                    DatasetItem(id=1,
+                        image=np.ones((16, 16, 3)),
+                        annotations=[
+                            Bbox(0, 4, 4, 8, label=2, attributes={
+                                'occluded': True,
+                                'visibility': 0.0,
```
nmanovic: Why here do we have visibility 0.0?

datumaro/tests/test_mot_format.py
```diff
@@ -0,0 +1,146 @@
...
+                            Bbox(1, 2, 4, 2, label=3, attributes={
+                                'occluded': False,
+                                'visibility': 1.0,
+                                'ignored': False,
+                            }),
+                        ]
+                    ),
+
+                    DatasetItem(id=3,
+                        image=np.ones((5, 4, 3)) * 3,
```
nmanovic: Why here we don't have attributes at all?

datumaro/tests/test_mot_format.py
zhiltsov-max: We had an object, which was `occluded`, but it's exact visibility was unknown.

datumaro/tests/test_mot_format.py
zhiltsov-max: `DatasetItem`s do not have `attribute`s yet.

datumaro/datumaro/util/__init__.py
zhiltsov-max: Nice variant, but it should probably include `normpath` call too, and have some handling for 1-element paths like ".." as it gives ['', '..'].

PR #1293, opened 2020-03-19
[Datumaro] LabelMe format

zhiltsov-max:
- Added LabelMe format support
- All plugin import errors are catched

### DIFF
--- a/datumaro/tests/test_labelme_format.py
+++ b/datumaro/tests/test_labelme_format.py
@@ -1,0 +1,120 @@
+import numpy as np
+
+from unittest import TestCase
+
+from datumaro.components.extractor import (Extractor, DatasetItem,
+    AnnotationType, Bbox, Mask, Polygon, LabelCategories
+)
+from datumaro.plugins.labelme_format import LabelMeImporter, LabelMeConverter
+from datumaro.util.test_utils import TestDir, compare_datasets
+
+
+class LabelMeConverterTest(TestCase):
+    def _test_save_and_load(self, source_dataset, converter, test_dir,
+            target_dataset=None, importer_args=None):
+        converter(source_dataset, test_dir)
+
+        if importer_args is None:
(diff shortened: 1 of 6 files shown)


datumaro/tests/test_labelme_format.py
```diff
@@ -0,0 +1,120 @@
...
+from unittest import TestCase
+
+from datumaro.components.extractor import (Extractor, DatasetItem,
+    AnnotationType, Bbox, Mask, Polygon, LabelCategories
+)
+from datumaro.plugins.labelme_format import LabelMeImporter, LabelMeConverter
+from datumaro.util.test_utils import TestDir, compare_datasets
+
+
+class LabelMeConverterTest(TestCase):
```
nmanovic: @zhiltsov-max , there is an opportunity to improve similar tests in Datumaro. Really want to see a small synthetic dataset in LabelMe format here (with actual xml files).

nmanovic: It is just an idea for future. Thus we can see that it actually can load such dataset. Also after load it you can save it again into another directory and "compare" using a cleaver algorithm. For now I don't see that the test cover the whole scenario.

nmanovic: You can generate a synthetic dataset using the original tools (to be sure that Datumaro can dump actual files).

zhiltsov-max: I don't think dataset files are actually can be compared easily for equality, or that they should be equal at all after import and export, but I agree that it isn't bad idea to check if actual small dataset can be read. It also gives an example of supported dataset layout. How do you think, should we store test datasets or we can generate them?

nmanovic: I believe it can be saved. Just keep some limitations like each dataset should be less than 10K in size (PNG should compress images if all pixels are the same).

PR #1297, opened 2020-03-20
React UI: Better exception handling

bsekachev:
* Added react boundary page which allows to restore broken UI 
* Added global error handler
* Added sendException logs 

Screenshot from 2020-03-23 12-37-34

### DIFF
--- a/cvat-ui/src/components/global-error-boundary/global-error-boundary.tsx
+++ b/cvat-ui/src/components/global-error-boundary/global-error-boundary.tsx
@@ -1,0 +1,165 @@
+// Copyright (C) 2020 Intel Corporation
+//
+// SPDX-License-Identifier: MIT
+
+import './styles.scss';
+import React from 'react';
+import { connect } from 'react-redux';
+import { Action } from 'redux';
+import { ThunkDispatch } from 'redux-thunk';
+import Result from 'antd/lib/result';
+import Text from 'antd/lib/typography/Text';
+import Paragraph from 'antd/lib/typography/Paragraph';
+import Collapse from 'antd/lib/collapse';
+import TextArea from 'antd/lib/input/TextArea';
+import Tooltip from 'antd/lib/tooltip';
+import copy from 'copy-to-clipboard';
+import ErrorStackParser from 'error-stack-parser';
+
+import { resetAfterErrorAsync } from 'actions/boundaries-actions';
+import { CombinedState } from 'reducers/interfaces';
+import logger, { LogType } from 'cvat-logger';
+
+interface StateToProps {
+    job: any | null;
+}
+
+interface DispatchToProps {
+    restore(): void;
+}
+
+interface State {
+    hasError: boolean;
+    error: Error | null;
+}
+
+function mapStateToProps(state: CombinedState): StateToProps {
+    return {
+        job: state.annotation.job.instance,
+    };
+}
+
+function mapDispatchToProps(dispatch: ThunkDispatch<CombinedState, {}, Action>): DispatchToProps {
+    return {
+        restore(): void {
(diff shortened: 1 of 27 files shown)


cvat-ui/src/components/global-error-boundary/global-error-boundary.tsx
```diff
@@ -0,0 +1,165 @@
...
+                            <ul>
+                                <li>
+                                    <Tooltip title='Copied!' trigger='click'>
+                                        {/* eslint-disable-next-line */}
+                                        <a onClick={() => {copy(message)}}> Copy </a>
+                                    </Tooltip>
+                                    the error message to clipboard
+                                </li>
+                                <li>
+                                    Notify your manager or submit the issue directly on
```
ActiveChooN: Maybe it can be useful to set more generic "administrator"

bsekachev:
@ActiveChooN, @nmanovic, @azhavoro, @zhiltsov-max 

Let's start since today enumerate versions of cvat-core, cvat-canvas and cvat-ui.
I can suggest following scheme:

```cvat-core```: 
update revision number ``x.y.[z]`` after bug fixes and some small changes (typos,descriptions, etc.)
update minor number ``x.[y].0`` after new functionality that doesn't break backward compatibility
update major number ``[x].0.0`` after new functionality that does break backward compatibility

``cvat-canvas`` and ``cvat-ui``:
minor and major number is the same like in used ``cvat-core`` to determine compatible core version
update revision number ``x.y.[z]`` after any changes that don't require new core version
OR
version consist of four numbers (the first two are responsible for core version, the third is responsible for new significant features, the last for minor bug fixes, typos, etc)

What do you think?

File: cvat-ui/src/components/global-error-boundary/global-error-boundary.tsx (hunk 1 above)
Comment (bsekachev) [reply cvat-ui/src/components/global-error-boundary/global-error-boundary.tsx]: Adjusted it

**Comment by zhiltsov-max** [comment]
You probably know about semantic versioning, but I'll just leave a link here: https://semver.org/

`cvat-core` - looks ok.

`cvat-canvas, -ui` are probably should just have their own versioning each, again with 3 numbers. The required `core` version can be just a convention for these modules, while the real dependency should be mentioned in some dependency manager (which will be `npm` I suppose).

The only question for me is will you be able to maintain this system easily, and, probably, how to automate the process. It is good for deployment, but what's with development and advancing few modules simultaneously (before publishing)?

**Comment by bsekachev** [comment]
@zhiltsov-max 

> should be mentioned in some dependency manager

Right now ``ui`` and ``canvas`` don't have ``core`` as a dependency in package json. They rely on ``cvat`` repository structure. But it is probably not very good practice. So, I'll investigate the issue. 

> how to automate the process

Some automation can be reached with npm commands like ``npm revision-update``, ``npm minor-update`` and something like this. 

> what's with development and advancing few modules simultaneously  

Not sure I understood you right. What if a patch requires change two modules or what if there are different patches in parallel?

**Comment by zhiltsov-max** [comment]
@bsekachev, 

> Not sure I understood you right...

Imagine the situation where we use dependency manager, which is probably the solution to go once we have distinct module versions. Will we be able to make changes to the local files in, for example, `cvat-core` and `cvat-ui`, and have the local debug server updated before we publish changes to the `develop` branch (or another public repo)?

**Comment by bsekachev** [comment]
@zhiltsov-max 

Local ``cvat-ui`` server is expected to be updated if we change its local files. I don't see any problems here.
Speaking about changing ``core`` and ``canvas`` it might be issue if we publish them on web.
But if we don't (and just include them as a local dependency)
```
npm install --save ../cvat-core
```

or add 
```
 "dependencies": {
    "cvat-core": "file:../cvat-core",
```
to package.json, it works.

PR #1299, opened 2020-03-23
[WIP] Fix data streaming

nmanovic:
- preview size was reduced till 128x128
- change implementation of progress notification for media extractors
- start, stop, step should work for images as well (untested)

### DIFF
--- a/cvat/apps/engine/media_extractors.py
+++ b/cvat/apps/engine/media_extractors.py
@@ -6,7 +6,7 @@ import os
 import tempfile
 import shutil
 import zipfile
-from io import BytesIO
+import io
 import itertools
 from abc import ABC, abstractmethod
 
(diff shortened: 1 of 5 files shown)


cvat/apps/engine/media_extractors.py
```diff
@@ -147,92 +150,84 @@ def __init__(self, source_path, step=1, start=0, stop=0):
...
+    def get_progress(self, pos):
         container = self._get_av_container()
         # Not for all containers return real value
-        length = container.streams.video[0].frames
-        return length
-
-    def __getitem__(self, k):
-        return next(itertools.islice(self, k, k + 1))
+        stream = container.streams.video[0]
+        return pos / stream.duration
```
azhavoro: I think it may be the same problem as for stream.frames, one of solutions: show number of extracted frames for status in this case. What do you think?

PR #1302, opened 2020-03-23
fixed image drawing for the both UI

azhavoro:


cvat-canvas/src/typescript/canvasView.ts
```diff
@@ -692,6 +692,7 @@ export class CanvasViewImpl implements CanvasView, Listener {
                         ctx.scale(image.renderWidth / image.imageData.width,
                             image.renderHeight / image.imageData.height);
                         ctx.putImageData(image.imageData, 0, 0);
+                        ctx.drawImage(this.background, 0, 0);
```
bsekachev: What is problem here?
Why ``putImageData`` is not enough?

azhavoro: > The current path, **transformation matrix**, shadow attributes, global alpha, the clipping region, and global composition operator must not affect the getImageData() and putImageData() methods.

from: https://www.w3.org/TR/2dcontext/#dom-context-2d-putimagedata

By this reason I have to redraw canvas to apply scale.

bsekachev: Is this fast operation? As far as I remember it was quite slow before

azhavoro: In this case putImageData takes ~1ms, drawImage < 1ms for fullHD and ~10 ms and <1ms for 4k.

bsekachev: It would be great if you left a comment in the code about why we need this line (just to do not forget)

azhavoro: Sure

PR #1303, opened 2020-03-23
Task creator validators & help message

bsekachev:


bsekachev:
@azhavoro 
So,
Git says I am not authorized to merge the PR even after approve. 
Can you?

PR #1306, opened 2020-03-23
Canvas failing and chunk loading status

ActiveChooN:


cvat-ui/src/actions/annotation-actions.ts
```diff
@@ -688,6 +688,7 @@ ThunkAction<Promise<void>, {}, {}, AnyAction> {
             });
 
             const data = await job.frames.get(toFrame, fillBuffer, frameStep);
+            if (data === undefined) return;
```
bsekachev: ``job.frames.get()`` is not expected to return ``undefined`` at all. @azhavoro right?
Probably there is a problem exact in this method

azhavoro: Yes, frames.get() should always return Instance of FrameData.

bsekachev:
I found one issue in this patch (at least I cannot reproduce it in data streaming branch)
Steps:
1) Start playing, wait till chunk is loaded end decoded
2) When playing is actually started, go to another frame

You can get an uncaught exception in console
Screenshot from 2020-03-25 15-46-31

It reproduces not every time, but quite often.

ActiveChooN:
@azhavoro Can you look at this? I guess this error does not apply to my PR.

azhavoro:
> @azhavoro Can you look at this? I guess this error does not apply to my PR.

ok

bsekachev:
@azhavoro 
Do you have any questions about streaming part? UI part looks good

cvat-core/src/frames.js
```diff
@@ -465,18 +465,12 @@
                     && cachedFrames.length < (this._size * 3) / 4) {
                     const maxFrame = cachedFrames ? Math.max(...cachedFrames) : frameNumber;
                     if (maxFrame < this._stopFrame) {
-                        this.makeFillRequest(maxFrame + 1, frameStep);
+                        await this.makeFillRequest(maxFrame + 1, frameStep);
```
azhavoro: Can we change this line as discussed today with smth like this?
```
this.makeFillRequest(maxFrame + 1, frameStep).catch(e => {
    if (e !== 'not needed') {
        throw e;
    }
});
```
Otherwise, we have to wait until `proactive` decoding resolve frames.

ActiveChooN: Guess it will not working as you expect. In simple example we still have same error.
изображение
@bsekachev what do you think about it?

bsekachev: @ActiveChooN 

Let's discuss it in skype

PR #1310, opened 2020-03-24
added validators for start_frame, stop_frane and chunk_size

azhavoro:


cvat/apps/engine/serializers.py
```diff
@@ -173,12 +173,38 @@ class Meta:
...
 
+    # pylint: disable=no-self-use
+    def validate_chunk_size(self, value):
+        if not value > 0:
+            raise serializers.ValidationError('Chunk size must be a positive integer')
+        return value
+
+    # pylint: disable=no-self-use
+    def validate_start_frame(self, value):
+        if not value > 0:
```
nmanovic: value >= 0?

cvat/apps/engine/serializers.py
```diff
@@ -173,12 +173,38 @@ class Meta:
...
+            raise serializers.ValidationError('Start frame must be a non negative integer')
+        return value
+
+    # pylint: disable=no-self-use
+    def validate_stop_frame(self, value):
+        if not value > 0:
```
nmanovic: What is the value for stop frame by default?

cvat/apps/engine/serializers.py
azhavoro: My bad, it's PositiveIntegerField at the Data model that can be 0 and manual validation is not needed in this case.

cvat/apps/engine/serializers.py
azhavoro: The same as above, I've deleted this validation.

PR #1314, opened 2020-03-24
React UI: Batch of fixes

bsekachev:
- Added validations for input number in different places
- Fixed some UI crashes
- Fixed minor bug with merging
- Added ability to undo points when editing
- Removed some old code

### DIFF
--- a/cvat-canvas/src/typescript/editHandler.ts
+++ b/cvat-canvas/src/typescript/editHandler.ts
@@ -96,10 +96,10 @@ export class EditHandlerImpl implements EditHandler {
         let mouseY: number | null = null;
 
         this.canvas.on('mousedown.edit', (e: MouseEvent): void => {
-            if (e.which === 1) {
+            if (e.button === 0) {
                 mouseX = e.clientX;
                 mouseY = e.clientY;
-            } else if (e.which === 3 && this.editLine) {
+            } else if (e.button === 2 && this.editLine) {
                 if (this.editData.state.shapeType === 'points'
                     || this.editLine.attr('points').split(' ').length > 2
                 ) {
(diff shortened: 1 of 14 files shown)


cvat-canvas/src/typescript/editHandler.ts
```diff
@@ -99,6 +99,12 @@ export class EditHandlerImpl implements EditHandler {
             if (e.which === 1) {
                 mouseX = e.clientX;
                 mouseY = e.clientY;
+            } else if (e.which === 3 && this.editLine) {
+                if (this.editData.state.shapeType === 'points'
+                    || this.editLine.attr('points').split(' ').length > 2
+                ) {
+                    (this.editLine as any).draw('undo');
+                }
```
ActiveChooN: Referenced to developer.mozilla.org/ru/docs/Web/API/MouseEvent/which, MouseEvent.which is deprecated, so it maybe better to use MouseEvent.button

cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/draw-shape-popover.tsx
```diff
@@ -117,7 +113,15 @@ function DrawShapePopoverComponent(props: Props): JSX.Element {
                         </Col>
                         <Col span={10}>
                             <InputNumber
-                                onChange={onChangePoints}
+                                onChange={(value: number | undefined) => {
+                                    if (typeof (value) === 'number') {
+                                        onChangePoints(Math.floor(
+                                            clamp(value, minimumPoints, Number.MAX_SAFE_INTEGER),
```
ActiveChooN: BTW, why is there two implementation of the clamp function? One in `utils/math.ts` other in `cvat-core/src/api.js`.

bsekachev: Despite everything ``core`` and ``ui`` are independent modules and can be divided in different repositories in the future

cvat-canvas/src/typescript/editHandler.ts
bsekachev: I didn't know. If it is, let's use button instead

PR #1315, opened 2020-03-25
Az/chunk size

azhavoro:


cvat/apps/engine/migrations/0025_auto_20200324_1222.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.2.10 on 2020-03-24 12:22
```
nmanovic: Can we splash the migration?

cvat/apps/engine/task.py
```diff
@@ -252,25 +253,42 @@ def update_progress(progress):
     compressed_chunk_writer = compressed_chunk_writer_class(db_data.image_quality)
     original_chunk_writer = original_chunk_writer_class(100)
 
+    # calculate chunk size if it isn't specified
+    if db_data.chunk_size is None:
+        if isinstance(compressed_chunk_writer, ZipCompressedChunkWriter):
+            w, h = extractor.get_image_size()
+            area = h * w
+            if area <= 1920 * 1080:
```
nmanovic: Does it make sense to have for small resolutions bigger chunk? Something like `max(1920 * 1080 * 18 // area * 2, 2)`

nmanovic: Maybe better to have `clip(1920 * 1080 * 18 // area * 2, 2, 72)`

cvat/apps/engine/task.py
```diff
@@ -252,25 +253,33 @@ def update_progress(progress):
     compressed_chunk_writer = compressed_chunk_writer_class(db_data.image_quality)
     original_chunk_writer = original_chunk_writer_class(100)
 
+    # calculate chunk size if it isn't specified
+    if db_data.chunk_size is None:
```
nmanovic: @azhavoro , will it be critical for current video decoder if somebody specified a wrong chunk_size (too small or even)?

cvat-core/webpack.config.js
```diff
@@ -52,7 +52,27 @@ const webConfig = {
                     sourceType: 'unambiguous',
                 },
             },
-        }],
+        }, {
+            test: /3rdparty\/.*\.worker\.js$/,
```
nmanovic: @azhavoro , why the change is necessary for the patch?

PR #1316, opened 2020-03-25
Fix AWS deployment 

nmanovic:
- Fix https://github.com/opencv/cvat/issues/1287
- Fix https://github.com/opencv/cvat/issues/1261 (the same as above)
- Fix https://github.com/opencv/cvat/issues/1264 (the same as above)

The idea of the fix is don't use antd.less (https://ant.design/docs/react/customize-theme). Processing of the file using less by webpack will lead to huge memory consumption. Thus the CVAT UI built will be failed (in some cases silently). It leads to many different problems (see issues above).

### DIFF
--- a/cvat-ui/package.json
+++ b/cvat-ui/package.json
@@ -32,8 +32,6 @@
     "eslint-plugin-react": "^7.17.0",
     "eslint-plugin-react-hooks": "^1.7.0",
     "html-webpack-plugin": "^3.2.0",
-    "less": "^3.11.1",
-    "less-loader": "^5.0.0",
     "node-sass": "^4.13.0",
     "postcss-loader": "^3.0.0",
     "postcss-preset-env": "^6.7.0",
--- a/cvat-ui/package.json
+++ b/cvat-ui/package.json
@@ -32,7 +32,7 @@
     "eslint-plugin-react": "^7.17.0",
     "eslint-plugin-react-hooks": "^1.7.0",
     "html-webpack-plugin": "^3.2.0",
-    "less": "^3.10.3",
+    "less": "^3.11.1",
     "less-loader": "^5.0.0",
     "node-sass": "^4.13.0",
     "postcss-loader": "^3.0.0",
@@ -42,7 +42,7 @@
     "style-loader": "^1.0.0",
     "tsconfig-paths-webpack-plugin": "^3.2.0",
     "typescript": "^3.7.3",
-    "webpack": "^4.41.2",
+    "webpack": "^4.42.1",
     "webpack-cli": "^3.3.8",
     "webpack-dev-server": "^3.8.0"
   },
--- a/cvat-ui/webpack.config.js
+++ b/cvat-ui/webpack.config.js
@@ -56,7 +56,7 @@ module.exports = {
                 options: {
                     importLoaders: 2,
                 },
-            }, 'postcss-loader', 'sass-loader']
+            }, 'postcss-loader']
         }, {
             test: /\.(css|scss)$/,
             exclude: /node_modules/,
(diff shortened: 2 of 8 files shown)


yohayonyon:
Hi @nmanovic,
I tried the code from your branch on my AWS and the build is still stuck.
Screenshot from 2020-03-25 16-56-22

nmanovic:
@yohayonyon , first of all how much time did you wait? For me the build takes some time. I used very slow machine from AWS with only 1G memory and it worked fine. Could you please provide the full log? Red warnings means nothing. Also `npm install` command downloads gigabytes of data. Performance of the step depends on internet speed.

yohayonyon:
Actually it happens pretty fast. I guess that the memory env var was defined.

Screenshot from 2020-03-25 21-35-00

cvat-ui/package.json
```diff
@@ -32,7 +32,7 @@
     "eslint-plugin-react": "^7.17.0",
     "eslint-plugin-react-hooks": "^1.7.0",
     "html-webpack-plugin": "^3.2.0",
-    "less": "^3.10.3",
+    "less": "^3.11.1",
```
bsekachev: Are ``less`` and ``less-loader`` still needed?

cvat-ui/webpack.config.js
```diff
@@ -50,13 +50,13 @@ module.exports = {
...
-                loader: 'less-loader',
+            test: /node_modules\/antd\/[\w\/]*.css$/,
+            use: ['style-loader', {
+                loader: 'css-loader',
                 options: {
-                    javascriptEnabled: true,
+                    importLoaders: 2,
                 },
-            }]
+            }, 'postcss-loader']
```
bsekachev: Maybe just removing ``exclued: /node_modules/`` in the rule below will work?

Or another option is to change ``importLoaders`` to ``1`` as it specified [here](https://github.com/webpack-contrib/css-loader#importloaders)

nmanovic:
> Actually it happens pretty fast. I guess that the memory env var was defined.
> 
> Screenshot from 2020-03-25 21-35-00

If you use 1G memory instance you need to down all containers before the build (otherwise you will get the problem above). I will recommend to use more powerful instance. In any case 1G isn't enough to annotate data and can lead to some problems. I would recommend at least 4G of memory.

yohayonyon:
When killing the containers the build is finished. Yet, the machine is reachable. the nginx welcome page is reached.

rasheeddo:
@yohayonyon I got that `cvat_ui failed to build` error before, can you try `docker compose down` then `sudo systemctl restart docker`, and build again? That works for me.

nmanovic:
@bsekachev ,

Looks like no difference between less and css version.

develop branch
```
nmanovic@Nikitas-MBP:~/Workspace/cvat/cvat-ui$ ls -lh dist/
total 24224
-rw-r--r--  1 nmanovic  staff   3.1M Mar 26 13:43 cvat-ui.min.js
-rw-r--r--  1 nmanovic  staff   8.7M Mar 26 13:43 cvat-ui.min.js.map
-rw-r--r--  1 nmanovic  staff   499B Mar 26 13:43 index.html
```

aws branch:
```
nmanovic@Nikitas-MBP:~/Workspace/cvat/cvat-ui$ ls -lh dist/
total 24272
-rw-r--r--  1 nmanovic  staff   3.1M Mar 26 13:48 cvat-ui.min.js
-rw-r--r--  1 nmanovic  staff   8.7M Mar 26 13:48 cvat-ui.min.js.map
-rw-r--r--  1 nmanovic  staff   499B Mar 26 13:48 index.html
```

bsekachev:
@nmanovic 
Great

PR #1318, opened 2020-03-25
React UI: Batch of fixes

### LINKED ISSUE #1284: Inconsistent labels between UI and CLI/API: "type" instead of "input_type"
UI shows `type` in generated labels but API and CLI expect `input_type` as in the swagger doc.

I created the label in the UI and copied the generated JSON. Manually and with the copy button.
```
[
  {
    "name": "label",
    "attributes": [
      {
        "name": "attr",
        "type": "select",
        "mutable": false,
        "values": [
          "val"
        ]
      }
    ]
  }
]
```
Using that JSON as a string with CLI fails to create the task with `500 Server Error: Internal Server Error for url: http://localhost:8080/api/v1/tasks`

```
python cli.py --auth admin:admin --server-host localhost --server-port 8080 create test_cli share "/data/1.jpg" "/data/2.jpg" --labels "[{\"name\": \"label\",\"attributes\": [{\"name\": \"attr\",\"type\": \"select\",\"mutable\": false,\"values\": [\"val\"]}]}]"
```

Changing `type` to `input_type` solves it. Even after task creation using `input_type` the UI shows `type`.

bsekachev:
- Fixed resize on right mouse button
- Fixed create object URL after first save, fixed URL itself
- Undo/redo returns frame where the change was

Resolved #1284


ActiveChooN:
@bsekachev 

>    * Undo/redo returns frame where the change was

How does it work? I'm trying to do something like that, but it does not work.

undo_return_to_frame

PR #1321, opened 2020-03-26
Az/fix for extractors

azhavoro:
Merge branch 'az/data_chunk_streaming' into nm/fix_data_streaming
fixed migration
fixed preview save
fixed stop frame for imagelist reader

### DIFF
--- a/cvat/apps/engine/media_extractors.py
+++ b/cvat/apps/engine/media_extractors.py
@@ -7,7 +7,6 @@ import tempfile
 import shutil
 import zipfile
 import io
-import itertools
 from abc import ABC, abstractmethod
 
 import av
@@ -60,7 +59,6 @@ class IMediaReader(ABC):
 
         return preview.convert('RGB')
 
-#Note step, start, stop have no affect
 class ImageListReader(IMediaReader):
     def __init__(self, source_path, step=1, start=0, stop=0):
         if not source_path:
@@ -69,7 +67,7 @@ class ImageListReader(IMediaReader):
         if stop == 0:
             stop = len(source_path)
         else:
-            stop = min(len(source_path), stop)
+            stop = min(len(source_path), stop + 1)
         step = max(step, 1)
         assert stop > start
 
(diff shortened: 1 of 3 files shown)


nmanovic:
@azhavoro , why so many changes? Can we simplify the patch?

Also need to check logic with _stop one more time. I believe it can be wrong in some cases (e.g. don't need to add +1 if user doesn't specify it because probably it is initialised correctly). In any case let's simplify the patch and I will review and check the logic.

azhavoro:
> @azhavoro , why so many changes? Can we simplify the patch?
> 
> Also need to check logic with _stop one more time. I believe it can be wrong in some cases (e.g. don't need to add +1 if user doesn't specify it because probably it is initialised correctly). In any case let's simplify the patch and I will review and check the logic.

Ok, lets split it to 2 PR: merge of data streaming branch and PR with changes of current branch.

cvat/apps/engine/media_extractors.py
```diff
@@ -58,9 +57,8 @@ def _get_preview(self, obj):
             preview = obj
         preview.thumbnail((128, 128))
 
-        return preview
+        return preview.convert('RGB')
```
nmanovic: Why do we need the line? What is the problem?

cvat/apps/engine/migrations/0024_auto_20191023_1025.py
```diff
@@ -68,7 +68,10 @@ def migrate_task_data(db_task_id, db_data_id, original_video, original_images, s
                     original_chunk_writer = Mpeg4ChunkWriter(100)
                     compressed_chunk_writer = ZipCompressedChunkWriter(image_quality)
 
-                    for chunk_idx, chunk_images in enumerate(reader.slice_by_size(chunk_size)):
+                    counter = itertools.count()
```
nmanovic: Will we have an issue with the code in the future? I don't see an alternative at the moment but if we change source code it may affect the migration.

cvat/apps/engine/media_extractors.py
azhavoro: I tried create a task with png RGBA image and got the following error
```
raise IOError("cannot write mode %s as JPEG" % im.mode)
OSError: cannot write mode RGBA as JPEG
```

cvat/apps/engine/migrations/0024_auto_20191023_1025.py
azhavoro: Potentially yes, but I cannot suggest an alternative at the moment except for duplicating the functions directly in the migration file.

PR #1328, opened 2020-03-29
Clean up memory for Auto Annotation

benhoff:
Trying to implement some circular buffers for auto annotation in order to avoid running out of RAM.

See here for the motivation: https://github.com/opencv/cvat/issues/1224

One thing that I don't understand is the `tracks` that is part of the `Results` class. 

https://github.com/opencv/cvat/blob/e7808cfb0322c1adcf61e7955b8b4a8c2badd0d2/cvat/apps/auto_annotation/inference.py#L34-L38

I assume that I need to implement a circular buffer for the `tracks` objects as well as the shapes. With this current PR, every 128 frames all the past tracks get thrown away. 

I assume this isn't desirable, but I've never used tracks, so I'm not sure what the expected behavior is.

@nmanovic, can you shed some light on what's expected here?


benhoff:
After doing some digging, it appears that `tracks` were added by this PR: https://github.com/opencv/cvat/pull/389

There is no reference to them in the auto annotation documentation. Prior to https://github.com/opencv/cvat/pull/389, `tracks` did not exist in the `Results` class. It looks like there is some sort of tracking functionality within CVAT somewhere, for videos? I'm not 100% sure, but it's mentioned in several issues.

The implementation of auto annotation currently does not record any of the tracking information. It gets dropped before the data is serialized. Maybe this is a future feature? Or a feature to help label shapes.

I'm not sure if it's used in the wild anywhere though. Which leads me to believe that implementing a circular buffer for shapes would be enough for now to fix the issues with annotating large videos. Thoughts?

HagegeR:
normally tracks have a kindof Time-Out property that deletes tracks when they did not get any update in the last X frames

benhoff:
Hmm, it would be possible to implement some sort of file based list interface, so that we could pretend that RAM is infinite. That would avoid the tradeoffs of figuring out the right amount of data to keep.

However, tracks don't have a public interface to add any data to them. So tacks looks like a future interface, and I'm not interested in over optimizing this PR. 

I'll clean up the commit history, do some additional testing, and submit this in close to it's current condition, without worrying about tracks.

zhiltsov-max:
@benhoff, what do you think about implementing and returning an iterator? Something like implemented in Datumaro here: https://github.com/opencv/cvat/blob/master/datumaro/datumaro/components/launcher.py#L32-L50
with an example of launcher here:  https://github.com/opencv/cvat/blob/master/datumaro/datumaro/plugins/openvino_launcher.py#L172-L179

benhoff:
@zhiltsov-max , I thought about it, but do to the serialization that is done I thought it was better to send the data in batches. 

Do you think that the overhead of serialization is small enough that it makes sense to implement an iterator?

See here for the serialization step:
https://github.com/opencv/cvat/blob/9344b067b1db54b641c36f651c4255c2857211f1/cvat/apps/auto_annotation/model_manager.py#L246

benhoff:
@zhiltsov-max, after looking at your code I realized that you were talking about a design optimization. If the patch works, I'd rather it gets merged. I don't want to spend more time on the patch unless it has bugs.

nmanovic:
@benhoff , thanks for the contribution! As usual your help is vital for our open source project.

@azhavoro, could you please look at the PR? Please test it and merge if it works as expected.

cvat/apps/auto_annotation/inference_engine.py
```diff
@@ -24,7 +23,7 @@ def make_plugin_or_core():
     use_core_openvino = False
     try:
         major, minor, reference = [int(x) for x in version.split('.')]
-        if major >= 2 and minor >= 1 and reference >= 37988:
```
nmanovic, line 27: @benhoff , why the condition isn't important anymore?

nmanovic:
@benhoff , about your questions regarding tracks. Automatic annotation doesn't support tracks. It detects shapes on each frame indivisually.

cvat/apps/auto_annotation/inference_engine.py
benhoff: The `reference` condition will continue to change. In order to test this, I had to compile dldt on my machine. My `reference` was completely different. 

My concern was that when they roll the release to a 3.1.XXXXX, the reference number might be smaller than `37988`, causing this line to malfunction.

This might also happen for a 2.2.XXXXX

PR #1331, opened 2020-03-30
Small preview and progress

azhavoro:


cvat/apps/engine/media_extractors.py
```diff
@@ -155,96 +154,88 @@ def __init__(self, source_path, step=1, start=0, stop=0):
...
-    def __iter__(self):
-        def decode_frames(container):
-            for packet in container.demux():
-                if packet.stream.type == 'video':
-                    for frame in packet.decode():
-                        yield frame
+    def _has_frame(self, i):
+        if i >= self._start:
+            if (i - self._start) % self._step == 0:
+                if self._stop == 0 or i <= self._stop:
```
nmanovic: @azhavoro , it seems that for images and for videos self._stop has different meaning (in case of images it doesn't include the latest frame). Can we easily unitfy?

azhavoro: already unified, we add +1 to the stop frame value in case of images.
https://github.com/opencv/cvat/blob/small_preview_and_progress/cvat/apps/engine/media_extractors.py#L75

nmanovic: @azhavoro , why in this case we have `i <= self._stop` here?

azhavoro: VideoReader class is not inherited from ImageListReader, stop + 1 doesn't affect VideoReader
stop value and iteration for ImageListReader:
```
...
stop = min(len(source_path), stop + 1)
...
for i in range(self._start, self._stop, self._step):
...
```

stop value and iteration for VideoReader:
```
...
stop=stop,
...
if i >= self._start:
    if (i - self._start) % self._step == 0:
        if self._stop == 0 or i <= self._stop:

```
Have I missed something?

nmanovic: @azhavoro , let me repeat the question: 

> @azhavoro , it seems that for images and for videos self._stop has different meaning (in case of images it doesn't include the latest frame). Can we easily unify?

nmanovic: @azhavoro , I just want to confirm that for videos we can use `i < self._stop` (or if we cannot let's change logic). In the future it will be very difficult to read the code and recall that two similar classes have different logic about the same variable.

azhavoro: @nmanovic I unified the iteration logic, also fixed corner case when user specified stop frame = 0.

PR #1339, opened 2020-03-31
React UI: Added dialog window with some help info about filters

bsekachev:
Screenshot from 2020-03-31 16-10-21
Screenshot from 2020-03-31 16-13-55


nmanovic:
@bsekachev , let's add the info into CHANGELOG

PR #1340, opened 2020-03-31
React UI: Special behaviour for the attribute value __undefined__

bsekachev:
- Invisible in UI
- Doesn't have a shortcut in AAM


nmanovic:
@ActiveChooN , it seems that a couple of patches from Boris need to review and test. Could you please allocate some time?

PR #1344, opened 2020-04-01
Fix point interpolation

### LINKED ISSUE #1336: dump annotation error - constructing empty shapely linestring
I'm getting this error when I try to dump my annotations:

Can not dump annotations for the task. Code: 500. Message: "cvat.apps.engine.utils.InterpreterError: ValueError at line 90: LineStrings must have at least 2 coordinate tuples\n"

strange thing is my annotations only have points, no polylines, so not sure why linestrings are being made. I'm a bit worried that I won't be able to export my work, that it will be lost.

attaching the docker log 
[cvat-log.txt](https://github.com/opencv/cvat/files/4406400/cvat-log.txt)

thanks!

zhiltsov-max:
Fixes #1336

- fixed single point interpolation

### DIFF
--- a/cvat/apps/engine/data_manager.py
+++ b/cvat/apps/engine/data_manager.py
@@ -293,7 +293,10 @@ class TrackManager(ObjectManager):
 
     @staticmethod
     def normalize_shape(shape):
-        points = np.asarray(shape["points"]).reshape(-1, 2)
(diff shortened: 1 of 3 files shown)


nmanovic:
@zhiltsov-max , could you please add the fix into CHANGELOG?

PR #1346, opened 2020-04-02
React UI: Implementation of bitmap in client

bsekachev:
Screenshot from 2020-04-02 13-33-33

Screenshot from 2020-04-02 13-35-15

Screenshot from 2020-04-02 13-35-09


ActiveChooN:
@bsekachev, if you place shape on the side of the canvas, shapes point still half visible, is it ok? And i have ability to change objects in bitmap mode, is this ok too?

bsekachev:
> @bsekachev, if you place shape on the side of the canvas, shapes point still half visible, is it ok? And i have ability to change objects in bitmap mode, is this ok too?

Both statements are expected in the implementation. I do not see any problems UX problems here. Do you?

ActiveChooN:
@bsekachev, no, just asking

PR #1349, opened 2020-04-02
React UI: Added option to display shape text always

bsekachev:
Screenshot from 2020-04-02 22-15-42
Screenshot from 2020-04-02 22-15-49

### DIFF
--- a/cvat-canvas/README.md
+++ b/cvat-canvas/README.md
@@ -108,7 +108,7 @@ Canvas itself handles:
         dragCanvas(enable: boolean): void;
         zoomCanvas(enable: boolean): void;
 
-        mode(): void;
+        mode(): Mode;
         cancel(): void;
         configure(configuration: Configuration): void;
     }
--- a/cvat-canvas/README.md
+++ b/cvat-canvas/README.md
@@ -50,6 +50,11 @@ Canvas itself handles:
         ZOOM_CANVAS = 'zoom_canvas',
     }
 
+    interface Configuration {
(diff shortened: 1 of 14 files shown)


cvat-ui/src/reducers/settings-reducer.ts
```diff
@@ -28,6 +28,7 @@ const defaultState: SettingsState = {
         autoSave: false,
         autoSaveInterval: 15 * 60 * 1000,
         aamZoomMargin: 100,
+        showObjectsTextAlways: true,
```
ActiveChooN, line 31: Why is it a good idea to enable this option by default?

bsekachev: I did it for test. Forgot to disable.

cvat-canvas/README.md
```diff
@@ -104,7 +108,9 @@ Canvas itself handles:
         dragCanvas(enable: boolean): void;
         zoomCanvas(enable: boolean): void;
 
+        mode(): void;
```
ActiveChooN, line 111: I see that `mode()` method returns Mode object

https://github.com/opencv/cvat/blob/feebec2670186f8f56f4314cb45d596d5dc13e8a/cvat-canvas/src/typescript/canvas.ts#L137-L139

Why did it change in documentation and Canvas interface?

bsekachev: It seems I did it accidently

PR #1355, opened 2020-04-03
Refactor frame provider

zhiltsov-max:
- refactored code to remove duplication
- probably, fixed bug with chunk caching - last loaded chunk id and reader were not stored, but passed


nmanovic:
@azhavoro , please review these changes.

PR #1367, opened 2020-04-06
Add chunk iterator cache to frame provider

zhiltsov-max:
- added chuck iterator cache to support random frame access without performance loss for sequential access
- added output type support for `get_frame()`


zhiltsov-max:
It may be better to put this logic into media extractors themselves.

PR #1396, opened 2020-04-13
Fixed uploading track annotations for multi-segment tasks

### LINKED ISSUE #1313: annotation upload corrupt for overlapping segments
i've noticed that if one has a video in CVAT which is segmented into several jobs and then uploads an annotation (containing annotations for one segment) the annotation cannot be changed and saved or deleted and saved if the annotation is bigger than the specific segment range. also the same is if there is an overlap between segments.

an example:

job 1: from 0 to 200 
job 2: from 195 to 395
job 3: from 390 to 590

so if the annotation has data for frames 195 to 395, the problem described above occurs.
also there will be a problem if one uploads an annotation for frames in the range 201 to 396.
im only able to edit and save the uploaded annotation if it is in the range of 201 to 395.

azhavoro:
## Fixed uploading track annotations for multi-segment tasks

resolve #1313 

### Motivation and context
Uploading track annotation in case of multi-segment task works incorrectly because imported tracked shapes is not filtered by frame and it leads to artifacts in the old UI and wrong dump file.
Also fixed bug with dump when track in the dump file may have no `closing` shape with `outside=True`.

### How has this been tested?
tested manually

### Checklist

- [x] I have raised an issue to propose this change ([required](https://github.com/opencv/cvat/issues))
- [x] My issue has received approval from the maintainers
- [x] I've read the [CONTRIBUTION](https://github.com/opencv/cvat/blob/develop/CONTRIBUTING.md) guide
- [x] I have linked related issues (#1313 )

### License

```python
#
```


nmanovic:
@azhavoro , do we have any related issues for the PR?

cvat/apps/annotation/annotation.py
```diff
@@ -83,7 +83,33 @@ def slice(self, start, stop):
         splitted_data = AnnotationIR()
         splitted_data.tags = copy.deepcopy(list(filter(is_frame_inside, self.tags)))
         splitted_data.shapes = copy.deepcopy(list(filter(is_frame_inside, self.shapes)))
-        splitted_data.tracks = copy.deepcopy(list(filter(lambda y: len(list(filter(is_frame_inside, y['shapes']))), self.tracks)))
+        splitted_data.tracks = []
+
+        for track_ in self.tracks:
+            if not list(filter(is_frame_inside, track_['shapes'])):
```
nmanovic, line 89: I don't look how the function is used by for the slice function the condition looks wrong. Let's say we have a track with two shapes only. First shape is on 0 frame and second shape is on 100 frame (0 - start, and 100 - outside). Let's say the slice function is called with start=10 and stop=20.  Looks like the track will be filtered out.

cvat/apps/annotation/annotation.py
```diff
@@ -83,7 +83,33 @@ def slice(self, start, stop):
...
+                continue
+
+            track = copy.deepcopy(track_)
+            track['frame'] = max(start, track['frame'])
+
+            segment_shapes = [s for s in track['shapes'] if is_frame_inside(s)]
+
+            if len(segment_shapes) < len(track['shapes']):
```
nmanovic, line 97: need to have a separate function, method "slice" for a track.

cvat/apps/annotation/annotation.py
```diff
@@ -83,7 +83,33 @@ def slice(self, start, stop):
...
+                interpolated_shapes = TrackManager.get_interpolated_shapes(track, 0, stop)
```
nmanovic, line 98: Why from 0?

CHANGELOG.md
```diff
@@ -43,6 +43,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - VOC format exports Upper case labels correctly in lower case (https://github.com/opencv/cvat/pull/1379)
 - Fixed polygon exporting bug in COCO dataset (https://github.com/opencv/cvat/issues/1387)
 - Task creation from remote files (https://github.com/opencv/cvat/pull/1392)
+- Uploading Track annotations for multi-segment tasks (https://github.com/opencv/cvat/pull/1396)
```
nmanovic, line 46:
```suggestion
- Uploading annotations for multi-segment tasks (https://github.com/opencv/cvat/pull/1396)
```

cvat/apps/engine/data_manager.py
```diff
@@ -133,6 +133,9 @@ def merge(self, objects, start_frame, overlap):
                     if j != -1:
                         self._modify_unmached_object(old_objects[j],
                             start_frame + overlap)
+                        # Need to reset cached interpolated shapes
+                        # because shapes was changed 
+                        old_objects[j]['interpolated_shapes'] = []
```
nmanovic, line 138: Should we have the condition inside _modify_unmached_object? Or it isn't a good solution by a reason?

cvat/apps/engine/data_manager.py
```diff
@@ -290,6 +290,9 @@ def _modify_unmached_object(obj, end_frame):
             shape["frame"] = end_frame
             shape["outside"] = True
             obj["shapes"].append(shape)
+            # Need to reset cached interpolated shapes
+            # because key shapes were changed
+            obj['interpolated_shapes'] = []
```
nmanovic, line 295: @azhavoro , is it a performance issue from your point of view? Just want to understand if the solution is good or bad. In general there are two possible solutions here:
- `obj['interpolated_shapes'] = []`
- interpolate till the end (in this case it is trivial because just need to copy the shape till end_frame).

cvat/apps/annotation/annotation.py
```diff
@@ -77,13 +77,58 @@ def serialize(self):
...
             return serializer.data
 
+    @staticmethod
+    def _is_shape_inside(shape, start, stop):
+        return start <= int(shape['frame']) <= stop
+
+    @staticmethod
+    def _is_track_inside(track, start, stop):
+        # should be a0 <= a1, b0 <= b1
+        def is_intervals_overlapped(a0, a1, b0, b1):
```
nmanovic, line 87:
```suggestion
        def has_overlap(a, b):
            return 0 <= min(b, stop) - max(a, start)  
```

cvat/apps/annotation/annotation.py
```diff
@@ -77,13 +77,58 @@ def serialize(self):
...
+            return 0 <= min(a1, b1) - max(a0, b0)
```
nmanovic, line 88:
```suggestion
```

cvat/apps/annotation/annotation.py
```diff
@@ -77,13 +77,58 @@ def serialize(self):
...
+
+        if not prev_shape["outside"] and prev_shape['frame'] <= stop:
+            return True
+
+        return False
+
+    @staticmethod
+    def _slice_track(track_, start, stop):
+        track = copy.deepcopy(track_)
+        segment_shapes = [s for s in track['shapes'] if AnnotationIR._is_shape_inside(s, start, stop)]
```
nmanovic, line 106: @azhavoro , if we have a track with only one shape with the frame less than start, segment_shapes will be empty.

cvat/apps/engine/data_manager.py
azhavoro: Let's use the interpolation approach, updated.

cvat/apps/annotation/annotation.py
azhavoro: this case will be handled later in 133 line

cvat/apps/engine/data_manager.py
```diff
@@ -290,6 +290,15 @@ def _modify_unmached_object(obj, end_frame):
             shape["frame"] = end_frame
             shape["outside"] = True
             obj["shapes"].append(shape)
+            # Need to update cached interpolated shapes
+            # because key shapes were changed
+            if "interpolated_shapes" in obj and obj["interpolated_shapes"]:
```
nmanovic, line 295:
```suggestion
            if obj.get("interpolated_shapes"):
```

cvat/apps/annotation/annotation.py
```diff
@@ -77,13 +77,57 @@ def serialize(self):
...
+
+        if len(segment_shapes) < len(track['shapes']):
+            interpolated_shapes = TrackManager.get_interpolated_shapes(track, start, stop)
+
+            for shape in interpolated_shapes:
+                if shape['frame'] == start and \
+                    (not segment_shapes or segment_shapes[0]['frame'] > start):
```
nmanovic, line 112: Is it possible that here `shape` will be with outside == True? Thus we will insert the shape but it looks wrong.

cvat/apps/annotation/annotation.py
```diff
@@ -77,13 +77,57 @@ def serialize(self):
         if serializer.is_valid(raise_exception=True):
             return serializer.data
 
+    @staticmethod
+    def _is_shape_inside(shape, start, stop):
+        return start <= int(shape['frame']) <= stop
```
nmanovic, line 82: One tricky case here. Should we return True here for shapes with outside == True?

cvat/apps/annotation/annotation.py
nmanovic: Potentially in the line we can get first shape with outside == True.

azhavoro: added filter to remove starting shapes with outside==True after processing

cvat/apps/annotation/annotation.py
azhavoro: added filter to remove starting shapes with outside==True after processing

cvat/apps/annotation/annotation.py
azhavoro: added filter to remove starting shapes with outside==True after processing

cvat/apps/annotation/annotation.py
```diff
@@ -77,13 +78,59 @@ def serialize(self):
...
+                    (not segment_shapes or segment_shapes[0]['frame'] > start):
+                    segment_shapes.insert(0, shape)
+                elif shape['frame'] == stop and \
+                    (not segment_shapes or segment_shapes[-1]['frame'] < stop):
+                    segment_shapes.append(shape)
+            drop_shape_count = sum(1 for _ in itertools.takewhile(lambda s: s['outside'], segment_shapes))
+            segment_shapes = segment_shapes[drop_shape_count:]
+            del track['interpolated_shapes']
+            for shape in segment_shapes:
+                del shape['keyframe']
```
nmanovic, line 122: Why should we delete the property for all segment_shapes? Could you please add a comment?

cvat/apps/annotation/annotation.py
```diff
@@ -77,13 +78,59 @@ def serialize(self):
...
+
+    @staticmethod
+    def _slice_track(track_, start, stop):
+        track = copy.deepcopy(track_)
+        segment_shapes = [s for s in track['shapes'] if AnnotationIR._is_shape_inside(s, start, stop)]
+
+        if len(segment_shapes) < len(track['shapes']):
+            interpolated_shapes = TrackManager.get_interpolated_shapes(track, start, stop)
+
+            for shape in interpolated_shapes:
```
nmanovic, line 111: @azhavoro , could you please think how to reorganize the code below? It looks messy. 

It seems that get_interpolated_shapes doesn't respect start and stop by a reason. Thus let's filter them:

```
scoped_shapes = [s for s in interpolated_shapes['shapes'] if AnnotationIR._is_shape_inside(s, start, stop)]
if scoped_shapes and scoped_shapes[0]['outside']:
    scoped_shapes.pop(0) 
```

If scoped_shapes empty then nothing to do. Otherwise if `not scoped_shapes[0]['keyframe']`, let's add the shape into segment_shapes at the beginning. If `not scoped_shapes[-1]['keyframe']`, lets' add the shape into segment_shapes at the end.

Let's assign `track['interpolated_shapes'] = scoped_shapes`.

Of course I didn't check the logic. You need just improve your code and I hope the code above will give you some ideas how to do that.

PR #1401, opened 2020-04-14
Update item "Types of shapes" in User Guide

TOsmanov:


cvat/apps/documentation/user_guide.md
```diff
@@ -382,19 +382,23 @@ The tool consists of:
     ![](static/documentation/images/image137.jpg)
 
 ### Types of shapes (basics)
+
 There are four shapes which you can annotate your images with:
-- ``box``
-- ``polygon``
-- ``polyline``
-- ``points``
-- ``auto segmentation``
```
nmanovic, line 390: Automatic segmentation will be available.

bsekachev: @nmanovic 

I would not say it is a separated shape. In the new UI it was implemented just as a checkbox with tooltip.
Screenshot from 2020-04-13 23-33-37

TOsmanov: But now it’s a way to create a polygon, I thought it would be more logical to move information about automatic segmentation to the section “Annotation with polygons”?

PR #1451, opened 2020-04-22
React UI: cuboids

ActiveChooN:
### Motivation and context

This PR transfer cuboid feature from old UI

изображение

### How has this been tested?

Manual testing

### Tasks
- [x] cvat-core: cuboids integration
- [x] cvat-core: cuboids unit-tests
- [x] cvat-ui: cuboid drawing components
- [x] cvat-ui: update statistics include info about cuboids
- [x] cvat-ui: cuboid projections configuration
- [x] cvat-canvas: cuboid drawing
- [x] cvat-canvas: cuboid restructuring for using only 8 canvas points
- [x] cvat-canvas: cuboid drag
- [x] cvat-canvas: cuboid resize (by @bsekachev )
- [x] cvat-canvas: cuboid opacity, selected, selected opacity, black stroke
- [x] cvat-canvas: cuboid copy/paste
- [x] cvat-canvas: cuboid merge, split, group

### Checklist

### License

```python
#
```


bsekachev:
@ActiveChooN 

I think we should add central points to resize. Resize using them significantly more convenient in some cases IMHO

bsekachev:
Great work!

Found only several issues during testing:
- Let's disable default pinned option for cuboids
- It seems to be z order doesn't work for cuboids in the new UI
- I found it convenient to resize the front corner points in two directions at the same time, but I am not sure about the back corner points. Maybe strict resizing only top/down for them? What do you think?
- Sometimes when I change frame I get exception in console:
```js
Uncaught (in promise) TypeError: Cannot read property 'center' of undefined
    at initializer.updateGrabPoints (svg.patch.ts:1036)
    at initializer.updateView (svg.patch.ts:992)
    at initializer.updateViewAndVM (svg.patch.ts:937)
    at initializer.attr (svg.patch.ts:838)
    at CanvasViewImpl.updateObjects (canvasView.ts:1098)
    at CanvasViewImpl.setupObjects (canvasView.ts:414)
    at CanvasViewImpl.notify (canvasView.ts:787)
    at CanvasModelImpl.notify (master.ts:41)
    at canvasModel.ts:365
```

this.getGrabPoints() returns an array of undefined values. I am not sure about stable way to reproduce. The same exception appears whenever I tried to paste a cuboid, so pasting doesn't work

- Finally in old UI when we hover mouse on a face, cube highlights the face and the front face. In new UI the whole cube is highlighted. I would suggest highlight only face under mouse but maybe there was a UI meaning. @nmanovic what do you think?

ActiveChooN:
@bsekachev 

> I found it convenient to resize the front corner points in two directions at the same time, but I am not sure about the back corner points. Maybe strict resizing only top/down for them? What do you think?

Sounds real good. Also we have vanishing point control on them so minimizing number of function is good way. Also @nmanovic proposed feature with reverse vanishing point control (case when back face is grater than front). What do you think about that?

 > this.getGrabPoints() returns an array of undefined values. I am not sure about stable way to reproduce. The same exception appears whenever I tried to paste a cuboid, so pasting doesn't work.

There is an issue with custom resize points. Patch should fix that

> It seems to be z order doesn't work for cuboids in the new UI

Hm, what do you meen? Can't reproduce any issues.

nmanovic:
> Finally in old UI when we hover mouse on a face, cube highlights the face and the front face. In new UI the whole cube is highlighted. I would suggest highlight only face under mouse but maybe there was a UI meaning. @nmanovic what do you think?

Now if mouse cursor under a specific place you can move it. I think it is better to highlight a specific place and get a visual hint what you can do with it something.

nmanovic:
- [ ] Need to improve improve editing experience. Follow Boris's recommendation I see a simple rule: if a point is placed on several planes at the same time it should adjust all of them in all possible directions. But it should be possible to keep one of arrows keys to look moving into one direction only for high-precision editing. What do you think?

nmanovic:
> * [ ]  Need to improve improve editing experience. Follow Boris's recommendation I see a simple rule: if a point is placed on several planes at the same time it should adjust all of them in all possible directions. But it should be possible to keep one of arrows keys to look move in one direction only for high-precision editing. What do you think?

Please ignore the comment. Need to think and experiment here. The problem I'm trying to solve is simple. In some cases it is not possible to move a place in some directions. Also in general I don't see a clear explanation why a specific point can move only in the specific direction.

bsekachev:
@ActiveChooN 

CuboidTrack is still pinned by default


>> Hm, what do you mean? Can't reproduce any issues.

I mean if you create two cubes and put one in front of another one, then increase zOrder value for one of them, the cube with increased zOrder value is expected to be in top when no one of them is activated. It is in old UI, but not in the new one (need to create task with ``zOrder`` activated)


>> proposed feature with reverse vanishing point control

Interesting, but I cannot imagine use case with 2D images where it is useful.

nmanovic:
@bsekachev , @ActiveChooN ,

One more idea (simple rule to follow):
- Each point can adjust cuboid only in 2 directions (because it is easy on 2D plane)
- Points in the corners can adjust only visible planes (move each in the direction which is parallel to plane's normal). Most of such points are on 1 or 2 visible planes and only one on 3 visible planes at the same time.
- Points in the middle of edges can adjust only visible planes. All such points on 1-2 visible planes. If both planes are invisible we don't draw a point (as we do now). Visible planes can be moved only in the direction which is parallel to the plane's normal.
- Invisible planes can be grabbed and moved by mouse in any direction (now it is particular true)
- What should we do with the point which is on 3 visible planes at the same time? It is an open question.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Re-Identification algorithm to merging bounding boxes automatically to the new UI (<https://github.com/opencv/cvat/pull/1406>)
 - Methods ``import`` and ``export`` to import/export raw annotations for Job and Task in ``cvat-core`` (<https://github.com/opencv/cvat/pull/1406>)
 - Versioning of client packages (``cvat-core``, ``cvat-canvas``, ``cvat-ui``). Initial versions are set to 1.0.0  (<https://github.com/opencv/cvat/pull/1448>)
+- Cuboids feature was migrated from old UI to new one.
```
bsekachev, line 12: Let's add PR link

nmanovic:
@ActiveChooN ,

It is a bug in old UI as well. But it should not be possible to draw cuboids like one below.

image

ActiveChooN:
@nmanovic do you mean the gap inside?

cvat-canvas/src/typescript/drawHandler.ts
```diff
@@ -236,14 +316,27 @@ export class DrawHandlerImpl implements DrawHandler {
     }
 
     private drawPolyshape(): void {
-        let size = this.drawData.numberOfPoints;
-        const sizeDecrement = function sizeDecrement(): void {
+        let size = this.drawData.shapeType === 'cuboid' ? 0 : this.drawData.numberOfPoints;
```
bsekachev, line 319: 'cuboid' ? 0 => 'cuboid' ? 4

cvat-canvas/src/typescript/drawHandler.ts
```diff
@@ -236,14 +316,27 @@ export class DrawHandlerImpl implements DrawHandler {
...
+            if (++size === 4) {
+                this.drawInstance.draw('done');
+            }
+        };
 
-        if (this.drawData.numberOfPoints) {
+        if (this.drawData.shapeType === 'cuboid') {
+            this.drawInstance.on('drawstart', sizeCuboidIncrement);
+            this.drawInstance.on('drawpoint', sizeCuboidIncrement);
+            this.drawInstance.on('undopoint', (): number => size--);
```
bsekachev, line 338: Probably we do not need this block and ``sizeCuboidIncrement``

nmanovic:
@ActiveChooN , @bsekachev ,

Please look at the example below. Should highlight visible edges?

image

cvat-canvas/src/typescript/drawHandler.ts
```diff
@@ -236,14 +316,27 @@ export class DrawHandlerImpl implements DrawHandler {
     }
 
     private drawPolyshape(): void {
-        let size = this.drawData.numberOfPoints;
-        const sizeDecrement = function sizeDecrement(): void {
+        let size = this.drawData.shapeType === 'cuboid' ? 0 : this.drawData.numberOfPoints;
+
+        if (this.drawData.shapeType === 'cuboid') this.addCrosshair();
```
bsekachev, line 321: Let's configure the crosshair with ``crosshair: boolean`` argument in DrawData

cvat-core/src/annotations-objects.js
```diff
@@ -1827,20 +1972,35 @@
...
+    class CuboidTrack extends PolyTrack {
+        constructor(data, clientID, color, injection) {
+            super(data, clientID, color, injection);
+            this.shapeType = ObjectShape.CUBOID;
+            for (const shape of Object.values(this.shapes)) {
+                checkNumberOfPoints(this.shapeType, shape.points);
+            }
+        }
+
+        // TODO: interpolation;
```
bsekachev, line 1984: Right now interpolation works based on algorithm for polyshapes?

bsekachev:
> Should highlight visible edges?

If we setup 100% opacity level for any shapes we don't see edges
Actually I do not see any problems here. This is one of reasons why option "Black borders" exists

cvat-core/src/annotations-objects.js
ActiveChooN: @bsekachev, no i've just adjusted algorithm for rectangles. I guess rectangle and cuboids are more similar.

bsekachev:
@nmanovic 

> One more idea (simple rule to follow):

I find current way of working with cuboids quite convenient. 
Maybe we should ask data annotation team

nmanovic:
> @nmanovic do you mean the gap inside?

Yes.

nmanovic:
> @nmanovic
> 
> > One more idea (simple rule to follow):
> 
> I find current way of working with cuboids quite convenient.
> Maybe we should ask data annotation team

Boris, I know because together with AlgoLux we went to the interface many months. But for me it is 
not intuitive why we move some sides in specific directions. No rules here. Just implemented this way by a reason. In some cases I cannot move a shape as I want and it is one more release.

We cannot ask our data annotation team because they don't annotate tasks with cuboids. I believe both ways will work. The only reason for the change is to make it more intuitive.

@ActiveChooN , how much time will it take to prototype the new behaviors from your perspective? We can commit the feature with the current interface but I will vote to experiment with the UX a little bit after that.

ActiveChooN:
@nmanovic 

> how much time will it take to prototype the new behaviors from your perspective?

It does not look very difficult, couple days, i guess.

> But it should not be possible to draw cuboids like one below.

And what about this problem. Is it about empty gap in cuboid between faces?

Tried to annotate few objects and have some thoughts. Dragging is very useful for adjusting cuboid track. But when you want to to just move cuboid in many cases you will encounter the problem that only one face will accidentally move. It happens when front face is covered by other one. So maybe we should prevent such behavior with face or cuboid dragging when only some key is pressed (shift for example).

bsekachev:
Small UX suggestion. Now front edge is always pink. I would suggest to choose another color (something more contrast for shape colors).
Screenshot from 2020-04-30 14-28-19

For example always black looks nice IMHO:
Screenshot from 2020-04-30 14-28-09

ActiveChooN:
@bsekachev we have black color for black borders feature. Maybe something about gray palette?

изображение

PR #1464, opened 2020-04-24
Ability to configure user agreements for the register user form

azhavoro:
### How has this been tested?
Manual testing

### Checklist

### License

```python
#
```


bsekachev:
@azhavoro 

How can I test it? 
I do not have any suggestions to agree with something during registering. It is probably because I do not have any agreements to be agree on the server

cvat-ui/src/reducers/interfaces.ts
```diff
@@ -105,6 +105,12 @@ export interface AboutState {
     initialized: boolean;
 }
 
+export interface UserAgreementsState {
+    userAgreements: any[];
```
bsekachev, line 109: I would suggest to describe an interface for a user agreement here and use it instead of any in ``cvat-ui``

cvat-ui/src/components/register-page/register-page.tsx
```diff
@@ -9,13 +9,15 @@ import Title from 'antd/lib/typography/Title';
...
+import RegisterForm, { RegisterData, UserAgreement } from './register-form';
 
 interface RegisterPageComponentProps {
     fetching: boolean;
+    userAgreements: any[];
     onRegister: (username: string, firstName: string,
         lastName: string, email: string,
-        password1: string, password2: string) => void;
+        password1: string, password2: string,
+        userAgreements: UserAgreement[]) => void;
```
bsekachev, line 20: Probably ``confirmations: UserConfirmation[]`` would be better.

cvat-ui/src/containers/register-page/register-page.tsx
```diff
@@ -6,21 +6,26 @@ import React from 'react';
...
-        password1: string, password2: string) => void;
+        password1: string, password2: string,
+        userAgreement: UserAgreement[]) => void;
 }
 
 function mapStateToProps(state: CombinedState): StateToProps {
     return {
-        fetching: state.auth.fetching,
+        fetching: state.auth.fetching || state.userAgreements.fetching,
+        userAgreements: state.userAgreements.userAgreements,
```
bsekachev, line 28: Looks confusing
Maybe: ``state.agreements.list``?

nmanovic:
@zhiltsov-max , could you please look at the patch?

cvat/apps/restrictions/admin.py
```diff
@@ -0,0 +1,8 @@
+# Copyright (C) 2020 Intel Corporation
```
zhiltsov-max, line 1: Do we really need to spawn empty files?

cvat/apps/restrictions/models.py
```diff
@@ -0,0 +1,8 @@
+# Copyright (C) 2020 Intel Corporation
```
zhiltsov-max, line 1: I believe accepted and declined agreements should be stored somewhere, but I can't find it.

cvat/apps/restrictions/tests.py
```diff
@@ -0,0 +1,8 @@
+# Copyright (C) 2020 Intel Corporation
```
zhiltsov-max, line 1: I suggest adding some tests, at least for the server.

cvat/settings/base.py
```diff
@@ -417,4 +417,8 @@ def generate_ssh_keys():
 LOCAL_LOAD_MAX_FILES_SIZE = 512 * 1024 * 1024  # 512 MB
 
 DATUMARO_PATH = os.path.join(BASE_DIR, 'datumaro')
-sys.path.append(DATUMARO_PATH)
\ No newline at end of file
+sys.path.append(DATUMARO_PATH)
+
+RESTRICTIONS = {
```
zhiltsov-max, line 422: Is it the place where agreements should be defined? If so, maybe some kind of plug-in folder (maybe with auto detection) could be better?

cvat-ui/src/reducers/interfaces.ts
azhavoro: done

azhavoro:
> @azhavoro
> 
> How can I test it?
> I do not have any suggestions to agree with something during registering. It is probably because I do not have any agreements to be agree on the server

Yes, you should change the default empty settings, an example can be found here https://github.com/opencv/cvat/blob/3e1748f1a169155fd9ca95a4b07086fad75815ae/cvat/apps/restrictions/tests.py

cvat/apps/restrictions/models.py
azhavoro: I'm not sure that we should store the agreements at the moment. For now if the user decline some of required agreement it cannot complete the registration and cannot continue to use the tool.
But I agree that this may be useful in the future.

cvat/apps/restrictions/tests.py
azhavoro: done

cvat/settings/base.py
azhavoro: Yes it's one of the possibles solution. I used the approach that we have used many times before.

cvat/apps/restrictions/models.py
zhiltsov-max: Isn't it required by some law?

cvat-ui/src/reducers/useragreements-reducer.ts
```diff
@@ -0,0 +1,50 @@
...
+            };
+        }
+        case UserAgreementsActionTypes.GET_USER_AGREEMENTS_SUCCESS:
+            return {
+                ...state,
+                fetching: false,
+                initialized: true,
+                list: action.payload,
+            };
+        case UserAgreementsActionTypes.GET_USER_AGREEMENTS_FAILED:
```
bsekachev, line 35: I would suggest to add this action also to notification-reducer (like for formats, etc.).
Otherwise if request is wrong, we do not see any notification about it and do not see any agreements on the register page

cvat-ui/src/reducers/notifications-reducer.ts
```diff
@@ -801,6 +805,21 @@ export default function (state = defaultState, action: AnyAction): Notifications
...
             };
         }
+        case UserAgreementsActionTypes.GET_USER_AGREEMENTS_FAILED: {
+            return {
+                ...state,
+                errors: {
+                    ...state.errors,
+                    formats: {
+                        ...state.errors.formats,
+                        fetching: {
```
bsekachev, line 815: probably ``userAgreements`` and ``...state.errors.userAgreements`` are right

azhavoro: sure, my bad
fixed

cvat/apps/authentication/auth.py
```diff
@@ -115,8 +115,8 @@ def is_job_owner(db_user, db_job):
 def is_job_annotator(db_user, db_job):
     db_task = db_job.segment.task
     # A job can be annotated by any user if the task's assignee is None.
-    has_rights = db_task.assignee is None or is_task_assignee(db_user, db_task)
-    if db_job.assignee is not None:
+    has_rights = (db_task.assignee is None and not settings.RESTRICTIONS['reduce_task_visibility']) or is_task_assignee(db_user, db_task)
+    if db_job.assignee is not None and not settings.RESTRICTIONS['reduce_task_visibility']:
```
nmanovic, line 119: Why do we need to modify the line? If db_job.assignee isn't None probably the user should be able to view the task/job. Right?

cvat/apps/restrictions/views.py
```diff
@@ -0,0 +1,40 @@
...
+    # to implement the method. By default, ViewSet doesn't provide it.
+    def get_serializer(self, *args, **kwargs):
+        pass
+
+    @staticmethod
+    @swagger_auto_schema(
+        method='get',
+        operation_summary='Method provides user agreements that the user must accept to register',
+        responses={'200': UserAgreementSerializer})
+    @action(detail=False, methods=['GET'], serializer_class=UserAgreementSerializer)
```
nmanovic, line 30: https://stackoverflow.com/questions/119312/urls-dash-vs-underscore

cvat/apps/restrictions/views.py
```diff
@@ -0,0 +1,40 @@
...
+    def user_agreements(request):
+        user_agreements = settings.RESTRICTIONS['user_agreements']
+        serializer = UserAgreementSerializer(data=user_agreements, many=True)
+        serializer.is_valid(raise_exception=True)
+        return Response(data=serializer.data)
+
+    @staticmethod
+    @action(detail=False, methods=['GET'], renderer_classes=(TemplateHTMLRenderer,))
```
nmanovic, line 38: https://stackoverflow.com/questions/119312/urls-dash-vs-underscore

cvat/apps/authentication/auth.py
azhavoro: Agree.

nmanovic:
@azhavoro , could you please confirm that you `have added description of my changes into CHANGELOG file`?

azhavoro:
> @azhavoro , could you please confirm that you `have added description of my changes into CHANGELOG file`?

@nmanovic done

PR #1474, opened 2020-04-28
Fixed 'Open task' button doesn't work

### LINKED ISSUE #1234: Task button doesn't work
I found that the task button in the annotation page  jump to `localhost:3000 ` in the most recent version on develope branch, which should be `CVAT_HOST:UI_PORT`. I found that the most recent version `docker-compost.yml` has no `UI_HOST` `UI_PORT` environment. A new `CVAT_HOST` environment has been specified.But In the code `cvat/settings/base.py` it still use `UI_HOST` and `UI_PORT` to get `UI_URL` .
```
UI_SCHEME = os.environ.get('UI_SCHEME', 'http')
UI_HOST = os.environ.get('UI_HOST', 'localhost')
UI_PORT = os.environ.get('UI_PORT', '3000')
CORS_ALLOW_CREDENTIALS = True
CSRF_TRUSTED_ORIGINS = [UI_HOST]
UI_URL = '{}://{}'.format(UI_SCHEME, UI_HOST)
```
It seems this cause the task button guide us to localhost:3000 in the annotation page.
I think this should be modify to use `CVAT_HOST` and the port should be 8080 in the newest version.

azhavoro:
## Fixed 'Open task' button in old annotation UI

resolve #1234 

### How has this been tested?
Manual testing

### Checklist

### License

```python
#
```

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -37,12 +37,7 @@
             ],
             "django": true,
             "cwd": "${workspaceFolder}",
-            "console": "internalConsole",
-            "env": {
-                "UI_SCHEME": "http",
-                "UI_HOST": "localhost",
-                "UI_PORT": "3000",
-            }
+            "console": "internalConsole"
         },
         {
             "name": "server: chrome",
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -37,7 +37,12 @@
             ],
             "django": true,
             "cwd": "${workspaceFolder}",
-            "console": "internalConsole"
+            "console": "internalConsole",
+            "env": {
+                "UI_SCHEME": "http",
+                "UI_HOST": "localhost",
+                "UI_PORT": "3000",
+            }
         },
         {
             "name": "server: chrome",
--- a/cvat/settings/base.py
+++ b/cvat/settings/base.py
@@ -187,21 +187,7 @@ MIDDLEWARE = [
     'dj_pagination.middleware.PaginationMiddleware',
 ]
 
-# Cross-Origin Resource Sharing settings for CVAT UI
-UI_SCHEME = os.environ.get('UI_SCHEME', 'http')
-UI_HOST = os.environ.get('UI_HOST')
-UI_PORT = os.environ.get('UI_PORT')
 UI_URL = ''
-if UI_HOST:
-    CORS_ALLOW_CREDENTIALS = True
-    CSRF_TRUSTED_ORIGINS = [UI_HOST]
-    UI_URL = '{}://{}'.format(UI_SCHEME, UI_HOST)
-
-    if UI_PORT and UI_PORT != '80':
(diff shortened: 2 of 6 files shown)


.vscode/launch.json
```diff
@@ -37,7 +37,12 @@
             ],
             "django": true,
             "cwd": "${workspaceFolder}",
-            "console": "internalConsole"
+            "console": "internalConsole",
+            "env": {
```
nmanovic, line 41: Why do we need these values? I can see that we have the by default in base.py.

nmanovic: OK. See some changes below. Please ignore the comment.

cvat/settings/base.py
```diff
@@ -189,17 +189,19 @@ def generate_ssh_keys():
...
-CORS_ALLOW_CREDENTIALS = True
-CSRF_TRUSTED_ORIGINS = [UI_HOST]
-UI_URL = '{}://{}'.format(UI_SCHEME, UI_HOST)
-
-if UI_PORT and UI_PORT != '80':
-    UI_URL += ':{}'.format(UI_PORT)
-
-CORS_ORIGIN_WHITELIST = [UI_URL]
-CORS_REPLACE_HTTPS_REFERER = True
+UI_HOST = os.environ.get('UI_HOST')
```
nmanovic, line 192: Why do we need to remove values by default?

azhavoro: Otherwise we will have to specify twice the same value for CVAT_HOST and UI_HOST that used for redirects by django backend, It's not good from my point of view.
In this patch I use the same approach as for UI server: default value is not specified and all urls will be relative, but these values are specified for the dev environments.
https://github.com/opencv/cvat/blob/3d4e7268e175557543932de115e0d7c03fae9e43/cvat-ui/package.json#L8

nmanovic: @azhavoro , let's try to  move the code to development.py configuration file.

azhavoro: @nmanovic Great idea! Done.

cvat/settings/base.py
```diff
@@ -187,19 +187,7 @@ def generate_ssh_keys():
...
-CORS_ALLOW_CREDENTIALS = True
-CSRF_TRUSTED_ORIGINS = [UI_HOST]
-UI_URL = '{}://{}'.format(UI_SCHEME, UI_HOST)
-
-if UI_PORT and UI_PORT != '80':
-    UI_URL += ':{}'.format(UI_PORT)
-
-CORS_ORIGIN_WHITELIST = [UI_URL]
-CORS_REPLACE_HTTPS_REFERER = True
+UI_URL = ''
```
nmanovic, line 190: Where is the URL used?

nmanovic: cvat/apps/authentication/decorators.py

PR #1479, opened 2020-04-30
Include empty images in exported annotations

### LINKED ISSUE #525: Export all images of a job even if no annotation is present
We have images of static scenes taken with different exposure time, lighting conditions, etc. We annotate only one image of a given scene and then duplicate the information using a Python script to the other pictures of the same scene.

The problem is that at the moment CVAT only export in the XML images that have at least one annotation. In order to workaround this, we draw a point in the corner of the first picture and propagate it to the whole set of picture and later filter out the points annotations. It would be nice if CVAT would also export images with empty annotations.

Something like the following:
```
<image height="2050" id="58" name="001-001.png" width="2448">
  <box label="A" occluded="0" xbr="2063.41" xtl="1832.21" ybr="1577.37" ytl="1269.10" />
  <box label="B" occluded="0" xbr="827.76" xtl="540.04" ybr="1895.92" ytl="1572.24" />
</image>
<image height="2050" id="58" name="001-002.png" width="2448" />
<image height="2050" id="58" name="001-003.png" width="2448" />
<image height="2050" id="58" name="001-004.png" width="2448" />
```

zhiltsov-max:
Closes #525 

### Motivation and context

This patch allows to export un-annotated images, so both annotated and un-annotated images are exported, where it makes sense to the target format.

### How has this been tested?

Unit tests

### Checklist

### License

```python
#
```

### DIFF
--- a/datumaro/datumaro/plugins/yolo_format/extractor.py
+++ b/datumaro/datumaro/plugins/yolo_format/extractor.py
@@ -93,12 +93,6 @@ class YoloExtractor(SourceExtractor):
                     (osp.splitext(osp.basename(p.strip()))[0], p.strip())
                     for p in f
                 )
-
-            for item_id, image_path in subset.items.items():
-                image_path = self._make_local_path(image_path)
-                if not osp.isfile(image_path) and item_id not in image_info:
-                    raise Exception("Can't find image '%s'" % item_id)
-
             subsets[subset_name] = subset
 
         self._subsets = subsets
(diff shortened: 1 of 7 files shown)


zhiltsov-max:
@nmanovic, done.

datumaro/datumaro/plugins/yolo_format/extractor.py
```diff
@@ -134,21 +127,30 @@ def _get(self, item_id, subset_name):
...
-                w = float(w)
-                h = float(h)
-                x = float(xc) - w * 0.5
-                y = float(yc) - h * 0.5
-                annotations.append(Bbox(
-                    round(x * image_width, 1), round(y * image_height, 1),
-                    round(w * image_width, 1), round(h * image_height, 1),
-                    label=label_id
-                ))
+                line = line.strip()
```
nmanovic, line 134: @zhiltsov-max , I don't think that it is a good idea to read the whole file into memory. I understand that the code below does it in any case but probably annotations.append can be clever in the future.

zhiltsov-max: These files contain annotations only for a single image, so it's unlikely to become a bottleneck. Whole dataset jsons in COCO hurt much more.

PR #1560, opened 2020-05-19
React UI: cuboid interpolation and cuboid drawing from rectangles

ActiveChooN:
from rectangles

### Motivation and context

This PR provide enabled cuboid tracks interpolation (bbox like) and drawing cuboids from rectangles.

rect2cuboid

### How has this been tested?

Manually

### Checklist

### License

```python
#
```


bsekachev:
I have some exceptions during drawing cube from rectangle: 

```
drawHandler.ts:480 Uncaught TypeError: Cannot read property 'update' of undefined
    at initializer.<anonymous> (drawHandler.ts:480)
    at initializer.fire (svg.js:3535)
    at PaintHandler.../cvat-canvas/node_modules/svg.draw.js/dist/svg.draw.js.PaintHandler.update (svg.draw.js:143)
    at svg.draw.js:83
(anonymous) @ drawHandler.ts:480
fire @ svg.js:3535
../cvat-canvas/node_modules/svg.draw.js/dist/svg.draw.js.PaintHandler.update @ svg.draw.js:143
(anonymous) @ svg.draw.js:83
drawHandler.ts:480 Uncaught TypeError: Cannot read property 'update' of undefined
    at initializer.<anonymous> (drawHandler.ts:480)
    at initializer.fire (svg.js:3535)
    at PaintHandler.../cvat-canvas/node_modules/svg.draw.js/dist/svg.draw.js.PaintHandler.update (svg.draw.js:143)
    at PaintHandler.../cvat-canvas/node_modules/svg.draw.js/dist/svg.draw.js.PaintHandler.stop (svg.draw.js:109)
    at PaintHandler.../cvat-canvas/node_modules/svg.draw.js/dist/svg.draw.js.PaintHandler.point (svg.draw.js:102)
    at initializer.draw (svg.draw.js:210)
    at initializer.constructor [as draw] (svg.patch.ts:35)
    at create.<anonymous> (drawHandler.ts:652)
```

bsekachev:
Quite strange bug with interpolation
Peek 2020-05-20 23-16

cvat-canvas/README.md
```diff
@@ -37,6 +37,11 @@ Canvas itself handles:
         EXTREME_POINTS = 'By 4 points'
     }
 
+    enum CuboidDrawingMethod {
+        CLASSIC = 'Classic cuboid draw method',
+        CORNER_POINTS = 'By 4 points',
```
bsekachev, line 42: What meaning do these messages mean? It isn't used in UI.

cvat-canvas/src/typescript/canvasModel.ts
```diff
@@ -46,6 +46,11 @@ export enum RectDrawingMethod {
     EXTREME_POINTS = 'By 4 points'
 }
 
+export enum CuboidDrawingMethod {
+    CLASSIC = 'Classic cuboid draw method',
+    CORNER_POINTS = 'By 4 points',
```
bsekachev, line 51: The same question about meaning

cvat-core/src/annotations-objects.js
```diff
@@ -1967,7 +1967,7 @@
         }
     }
 
-    class CuboidTrack extends PolyTrack {
+    class CuboidTrack extends Track {
```
bsekachev, line 1970: If we inherit CuboidTrack from Track, we should inherit CuboidShape from Shape

bsekachev:
I am unable to cancel drawing a cuboid (from rectangle). UI just freezes.

cvat-canvas/src/typescript/svg.patch.ts
```diff
@@ -860,7 +860,7 @@ function getTopDown(edgeIndex: EdgeIndex): number[] {
             });
             this.on('mouseover', () => {
                 edges.forEach((edge: SVG.Element) => {
-                    this.strokeOffset = 2.5;
+                    this.strokeOffset = this.node.classList.contains('cvat_canvas_shape_activated') ? 2.5 : 1.75;
```
bsekachev, line 863: Let's escape of magic numbers in the code. If you need a const, it is better to declare it in ``consts.ts`` file

cvat-canvas/README.md
ActiveChooN: @bsekachev, It's similar to `RectDrawingMethod` and have same naming and use cases. Can you explain problem with more details?

bsekachev: @ActiveChooN 

No problem. Just question. 
I expect that for example line ``Classic cuboid draw method`` is used in UI somewhere (for example near the checkbox field). But it isn't. So, in this case it plays only role of comment. 

And now I see that for ``RectDrawingMethod`` we also do not use these value to display them in UI (but we might use), but values in UI are equal to values of the enumeration.

bsekachev:
Client part looks great

@zhiltsov-max 
When I try dumping a task with cuboids in some cases I get error:
```Unknown shape type: 'cuboid'```. What do you think, is it okay? Maybe just ignore unsupported objects when we dump in a specific format?

zhiltsov-max:
@bsekachev, this is the problem we discussed (#1568). There is an excluding `if` switch, which should have covered cuboids when they were introduced, but it did not.

PR #1570, opened 2020-05-20
CVAT UI: tag annotation workspace

ActiveChooN:
### Motivation and context

This PR provide a new workspace for more comfortable tag annotation process.

изображение

### How has this been tested?

Manually

### Checklist

### License

```python
#
```


bsekachev:
@ActiveChooN 

In the implementation, when you add a tag, CVAT immediately switches frame. 
It can be convenient in tasks where we have strict condition: one frame, max one tag. But it creates significant inconveniences when frame can include several tags. What is annotation scenario in this case? 

- Also need to add a built-in search in label selector (antd should support it).
- Probably some icons for buttons "add tag" and "skip frame" would be useful to UX. 
- What do you think about case when the same frame includes several tags with the same label? Is it valid?
- Eslint shows quite a lot of issues in the patch. Could you please fix them?

Feature idea: 
We have enough free space in the sidebar. What if we show list of buttons with recently created labels? Pressing the button creates a tag with the corresponding label. So, you do not need to search often used labels in the list. It should accelerate the process. What's more, it is possible, to add shortcuts for these labels.

ActiveChooN:
@bsekachev 

> What is annotation scenario in this case?

I imagine it like that. User selects one tag for annotation, for example cat, and annotate all images with this tag. So there is two conditions, tag is present or not on this image. In both cases it will be good to go further without additional clicks. And when annotation with this tag is ended user selects another tag and run all process from the beginning.  

> What do you think about case when the same frame includes several tags with the same label? Is it valid?

I guess it will be good question for data annotators. In my opinion you can tag image with several tags with different attributes (for example cars with different color).

> Feature idea:

I have another idea. It can map keys 0-9 to first 10 labels of the task. Or maybe we should use your solution. But in my idea user should only annotate labels in that mode. For attributes they should use AAM.

bsekachev:
@ActiveChooN 

> User selects one tag for annotation, for example cat, and annotate all images with this tag. So there is two conditions, tag is present or not on this image. In both cases it will be good to go further without additional clicks. 

On the one hand yes. But what if dataset contains a lot of labels? In this case an annotator needs to go through all images N times. It looks as a very time consuming process. I suppose need to communicate with our team about the question. 

> I guess it will be good question for data annotators.

So, will you ask them?

> I have another idea.

In a dataset with a large number of labels often used labels aren't always first 10 labels. But actually I do not remember case when we annotate a dataset with number of labels > 10. So, first 10 labels also could be a great solution. By the way, we can allow annotators to setup these buttons as they want in the future.

PR #1463, opened 2020-04-24
Support relative paths in import and export

zhiltsov-max:
Depends on #1715

### Motivation and context

Added support for importing annotations with relative paths.
Fixed file name matching when importing annotations with relative paths and the target task has been created with images from file share.

How to test:
- create a task with relative paths
- import and export annotations. Images with same names lying in different directories should be handled normally.

### How has this been tested?

Unit tests

### Checklist

- [x] I submit my changes into the `develop` branch
- [x] I have added description of my changes into [CHANGELOG](https://github.com/opencv/cvat/blob/develop/CHANGELOG.md) file
- [ ] I have updated the [documentation](
  https://github.com/opencv/cvat/blob/develop/README.md#documentation) accordingly
- [x] I have added tests to cover my changes
- [ ] I have linked related issues ([read github docs](
  https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword))

### License

- [x] I submit _my code changes_ under the same [MIT License](
  https://github.com/opencv/cvat/blob/develop/LICENSE) that covers the project.
  Feel free to contact the maintainers if that's a concern.
- [ ] I have updated the license header for each file (see an example below)

```python
# Copyright (C) 2020 Intel Corporation
#
# SPDX-License-Identifier: MIT
```

PR #1636, opened 2020-06-03
Add directory check on image save

zhiltsov-max:
Preparation commit for #1463 

Added a check for directory existence and a parameter for `save_image` function to avoid silent fails when saving images into non existent dirs.

### DIFF
--- a/datumaro/datumaro/util/image.py
+++ b/datumaro/datumaro/util/image.py
@@ -5,11 +5,12 @@
 
 # pylint: disable=unused-import
 
+from enum import Enum
 from io import BytesIO
 import numpy as np
+import os
 import os.path as osp
 
-from enum import Enum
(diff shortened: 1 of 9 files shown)


datumaro/datumaro/util/image.py
```diff
@@ -45,7 +46,17 @@ def load_image(path):
...
-def save_image(path, image, **kwargs):
+def save_image(path, image, create_dir=False, **kwargs):
+    # NOTE: Check destination path for existence
+    # OpenCV silently fails if target directory does not exist
+    dst_dir = osp.dirname(path)
+    if dst_dir:
+        if create_dir:
+            os.makedirs(dst_dir, exist_ok=True)
+        elif not osp.isdir(dst_dir):
+            raise NotADirectoryError(
```
nmanovic, line 57: FileNotFoundError - Raised when a file or directory is requested but doesn’t exist. Corresponds to errno ENOENT.

PR #1696, opened 2020-06-10
Allow custom model inference and upload

### LINKED ISSUE #1028: Use custom model for tf_annotation
This is more of feature request rather than a bug. I would like to use my custom model trained using tf object detection( instead of Faster RCNN Inception Resnet v2 Atrous Coco Model from tensorflow detection model zoo to  "Run TF Annotation" . Is it possible to do this? Is there any documentation regarding this?

san999p:
This patch allows users to upload any compatible object detection or segmentation model and run automatic annotation. This will close #1028 

### Motivation and context

Right now, you can only upload custom models that are supported by OpenVINO. Tensorflow Object Detection API is widely used and it would be helpful if we allow support for custom model inference. The same goes for MaskRCNN.

For object detection, you can upload .pb file for model and .csv file for classes. For auto segmentation, you can upload .h5 for model and .csv for classes. The format of csv is slighly different for both of them which we can put it in the docs.

### How has this been tested?

This has been tested on local machine.

### Checklist

- [`x` ] I submit my changes into the `develop` branch

I will update CHANGELOG and add documentation once this has been approved. I expect some changes based on your suggestion. Please feel free to suggest any changes that can make this better.

### License

```python
#
```
requesting for a review @nmanovic @azhavoro

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -27,7 +27,7 @@
             "request": "launch",
             "stopOnEntry": false,
             "justMyCode": false,
-            "pythonPath": "${config:python.interpreterPath}",
+            "pythonPath": "${config:python.pythonPath}",
             "program": "${workspaceRoot}/manage.py",
             "args": [
                 "runserver",
@@ -58,7 +58,7 @@
             "request": "launch",
             "stopOnEntry": false,
             "justMyCode": false,
-            "pythonPath": "${config:python.interpreterPath}",
+            "pythonPath": "${config:python.pythonPath}",
             "program": "${workspaceRoot}/manage.py",
             "args": [
                 "rqworker",
@@ -77,7 +77,7 @@
             "request": "launch",
             "stopOnEntry": false,
             "justMyCode": false,
-            "pythonPath": "${config:python.interpreterPath}",
+            "pythonPath": "${config:python.pythonPath}",
             "program": "${workspaceRoot}/manage.py",
             "args": [
                 "rqscheduler",
@@ -93,7 +93,7 @@
             "request": "launch",
             "justMyCode": false,
             "stopOnEntry": false,
-            "pythonPath": "${config:python.interpreterPath}",
+            "pythonPath": "${config:python.pythonPath}",
             "program": "${workspaceRoot}/manage.py",
             "args": [
                 "rqworker",
@@ -112,7 +112,7 @@
             "request": "launch",
             "justMyCode": false,
             "stopOnEntry": false,
-            "pythonPath": "${config:python.interpreterPath}",
+            "pythonPath": "${config:python.pythonPath}",
             "program": "${workspaceRoot}/manage.py",
             "args": [
                 "update_git_states"
@@ -128,7 +128,7 @@
             "request": "launch",
             "justMyCode": false,
             "stopOnEntry": false,
-            "pythonPath": "${config:python.interpreterPath}",
+            "pythonPath": "${config:python.pythonPath}",
             "program": "${workspaceRoot}/manage.py",
             "args": [
                 "migrate"
@@ -144,7 +144,7 @@
             "request": "launch",
             "justMyCode": false,
             "stopOnEntry": false,
-            "pythonPath": "${config:python.interpreterPath}",
+            "pythonPath": "${config:python.pythonPath}",
             "program": "${workspaceRoot}/manage.py",
             "args": [
                 "test",
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -27,7 +27,7 @@
             "request": "launch",
             "stopOnEntry": false,
             "justMyCode": false,
-            "pythonPath": "${config:python.pythonPath}",
+            "pythonPath": "${config:python.interpreterPath}",
             "program": "${workspaceRoot}/manage.py",
             "args": [
                 "runserver",
@@ -58,7 +58,7 @@
             "request": "launch",
             "stopOnEntry": false,
             "justMyCode": false,
-            "pythonPath": "${config:python.pythonPath}",
+            "pythonPath": "${config:python.interpreterPath}",
             "program": "${workspaceRoot}/manage.py",
             "args": [
                 "rqworker",
@@ -77,7 +77,7 @@
             "request": "launch",
             "stopOnEntry": false,
             "justMyCode": false,
-            "pythonPath": "${config:python.pythonPath}",
+            "pythonPath": "${config:python.interpreterPath}",
             "program": "${workspaceRoot}/manage.py",
             "args": [
                 "rqscheduler",
@@ -93,7 +93,7 @@
             "request": "launch",
             "justMyCode": false,
             "stopOnEntry": false,
-            "pythonPath": "${config:python.pythonPath}",
+            "pythonPath": "${config:python.interpreterPath}",
             "program": "${workspaceRoot}/manage.py",
             "args": [
                 "rqworker",
@@ -112,7 +112,7 @@
             "request": "launch",
             "justMyCode": false,
             "stopOnEntry": false,
-            "pythonPath": "${config:python.pythonPath}",
+            "pythonPath": "${config:python.interpreterPath}",
             "program": "${workspaceRoot}/manage.py",
             "args": [
                 "update_git_states"
@@ -128,7 +128,7 @@
             "request": "launch",
             "justMyCode": false,
             "stopOnEntry": false,
-            "pythonPath": "${config:python.pythonPath}",
+            "pythonPath": "${config:python.interpreterPath}",
             "program": "${workspaceRoot}/manage.py",
             "args": [
                 "migrate"
@@ -144,7 +144,7 @@
             "request": "launch",
             "justMyCode": false,
             "stopOnEntry": false,
-            "pythonPath": "${config:python.pythonPath}",
+            "pythonPath": "${config:python.interpreterPath}",
             "program": "${workspaceRoot}/manage.py",
             "args": [
                 "test",
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,5 @@
 {
-    "python.pythonPath": ".env/bin/python",
+    "python.pythonPath": "/usr/bin/python3",
     "eslint.enable": true,
     "eslint.validate": [
         "javascript",
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,5 @@
 {
-    "python.pythonPath": "/usr/bin/python3",
+    "python.pythonPath": ".env/bin/python",
     "eslint.enable": true,
     "eslint.validate": [
         "javascript",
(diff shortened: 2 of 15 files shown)


nmanovic:
@azhavoro , could you please review, test, and comment?

cvat/apps/tf_annotation/views.py
```diff
@@ -18,329 +19,370 @@
...
-                            xmin, ymin, xmax, ymax = _normalize_box(boxes[0][i], width, height)
-                            label = labels_mapping[classes[0][i]]
-                            if label not in result:
-                                result[label] = []
-                            result[label].append([image_num, xmin, ymin, xmax, ymax])
-        finally:
-            sess.close()
-            del sess
-    return result
+	(im_width, im_height) = image.size
```
azhavoro, line 28: It seems you use tab as indentation, please revert.

cvat/apps/auto_segmentation/views.py
```diff
@@ -211,37 +248,58 @@ def create(request, tid):
...
-            "microwave": 69, "oven": 70, "toaster": 71, "sink": 72, "refrigerator": 73,
-            "book": 74, "clock": 75, "vase": 76, "scissors": 77, "teddy_bear": 78, "hair_drier": 79,
-            "toothbrush": 80
-            }
-
-        labels_mapping = {}
-        for key, labels in db_labels.items():
-            if labels in auto_segmentation_labels.keys():
-                labels_mapping[auto_segmentation_labels[labels]] = key
+        if int(mid) == 989898:
```
azhavoro, line 251: Please use django fixtures or post migrate signals to add proper records to the database. It will allow to not use a magic numbers.

cvat/apps/auto_segmentation/views.py
```diff
@@ -86,33 +112,37 @@ class InferenceConfig(coco.CocoConfig):
 
     ## RUN OBJECT DETECTION
     result = {}
-    frames = frame_provider.get_frames(frame_provider.Quality.ORIGINAL)
-    for image_num, (image_bytes, _) in enumerate(frames):
+    image_num = 0
+    image_list = list(frame_provider.get_frames(frame_provider.Quality.ORIGINAL))
```
azhavoro, line 116: Could there be memory issue here? Why do you need cast to `list` here?

cvat/apps/auto_segmentation/views.py
```diff
@@ -59,19 +77,27 @@ def _convert_to_segmentation(mask):
...
-    class InferenceConfig(coco.CocoConfig):
+    class InferenceConfig(CocoConfig):
         # Set batch size to 1 since we'll be running inference on
         # one image at a time. Batch size = GPU_COUNT * IMAGES_PER_GPU
         GPU_COUNT = 1
         IMAGES_PER_GPU = 1
+        NUM_CLASSES = num_c
+        local_device_protos = device_lib.list_local_devices()
+        num_gpus = len([x.name for x in local_device_protos if x.device_type == 'GPU'])
+        if num_gpus in [2,4,8]:
```
azhavoro, line 99: `if num_gpus % 2 == 0`?

cvat/apps/auto_annotation/model_manager.py
```diff
@@ -144,16 +148,26 @@ def save_file_as_tmp(data):
         dl_model_id = create_empty(owner=owner)
 
     run_tests = bool(model_file or weights_file or labelmap_file or interpretation_file)
+    if is_custom in ["tensorflow","maskrcnn"]:
```
azhavoro, line 151: I see that these string constants are used many times, could you add enum for them?

cvat/apps/auto_annotation/model_manager.py
```diff
@@ -31,7 +31,7 @@ def _remove_old_file(model_file_field):
         os.remove(model_file_field.name)
 
 def _update_dl_model_thread(dl_model_id, name, is_shared, model_file, weights_file, labelmap_file,
-        interpretation_file, run_tests, is_local_storage, delete_if_test_fails, restricted=True):
+        interpretation_file, run_tests, is_local_storage, delete_if_test_fails, is_custom, restricted=True):
```
azhavoro, line 34: `is_custom` -> `framework`?

components/auto_segmentation/install.sh
```diff
@@ -3,11 +3,11 @@
 
 set -e
 
-MASK_RCNN_URL=https://github.com/matterport/Mask_RCNN
+MASK_RCNN_URL=https://github.com/onepanelio/Mask_RCNN
```
azhavoro, line 6: The same as above.

.vscode/settings.json
```diff
@@ -1,5 +1,5 @@
 {
-    "python.pythonPath": ".env/bin/python",
```
azhavoro, line 2: Please revert these changes.

.vscode/launch.json
```diff
@@ -27,7 +27,7 @@
             "request": "launch",
             "stopOnEntry": false,
             "justMyCode": false,
-            "pythonPath": "${config:python.pythonPath}",
```
azhavoro, line 30: Please revert these changes.

components/auto_segmentation/install.sh
san999p: We made several changes in Mask_RCNN repo to make it work with multiple gpus. So, if we change this then it won't work with multiple GPUs.

cvat/apps/auto_annotation/model_manager.py
san999p: ok, makes sense. Will update this.

cvat/apps/auto_annotation/model_manager.py
san999p: Sure, do we have any specific file where we should create enum class? or do you have any suggestion where this should be created?

cvat/apps/auto_segmentation/views.py
san999p: Yeah, it can be changed. We only had 2,4 and 8 GPUs, so.

cvat/apps/auto_segmentation/views.py
san999p: Thanks. I need to check that. I will get back to you. Just to summarize what we do here, we assigned a specific unique number to the default models because we had `int` in the request.

nmanovic:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- cvat/apps/auto_annotation/model_manager.py  1
- cvat/apps/tf_annotation/views.py  6
- cvat/apps/auto_segmentation/views.py  8
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/cvat/cvat/pullRequest?prid=5677871&bid=18586967)

cvat/apps/auto_segmentation/views.py
san999p: We didn't face any issues while testing, but yes I agree with you. We need a length of the frame_provider to figure out distribution among available GPUs. The Generator class does not have length function. Let me see if we can have an alternative way. Do you have any suggestions?

cvat/apps/tf_annotation/views.py
san999p: Ok, will do.

cvat/apps/auto_segmentation/views.py
san999p: We didn't face any issue while testing, but I agree with you. We need a length of a frame_provider to distribute images among all available gpus. The generator class does not have length, so used it. Let me see if we can come up with an alternative. Do you have any suggestions?

.vscode/settings.json
san999p: changed.

cvat/apps/auto_segmentation/views.py
san999p: I made the change.

cvat/apps/auto_annotation/model_manager.py
san999p: Updated files.

cvat/apps/tf_annotation/views.py
san999p: Updated file.

nmanovic:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- cvat/apps/auto_segmentation/views.py  8
- cvat/apps/tf_annotation/views.py  6
- cvat/apps/auto_annotation/model_manager.py  1
         
```


See the complete overview on [Codacy](https://app.codacy.com/manual/cvat/cvat/pullRequest?prid=5677871&bid=18586967)

nmanovic:
@san999p , let's meet next week and discuss the PR. I want to present to you an alternative solution using nuclio serverless platform (https://github.com/nuclio/nuclio): https://github.com/opencv/cvat/tree/nm/serverless.

san999p:
sure @nmanovic let me know what time works for you. Maybe send a calendly link?

cvat/apps/auto_segmentation/views.py
azhavoro: `frame_provider` has `__len__` definition https://github.com/opencv/cvat/blob/develop/cvat/apps/engine/frame_provider.py#L83

cvat/apps/auto_annotation/model_manager.py
azhavoro: I think this file will be ok.

san999p: ok thanks

cvat/apps/auto_segmentation/views.py
san999p: thanks, I will update this.

nmanovic:
@san999p , I will schedule a meeting this week. I will contact Rush and Donald. You can look at the PR to understand future changes: https://github.com/opencv/cvat/pull/1767

kishan-sib:
@san999p @nmanovic 
I think there is better way to support Tensorflow models. There should be generic flow to execute any tensorflow models rather than defining MASK_RCNN as separate framework.

We have simple implementation where user can:
- Create model using Tensorflow frozen .pb file, inference script and label config file
- Auto annotation will work in same flow.

This gives user freedom to use their own models for automation and also they can use any models from TF model zoo.

Let me know your views on this. Will submit PR soon, may be it will give more clear idea.

nmanovic:
@kishan-sib ,

Let's look at your implementation. The main problem that every DL model needs an `inference script` which translates output layer to actual annotations which CVAT can understand. The script can depends on different python modules, it can be implemented using different programming languages, potentially need to support several TF frameworks and custom extensions. Too many problems which need to solve.

It is the primary reason why I'm looking at "serverless approach" to resolve all issues because each DL model will be a separate container. You can look at Mask RCNN description as serverless function for nuclio framework here: https://github.com/opencv/cvat/tree/nm/serverless/serverless/tensorflow/matterport/mask_rcnn/nuclio

savan77:
I think #1767 closes this PR.

nmanovic:
@savan77 , thanks for the input. I will close the PR.

PR #1764, opened 2020-06-19
Allow import of single boxes in MOT format

zhiltsov-max:
Implements the use case found in #1463 

### Motivation and context

Extends MOT format support with import of single boxes (without tracks). This is an extension for the original format, as it does not expect non-track annotations.

### How has this been tested?

Unit test

### Checklist

### License

```python
#
```


cvat/apps/dataset_manager/formats/mot.py
```diff
@@ -45,6 +45,17 @@ def _import(src_file, task_data):
 
                 track_id = ann.attributes.get('track_id')
                 if track_id is None:
+                    # Extension. Import regular boxes:
+                    task_data.add_shape(task_data.LabeledShape(
```
nmanovic, line 49: Why don't create a track with only one bounding box?

zhiltsov-max: Is it any better? This would not preserve the original annotations, as all tracks are exported as tracks, even single-framed. We can, possibly, convert regular boxes to such tracks.

nmanovic: MOT is for tracks only. Right? How are you going to dump a track which has two shapes:
- First with outside=False, frame=0
- Second with outside=True, frame=1?

zhiltsov-max: MOT is mostly for tracks, look here: https://motchallenge.net/instructions/. Almost every version has `<[track] id>` field set (idx 2), but not `MOT17det`, which contains just boxes and `-1` as `id`. So we can safely assume that bboxes are supported via this extension.

nmanovic: OK

PR #1807, opened 2020-06-26
Match image by name (instead of id) on CVAT upload

artemisart:
### Motivation and context

Currently, the cvat annotation import (upload) matches annotations to frames by id, so this works only on very specific cases like backups for the same task. When moving annotations between tasks, if the order doesn't match exactly or if there are more frames, or (in our case) if we generate cvat annotations for pre-annotation, nothing matched and it's useless.
This PR just uses the same matching as for coco files, so it works both for old use-cases like backup (if the ID matched, then the filenames will match) but also pre-annotation and others.

### How has this been tested?

Manually, uploaded some files.

### Checklist

### License

```python
#
```

### DIFF
--- a/cvat/apps/dataset_manager/formats/cvat.py
+++ b/cvat/apps/dataset_manager/formats/cvat.py
@@ -432,7 +432,7 @@ def load(file_object, annotations):
                 )
             elif el.tag == 'image':
                 image_is_opened = True
-                frame_id = int(el.attrib['id'])
+                frame_id = annotations.match_frame(el.attrib['name'])
             elif el.tag in supported_shapes and (track is not None or image_is_opened):
                 attributes = []
                 shape = {
--- a/cvat/apps/dataset_manager/formats/cvat.py
+++ b/cvat/apps/dataset_manager/formats/cvat.py
@@ -9,8 +9,10 @@ from collections import OrderedDict
 from glob import glob
 from tempfile import TemporaryDirectory
 
+from cvat.apps.dataset_manager.bindings import match_dm_item
 from cvat.apps.dataset_manager.util import make_zip_archive
 from cvat.apps.engine.frame_provider import FrameProvider
+from datumaro.components.extractor import DatasetItem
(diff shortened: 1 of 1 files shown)


cvat/apps/dataset_manager/formats/cvat.py
```diff
@@ -432,7 +432,7 @@ def load(file_object, annotations):
                 )
             elif el.tag == 'image':
                 image_is_opened = True
-                frame_id = int(el.attrib['id'])
+                frame_id = annotations.match_frame(el.attrib['name'])
```
zhiltsov-max, line 435: I think you should use https://github.com/opencv/cvat/blob/develop/cvat/apps/dataset_manager/bindings.py#L571 here, which will fall back to ids and other things in the case image name did not match. This can happen, for example, in the case when  the task was created from files on a shared storage, as image names in the task will include their absolute paths, in which case the new variant doesn't work. This problem would require analysis of the whole input file, but ids can simplify things.

Another simple variant - use the name, otherwise use the id and check if basenames match, if there is one provided.

artemisart: Indeed, I thought match_frame already matched on the basename and not the full path but not sure, and I didn't follow the updates of the importer.

zhiltsov-max: You can find an example of a `DataseItem` for these purposes here: https://github.com/opencv/cvat/blob/develop/cvat/apps/dataset_manager/formats/yolo.py#L40

I think, an approach in the last line above will just fit.

azhavoro: Hi, any updates?

artemisart:
Yes, sorry I didn't have the time to update and test yet but it should be good now.

cvat/apps/dataset_manager/formats/cvat.py
artemisart: Done

PR #1808, opened 2020-06-26
Avoid hard-coding container names and use docker-compose…

artemisart:
… instead of docker

### Motivation and context

We can't run multiple cvat on the same host because the container names are hardcoded (and clashes with the already running containers). Omitting `container_names` allows to use the default ones which are the name of the service, prefixed by the project name ([--project-name](https://docs.docker.com/compose/reference/overview/) which is by default the directory name) so eg. `cvat` becomes `cvat_cvat_1` if we cloned the repo to `cvat/` (the `_1` is because it's suffixed to allow multiple containers of the same image).
So now we just have to clone to another directory, change the 8080 port, and it works because everything is prefixed. Small change: it's better to use `docker-compose something` because it will find the containers of the current project, instead of having to prefix the names manually (ie. use `docker-compose exec cvat bash` instead of `docker exec -it cvat_cvat_1 bash`). I updated the documentation.

### How has this been tested?

Currently running 3 cvats on the same host without issues.

### Checklist

### License

```python
#
```

### DIFF
--- a/components/analytics/docker-compose.analytics.yml
+++ b/components/analytics/docker-compose.analytics.yml
@@ -1,7 +1,6 @@
 version: '2.3'
 services:
   cvat_elasticsearch:
-    container_name: cvat_elasticsearch
     image: cvat_elasticsearch
     networks:
       default:
@@ -16,7 +15,6 @@ services:
     restart: always
 
   cvat_kibana:
-    container_name: cvat_kibana
     image: cvat_kibana
     networks:
       default:
@@ -30,7 +28,6 @@ services:
     restart: always
 
   cvat_kibana_setup:
-    container_name: cvat_kibana_setup
     image: cvat
     volumes: ['./components/analytics/kibana:/home/django/kibana:ro']
     depends_on: ['cvat']
@@ -42,7 +39,6 @@ services:
       no_proxy: elasticsearch,kibana,${no_proxy}
 
   cvat_logstash:
-    container_name: cvat_logstash
     image: cvat_logstash
     networks:
       default:
--- a/cvat/apps/documentation/faq.md
+++ b/cvat/apps/documentation/faq.md
@@ -20,13 +20,13 @@ docker-compose up -d
 ```
 
 Sometimes the update process takes a lot of time due to changes in the database schema and data.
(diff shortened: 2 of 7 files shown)


azhavoro:
@artemisart  Hi, sorry for the late reply and thanks for the PR, in general I agree with you that it is better not to use hard-coded names for containers, and this problem is especially relevant for multiple installation of CVAT. But I think that in most cases our users have only one installation, and it is better to have a predictable container name regardless of the compose project name and some suffix. You also missed one important detail in the documentation: any docker-compose command should run in the cvat directory. For example, I don’t need to know what name compose was generated for the `cvat` service or directory where the cvat code is located to run the command `docker-compose logs ...`, so I just ask the user to run `docker logs cvat`.
Anyway for multiple installations , it is possible to override the container name in the 'docker-compose.override.yml' file.

nmanovic:
@artemisart ,

Thanks for the PR. Really great to see contributions from the community. Don't hesitate to send us new ones but discuss them before to avoid extra work on your side.

> Anyway for multiple installations , it is possible to override the container name in the 'docker-compose.override.yml' file.

It looks like we have a clear way to workaround the issue. You need to update docker-compose.override.yml in any case to resolve conflicts between ports. The same can be done for container names.

I will close the PR.

artemisart:
@azhavoro @nmanovic I understand that not hard coding container names introduces a little bit of complexity, but it could just be detailed in the documentation to use docker-compose in the right folder? I think the alternative of overriding every container name, including the extensions, is harder.

nmanovic:
@artemisart , let's start with a documentation how to solve the particular problem using docker-compose.override.yml. To run CVAT on the same machine you have to update ports in any case. Thus it makes sense to ask to update container names as well. Meaningful names are better than the random ones.

dschoerk:
Just arrived here because I also ran into the issue of not being able to run multiple CVAT instances on the same machine. My use-case is that I have a CVAT v2.1.0 installation and was trying to clone it before upgrading. I'll give it a try with docker-compose.override.yml at the next attempt :) From my experience it is rather unusual to hardcode the container names.

PR #1820, opened 2020-06-29
[CVAT-UI]: Appearance settings in AAM, keyframe navigation and other buttons in AAM

### LINKED ISSUE #1422: CVAT new UI: select opacity in the AAM

### Expected Behaviour

Added the opacity regulation to the Attribute Annotation mode
### Current Behaviour

No opportunity to regulate opacity in the Attribute Annotation mode
### Possible Solution

Add the corresponding button

### Context

To recognize some details or small attributes when we work in the AAM we use opacity. To regulate it we need to switch to the Standard mode and vice versa. We'd like to have an opportunity to change opacity regardless of the chosen annotation mode.

### LINKED ISSUE #1423: CVAT new UI: add the "lock" and the "move to key frame" functions to Attribute annotation mode

### Expected Behaviour

Expected that we have an opportunity to use some functions when we work in the Attribute Annotation mode:
1. Objects locking
2. Moving to key frames
Previously all of functions were available:
image

### Current Behaviour

At the moment there are no available actions for an object in this mode:
image
 
### Possible Solution

Add the functions "locking" and "moving to key frames" (incl. hot keys) to the AAM

### Context

Mostly uses the "locking" and "moving to key frames". "Locking" is useful to avoid any changes we don't want to do.
"Moving to key frames" uses very often when we work in the AAM, look after some object and change its attributes.

bsekachev:
### Motivation and context
Implementation of some features that have been requested by internal team

Changes:
- Component/container refactoring to reuse existing components in different places. Some files have been spitted because a lot of components in the same file. 
- Appearance block in AAM
- Block with some switchers in AAM

Resolve #1423
Resolve #1422

### How has this been tested?
Manual testing

### Checklist

### License

```python
#
```

Screenshot from 2020-06-29 22-49-42


ActiveChooN:
@bsekachev, get some issue when changed between frames/objects.

aam_tags_canvas_failure_1

bsekachev:
@ActiveChooN 

Try to hide a rect or do it outside, then open AAM and go to this shape and you get the same result in develop,
But in this patch I tried to fix it. It seems not completely. I need specific info about shape where you get such a result

ActiveChooN:
@bsekachev, if I enter the wrong filter then the filter input field disappears. So i need to change standard mode to fix filter.

ActiveChooN:
@bsekachev, also i have an idea. It is not directly related with this PR. I guess it will be more convenient to start attributes shortcuts ordering from 1 not from 0, like it placed on keyboard. So label attribute shortcuts will be 1, 2 ... 9, 0 and not 0, 1, 2 ... 9.

bsekachev:
> @bsekachev, also i have an idea. It is not directly related with this PR. I guess it will be more convenient to start attributes shortcuts ordering from 1 not from 0, like it placed on keyboard. So label attribute shortcuts will be 1, 2 ... 9, 0 and not 0, 1, 2 ... 9.

We use 0,1,2,3 because of historycal reasons. Our annotators use numpad keyboard. 
We can discuss the idea on the meeting, but I think there won't be positive feedback from annotators

bsekachev:
> @bsekachev, if I enter the wrong filter or filter with empty result then the filter input field disappears. So i need to change standard mode to fix filter.

Again, the same in develop. But I fixed it here.

PR #1969, opened 2020-07-30
Increase the rate of allowed requests

mpeven:
Increase the amount of requests that an "unauthenticated" user can perform, to prevent throttling errors when using the CLI.

### Motivation and context

Hitting the /login endpoint from an unauthenticated user can trigger throttling. While throttling should be kept, as it is good practice, the rate can be increased as the user may have valid authentication. A reasonable level of 100 requests/minute should be enough to enforce good practices while preventing headaches for users trying to automate tasks with the CLI. 

(This rate was chosen to be similar to the [rate limiting docs](https://laravel.com/docs/7.x/routing#rate-limiting) for Laravel)

### How has this been tested?

Manually.

### Checklist

~~- [ ] I have increased versions of npm packages if it is necessary ([cvat-canvas](https://github.com/opencv/cvat/tree/develop/cvat-canvas#versioning),
[cvat-core](https://github.com/opencv/cvat/tree/develop/cvat-core#versioning), [cvat-data](https://github.com/opencv/cvat/tree/develop/cvat-data#versioning) and [cvat-ui](https://github.com/opencv/cvat/tree/develop/cvat-ui#versioning))~~

### License

```python
#
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Stylelint filetype scans (<https://github.com/opencv/cvat/pull/1952>)
 - Fixed toolip closing issue (<https://github.com/opencv/cvat/pull/1955>)
 - Clearing frame cache when close a task (<https://github.com/opencv/cvat/pull/1966>)
+- Increase rate of throttling policy for unauthenticated users (<https://github.com/opencv/cvat/pull/1969>)
 
 ### Security
 -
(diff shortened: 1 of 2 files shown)


mpeven:
@nmanovic if it helps in the review - I discussed this on Gitter with @azhavoro and he agrees this is a correct change to make.

nmanovic:
@mpeven , thanks for the PR. 👍 Let me think a little bit about the issue. I understand the problem. Your new limit looks reasonable and I don't think that it is a security issue (100 per minute is pretty small). I will come back soon.

Issue #1965, opened 2020-07-30
When developing without Docker it always need to run nuclio container


bsekachev:
### My actions before raising this issue
- [x] Read/searched [the docs](https://github.com/opencv/cvat/tree/master#documentation)
- [x] Searched [past issues](/issues)

### Expected Behaviour
If nuclio dashboard is off, basic functionality of CVAT is expected to work. 

### Current Behaviour
When loading UI, server raises error: 
``Error: Request failed with status code 503. "HTTPConnectionPool(host='localhost', port=8070): Max retries exceeded with url: /api/functions (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f0a296f9b00>: Failed to establish a new connection: [Errno 111] Connection refused',))".``

### Possible Solution
If nuclio server is unavaliable, server can return empty list of functions

### Your Environment

- Git hash commit (`git log -1`):  0062ecdec
- Docker version `docker version` (e.g. Docker 17.0.05): no
- Are you using Docker Swarm or Kubernetes? no
- Operating System and version (e.g. Linux, Windows, MacOS): Ubuntu 18.04

nmanovic:
We decided to close the bug for a while. The behavior is expected.

PR #1974, opened 2020-07-31
stop dextr plugin support check from crashing app

TravisYeah:
The plugin checker checks for dextr support by requesting the lambda
list but it returns a 503 error when the lambda service is not running
which crashes the app. This change catches dextr support check errors
and returns false.

Fixes: #1965

### Checklist

### License

```python
#
```

### DIFF
--- a/cvat-ui/src/actions/plugins-actions.ts
+++ b/cvat-ui/src/actions/plugins-actions.ts
@@ -39,17 +39,34 @@ export function checkPluginsAsync(): ThunkAction {
             DEXTR_SEGMENTATION: false,
         };
 
-        const promises: Promise<boolean>[] = [
-            PluginChecker.check(SupportedPlugins.ANALYTICS),
-            PluginChecker.check(SupportedPlugins.GIT_INTEGRATION),
-            PluginChecker.check(SupportedPlugins.DEXTR_SEGMENTATION),
-        ];
-
         try {
+            let pluginCheckError: any = null;
+
+            const promises: Promise<boolean>[] = [
+                PluginChecker.check(SupportedPlugins.ANALYTICS).catch((err) => {
+                    pluginCheckError = err;
+                    return false;
+                }),
+                PluginChecker.check(SupportedPlugins.GIT_INTEGRATION).catch(
+                    (err) => {
+                        pluginCheckError = err;
+                        return false;
+                    },
+                ),
+                PluginChecker.check(SupportedPlugins.DEXTR_SEGMENTATION).catch(
+                    (err) => {
+                        pluginCheckError = err;
+                        return false;
+                    },
+                ),
+            ];
+
             const values = await Promise.all(promises);
             [plugins.ANALYTICS, plugins.GIT_INTEGRATION,
                 plugins.DEXTR_SEGMENTATION] = values;
             dispatch(pluginActions.checkedAllPlugins(plugins));
+
+            if (pluginCheckError instanceof Error) throw pluginCheckError;
         } catch (error) {
             dispatch(pluginActions.raisePluginCheckError(error));
         }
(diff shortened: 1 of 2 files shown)


bsekachev:
@TravisYeah 

Hi,
We have discussed this issue internally and decided to left all as is because it is expected, that nuclio is always runned with CVAT. Is not it true for you? How the issue does affect you?
I am not sure that just silence is the best solution, because it can confuse users who expect DEXTR to work, but forgot to run nuclio dashboard (at least in current implementation they get a notification).

Probably need to discuss with @nmanovic

nmanovic:
@ActiveChooN , could you please look at the PR? It looks good from my point of view.

cvat-ui/src/actions/plugins-actions.ts
```diff
@@ -39,17 +39,34 @@ export function checkPluginsAsync(): ThunkAction {
...
+                PluginChecker.check(SupportedPlugins.ANALYTICS).catch((err) => {
+                    pluginCheckError = err;
+                    return false;
+                }),
+                PluginChecker.check(SupportedPlugins.GIT_INTEGRATION).catch(
+                    (err) => {
+                        pluginCheckError = err;
+                        return false;
+                    },
+                ),
```
ActiveChooN, line 55: This cases are wrapped with `try catch` in plugin-checker so i don't think catches blocks are necessary here. Also you can dispatch `raisePluginCheckError` just here without having to re-throw an error.

ActiveChooN:
@nmanovic, @bsekachev, should end users see the notification every time they reload the page? Maybe there are some installation without any functions and nuclio itself. There is an axios error with 503 code in console for administrators and developers already. If yes, message `Could not initialize plugins state` is irrelevant here due to we have initialized it already. Maybe we should have another checker for nuclio with relevant message and dextr plugin checker should return true or false without throwing any error, like other checkers do. What is your opinion?

cvat-ui/src/actions/plugins-actions.ts
TravisYeah: The catch statement does not dispatch `checkedAllPlugins` and even if it does it will not know what plugins are actually checked because the `Promise.all` will throw immediately after any promise returns a rejection error. This means that the screen just sits in the `fetching` state forever. My change allows each plugin to be checked before an alert is raised then it dispatches `checkedAllPlugins` with the finished plugin check list so that the page will load (set `fetching` to `false` and `initialized` to `true`) but still alert the user when one or more plugins fail.

I made the change because I am having issues getting nuclio to work but I haven't created an issue yet because I need to put some more time into it (and I think it is a nuclio issue). Seems to be my private ca certificate blocking nuclio from running properly.

nmanovic: @TravisYeah, could you please report your problem with Nuclio? Even you cannot debug it on your side yet it is really important to identify all of them as early as possible.

bsekachev: Instead of this construction I would suggest to use ``Promise.allSettled()`` instead of ``Promise.all()`` below that returns results of promises.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled

bsekachev:
@ActiveChooN 

> should end users see the notification every time they reload the page?

I suppose yes, because CVAT is a single page application. From this point of view, it is the same like you run desktop app and every time when there are errors, a user should know that something is going wrong.

> If yes, message Could not initialize plugins state is irrelevant here due to we have initialized it already.

I agree that probably we can find a more relevant message (like ``Could not fetch plugins information from the server`` or smth like this), but need to remember that it is only title. Message body is dynamic and depends on raised exceptions (but in this case it outputs only in console because of size, maybe it is wrong)

bsekachev:
@TravisYeah 

Hi, really appreciate your work. We discussed this issue internally and decided to change logic a bit. 
So, we prepared another PR to fix the issue #2008
Anyway, thx for your help and time

PR #1983, opened 2020-08-04
Preparation for testing. Add tests for issue 1599.

dvkruchinin:
### Motivation and context
Re-configure .travis.yml to run Cypress from Docker host.
Add auto creation superuser for login to CVAT and testing.
Remove unnecessary line from Dockerfile.ci 
Add tests for issue 1599

### Checklist

### License

```python
#
```


azhavoro:
@dvkruchinin What is the licence of the `cars.jpg` file?

dvkruchinin:
> @dvkruchinin What is the licence of the `cars.jpg` file?

This image is freely available on the Internet. I didn't find any mention of a license.

nmanovic:
> > @dvkruchinin What is the licence of the `cars.jpg` file?
> 
> This image is freely available on the Internet. I didn't find any mention of a license.

@dvkruchinin , it is not acceptable. You can use any image which you can find in our repo.

dvkruchinin:
> > > @dvkruchinin What is the licence of the `cars.jpg` file?
> > 
> > 
> > This image is freely available on the Internet. I didn't find any mention of a license.
> 
> @dvkruchinin , it is not acceptable. You can use any image which you can find in our repo.

Done. Replaced the image.

nmanovic:
@dvkruchinin , can we generate an image on the fly? Do we need the actual content of the image?

dvkruchinin:
> @dvkruchinin , can we generate an image on the fly? Do we need the actual content of the image?
can we generate an image on the fly? - it is for investigation. It is not clear what is meant by image generation. Alternatively, I can add several images to the directory and use them if necessary.
Do we need the actual content of the image? - in this moment no. Current tasks do not require the actual content.

nmanovic:
@dvkruchinin , in this case I will ask to generate the content. It will allow us to create tasks with any number of images.

dvkruchinin:
> @dvkruchinin , in this case I will ask to generate the content. It will allow us to create tasks with any number of images.

This will take longer than planned. I suggest that solve this in a separate task and create a separate PR.

dvkruchinin:
I think it would be better to remove the annotation task creation functionality from this PR. I have an idea to rework this functionality in the form of a Cypress command that can be called in tests for which need to create an annotation task.

nmanovic:
> > @dvkruchinin , in this case I will ask to generate the content. It will allow us to create tasks with any number of images.
> 
> This will take longer than planned. I suggest that solve this in a separate task and create a separate PR.

It is fine. Let's re-estimate the task.

nmanovic:
> I think it would be better to remove the annotation task creation functionality from this PR. I have an idea to rework this functionality in the form of a Cypress command that can be called in tests for which need to create an annotation task.

I believe it is a good idea. Let's discuss a plan for the PR offline.

tests/cypress/integration/auth_page.js
```diff
@@ -7,27 +7,27 @@
...
-    cy.get('[type="submit"]')
-  })
+    it('"Sign in" button is exists', () => {
+        cy.get('[type="submit"]')
+    })
 
-  it('Check plaseholder "Username"', () => {
-    cy.get('input').invoke('attr', 'placeholder').should('contain', 'Username')
-  })
+    it('Check plaseholder "Username"', () => {
```
bsekachev, line 22: placeholder*

tests/cypress/integration/auth_page.js
```diff
@@ -7,27 +7,27 @@
...
+        cy.get('input').invoke('attr', 'placeholder').should('contain', 'Username')
+    })
 
-  it('Check plaseholder "Password"', () => {
-    cy.get('[type="password"]')
-  })
+    it('Check plaseholder "Password"', () => {
```
bsekachev, line 26: placeholder*

tests/cypress/integration/issue_1599_ch_user_registration.js
```diff
@@ -0,0 +1,50 @@
...
+context('Issue 1599 (Chinese alphabet).', () => {
+    before(() => {
+        cy.visit('auth/register')
+        cy.url().should('include', '/auth/register')
+    })
+
+    describe('User registration using the Chinese alphabet.', () => {
+        it('Filling in the plaseholder "First name"', () => {
+          cy.get('[placeholder="First name"]').type('测试者')
+          cy.contains('Please specify a first name').should('not.exist')
```
bsekachev, line 18: I would suggest to check this test in following way: 

``cy.get('[placeholder="First name"]').should('not.have.class', 'has-error')``

Because we can just change these messages and it would be better if tests won't be broken in this case.

tests/cypress/integration/issue_1599_ch_user_registration.js
```diff
@@ -0,0 +1,50 @@
...
+        })
+
+        it('Filling in the plaseholder "Last name"', () => {
+              cy.get('[placeholder="Last name"]').type('测试')
+              cy.contains('Please specify a last name').should('not.exist')
```
bsekachev, line 23: The same like above

tests/cypress/integration/issue_1599_ch_user_registration.js
```diff
@@ -0,0 +1,50 @@
...
+        it('Filling in the plaseholder "Confirm password"', () => {
+              cy.get('[placeholder="Confirm password"]').type('Qwerty123!')
+        })
+
+        it('Click to "Submit" button', () => {
+              cy.get('[type="submit"]').click()
+        })
+
+        it('Successful registration', () => {
+              cy.url().should('include', 'task')
```
bsekachev, line 47: Maybe ``should('include', '/tasks')``?

tests/cypress/integration/issue_1599_pl_user_registration.js
```diff
@@ -0,0 +1,50 @@
+/*
```
bsekachev, line 1: The same comments like in the previous file

tests/cypress/plugins/imageGenerator/addPlugin.js
```diff
@@ -0,0 +1,43 @@
...
+"use strict";
+
+Object.defineProperty(exports, "__esModule", {
+    value: true
+});
+exports.imageGenerator = imageGenerator;
+
+var jimp = require('jimp');
+var path = require('path');
+var fs = require('fs-extra');
```
bsekachev, line 16: Let's use ``const`` and ``let`` instead of ``var`` everywhere. ``const`` is used to names that cannot be reassigned in the future, else ``let`` is used. There are some benefits of using these keywords in JS.

tests/cypress/plugins/imageGenerator/index.d.ts
```diff
@@ -0,0 +1,17 @@
...
+
+declare namespace Cypress {
+    interface Chainable<Subject> {
+        /**
+         * @example
+         * cy.imageGenerator('./fixtures', 'image.png', 500, 500, 'gray', 'Hello world!')
+         */
+        downloadFile(directory: string, fileName: string, width: number, height: number, color: string, message: string): Chainable<any>
+    }
+}
```
bsekachev, line 17: Why we need TypeScript declaration file if all tests are written on JavaScript?

tests/cypress/plugins/index.js
```diff
@@ -0,0 +1,16 @@
...
+ */
+
+/// <reference types="cypress" />
+/**
+ * @type {Cypress.PluginConfig}
+ */
+
+const {imageGenerator} = require('../plugins/imageGenerator/addPlugin')
+
+module.exports = (on, config) => {
```
bsekachev, line 14: ``config`` is unused argument

tests/cypress/plugins/imageGenerator/addPlugin.js
```diff
@@ -0,0 +1,43 @@
...
+ * Copyright (C) 2020 Intel Corporation
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+"use strict";
+
+Object.defineProperty(exports, "__esModule", {
+    value: true
+});
```
bsekachev, line 11: Hm, I've read about this and as I understand it is need in order to CommonJS modules were able to correctly import ES6 modules. Could you please explain me why we need it here?

As I see both this module and module where you import it, are CommonJS

tests/cypress/plugins/imageGenerator/addPlugin.js
```diff
@@ -0,0 +1,43 @@
...
+function imageGenerator(args) {
+    var directory = args.directory;
+    var fileName = args.fileName;
+    var width = args.width;
+    var height = args.height;
+    var color = args.color;
+    var message = args.message;
+    var image = new jimp(width, height, color, function (err, image) {
+        if (err) throw err;
+    });
```
bsekachev, line 27: I suspect this function to be async. It is why it has callback as the latest argument. 
From this point of view it would be more reliable to do all actions (printing text, and saving it) with the image inside the callback.

tests/cypress/plugins/imageGenerator/addPlugin.js
```diff
@@ -0,0 +1,43 @@
...
+    var posX = 10;
+    var posY = 10;
+    jimp.loadFont(jimp.FONT_SANS_64_BLACK).then(function (font) {
+        image.print(font, posX, posY, message);
+        return image;
+    }).then(function (image) {
+        var file = path.join(directory, fileName);
+        return image.write(file);
+    });
```
bsekachev, line 36: This code looks like async. It means that in common case image doesn't exist when return statement works.

tests/cypress/integration/auth_page.js
dvkruchinin: Fixed. Thanks.

tests/cypress/integration/auth_page.js
dvkruchinin: Fixed. Thanks.

tests/cypress/integration/issue_1599_ch_user_registration.js
dvkruchinin: Done. Replaced.

tests/cypress/integration/issue_1599_ch_user_registration.js
dvkruchinin: Replaced.

tests/cypress/integration/issue_1599_ch_user_registration.js
dvkruchinin: Yes. Thanks. It's safer that way. Replaced.

tests/cypress/integration/issue_1599_pl_user_registration.js
dvkruchinin: Done.

tests/cypress/plugins/imageGenerator/addPlugin.js
dvkruchinin: Thanks. Replaced.

tests/cypress/plugins/imageGenerator/index.d.ts
dvkruchinin: Yes, this file is not needed. Removed.

tests/cypress/plugins/index.js
dvkruchinin: Until deleted this argument.

tests/cypress/plugins/imageGenerator/addPlugin.js
dvkruchinin: Initially, I compiled the file using the babel.  These lines are not needed. Removed.

tests/cypress/plugins/imageGenerator/addPlugin.js
dvkruchinin: Rewritten the function. Thanks.

tests/cypress/integration/issue_1599_ch_user_registration.js
```diff
@@ -0,0 +1,50 @@
...
+        it('Filling in the placeholder "Confirm password"', () => {
+            cy.get('[placeholder="Confirm password"]').type('Qwerty123!')
+        })
+
+        it('Click to "Submit" button', () => {
+            cy.get('[type="submit"]').click()
+        })
+
+        it('Successful registration', () => {
+            cy.url().should('include', '/task')
```
bsekachev, line 47: Still ``/task`` instead of ``/tasks``

tests/cypress/integration/issue_1599_pl_user_registration.js
```diff
@@ -0,0 +1,50 @@
...
+        it('Filling in the placeholder "Confirm password"', () => {
+            cy.get('[placeholder="Confirm password"]').type('Qwerty123!')
+        })
+
+        it('Click to "Submit" button', () => {
+            cy.get('[type="submit"]').click()
+        })
+
+        it('Successful registration', () => {
+            cy.url().should('include', '/task')
```
bsekachev, line 47: Still ``/task`` instead of ``/tasks``

tests/cypress/integration/issue_1599_ch_user_registration.js
dvkruchinin: Fixed.

tests/cypress/integration/issue_1599_pl_user_registration.js
dvkruchinin: Fixed.

PR #1988, opened 2020-08-05
Siammask tracker as DL serverless function

nmanovic:
### Motivation and context

Add support of tracker serverless functions into Lambda Manager. Also add [SiamMask](https://github.com/foolwood/SiamMask) tracker as a DL serverless function.
The whole idea is trivial. To transform a tracker to DL serverless function it should return its `state` on each call. In the beginning the state is None and we send `shape` instead to initialize the tracker. After that it is necessary to send `state` for each new call of the function to continue the tracking procedure.

**An idea for the future**: now we use requests only to submit data to DB. It will be great if we have two types of requests: online and offline. Online requests will send results directly to UI at the end. Offline requests will send results to DB as now.

**Note**: it seems it will not be easy to convert other trackers to DL serverless functions because they hide own state usually. Every time need to implement encode_state and decode_state functions to serialize/deserialize a state to/from JSON. It should not be a problem for python native trackers because python cannot truly incapsulate anything but it is necessary to understand what is a state for a tracker.

**Note**: it was pain to integrate conda to docker. It is a good example that implementation of a DL serverless function can be non-trivial.

### How has this been tested?

Manually. I used model_handler.py from the PR with demo.py from SiamMask GitHub repo and saw that I got the same tracking results.

### Checklist

### License

```python
#
```


nmanovic:
@azhavoro , could you please look at the PR? I see that we have a problem with build. Could you please investigate and fix?

PR #2013, opened 2020-08-11
Cypress test for issue 1919

dvkruchinin:
### Motivation and context
Add Cypress test for issue [1919](https://github.com/opencv/cvat/issues/1919)

### Checklist

### License

```python
#
```

### DIFF
--- a/tests/cypress/integration/issue_1919_check_text_attr.js
+++ b/tests/cypress/integration/issue_1919_check_text_attr.js
@@ -1,0 +1,64 @@
+/*
+ * Copyright (C) 2020 Intel Corporation
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+/// <reference types="cypress" />
+
+context('Check label attribute changes', () => {
+
+    const issueId='1919'
+    const labelName=`Issue ${issueId}`
+    const taskName=`New annotation task for ${labelName}`
+    const attrName=`Attr for ${labelName}`
+    const textDefaultValue='Some default value for type Text'
+    const image=`image_${issueId}.png`
+    const newLabelAttrValue = 'New attribute value'
+    const width='800'
+    const height='800'
+    const posX=10
+    const posY=10
+    const color='gray'
+
+    before(() => {
+        cy.visit('auth/login')
+        cy.login()
+        cy.get('[type="submit"]').click()
+        cy.imageGenerator('cypress/fixtures', image, width, height, color, posX, posY, labelName)
+        cy.createAnnotationTask(taskName, labelName, attrName, textDefaultValue, image)
+    })
+
+    describe(`Open task "${taskName}"`, () => {
+        it('The annotation task present in the list', () => {
+            cy.contains('strong', taskName)
+            .parent()
+            .parent()
+            .parent()
+            .contains('a', 'Open').click()
+        })
+        it('Open a job', () => {
+            cy.contains('a', 'Job #').click()
+            cy.url().should('include', '/jobs')
+        })
+        it('Create a shape', () => {
+            cy.createShape(309, 431, 616, 671)
+        })
+        it('Open object menu on the created shape', () => {
+            cy.get('#cvat_canvas_shape_1').trigger('mousemove').rightclick()
+        })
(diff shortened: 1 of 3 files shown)


tests/cypress/integration/issue_1919_check_text_attr.js
```diff
@@ -0,0 +1,64 @@
...
+    const width='800'
+    const height='800'
+    const posX=10
+    const posY=10
+    const color='gray'
+
+    before(() => {
+        cy.visit('auth/login')
+        cy.login()
+        cy.get('[type="submit"]').click()
```
bsekachev, line 27: Shouldn't ``cy.get('[type="submit"]').click()`` be a part of ``cy.login()``?

tests/cypress/integration/issue_1919_check_text_attr.js
```diff
@@ -0,0 +1,64 @@
...
+
+    const issueId='1919'
+    const labelName=`Issue ${issueId}`
+    const taskName=`New annotation task for ${labelName}`
+    const attrName=`Attr for ${labelName}`
+    const textDefaultValue='Some default value for type Text'
+    const image=`image_${issueId}.png`
+    const newLabelAttrValue = 'New attribute value'
+    const width='800'
+    const height='800'
```
bsekachev, line 19: Why we use strings instead of numbers here?

tests/cypress/integration/issue_1919_check_text_attr.js
```diff
@@ -0,0 +1,64 @@
...
+            cy.url().should('include', '/jobs')
+        })
+        it('Create a shape', () => {
+            cy.createShape(309, 431, 616, 671)
+        })
+        it('Open object menu on the created shape', () => {
+            cy.get('#cvat_canvas_shape_1').trigger('mousemove').rightclick()
+        })
+        it('Open object menu details', () => {
+            cy.get('.cvat-canvas-context-menu > [style="display: flex; margin-bottom: 1px;"] > #cvat-objects-sidebar-state-item-1 > .ant-row > .ant-collapse > .ant-collapse-item > .ant-collapse-header > span')
```
bsekachev, line 51: Can we escape such long and complex selectors? They are extremely unreliable. Any minor UI change here and these tests will be broken. If you need more specific css selectors, I believe we can add them. In this case we might use for example something like this: ``#cvat-objects-sidebar-state-item-1 .ant-collapse-header span``

The same comment to selectors below

tests/cypress/integration/issue_1919_check_text_attr.js
```diff
@@ -0,0 +1,64 @@
...
+        cy.createAnnotationTask(taskName, labelName, attrName, textDefaultValue, image)
+    })
+
+    describe(`Open task "${taskName}"`, () => {
+        it('The annotation task present in the list', () => {
+            cy.contains('strong', taskName)
+            .parent()
+            .parent()
+            .parent()
+            .contains('a', 'Open').click()
```
bsekachev, line 38: Can we get this button in a simpler way? ``cy.get('[href="/tasks/${taskID}"]').click()``

tests/cypress/integration/issue_1919_check_text_attr.js
dvkruchinin: Yes it's better that way. Moved.

tests/cypress/integration/issue_1919_check_text_attr.js
dvkruchinin: Jimp successfully handles this format as well. But it is better to use a numeric format. Thanks.

tests/cypress/integration/issue_1919_check_text_attr.js
dvkruchinin: I tried to shorten the selector.

tests/cypress/integration/issue_1919_check_text_attr.js
dvkruchinin: `${taskID}` unknown in advance. I tried using a slightly different solution.

PR #2014, opened 2020-08-11
Label color

### LINKED ISSUE #1480: Feature: custom label colors
Is this possible?
Like red for person,blue for car,so whenever I annotate a car it should've blue color and whenever I annotate person it should've red color.

### LINKED ISSUE #1572: Allow users to modify colors of labels
- [X] Read/searched [the docs](https://github.com/opencv/cvat/tree/master#documentation)
- [X] Searched [past issues](/issues)

### Expected Behaviour
Enable users to add more colors based on hex value. This gives the control to the users and allows them to represent classes in any fashion they need.

### Current Behaviour
Currently, the number of colors users can change is restricted to 24 colors. 

### Context
Currently the colors that are provided the pallet are all of the same shades and when we want to present any work related to annotations, it makes us stick to the colors provided.
For example:
Let's say I have a class - road and I have drawn the segmentation mask for the same on an image. To represent that it is a road, I'd want to choose a grey color. So when a person looks at the annotation, he can understand what the segmentation mask is trying to represent without having to think much about it. But now, I have to stick to the provided colors.

ActiveChooN:
### Motivation and context

This PR helps user to change predefined label color. When creating a label, it is assigned a color from a color map. Also ability to change this color from UI is available.

изображение

Resolve #1480, resolve #1572.

### How has this been tested?

Manually

### Checklist

### License

```python
#
```


cvat/apps/engine/migrations/0026_labelcolor.py
```diff
@@ -0,0 +1,33 @@
...
+
+    dependencies = [
+        ('engine', '0025_auto_20200324_1222'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='label',
+            name='color',
+            field=models.CharField(default='', max_length=32),
```
zhiltsov-max, line 27: Shouldn't the length be 6 or 8 or so?

cvat/apps/dataset_manager/formats/utils.py
```diff
@@ -0,0 +1,64 @@
...
+    with open(file_path) as f:
+        for line in f:
+            line = line.strip()
+            if not line or line[0] == '#':
+                continue
+            _, label, color = line.split(':')
+            colors[label] = tuple(map(int, color.split(',')))
+    return colors
+
+def normalize_label(label):
```
zhiltsov-max, line 27: The function is not used anymore, but it should be to avoid problems with symbol case and special characters.

The logic is:
- sort labels to provide stable results (across different tasks with same labels)
- find matching labels using normalization
- if have conflicts, the first match uses the table
- generate random colors for the rest

cvat/apps/dataset_manager/formats/mask.py
```diff
@@ -2,19 +2,17 @@
...
 from cvat.apps.dataset_manager.util import make_zip_archive
-from datumaro.cli.util import make_file_name
 from datumaro.components.project import Dataset
-from datumaro.util.mask_tools import generate_colormap
 
 from .registry import dm_env, exporter, importer
+from .utils import make_colormap
 
 
 @exporter(name='Segmentation mask', ext='ZIP', version='1.1')
```
zhiltsov-max, line 18: `background` must always be the `#0` color here.

cvat/apps/dataset_manager/formats/utils.py
ActiveChooN: Now colors are set at  the time of label creation. So i guess sorting by its name can not provide stable result now. I can not create labels with the same name, so are there any conflicts?

zhiltsov-max: Conflicts appear from labels like `BACKGROUND`, `_BA_CK_gro_und_`, `back_GROUND` etc after their normalization. I feel that sorting might be unnecessary anymore, or should be done in label creation code, when it is done in batch mode.

bsekachev:
@ActiveChooN 

When we change label color on the annotation view (Enter shortcut), is it expected to be changed on a server? 
Annotation view color picker still has default preset of colors, are you planning to change it?

nmanovic:
@ActiveChooN ,

What is the status of the PR?

Several comments from my side:
- Need to have a button "OK" when you select a color, "Cancel", and "Reset to default"

ActiveChooN:
@nmanovic, finishing work on annotation view. I'm totally agree about "reset to default". But changing color proceed immediately after picking. So I don't think button "Ok" is necessary. What do you think?

nmanovic:
> finishing work on annotation view. I'm totally agree about "reset to default". But changing color proceed immediately after picking. So I don't think button "Ok" is necessary. What do you think?

I think that need to replace "Label color" by a color pictogram for sure. I think that colors for default labels should be reviewed one more time. Also press X isn't intuitive way to choose a color.

nmanovic:
@ActiveChooN , looks like colors of labels are not matched between the task view and annotation view. Is it something you are working right now?

nmanovic:
Let's add a similar menu:

image

nmanovic:
Create new task dialog:
-	Use color picker icon instead of “Label color”
-	The color picker should show the current color

Color picker dialog:
-	It should have “OK”, “Cancel” and “Reset” buttons (see Gimp screenshot)

Job view:
-	Use “color by” == “label” by default
-	Use color picker dialog implemented for create new task dialog
-	Use color picker icon to choose a color
-	You can choose color only if “color by” equals to “instance” or “group”. Thus “Enter” and “color picker” is enabled only when “color by” != “label”.
-  Don't save selected colors on the server

cvat/apps/engine/migrations/0028_labelcolor.py
```diff
@@ -0,0 +1,35 @@
...
+from django.db import migrations, models
+from cvat.apps.dataset_manager.formats.utils import get_color_from_label_name, normalize_label, rgb2hex
+
+def alter_label_colors(apps, schema_editor):
+    Label = apps.get_model('engine', 'Label')
+    Task = apps.get_model('engine', 'Task')
+
+    for task in Task.objects.all():
+        labels = Label.objects.filter(task_id=task.id)
+        normalized_names = [normalize_label(l.name) for l in labels]
```
nmanovic, line 11: I see the code in 3 places. Need to remove code cloning because the logic isn't trivial.

Will `Road` and `road` have the same color? Is it OK to have the same color for different classes?

ActiveChooN: @nmanovic, sorry for a long wait. Realized that default python hash function has a random seed every run.

> I see the code in 3 places. Need to remove code cloning because the logic isn't trivial.

Done.

> Will Road and road have the same color? Is it OK to have the same color for different classes?

If that labels will appear in the task only once they will have the same color. But if label will appear in the task more than one time, for the first label color will be taken from the color map and color will be calculated with hash for others.

PR #2092, opened 2020-08-27
Cypress tests via Firefox browser.

dvkruchinin:
Added a Firefox browser for running Cypress tests.
Added browser verification functionality.
Added user and tasks removing functionality.

### Checklist

### License

```python
#
```

### DIFF
--- a/tests/cypress/integration/remove_users_tasks.js
+++ b/tests/cypress/integration/remove_users_tasks.js
@@ -1,0 +1,69 @@
+/*
+ * Copyright (C) 2020 Intel Corporation
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+/// <reference types="cypress" />
+
+let authKey = ''
+
+describe('Delete users and tasks created during the test run.', () => {
+    it('Get token', () => {
+        cy.request({
+            method: 'POST',
+            url: '/api/v1/auth/login',
+            body: {
+                username: Cypress.env('user'),
+                password: Cypress.env('password')
+            }
+        })
+        .then((responce) => {
+            authKey = responce['body']['key']
+        })
+    })
+    it('Get a list of users and delete all except id:1', () => {
+        cy.request({
+            url: '/api/v1/users',
+            headers: {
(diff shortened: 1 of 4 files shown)


tests/cypress/integration/remove_users_tasks.js
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+/// <reference types="cypress" />
+
+let authKey = ''
+
+describe('Delete users and tasks created during the test run.', () => {
+    it('Get token', () => {
```
bsekachev, line 12: I would suggest to make this method ``async`` and use ``await`` inside instead of promises.

tests/cypress/integration/remove_users_tasks.js
```diff
@@ -0,0 +1,69 @@
...
+            }
+        })
+    })
+    it('Get a list of tasks and delete them all', ()=> {
+        cy.request({
+            url: '/api/v1/tasks',
+            headers: {
+                Authorization: `Token ${authKey}`
+            }
+        })
```
bsekachev, line 54: Can we use ``cvat-core`` instead of these direct requests here and above?
For example to get task list you can use ``cvat.tasks.get(...)``.
What's more as far as I remember, this request doesn't return all task (only first 10). Then need to use parameter ``page`` to get more

One thing ``cvat-core`` isn't able to do is to delete a user. We can leave this case as is, or I can add api method for this. We can have a call and discuss it

tests/cypress/integration/remove_users_tasks.js
dvkruchinin: Done. Added method `async` and use `await` for responce.

tests/cypress/integration/remove_users_tasks.js
dvkruchinin: In the current moment, as we agreed in discuss, I changed the link to search for tasks by adding `?page_size=1000`

bsekachev (approved):

Issue #1769, opened 2020-06-21
Bug: Polygon must have at least 3 points


cinjon:
I am using cvat.org to test out the service. I used the MaskRCNN feature to annotate a video. After completion, I went into the job and found that I could not open it. I get this error in the console:

`cvat-app.tsx:192 Error: Polygon must have at least 3 points, but got 2`

I also see the same error in the window:

`Error: Polygon must have at least 3 points, but got 2`

However, I can open it in the Old UI no problem. This seems like a bug.

bsekachev:
@cinjon 

Could you provide annotations data that you get from server to client? 
I suspicious there are wrong polygons.

cinjon:
I just tried dumping as COCO 1.0 Annotations and got the error:

```
Could not dump annotations for the task 1690
Error: Request failed with status code 500. "AssertionError: Wrong polygon points: [395.0, 85.5, 395.0, 85.5]\n".
```

bsekachev:
@cinjon 

It seems there is a bug in ``MaskRCNN``. Polygon mustn't have less then 3 points. But the algorithm have returned the wrong result.
Another idea is instead of the error we might output a warning and continue working in UI. 

Let's look into this.

bsekachev:
Speaking about dumping. @zhiltsov-max what do you think? Should we have the assert?

zhiltsov-max:
We have wrong annotations, assert is here, and it fires. Moreover, these annotations are incompatible with the official COCO python API.

savan77:
I am also facing the same issue. The annotations were generated by the MaskRCNN model. Did you guys find any workaround?  @zhiltsov-max @bsekachev

zhiltsov-max:
@savan77, there is 7 annotations with less than 6 values: 52, 100, 154, 176, 184, 285, 299. You're right about the function in COCO API.

rushtehrani:
Is there a PR for this already? Looked through the PRs but couldn’t see anything. Our workaround right now is to remove the check in UI code.

zhiltsov-max:
@savan77, actually, lambda functions have already come. It would be great, if you could prepare a PR with this fix. I also think that it can be interesting to check for small area polygons (e.g. < 1 px^2), instead of just 4-value arrays.

savan77:
@zhiltsov-max That's great. I think it will be easier to just check for length not only logically but also computationally. Let me know what do you think, happy to send a pr.

zhiltsov-max:
[...]

PR #2100, opened 2020-08-28
fix: discard polygons with length of 4 or less

savan77:
It was found that Matterport's MaskRCNN implementation generates polygons with length of 4 which creates lot of issues since pycocotools considers them as a bounding box. It looks like these polygons with length of 4 are just garbage values. That is actually a single point repeated twice. So, we can just filter them out.
More discussion around this topic can be found here #1769.

### Motivation and context

It will solve issues where users were not able to see the job because there were polygons with length of 4. It will also solve an issue with export dataset feature as we go through pycocotools there as well.

### How has this been tested?

This has been tested locally as well as on Onepanel.

### Checklist

- [`x`] I submit my changes into the `develop` branch
- [`x` ] I have added description of my changes into [CHANGELOG](https://github.com/opencv/cvat/blob/develop/CHANGELOG.md) file

### License

- [`x` ] I submit _my code changes_ under the same [MIT License](

```python
#
```


serverless/tensorflow/matterport/mask_rcnn/nuclio/model_loader.py
```diff
@@ -63,7 +63,7 @@ def infer(self, image, threshold):
                 contour = np.flip(contour, axis=1)
                 # Approximate the contour and reduce the number of points
                 contour = approximate_polygon(contour, tolerance=2.5)
-                if len(contour) < 3:
+                if len(contour) < 5:
```
zhiltsov-max, line 66: It might be better to use 6 here, though.

zhiltsov-max (approved):

serverless/tensorflow/matterport/mask_rcnn/nuclio/model_loader.py
savan77: Agreed. Updated.

nmanovic:
@savan77 , need to add a line into CHANGELOG. Also please check `I submit my code changes under the same MIT License`.  Probably server itself should have corresponding checks and don't accept "bad" polygons.

savan77:
@savan77 Made changes as per your suggestions. Yeah, that might one possible way. Let me know what do you guys think. For now, this will solve multiple errors (https://github.com/opencv/cvat/issues/1769) that arises because of MaskRCNN generating polygons with two points (both are same).

nmanovic (approved):

nmanovic (approved):

nmanovic (approved):

PR #2054, opened 2020-08-19
Improved interface of interactors on UI

bsekachev:
### How has this been tested?
Manual testing

### Checklist

- [x] I submit my changes into the `develop` branch
- [x] I have added description of my changes into [CHANGELOG](https://github.com/opencv/cvat/blob/develop/CHANGELOG.md) file
- [ ] I have updated the [documentation](
  https://github.com/opencv/cvat/blob/develop/README.md#documentation) accordingly
- [ ] I have added tests to cover my changes
- [ ] I have linked related issues ([read github docs](
  https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword))
- [x] I have increased versions of npm packages if it is necessary ([cvat-canvas](https://github.com/opencv/cvat/tree/develop/cvat-canvas#versioning),
[cvat-core](https://github.com/opencv/cvat/tree/develop/cvat-core#versioning), [cvat-data](https://github.com/opencv/cvat/tree/develop/cvat-data#versioning) and [cvat-ui](https://github.com/opencv/cvat/tree/develop/cvat-ui#versioning))

### License

- [x] I submit _my code changes_ under the same [MIT License](
  https://github.com/opencv/cvat/blob/develop/LICENSE) that covers the project.
  Feel free to contact the maintainers if that's a concern.
- [x] I have updated the license header for each file (see an example below)

```python
# Copyright (C) 2020 Intel Corporation
#
# SPDX-License-Identifier: MIT
```

Checklist:
- [x] UI notifications & lock when the process is active (UI feedback)
- [x] Freeze history when updated points
- [x] Add description as a tooltip
- [x] Change icon for AI tools
- [x] Remove old DEXTR code
- [x] CTRL modifier key to maintain requests
- [ ] Update server interactors API (need some additional parameters and negative points)

PR #2102, opened 2020-08-31
Running detectors on the fly

### LINKED ISSUE #197: Support on-the-fly tensorflow inference
Currently, I believe the inferences runs entirely first, before any annotation can begin. This can be pretty slow, especially on slow hardware and large videos etc. I'm requesting that the model runs 'on demand'. To prevent a visible slow-down to the user, we should buffer (e.g. when I start annotating frame N, inferences starts on frame N + 3, so that it's complete by the time I get to it). This seems effective as I can start annotating nearly immediately, and there's no noticeable slow-down while annotating (as the time taken annotating a frame - which things are currently idea - should hopefully exceed model evaluation time).

Note - we'd actually have to consider how to add in support for e.g. evaluating the model only every Nth frame (e.g. when I'm annotating I generally only annotate every 10th frame, and use interpolation for the rest - so it makes sense to only run tensorflow on every 10th frame too).

This seems reasonably feasible - is there likely to be interest?

bsekachev:
### Motivation and context
Our customers asked as to implement the feature that allows to annotate only one specific frame with automatic detectors 
Resolve #197
The PR depends on #2054

### How has this been tested?
Manual testing

### Checklist

### License

```python
#
```

- [x] Show only detectors on annotation view
- [ ] Add ability to cancel annotation request
- [x] Prettify UI


ActiveChooN (approved):

nmanovic:
Please don't assume that if we map all labels from the task, it is enough. In general I can want to map multiple labels to one. For example, { car -> vehicle, bus -> vehicle, truck -> vehicle }.


Can we adjust labels by columns? The new version:
image

The dialog below looks much better in the old version:
image

bsekachev:
@nmanovic Sure. I will fix it.

nmanovic (approved):
Colors for labels should be taken from the task itself. Now "person" each time has own random color.

PR #2136, opened 2020-09-07
UI Tracking with serverless functions

bsekachev:
Screenshot from 2020-09-07 11-27-26
Screenshot from 2020-09-07 11-27-32
Screenshot from 2020-09-07 11-27-11

### Motivation and context
The patch uses server-side implemented serverless tracking to get automatic bounding boxes on client.

### How has this been tested?
Manual testing 

### Checklist

### License

```python
#
```

### DIFF
--- a/cvat-ui/src/actions/annotation-actions.ts
+++ b/cvat-ui/src/actions/annotation-actions.ts
@@ -1424,12 +1424,21 @@ export function repeatDrawShapeAsync(): ThunkAction {
 
         let activeControl = ActiveControl.CURSOR;
         if (activeInteractor) {
-            canvasInstance.interact({
-                enabled: true,
-                shapeType: 'points',
-                minPosVertices: 4, // TODO: Add parameter to interactor
-            });
-            dispatch(interactWithCanvas(activeInteractor, activeLabelID));
+            if (activeInteractor.type === 'tracker') {
+                canvasInstance.interact({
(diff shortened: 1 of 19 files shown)


cvat-ui/src/actions/annotation-actions.ts
```diff
@@ -1424,12 +1436,21 @@ export function repeatDrawShapeAsync(): ThunkAction {
...
+                canvasInstance.interact({
+                    enabled: true,
+                    shapeType: 'rectangle',
+                });
+                dispatch(interactWithCanvas(activeInteractor, activeLabelID));
+            } else {
+                canvasInstance.interact({
+                    enabled: true,
+                    shapeType: 'points',
+                    minPosVertices: 4, // TODO: Add parameter to interactor
```
nmanovic, line 1449: Please use parameter min_pos_points. By default, the value should be 1

bsekachev: In JS I would prefer to use camelCase even if from the server we get ``snake_case`` as it done for all model fields in cvat-core (model, user, etc.). Knowing exact fields that we get from server we always can rename them in ``cvat-core`` 

If we set value ``1`` here, DEXTR will not work correctly. First need to add parameter to all interactors

nmanovic (approved):

PR #2192, opened 2020-09-17
Nuclio as a plugin in CVAT, improved system to check installed plugins

dmitryagapov:
### Checklist

### License

```python
#
```


cvat-core/src/server-proxy.js
```diff
@@ -799,7 +799,7 @@
             }
 
             async function cancelLambdaRequest(requestId) {
-                const { backendAPI } = config;
+                const {backendAPI} = config;
```
bsekachev, line 802: Please, setup eslint in your dev environent (install eslint extension for vscode and all npm packages).
Here is eslint error: ``eslintobject-curly-spacing``

cvat-core/src/server-proxy.js
```diff
@@ -812,6 +812,16 @@
                 }
             }
 
+            async function getAllowedApps() {
+                const {backendAPI} = config;
```
bsekachev, line 816: The same

cvat-core/src/server-proxy.js
```diff
@@ -812,6 +812,16 @@
...
+                try {
+                    const response = await Axios.get(`${backendAPI}/apps/`);
```
bsekachev, line 818: Please, add a proxy setting to the request to be consistent (look at other requests)

cvat-ui/src/actions/meta-action.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    try {
+        const allowedApps: string[] = await core.allowedApps.list();
+
+        const data: MetaState = {
+            showTasksButton: allowedApps.includes('tasks'),
+            showAnalyticsButton: allowedApps.includes('analytics'),
+            showModelsButton: allowedApps.includes('serverless'),
+        };
+        console.log(data);
```
bsekachev, line 37: Please, remove the statement here

cvat-core/src/api.js
```diff
@@ -470,34 +469,42 @@ function build() {
...
                 return result;
             },
         },
+
+        allowedApps: {
+            async list() {
+                const result = await PluginRegistry.apiWrapper(cvat.allowedApps.list);
+                return result;
+            },
+        },
```
bsekachev, line 491: Why not to use ``plugins.list()`` to get this information?
``plugin.list()`` is used to get available apps from the server ``analytics``, ``git``.

cvat-ui/src/reducers/meta-reducer.ts
```diff
@@ -0,0 +1,46 @@
+import { MetaState } from './interfaces';
```
bsekachev, line 1: ``Meta`` is not an informative name. Let's find another. And why did you create new reducer & actions instead of using ``plugins`` that are dedicated for that?

cvat-ui/src/components/header/header.tsx
```diff
@@ -256,13 +269,14 @@ function HeaderContainer(props: Props): JSX.Element {
 
         </Menu>
     );
+    console.log(showTasksButton);
```
bsekachev, line 272: Extra ``console.log`` statement

bsekachev (changes requested):

bsekachev:
Why have you decided to hide ``Models`` button when optionally?
1. We are still able to go manually to ``/models``
2. Early CVAT showed instruction to deploy a new model if no models available. So, user understood that CVAT is able to work with models.
3. There are another components (like button ``Automatic Annotation`` (and probably others) which we should hide in this case.

cvat-core/src/api.js
dmitryagapov: Because the same elements shoud have the same behavior. But now analytics, models and tasks implementation is different. And implementation of analitycs and git different also. It first step to start working with them in the same way.

cvat-ui/src/reducers/meta-reducer.ts
dmitryagapov: Meta means that it is the different level of abstraction for other state element. Because based on this data we decide how much functionality of application will be available.

dmitryagapov:
This solution based on explanation of @nmanovic.

cvat-core/src/api.js
bsekachev: What do you mean speaking that implementation is different?
Right now ``models`` are built-in like ``tasks``. 

Checking whether server has applications ``git`` and ``analytics`` happens from ``cvat-ui`` (in ``plugin-cheker.ts``) what is of course wrong approach (need to test it via ``cvat-core`` [where we might add method like ``serverList`` to namespace ``plugins``]), but in context of cvat client, ``analytics`` and ``git`` are plugins.

The patch introduced some rather abstract ``meta`` what is actually looks like the same what is plugins.

bsekachev: And speaking about the PR's name. How is it related with these code changes?
I just try to understand what task you are trying to solve

dmitryagapov: The goal of this pr hide the models button

cvat-ui/src/reducers/meta-reducer.ts
bsekachev: > Because based on this data we decide how much functionality of application will be available.

Sounds like plugins :)

It looks like there are some things to discuss. Let's do it with the team on today's meeting or have a different call.

bsekachev (dismissed):

bsekachev:
@zhiltsov-max @nmanovic Hey guys, could you please look at server part of the patch?

cvat/apps/plugins/views.py
```diff
@@ -0,0 +1,24 @@
...
+
+    def list(self, request):
+        result = {
+            'GIT_INTEGRATION': False,
+            'ANALYTICS': False,
+            'MODELS': False,
+        }
+        if apps.is_installed('cvat.apps.git'):
+            result['GIT_INTEGRATION'] = True
+        if os.environ.get("CVAT_ANALYTICS", '0') == '1':
```
nmanovic, line 20: Let's use https://docs.python.org/3/distutils/apiref.html#distutils.util.strtobool

components/analytics/docker-compose.analytics.yml
```diff
@@ -26,18 +26,18 @@ services:
       context: ./components/analytics/kibana
       args:
         ELK_VERSION: 6.4.0
-    depends_on: ['cvat_elasticsearch']
+    depends_on: [ 'cvat_elasticsearch' ]
```
nmanovic, line 29: What is the reason to add extra spaces here?

components/analytics/docker-compose.analytics.yml
```diff
@@ -26,18 +26,18 @@ services:
...
     image: cvat/server
-    volumes: ['./components/analytics/kibana:/home/django/kibana:ro']
-    depends_on: ['cvat']
+    volumes: [ './components/analytics/kibana:/home/django/kibana:ro' ]
+    depends_on: [ 'cvat' ]
     working_dir: '/home/django'
-    entrypoint: ['bash', 'wait-for-it.sh', 'elasticsearch:9200', '-t', '0', '--',
-      '/bin/bash', 'wait-for-it.sh', 'kibana:5601', '-t', '0', '--',
-      '/usr/bin/python3', 'kibana/setup.py', 'kibana/export.json']
+    entrypoint: [ 'bash', 'wait-for-it.sh', 'elasticsearch:9200', '-t', '0', '--',
```
nmanovic, line 38: Could you please keep indentation as is?

cvat-ui/src/components/cvat-app.tsx
```diff
@@ -316,7 +319,8 @@ class CVATApplication extends React.PureComponent<CVATAppProps & RouteComponentP
                                         <Route exact path='/tasks/create' component={CreateTaskPageContainer} />
                                         <Route exact path='/tasks/:id' component={TaskPageContainer} />
                                         <Route exact path='/tasks/:tid/jobs/:jid' component={AnnotationPageContainer} />
-                                        <Route exact path='/models' component={ModelsPageContainer} />
+                                        { isModelPluginActive
```
nmanovic, line 322: I agree with Boris. It is better to show instructions how to enable the feature in UI.

cvat-ui/src/components/header/header.tsx
```diff
@@ -276,38 +281,40 @@ function HeaderContainer(props: Props): JSX.Element {
...
-                    className='cvat-header-button'
-                    type='link'
-                    value='models'
-                    href='/models'
-                    onClick={
-                        (event: React.MouseEvent): void => {
-                            event.preventDefault();
-                            history.push('/models');
+
+                {isModelsPluginActive && (
```
nmanovic, line 285: Show instructions if models plugin isn't available.

cvat-ui/src/components/header/header.tsx
```diff
@@ -276,38 +281,40 @@ function HeaderContainer(props: Props): JSX.Element {
...
-                                    event.preventDefault();
-                                    // false positive
-                                    // eslint-disable-next-line
-                                    window.open(`${tool.server.host}/analytics/app/kibana`, '_blank');
-                                }
+                    >
+                        Models
+                    </Button>
+                )}
+                {isAnalyticsPluginActive && (
```
nmanovic, line 301: Show instructions if analytics plugin isn't available.

cvat/apps/plugins/views.py
```diff
@@ -0,0 +1,24 @@
+import os
+
+from django.apps import apps
+from rest_framework import viewsets
+from rest_framework.response import Response
+
+
+# Create your views here.
+
+class Plugins(viewsets.ViewSet):
```
nmanovic, line 10: Let's implement it not as a separate app but as a part of the engine app (/api/v1/server/plugins). In this case the feature is so tiny feature that I don't think we need a separate django app.

cvat/apps/lambda_manager/urls.py
```diff
@@ -2,9 +2,10 @@
 #
 # SPDX-License-Identifier: MIT
 
+from django.urls import include
```
zhiltsov-max, line 5: `import include, path`

cvat-ui/src/components/cvat-app.tsx
dmitryagapov: I think it better to do in separate PR.

bsekachev:
Let's merge the PR and then add some instructions to activate analytics and models if they are disabled. 
I will create an issue in advance and assign it to @dmitryagapov

PR #2219, opened 2020-09-24
Size limit improvement

PMazarovich:
### Motivation and context

Sometimes it is required to have more flexibility in setting limits for data files in CVAT.
Now all you need to increase max video size is creating 3 environment variables.

### How has this been tested?

It was tested manually

### Checklist

### License

```python
#
```


cvat/settings/base.py
```diff
@@ -397,10 +397,10 @@ def generate_ssh_keys():
     LOGGING['loggers']['cvat.server']['handlers'] += ['logstash']
     LOGGING['loggers']['cvat.client']['handlers'] += ['logstash']
 
-DATA_UPLOAD_MAX_MEMORY_SIZE = 100 * 1024 * 1024  # 100 MB
-DATA_UPLOAD_MAX_NUMBER_FIELDS = None   # this django check disabled
-LOCAL_LOAD_MAX_FILES_COUNT = 500
-LOCAL_LOAD_MAX_FILES_SIZE = 512 * 1024 * 1024  # 512 MB
+DATA_UPLOAD_MAX_MEMORY_SIZE = int(os.getenv('CVAT_MAX_MEMORY_SIZE', '104857600'))  # 100 MB default
```
nmanovic, line 400: If we add an env variable for every item in the config file, it will be mess. If you need to change settings, there is a better solution:
- Create your own `mysettings.py` file (name it as you like)
- Inside the `mysettings.py` file import appropriate CVAT configuration (e.g.  `from cvat.settings.production import *`)
- Modify necessary variables (e.g. DATA_UPLOAD_MAX_MEMORY_SIZE)
- Use DJANGO_SETTINGS_MODULE to point to the new settings file for your instance.

The approach is more flexible. You can redefine your settings as you like.

Do you still need the PR or the described approach will work for you?

nmanovic:
@PMazarovich , I will close the PR. I don't see any value to redefine multiple variables inside settings.py. It is better to redefine them inside own settings.py file and point django to the new settings.py file.

412b:
@nmanovic I'm a bit too late for this one, but the value I see here is that one doesn't need to create custom docker image to have those values adjusted (of course, mounting some fs with the settings can also be an option, but env vars for docker image is still way simplier)...

nmanovic:
@412b , It is approach our team uses inside. I understand that it is easier. But need to maintain complexity. Settings with 100 of env variables are not something which we want to get at the end.

Could you please describe your use case? How are you using CVAT internally?

@azhavoro , @bsekachev , @vnishukov , @zhiltsov-max 

Please share your option about that. Let's vote and find a consensus.

412b:
@nmanovic we're using cvat as docker images (server and ui) run on ECS Fargate in the AWS with EFS as local FS for persistent data, RDS for postgres and ElastiCache for redis.
due to the fact that we have EFS, using `share` folder is not really different from using web upload and 500MiB is not really enough due to our videos being 1+ hours, scale of x*10k (sometimes 100k) frames.
I'm also not in favour of exporting a lot of env vars, but our goal is to have vanilla cvat docker images and do all of the needed overrides in most simple and straight-forward way.
of course, in worst case we can give it a try with having custom settings reside somewhere in the EFS as well, but that's not architectually transparent and adds just more layers to the setup itself.

as a side note I can always refer to Typesafe's/Lightbend's Config, which allows overriding literally any configuration setting via environment. perhaps, it can be not env vars, but some kind of args being passed, though it's a bit tricky due to supervisord in front of cvat itself.

zhiltsov-max:
Django has lots of configuration options, so I don't feel like provision of tons of env vars is a good approach. However, I think, it could be useful to provide a variable with a path to a local config file, which would be mounted into the container. Either way, it will require docker-compose modification.

nmanovic:
@412b ,

> I'm also not in favour of exporting a lot of env vars, but our goal is to have vanilla cvat docker images and do all of the needed overrides in most simple and straight-forward way. of course, in worst case we can give it a try with having custom settings reside somewhere in the EFS as well, but that's not architectually transparent and adds just more layers to the setup itself.

I agree with Maxim. In the case it is better to update 1 file. I believe the number of variables which you want to redefined will grow over time. I have to close the PR because CVAT has an official way (django supported) to redefine all changes in a custom settings file.

PR #2234, opened 2020-09-28
Allow to login into UI with token from api/v1/auth/login

tdowgiel:
### Motivation and context

Allow to automatically log into CVAT-UI by 3rd part apps on the same cluster

### How has this been tested?
1. In bash:
```bash
curl localhost:8080/api/v1/auth/login -X POST -H "Content-Type: application/json" -d '{"username": "some_name", "password": "some_password"}' -v # Get csrftoken & session id from the response
```
2. Make URL: http://localhost:8080/auth/login-with-token/:sessionid/:csrftoken
3. Paste it into a browser

### Checklist

### License

```python
#
```


bsekachev:
Hi, @tdowgiel 
Is this PR ready? Can we review it?

tdowgiel:
Thanks for asking, I’m in middle of testing it. Probably it won’t change a lot. I will give you sign when it will be fully tested.

bsekachev:
@tdowgiel 

Okay, nevertheless, let me comment some code-style issues and ideas about this PR.

cvat-ui/src/components/login-with-token/login-with-token.tsx
```diff
@@ -0,0 +1,21 @@
...
+// SPDX-License-Identifier: MIT
+
+import React from 'react';
+import {Redirect, useParams} from "react-router";
+import {useCookies } from "react-cookie";
+
+export default function LoginWithTokenComponent(){
+    const { sessionId, token } = useParams()
+    const [cookies, setCookie] = useCookies(['sessionid', 'csrftoken'])
+    const expires = new Date(new Date().setFullYear(new Date().getFullYear() + 1))
```
bsekachev, line 12: Does the cookies actually expire after one year? Does it correspond to the server?

bsekachev:
Also could you please increase minor version of ``cvat-ui`` (``npm version minor`` in ``cvat/cvat-ui``). You've checked this item in the PR description, but I don't see it is done

cvat-ui/src/components/login-with-token/login-with-token.tsx
tdowgiel: token expires after one year. However session id after 2 weeks, thanks for asking. Fixed

cvat-ui/src/components/login-with-token/login-with-token.tsx
```diff
@@ -0,0 +1,21 @@
...
+    setCookie('sessionid', sessionId, {path: '/', expires })
+    setCookie('csrftoken', token, {path: '/', expires})
```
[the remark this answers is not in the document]
tdowgiel: Sorry. I i'm only a little bit experienced with react. And used to use FC as standard. How i could pass url parameters into 
CVATApplication ?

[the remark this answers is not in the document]
bsekachev: Maybe something like this:

```ts
import { matchPath } from 'react-router';
...

 const match = matchPath('/auth/login-with-token/:sessionId/:token', {
            path: location.pathname,
            exact: true,
        });

if (match) {
    const { sessionId, token } = match.params;
    ....
}
```

cvat-ui/src/components/login-with-token/login-with-token.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+    const expires1y = new Date(new Date().setFullYear(new Date().getFullYear() + 1))
+    const expires2w = new Date(new Date().setDate(new Date().getDate() + 13))
+
+    setCookie('sessionid', sessionId, {path: '/', expires: expires2w })
+    setCookie('csrftoken', token, {path: '/', expires: expires1y})
+
+    if ( cookies['sessionid'] && cookies['csrftoken']) {
+        window.location.reload();
+        <Redirect to="/tasks" />
```
bsekachev, line 21: I suppose redirect after reload is useless. Moreover it should be wrapped into ``return ();``
Maybe something like ``window.location = window.location.origin;`` would be more suitable

bsekachev:
Hi, @tdowgiel, I hope you are doing well.

I tried this PR and it doesn't seem to be a working solution. 
Maybe I did something wrong, but in this case could you please provide steps to test it? (I just copied sessionid and token from active session, made URL, pasted it, and browser infinitely reloads the page). 

Also POST request to ``http://localhost:8080/api/v1/auth/login`` with credentials returns json with ``key`` field, which is actually used to authorization via ``cvat-core`` and stored in the local storage, but I do not see it is used somehow in this patch.

cvat-ui/src/components/login-with-token/login-with-token.tsx
tdowgiel: For some reason. I need to reload window to apply this changes.  However, i have fixed issue with endless reloading when wrong token, by moving this to useEffect

tdowgiel:
> Hi, @tdowgiel, I hope you are doing well.
> 
> I tried this PR and it doesn't seem to be a working solution.
> Maybe I did something wrong, but in this case could you please provide steps to test it? (I just copied sessionid and token from active session, made URL, pasted it, and browser infinitely reloads the page).
> 
> Also POST request to `http://localhost:8080/api/v1/auth/login` with credentials returns json with `key` field, which is actually used to authorization via `cvat-core` and stored in the local storage, but I do not see it is used somehow in this patch.

I have tested it in many ways. Response used in cvat-core is not only returning key but also setting cookie because of headers. Please look on example response and set-cookie header:
```
Date: Fri, 09 Oct 2020 13:35:16 GMT
Server: WSGIServer/0.2 CPython/3.8.3
Content-Type: application/json
Vary: Accept, Cookie, Origin
Allow: POST, OPTIONS
X-Frame-Options: DENY
Content-Length: 50
X-Content-Type-Options: nosniff
Referrer-Policy: same-origin
Set-Cookie: csrftoken=bAd9852nQZipwTvFnuELs29mdwowc3BGjuCsjpMfLrWT3txGVtaDp4jIboMBU1cq; expires=Fri, 08 Oct 2021 13:35:16 GMT; Max-Age=31449600; Path=/; SameSite=Lax sessionid=i8xqtviqysnfbi9n63rmhici3rb1nfo7; expires=Fri, 23 Oct 2020 13:35:16 GMT; HttpOnly; Max-Age=1209600; Path=/; SameSite=Lax
```
This is scenario:
I have access to the server with cvat.  I want to integrate CVAT with my other webapp. And i want to share users between my app and cvat. Also core goal is to have user already logged in when he decide to use cvat.

1. Create django user on CLI
2. My backend send python request to cvat server to login django user
3. CVAT django server returns key, but also mentioned before 'Set Cookie' params. 
4. User autologs in to CVAT thanks to /login-with-token/ parameters

-

cvat-ui/src/components/login-with-token/login-with-token.tsx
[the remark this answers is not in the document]
tdowgiel: As discussed. If it is possible to leave this as separate component  it will be easier to integrate.

cvat-ui/src/components/login-with-token/login-with-token.tsx
```diff
@@ -0,0 +1,26 @@
+// Copyright (C) 2020 Intel Corporation
+//
+// SPDX-License-Identifier: MIT
+
+import React, { useEffect } from 'react';
+import { Redirect, useParams } from 'react-router';
+import { useCookies } from 'react-cookie';
+
+export default function LoginWithTokenComponent(){
```
bsekachev: Please, add return type for the function ``JSX.Element``

cvat-ui/src/components/cvat-app.tsx
```diff
@@ -10,6 +10,7 @@ import notification from 'antd/lib/notification';
 import Spin from 'antd/lib/spin';
 import Text from 'antd/lib/typography/Text';
 import GlobalErrorBoundary from 'components/global-error-boundary/global-error-boundary';
+import LoginWithTokenComponent from './login-with-token/login-with-token';
```
bsekachev, line 13: Please, use absolute import: ``components/login-with-token/login-with-token``, or put the line after all absolute (linter says that relative import should occur after absolute)

bsekachev:
That's amazing, thanks for your patience and for the contribution!

PR #2266, opened 2020-10-07
Cypress test. Actions on polygon.

dvkruchinin:
Try to draw a polygon shape/track
Try to use parameter “number of points”, polygon drawing should automatically stop after this number of points is achieved
Try to switch a label and draw one more polygon
Check if everything works here.

Adapt the tests to use the updated polygon drawing function.
Redesigned the polygon drawing function for using an object as an argument.

### Checklist

### License

```python
#
```

### DIFF
--- a/tests/cypress/integration/case_10_polygon_shape_track_label_points.js
+++ b/tests/cypress/integration/case_10_polygon_shape_track_label_points.js
@@ -1,0 +1,173 @@
+/*
+ * Copyright (C) 2020 Intel Corporation
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+/// <reference types="cypress" />
+
+context('Actions on polygon', () => {
+
+    const caseId = '10'
+    const labelName = `Case ${caseId}`
+    const taskName = `New annotation task for ${labelName}`
+    const attrName = `Attr for ${labelName}`
+    const textDefaultValue = 'Some default value for type Text'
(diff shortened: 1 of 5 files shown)


tests/cypress/integration/case_10_polygon_shape_track_label_points.js
```diff
@@ -0,0 +1,173 @@
...
+            ],
+        complete: true,
+        numberOfPoints: null
+    }
+
+    before(() => {
+        cy.visit('auth/login')
+        cy.login()
+        cy.imageGenerator('cypress/fixtures', image, width, height, color, posX, posY, labelName)
+        cy.createAnnotationTask(taskName, labelName, attrName, textDefaultValue, image)
```
bsekachev, line 105: I would suggest do not create a different annotation task for each test in the future (if test doesn't require it). It would be much more efficient if we created a couple of annotation tasks (video, images) with several labels and attributes and used it for all such tests. We might consider such optimization like a separated request. What do you think?

bsekachev (approved):

tests/cypress/integration/case_10_polygon_shape_track_label_points.js
dvkruchinin: Yes, this can be considered as part of a separate task. I will need to think about how best to combine the tests. And of course it will speed up the tests.

PR #2295, opened 2020-10-09
The change in the measurement of CVAT-UI code coverage.

dvkruchinin:
Update Cypress to 5.3 (Firefox support is out of beta)
The use of the plugin "@cypress/code-coverage" for running tests after code instrumentations. 

### Motivation and context
Speed up passing tests after code instrumentation using the Firefox browser. This change made it possible to reduce the launch time of each test by more than 3 times. In addition, don't need to run tests via Google Chrome to measure coverage.

### Checklist

### License

```python
#
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,25 +33,23 @@ before_script:
   - chmod a+rwx ${HOST_COVERAGE_DATA_DIR}
 
 script:
-  - docker-compose -f docker-compose.yml -f docker-compose.ci.yml run cvat_ci /bin/bash -c 'coverage run -a manage.py test cvat/apps utils/cli && mv .coverage ${CONTAINER_COVERAGE_DATA_DIR}'
-  - docker-compose -f docker-compose.yml -f docker-compose.ci.yml run cvat_ci /bin/bash -c 'cd cvat-data && npm install && cd ../cvat-core && npm install && npm run test && coveralls-lcov -v -n ./reports/coverage/lcov.info > ${CONTAINER_COVERAGE_DATA_DIR}/coverage.json'
   - docker-compose up -d
-  # Create superuser
-  - docker exec -it cvat bash -ic "echo \"from django.contrib.auth.models import User; User.objects.create_superuser('${DJANGO_SU_NAME}', '${DJANGO_SU_EMAIL}', '${DJANGO_SU_PASSWORD}')\" | python3 ~/manage.py shell"
+  - docker exec -it cvat bash -ic "python3 ~/manage.py migrate && echo \"from django.contrib.auth.models import User; User.objects.create_superuser('${DJANGO_SU_NAME}', '${DJANGO_SU_EMAIL}', '${DJANGO_SU_PASSWORD}')\" | python3 ~/manage.py shell"
   # End-to-end testing
-  - if [[ $TRAVIS_EVENT_TYPE == "cron" || $TRAVIS_BRANCH == "dkru/measure_coverage_experiments_for_testing" ]];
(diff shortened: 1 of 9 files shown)


.travis.yml
```diff
@@ -33,29 +33,23 @@ before_script:
   - chmod a+rwx ${HOST_COVERAGE_DATA_DIR}
 
 script:
-  - docker-compose -f docker-compose.yml -f docker-compose.ci.yml run cvat_ci /bin/bash -c 'coverage run -a manage.py test cvat/apps utils/cli && mv .coverage ${CONTAINER_COVERAGE_DATA_DIR}'
```
bsekachev, line 36: Can we leave it as is?

tests/package.json
```diff
@@ -1,12 +1,12 @@
 {
   "scripts": {
-    "cypress:run:chrome": "cypress run --headless --browser chrome",
+    "cypress:run:chrome": "cypress run --env coverage=false --headless --browser chrome",
```
bsekachev, line 3: Is this enabled by default?
Can we disable coverage for Firefox by default? (by passing the same argument) And override it in .travis.yml?

bsekachev: Do you use this arg in the code explicitly?

.travis.yml
```diff
@@ -33,29 +33,23 @@ before_script:
...
       npm run report:coverage:lcov;
     else
       cd ./tests && npm install &&
       npm run cypress:run:chrome && cd ..;
     fi;
-
-
-after_success:
-  # https://coveralls-python.readthedocs.io/en/latest/usage/multilang.html
-  - docker-compose -f docker-compose.yml -f docker-compose.ci.yml run cvat_ci /bin/bash -c 'ln -s ${CONTAINER_COVERAGE_DATA_DIR}/.git . && ln -s ${CONTAINER_COVERAGE_DATA_DIR}/.coverage . && ln -s ${CONTAINER_COVERAGE_DATA_DIR}/coverage.json . && coveralls --merge=coverage.json'
```
bsekachev, line 61: I've just realized that the coverage is computed also for a server part. Could you please investigate if we could have two different badges for client and for server? If it is impossible, anyway we need left coverage for server part as is and merge results for the same badge (maybe also we should to run server coverage only on cron builds)

.travis.yml
dvkruchinin: In this case, the coverage measurement readings will alternate depending on the type of CI being triggered. This command also sends data to the coveralls server.

tests/package.json
dvkruchinin: Yes, coverage measurement is enabled by default in the plugin.
Alternatively, you can make 2 commands for each launch. With and without coverage measurement.
I use it as an argument for running through npm.

.travis.yml
dvkruchinin: Well. I will investigate this topic. It will probably be easier to combine the results.

dvkruchinin:
Redesigned the logic for running tests and measuring coverage. In cron, run tests via Firefox without measuring coverage. For the rest, run tests via Chrome with coverage measurement. Combining coverage measurement results for core and 2e2 tests.

bsekachev:
@dvkruchinin 
Please, let me know when the PR is ready. Just remove WIP tag

dvkruchinin:
> @dvkruchinin
> Please, let me know when the PR is ready. Just remove WIP tag

I got a crash error for the first test after rebuilding containers after code instrumentation. When logging in, a modal window appeared with the text "Could not get user agreements from the server". To avoid this, I added a 10-second wait to the first test.

bsekachev (approved):

PR #2345, opened 2020-10-22
Don't export outside frames in frame-based formats

### LINKED ISSUE #2339: Outside property not used for rectangle tracks at CVAT for Images 1.1

We use interpolation of rectangles to annotate objects in annotation mode. Occasionally a rectangle is not visible so we mark it as "outside". The dumped annotation xml using "CVAT for Images 1.1" still includes entries for the "outside" rectangles at the frames where they were marked as "outside" without any indication that they are actually outside. Using CVAT version v.1.1.0

Related: https://github.com/openvinotoolkit/cvat/issues/1620 and https://github.com/openvinotoolkit/cvat/pull/1729

### Expected Behaviour

The expected behaviour is to exclude entries for the rectangles that are marked as "outside" (as was the case with "hidden" at v.0.6.1).

### Current Behaviour

As a test, we dumped the annotations using the "YOLO 1.1" format directly and the "outside" rectangles were excluded. Dumping with "CVAT for Videos 1.1" provides an "outside" attribute to the resulting xml however the output format of the xml is incompatible with the post-processing steps we developed using the previous CVAT version.

1. Create a new task
1. Draw new Rectangle as Track (doesn't matter if 2-points or 4-points)  
1. Mark the rectangle at a random frame as "outside" 
1. Save
1. Dump annotations as "CVAT for Images 1.1"

### Context

I can't be sure if it a bug or desired behaviour introduced in the new version, however the usability of "hidden" from version 0.6.1 to hide object instances when annotating with interpolation mode cannot be reproduced neither with "hidden" nor with "outside".

### Your Environment

not relevant

### Next steps
You may [join our Gitter](https://gitter.im/opencv-cvat/public) channel for community support.

zhiltsov-max:
### Motivation and context

Fixes the problem with `CVAT for images` export - first track frames marked `outside` were presented in the output files.
Closes #2339 

### How has this been tested?

Unit test, manual test.

### Checklist

### License

```python
#
```

### DIFF
--- a/cvat/apps/dataset_manager/bindings.py
+++ b/cvat/apps/dataset_manager/bindings.py
@@ -268,6 +268,8 @@ class TaskData:
         for shape in sorted(anno_manager.to_shapes(self._db_task.data.size),
                 key=lambda shape: shape.get("z_order", 0)):
             if 'track_id' in shape:
+                if shape['outside']:
+                    continue
                 exported_shape = self._export_tracked_shape(shape)
             else:
                 exported_shape = self._export_labeled_shape(shape)
@@ -449,10 +451,9 @@ class TaskData:
         return None
 
 class CvatTaskDataExtractor(datumaro.SourceExtractor):
-    def __init__(self, task_data, include_images=False, include_outside=False):
+    def __init__(self, task_data, include_images=False):
         super().__init__()
         self._categories = self._load_categories(task_data)
-        self._include_outside = include_outside
 
         dm_items = []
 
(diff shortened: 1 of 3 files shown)


azhavoro (approved):

cvat/apps/dataset_manager/bindings.py
```diff
@@ -268,6 +268,8 @@ def get_frame(idx):
         for shape in sorted(anno_manager.to_shapes(self._db_task.data.size),
                 key=lambda shape: shape.get("z_order", 0)):
             if 'track_id' in shape:
+                if shape['outside']:
```
nmanovic, line 271: @zhiltsov-max , will the code work correctly for the case when a user wants to dump annotations in CVAT for video format? Because the format has to include outside frames with explicit `outside=1`. Could you please remind me how the case will be handled?

zhiltsov-max: This function is not used in that format.

nmanovic (approved):

PR #2369, opened 2020-10-28
Cypress test for check merge/split features.

dvkruchinin:
Cypress test for check merge/split features.

Reworking cypress command "checkObjectParameters". The logic is based on getting the ".cvat_canvas_shape " class and getting the ID number "#cvat_canvas_shape_" based on it. Next, use this value to check the parameters for "#cvat-objects-sidebar-state-item -". This is necessary for the function to work correctly if objects are created on different frames.
Try to fix the error "Cypress failed to make a connection to the Chrome DevTools Protocol". Which is often found in CI.
Moving selection label name when going to AAM to cypress command.
Fix some test for avid fail when the label order is reversed.

### Motivation and context
Reproduce a test scenario using Cypress.

### Checklist

### License

```python
#
```


tests/cypress/support/commands.js
```diff
@@ -217,10 +217,17 @@ Cypress.Commands.add('closeSettings', () => {
         });
 });
 
-Cypress.Commands.add('changeAnnotationMode', (mode) => {
+Cypress.Commands.add('changeAnnotationMode', (mode, labelName) => {
```
bsekachev, line 220: I would suggest renaming to ``changeWorkspace``.

tests/cypress/support/commands.js
```diff
@@ -217,10 +217,17 @@ Cypress.Commands.add('closeSettings', () => {
...
     cy.get('.cvat-workspace-selector').click();
     cy.get('.ant-select-dropdown-menu-item').contains(mode).click();
     cy.get('.cvat-workspace-selector').should('contain.text', mode);
+    if (mode === 'Attribute annotation') {
+        // Select the necessary label in any case
+        cy.get('.attribute-annotation-sidebar-basics-editor').within(() => {
+            cy.get('.ant-select-selection').click();
+        });
+        cy.get('.ant-select-dropdown-menu-item').contains(labelName).click();
+    }
```
bsekachev, line 230: Let it me a separated command. Like: ``changeLabel`` where mode can be detected automatically and corresponding elements are used to label changing. This command doesn't look like a right place to this ``if``

tests/cypress/support/commands.js
```diff
@@ -124,10 +122,12 @@ Cypress.Commands.add('switchLabel', (labelName) => {
 });
 
 Cypress.Commands.add('checkObjectParameters', (objectParameters, objectType) => {
-    cy.get('.cvat-objects-sidebar-state-item').then((objectSidebar) => {
-        cy.get(`#cvat_canvas_shape_${objectSidebar.length}`).should('exist').and('be.visible');
-        cy.get(`#cvat-objects-sidebar-state-item-${objectSidebar.length}`)
-        .should('contain', objectSidebar.length).and('contain', `${objectType} ${objectParameters.type.toUpperCase()}`).within(() => {
+    cy.get('.cvat_canvas_shape').last().should('have.attr', 'id').then(($cvatCanvasShapeId) => {
```
bsekachev, line 125: If we use ``last()`` I am not sure it will work always because there exist different orderings. We can test them and this logic will be broken. To get exactly latest element we might choose an object with the highest ID. Also getting the latest element id could be a different cypress command.

tests/cypress/support/commands.js
```diff
@@ -124,10 +122,12 @@ Cypress.Commands.add('switchLabel', (labelName) => {
...
+        const arrCvatCanvasShapeId = $cvatCanvasShapeId.split('_')
+        const idNumCanvasShape = arrCvatCanvasShapeId[arrCvatCanvasShapeId.length - 1]
```
bsekachev, line 127: id = +$cvatCanvasShapeId.match(/\d+$/);

tests/cypress/integration/actions_tasks_objects/case_13_merge_split_features.js
```diff
@@ -0,0 +1,132 @@
...
+    describe(`Testing case "${caseId}"`, () => {
+        it('Create rectangle shape on first frame', () => {
+            checkFrameNumber(frameNum);
+            cy.createRectangle(createRectangleShape2Points);
+            cy.get('#cvat_canvas_shape_1').should('have.attr', 'x').then(xCoords => {
+                xCoordinatesObjectFirstFrame = Math.floor(xCoords);
+            });
+        });
+        it('Create rectangle shape on third frame with another position', () => {
+            cy.get('.cvat-player-next-button').click().click();
```
bsekachev, line 53: Just to the future. You can input specify a frame in ``.cvat-player-frame-selector`` and to not click these buttons several times.

tests/cypress/support/commands.js
dvkruchinin: Renamed.

tests/cypress/support/commands.js
dvkruchinin: Made it a separate command. But I don't think you can completely avoid this condition. Cypress will try to search for the change label field where it doesn't exist. What will result in an error

tests/cypress/support/commands.js
dvkruchinin: Changed to use the maximum value from ```#cvat_canvas_shape_```

tests/cypress/support/commands.js
dvkruchinin: Thanks. Use this.

tests/cypress/integration/actions_tasks_objects/case_13_merge_split_features.js
dvkruchinin: Redesigned the function to enter the required frame number.

bsekachev (approved):

Issue #204, opened 2018-11-21
Any way to avoid copying and compressing files when creating new task?


Paperone80:
Hi,

Is there any way to avoid copying and compressing images when a new task is created with Source: Shared? Also, source is bound read-only so it shouldn't overwrite anything.
Not sure what the reason is but I have some high resolution imagery and I am loosing the important details to set proper attributes and polygon masks. Also, it takes up unnecessary time and disk space. Thanks.

I am using cvat github version from 2018-11-19.

gzvulon:
Same here, I have a lot of HD images on net share and s3,
I'd like to map them to /share dir and use only links, without any data copy.

vfdev-5:
any updates on this ?

nmanovic:
@vfdev-5 , we are going to reimplement our way to serve data from server (https://github.com/opencv/cvat/tree/az/video_stream). I hope to see the functionality merged in a month or so. After it is merged we will think how to implement the feature. Probably it will be possible to provide data in a pre-defined format. I hope to see the feature in v1.0.0. But we cannot promise.

nmanovic:
Some notes for future reference.

Pipeline to use original data:
1. Prepare data in a format which CVAT can understand and put them onto your remote storage:
     * directory with images
     * Use CVAT script (will be provided) to prepare data in right format (chunks)
     * Use CVAT script (will be provided) to prepare data for "protected" access (e.g. S3 with credentials). Thus instead of original data the user will upload some text files with links on original data and meta information about data.
2. CVAT will use the remote storage as is and don't try to copy files internally. There are three main use cases:
    * For `directory with images` CVAT will convert them into "own format" on the fly and cache using DiskCache (http://www.grantjenks.com/docs/diskcache/) in a temporary directory (thus next access should be fast and storage size will be limited).
    * Serve original data as is if data was prepared using CVAT script. Aka remote links on the required data. The user already prepared original data for us, don't need to do anything else. But for compressed data we need to compress them on the fly and cache using DiskCache (http://www.grantjenks.com/docs/diskcache/) in a temporary directory (thus next access should be fast and storage size will be limited). Because the original data was prepared as small chunks it will be fast enough to prepare compressed data.
3. cvat-data will be responsible to accept prepared data and transform them to actual images with meta information.
    * chunkN.zip with images will be unzipped
    * chunkN.mp4 with video frames will be decoded
    * chunkN.txt with links to data will be converted to images if credentials are provided by the client.

[...]

PR #2377, opened 2020-10-30
Share without copying & mount cloud storages

Marishka17:
### Motivation and context
Related issue: #204

This PR will allow not copy files inside CVAT when creating a task using share files.
Also PR contains documentation about mounting cloud storage(e.g AWS S3, Azure container, Google Drive) as FUSE. 

### How has this been tested?

Manually, tests
### Checklist
- [x] fix uploaded archive/zip
- [x] check automount using fstab
- [x] check automount using systemd

### Checklist

### License

```python
#
```


nmanovic:
@Marishka17 , please check the following use case:
- Mount a local directory to CVAT server with photos
- Try to classify images. For example, if the directory has your family photos, you can have labels `me`, `mother`, `father`, `brother`, etc.

Expected results:
- Task should be created mostly immediately
- Navigation by the task should be without significant delays
- You can annotate these photos with the annotation speed more than 1000 tags/hour. 
- Photos are not copied into the CVAT container (one copy of data)
-

cvat-ui/src/components/create-task-page/advanced-configuration-form.tsx
```diff
@@ -26,11 +26,14 @@ export interface AdvancedConfiguration {
     useZipChunks: boolean;
     dataChunkSize?: number;
     useCache: boolean;
+    activeTab: string;
```
bsekachev, line 29: activeTab shouldn't be a part of configuration. It is a prop.

cvat-ui/src/components/create-task-page/advanced-configuration-form.tsx
```diff
@@ -26,11 +26,14 @@ export interface AdvancedConfiguration {
...
+    copyData?: boolean;
 }
 
 type Props = FormComponentProps & {
     onSubmit(values: AdvancedConfiguration): void;
     installedGit: boolean;
+    activeTab: string;
```
bsekachev, line 36: let's rename to ``activeFileManagerTab``

cvat-ui/src/components/create-task-page/create-task-content.tsx
```diff
@@ -42,6 +42,7 @@ const defaultState = {
         lfs: false,
         useZipChunks: true,
         useCache: true,
+        activeTab: 'local',
```
bsekachev, line 45: Let's push it out of the nested advanced object

cvat-ui/src/components/create-task-page/advanced-configuration-form.tsx
```diff
@@ -100,6 +103,7 @@ class AdvancedConfigurationForm extends React.PureComponent<Props> {
                     onSubmit({
                         ...values,
                         frameFilter: values.frameStep ? `step=${values.frameStep}` : undefined,
+                        activeTab: activeTab,
```
bsekachev, line 106: Why do you need it in advanced configuration?

cvat/apps/engine/models.py
```diff
@@ -54,6 +54,18 @@ def choices(cls):
     def __str__(self):
         return self.value
 
+class UploadedDataStorageLocationChoice(str, Enum):
```
nmanovic, line 57: I will vote for a shorter name like StorageChoice or something like that.

bsekachev (dismissed):

bsekachev:
@Marishka17  I've just tested the PR in local development.
I have a share directory with some files. I selected on of files and set copy_data to True in advanced dialog. 
After that I created the task, went to share directory, renamed the file and everything still works well. Shoudn't this PR allow to avoid copying? 
Also in ``data/data/<id>/raw`` I see the file, so, it was copied. Am I wrong with understanding something?

cvat/apps/engine/models.py
Marishka17: Renamed

Marishka17:
@bsekachev, by default files are not copied to `../data/id/raw/` and customer works with files from share. If customer selects the `copy data` checkbox, then files is copied to `../data/id/raw/` and the he works with files inside CVAT.
Is there something wrong?

bsekachev:
> @bsekachev, by default files are not copied to `../data/id/raw/` and customer works with files from share. If customer selects the `copy data` checkbox, then files is copied to `../data/id/raw/` and he works with files inside CVAT.
> Is there something wrong?

Oh, sure, got it.

bsekachev:
@Marishka17 
One enhancement I could suggest is when we have a task with data on the share and file is not available on the share anymore (I just renamed a file to emulate it), we have infinite loading of an annotation view. No any errors occur client side, client just receives empty zip archive. I believe it would be better if UI says that the file disappear. 

It could be a different PR I suppose.

nmanovic:
> @Marishka17
> One enhancement I could suggest is when we have a task with data on the share and file is not available on the share anymore (I just renamed a file to emulate it), we have infinite loading of an annotation view. No any errors occur client side, client just receives empty zip archive. I believe it would be better if UI says that the file disappear.
> 
> It could be a different PR I suppose.

I think we need to address it in the PR. It is a bug and could lead to a lot of issues which will difficult to investigate.

nmanovic:
@azhavoro , could you please look at the PR and provide your decision?

cvat-core/src/server-proxy.js
```diff
@@ -641,6 +641,8 @@
                         },
                     );
                 } catch (errorData) {
+                    errorData.message = "";
+                    errorData.response.data = String.fromCharCode.apply(null, new Uint8Array(errorData.response.data));
```
bsekachev, line 645: Please, try something like this:

```js
throw generateError({
    ...errorData,
    message: '',
    response: {...errorData.response, data: String.... },
})
```

For strings use ``'`` instead of ``"``
```

azhavoro:
@Marishka17  PR is ok for me, please fix the conflict with `views.py`

azhavoro (approved):

bsekachev (approved):

iraadit:
Is the "Share without copying" working with the cli?

Marishka17:
@iraadit, there is currently no way to use this feature using the cli. Could you please create the issue?

iraadit:
> @iraadit, there is currently no way to use this feature using the cli. Could you please create the issue?

Hi, I created the new issue #2862

Thank you

PR #2475, opened 2020-11-23
Cypress test. Filters features.

dvkruchinin:
### Motivation and context
Cypress test. Filters features.
I also slightly modified 2 tests in which elements are selected by label name on the labels tab. Made a selection using a regular expression.

### Checklist

### License

```python
#
```

### DIFF
--- a/tests/cypress/integration/actions_tasks_objects/case_18_filters_functionality.js
+++ b/tests/cypress/integration/actions_tasks_objects/case_18_filters_functionality.js
@@ -8,44 +8,14 @@ import { taskName } from '../../support/const';
 
 context('Filters functionality.', () => {
     const caseId = '18';
-    const labelPolygonShape = 'polygon shape';
-    const additionalAttrsLabelPolygonShape = [
+    const labelShape = 'shape 3 points';
+    const additionalAttrsLabelShape = [
         { additionalAttrName: 'type', additionalValue: 'shape', typeAttribute: 'Text' },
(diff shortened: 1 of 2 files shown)


ActiveChooN:
@dvkruchinin. Works fine, but test has slightly long duration about 250 seconds. Can we reduce it with creating task via API?

dvkruchinin:
@ActiveChooN 
> test has slightly long duration about 250 seconds. Can we reduce it with creating task via API?

In CI, the annotation task already exists when the test is run. The test does not create it again, but only adds labels (In CI, the test time is about 50 seconds.). I also think it would be wrong to use the API in this tests since they are focused on testing the UI.
Alternatively, I can reduce the number of labels added and the number of objects created.

ActiveChooN:
@dvkruchinin,

> I also think it would be wrong to use the API in this tests since they are focused on testing the UI.

Agree with you, can we use raw label editor in that case?

dvkruchinin:
@ActiveChooN 
> can we use raw label editor in that case?

I think yes. Will check.

dvkruchinin:
@ActiveChooN
 
> > can we use raw label editor in that case?
> 
> I think yes. Will check.

I checked this point. It turns out the same time, because Cypress is able to enter strings. Accordingly, it enters all json as a single string. I will reduce the number of added labels.

ActiveChooN (approved):
LGTM,
@bsekachev, can you please merge this PR?

PR #2504, opened 2020-11-30
Task view hotfix

azhavoro:
Fixed the tasks view (infinite loading) when it is impossible to get a preview of the task

### How has this been tested?
Manually

### Checklist

### License

~~- [] I have updated the license header for each file (see an example below)~~

```python
#
```

### DIFF
--- a/cvat-ui/src/actions/tasks-actions.ts
+++ b/cvat-ui/src/actions/tasks-actions.ts
@@ -93,7 +93,7 @@ export function getTasksAsync(query: TasksQuery): ThunkAction<Promise<void>, {},
         }
 
         const array = Array.from(result);
-        const promises = array.map((task): string => (task as any).frames.preview().catch(''));
+        const promises = array.map((task): string => (task as any).frames.preview().catch(() => ''));
 
         dispatch(getInferenceStatusAsync());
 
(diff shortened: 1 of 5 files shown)


cvat-ui/src/actions/tasks-actions.ts
```diff
@@ -93,7 +93,7 @@ export function getTasksAsync(query: TasksQuery): ThunkAction<Promise<void>, {},
         }
 
         const array = Array.from(result);
-        const promises = array.map((task): string => (task as any).frames.preview().catch(''));
+        const promises = array.map((task): string => (task as any).frames.preview().catch(() => ''));
```
ActiveChooN, line 96: If we need only to ignore raised errors we can use lambda with empty body like this. Guess, it looks more laconic.
```suggestion
        const promises = array.map((task): string => (task as any).frames.preview().catch(() => {}));
```

azhavoro: > If we need only to ignore raised errors we can use lambda with empty body like this. Guess, it looks more laconic.

I agree, but `getTasksSuccess` function expects string[] (it should be base64-encoded images) as the second argument:
```ts
 getTasksSuccess(array: any[], previews: string[], count: number, query: TasksQuery)

ActiveChooN: Sure, agree with you

ActiveChooN (approved):

bsekachev (approved):

PR #2531, opened 2020-12-04
Logstash+Kibana config flexibility improvements

PMazarovich:
### Motivation and context

It is important to be able to connect to remote instances of Logstash or Kibana. With this changes you can simply configure any address (with or without credentials) via environment variables.
As example you can use this with AWS: for Logstash you can use Fargate, for the rest - Amazon Elasticsearch, which provide access to Kibana by URL.

### How has this been tested?
It was tested manually.

### Checklist

- [ x] I submit my changes into the `develop` branch

```python
#
```


azhavoro:
@PMazarovich, thanks for the your contribution!

docker-compose.yml
```diff
@@ -63,6 +63,8 @@ services:
     container_name: cvat_ui
     image: cvat/ui
     restart: always
+    environment:
+        CVAT_KIBANA_URL: http://localhost:8080/analytics/app/kibana
```
azhavoro, line 67: If the `cvat_proxy` service does not use the default CVAT_HOST or port, then Kibana will not be available at http://localhost:8080/analytics/app/kibana, so this setting must be changed consistently. From my point of view, this is not good.
Inside the cvat backend uses a reverse proxy for proxy requests in kibana, I would suggest making this django app customizable and pass CVAT_KIBANA_URL env variable into cvat instead of the current implementation. Do you see any problems with this approach?

PMazarovich: We do not use "default" `cvat_proxy` service (also `cvat_logstash`, `cvat_elasticsearch` and `cvat_kibana`), instead all these services from AWS or using Fargate, and accessed via Application Load Balancer. So it is necessary for us to have this cvat_kibana_url inside cvat_ui container, not in cvat one.
Or did I misunderstand something and there is some additional reverse proxy logic there?

azhavoro: @PMazarovich sorry for the late response
I mean your changes breaks the current approach to setting the cvat hostname with a single env variable (CVAT_HOST). For example, if I want to use anything other than `localhost` as the cvat hostname with the default ELK, with your changes, I have to set 2 environment variables accordingly: CVAT_HOST and CVAT_KIBANA_URL.
I just would suggest proxy all requests to kibana independently of kibana URL as it implemented now.
Also one more issue: regardless on CVAT_KIBANA_URL setting UI display the Analytics tab as origin URL
image

PMazarovich: In our case it is impossible to leave only CVAT_HOST variable because AWS kibana is accessible through this path: /_plugin/kibana/, not this one: /analytics/app/kibana. That is why I suggest to leave approach from the PR. Of course, you can change this in header.tsx, but, as for me, it is better to create another ENV rather than changing code. As for your notice about href, I'll perform another commit this week, that solves the problem.

bsekachev: Agree with Andrey, proposed changes look complicated. IMHO a client should work only with its server and if analytics located somewhere else, the server is expected to proxy request where need. Other approaches over complicate the architecture and configuration.

PMazarovich:
Changes with corrections are performed taking into account the wishes of developers and minimized in general.

azhavoro:
@PMazarovich  LGTM, could you please add a note to changelog.md about this change? Thanks!

PMazarovich:
Changelog updated.

azhavoro (approved):

PR #2534, opened 2020-12-07
CVAT-3D Milestone1

manasars:
CVAT-3D : Create a 3D task - Support to upload PCD/BIN files in archived zip.

Changes include:
Added Bin as accepted MIME type under Image Set
Added Dimension Validation Class to verify PCD vs PhotoCD and BIN to PCD conversion.
Added Support for Point Cloud and BIN Files for 3D Tasks.
Added an extra attribute dimension across API's and Formats.
Modified MPEG and Zip Chunk Writer to support PCD

Test Cases:
Manual Unit testing done locally.
Existing test cases work as expected.

- [x ] I submit my changes into the `develop` branch
- [ x] I have added tests to cover my changes

- [ x] I submit _my code changes_ under the same [MIT License](


cvat/apps/engine/media_extractors.py
```diff
@@ -179,25 +194,48 @@ def _make_name():
...
         file_list = [f for f in self._zip_source.namelist() if get_mime(f) == 'image']
         super().__init__(file_list, step, start, stop)
 
     def __del__(self):
         self._zip_source.close()
 
-    def get_preview(self):
+    def get_preview(self, dimension="2d"):
+        if dimension == "3d":
+            fp = open(os.path.join(os.path.dirname(__file__), 'assets/3d_preview.jpeg'), "rb")
```
zhiltsov-max, line 207: I believe the source of 3d preview is to be changed in the future to a real preview.

cvat/apps/engine/media_extractors.py
```diff
@@ -326,17 +365,21 @@ class ZipCompressedChunkWriter(IChunkWriter):
...
         image_sizes = []
         with zipfile.ZipFile(chunk_path, 'x') as zip_chunk:
-            for idx, (image, _ , _) in enumerate(images):
-                w, h, image_buf = self._compress_image(image, self._image_quality)
+            for idx, (image, _, _) in enumerate(images):
+                if self._dimension == "2d":
+                    [w, h, image_buf], extension = self._compress_image(image, self._image_quality), "jpeg"
+                else:
+                    properties = get_pcd_properties(image)
+                    w, h, image_buf, extension  = int(properties["WIDTH"]), int(properties["HEIGHT"]), image, "pcd"
```
zhiltsov-max, line 373: Let's split the assignments into few lines.

cvat/apps/engine/media_extractors.py
```diff
@@ -326,17 +365,21 @@ class ZipCompressedChunkWriter(IChunkWriter):
...
+                arcname = '{:06d}.{}'.format(idx, extension)
                 zip_chunk.writestr(arcname, image_buf.getvalue())
 
         return image_sizes
 
 class Mpeg4ChunkWriter(IChunkWriter):
-    def __init__(self, _):
-        super().__init__(17)
+    def __init__(self, _, dimension="2d"):
+        super().__init__(17, dimension)
```
zhiltsov-max, line 382: I suggest passing `dimension` as a named parameter.

cvat/apps/engine/media_extractors.py
```diff
@@ -510,3 +553,190 @@ def _is_zip(path):
...
+        self.converted_files.append(pcd_path)
+        return pcd_path.split(actual_path)[-1][1:]
+
+    def pcd_operation(self, file_path, actual_path):
+        if self.validate_pcd(file_path):
+            return file_path.split(actual_path)[-1][1:]
+        else:
+            return file_path
+
+    def validate(self):
```
zhiltsov-max, line 616: Try to split the function into smaller parts, probably using local functions, if possible.

zhiltsov-max (commented):
Changes regarding formats look ok. The only thing I can suggest is adding some kind of enumeration for available values for `dimension` args.

cvat/apps/dataset_manager/serializers.py
```diff
@@ -10,6 +10,7 @@ class DatasetFormatSerializer(serializers.Serializer):
     ext = serializers.CharField(max_length=64, source='EXT')
     version = serializers.CharField(max_length=64, source='VERSION')
     enabled = serializers.BooleanField(source='ENABLED')
+    dimension = serializers.CharField(max_length=64, source='DIMENSION')
```
nmanovic, line 13: Why do you need 64 chars here? It is better to have choice here between 2D and 3D.

cvat/apps/dataset_manager/formats/registry.py
```diff
@@ -45,29 +45,30 @@ def __call__(self, *args, **kwargs):
...
+            enabled=enabled, dimension=dimension)
         key = t.DISPLAY_NAME
         assert key not in EXPORT_FORMATS, "Export format '%s' already registered" % name
         EXPORT_FORMATS[key] = t
         return t
     return wrap_with_params
 
 IMPORT_FORMATS = {}
-def importer(name, version, ext, display_name=None, enabled=True):
+def importer(name, version, ext, display_name=None, enabled=True, dimension="2d"):
```
nmanovic, line 67: I believe `2D` and `3D` should be enum values. Look at similar constants and how they are defined in server's code.

cvat/apps/engine/media_extractors.py
nmanovic: I agree. I will vote to leave it unimplemented or generate a real preview during data preparation.

cvat/apps/engine/media.mimetypes
```diff
@@ -107,6 +107,7 @@ image/jp2	jp2
 image/x-portable-bitmap	pbm
 image/x-eps	epsf
 image/x-photo-cd	pcd
+image/bin	bin
```
nmanovic, line 110: Probably we need two media types in accordance with https://en.wikipedia.org/wiki/Media_type
image/x.kitti-velodyne bin
image/x.point-cloud-data pcd

cvat/apps/engine/media_extractors.py
```diff
@@ -179,25 +194,48 @@ def _make_name():
 
 class ZipReader(ImageListReader):
     def __init__(self, source_path, step=1, start=0, stop=None):
-        self._zip_source = zipfile.ZipFile(source_path[0], mode='r')
+        self._zip_source = zipfile.ZipFile(source_path[0], mode='a')
```
nmanovic, line 197: What is a reason to change mode here?

cvat/apps/engine/media_extractors.py
```diff
@@ -29,6 +32,18 @@ def get_mime(name):
 
     return 'unknown'
 
+def get_pcd_properties(file):
+        kv = {}
+        for line_no, line in enumerate(file):
```
nmanovic, line 37: why do you need line_no? I will propose to move parsing of PCD files into a separate py module.

cvat/apps/engine/media_extractors.py
```diff
@@ -510,3 +553,190 @@ def _is_zip(path):
...
+        self.converted_files = []
+
+    @staticmethod
+    def validate_pcd(path):
+        """
+            Validate the point cloud file along with version check
+            On failure it is considered as PhotoCD format
+        """
+
+        pcd_version = ["0.7", "0.6", "0.5", "0.4", "0.3", "0.2", "0.1", ".7", ".6", ".5", ".4", ".3", ".2", ".1"]
```
nmanovic, line 574: It should be a feature of a module which parses PCD files. I don't like the idea to spread the code over the file.

cvat/apps/engine/media_extractors.py
```diff
@@ -510,3 +553,190 @@ def _is_zip(path):
...
+        with open(path, "rb") as file_read:
+            data = file_read.read(70)
```
nmanovic, line 576: I saw code to parse the header above. It is better to reuse it.

cvat/apps/engine/models.py
```diff
@@ -202,6 +202,7 @@ class Task(models.Model):
     status = models.CharField(max_length=32, choices=StatusChoice.choices(),
         default=StatusChoice.ANNOTATION)
     data = models.ForeignKey(Data, on_delete=models.CASCADE, null=True, related_name="tasks")
+    dimension = models.CharField(max_length=2, default="2d")
```
nmanovic, line 205: the serializer has max_length=64 parameter. Probably it should use choices parameter.

cvat/apps/engine/models.py
```diff
@@ -265,6 +266,17 @@ class RemoteFile(models.Model):
     class Meta:
         default_permissions = ()
 
+
+class RelatedFile(models.Model):
```
nmanovic, line 270: Why do we need the db model? I expect that 3d data will be handled  the same way as regular images. Probably I miss something.

cvat/apps/engine/media_extractors.py
```diff
@@ -510,3 +553,190 @@ def _is_zip(path):
...
+            return file_path
+
+    def validate(self):
+        """
+            Validate the directory structure for kitty and point cloud format.
+        """
+        if not self.path:
+            return
+        actual_path = self.path
+        for root, dirs, files in os.walk(actual_path):
```
nmanovic, line 623: @zhiltsov-max , should the code (import velodyne dataset) migrate to datumaro? Or we prefer to keep it here for a while.

cvat-core/src/api-implementation.js
```diff
@@ -176,6 +176,7 @@
                 search: isString,
                 status: isEnum.bind(TaskStatus),
                 mode: isEnum.bind(TaskMode),
+                dimension: isString
```
vnishukov, line 179: In case of dimension has to be only 2D and 3D, so it might be better solution to make an enum for it

cvat-ui/src/components/actions-menu/actions-menu.tsx
```diff
@@ -21,7 +21,7 @@ interface Props {
     dumpActivities: string[] | null;
     exportActivities: string[] | null;
     inferenceIsActive: boolean;
-
+    taskDimension: string;
```
vnishukov, line 24: Same here (enum)

cvat-ui/src/components/actions-menu/dump-submenu.tsx
```diff
@@ -19,15 +19,17 @@ interface Props {
     menuKey: string;
     dumpers: any[];
     dumpActivities: string[] | null;
+    taskDimension: string;
```
vnishukov, line 22: Enum

cvat-ui/src/components/actions-menu/load-submenu.tsx
```diff
@@ -14,15 +14,17 @@ interface Props {
     loaders: any[];
     loadActivity: string | null;
     onFileUpload(file: File): void;
+    taskDimension: string;
```
vnishukov, line 17: Enum

cvat/apps/engine/media_extractors.py
vnishukov: In SoW we have chapter **1.2.2. 3D-tasks must contain a thumbnail of the lidar data and a “3D” label** so the content of the thumbnail must be a frame from real data of particular PCD or BIN file

cvat/apps/engine/media_extractors.py
```diff
@@ -179,25 +194,48 @@ def _make_name():
...
+        self._zip_source = zipfile.ZipFile(source_path[0], mode='a')
         self.extract_dir = source_path[1] if len(source_path) > 1 else None
         file_list = [f for f in self._zip_source.namelist() if get_mime(f) == 'image']
         super().__init__(file_list, step, start, stop)
 
     def __del__(self):
         self._zip_source.close()
 
-    def get_preview(self):
+    def get_preview(self, dimension="2d"):
```
azhavoro, line 205: From my point of view it would be better to inherit a new class to support 3d data. In this case it's not needed to add dummy arguments (I mean `dimension`) to the other readers that will never support them.
You can determinate the proper reader here https://github.com/openvinotoolkit/cvat/blob/0fd3742c5a50d55e09258a5dc241ce91d72c3bc4/cvat/apps/engine/task.py#L109 by using your `ValidateDimention` class.

cvat/apps/engine/media_extractors.py
```diff
@@ -179,25 +194,48 @@ def _make_name():
...
 
+    def add_files(self, source_path):
+        root_path = os.path.split(self._zip_source.filename)[0]
+        for path in source_path:
+            self._zip_source.write(path, path.replace(root_path, ""))
+
+    def get_zip_filename(self):
+        return self._zip_source.filename
+
+    def initialize_for_3d(self, source_path, step=1, start=0, stop=None):
```
azhavoro, line 231: If you inherit a new class, this method is no longer needed

cvat-ui/src/components/actions-menu/dump-submenu.tsx
```diff
@@ -19,15 +19,17 @@ interface Props {
...
 
 export default function DumpSubmenu(props: Props): JSX.Element {
-    const { taskMode, menuKey, dumpers, dumpActivities } = props;
+    const { taskMode, menuKey, dumpers, dumpActivities, taskDimension } = props;
 
     return (
         <Menu.SubMenu key={menuKey} title='Dump annotations'>
             {dumpers
                 .sort((a: any, b: any) => a.name.localeCompare(b.name))
+                .filter((dumper: any): JSX.Element => dumper.dimension === taskDimension)
```
bsekachev, line 32: filter callback returns boolean, not JSX.Element

cvat-ui/src/components/actions-menu/export-submenu.tsx
```diff
@@ -11,15 +11,17 @@ interface Props {
...
 
 export default function ExportSubmenu(props: Props): JSX.Element {
-    const { menuKey, exporters, exportActivities } = props;
+    const { menuKey, exporters, exportActivities, taskDimension } = props;
 
     return (
         <Menu.SubMenu key={menuKey} title='Export as a dataset'>
             {exporters
                 .sort((a: any, b: any) => a.name.localeCompare(b.name))
+                .filter((exporter: any): JSX.Element => exporter.dimension === taskDimension)
```
bsekachev, line 24: filter callback returns boolean, not JSX.Element

cvat-ui/src/components/actions-menu/load-submenu.tsx
```diff
@@ -14,15 +14,17 @@ interface Props {
...
 
 export default function LoadSubmenu(props: Props): JSX.Element {
-    const { menuKey, loaders, loadActivity, onFileUpload } = props;
+    const { menuKey, loaders, loadActivity, onFileUpload, taskDimension } = props;
 
     return (
         <Menu.SubMenu key={menuKey} title='Upload annotations'>
             {loaders
                 .sort((a: any, b: any) => a.name.localeCompare(b.name))
+                .filter((loader: any): JSX.Element => loader.dimension === taskDimension)
```
bsekachev, line 27: filter callback returns boolean, not JSX.Element

cvat-ui/src/components/model-runner-modal/detector-runner.tsx
```diff
@@ -118,7 +118,8 @@ function DetectorRunner(props: Props): JSX.Element {
                 <Col span={4}>Model:</Col>
                 <Col span={20}>
                     <Select
-                        placeholder='Select a model'
+                        placeholder={task.dimension ==='2d'?'Select a model' : "No model available"}
```
bsekachev, line 121: Please, add missed spaces. Doesn't precommit hook with automatic code linting work for you?

cvat-ui/src/components/model-runner-modal/detector-runner.tsx
```diff
@@ -118,7 +118,8 @@ function DetectorRunner(props: Props): JSX.Element {
...
+                        disabled = {task.dimension !== '2d'}
```
bsekachev, line 122: Please, remove extra spaces

bsekachev (changes requested):

cvat/apps/engine/models.py
bsekachev: I would suggest implement it via enums. Please, look how "status" field was implemented with choices argument.

bsekachev (changes requested):

cvat/apps/dataset_manager/serializers.py
manasars: Yes we will change it to have choices.

cvat/apps/engine/media_extractors.py
manasars: We add the converted (bin to PCD), into the in memory zip again, thus we require it to be "a" mode, against creating an entirely new zip

cvat/apps/engine/models.py
manasars: We have used the db model RelatedFile in-order to place the contextual image corresponding to the PCD files.
Also in future we can have support for multiple images mapped to a single PCD file.

cvat/apps/engine/media_extractors.py
manasars: We have left the changes as such, will change it to real preview in the next milestone.

cvat/apps/engine/media_extractors.py
manasars: Made changes accordingly.

cvat/apps/engine/media_extractors.py
manasars: Made changes accordingly.

cvat/apps/dataset_manager/serializers.py
manasars: Made changes accordingly.

cvat/apps/dataset_manager/formats/registry.py
manasars: Made changes accordingly.

cvat/apps/engine/media.mimetypes
manasars: We have added the two media types as suggested, but in the code there is an existing mime-type for pcd i.e 'image/x-photo-cd	pcd', kindly let us know if the current one will still be applicable.

cvat/apps/engine/media_extractors.py
manasars: We have included all the parsing of PCD/BIN files in the class ValidateDimension.

cvat/apps/engine/media_extractors.py
manasars: We have included all the parsing of PCD/BIN files in the class ValidateDimension.

cvat/apps/engine/models.py
manasars: Made changes accordingly.

cvat-core/src/api-implementation.js
manasars: Made changes accordingly.

cvat-ui/src/components/actions-menu/actions-menu.tsx
manasars: Since there are no redux actions performed on the taskDimension, we have used string here.
If we need to change to enum is there any common location where we can place and use it in whole of cvat-ui, kindly suggest.

cvat/apps/engine/media_extractors.py
manasars: Currently there is an existing mime-type for pcd i.e 'image/x-photo-cd pcd' in mime-types, hence we are distinguishing the 3D and 2D based on headers in the PCD file, if the existing mime-type can be deleted then we can implement in the similar manner mentioned.

cvat-ui/src/components/actions-menu/dump-submenu.tsx
manasars: Made changes accordingly.

cvat-ui/src/components/actions-menu/export-submenu.tsx
manasars: Made changes accordingly.

cvat-ui/src/components/actions-menu/load-submenu.tsx
manasars: Made changes accordingly.

cvat-ui/src/components/model-runner-modal/detector-runner.tsx
manasars: Made changes accordingly.

cvat-ui/src/components/actions-menu/actions-menu.tsx
vnishukov: For now you can place it to "**/cvat-ui/src/reducers/interfaces.ts**" like this, for example:
```
export enum TaskDimension {
    D2 = '2D',
    D3 = '3D',
}
```

right after, for example:
```
export enum TaskStatus {
    ANNOTATION = 'annotation',
    REVIEW = 'validation',
    COMPLETED = 'completed',
}
```

I would propose try to not use a plain strings, especially if it's some kind of constant options or some stuff like this

cvat-ui/src/components/actions-menu/dump-submenu.tsx
vnishukov: Corresponding above, something like this:
```
taskDimension: TaskDimension.D2 | TaskDimension.D3
```

vnishukov (changes requested):
Enum for cvat_ui

cvat/apps/engine/models.py
```diff
@@ -19,6 +19,17 @@ def get_prep_value(self, value):
             return value[:self.max_length]
         return value
 
+class DimensionType(str, Enum):
+    THREED = '3d'
```
nmanovic, line 23: Something like `DIM_2D` and `DIM_3D` is better.

nmanovic:
I will review the patch  tomorrow. Please resolve conflicts.

cvat-ui/src/components/actions-menu/actions-menu.tsx
manasars: Changed code accordingly.

cvat-ui/src/components/actions-menu/dump-submenu.tsx
manasars: Changed code accordingly.

cvat-ui/src/components/actions-menu/load-submenu.tsx
manasars: Changed code accordingly.

cvat/apps/engine/models.py
manasars: Changed code accordingly.

cvat-ui/src/reducers/interfaces.ts
```diff
@@ -569,3 +569,8 @@ export interface CombinedState {
     shortcuts: ShortcutsState;
     review: ReviewState;
 }
+
+export enum TaskDimension {
+    DIM_3D = '3d',
```
nmanovic, line 574: I see similar enums in cvat-core/src/enums.js. What is a reason to have them here as well?

nmanovic: Also I see that in python code you use DimensionType. Probably it is better to have similar name here as well instead of TaskDimension.

cvat/apps/engine/media_extractors.py
nmanovic: I agree. Instead of multiple "if" in the class, let us create a new one like LidarDataReader or something like that.

cvat/apps/engine/media_extractors.py
```diff
@@ -264,7 +291,7 @@ def _get_av_container(self):
             self._source_path[0].seek(0) # required for re-reading
         return av.open(self._source_path[0])
 
-    def get_preview(self):
+    def get_preview(self, dimension=DimensionType.DIM_2D):
```
nmanovic, line 294: If you have LidarDataReader, you will not need the extra argument. It isn't necessary here in VideoReader for sure.

cvat/apps/engine/migrations/0036_auto_20201216_0943.py
```diff
@@ -0,0 +1,33 @@
...
+
+    dependencies = [
+        ('engine', '0035_data_storage'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='task',
+            name='dimension',
+            field=models.CharField(choices=[('3d', 'DIM_3D'), ('2d', 'DIM_2D')], default=cvat.apps.engine.models.DimensionType['DIM_2D'], max_length=2),
```
nmanovic, line 18: from cvat.apps.engine.models import DimensionType

Use DimensionType.choices() method here.

nmanovic (commented):
Where did you get files for test_pointcloud_pcd.zip and test_velodyne_points.zip. Are they under MIT license? Any privacy or legal issues with data?

I will recommend generate synthetic data for testings. Thus you will be able to reduce size of archives and be sure that data is absolutely legal.

cvat/apps/engine/media_extractors.py
vnishukov: Let's get rid of the old one

cvat-ui/src/components/model-runner-modal/detector-runner.tsx
vnishukov: Also please do not use **--no-verify** flag on commit. The pre-commit hooks must be passed before push

cvat-core/src/enums.js
```diff
@@ -33,6 +33,20 @@
...
+    /**
+     * Task dimension
+     * @enum
+     * @name TaskDimension
+     * @memberof module:API.cvat.enums
+     * @property {string} DIMENSION_2D '2d'
+     * @property {string} DIMENSION_3D '3d'
+     * @readonly
+     */
+    const TaskDimension = Object.freeze({
```
vnishukov, line 45: Let's use unified name here **DimensionType** (like in python code)

cvat-ui/src/components/actions-menu/dump-submenu.tsx
vnishukov: Lets' use unified name here **DimensionType** (like in Python code)

cvat/apps/engine/media.mimetypes
vnishukov: Let's get rid of old one

cvat-ui/src/components/actions-menu/dump-submenu.tsx
vnishukov: @bsekachev can we use **cvat-core/src/enums.js** here? (it might be a hard link of 2 separate modules, I believe)

cvat/apps/engine/media_extractors.py
manasars: Made changes accordingly.

cvat/apps/engine/media.mimetypes
manasars: Thanks, Made changes accordingly.

cvat-ui/src/reducers/interfaces.ts
manasars: Made changes accordingly.

cvat/apps/engine/migrations/0036_auto_20201216_0943.py
manasars: This is an auto-generated file, do we have to modify to as requested, kindly confirm.

cvat-ui/src/components/model-runner-modal/detector-runner.tsx
manasars: We are not using the --no-verify flag on commit, but looks like the pre-commit hook is not giving us any error.

cvat-ui/src/reducers/interfaces.ts
manasars: Made changes accordingly.

cvat-core/src/enums.js
manasars: Made changes accordingly

cvat-ui/src/components/model-runner-modal/detector-runner.tsx
vnishukov: Ok, so just put the spaces by hand please:
**==='2d'?'Select a model'** -> **=== '2d' ? 'Select a model'**

cvat/apps/engine/media_extractors.py
manasars: Made changes accordingly

vnishukov (approved):

cvat-ui/src/components/model-runner-modal/detector-runner.tsx
manasars: The space was added as a part of earlier commit, so I do see a space currently in the file, kindly confirm

cvat/apps/engine/media_extractors.py
manasars: Image dataset and pcd dataset can be in a zip format, to classify if the dataset is for a 2D task or 3D task.
If we create another class as "LidarDataReader", we need to unzip the zip file twice, once for extracting the data in "ZipReader",then again in "LidarReader" class since zip is a common media-type.

https://github.com/openvinotoolkit/cvat/blob/0877de4885232cdfa7d6b8e3d9516a756df06f99/cvat/apps/engine/media_extractors.py#L545

Instead we modified the "ZipReader" class to and used reconcile method to distinguish between the dimensionType and now the multiple if's are not present.
It is only checked once on initialisation.

cvat-ui/src/components/model-runner-modal/detector-runner.tsx
vnishukov: I see. I have approved changes

nmanovic:
@dvkruchinin , could you please help us to investigate problems with tests in the PR? Please put here comments ASAP.

```
┌────────────────────────────────────────────────────────────────────────────────────────────────┐
9793  │ Tests:        6                                                                                │
9794  │ Passing:      0                                                                                │
9795  │ Failing:      2                                                                                │
9796  │ Pending:      0                                                                                │
9797  │ Skipped:      4                                                                                │
9798  │ Screenshots:  2                                                                                │
9799  │ Video:        false                                                                            │
9800  │ Duration:     56 seconds                                                                       │
9801  │ Spec Ran:     actions_tasks_objects/case_24_delete_unlock_lock_object.js                       │
9802  └────────────────────────────────────────────────────────────────────────────────────────────────┘
9803
9804
9805  (Screenshots)
9806
9807  -  /home/travis/build/openvinotoolkit/cvat/tests/cypress/screenshots/actions_tasks_     (1280x720)
9808     objects/case_24_delete_unlock_lock_object.js/Prepare to testing (failed).png                   
9809  -  /home/travis/build/openvinotoolkit/cvat/tests/cypress/screenshots/actions_tasks_     (1280x720)
9810     objects/case_24_delete_unlock_lock_object.js/Delete unlocklock object -- Create                
9811     and delete object via Delete shortcut -- before all hook (failed).png                          
9812
9813
9814────────────────────────────────────────────────────────────────────────────────────────────────────
9815                                                                                                    
9816  Running:  actions_tasks_objects/case_25_canvas_bitmap_feature.js                        (18 of 72)
9817
9818
9819  1) Prepare to testing
9820  Canvas bitmap feature
9821    2) "before all" hook for "Bitmap not visible."
9822
9823
9824  0 passing (57s)
9825  2 failing
9826
9827  1) Prepare to testing:
9828     AssertionError: Timed out retrying: expected 'http://localhost:8080/auth/login' to include '/tasks'
9829      at Context.eval (http://localhost:8080/__cypress/tests?p=cypress/support/index.js:678:12
```

nmanovic:
@manasars , please investigate problems with tests in parallel. All tests should be passed to merge a PR into our repo.

dvkruchinin:
> Please put here comments ASAP.

@nmanovic 
All tests are not successful in this PR due to an error when creating django superuser
```django.db.utils.ProgrammingError: relation "auth_user" does not exist```
```LINE 1: INSERT INTO "auth_user" ("password", "last_login", "is_super...```
As a result, it was not possible to log in to the application under the superuser.

**Comment by dvkruchinin** [comment]
In this case, an error occurred in the test issue_1568_cuboid_dump_annotation.js
Due to the lack of opening submenu for Menu - > Dump annotations.
How to reproduce:
1. Create a task
2. Go to menu
3. Try to move cursor to "Dump annotations" or "Upload annotations" or "Export as a dataset". Submenu should appear.
But it doesn't appear.

**Comment by manasars** [comment]
Hi @nmanovic,

We have generated synthetic data for pcd, bin and img test files.
Resolved build errors, 1 test case shows as pending and notified the same on mail.
Kindly review PR.

Thank You.

**Comment by nmanovic** [comment]
@manasars ,

Could you please inspect logs from travis-ci next time before ask me to review the patch again?

```
raceback (most recent call last):
  File "/home/django/utils/cli/tests/test_cli.py", line 41, in setUpClass
    super().setUpClass()
  File "/opt/venv/lib/python3.8/site-packages/django/test/testcases.py", line 1123, in setUpClass
    cls.setUpTestData()
  File "/home/django/utils/cli/tests/test_cli.py", line 54, in setUpTestData
    create_db_users(cls)
  File "/home/django/cvat/apps/engine/tests/test_rest_api.py", line 45, in create_db_users
    user_admin = User.objects.create_superuser(username="admin", email="",
  File "/opt/venv/lib/python3.8/site-packages/django/contrib/auth/models.py", line 157, in create_superuser
    return self._create_user(username, email, password, **extra_fields)
  File "/opt/venv/lib/python3.8/site-packages/django/contrib/auth/models.py", line 140, in _create_user
    user.save(using=self._db)
  File "/opt/venv/lib/python3.8/site-packages/django/contrib/auth/base_user.py", line 67, in save
    super().save(*args, **kwargs)
  File "/opt/venv/lib/python3.8/site-packages/django/db/models/base.py", line 753, in save
    self.save_base(using=using, force_insert=force_insert,
  File "/opt/venv/lib/python3.8/site-packages/django/db/models/base.py", line 790, in save_base
    updated = self._save_table(
  File "/opt/venv/lib/python3.8/site-packages/django/db/models/base.py", line 895, in _save_table
    results = self._do_insert(cls._base_manager, using, fields, returning_fields, raw)
  File "/opt/venv/lib/python3.8/site-packages/django/db/models/base.py", line 933, in _do_insert
    return manager._insert(
  File "/opt/venv/lib/python3.8/site-packages/django/db/models/manager.py", line 85, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/opt/venv/lib/python3.8/site-packages/django/db/models/query.py", line 1249, in _insert
    return query.get_compiler(using=using).execute_sql(returning_fields)
  File "/opt/venv/lib/python3.8/site-packages/django/db/models/sql/compiler.py", line 1397, in execute_sql
    cursor.execute(sql, params)
  File "/opt/venv/lib/python3.8/site-packages/cacheops/transaction.py", line 93, in execute
    result = self._no_monkey.execute(self, sql, params)
  File "/opt/venv/lib/python3.8/site-packages/django/db/backends/utils.py", line 66, in execute
    return self._execute_with_wrappers(sql, params, many=False, executor=self._execute)
  File "/opt/venv/lib/python3.8/site-packages/django/db/backends/utils.py", line 75, in _execute_with_wrappers
    return executor(sql, params, many, context)
  File "/opt/venv/lib/python3.8/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
  File "/opt/venv/lib/python3.8/site-packages/django/db/utils.py", line 90, in __exit__
    raise dj_exc_value.with_traceback(traceback) from exc_value
  File "/opt/venv/lib/python3.8/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
  File "/opt/venv/lib/python3.8/site-packages/django/db/backends/sqlite3/base.py", line 413, in execute
    return Database.Cursor.execute(self, query, params)
django.db.utils.IntegrityError: UNIQUE constraint failed: auth_user.username
----------------------------------------------------------------------
Ran 132 tests in 45.361s
FAILED (errors=30)
Destroying test database for alias 'default'...
The command "docker-compose -f docker-compose.yml -f docker-compose.ci.yml run cvat_ci /bin/bash -c 'coverage run -a manage.py test cvat/apps utils/cli && mv .coverage ${CONTAINER_COVERAGE_DATA_DIR}'" exited with 1.
```

File: Dockerfile [hunk 31]
```diff
@@ -73,6 +73,8 @@ RUN apt-get update && \
     DEBIAN_FRONTEND=noninteractive apt-get --no-install-recommends install -yq \
         apache2 \
         libapache2-mod-xsendfile \
+        libgomp1 \
+        libgl1-mesa-glx \
```
azhavoro, line 77: @manasars What is the full list of apt packages that will be installed along with these packages?

manasars: Hi @azhavoro ,
We have changed libgl1-mesa-glx to libgl1 now, and below are the dependent packages:

Package: libgomp1
Depends: gcc-10-base (= 10-20200411-0ubuntu1), libc6 (>= 2.17)
Breaks: gcc-4.3 (<< 4.3.6-1), gcc-4.4 (<< 4.4.6-4), gcc-4.5 (<< 4.5.3-2)

Package: libgl1
Depends: libc6 (>= 2.14), libglvnd0 (= 1.3.1-1), libglx0 (= 1.3.1-1)
Filename: pool/main/libg/libglvnd/libgl1_1.3.1-1_amd64.deb

manasars:
Hi @nmanovic ,

The ci/cd build has passed, latest code pushed to PR.

cvat-ui/src/reducers/interfaces.ts
```diff
@@ -569,3 +569,8 @@ export interface CombinedState {
     shortcuts: ShortcutsState;
     review: ReviewState;
 }
+
+export enum DimensionType {
```
nmanovic, line 573: @manasars , we have similar definition in cvat-core/src/enums.js. What is the reason for that?

cvat/apps/engine/media_extractors.py
```diff
@@ -179,7 +183,8 @@ def _make_name():
 
 class ZipReader(ImageListReader):
     def __init__(self, source_path, step=1, start=0, stop=None):
-        self._zip_source = zipfile.ZipFile(source_path[0], mode='r')
+        self._dimension = DimensionType.DIM_2D
```
nmanovic, line 186: Andrey and I asked you to reimplement the class. It is much better to create a class like LidarDataReader inherited from ZipReader and redefine all corresponding methods. Thus you will not need to have `if self._dimension == DimensionType.DIM_3D:` mostly in every method. Let me know if you disagree by a reason.

cvat/apps/engine/media_extractors.py
```diff
@@ -283,8 +312,9 @@ def get_image_size(self, i):
         return image.width, image.height
 
 class IChunkWriter(ABC):
-    def __init__(self, quality):
+    def __init__(self, quality, dimension=DimensionType.DIM_2D):
```
nmanovic, line 315: Again better to inherit here. For example, IChunkWriter for common code, IChunk2dWriter(IChunkWriter) and IChunk3dWriter(IChunkWriter) or something like that.

nmanovic:
@manasars , an extra level of directories + temporary files inside archives. Could you please fix that?

test_pointcloud_pcd.zip\test_pointcloud_pcd\test_pointcloud_pcd\related_images\000001_pcd

image

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -1840,6 +1844,47 @@ def setUpClass(cls):
...
+        zip_file = zipfile.ZipFile(path)
+        for info in zip_file.namelist():
+            if info.rsplit(".", maxsplit=1)[-1] == "pcd":
+                with zip_file.open(info, "r") as file:
+                    data = ValidateDimension.get_pcd_properties(file)
+                    image_sizes.append((int(data["WIDTH"]), int(data["HEIGHT"])))
+        cls._image_sizes[filename] = image_sizes
+
+        filename = "test_velodyne_points.zip"
+        path = os.path.join(os.path.dirname(__file__), 'assets', 'test_velodyne_points.zip')
```
nmanovic, line 1860:
```suggestion
        path = os.path.join(os.path.dirname(__file__), 'assets', filename)
```

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -1840,6 +1844,47 @@ def setUpClass(cls):
             zip_archive.write(data.read())
         cls._image_sizes[filename] = img_sizes
 
+        filename = "test_pointcloud_pcd.zip"
+        path = os.path.join(os.path.dirname(__file__), 'assets', 'test_pointcloud_pcd.zip')
```
nmanovic, line 1848:
```suggestion
        path = os.path.join(os.path.dirname(__file__), 'assets', filename)
```

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -1840,6 +1844,47 @@ def setUpClass(cls):
...
+        image_sizes = []
+        # create zip instance
+
+        zip_file = zipfile.ZipFile(path, mode='a')
+
+        source_path = []
+        root_path = os.path.abspath(os.path.split(path)[0])
+
+        for info in zip_file.namelist():
+            if info.rsplit(".", maxsplit=1)[-1] == "bin":
```
nmanovic, line 1870: probably it is better to use os.path.splitext here.

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -1840,6 +1844,47 @@ def setUpClass(cls):
...
+                zip_file.extract(info, root_path)
+                bin_path = os.path.abspath(os.path.join(root_path, info))
+                source_path.append(ValidateDimension.convert_bin_to_pcd(bin_path))
+
+        for path in source_path:
+            zip_file.write(path, os.path.abspath(path.replace(root_path, "")))
+
+        for info in zip_file.namelist():
+            if info.rsplit(".", maxsplit=1)[-1] == "pcd":
```
nmanovic, line 1879: os.path.splitext here?

nmanovic:
> @manasars , an extra level of directories + temporary files inside archives. Could you please fix that?
> 
> test_pointcloud_pcd.zip\test_pointcloud_pcd\test_pointcloud_pcd\related_images\000001_pcd
> 
> image

The same problem with another archive.

cvat/apps/engine/media_extractors.py
manasars: As mentioned in the comments earlier, We distinguish the mime types only after we initialise the zipWriter class as initially we only enquire on what is the extension of the main file, which here is a zip (MediaFile). After the extraction along with picking the  names of possible image files, the actual zip is deleted by keeping file pointer open in memory. 

Upon deleting of zipExtractor object, the __del__ would close the open file pointer, which is being actively
utilized further in the process. Thus introduction of LidarData Class would disrupt the flow and make it required to 
create another zip file or abandon the deletion in the ZipExtractor. 
Also if the ` 2 if` are removed, then we need to introduce `if` in places such as FrameProvider, task.py and possibly when
import export is being carried based on dimension.
Kindly let us know your comments on the same.

cvat/apps/engine/media_extractors.py
manasars: The only parameter we require for distinguishing 2D and 3D task is _dimension which by default is always 2D.
Only in the case of zipCompressedChunkWriter we require the same and only at this point we have an `if`. 
If another class is created for the same, such as zipCompressed3DChunkWriter along with IChunk3dWriter, then we require to place multiple `if` in frameProvider and task.py. Please let me know if you have any different opinion.

azhavoro: Does an additional type that is defined here https://github.com/openvinotoolkit/cvat/blob/develop/cvat/apps/engine/models.py#L34-L37 solve this problem?

manasars: Hi @azhavoro,

The additional type mentioned here will not help since the ext type 'zip' can have both images and pcd data and the additional type list accepts list as input whereas we have zip as input.

cvat/apps/engine/tests/test_rest_api.py
manasars: Made changes accordingly.

cvat/apps/engine/tests/test_rest_api.py
manasars: Made changes accordingly.

cvat/apps/engine/tests/test_rest_api.py
manasars: Made changes accordingly.

cvat/apps/engine/tests/test_rest_api.py
manasars: Made changes accordingly.

cvat-ui/src/reducers/interfaces.ts
manasars: As mentioned in earlier review inputs, we have similar implementation of ReviewStatus in both the places, hence we implemented similarly.

nmanovic:
@bsekachev , do you have any questions to the client or server part? Could you please review?

bsekachev:
@nmanovic 

I do not have critical comments about the patch. Looks like my comments and the most comments of other guys have been resolved. There are several unresolved conversations above opened not by me. 
Also I've noticed newly appeared client-side code issues (not critical) and the most of them are expected to be highlighted or even fixed automatically by static linters on precommit hook. Apparently it doesn't work or just ignored. And I believe we should clarify this moment with @manasars because getting rid of such the issues on code review was exact purpose of the linters integration.

cvat-core/src/session.js
```diff
@@ -974,6 +974,7 @@
                 use_zip_chunks: undefined,
                 use_cache: undefined,
                 copy_data: undefined,
+                dimension: undefined
```
bsekachev, line 977: Missed trailing comma

cvat-core/src/session.js
```diff
@@ -1452,6 +1453,16 @@
                             updatedFields = fields;
                         },
                     },
+                    dimension: {
+                    /**
+                     * @name enabled
+                     * @type {string}
```
bsekachev, line 1459: probably should be module:API.cvat.enums.TaskDimension

cvat-ui/src/components/actions-menu/actions-menu.tsx
```diff
@@ -11,6 +11,7 @@ import { MenuInfo } from 'rc-menu/lib/interface';
 import DumpSubmenu from './dump-submenu';
 import LoadSubmenu from './load-submenu';
 import ExportSubmenu from './export-submenu';
+import {DimensionType} from "../../reducers/interfaces";
```
bsekachev, line 14: Please, use single quotes

bsekachev: Missed spaces in {}

cvat-ui/src/components/actions-menu/actions-menu.tsx
```diff
@@ -22,7 +23,7 @@ interface Props {
     dumpActivities: string[] | null;
     exportActivities: string[] | null;
     inferenceIsActive: boolean;
-
+    taskDimension: DimensionType.DIM_3D | DimensionType.DIM_2D;
```
bsekachev, line 26: Just ``taskDimension: DimensionType;``

cvat-ui/src/components/actions-menu/dump-submenu.tsx
```diff
@@ -6,6 +6,7 @@ import React from 'react';
 import Menu from 'antd/lib/menu';
 import { DownloadOutlined, LoadingOutlined } from '@ant-design/icons';
 import Text from 'antd/lib/typography/Text';
+import {DimensionType} from '../../reducers/interfaces'
```
bsekachev, line 9: Missed spaces in {}

cvat-ui/src/components/actions-menu/dump-submenu.tsx
```diff
@@ -19,15 +20,17 @@ interface Props {
     menuKey: string;
     dumpers: any[];
     dumpActivities: string[] | null;
+    taskDimension: DimensionType.DIM_3D | DimensionType.DIM_2D;
```
bsekachev, line 23: Just ``taskDimension: DimensionType;``

cvat-ui/src/components/actions-menu/export-submenu.tsx
```diff
@@ -5,21 +5,24 @@
 import React from 'react';
 import Menu from 'antd/lib/menu';
 import Text from 'antd/lib/typography/Text';
+import {DimensionType} from '../../reducers/interfaces'
```
bsekachev, line 8: Missed spaces in {}

cvat-ui/src/components/actions-menu/export-submenu.tsx
```diff
@@ -5,21 +5,24 @@
...
 import { ExportOutlined, LoadingOutlined } from '@ant-design/icons';
 
 interface Props {
     menuKey: string;
     exporters: any[];
     exportActivities: string[] | null;
+    taskDimension: DimensionType.DIM_3D | DimensionType.DIM_2D;
```
bsekachev, line 15: Just ``taskDimension: DimensionType;``

cvat-ui/src/components/actions-menu/load-submenu.tsx
```diff
@@ -7,24 +7,25 @@ import Menu from 'antd/lib/menu';
 import Upload from 'antd/lib/upload';
 import Button from 'antd/lib/button';
 import Text from 'antd/lib/typography/Text';
+import {DimensionType} from "../../reducers/interfaces";
```
bsekachev, line 10: Missed spaces

bsekachev: Please, use single quotes

cvat-ui/src/components/annotation-page/top-bar/annotation-menu.tsx
```diff
@@ -161,6 +161,7 @@ export default function AnnotationMenuComponent(props: Props): JSX.Element {
                 dumpers,
                 dumpActivities,
                 menuKey: Actions.DUMP_TASK_ANNO,
+                taskDimension:jobInstance.task.dimension
```
bsekachev, line 164: Missed space

cvat-ui/src/components/annotation-page/top-bar/annotation-menu.tsx
```diff
@@ -169,11 +170,13 @@ export default function AnnotationMenuComponent(props: Props): JSX.Element {
                     onClickMenuWrapper(null, file);
                 },
                 menuKey: Actions.LOAD_JOB_ANNO,
+                taskDimension:jobInstance.task.dimension
```
bsekachev, line 173: Missed space

cvat-ui/src/components/annotation-page/top-bar/annotation-menu.tsx
```diff
@@ -169,11 +170,13 @@ export default function AnnotationMenuComponent(props: Props): JSX.Element {
...
             })}
             {ExportSubmenu({
                 exporters: dumpers,
                 exportActivities,
                 menuKey: Actions.EXPORT_TASK_DATASET,
+                taskDimension:jobInstance.task.dimension
```
bsekachev, line 179: Missed space

cvat-ui/src/components/model-runner-modal/detector-runner.tsx
```diff
@@ -14,6 +14,7 @@ import Text from 'antd/lib/typography/Text';
 import InputNumber from 'antd/lib/input-number';
 import Button from 'antd/lib/button';
 import notification from 'antd/lib/notification';
+import {DimensionType} from '../../reducers/interfaces'
```
bsekachev, line 17: Missed space

cvat-ui/src/reducers/interfaces.ts
bsekachev: @nmanovic Need to rewrite cvat-core in typescript to avoid this.

manasars:
Hi @bsekachev ,
The pre-commit hook had stopped working earlier due to some git config.
We have modified the same and now the lint changes are checked.
Fixed all the mentioned lint errors.

cvat-core/src/annotation-formats.js
```diff
@@ -117,6 +129,16 @@
                      */
                     get: () => data.enabled,
                 },
+                dimension: {
+                    /**
+                     * @name dimension
+                     * @type {string}
```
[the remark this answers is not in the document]
manasars: Made changes accordingly.

cvat-core/src/annotation-formats.js
```diff
@@ -58,6 +59,16 @@
...
```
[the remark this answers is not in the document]
manasars: Made changes accordingly.

cvat-core/src/annotation-formats.js
```diff
@@ -15,6 +15,7 @@
                 format: initialData.ext,
                 version: initialData.version,
                 enabled: initialData.enabled,
+                dimension: initialData.dimension
```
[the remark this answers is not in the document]
manasars: Made changes accordingly.

cvat-core/src/session.js
manasars: Made changes accordingly.

cvat-core/src/session.js
manasars: Made changes accordingly.

cvat-ui/src/components/annotation-page/top-bar/annotation-menu.tsx
manasars: Made changes accordingly.

cvat-ui/src/components/annotation-page/top-bar/annotation-menu.tsx
manasars: Made changes accordingly.

cvat-ui/src/components/annotation-page/top-bar/annotation-menu.tsx
manasars: Made changes accordingly.

cvat-ui/src/components/model-runner-modal/detector-runner.tsx
manasars: Made changes accordingly.

bsekachev (approved):

nmanovic (approved):

cvat/apps/engine/media_extractors.py
g-kartik: Please add docstrings for each method

Issue #1418, opened 2020-04-17
CVAT new UI: find the way to simplify the syntax for filters


aschernov:
### My actions before raising this issue
- [x] Read/searched [the docs](https://github.com/opencv/cvat/tree/master#documentation)
- [x] Searched [past issues](/issues)

### Expected Behaviour

The variety of filters should be visually and manually easy and quick to use. Without difficult syntax and with flexible opportunities to quickly correct or change them.
### Current Behaviour

1. The syntax for simple labels became harder to write. For example:
**before**: _vehicle_ **after**: _label =="vehicle"_
2. There is no way to correct a filter if we made some mistake. We can only delete that we wrote before and write a filter from the beginning again:
![image](https://user-images.githubusercontent.com/45849884/79554219-6d9aa800-80a6-11ea-8b3f-3be90284136c.png)

### Possible Solution

Simplify the syntax. Add the opportunity to change symbols in previous filters. Create a comfortable for using panel of filters.

### Context

We spend much time to understand how to use some filters correctly and how to write them. Not all users can do it. It influences the validation process.

PR #2743, opened 2021-01-29
Annotations filters UI

vnishukov:
### Motivation and context
[1418: CVAT new UI: find the way to simplify the syntax for filters](https://github.com/openvinotoolkit/cvat/issues/1418)

### How has this been tested?

- Go to annotation page
- Click to filters input
- Play around with form
- **Shift+Left click: add open bracket for item**
- **Shift+Right click: add close bracket for item**
- **Alt+Left click: remove open bracket for item**
- **Alt+Right click: remove close bracket for item**
- **Right click on filter input open confirm dialog to clean up filters**

### Checklist

### License

```python
#
```


vnishukov:
**FiltersInput** component has to be removed after review

cvat-ui/package-lock.json
```diff
@@ -12886,6 +12926,7 @@
       "requires": {
         "axios": "^0.21.1",
         "browser-or-node": "^1.2.1",
+        "cvat-data": "file:../cvat-data",
```
bsekachev, line 12929: This line is probably the reason of CI fail. Try to delete it.

nmanovic:
@vnishukov , by a reason first image is zoomed when I open a job. Can it be connected with your changes?

nmanovic:
In help message:
- Can we replace `[Right] and [Left]` on icons for left and right buttons?
- Do we need ServerID and ClientID? The user can see ID. I will prefer to keep only ID = ClientID.
- `CVAT uses json queries to perform search.`. Why do we need mention that? Is it important?

nmanovic:
Can we have an icon to clear filter somewhere instead of context menu?

image

nmanovic:
Why did you decide to use hotkeys to add `(` and `)`?

nmanovic:
@vnishukov , did you look at https://github.com/ukrbublik/react-awesome-query-builder? Can it help?

vnishukov:
> In help message:
> 
> * Can we replace `[Right] and [Left]` on icons for left and right buttons?
      This is a dynamic values fetched from global shortcuts map, don't think we shall hardcode it
> * Do we need ServerID and ClientID? The user can see ID. I will prefer to keep only ID = ClientID.
      This developed as it was requested in concept. Maybe we can ask @bsekachev?
> * `CVAT uses json queries to perform search.`. Why do we need mention that? Is it important?
      This was mentioned before. JSON queries is still used to apply the filters, that's why it is still present. Should I remove it?

vnishukov:
@nmanovic 

> Why did you decide to use hotkeys to add `(` and `)`?

I'm absolutely sure this is more comfortable than it was described in concept (additional combine button in change filter panel). We should improve UX and make it possible for users to achieve theirs goals using minimum of clicks, navigations, etc. In current approach user can group filters by single clicks for each of them. In the approach described in concept it will be couple of clicks, and extra button. But the main benefit here is that multigrouping (two or more braces) in the current approach we still handle it by two clicks, in other approach it requires an "by hand" input in filters pane which is slowly and takes more efforts from user. Let's give a chance to test it but users as well?

vnishukov:
@nmanovic 

> @vnishukov , did you look at https://github.com/ukrbublik/react-awesome-query-builder? Can it help?

Current logic of json queries was not changed, as it works fine. What benefits will we have to change it with something else instead?

nmanovic:
> @nmanovic
> 
> > @vnishukov , did you look at https://github.com/ukrbublik/react-awesome-query-builder? Can it help?
> 
> Current logic of json queries was not changed, as it works fine. What benefits will we have to change it with something else instead?

@vnishukov , I see your point. @bsekachev , @ActiveChooN , what do you think?

bsekachev:
@ActiveChooN Can I ask you to review the PR?

vnishukov:
> @vnishukov , by a reason first image is zoomed when I open a job. Can it be connected with your changes?

This is the one of the reason of failed tests. I didn't make any specific with images, changes affect only right side panel, but seems somehow it affects annotation area as well

bsekachev:
Hi, I tried the patch and here are some thoughts:

1. Why do we need "@babel/plugin-proposal-nullish-coalescing-operator"? Let's just use "||" to get default value. 
2. Bug: Image scaled after opening a job (already mentioned)
3. Delete by right mouse click: not obvious (why not left icon as it was?)
4. Disappeared feature: the latest 10 filters list to quick select on of them
5. Personally, it took me 10 minutes to understand how to add brackets in current implementation
6. Bug: the object with ID 6 is inaccessible (literally it is out of scroll area, scroll is set to maximum on the screenshot)
Screenshot from 2021-02-09 18-30-48

7. Pressing right button (to remove a bracket) looks quite risky because in case when I pressed it before Alt (or missed the key in some reasons) it opens popover with suggestion to remove filter and button "Yes" close to the cursor
8. Bug: In attribute annotation mode we still have old component for filters
9. Bug: Not critical but looks strange when a label does not have any attributes
Screenshot from 2021-02-09 18-38-27

10. Bug: I set empty_frame=="false" and CVAT still navigates by empty frames. How is it expected to work? We already have another feature to empty frames navigation actually
11. Probably bug: boolean attributes have only one option (I understand that it could be replaced with ==, !==, but for me personally it looks strange (maybe @ActiveChooN  think in other way)
Screenshot from 2021-02-09 18-41-59

12. Disappeared feature: filters like width > height; height > width; height == width, etc
13. Type filter does not contain "Tag" option
14. Bug: Filter: occluded=="true" hides occluded object (occluded=="false" also hides occluded object as it expected)

I also looked at https://github.com/ukrbublik/react-awesome-query-builder and I would say it looks promising. IMO UX is simplier and the builder is quite flexible. I have some concerns related with output format (out of the box it is different, but GitHub says "Export to MongoDb, SQL, JsonLogic or your **custom forma**t", so maybe we will be able to setup it properly or adjust format that cvat-core expects (it is also possible because cvat-core expects custom format, not jsonpath and we have our code for convertation which could be adjusted). And additionally we need  to be sure that this library is able to provide necessary flexibility (in possible queries). After quick review it looks like it is able more or less, but need more investigations.

bsekachev:
@vnishukov @ActiveChooN 

Let's have a meeting tomorrow about the patch, guys. I'll send an invite.

bsekachev:
@nmanovic 

> Do we need ServerID and ClientID? The user can see ID. I will prefer to keep only ID = ClientID. 

Previously when dump file included server IDs some people used serverID to find objects from dump file in UI. 
Now a dump file does not include id anymore, so I think it can be removed.

cvat-ui/src/components/annotation-page/annotation-filter/annotation-filter-item/annotation-filter-item.tsx
```diff
@@ -0,0 +1,115 @@
+// Copyright (C) 2020-2021 Intel Corporation
+//
+// SPDX-License-Identifier: MIT
+
+import { Tag } from 'antd';
```
bsekachev, line 5: Let's use the same import policy, now project uses modular import: ``import Tag from /antd/lib/Tag``
For information: https://nsisodiya.medium.com/reduce-size-in-ant-design-bundle-size-c35dad9ce3a8

cvat-ui/src/components/annotation-page/annotation-filter/annotation-filter-item/annotation-filter-item.tsx
```diff
@@ -0,0 +1,115 @@
...
+        case ActionType.removeRight:
+            state.right.pop();
+            return { ...state };
+        default:
+            return state;
+    }
+};
+function AnnotationFilterItem({
+    item, onEdit, onDelete, onGrouping,
+}: Props): ReactElement {
```
bsekachev, line 56: ReactElement > JSX.Element

cvat-ui/src/components/annotation-page/annotation-filter/annotation-filter-item/annotation-filter-item.tsx
```diff
@@ -0,0 +1,115 @@
...
+}
+
+const reducer = (state: State, action: { type: ActionType; payload?: any }): State => {
+    switch (action.type) {
+        case ActionType.addLeft:
+            return { ...state, left: [...state.left, '('] };
+        case ActionType.addRight:
+            return { ...state, right: [...state.right, ')'] };
+        case ActionType.removeLeft:
+            state.left.pop();
```
bsekachev, line 45: state.right property before reducing is equal to state.right property after.
It is better to use something destructive like that:
``return { ...state, left: state.left.slice(0, -1) };``

cvat-ui/src/components/annotation-page/annotation-filter/annotation-filter-item/annotation-filter-item.tsx
```diff
@@ -0,0 +1,115 @@
...
+            return { ...state };
+        case ActionType.removeRight:
+            state.right.pop();
+            return { ...state };
```
bsekachev, line 49: The same

cvat-ui/src/components/annotation-page/annotation-filter/annotation-filter-item/annotation-filter-item.tsx
```diff
@@ -0,0 +1,115 @@
...
+    anotherAttributeValue: string;
+    left: string[];
+    right: string[];
+}
+
+interface Props {
+    item: State;
+    onEdit: Function;
+    onDelete: Function;
+    onGrouping: Function;
```
bsekachev, line 28: Can we use precise type? Function is too abstract

cvat-ui/src/components/annotation-page/annotation-filter/annotation-filter-pane/annotation-filter-pane.tsx
```diff
@@ -0,0 +1,185 @@
...
+                    style={{ cursor: filters.length ? 'default' : 'pointer' }}
+                    onContextMenu={(e: React.MouseEvent<HTMLElement, MouseEvent>) => confirmClearFilters(e)}
+                >
+                    {filters?.length ? (
+                        <>
+                            {filters.map((item: any) => (
+                                <AnnotationFilterItem
+                                    key={item.id}
+                                    item={item}
+                                    onGrouping={(filter: any) => {
```
bsekachev, line 124: Please, avoid using ``any``, I believe it would be easy to specify precise types here

cvat-ui/src/components/annotation-page/annotation-filter/annotation-filter-pane/annotation-filter-pane.tsx
```diff
@@ -0,0 +1,185 @@
...
+// SPDX-License-Identifier: MIT
+
+/* eslint-disable jsx-a11y/no-static-element-interactions */
+/* eslint-disable jsx-a11y/click-events-have-key-events */
+
+import { FilterOutlined, PlusOutlined, QuestionOutlined } from '@ant-design/icons';
+import {
+    changeAnnotationsFilters as changeAnnotationsFiltersAction,
+    fetchAnnotationsAsync,
+} from 'actions/annotation-actions';
```
bsekachev, line 12: Please, do internal import after 3rdparty import

cvat-ui/src/components/annotation-page/annotation-filter/annotation-filter-pane/annotation-filter-pane.tsx
```diff
@@ -0,0 +1,185 @@
...
+    ReactElement, useEffect, useRef, useState,
+} from 'react';
+import { connect } from 'react-redux';
+import { v4 as uuidv4 } from 'uuid';
+import AnnotationFilterItem from '../annotation-filter-item/annotation-filter-item';
+import AnnotationFilterPanel from '../annotation-filter-panel/annotation-filter-panel';
+import './annotation-filter-pane.scss';
+
+interface DispatchToProps {
+    changeAnnotationsFilters(value: any): void;
```
bsekachev, line 24: Probably its type string[], not any. But I may be wrong

cvat-ui/src/cvat-canvas-wrapper.ts
```diff
@@ -1,4 +1,4 @@
-// Copyright (C) 2020 Intel Corporation
+// Copyright (C) 2021 Intel Corporation
```
bsekachev, line 1: Please, abort changes in this file

cvat-ui/src/icons.tsx
```diff
@@ -1,4 +1,4 @@
-// Copyright (C) 2020 Intel Corporation
+// Copyright (C) 2021 Intel Corporation
```
bsekachev, line 1: Please, abort changes in this file

cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/controls-side-bar.tsx
```diff
@@ -1,4 +1,4 @@
-// Copyright (C) 2020 Intel Corporation
+// Copyright (C) 2021 Intel Corporation
```
bsekachev, line 1: Should be 2020-2021

cvat-ui/src/components/annotation-page/standard-workspace/objects-side-bar/objects-list-header.tsx
```diff
@@ -1,22 +1,21 @@
-// Copyright (C) 2020 Intel Corporation
+// Copyright (C) 2021 Intel Corporation
```
bsekachev, line 1: Should be 2020-2021

cvat-ui/src/components/annotation-page/standard-workspace/objects-side-bar/objects-list-header.tsx
```diff
@@ -1,22 +1,21 @@
...
+    EyeOutlined,
+    LockFilled,
+    UnlockOutlined,
 } from '@ant-design/icons';
+import { Col, Row } from 'antd/lib/grid';
 import Tooltip from 'antd/lib/tooltip';
-
-import AnnotationsFiltersInput from 'components/annotation-page/annotations-filters-input';
 import StatesOrderingSelector from 'components/annotation-page/standard-workspace/objects-side-bar/states-ordering-selector';
+import React from 'react';
```
bsekachev, line 16: Please, do 3rdparty import before internal import
What is a reason to change import order in this file?

cvat-ui/src/components/annotation-page/annotation-filter/annotation-filter.enum.ts
```diff
@@ -0,0 +1,100 @@
...
+    operator,
+    value,
+    attribute,
+    attributeOperator,
+    attributeValue,
+    anotherAttributeLabel,
+    anotherAttributeValue,
+    fillState,
+    partialReset,
+    reset,
```
bsekachev, line 29: I expect all actions to be a verb (or contain a verb)
For example I cannot understand what such actions like ``concatenator``, ``value``, ``operator``, ``attribute``, etc. do.
Also to follow the current codestyle let actions to be written in upper case: FILTER_BY, FILL_STATE, PARTIAL_RESET.

bsekachev:
Some common comments after code review:
1. Using ``any`` and some abstract types is too often, we lose the main benefit of TypeScript in this case
2. I see code style and header changes in some files which are not expected to be affected by the patch

vnishukov:
@bsekachev 
> 1. Why do we need "@babel/plugin-proposal-nullish-coalescing-operator"? Let's just use "||" to get default value.
Using `||` we apply right branch value when the left branch value is false but not only includes `undefined` and `null` but also `0` and `''`.
I believe `??` is much more specific as to apply only for `undefined` and `null` and not for `0` and `''` (which more like a js assertions side effect)

bsekachev:
> Using || we apply right branch value when the left  

I read about the difference, but in the code I do not see where it is important in our case. Could you please direct me?

cvat-ui/src/components/annotation-page/annotation-filter/annotation-filter-panel/annotation-filter-panel.tsx
```diff
@@ -0,0 +1,576 @@
...
+                    isFilled(StateFields.anotherAttributeValue);
+            }
+        } else {
+            isValid = isValid && isFilled(StateFields.operator);
+        }
+        setFormValid(isFirst ? isValid : isValid && isFilled(StateFields.concatenator));
+    }, [state]);
+
+    useEffect(() => {
+        setTimeout(() => {
```
ActiveChooN, line 204: I guess it's better to use recap for timeout in `useEffect` hook

ActiveChooN (commented):
Most problems was described by @bsekachev and i have to add only few things:
1. It's better to save state of the filter input for reducing clicks for similar filters (in case we want to select some obejcts with differnt clientID).
2. There is no way to insert filter in the middle of the query, only remove  secand half of the query.
3. I guess it can be really nice feature to insert query as a string like it works for label editor. It can be helpful for common filters.

> Probably bug: boolean attributes have only one option (I understand that it could be replaced with ==, !==, but for me personally it looks strange (maybe @ActiveChooN think in other way)

I did not notice anithing strange.

vnishukov:
Reimplemented with new UI in https://github.com/openvinotoolkit/cvat/pull/2871 using https://github.com/ukrbublik/react-awesome-query-builder

PR #2784, opened 2021-02-09
Add "id" to sidebar issue item. Update Cypress test.

dvkruchinin:
### Motivation and context
At the moment, test checks the highlighting of the issue region when hovering the mouse over the sidebar issue item elements in order from top to bottom. There may be a situation where the server returns these elements in reverse order. In this case, the test will fail with an error. This patch adds an ```id``` to the sidebar issue elements. The ```id``` corresponds to the ```id``` of the issue region. And the test checks that when hover the mouse over the sidebar issue item with ```id``` N, the element of the issue region with the same ```id``` will be highlighted.

### Checklist

### License

```python
#
```

### DIFF
--- a/tests/cypress/integration/actions_users/registration_involved/case_28_review_pipeline_feature.js
+++ b/tests/cypress/integration/actions_users/registration_involved/case_28_review_pipeline_feature.js
@@ -352,21 +352,13 @@ context('Review pipeline feature', () => {
 
         it('Select an issue on sidebar. Issue indication has changed the color for highlighted issue', () => {
             cy.collectIssueRegionId().then(($issueRegionList) => {
-                for (let i = 0; i < $issueRegionList.length; i++) {
-                    cy.get(`#cvat-objects-sidebar-issue-item-${$issueRegionList[i]}`)
-                        .trigger('mousemove')
-                        .trigger('mouseover');
-                    cy.get(`#cvat_canvas_issue_region_${$issueRegionList[i]}`).should(
-                        'have.attr',
-                        'fill',
-                        'url(#cvat_issue_region_pattern_2)',
-                    );
-                    cy.get(`#cvat-objects-sidebar-issue-item-${$issueRegionList[i]}`).trigger('mouseout');
-                    cy.get(`#cvat_canvas_issue_region_${$issueRegionList[i]}`).should(
-                        'have.attr',
-                        'fill',
-                        'url(#cvat_issue_region_pattern_1)',
-                    );
+                for (const issueRegionID of $issueRegionList) {
+                    const objectsSidebarIssueItem = `#cvat-objects-sidebar-issue-item-${issueRegionID}`;
+                    const canvasIssueRegion = `#cvat_canvas_issue_region_${issueRegionID}`;
+                    cy.get(objectsSidebarIssueItem).trigger('mousemove').trigger('mouseover');
+                    cy.get(canvasIssueRegion).should('have.attr', 'fill', 'url(#cvat_issue_region_pattern_2)');
+                    cy.get(objectsSidebarIssueItem).trigger('mouseout');
+                    cy.get(canvasIssueRegion).should('have.attr', 'fill', 'url(#cvat_issue_region_pattern_1)');
                 }
             });
         });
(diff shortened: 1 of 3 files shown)


tests/cypress/integration/actions_users/registration_involved/case_28_review_pipeline_feature.js
```diff
@@ -351,25 +351,23 @@ context('Review pipeline feature', () => {
...
-                        cy.get(sidebarIssueItems[i]).trigger('mouseout');
-                        cy.get(`#cvat_canvas_issue_region_${$issueRegionList[index]}`).should(
-                            'have.attr',
-                            'fill',
-                            'url(#cvat_issue_region_pattern_1)',
-                        );
-                        index++;
-                    }
-                });
+                for (let i = 0; i < $issueRegionList.length; i++) {
```
bsekachev, line 355: I believe construction for .. of is more suitable here
``for (const issueRegionID of issueRegionList)``

tests/cypress/integration/actions_users/registration_involved/case_28_review_pipeline_feature.js
```diff
@@ -351,25 +351,23 @@ context('Review pipeline feature', () => {
...
+                    cy.get(`#cvat-objects-sidebar-issue-item-${$issueRegionList[i]}`)
```
bsekachev, line 356: `#cvat-objects-sidebar-issue-item-${issueRegionID}`

tests/cypress/integration/actions_users/registration_involved/case_28_review_pipeline_feature.js
bsekachev: BTW, I looked at implementation of ``collectIssueRegionId`` command and I see the regex there does not consider negative IDs
But when new issue is created and not saved on the server yet, it has exactly negative ID. It could lead to unexpected bugs in the future
Possible solution is to change regex to ``.match(/\-?\d+$/)``
What do you think?

tests/cypress/integration/actions_users/registration_involved/case_28_review_pipeline_feature.js
bsekachev: The same below

tests/cypress/integration/actions_users/registration_involved/case_28_review_pipeline_feature.js
```diff
@@ -351,25 +351,23 @@ context('Review pipeline feature', () => {
...
+                        .trigger('mousemove')
+                        .trigger('mouseover');
+                    cy.get(`#cvat_canvas_issue_region_${$issueRegionList[i]}`).should(
+                        'have.attr',
+                        'fill',
+                        'url(#cvat_issue_region_pattern_2)',
+                    );
+                    cy.get(`#cvat-objects-sidebar-issue-item-${$issueRegionList[i]}`).trigger('mouseout');
```
bsekachev, line 364: If we use it twice, we could create a const to store the value in iteration

And also const for ``cvat-objects-sidebar-issue-item``

tests/cypress/integration/actions_users/registration_involved/case_28_review_pipeline_feature.js
dvkruchinin: > I believe construction for .. of is more suitable here

You're right. It's more convenient. Thanks.

> Possible solution is to change regex to

Right. This may come in handy in the future. It also allowed you to slightly reduce the code in the ```checkIssueRegion``` command

tests/cypress/integration/actions_users/registration_involved/case_28_review_pipeline_feature.js
dvkruchinin: Done.

tests/cypress/integration/actions_users/registration_involved/case_28_review_pipeline_feature.js
dvkruchinin: Fixed it.

bsekachev (approved):

PR #2871, opened 2021-02-26
Annotations filters new UI

### LINKED ISSUE #1418: CVAT new UI: find the way to simplify the syntax for filters

vnishukov:
### Motivation and context

Resolved #1418

### How has this been tested?

Go to annotation page, click filter Icon on top bar right (see examples images below)
image
image
image
image

### Checklist

### License

```python
#
```


nmanovic:
@vnishukov , need to fix the problem with CI build.

bsekachev:
Hi, I tried the patch. Mostly working fine, but several comments/questions below.
@ActiveChooN Please, try to run also this patch and left some comments about.

1. When I open CVAT in small windows (first screenshot), layout looks alright (**except of minor issues with position of sort by selector, it is expected to be right-aligned**)
Screenshot from 2021-03-01 16-46-10

But when I open CVAT in my different configuration (Full HD), something wrong with layout
Screenshot from 2021-03-01 16-46-18

2. I have attributes ``type`` (radio) and ``color`` (select) and I want to set equality filter on them (like on the screenshot below), but I can't because there are not option to select ``color`` attribute, only ``text``.
Screenshot from 2021-03-01 16-50-27

3. I would vote for changing color to blue
Screenshot from 2021-03-01 16-54-35

4. Probably "Tag" is missed in possible types. (I see "Tag" in "Shapes", but "Tag" is not a shape at all)
Screenshot from 2021-03-01 16-57-04

5. Is comparison label with these fields valid? (the same for type and shape)
Screenshot from 2021-03-01 17-02-14

6. Is raw filter editor difficult to implement? (I see it was cut from the interface)

cvat-core/package.json
```diff
@@ -41,7 +41,7 @@
     "form-data": "^2.5.0",
     "jest-config": "^26.6.3",
     "js-cookie": "^2.2.0",
-    "jsonpath": "^1.1.0",
+    "json-logic-js": "^2.0.0",
```
bsekachev, line 44: Why so many changes in package-lock if you only installed one package?

.github/workflows/eslint.yml
```diff
@@ -29,7 +29,7 @@ jobs:
 
             echo "ESLint version: "`npx eslint --version`
             echo "The files will be checked: "`echo ${changed_files_eslint}`
-            npx eslint ${changed_files_eslint} -f node_modules/eslint-detailed-reporter/lib/detailed.js -o ./eslint_report/eslint_checks.html
+            npx eslint --no-error-on-unmatched-pattern ${changed_files_eslint} -f node_modules/eslint-detailed-reporter/lib/detailed.js -o ./eslint_report/eslint_checks.html
```
bsekachev, line 32: What are you trying to solve?

cvat-core/tests/internal/filter.js.bkp
```diff
@@ -0,0 +1,121 @@
+// Copyright (C) 2020-2021 Intel Corporation
```
bsekachev, line 1: Need to update these test, not just drop them out. We can do it in separated PR, but in this case, please, create an issue.

cvat-ui/package-lock.json
```diff
@@ -1081,6 +1081,19 @@
       "resolved": "https://registry.npmjs.org/@ctrl/tinycolor/-/tinycolor-3.2.0.tgz",
       "integrity": "sha512-cP1tbXA1qJp/er2CJaO+Pbe38p7RlhV9WytUxUe79xj++Q6s/jKVvzJ9U2dF9f1/lZAdG+j94A38CsNR+uW4gw=="
     },
+    "@date-io/core": {
```
bsekachev, line 1084: Too many changes in this file also. I checked it out to develop, then performed query-builder installation and git stat gives me 68 changes in this file.  In your case I see 11945 changes, what is a reason?

cvat-core/package.json
vnishukov: I don't do nothing special. Just npm i. Is it important to store auto generated file in particular structure, while generation of lock files not guaranteed same structure through different iteration?

.github/workflows/eslint.yml
vnishukov: Original issue:

ESLint version: v7.11.0

The files will be checked: cvat-core/src/annotations-collection.js cvat-core/src/annotations-filter.js cvat-core/src/annotations.js cvat-core/src/session.js cvat-core/tests/internal/filter.js cvat-ui/src/actions/annotation-actions.ts cvat-ui/src/components/annotation-page/annotation-page.tsx **cvat-ui/src/components/annotation-page/annotations-filters-input.tsx** cvat-ui/src/components/annotation-page/attribute-annotation-workspace/attribute-annotation-sidebar/attribute-annotation-sidebar.tsx cvat-ui/src/components/annotation-page/standard-workspace/objects-side-bar/objects-list-header.tsx cvat-ui/src/components/annotation-page/top-bar/filters-modal.tsx cvat-ui/src/components/annotation-page/top-bar/right-group.tsx cvat-ui/src/components/annotation-page/top-bar/top-bar.tsx cvat-ui/src/containers/annotation-page/standard-workspace/objects-side-bar/objects-list.tsx cvat-ui/src/containers/annotation-page/top-bar/filters-modal.tsx cvat-ui/src/containers/annotation-page/top-bar/top-bar.tsx cvat-ui/src/icons.tsx cvat-ui/src/reducers/annotation-reducer.ts cvat-ui/src/reducers/interfaces.ts tests/cypress/integration/actions_tasks_objects/case_18_filters_functionality.js

Oops! Something went wrong! :(

ESLint: 7.11.0

No files matching the pattern **"cvat-ui/src/components/annotation-page/annotations-filters-input.tsx"** were found.

Please check for typing mistakes in the pattern.

Error: Process completed with exit code 2.

This file **cvat-ui/src/components/annotation-page/annotations-filters-input.tsx** was deleted, but condition collect all files in PR.

cvat-core/tests/internal/filter.js.bkp
vnishukov: Will create an issue

cvat-ui/package-lock.json
vnishukov: I don't do nothing special. Just npm i. Is it important to store auto generated file in particular structure, while generation of lock files not guaranteed same structure through different iteration?

vnishukov:
@bsekachev

> Hi, I tried the patch. Mostly working fine, but several comments/questions below.
> @ActiveChooN Please, try to run also this patch and left some comments about.
> 
> 1. When I open CVAT in small windows (first screenshot), layout looks alright (**except of minor issues with position of sort by selector, it is expected to be right-aligned**)
>    Screenshot from 2021-03-01 16-46-10
> 
> But when I open CVAT in my different configuration (Full HD), something wrong with layout
> Screenshot from 2021-03-01 16-46-18
> 
> 1. I have attributes `type` (radio) and `color` (select) and I want to set equality filter on them (like on the screenshot below), but I can't because there are not option to select `color` attribute, only `text`.
>    Screenshot from 2021-03-01 16-50-27
> 2. I would vote for changing color to blue
>    Screenshot from 2021-03-01 16-54-35
> 3. Probably "Tag" is missed in possible types. (I see "Tag" in "Shapes", but "Tag" is not a shape at all)
>    Screenshot from 2021-03-01 16-57-04
> 4. Is comparison label with these fields valid? (the same for type and shape)
>    Screenshot from 2021-03-01 17-02-14
> 5. Is raw filter editor difficult to implement? (I see it was cut from the interface)

0. Will check top-bart alignment
1. As I understood query-builder allow to compare same types of fields. In our case radio will cast to text
2. Will change to blue
3. Probably yes. It was developed based on initial concept
4. Will doublecheck what happening here
5. I don't think is difficult but can we make it with separate enhancement PR?

bsekachev:
@vnishukov 

> As I understood query-builder allow to compare same types of fields. In our case radio will cast to text

Radio attribute and text attribute are the same type, I agree (string). But we also have select attribute (see at right on the screenshot, color attribute) that is also string, but the list is lack of this select attribute. It contains only the text attribute

> I don't think is difficult but can we make it with separate enhancement PR?

Sure.

cvat-ui/package-lock.json
bsekachev: You do not need ``npm i`` because it updates all dependencies from ``package.json``. Instead of it you need only install one more dependency. If you with to update all, please, send it as a separated PR.  To install all dependencies from ``package-lock.json`` you need ``npm ci``

.github/workflows/eslint.yml
bsekachev: Got it. Was not it fixed in #2888?

cvat-core/tests/internal/filter.js.bkp
bsekachev: Okay. This ``.bpk`` file probably should be removed since we have all these files on git, right?

bsekachev:
@vnishukov Please, work with @dvkruchinin if you have difficulties with CI.

.github/workflows/eslint.yml
vnishukov: Maybe, but not merged to develop for sure. Any way it ill be inlined in next develop merge I think

cvat-core/tests/internal/filter.js.bkp
vnishukov: OK

cvat-ui/package-lock.json
vnishukov: OK

vnishukov:
@bsekachev 

> @vnishukov Please, work with @dvkruchinin if you have difficulties with CI.

Root cause already defined. Will be fixed soon

cvat-ui/src/components/annotation-page/top-bar/filters-modal.tsx
```diff
@@ -0,0 +1,266 @@
...
+import Button from 'antd/lib/button';
+import Modal from 'antd/lib/modal';
+import React, { useEffect, useState } from 'react';
+import {
+    Builder, Config, ImmutableTree, JsonLogicTree, Query, Utils as QbUtils,
+} from 'react-awesome-query-builder';
+import AntdConfig from 'react-awesome-query-builder/lib/config/antd';
+import 'react-awesome-query-builder/lib/css/styles.css';
+import { useDispatch, useSelector } from 'react-redux';
+import { CombinedState } from 'reducers/interfaces';
```
ActiveChooN, line 17: Can you organize imports, please? For example react import should be at the top of the file. For ant components you can use separate imports like `ant/lib/<component>` for bundle minification. Local import should be after all external components import.

cvat-ui/src/components/annotation-page/top-bar/right-group.tsx
```diff
@@ -2,27 +2,40 @@
...
+        showFilters, showStatistics, changeWorkspace, workspace, jobInstance,
     } = props;
 
+    const filters = useSelector((state: CombinedState) => state.annotation.annotations.filters);
+    const dispatch = useDispatch();
+    const filtersClickHandler = (e: React.MouseEvent<HTMLElement, MouseEvent>): void => {
+        if (e.altKey) {
+            dispatch(changeAnnotationsFilters([]));
+            dispatch(fetchAnnotationsAsync());
+        } else {
```
ActiveChooN, line 34: Guess it can be better to add popup with helper for such case

ActiveChooN (commented):
Guess it would be a little bit better if button will be gray with more margin at the top.

изображение

If i click on one of these empty dropdown values, UI will crush.

изображение

Guess it can happend due to old values in `filtersHistory` in local storage.

изображение

cvat-ui/src/components/annotation-page/top-bar/right-group.tsx
vnishukov: @ActiveChooN 
This was an additional quick way to clear the filters. We already have particular button on the panel. I believe in this case we can just remove alt-click listener.

cvat-ui/src/components/annotation-page/top-bar/filters-modal.tsx
vnishukov: OK

bsekachev:
Hi @vnishukov 

I've just remembered about feature ``createObjectURL`` which works for objects saved on the sever. 
Please, be sure it also works in new filters implementation because filters are used for this feature (``serverID`` field to be more specific)

vnishukov:
> 4. Is comparison label with these fields valid? (the same for type and shape)
>    Screenshot from 2021-03-01 17-02-14

Yes it's valid because all of these stuffs are the same type in terms of _react-awesome-query_ library (`type: select`)

bsekachev:
> Yes it's valid because all of these stuffs are the same type in terms of react-awesome-query library (type: select)

From the point of view UX I am not sure that these comparisons are valid. Do you think it is?
And why do ``clientID`` and ``serverID`` for example have ``select`` type if they are numbers and cannot be changed?

vnishukov:
> > Yes it's valid because all of these stuffs are the same type in terms of react-awesome-query library (type: select)
> 
> And why do `clientID` and `serverID` for example have `select` type if they are numbers and cannot be changed?

Well, it is number in terms of each select item, but in terms of fields config it's a single select. This is migrated from initial concept where we wanted to list all of the client and server ids in the dropbox. We actually can replace it with something like number or range of numbers, but I'm not sure if it will be clear for user how exactly he/she should populate these field(s) and with what values. What do you think about that?

bsekachev:
> Or we can get rid of clientID and serverID to be appeared in the right side of comparison as well.
I believe the last one it's a best and most logical solution.
What do you think about that?

As I understand you already done it. Agree, I do not see any cases when we need clientID, serverID at the right side of comparison (only at left).

bsekachev:
Hi @vnishukov 

I tested the patch one more time.
Navigation "Go back/next with a filter" does not work. It should navigate to the nearest frame that consist any objects (considering applied filter)

"Create object URL" does not work. It creates URL, but when I try to open it all objects are still visible and no any filters applied (at the same time filters button is blue, so apparently UI considers it to be activated)

Logic with attribute still not clear. I have attribute "color" (select) and attribute "type" (radio buttons), both possible values are string[], but the constructor  does not allow to compare them. 
Possible pairs for "color" attribute: label, type, shape. Possible pairs for "type": attribute "text" (which is text)

bsekachev:
One more possible issue found:

I have "label 1" which contains attribute "type" (radio button)
I have "label 2" which contains attribute "type" (number)

Looks like constructor doesn't work correctly in this case. It considers "type" as number attribute only. Should not we link attributes with their parent labels? Something like: "type (label 1)", "type (label 2)"?

.github/workflows/eslint.yml
nmanovic: @vnishukov , it was merged into develop. Probably we can remove the line.

cvat-core/package.json
nmanovic: @bsekachev , It looks like the number of changes are reasonable now.

nmanovic:
- [ ] Server ID should be removed
- [ ] Client ID should be renamed to Shape ID
- [ ] Shape ID should be a number (not select). Nobody will select from 1000 values. If you don't know the exact number, nothing to select.
- [ ] Fix bug: add rule for "Server ID" or "Client ID", as select you will see only 1 variant (let's say "1"). Enter "2" and it will crash UI.
TypeError
(t.children || t.value).toLowerCase is not a function

TypeError: (t.children || t.value).toLowerCase is not a function
    at t.filterOption (http://localhost:8080/cvat-ui.dffe796fde25f2595563.min.js:388:377502)
    at http://localhost:8080/cvat-ui.dffe796fde25f2595563.min.js:24:146169
    at Array.forEach (<anonymous>)
    at y (http://localhost:8080/cvat-ui.dffe796fde25f2595563.min.js:24:145984)
    at http://localhost:8080/cvat-ui.dffe796fde25f2595563.min.js:36:225400
    at Object.pi [as useMemo] (http://localhost:8080/cvat-ui.dffe796fde25f2595563.min.js:348:61567)
    at t.useMemo (http://localhost:8080/cvat-ui.dffe796fde25f2595563.min.js:332:5984)
    at x (http://localhost:8080/cvat-ui.dffe796fde25f2595563.min.js:36:225346)
    at Ko (http://localhost:8080/cvat-ui.dffe796fde25f2595563.min.js:348:57930)
    at Pi (http://localhost:8080/cvat-ui.dffe796fde25f2595563.min.js:348:65886)
- [ ] Should select and radio have the same type (`select`)? No radio is converted to string

bsekachev:
@nmanovic 

>  Server ID should be removed

It cannot be removed because "Create object URL" works via serverID.

nmanovic:
> @nmanovic
> 
> > Server ID should be removed
> 
> It cannot be removed because "Create object URL" works via serverID.
> 
> > Shape ID should be a number (not select). Nobody will select from 1000 values
> 
> It was Data Annotation Team request

@bsekachev , let's discuss it one more time. Need to understand how they are using the feature.

bsekachev:
> > @nmanovic
> > > Server ID should be removed
> > 
> > 
> > It cannot be removed because "Create object URL" works via serverID.
> > > Shape ID should be a number (not select). Nobody will select from 1000 values
> > 
> > 
> > It was Data Annotation Team request
> 
> @bsekachev , let's discuss it one more time. Need to understand how they are using the feature.

I've already sent a letter to Andrey. Agree, that this using looks strange.

bsekachev:
@nmanovic @ActiveChooN 

Issues with this patch:

- It is not clear how to implement input for Object ID that support both predefined and custom values "out-of-the-box" 
Related issue: https://github.com/ukrbublik/react-awesome-query-builder/issues/191 ('select' type does not support custom values since antd does not support them)
I tried to use 'multiselect' instead of 'select' because it supports ``allowCustomValues`` configuration. Nevertheless it does not work with JsonLogic, returning false when comparison. 

- When we compare two attributes (like: an attribute "color" of label "car" equals to attribute "wheel color" of label "car" ), it shows also objects which are lack of these attributes because formally they are equal in this case (``undefined === undefined``). It is workaround-ed by adding label name with ``&&`` operator to the filter expression. 

- ~If a label/an attribute contains ``.`` in it's name filters don't work~ 
Good news. It was also workaround-ed by replacing all ``.`` characters by another dot-like character (internally only, no any changes for a user).

cvat-ui/src/components/annotation-page/attribute-annotation-workspace/attribute-annotation-sidebar/attribute-annotation-sidebar.tsx
```diff
@@ -2,32 +2,28 @@
...
 import Layout, { SiderProps } from 'antd/lib/layout';
 import { SelectValue } from 'antd/lib/select';
-import { Row, Col } from 'antd/lib/grid';
 import Text from 'antd/lib/typography/Text';
-import { MenuFoldOutlined, MenuUnfoldOutlined } from '@ant-design/icons';
-
-import { ThunkDispatch } from 'utils/redux';
-import { Canvas } from 'cvat-canvas-wrapper';
-import { LogType } from 'cvat-logger';
+import React, { useEffect, useState } from 'react';
```
ActiveChooN, line 9: PR looks good. Only ask to normilize imports here and in `object-list-header.tsx`

ActiveChooN (dismissed):

PR #2921, opened 2021-03-09
Cypress test. Canvas 3D functionality. Basic actions

dvkruchinin:
### Motivation and context
Add Cypress tests.
Used for testing ```.pcd``` files are created using [PCL](https://github.com/PointCloudLibrary/pcl).
Since ```.pcd``` is an image and when interacting with it in the DOM, there are no changes to the classes/attributes of the element, so visual regression testing is used based on comparing images using the ```jimp``` library.

### Checklist

### License

```python
#
```


tests/cypress/integration/canvas3d_functionality/case_56_canvas3d_functionality_basic_actions.js
```diff
@@ -0,0 +1,158 @@
...
+        });
+    }
+
+    function checkFilenameWrapperText(expectedText) {
+        cy.get('.cvat-player-filename-wrapper').should('contain.text', expectedText);
+    }
+
+    function pressKeyPerspectiveDirections(key, screenshotNameBefore, screenshotNameAfter) {
+        cy.get('.cvat-canvas3d-perspective').trigger('mouseover').screenshot(screenshotNameBefore);
+        cy.get('body').type(`{alt}${key}`).wait(300); //Wait to change point cloud position
```
bsekachev, line 26: Shouldn't we use keydown here with corresponding keycode? (as in other tests)
Can we get rid of wait() function? It doesn't look reliable.

tests/cypress/integration/canvas3d_functionality/case_56_canvas3d_functionality_basic_actions.js
```diff
@@ -0,0 +1,158 @@
...
+        cy.get('.cvat-canvas3d-perspective').screenshot(screenshotNameAfter);
+        compareImages(`${screenshotNameBefore}.png`, `${screenshotNameAfter}.png`);
+    }
+
+    function pressKeyArrowDirections(key, screenshotNameBefore, screenshotNameAfter) {
+        cy.get('.cvat-canvas3d-perspective').trigger('mouseover').screenshot(screenshotNameBefore);
+        cy.get('body').type(key).wait(300); //Wait to change point cloud position
```
bsekachev, line 33: The same question

tests/cypress/integration/canvas3d_functionality/case_56_canvas3d_functionality_basic_actions.js
```diff
@@ -0,0 +1,158 @@
...
+        cy.get('.cvat-canvas3d-perspective').screenshot(screenshotNameAfter);
+        compareImages(`${screenshotNameBefore}.png`, `${screenshotNameAfter}.png`);
+    }
+
+    function wheelDirection(element, deltaY, screenshotNameBefore, screenshotNameAfter) {
+        cy.get(element)
+            .screenshot(screenshotNameBefore)
+            .trigger('wheel', { deltaY: deltaY })
+            .wait(300) // Wait change points cloud position
```
bsekachev, line 42: And here

tests/cypress/integration/canvas3d_functionality/case_56_canvas3d_functionality_basic_actions.js
```diff
@@ -0,0 +1,158 @@
...
+                    cy.get('[aria-label="camera"]').should('exist').and('be.visible');
+                });
+        });
+
+        it('Check workspace selector.', () => {
+            cy.get('.cvat-workspace-selector').should('contain.text', 'Standard 3D').click();
+            cy.get('.cvat-workspace-selector-dropdown')
+                .not('.ant-select-dropdown-hidden')
+                .within(() => {
+                    cy.get('[title="Standard 3D"]').should('have.class', 'ant-select-item-option-active');
```
bsekachev, line 102: Let's not to use ant classes. We can add classes on our side if necessary. Can you do it or do you need any help?

tests/cypress/integration/canvas3d_functionality/case_56_canvas3d_functionality_basic_actions.js
```diff
@@ -0,0 +1,158 @@
...
+        cy.compareImages(`${screenshotsPath}/${imgBefore}`, `${screenshotsPath}/${imgAfter}`).then((diffPercent) => {
+            expect(diffPercent).to.be.gt(0);
+        });
+    }
+
+    function checkFilenameWrapperText(expectedText) {
+        cy.get('.cvat-player-filename-wrapper').should('contain.text', expectedText);
+    }
+
+    function pressKeyPerspectiveDirections(key, screenshotNameBefore, screenshotNameAfter) {
```
bsekachev, line 24: The function name does not correspond to the things the function actually does.
I would suggest renaming it to ``testPerspectiveChangeOnKeyPress`` for example.

tests/cypress/integration/canvas3d_functionality/case_56_canvas3d_functionality_basic_actions.js
```diff
@@ -0,0 +1,158 @@
...
+        cy.get('.cvat-canvas3d-perspective').trigger('mouseover').screenshot(screenshotNameBefore);
+        cy.get('body').type(`{alt}${key}`).wait(300); //Wait to change point cloud position
+        cy.get('.cvat-canvas3d-perspective').screenshot(screenshotNameAfter);
+        compareImages(`${screenshotNameBefore}.png`, `${screenshotNameAfter}.png`);
+    }
+
+    function pressKeyArrowDirections(key, screenshotNameBefore, screenshotNameAfter) {
```
bsekachev, line 31: The same comment about the function name

tests/cypress/integration/canvas3d_functionality/case_56_canvas3d_functionality_basic_actions.js
```diff
@@ -0,0 +1,158 @@
...
+        'cypress/screenshots/canvas3d_functionality/case_56_canvas3d_functionality_basic_actions.js';
+
+    function compareImages(imgBefore, imgAfter) {
+        cy.compareImages(`${screenshotsPath}/${imgBefore}`, `${screenshotsPath}/${imgAfter}`).then((diffPercent) => {
+            expect(diffPercent).to.be.gt(0);
+        });
+    }
+
+    function checkFilenameWrapperText(expectedText) {
+        cy.get('.cvat-player-filename-wrapper').should('contain.text', expectedText);
```
bsekachev, line 21: Do we really need this function that consist of one line?

tests/cypress/integration/canvas3d_functionality/case_56_canvas3d_functionality_basic_actions.js
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    function pressKeyArrowDirections(key, screenshotNameBefore, screenshotNameAfter) {
+        cy.get('.cvat-canvas3d-perspective').trigger('mouseover').screenshot(screenshotNameBefore);
+        cy.get('body').type(key).wait(300); //Wait to change point cloud position
+        cy.get('.cvat-canvas3d-perspective').screenshot(screenshotNameAfter);
+        compareImages(`${screenshotNameBefore}.png`, `${screenshotNameAfter}.png`);
+    }
+
+    function wheelDirection(element, deltaY, screenshotNameBefore, screenshotNameAfter) {
```
bsekachev, line 38: The same comment about the function name

bsekachev (changes requested):
Will the tests for context images be written later?

dvkruchinin:
> Will the tests for context images be written later?

I can add this test, but it will have to be skipped. Because I never managed to display the context image. Maybe it's a bug. I was passing on information @vnishukov . 
For example if create a task use ```cvat/apps/engine/tests/assets/test_pointcloud_pcd.zip``` than a context image not displayed.
But displayed if use the ```cvat/apps/engine/tests/assets/test_velodyne_points.zip```.

bsekachev:
> > Will the tests for context images be written later?
> 
> I can add this test, but it will have to be skipped. Because I never managed to display the context image. Maybe it's a bug. I was passing on information @vnishukov .

Hi, @vnishukov . Have you discussed the issue on today's meeting?

@dvkruchinin In this case let's wait till it will be fixed.

bsekachev:
@dvkruchinin 

Could you please resolve my comments in this PR? After that we will merge it. Tests for context image are postponed.

tests/cypress/integration/canvas3d_functionality/case_56_canvas3d_functionality_basic_actions.js
dvkruchinin: You don't have to use this for type (). Since this function passes kyeCode. For example ```Details: "{ code: AltLeft, which: 18 }"```
Cypress executes the command too quickly, and without using ```wait()```, the points cloud changes do not have time to occur. In this case, there will be no percentage change in pixels when comparing images.

tests/cypress/integration/canvas3d_functionality/case_56_canvas3d_functionality_basic_actions.js
dvkruchinin: Answered above.

tests/cypress/integration/canvas3d_functionality/case_56_canvas3d_functionality_basic_actions.js
dvkruchinin: Answered above.

tests/cypress/integration/canvas3d_functionality/case_56_canvas3d_functionality_basic_actions.js
dvkruchinin: I agree that it is better not to use them. I can spend some time adding additional classes. But I'll probably need some help.

tests/cypress/integration/canvas3d_functionality/case_56_canvas3d_functionality_basic_actions.js
bsekachev: > Cypress executes the command too quickly, and without using wait(), the points cloud changes do not have time to occur. In this case, there will be no percentage change in pixels when comparing images.

Can we wrap the following commands to setTimeout without exact timeout value?

Theoretically event ``cy.get('body').type(`{alt}${key}`)`` generates a macrotask which will be done after current stack execution. If we wrap the next instructions to setTimeout, it also generates a macrotask which will be done after the previous one. 
```
setTimeout(() => {
    cy.get('.cvat-canvas3d-perspective').screenshot(screenshotNameAfter);
    compareImages(`${screenshotNameBefore}.png`, `${screenshotNameAfter}.png`);
})
```

dvkruchinin: In this case, I get an error.
```
Error:                       CypressError: The following error originated from your test code, not from Cypress.
  > Cypress detected that you returned a promise from a command while also invoking one or more cy commands in that promise.
The command that returned the promise was:
  > `cy.screenshot()`
The cy command you invoked inside the promise was:
  > `cy.get()`
Because Cypress commands are already promise-like, you don't need to wrap them or return your own promise.
Cypress will resolve your command with whatever the final Cypress command yields.
The reason this is an error instead of a warning is because Cypress internally queues commands serially whereas Promises execute as soon as they are invoked. Attempting to reconcile this would prevent Cypress from ever resolving.
```
I will try to regenerate the ```.pcd``` files. Perhaps this will help to solve this problem. Since in this view, points cloud is far away in the image and minimal changes are not visible in the screenshot.

tests/cypress/integration/canvas3d_functionality/case_56_canvas3d_functionality_basic_actions.js
dvkruchinin: Renamed

tests/cypress/integration/canvas3d_functionality/case_56_canvas3d_functionality_basic_actions.js
dvkruchinin: Renamed

tests/cypress/integration/canvas3d_functionality/case_56_canvas3d_functionality_basic_actions.js
dvkruchinin: Function removed.

tests/cypress/integration/canvas3d_functionality/case_56_canvas3d_functionality_basic_actions.js
dvkruchinin: Renamed.

tests/cypress/integration/canvas3d_functionality/case_56_canvas3d_functionality_basic_actions.js
dvkruchinin: To exclude the use of ```wait()```, I replaced it with an additional ```click()``` on the element. For additional time, change the orientation of points cloud.

tests/cypress/integration/canvas3d_functionality/case_56_canvas3d_functionality_basic_actions.js
dvkruchinin: To exclude the use of these classes, I reworked the attempts to click on the blocked menu items and check the selector text. It should not change at the same time.

bsekachev:
Locally I always get failed the latest scenario.
Do you have any thoughts?

Screenshot from 2021-03-15 18-46-48

tests/cypress/integration/canvas3d_functionality/case_56_canvas3d_functionality_basic_actions.js
bsekachev: Tested it locally and I do not have any issues if I remove both wait() and click(). It is quite strange.

bsekachev:
After a lot of testing I have came around another stable error (the previous has gone and I have not changed code anyhow)

Screenshot from 2021-03-15 20-58-58

dvkruchinin:
Without waiting, I get an error comparing images only on the wheel. I think we can't do without wait() here, since cypress is waiting for the wheel to execute, but the image does not change points cloud immediately. And even at the time of creating the screenshot, points cloud is still moving. Although the wheel event is already over at this time.
I think it will be easier for us to meet to make a common decision.

dvkruchinin:
I've reworked it a bit ```.pcd``` files. Added points cloud to 100,000 points for a clearer display. I also replaced ```wait()``` for wheel with top/side/front for a loop. This gives a more reliable display of the image change.

bsekachev:
> I think we can't do without wait() here, since cypress is waiting for the wheel to execute, but the image does not change points cloud immediately

Cypress ``trigger()`` creates a synthetic event and it calls corresponding listeners (looks like it works in a sync way, try to perform simple example below in a browser console)

```js
window.document.addEventListener('wheel', () => { alert('wheel listener') }) // second
alert('a') // first
window.document.dispatchEvent(new WheelEvent('wheel'))
alert('b') // third
```

But according to the profiler, the browser performs requestAnimationFrame callbacks when IDLE to redraw canvas content and I do not know ways to perform any code after (setTimeout, requestIdleCallback do not work in Cypress tests). 

Cypress also says following:

> However, most of the time you don’t even have to worry about animations. Why not? Cypress will automatically wait for elements to stop animating prior to interacting with them via action commands like .click() or .type(). 
https://docs.cypress.io/faq/questions/using-cypress-faq.html#Some-of-my-elements-animate-in-how-do-I-work-around-that

So, itself it should work out with animations, but I am not sure if it only about css animations (transitions) or also about JS animations (requestAnimationFrame). I would make a suggestions that it is only about the first one. Probably asking community or developers is a good idea. Maybe them would help us with an idea how better to work with it.

They are just my thoughts after some investigation. Speaking about the PR, let's merge it now as is (I tried to run it several times and it worked). If it will be wrong later, we can just skip it and improve.

bsekachev (approved):

PR #2933, opened 2021-03-10
Rework Cypress test "Register user, change password, login with new password".

dvkruchinin:
### Motivation and context
The test uses the credential generation function. It may match so that the generated password will not meet the CVAT requirements. In this case, the test will fail.

### Checklist

### License

```python
#
```

### DIFF
--- a/tests/cypress/support/commands.js
+++ b/tests/cypress/support/commands.js
@@ -53,7 +53,7 @@ Cypress.Commands.add('deletingRegisteredUsers', (accountToDelete) => {
     }).then((responce) => {
         const authKey = responce['body']['key'];
         cy.request({
-            url: '/api/v1/users',
+            url: '/api/v1/users?page_size=all',
             headers: {
                 Authorization: `Token ${authKey}`,
             },
--- a/tests/cypress/support/commands.js
+++ b/tests/cypress/support/commands.js
@@ -41,6 +41,40 @@ Cypress.Commands.add('userRegistration', (firstName, lastName, userName, emailAd
     }
 });
 
+Cypress.Commands.add('deletingRegisteredUsers', () => {
+    cy.request({
+        method: 'POST',
+        url: '/api/v1/auth/login',
+        body: {
+            username: Cypress.env('user'),
+            email: Cypress.env('email'),
+            password: Cypress.env('password'),
+        },
+    }).then(async (responce) => {
+        const authKey = await responce['body']['key'];
+        cy.request({
+            url: '/api/v1/users',
+            headers: {
+                Authorization: `Token ${authKey}`,
+            },
+        }).then(async (responce) => {
+            const responceResult = await responce['body']['results'];
+            for (let user of responceResult) {
+                let userId = user['id'];
+                if (userId !== 1) {
+                    cy.request({
+                        method: 'DELETE',
+                        url: `/api/v1/users/${userId}`,
+                        headers: {
+                            Authorization: `Token ${authKey}`,
+                        },
+                    });
+                }
+            }
+        });
+    });
+});
+
 Cypress.Commands.add(
     'createAnnotationTask',
     (
--- a/tests/cypress/support/commands.js
+++ b/tests/cypress/support/commands.js
@@ -41,7 +41,7 @@ Cypress.Commands.add('userRegistration', (firstName, lastName, userName, emailAd
     }
 });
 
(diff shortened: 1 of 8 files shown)


bsekachev:
@dvkruchinin 


Hi, what do you think about my proposal on email? (removing created users after the test)
And can we do it everytime regardless on test passing status (I mean remove all created users anyway if test was failed or not, like ``finally`` works in JS)

dvkruchinin:
@bsekachev 
Unfortunately, cypress currently does not have full functionality for running commands after a failed test. There is an experimental functionality. But I'm not sure we should use it right now. I added the ability to delete user accounts if the test uses user registration. Perhaps this option will add convenience to testing.
In any case, can use spec ```tests/cypress/integration/remove_users_tasks_projects.js``` to delete user accounts, created tasks, or projects.

tests/cypress/support/commands.js
```diff
@@ -41,6 +41,40 @@ Cypress.Commands.add('userRegistration', (firstName, lastName, userName, emailAd
...
+                let userId = user['id'];
+                if (userId !== 1) {
+                    cy.request({
+                        method: 'DELETE',
+                        url: `/api/v1/users/${userId}`,
+                        headers: {
+                            Authorization: `Token ${authKey}`,
+                        },
+                    });
+                }
```
bsekachev, line 72: I am not sure this is the best idea.

Locally we usually have different users and these users created some tasks, projects and annotations.
Running this test once we will loose all the data in such approach. I believe created users should be a parameter of this command.

bsekachev (changes requested):

tests/cypress/support/commands.js
```diff
@@ -41,6 +41,40 @@ Cypress.Commands.add('userRegistration', (firstName, lastName, userName, emailAd
...
+    cy.request({
+        method: 'POST',
+        url: '/api/v1/auth/login',
+        body: {
+            username: Cypress.env('user'),
+            email: Cypress.env('email'),
+            password: Cypress.env('password'),
+        },
+    }).then(async (responce) => {
+        const authKey = await responce['body']['key'];
```
bsekachev, line 54: Is this operation really ``async``? Should we use ``await``?

tests/cypress/support/commands.js
```diff
@@ -41,6 +41,40 @@ Cypress.Commands.add('userRegistration', (firstName, lastName, userName, emailAd
...
+        cy.request({
+            url: '/api/v1/users',
+            headers: {
+                Authorization: `Token ${authKey}`,
+            },
+        }).then(async (responce) => {
+            const responceResult = await responce['body']['results'];
```
bsekachev, line 61: The same question

bsekachev, line 61: Probably we should use linear promise chain here (now you use nested promise handlers)):

```js
cy.request({
       ....
    }).then(() => {
       .....
       return cy.request({ ... })  
    }).then(() => {
       .......
       delete users in circle
    })
```

tests/cypress/support/commands.js
dvkruchinin: Oh, I understood. Modified the command to use the username to delete as a parameter.

tests/cypress/support/commands.js
dvkruchinin: Yes, this is not necessary here.

tests/cypress/support/commands.js
dvkruchinin: Answered above.

dvkruchinin: Rework in a slightly different form.

tests/cypress/support/commands.js
```diff
@@ -41,6 +41,43 @@ Cypress.Commands.add('userRegistration', (firstName, lastName, userName, emailAd
...
+        url: '/api/v1/auth/login',
+        body: {
+            username: Cypress.env('user'),
+            email: Cypress.env('email'),
+            password: Cypress.env('password'),
+        },
+    }).then((responce) => {
+        const authKey = responce['body']['key'];
+        cy.request({
+            url: '/api/v1/users',
```
bsekachev, line 56: Only ten results will be returned. Use ``?page_size=all`` to get all the users.

dvkruchinin: Added.

bsekachev (approved):

PR #3056, opened 2021-04-02
Az/import export tasks

azhavoro:
### Checklist

### License

```python
#
```


cvat/apps/engine/backup.py
```diff
@@ -0,0 +1,354 @@
...
+                arcname = os.path.normpath(os.path.join(target_dir, os.path.relpath(root, source_dir), filename))
+                zip_object.write(filename=file_path, arcname=arcname)
+
+    def _write_data(self, zip_object):
+        data_dir = self._db_data.get_data_dirname()
+        self._write_directory(source_dir=data_dir, zip_object=zip_object, target_dir=self.DATA_DIRNAME)
+
+    def _write_task(self, zip_object):
+        task_dir = self._db_task.get_task_dirname()
+        self._write_directory(source_dir=task_dir, zip_object=zip_object, target_dir=self.TASK_DIRNAME, exclude=('artifacts', 'export_cache', 'logs'))
```
zhiltsov-max, line 48: Maybe we need to copy only specific dirs instead of excluding few names?
Maybe we need to take specific names from the task (eg. `task.get_task_dirname()`)?

cvat/apps/engine/backup.py
```diff
@@ -0,0 +1,354 @@
...
+        manifest['data']['frames'] = frame_meta_serializer.data
+        for segment in segments:
+            for job in segment['jobs']:
+                job['annotations'] = job_annotations[job['id']]
+        manifest['segments'] = segments
+
+        zip_object.writestr(self.MANIFEST_FILENAME, data=JSONRenderer().render(manifest))
+
+    def export_to(self, filename):
+        with ZipFile(filename, 'x') as output_file:
```
zhiltsov-max, line 102: > 'x': open for exclusive creation, failing if the file already exists

Why do we need to fail instead of overwriting?

cvat/apps/engine/backup.py
```diff
@@ -0,0 +1,354 @@
...
+        version = self._manifest.pop('version')
+        try:
+            return self.Version(version)
+        except ValueError:
+            raise ValueError('{} version is not supported'.format(version))
+
+    def _prepare_meta(self, allowed_keys, meta):
+        for key in list(meta.keys()):
+            if key not in allowed_keys:
+                del meta[key]
```
zhiltsov-max, line 134: I suggest to report dropped fields.

cvat-core/src/server-proxy.js
```diff
@@ -465,6 +465,63 @@
                 });
             }
 
+            async function backupTask(id) {
```
bsekachev, line 468: Why use "backup" name anywhere in the PR? Opposite of "import" is "export" (as in PR title)

cvat-ui/src/actions/tasks-actions.ts
```diff
@@ -213,6 +219,55 @@ export function loadAnnotationsAsync(
...
+
+    return action;
+}
+
+function importTaskSuccess(taskId: number): AnyAction {
+    const action = {
+        type: TasksActionTypes.IMPORT_TASK_SUCCESS,
+        payload: {
+            taskId,
+            importing: false,
```
bsekachev, line 237: literal values should not be a part of payload because we can specify them in reducer based on action type

cvat-ui/src/actions/tasks-actions.ts
```diff
@@ -213,6 +219,55 @@ export function loadAnnotationsAsync(
...
+
+    return action;
+}
+
+function importTaskFailed(error: any): AnyAction {
+    const action = {
+        type: TasksActionTypes.IMPORT_TASK_FAILED,
+        payload: {
+            error,
+            importing: false,
```
bsekachev, line 249: literal values should not be a part of payload because we can specify them in reducer based on action type

cvat-ui/src/actions/tasks-actions.ts
```diff
@@ -213,6 +219,55 @@ export function loadAnnotationsAsync(
     };
 }
 
+function importTask(): AnyAction {
+    const action = {
+        type: TasksActionTypes.IMPORT_TASK,
+        payload: {},
+        importing: true,
```
bsekachev, line 226: literal values should not be a part of payload because we can specify them in reducer based on action type

cvat-ui/src/actions/tasks-actions.ts
```diff
@@ -213,6 +219,55 @@ export function loadAnnotationsAsync(
...
+}
+
+export function importTaskAsync(
+    file: File,
+): ThunkAction<Promise<void>, {}, {}, AnyAction> {
+    return async (dispatch: ActionCreator<Dispatch>): Promise<void> => {
+        try {
+            dispatch(importTask());
+            let taskInstance = new cvat.classes.Task({});
+            taskInstance = await taskInstance.import(file);
```
bsekachev, line 263: Should not .import() be a static method of Task class?

cvat-ui/src/actions/tasks-actions.ts
```diff
@@ -267,6 +322,57 @@ export function exportDatasetAsync(task: any, exporter: any): ThunkAction<Promis
...
+        type: TasksActionTypes.BACKUP_TASK_SUCCESS,
+        payload: {
+            taskID,
+        },
+    };
+
+    return action;
+}
+
+function backupTaskFailed(taskID: any, error: any): AnyAction {
```
bsekachev, line 347: taskID should be a number
error should be an instance of Error

cvat-ui/src/actions/tasks-actions.ts
```diff
@@ -267,6 +322,57 @@ export function exportDatasetAsync(task: any, exporter: any): ThunkAction<Promis
...
+        type: TasksActionTypes.BACKUP_TASK,
+        payload: {
+            taskID,
+        },
+    };
+
+    return action;
+}
+
+function backupTaskSuccess(taskID: any): AnyAction {
```
bsekachev, line 336: taskID should be a number

cvat-ui/src/reducers/tasks-reducer.ts
```diff
@@ -238,6 +240,52 @@ export default (state: TasksState = defaultState, action: AnyAction): TasksState
                 },
             };
         }
+        case TasksActionTypes.BACKUP_TASK: {
+            const { taskID } = action.payload;
+            const { backups } = state.activities;
+
+            backups[taskID] = true;
```
bsekachev, line 247: let's avoid changing current state in this way, in common case we do not know what code refers to the object

cvat-ui/src/reducers/tasks-reducer.ts
```diff
@@ -238,6 +240,52 @@ export default (state: TasksState = defaultState, action: AnyAction): TasksState
...
+
+            return {
+                ...state,
+                activities: {
+                    ...state.activities,
+                    backups: {
+                        ...backups,
```
bsekachev, line 254: Try something like this instead of changing current state

``backups: { ...backups, ...Object.fromEntries([[taskID, true]]) })``

cvat-ui/src/reducers/tasks-reducer.ts
```diff
@@ -238,6 +240,52 @@ export default (state: TasksState = defaultState, action: AnyAction): TasksState
...
+                    },
+                },
+            };
+        }
+        case TasksActionTypes.BACKUP_TASK_FAILED:
+        case TasksActionTypes.BACKUP_TASK_SUCCESS: {
+            const { taskID } = action.payload;
+            const { backups } = state.activities;
+
+            delete backups[taskID];
```
bsekachev, line 264: The same comment.

cvat-ui/src/reducers/tasks-reducer.ts
```diff
@@ -238,6 +240,52 @@ export default (state: TasksState = defaultState, action: AnyAction): TasksState
...
+
+            return {
+                ...state,
+                activities: {
+                    ...state.activities,
+                    backups: {
+                        ...backups,
```
bsekachev, line 271: I would suggest to try something like this

```
import { omit } from 'lodash';
backups: omit(backups, [taskID]);
```

cvat-ui/src/components/actions-menu/actions-menu.tsx
```diff
@@ -127,6 +131,10 @@ export default function ActionsMenuComponent(props: Props): JSX.Element {
             <Menu.Item disabled={inferenceIsActive} key={Actions.RUN_AUTO_ANNOTATION}>
                 Automatic annotation
             </Menu.Item>
+            <Menu.Item key={Actions.BACKUP_TASK}>
+                {backupIsActive && <LoadingOutlined style={{ marginLeft: 10 }} />}
```
bsekachev, line 135: Can we put static styles to ``.scss`` files?

cvat-ui/src/components/tasks-page/top-bar.tsx
```diff
@@ -37,6 +41,29 @@ export default function TopBarComponent(props: VisibleTopBarProps): JSX.Element
...
+                        <Button
+                            size='large'
+                            id='cvat-import-task-button'
+                            type='primary'
+                            disabled={taskImporting}
+                            icon={<UploadOutlined />}
+                        >
+
+                            Import Task
+                            {taskImporting && <LoadingOutlined style={{ marginLeft: 10 }} />}
```
bsekachev, line 63: Can we put static styles in .scss?

bsekachev:
Tried this patch also. Some found issues:

- Shifted layout for top bar. Also need to add space between Import and create.
Screenshot from 2021-04-07 10-29-32

- Since import can be long procedure, probably we need to add a notification popup with offer to open imported task (as in case with created tasks)

- Unable to open imported video/3d task.
```
Error during fetching a job
TypeError: Reduce of empty array with no initial value
```

- 3D task importing issue: The task was imported, but rq_default failed. 
- If I provide a wrong archive (for example with annotations), no any error messages on UI and rq_default also failed

nmanovic:
@azhavoro , can we avoid limits 500 images, if we import a task? Does it handle properly?

nmanovic:
@azhavoro , when are we going to support the same functionality for projects?

cvat-ui/src/actions/tasks-actions.ts
```diff
@@ -267,6 +316,57 @@ export function exportDatasetAsync(task: any, exporter: any): ThunkAction<Promis
...
+        try {
+            const url = await taskInstance.export();
+            const downloadAnchor = window.document.getElementById('downloadAnchor') as HTMLAnchorElement;
+            downloadAnchor.href = url;
+            downloadAnchor.click();
+        } catch (error) {
+            dispatch(exportTaskFailed(taskInstance.id, error));
+        }
+
+        dispatch(exportTaskSuccess(taskInstance.id));
```
nmanovic, line 366: Probably you can put the line under try. It looks more naturally.

utils/cli/core/definition.py
```diff
@@ -289,4 +289,51 @@ def argparse(s):
...
+    'task_id',
+    type=int,
+    help='task ID'
+)
+export_task_parser.add_argument(
+    'filename',
+    type=str,
+    help='output file'
+)
+export_task_parser.add_argument(
```
nmanovic, line 312: I don't think we need the argument. Let's have something by default.

utils/cli/core/definition.py
```diff
@@ -289,4 +289,51 @@ def argparse(s):
...
+    'import',
+    description='import a CVAT task.'
+)
+import_task_parser.add_argument(
+    'filename',
+    type=str,
+    help='upload file'
+)
+import_task_parser.add_argument(
+    '--import_verification_period',
```
nmanovic, line 334: I don't think we need the argument. Let's have something by default.

utils/cli/core/core.py
```diff
@@ -198,6 +198,54 @@ def tasks_upload(self, task_id, fileformat, filename, **kwargs):
             "with annotation file {} finished".format(filename)
         log.info(logger_string)
 
+    def tasks_export(self, task_id, filename, export_verification_period=3, **kwargs):
+        """ Export and download a whole task """
+        url = self.api.tasks_id(task_id)
+        export_url = url + '?action=export'
+        response = self.session.get(export_url)
```
nmanovic, line 205: Looks like exactly the lines are in the beginning of while loop. Move sleep at the end of while and you can remove these 2 lines.

cvat/apps/engine/views.py
```diff
@@ -40,18 +41,29 @@
...
-)
-from cvat.apps.engine.serializers import (
-    AboutSerializer, AnnotationFileSerializer, BasicUserSerializer,
-    DataMetaSerializer, DataSerializer, ExceptionSerializer,
-    FileInfoSerializer, JobSerializer, LabeledDataSerializer,
-    LogEventSerializer, ProjectSerializer, ProjectSearchSerializer, ProjectWithoutTaskSerializer,
-    RqStatusSerializer, TaskSerializer, UserSerializer, PluginsSerializer, ReviewSerializer,
-    CombinedReviewSerializer, IssueSerializer, CombinedIssueSerializer, CommentSerializer
-)
+from cvat.apps.engine.models import (Comment, DimensionType, Image, Issue, Job,
```
nmanovic, line 44: Could you please revert these changes? Does our pre-commit hook works this way? The indent should be 4 spaces.

cvat/apps/engine/task.py
```diff
@@ -5,28 +5,30 @@
...
+import rq
 from django.conf import settings
 from django.db import transaction
-from distutils.dir_util import copy_tree
 
-from . import models
-from .log import slogger
+from cvat.apps.engine import models
+from cvat.apps.engine.log import slogger
+from cvat.apps.engine.media_extractors import (MEDIA_TYPES, Mpeg4ChunkWriter,
```
nmanovic, line 23: Please revert this changes? The indent should be 4 spaces.

cvat/apps/documentation/user_guide.md
```diff
@@ -284,6 +284,7 @@ Go to the [Django administration panel](http://localhost:8080/admin). There you
       1. Assigned to — is used to assign a task to a person. Start typing an assignee’s name and/or
          choose the right person out of the dropdown list.
     - `Jobs` — is a list of all jobs for a particular task. Here you can find the next data:
```
nmanovic, line 286: I think you want to revert the file. Otherwise it will affect documentation PR.

cvat/apps/dataset_manager/formats/README.md
```diff
@@ -536,14 +536,13 @@ archive.zip/
   - `score` (number) - the annotation `score` field
   - arbitrary attributes - will be stored in the `attributes` annotation section
 
-
-*Note*: there is also a [support for COCO keypoints over Datumaro](https://github.com/openvinotoolkit/cvat/issues/2910#issuecomment-726077582)
```
nmanovic, line 540: OK. Probably it is a good idea to revert all files which you don't want to change.

cvat-core/src/server-proxy.js
```diff
@@ -490,6 +490,59 @@
...
+                    }
+
+                    setTimeout(request);
+                });
+            }
+
+            async function importTask(file) {
+                const { backendAPI } = config;
+
+                let annotationData = new FormData();
```
nmanovic, line 520: What is the reason behind `annotationData` name? Should it be payload or something like that?

cvat/apps/engine/views.py
```diff
@@ -360,20 +372,133 @@ def get_permissions(self):
...
+    def create(self, request):
+        action = self.request.query_params.get('action', None)
+        if action is None:
+            return super().create(request)
+        elif action == 'import':
+            self._validate_task_limit(owner=self.request.user)
+            if 'rq_id' in request.data:
+                rq_id = request.data['rq_id']
+            else:
+                rq_id = "{}@/api/v1/tasks/{}?action_import".format(request.user, uuid.uuid4())
```
nmanovic, line 391: Why uuid here? Dataset export and annotations import/export don't use uuid.

nmanovic: OK. The task doesn't exist here. You are  trying to use something here. It is fine.

nmanovic:
It is possible to press `Export Task` multiple times. Sometimes it can lead to strange empty files.

nmanovic:
Let's clean up all temporary files from the archive. For example, index.json and preview.jpg. All these files are generated now during tasks creation. They can be changed in the future. At the same time manifest.jsonln looks like a stable public API and we can keep it inside the archive.

nmanovic:
manifest.json has a lot of fields which probable has to be removed. Let's discuss.

[manifest.json.txt](https://github.com/openvinotoolkit/cvat/files/6296950/manifest.json.txt)

- Do you think annotations should be in a separate file? We have a well defined method to import/export annotations. Why should we keep them in manifest.json? One more annotation format?
- Could you please analyze the manifest.json and say which fields are absolutely necessary to re-create the task correctly on another instance? I don't understand why we need `id`, `project_id`, `mode`, `owner`, `assignee`, etc...

nmanovic:
I will vote for the following archive structure:

```
task_archive.zip
|
--- data
    |
    --- <images or video file>
--- task.json
--- annotations.json
```

nmanovic:
@azhavoro , let's update HLD and put it as a markdown document as advanced documentation.

cvat/apps/documentation/backup_hld.md
```diff
@@ -0,0 +1,57 @@
...
+  - endpoint: `/api/v1/tasks/{id}?action=export​`
+  - method: `GET`
+  - returns: zip archive
+
+  The zip archive has the following structure:
+
+  ```
+   .
+   ├── data
+   │   ├── compressed
```
nmanovic, line 25: Does it really have compressed and original directories?

nmanovic:
@dmitriikhurtin , could you please look at artifacts for remark linter? Can we have more information about problems in markdown files?

nmanovic:
@azhavoro , could you please attach export archives for a small task with images and a small task with a video file? I can see some fields which I think can  be omitted. Let's discuss.

dmitriikhurtin:
I think markdown file is correct. Check script works wrong. If you look at remark report, you can see something  like _[{"path":"cvat/apps/documentation/backup_hld.md","cwd":"/home/runner/work/cvat/cvat","history":["cvat/apps/documentation/backup_hld.md"],"messages":[]}]_. 
The "messages" field contains empty array. It means there aren't errors, but check script call following code
_get_report=`cat ./remark_report/remark_report.json  | jq -r '.[]'`_
and 
_if [[ ! -z ${get_report} ]]; then_
is TRUE.

    _npx remark --report json --no-stdout ${changed_files_remark} 2> ./remark_report/remark_report.json
    get_report=`cat ./remark_report/remark_report.json  | jq -r '.[]'`
     if [[ ! -z ${get_report} ]]; then
      pip install json2html
      python ./tests/json_to_html.py ./remark_report/remark_report.json
      exit 1
    fi_

I suggest to add remark option -q (output only warnings and errors)
 __npx remark -q --report json --no-stdout ${changed_files_remark} 2> ./remark_report/remark_report.json_

azhavoro:
> @azhavoro , could you please attach export archives for a small task with images and a small task with a video file? I can see some fields which I think can be omitted. Let's discuss.

[task_video_backup_2021_04_21_10_49_21.zip](https://github.com/openvinotoolkit/cvat/files/6350249/task_video_backup_2021_04_21_10_49_21.zip)
[task_images_backup_2021_04_09_11_34_10.zip](https://github.com/openvinotoolkit/cvat/files/6350251/task_images_backup_2021_04_09_11_34_10.zip)

cvat/apps/documentation/backup_hld.md
azhavoro: Yes, if the task was created with `use_cache=false` option. In some cases the task does not contain raw data and we cannot recreate it from raw. Anyway, I think that recreating the task from the raw data could lead to potential problems (e.g. different versions of ffmpeg).

nmanovic:
@dvkruchinin , could you please fix the script in this case?

nmanovic:
@azhavoro , the list of questions from me below:
- manifest.jsonl has information about images. Should we duplicate it inside tasks.json? Can we omit `"frames"`?
- What is the value to have `raw` directory inside `data` directory? When will it help?
- Can we omit `"mode": "annotation"`?
- Can we omit `"overlap": 0`?
- Can we omit `"segment_size": 0`?
- Can we omit ` "dimension": "2d"`?
- Can we omit `"version": "v1"`? I don't remember how we are using it.
- Can we have `chunk_type` instead of `compressed_chunk_type` and `original_chunk_type`?
- Can we remove `client_files`, `server_files`, `remote_files`?
- Can we remove `use_zip_chunks`? Does `chunk_type` cover the field?
- Can we remove `copy_data`?
- Can we remove `use_cache` if we have `storage_method`?
- Can we omit `storage`?
- In the future segments concept will be removed. It is only one way how we can create multiple jobs. Can we remove the abstructions and keep only array of jobs?
-

dvkruchinin:
@nmanovic 

> could you please fix the script in this case?

I have prepared a PR https://github.com/openvinotoolkit/cvat/pull/3117 for this

azhavoro:
> @azhavoro , the list of questions from me below:


* manifest.jsonl has information about images. Should we duplicate it inside tasks.json? Can we omit `"frames"`? - manifest.jsonl doesn't contain all neccesary info info for 3d tasks
* What is the value to have `raw` directory inside `data` directory? When will it help? - In general, the data directory can contain raw, original, compressed files. If the request was created with cache = True, only raw will be exported. 
* Can we omit `"mode": "annotation"`? - I would prefer keep this field, it looks like this field is only used to highlight the default dump format in the UI, I believe it's impossible to guess this value, because it can be changed in admin view.
* Can we omit `"overlap": 0`? - No, this field value is needed for dump annotation for example
* Can we omit `"segment_size": 0`? - Looks like this field can be omitted.
* Can we omit ` "dimension": "2d"`? - I think no, this value is used in many places
* Can we omit `"version": "v1"`? I don't remember how we are using it. - It's version of exporter/importer format mainly for the future
* Can we have `chunk_type` instead of `compressed_chunk_type` and `original_chunk_type`? - No, it depends on data type and `use_zip_chunks`  option that is not stored in the db
* Can we remove `client_files`, `server_files`, `remote_files`? - Think yes.
* Can we remove `use_zip_chunks`? Does `chunk_type` cover the field? - This field can be omitted, it's side effect from DataSerializer
* Can we remove `copy_data`? - Yes
* Can we remove `use_cache` if we have `storage_method`? - Looks `storage_method` will be enough.
* Can we omit `storage`? - No, discussed with Maria
* In the future segments concept will be removed. It is only one way how we can create multiple jobs. Can we remove the abstructions and keep only array of jobs? - Ok

nmanovic:
* manifest.jsonl has information about images. Should we duplicate it inside tasks.json? Can we omit "frames"?` - manifest.jsonl doesn't contain all neccesary info info for 3d tasks
```
let's implement that for 3d case
```
* What is the value to have raw directory inside data directory? When will it help? - In general, the data directory can contain raw, original, compressed files. If the request was created with cache = True, only raw will be exported.
```
don't export compressed, original. Send exception if raw data is absent. Put vision data directly to data directory
```
* Can we omit "mode": "annotation"? - I would prefer keep this field, it looks like this field is only used to highlight the default dump format in the UI, I believe it's impossible to guess this value, because it can be changed in admin view. 
```
let's remove the field from export/import of a task and create an issue that we need to remove the field globally
```
* Can we omit "overlap": 0? - No, this field value is needed for dump annotation for example
```
let's calculate it using start/stop frames for a job
```
* Can we omit "dimension": "2d"? - I think no, this value is used in many places
```
let's try to predict the field from the data itself
```
* Can we omit "version": "v1"? I don't remember how we are using it. - It's version of exporter/importer format mainly for the future
```
probably better to have major.minor format for the version
```
* Can we have chunk_type instead of compressed_chunk_type and original_chunk_type? - No, it depends on data type and use_zip_chunks option that is not stored in the db
```
original has always "video" type, let's rename compressed_chunk_type to chunk_type
```
* frame filter
```
Instead of frame_filter we should have corresponding information inside dataset manifest.
```

nmanovic:
@azhavoro , could you please check that the latest version of the branch works. I don't see the import/export in UI and also CLI doesn't work. If it is the case, I don't understand why tests are green.

cvat/apps/engine/task.py
```diff
@@ -262,20 +268,18 @@ def _create_thread(tid, data):
...
-                source_files=list(validate_dimension.related_files.keys()),
-                step=db_data.get_frame_step(),
-                start=db_data.start_frame,
-                stop=data['stop_frame'],
-                dimension=DimensionType.DIM_3D,
-
-            )
-            extractor.add_files(validate_dimension.converted_files)
+    validate_dimension.set_path(upload_dir)
+    validate_dimension.validate()
```
Marishka17, line 272: 1) All data on the share point is validated (if a share is used without copying).
2) Accordingly, if there is data without extension, then the task is not created.

**Example data share structure:**

 - sub_images/
   - image_without_extention
   - image_2.jpeg
   - image_3.jpeg
  - **image_1.jpeg (selected)**
 - **image_2.jpeg (selected)**
```
File "/home/maya/Documents/cvat/.env/lib/python3.8/site-packages/rq/worker.py", line 936, in perform_job
rv = job.perform()
File "/home/maya/Documents/cvat/.env/lib/python3.8/site-packages/rq/job.py", line 684, in perform
File "/home/maya/Documents/cvat/.env/lib/python3.8/site-packages/rq/job.py", line 690, in _execute
File "/usr/lib/python3.8/contextlib.py", line 75, in inner
return func(*args, **kwds)
File "/home/maya/Documents/cvat/cvat/apps/engine/task.py", line 272, in _create_thread
validate_dimension.validate()
File "/home/maya/Documents/cvat/cvat/apps/engine/media_extractors.py", line 856, in validate
self.validate_default(root, actual_path, files)
File "/home/maya/Documents/cvat/cvat/apps/engine/media_extractors.py", line 821, in validate_default
pcd_files = self.process_files(root, actual_path, files)
File "/home/maya/Documents/cvat/cvat/apps/engine/media_extractors.py", line 758, in process_files
file_name, file_extension = file.rsplit('.', maxsplit=1)
ValueError: not enough values to unpack (expected 2, got 1)
```

On develop branch such case not reproduced, because the validation is started only for `zip` media type.

cvat/apps/engine/backup.py
```diff
@@ -0,0 +1,400 @@
...
+                )
+
+    def _write_data(self, zip_object):
+        if self._db_data.storage == StorageChoice.LOCAL:
+            self._write_directory(
+                source_dir=self._db_data.get_upload_dirname(),
+                zip_object=zip_object,
+                target_dir=self.DATA_DIRNAME,
+            )
+        else:
```
Marishka17, line 148: could you please add:
```python
elif self._db_data.storage == StorageChoice.SHARE:
    ...
else:
    raise NotImplementedError()
```

utils/cli/core/core.py
```diff
@@ -202,6 +202,52 @@ def tasks_upload(self, task_id, fileformat, filename, **kwargs):
...
+        logger_string = "Task {} has been exported sucessfully. ".format(task_id) +\
+            "to {}".format(os.path.abspath(filename))
+        log.info(logger_string)
+
+    def tasks_import(self, filename, import_verification_period=3, **kwargs):
+        """ Import a task"""
+        url = self.api.tasks + '?action=import'
+        response = self.session.post(
+                url,
+                files={'task_file': open(filename, 'rb')}
```
Marishka17, line 232: Maybe add a handler if such file not exists?

Marishka17 (commented):
Maybe need to update `utils/cli/README.md`

utils/cli/core/core.py
azhavoro: little bit improved, in this case error message will be like this, I think it's enough for user
```
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/343.zip'
```

nmanovic:
@azhavoro , I have found an issue. Multiple times to import the same video task and got the following error a couple of times:

```
2021-05-14 13:00:16,244 DEBG 'runserver' stderr output:
[Fri May 14 13:00:16.243927 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430] [2021-05-14 13:00:16,243] ERROR django.request: Internal Server Error: /api/v1/tasks
[Fri May 14 13:00:16.243959 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430] Traceback (most recent call last):
[Fri May 14 13:00:16.243966 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]   File "/opt/venv/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
[Fri May 14 13:00:16.243972 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]     response = get_response(request)
[Fri May 14 13:00:16.243977 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]   File "/opt/venv/lib/python3.8/site-packages/django/core/handlers/base.py", line 181, in _get_response
[Fri May 14 13:00:16.243983 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]     response = wrapped_callback(request, *callback_args, **callback_kwargs)
[Fri May 14 13:00:16.243989 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]   File "/opt/venv/lib/python3.8/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
[Fri May 14 13:00:16.243994 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]     return view_func(*args, **kwargs)
[Fri May 14 13:00:16.243999 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]   File "/opt/venv/lib/python3.8/site-packages/rest_framework/viewsets.py", line 114, in view
[Fri May 14 13:00:16.244005 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]     return self.dispatch(request, *args, **kwargs)
[Fri May 14 13:00:16.244010 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]   File "/opt/venv/lib/python3.8/site-packages/rest_framework/views.py", line 505, in dispatch
[Fri May 14 13:00:16.244015 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]     response = self.handle_exception(exc)
[Fri May 14 13:00:16.244021 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]   File "/opt/venv/lib/python3.8/site-packages/rest_framework/views.py", line 465, in handle_exception
[Fri May 14 13:00:16.244026 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]     self.raise_uncaught_exception(exc)
[Fri May 14 13:00:16.244031 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]   File "/opt/venv/lib/python3.8/site-packages/rest_framework/views.py", line 476, in raise_uncaught_exception
[Fri May 14 13:00:16.244037 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]     raise exc
[Fri May 14 13:00:16.244041 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]   File "/opt/venv/lib/python3.8/site-packages/rest_framework/views.py", line 502, in dispatch
[Fri May 14 13:00:16.244047 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]     response = handler(request, *args, **kwargs)

2021-05-14 13:00:16,244 DEBG 'runserver' stderr output:
[Fri May 14 13:00:16.244052 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]   File "/opt/venv/lib/python3.8/site-packages/django/utils/decorators.py", line 43, in _wrapper
[Fri May 14 13:00:16.244061 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]     return bound_method(*args, **kwargs)
[Fri May 14 13:00:16.244066 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]   File "/home/django/cvat/apps/engine/views.py", line 402, in create
[Fri May 14 13:00:16.244071 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]     os.close(rq_job.meta['tmp_file_descriptor'])
[Fri May 14 13:00:16.244076 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430] KeyError: 'tmp_file_descriptor'
[Fri May 14 13:00:16.244297 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430] ERROR - 2021-05-14 13:00:16,243 - log - Internal Server Error: /api/v1/tasks
[Fri May 14 13:00:16.244305 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430] Traceback (most recent call last):
[Fri May 14 13:00:16.244310 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]   File "/opt/venv/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
[Fri May 14 13:00:16.244315 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]     response = get_response(request)
[Fri May 14 13:00:16.244320 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]   File "/opt/venv/lib/python3.8/site-packages/django/core/handlers/base.py", line 181, in _get_response
[Fri May 14 13:00:16.244326 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]     response = wrapped_callback(request, *callback_args, **callback_kwargs)
[Fri May 14 13:00:16.244331 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]   File "/opt/venv/lib/python3.8/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
[Fri May 14 13:00:16.244337 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]     return view_func(*args, **kwargs)
[Fri May 14 13:00:16.244342 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]   File "/opt/venv/lib/python3.8/site-packages/rest_framework/viewsets.py", line 114, in view
[Fri May 14 13:00:16.244348 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]     return self.dispatch(request, *args, **kwargs)
[Fri May 14 13:00:16.244353 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]   File "/opt/venv/lib/python3.8/site-packages/rest_framework/views.py", line 505, in dispatch
[Fri May 14 13:00:16.244358 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]     response = self.handle_exception(exc)
[Fri May 14 13:00:16.244363 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]   File "/opt/venv/lib/python3.8/site-packages/rest_framework/views.py", line 465, in handle_exception
[Fri May 14 13:00:16.244368 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]     self.raise_uncaught_exception(exc)
[Fri May 14 13:00:16.244373 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]   File "/opt/venv/lib/python3.8/site-packages/rest_framework/views.py", line 476, in raise_uncaught_exception
[Fri May 14 13:00:16.244379 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]     raise exc
[Fri May 14 13:00:16.244383 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]   File "/opt/venv/lib/python3.8/site-packages/rest_framework/views.py", line 502, in dispatch
[Fri May 14 13:00:16.244389 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]     response = handler(request, *args, **kwargs)
[Fri May 14 13:00:16.244394 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]   File "/opt/venv/lib/python3.8/site-packages/django/utils/decorators.py", line 43, in _wrapper
[Fri May 14 13:00:16.244406 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]     return bound_method(*args, **kwargs)
[Fri May 14 13:00:16.244411 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]   File "/home/django/cvat/apps/engine/views.py", line 402, in create
[Fri May 14 13:00:16.244417 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430]     os.close(rq_job.meta['tmp_file_descriptor'])
[Fri May 14 13:00:16.244422 2021] [wsgi:error] [pid 633:tid 140235986036480] [remote 172.28.0.6:53430] KeyError: 'tmp_file_descriptor'
```

nmanovic:
@azhavoro , also the list of tasks isn't updated. Need to refresh the whole site or go to `projects` and come back to `tasks`.

nmanovic:
@azhavoro , I see great process with the layout of the exported archive. Could we do one more step forward?

- Let's use label's name as its `id` and remove `id: number` from the specification. The same change should work for attributes as well.
- Let's use local index of a job as its `id` and remove `id: number` from the specification
- Let's use local index of a shape as its `id` and remove `id: number` from the specification
- Let's use an array inside annotations.json where first dictionary corresponds to first job, second dictionary corresponds to second job, and so on.

Can it be implemented from your point of view? Will we loose any information?

nmanovic:
Another exception when try to export a task:

```
2021-05-14 13:35:27,778 DEBG 'rqworker_default_0' stderr output:
ERROR - 2021-05-14 13:35:27,778 - worker - Traceback (most recent call last):
  File "/opt/venv/lib/python3.8/site-packages/rq/worker.py", line 936, in perform_job
    rv = job.perform()
  File "/opt/venv/lib/python3.8/site-packages/rq/job.py", line 684, in perform
  File "/opt/venv/lib/python3.8/site-packages/rq/job.py", line 690, in _execute
  File "/home/django/cvat/apps/dataset_manager/views.py", line 114, in backup_task
    task_exporter.export_to(temp_file)
  File "/home/django/cvat/apps/engine/backup.py", line 241, in export_to
    self._write_manifest(output_file)
  File "/home/django/cvat/apps/engine/backup.py", line 215, in _write_manifest
    task['jobs'] = serialize_jobs()
  File "/home/django/cvat/apps/engine/backup.py", line 197, in serialize_jobs
    job_data = self._prepare_job_meta(job_serializer.data)
  File "/opt/venv/lib/python3.8/site-packages/rest_framework/serializers.py", line 562, in data
    ret = super().data
  File "/opt/venv/lib/python3.8/site-packages/rest_framework/serializers.py", line 260, in data
    self._data = self.to_representation(self.instance)
  File "/opt/venv/lib/python3.8/site-packages/rest_framework/serializers.py", line 529, in to_representation
    ret[field.field_name] = field.to_representation(attribute)
  File "/opt/venv/lib/python3.8/site-packages/rest_framework/serializers.py", line 529, in to_representation
    ret[field.field_name] = field.to_representation(attribute)
  File "/opt/venv/lib/python3.8/site-packages/rest_framework/relations.py", line 370, in to_representation
    assert 'request' in self.context, (
AssertionError: `HyperlinkedIdentityField` requires the request in the serializer context. Add `context={'request': request}` when instantiating the serializer.
```

nmanovic:
@azhavoro , also I see "index.json" file which should be created on the fly.

nmanovic:
Now we have tasks.json

```json
{
    "name": "tracking video",
    "bug_tracker": "",
    "status": "annotation",
    "labels": [
        {
            "id": 12,
            "name": "person",
            "color": "#c06060",
            "attributes": []
        }
    ],
    "subset": "",
    "version": "1.0",
    "data": {
        "chunk_size": 72,
        "image_quality": 70,
        "start_frame": 0,
        "stop_frame": 794,
        "storage_method": "file_system",
        "storage": "local",
        "chunk_type": "imageset"
    },
    "jobs": [
        {
            "start_frame": 0,
            "stop_frame": 794,
            "id": 6,
            "status": "annotation"
        }
    ]
}
```

Can we have something like json below?

```
{
    "name": "tracking video",
    "bug_tracker": "",
    "status": "annotation",
    "labels": [
        {
            "name": "person",
            "color": "#c06060",
            "attributes": []
        }
    ],
    "subset": "",
    "version": "1.0",
    "data": {
        "chunk_size": 72,
        "image_quality": 70,
        "start_frame": 0,
        "stop_frame": 794,
        "storage_method": "file_system",
        "storage": "local",
        "chunk_type": "imageset"
    },
    "jobs": [
        {
            "start_frame": 0,
            "stop_frame": 794,
            "status": "annotation"
        }
    ]
}
```

nmanovic:
For annotations.json before:

```json
{
    "16": {
        "version": 3,
        "tags": [],
        "shapes": [
            {
                "type": "rectangle",
                "occluded": false,
                "z_order": 0,
                "points": [
                    971.5869140625,
                    462.3623046875,
                    1020.4619140625,
                    582.5683822631836
                ],
                "frame": 0,
                "label_id": 22,
                "group": 0,
                "source": "manual",
                "attributes": []
            },
            {
                "type": "rectangle",
                "occluded": false,
                "z_order": 0,
                "points": [
                    1190.86328125,
                    590.4931640625,
                    1256.9105758666992,
                    715.9830322265625
                ],
                "frame": 0,
                "label_id": 22,
                "group": 0,
                "source": "manual",
                "attributes": []
            },
            {
                "type": "rectangle",
                "occluded": false,
                "z_order": 0,
                "points": [
                    527.7490234375,
                    516.5205078125,
                    717.9652404785156,
                    721.2671356201172
                ],
                "frame": 0,
                "label_id": 22,
                "group": 0,
                "source": "manual",
                "attributes": []
            }
        ],
        "tracks": []
    }
}
```

And after my suggestions:

```json
[
    {
        "version": 3,
        "tags": [],
        "shapes": [
            {
                "type": "rectangle",
                "occluded": false,
                "z_order": 0,
                "points": [
                    971.5869140625,
                    462.3623046875,
                    1020.4619140625,
                    582.5683822631836
                ],
                "frame": 0,
                "label": "person",
                "group": 0,
                "source": "manual",
                "attributes": []
            },
            {
                "type": "rectangle",
                "occluded": false,
                "z_order": 0,
                "points": [
                    1190.86328125,
                    590.4931640625,
                    1256.9105758666992,
                    715.9830322265625
                ],
                "frame": 0,
                "label": "person",
                "group": 0,
                "source": "manual",
                "attributes": []
            },
            {
                "type": "rectangle",
                "occluded": false,
                "z_order": 0,
                "points": [
                    527.7490234375,
                    516.5205078125,
                    717.9652404785156,
                    721.2671356201172
                ],
                "frame": 0,
                "label": "person",
                "group": 0,
                "source": "manual",
                "attributes": []
            }
        ],
        "tracks": []
    }
]
```

nmanovic:
In tasks.json we have "storage" and "storage_method". Do we really need both?

bsekachev:
If we export a task from a project, we can't import it back to the project. Is it known issue? We just do not have "Import task" button on the project page. The only way is to import and then move the task to the project (using this way in current implementation we lose all the annotated attributes).

``task.json`` contains project field ``subset``. How is it used? 

``django: defaul`` process fails if I send an archive without ``task.json``

Extra offset of "Create new task" button:

Develop:
Screenshot from 2021-05-31 12-53-37

This branch:
Screenshot from 2021-05-31 12-53-20

bsekachev:
I exported a task with a finished job. Imported it and job in annotation status. I suppose the same works for jobs in reviews.

Also video task has overlap (5 default). Export, import it and now it has overlap 0.

bsekachev:
One more possible problem: Looks like issues aren't dumped together with annotations.

nmanovic:
- [ ] index.json should be removed
- [ ] If I try to upload dump of annotations, I get 500 error code. It should be 400.
(Error: Request failed with status code 500. "KeyError: \"There is no item named 'task.json' in the archive\"\n".)

nmanovic:
- [ ] In CVAT for images we have floating point numbers with 2 digits after the dot. Should we apply the same approach here? What do you think? Personally I don't have a strong opinion on the subject. Looks like it is good to have exactly the same as we have in DB. But at the same time for annotation formats we took another decision.

cvat/apps/dataset_manager/views.py
```diff
@@ -97,6 +97,40 @@ def clear_export_cache(task_id, file_path, file_ctime):
...
         raise
 
+def backup_task(task_id, output_path):
+    try:
+        db_task = Task.objects.get(pk=task_id)
+
+        cache_dir = get_export_cache_dir(db_task)
+        output_path = osp.join(cache_dir, output_path)
+
+        task_time = timezone.localtime(db_task.updated_date).timestamp()
```
nmanovic, line 107: Do we need to adjust time for clearning_job somewhere here? I believe we have multiple races here. Several users can try to backup the same task in parallel. Clearning job can remove our file after all our successful checks.

Races are fine if they don't lead to unexpected results.

nmanovic (approved):
Let's merge the branch as is and try to fix remaining issues in the next PR.

cvat/apps/engine/backup.py
```diff
@@ -0,0 +1,550 @@
...
+            db_segments = list(self._db_task.segment_set.all().prefetch_related('job_set'))
+            db_segments.sort(key=lambda i: i.job_set.first().id)
+            db_jobs = (s.job_set.first() for s in db_segments)
+            return db_jobs
+        return ()
+
+class TaskExporter(_TaskBackupBase):
+    def __init__(self, pk, version=Version.V1):
+        self._db_task = models.Task.objects.prefetch_related('data__images').select_related('data__video').get(pk=pk)
+        self._db_data = self._db_task.data
```
bsekachev, line 194: @azhavoro 
Probably there is the same problem like in the recent fix.
select_related(data__..) breaks prefetch_related(data__...)

PR #3070, opened 2021-04-06
Added hotkeys to change labels 

bsekachev:
### Motivation and context
Resolved #2990
Resolved #2317

#### From design
**Task:** Hotkeys for labels.
**Solution:** Add key selector button to labels panel in the left sidebar. Clicking on the button will open popover with number selection  for 0-9. Unpresented labels will be covered with opacity. [Num key + some modifier key] handler will be added to the standard workspace.

I suppose it should change a label of activated object or change a label to create the next object if no any objects activated

Screenshot from 2021-04-06 20-11-24

Screenshot from 2021-04-06 18-24-12

## HOW TO USE?

Activate (or deactivate if need to change default label) object and press Ctrl+Num (0...9)
Default binding can be changed like shown on the screenshot below

### How has this been tested?
Manual testing

### Checklist

- [x] I submit my changes into the `develop` branch
- [x] I have added description of my changes into [CHANGELOG](https://github.com/opencv/cvat/blob/develop/CHANGELOG.md) file
- [ ] I have updated the [documentation](
  https://github.com/opencv/cvat/blob/develop/README.md#documentation) accordingly
- [ ] I have added tests to cover my changes
- [x] I have linked related issues ([read github docs](
  https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword))
- [x] I have increased versions of npm packages if it is necessary ([cvat-canvas](https://github.com/opencv/cvat/tree/develop/cvat-canvas#versioning),
[cvat-core](https://github.com/opencv/cvat/tree/develop/cvat-core#versioning), [cvat-data](https://github.com/opencv/cvat/tree/develop/cvat-data#versioning) and [cvat-ui](https://github.com/opencv/cvat/tree/develop/cvat-ui#versioning))

### License

- [x] I submit _my code changes_ under the same [MIT License](
  https://github.com/opencv/cvat/blob/develop/LICENSE) that covers the project.
  Feel free to contact the maintainers if that's a concern.
- [x] I have updated the license header for each file (see an example below)

```python
# Copyright (C) 2021 Intel Corporation
#
# SPDX-License-Identifier: MIT
```

PR #3071, opened 2021-04-06
Label hotkeys

ActiveChooN:
### Motivation and context

Added hotkeys to change active object label or default label for drawing with numerical keys. Keys for labels could be changed via according buttons on control side bar panel in labels pane.

### How has this been tested?

Manually

### Checklist

### License

```python
#
```


bsekachev:
@ActiveChooN 

Hmm.. What should we do with two PRs implementing the feature in a bit different way?

bsekachev:
Tried this patch. Occasionally get error:
Screenshot from 2021-04-06 18-27-32

But I like this dropdown more:
Screenshot from 2021-04-06 18-30-52

What do you think if I move the dropdown to my version of patch and we will work on it?

ActiveChooN:
> What do you think if I move the dropdown to my version of patch and we will work on it?
@bsekachev, sure. And i want to ask you keep old label item component style, because this style is applicable to object component due to it has normal height, but not to new version, because it is quite too contrast with small size of label item component.

bsekachev:
@ActiveChooN 

I am not sure I understand you clearly, but style of label items was changed to be consistent with object.
Do you think we should not follow this way?

ActiveChooN:
> takes too many space

@bsekachev , i'm agree with you, but i'm not sure about making all background colorful. Guess it can be better to make side color block and keep background color grey.

bsekachev:
These changes moved to #3070

Issue #2473, opened 2020-11-23
Cannot upload YOLO annotations generated from dump annotations


kairikibear:
### My actions before raising this issue
- [V ] Read/searched [the docs](https://github.com/opencv/cvat/tree/master#documentation)
- [V ] Searched [past issues](/issues)

I have a dataset i want to label, but i want to incorporate exisitng labels of the dataset (detection dataset, i want to add attributes)

I experimented to see if it was possible through the upload annotation feature and did exactly this :
1. Created a small toy dataset
2. Labelled the images
3. Saved the annots
4. Dump the annotations to yolo format
5. Upload again with yolo format

And instead of working as intended, i have met with this error : 
![image](https://user-images.githubusercontent.com/70246267/99949507-bab5c980-2dad-11eb-9b1a-41d4a0f98525.png)

I immediately checked the log and traced the error all the way to this [file](https://github.com/openvinotoolkit/datumaro/blob/develop/datumaro/plugins/yolo_format/extractor.py#L156)

in which i surmise the error comes from the absence of the 'size' property of 'image' being None, which is funnily caused by the absence of an images.meta file judging from these [lines](https://github.com/openvinotoolkit/datumaro/blob/develop/datumaro/plugins/yolo_format/extractor.py#L48-L60), documented nowhere. 

I also thought that somehow i did something that makes image info is not passed to the constructor of the extractor, but seems like [this](https://github.com/openvinotoolkit/datumaro/blob/b71610368e7eb7a3b63aea74d228a7e0aca1eddc/datumaro/components/project.py#L318) shows it was not the case.

### Expected Behaviour

The yolo annotation can be uploaded normally without anything on the way.

### Current Behaviour

The error aforementioned

### Possible Solution

Better documentations regarding such subjects if it was NOT documented, or bugfix if it was a bug / error

### Steps to Reproduce (for bugs)

Mentioned above

### Your Environment

This commit hash 3fb3f677ebf95ed90ff404fb9ffe52f3c8fe55b3
Website says version 1.1rc0

zhiltsov-max:
Hi.
It is more of a bug than something undocumented, because this image information is passed [here](https://github.com/openvinotoolkit/cvat/blob/develop/cvat/apps/dataset_manager/formats/yolo.py#L37-L49) and the user is not expected to provide it. Could you help with reproducing this issue? How did you add images to the task?

kairikibear:
[...]

PR #3091, opened 2021-04-15
Fix import annotations for frames with dots in name

kirill-sizov:
### Motivation and context

Fixes https://github.com/openvinotoolkit/cvat/issues/2473

### How has this been tested?
 

Unit test

### Checklist

### License

```python
#
```


cvat/apps/dataset_manager/bindings.py
```diff
@@ -651,7 +652,7 @@ def import_dm_annotations(dm_dataset, task_data):
 
     for item in dm_dataset:
         frame_number = task_data.abs_frame_id(
-            match_dm_item(item, task_data, root_hint=root_hint))
+            match_dm_item(DatasetItem(id=item.id + '.'), task_data, root_hint=root_hint))
```
zhiltsov-max, line 655: This solution does not look right.

zhiltsov-max (changes requested):

cvat/apps/dataset_manager/bindings.py
kirill-sizov: Fixed this solution

cvat/apps/dataset_manager/formats/cvat.py
```diff
@@ -441,7 +441,7 @@ def load(file_object, annotations):
             elif el.tag == 'image':
                 image_is_opened = True
                 frame_id = annotations.abs_frame_id(match_dm_item(
-                    DatasetItem(id=el.attrib['name'],
+                    DatasetItem(id=osp.splitext(el.attrib['name'])[0],
```
zhiltsov-max, line 444: This change will break support for images with same filename and different extension in CVAT format, which is the only format that supports it.

cvat/apps/dataset_manager/tests/test_formats.py
```diff
@@ -496,6 +511,21 @@ def test_frames_outside_are_not_generated(self):
             self.assertTrue(frame.frame in range(6, 10))
         self.assertEqual(i + 1, 4)
 
+    def test_can_import_annotations_for_image_with_dots_in_filename(self):
+        export_formats = [f.DISPLAY_NAME for f in dm.views.get_export_formats()]
+        formats = [f.DISPLAY_NAME for f in dm.views.get_import_formats() if
+            f.DISPLAY_NAME in export_formats]
+        for format_name in formats:
+            if format_name == "VGGFace2 1.0":
+                self.skipTest("Format is disabled")
```
zhiltsov-max, line 520: Why this format is disabled?

cvat/apps/dataset_manager/tests/test_formats.py
```diff
@@ -496,6 +511,21 @@ def test_frames_outside_are_not_generated(self):
             self.assertTrue(frame.frame in range(6, 10))
         self.assertEqual(i + 1, 4)
 
+    def test_can_import_annotations_for_image_with_dots_in_filename(self):
+        export_formats = [f.DISPLAY_NAME for f in dm.views.get_export_formats()]
+        formats = [f.DISPLAY_NAME for f in dm.views.get_import_formats() if
```
zhiltsov-max, line 516: This definition will miss few formats, check this declaration https://github.com/openvinotoolkit/cvat/blob/develop/cvat/apps/engine/tests/test_rest_api.py#L4414

cvat/apps/dataset_manager/tests/test_formats.py
kirill-sizov: I skipped the test for this format because VGGFace does not support annotations that are generated in the `_generate_annotations(...)` function. Same problem with ICDAR formats. I can specify annotations for these formats, but this will lead to a lot of code. So I would prefer to remove this test and include images with dots in filename in this test https://github.com/openvinotoolkit/cvat/blob/4ceb5da2447805c29ca4e20108bb2a2fb31ca6d3/cvat/apps/engine/tests/test_rest_api.py#L2966

zhiltsov-max: Actually, the real problem was that the reason of skipping was not described in tests. Thus, a reader can't judge, whether this test already can be enabled or not.

zhiltsov-max (approved):

PR #3096, opened 2021-04-15
Fix bug with hung web worker

### LINKED ISSUE #3080: UI fails to load previously cached frames when working with large images

### Expected Behaviour

A user can navigate through all frames in a job regardless of image size.

### Current Behaviour

The UI gets stuck in a loading state when trying to go backwards through already loaded frames.

### Possible Solution

This seems to be some issue with retrieving frames that have been cached, because the UI always stops at one of the image chunk points.

Here is a live example uploaded on cvat.org: https://cvat.org/tasks/66868/jobs/97576
1. Create a task with 100 images (all 2048x2448)
1. Open the job and step through all frames
1. Once at the end, step backwards through all frames
1. At some point, the UI shows the loading icon and fails to load any previous images. You can however step back forward, but everything backwards fails to load.

maximusunc:
* Added upper limit to stop property of active chunk request

### Motivation and context

Fixes #3080. The unzip_imgs web worker was not getting terminated when loading the last chunk in a task. I added a stopFrame limit to the chunk request so that the web worker gets properly terminated once it unzips all images.

### How has this been tested?

I manually tested this by uploading 71 frames, all 2048 x 2448 to exceed the cache limit, and made sure that the web worker is correctly terminated after loading the last chunk. I also made sure that scrolling back through the previous frames worked and that frames that had been removed from the cache are successfully loaded back in.

### Checklist

### License

```python
#
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Export of instance masks with holes (<https://github.com/openvinotoolkit/cvat/pull/3044>)
 - Changing a label on canvas does not work when 'Show object details' enabled (<https://github.com/openvinotoolkit/cvat/pull/3084>)
+- Make sure frame unzip web worker correctly terminates after unzipping all images in a requested chunk (<https://github.com/openvinotoolkit/cvat/pull/3096>)
 
 ### Security
 
--- a/cvat-core/package-lock.json
+++ b/cvat-core/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "cvat-core",
-  "version": "3.12.0",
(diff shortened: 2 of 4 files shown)


nmanovic:
@maximusunc , thanks for the contribution. Could you please add information about the fix into CHANGELOG.md?

nmanovic:
@maximusunc , could you please update the version of cvat-core? Run `npm version patch` command.

maximusunc:
Hi @nmanovic Looking at the test output, it looks like the tests timed out while running some canvas3d tests. I don't think my changes had anything to do with this test failure, but I will wait to hear back from you.

nmanovic (approved):
@maximusunc , thanks for the fix. For we it looks fine. All checks for the PR is OK.

PR #3188, opened 2021-05-13
Cypress test. Fix "Changing a default value for an attribute"

dvkruchinin:
### Motivation and context
Any attempts to wait for the attribute value to change to positive failed. I reworked it so that the label change button was visible otherwise I got the ```Detached from the DOM``` error. Perhaps at this time react finishes processing the asynchronous function and the necessary elements get positive values.

### Checklist

### License

```python
#
```

### DIFF
--- a/tests/cypress/integration/actions_tasks3/case_44_changing_default_value_for_attribute.js
+++ b/tests/cypress/integration/actions_tasks3/case_44_changing_default_value_for_attribute.js
@@ -51,11 +51,13 @@ context('Changing a default value for an attribute.', () => {
             cy.get('.cvat-label-constructor-updater').within(() => {
                 cy.get('.cvat-attribute-inputs-wrapper').then((wrapper) => {
                     for (let i = 0; i < wrapper.length; i++) {
+                        cy.task('log', `Before while ${wrapper[i].getAttribute('cvat-attribute-id')}`)
                         // Waiting for "cvat-attribute-id" value be greater then 0
                         while (Number(wrapper[i].getAttribute('cvat-attribute-id')) < 0) {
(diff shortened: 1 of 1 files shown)


tests/cypress/integration/actions_tasks3/case_44_changing_default_value_for_attribute.js
```diff
@@ -46,8 +46,12 @@ context('Changing a default value for an attribute.', () => {
                 cy.contains(new RegExp(`^${additionalLabel}$`))
                     .parents('.cvat-constructor-viewer-item')
                     .find('[aria-label="edit"]')
-                    .click({ force: true });
+                    .should('be.visible')
+                    .then((e) => {
+                        cy.get(e).click();
+                    });
```
bsekachev, line 52: @dvkruchinin 
Could you please explain how does it fix the issue?

dvkruchinin: I think that the issue was due to the fact that the click on the ```[aria-label="edit"]``` was carried out with the argument ```force: true``` and at this time attributes ```cvat-attribute-id``` haven't received positive values yet. When I cleaned up ```force: true``` then I got an error ```Detached from the DOM``` when click on ```[aria-label="edit"]```. Perhaps this is one of the ```react``` reactions. I tested this in my fork for about 20 restarts. Unfortunately, there is no complete certainty that this will completely solve the problem. For a similar issue, the issue is still open in the cypress repository.

bsekachev: Okay, lets look how it will behave in real cases

bsekachev (approved):

PR #3234, opened 2021-05-25
CVAT-3D milestone6

manasars:
Hi @bsekachev , @nmanovic , @zhiltsov-max 

CVAT 3D Milestone 6 changes:

Added support for Dump annotations, Export Annotations and Upload annotations in PCD and Kitti formats.
The code changes are only in 4 files i.e  bindings.py, registry.py and added new datatsets pointcloud.py and velodynepoint.py.

The rest of the files are from M5 base branch, was waiting for it to be merged but since comments are in progress created one for M6.

- We shall add the changes in CHANGELOG after M5 code is merged .
- (https://github.com/opencv/cvat/blob/develop/CHANGELOG.md) file
Datumaro PR - https://github.com/openvinotoolkit/datumaro/pull/245

- [x] I have updated the license header for each file


bsekachev:
Tested the patch roughly. Found issues are:
1. In 2D when you are pasting object holding Ctrl, you can paste copied object several times. It does not work in 3D mode.
2. List of shortcuts shows some shortcuts unapplicable in 3D (switch keyframe, to background, to foreground, next/previous keyframe)
3. Changing color does not work correctly (when Color by Instance is activated). Using both UI and hotkey (Enter). It does not change color of the shape on canvas.
4. The issue with selection appeared again with M6. On M5 selection works when hover an object on the canvas. On M6 you need to click it.

bsekachev:
Also I did ``pip install -r cvat/requirements/development.txt``, but when I am trying to dump annotations/dataset, I get error:

```
raceback (most recent call last):
  File "/home/user/cvat/.env/lib/python3.8/site-packages/rq/worker.py", line 936, in perform_job
    rv = job.perform()
  File "/home/user/cvat/.env/lib/python3.8/site-packages/rq/job.py", line 684, in perform
  File "/home/user/cvat/.env/lib/python3.8/site-packages/rq/job.py", line 690, in _execute
  File "/home/user/cvat/cvat/apps/dataset_manager/views.py", line 84, in export_task_as_dataset
    return export_task(task_id, dst_format, server_url=server_url, save_images=True)
  File "/home/user/cvat/cvat/apps/dataset_manager/views.py", line 61, in export_task
    task.export_task(task_id, temp_file, dst_format,
  File "/home/user/cvat/cvat/apps/dataset_manager/task.py", line 749, in export_task
    task.export(f, exporter, host=server_url, save_images=save_images)
  File "/home/user/cvat/cvat/apps/dataset_manager/task.py", line 631, in export
    exporter(dst_file, task_data, **options)
  File "/home/user/cvat/cvat/apps/dataset_manager/formats/registry.py", line 37, in __call__
    f_or_cls(*args, **kwargs)
  File "/home/user/cvat/cvat/apps/dataset_manager/formats/pointcloud.py", line 31, in _export_images
    dataset = Dataset.from_extractors(CvatTaskDataExtractor(
  File "/home/user/cvat/cvat/apps/dataset_manager/bindings.py", line 530, in __init__
    dm_anno = self._read_cvat_anno(frame_data, task_data)
  File "/home/user/cvat/cvat/apps/dataset_manager/bindings.py", line 665, in _read_cvat_anno
    anno = datumaro.Cuboid(id=anno_id, points=anno_points,
AttributeError: module 'datumaro.components.extractor' has no attribute 'Cuboid'
```

manasars:
> Also I did `pip install -r cvat/requirements/development.txt`, but when I am trying to dump annotations/dataset, I get error:
> 
> ```
> raceback (most recent call last):
>   File "/home/user/cvat/.env/lib/python3.8/site-packages/rq/worker.py", line 936, in perform_job
>     rv = job.perform()
>   File "/home/user/cvat/.env/lib/python3.8/site-packages/rq/job.py", line 684, in perform
>     self._result = self._execute()
>   File "/home/user/cvat/.env/lib/python3.8/site-packages/rq/job.py", line 690, in _execute
>     return self.func(*self.args, **self.kwargs)
>   File "/home/user/cvat/cvat/apps/dataset_manager/views.py", line 84, in export_task_as_dataset
>     return export_task(task_id, dst_format, server_url=server_url, save_images=True)
>   File "/home/user/cvat/cvat/apps/dataset_manager/views.py", line 61, in export_task
>     task.export_task(task_id, temp_file, dst_format,
>   File "/home/user/cvat/cvat/apps/dataset_manager/task.py", line 749, in export_task
>     task.export(f, exporter, host=server_url, save_images=save_images)
>   File "/home/user/cvat/cvat/apps/dataset_manager/task.py", line 631, in export
>     exporter(dst_file, task_data, **options)
>   File "/home/user/cvat/cvat/apps/dataset_manager/formats/registry.py", line 37, in __call__
>     f_or_cls(*args, **kwargs)
>   File "/home/user/cvat/cvat/apps/dataset_manager/formats/pointcloud.py", line 31, in _export_images
>     dataset = Dataset.from_extractors(CvatTaskDataExtractor(
>   File "/home/user/cvat/cvat/apps/dataset_manager/bindings.py", line 530, in __init__
>     dm_anno = self._read_cvat_anno(frame_data, task_data)
>   File "/home/user/cvat/cvat/apps/dataset_manager/bindings.py", line 665, in _read_cvat_anno
>     anno = datumaro.Cuboid(id=anno_id, points=anno_points,
> AttributeError: module 'datumaro.components.extractor' has no attribute 'Cuboid'
> ```

@bsekachev : Are you trying Dump Annotations for 3D in the same branch , kindly confirm

bsekachev:
> @bsekachev : Are you trying Dump Annotations for 3D in the same branch , kindly confirm

Actually I used cvat-3D-M6, now I see cvat-3D-Milstone6 is the right

bsekachev:
@manasars 

I've changed the branch, and still get similar exception, but a bit another.

```
  File "/home/user/cvat/utils/dicom_converter/.env/lib/python3.8/site-packages/rq/worker.py", line 936, in perform_job
    rv = job.perform()
  File "/home/user/cvat/utils/dicom_converter/.env/lib/python3.8/site-packages/rq/job.py", line 684, in perform
  File "/home/user/cvat/utils/dicom_converter/.env/lib/python3.8/site-packages/rq/job.py", line 690, in _execute
  File "/home/user/cvat/cvat/apps/dataset_manager/views.py", line 87, in export_task_annotations
    return export_task(task_id, dst_format, server_url=server_url, save_images=False)
  File "/home/user/cvat/cvat/apps/dataset_manager/views.py", line 61, in export_task
    task.export_task(task_id, temp_file, dst_format,
  File "/home/user/cvat/cvat/apps/dataset_manager/task.py", line 749, in export_task
    task.export(f, exporter, host=server_url, save_images=save_images)
  File "/home/user/cvat/cvat/apps/dataset_manager/task.py", line 631, in export
    exporter(dst_file, task_data, **options)
  File "/home/user/cvat/cvat/apps/dataset_manager/formats/registry.py", line 37, in __call__
    f_or_cls(*args, **kwargs)
  File "/home/user/cvat/cvat/apps/dataset_manager/formats/pointcloud.py", line 21, in _export_images
    dataset = Dataset.from_extractors(CvatTaskDataExtractor(
  File "/home/user/cvat/cvat/apps/dataset_manager/bindings.py", line 535, in __init__
    dm_anno = self._read_cvat_anno(frame_data, task_data)
  File "/home/user/cvat/cvat/apps/dataset_manager/bindings.py", line 664, in _read_cvat_anno
    anno = datumaro.Cuboid3D(id=anno_id, points=anno_points,
AttributeError: module 'datumaro.components.extractor' has no attribute 'Cuboid3D'
```
What is wrong? Of course I did ``pip install`` again and restarted the server.

Issue with shape selection fixed in this branch.

nmanovic:
@manasars ,

Could you please fix tests?

```
======================================================================
960FAIL: test_stats (tests.test_ops.TestOperations)
961----------------------------------------------------------------------
962Traceback (most recent call last):
963  File "/home/travis/build/openvinotoolkit/datumaro/tests/test_ops.py", line 139, in test_stats
964    self.assertEqual(expected, actual)
965AssertionError: {'ima[265 chars]2}, 'polygon': {'count': 0}, 'polyline': {'cou[1278 chars]]}}}} != {'ima[265 chars]2}, 'mask': {'count': 1}, 'points': {'count': [1302 chars]]}}}}
966Diff is 5243 characters long. Set self.maxDiff to None to see it.
967
968======================================================================
969FAIL: test_stats_with_empty_dataset (tests.test_ops.TestOperations)
970----------------------------------------------------------------------
971Traceback (most recent call last):
972  File "/home/travis/build/openvinotoolkit/datumaro/tests/test_ops.py", line 190, in test_stats_with_empty_dataset
973    self.assertEqual(expected, actual)
974AssertionError: {'ima[224 chars]0}, 'polygon': {'count': 0}, 'polyline': {'cou[421 chars]]}}}} != {'ima[224 chars]0}, 'mask': {'count': 0}, 'points': {'count': [447 chars]]}}}}
975Diff is 1489 characters long. Set self.maxDiff to None to see it.
```

manasars:
> @manasars
> 
> I've changed the branch, and still get similar exception, but a bit another.
> 
> ```
> Traceback (most recent call last):
>   File "/home/user/cvat/utils/dicom_converter/.env/lib/python3.8/site-packages/rq/worker.py", line 936, in perform_job
>     rv = job.perform()
>   File "/home/user/cvat/utils/dicom_converter/.env/lib/python3.8/site-packages/rq/job.py", line 684, in perform
>     self._result = self._execute()
>   File "/home/user/cvat/utils/dicom_converter/.env/lib/python3.8/site-packages/rq/job.py", line 690, in _execute
>     return self.func(*self.args, **self.kwargs)
>   File "/home/user/cvat/cvat/apps/dataset_manager/views.py", line 87, in export_task_annotations
>     return export_task(task_id, dst_format, server_url=server_url, save_images=False)
>   File "/home/user/cvat/cvat/apps/dataset_manager/views.py", line 61, in export_task
>     task.export_task(task_id, temp_file, dst_format,
>   File "/home/user/cvat/cvat/apps/dataset_manager/task.py", line 749, in export_task
>     task.export(f, exporter, host=server_url, save_images=save_images)
>   File "/home/user/cvat/cvat/apps/dataset_manager/task.py", line 631, in export
>     exporter(dst_file, task_data, **options)
>   File "/home/user/cvat/cvat/apps/dataset_manager/formats/registry.py", line 37, in __call__
>     f_or_cls(*args, **kwargs)
>   File "/home/user/cvat/cvat/apps/dataset_manager/formats/pointcloud.py", line 21, in _export_images
>     dataset = Dataset.from_extractors(CvatTaskDataExtractor(
>   File "/home/user/cvat/cvat/apps/dataset_manager/bindings.py", line 535, in __init__
>     dm_anno = self._read_cvat_anno(frame_data, task_data)
>   File "/home/user/cvat/cvat/apps/dataset_manager/bindings.py", line 664, in _read_cvat_anno
>     anno = datumaro.Cuboid3D(id=anno_id, points=anno_points,
> AttributeError: module 'datumaro.components.extractor' has no attribute 'Cuboid3D'
> ```
> 
> What is wrong? Of course I did `pip install` again and restarted the server.
> 
> Issue with shape selection fixed in this branch.

Hi @bsekachev ,

We have changed the datumaro to point to the local branch which has M6 changes in base.txt file.
So might need to do a docker-prune for the changes in datumaro repository to be reflected.

bsekachev:
> We have changed the datumaro to point to the local branch which has M6 changes in base.txt file.
So might need to do a docker-prune for the changes in datumaro repository to be reflected.

I do not use docker. I've setup local environment. Looks like removing the virtual environment and creating it from scratch helped, but doing that and ``docker-prune`` is not an option for the most of CVAT users, who are not experienced in docker or programming. Will it be a problem for them after we merge the changes to the main datumaro repository, or it is going to work without extra actions?

manasars:
> > We have changed the datumaro to point to the local branch which has M6 changes in base.txt file.
> > So might need to do a docker-prune for the changes in datumaro repository to be reflected.
> 
> I do not use docker. I've setup local environment. Looks like removing the virtual environment and creating it from scratch helped, but doing that and `docker-prune` is not an option for the most of CVAT users, who are not experienced in docker or programming. Will it be a problem for them after we merge the changes to the main datumaro repository, or it is going to work without extra actions?

After the datumaro changes are merged, it will work without any extra actions just like other datasets in 2D.

MashaSS:
@manasars In point cloud dump format only nine values for points are needed. "Position": x, y, z, "rotation": x, y, z, "dimensions":  x, y, z. But requests have other 7 extra values and they are always equal to zero regardless of the cuboid position. Could you please explain what these extra values are for?

Screenshot_2

manasars:
> Tested the patch roughly. Found issues are:
> 
> 1. In 2D when you are pasting object holding Ctrl, you can paste copied object several times. It does not work in 3D mode.
The key controls are handled separately in 3D(using Threejs) , and so on click of Ctrl the screen operations are disabled(as per the change in M5 for object activation issue). 
So if the user needs to copy paste several times, they can still do Ctrl + V multiple times to achieve the same.
> 2. List of shortcuts shows some shortcuts unapplicable in 3D (switch keyframe, to background, to foreground, next/previous keyframe)
Resolved
>    ~3. Changing color does not work correctly (when Color by Instance is activated). Using both UI and hotkey (Enter). It does not change color of the shape on canvas.~
>    ~4. The issue with selection appeared again with M6. On M5 selection works when hover an object on the canvas. On M6 you need to click it.~
> 3. The issue with updating text attributes. It is very laggy. Try to compare changing text attributes in 2D and in 3D.
This issue occurs since on any key press the canvas sends an update to redux(as per the normal design) and in 3D since it uses Threejs and not SVG, this causes a slight lag.

@bsekachev : We have fixed the tests and conflicts.

bsekachev:
Hi @manasars 

Some found issues with ``pinned`` property:
On the scenario below:
1. Object cannot be deactivated
2. Additionally fitting leads to issues on side views

Peek 2021-06-08 10-05

3. With active pinned property clicked at the center of a shape, then just move the cursor (without release it) at corners, the object starts resizing (and rotating) what is a bug
Peek 2021-06-08 10-08

> @bsekachev : We have fixed the tests and conflicts.

4. Some of shortcuts still not applicable for 3D:
- Switch keyframe
- Next keyframe
- Previous keyframe
5. Issue "paste with ctrl hold" was not fixed as I see, please let me know if I misunderstand something
6. Issue with lagging text aditing was not fixed as I see, please let me know if I misunderstand something

dvkruchinin:
Hi @manasars 
The functionality for the "Point Cloud Format" annotation upload has stopped working. Please take a look PR #3263 
```
Error: Request failed with status code 500. "KeyError: None\n".
e @ cvat-app.tsx:209
showErrors @ cvat-app.tsx:220
componentDidUpdate @ cvat-app.tsx:119
os @ react-dom.production.min.js:212
pl @ react-dom.production.min.js:255
t.unstable_runWithPriority @ scheduler.production.min.js:19
Ba @ react-dom.production.min.js:122
fl @ react-dom.production.min.js:248
Qs @ react-dom.production.min.js:239
(anonymous) @ react-dom.production.min.js:123
t.unstable_runWithPriority @ scheduler.production.min.js:19
Ba @ react-dom.production.min.js:122
Ga @ react-dom.production.min.js:123
Wa @ react-dom.production.min.js:122
el @ react-dom.production.min.js:240
notify @ Subscription.js:19
t.notifyNestedSubs @ Subscription.js:90
t.handleChangeWrapper @ Subscription.js:95
v @ redux.js:296
(anonymous) @ redux-logger.js:1
(anonymous) @ index.js:11
dispatch @ redux.js:659
(anonymous) @ annotation-actions.ts:391
async function (async)
(anonymous) @ annotation-actions.ts:356
(anonymous) @ index.js:8
loadAnnotations @ annotation-menu.tsx:73
onClickMenu @ annotation-menu.tsx:137
onOk @ annotation-menu.tsx:105
onClick @ ActionButton.js:92
Q @ button.js:227
s @ react-dom.production.min.js:14
p @ react-dom.production.min.js:14
(anonymous) @ react-dom.production.min.js:14
v @ react-dom.production.min.js:15
it @ react-dom.production.min.js:52
at @ react-dom.production.min.js:51
st @ react-dom.production.min.js:52
pt @ react-dom.production.min.js:56
z @ react-dom.production.min.js:287
V @ react-dom.production.min.js:19
Zt @ react-dom.production.min.js:70
Jt @ react-dom.production.min.js:69
t.unstable_runWithPriority @ scheduler.production.min.js:19
Ba @ react-dom.production.min.js:122
D @ react-dom.production.min.js:287
$t @ react-dom.production.min.js:68
```

manasars:
> @manasars In point cloud dump format only nine values for points are needed. "Position": x, y, z, "rotation": x, y, z, "dimensions": x, y, z. But requests have other 7 extra values and they are always equal to zero regardless of the cuboid position. Could you please explain what these extra values are for?
> 
> Screenshot_2

Hi @MashaSS : For point cloud(3D), the other 7 values are not applicable and by default will be zero.

bsekachev:
> Hi @MashaSS : For point cloud(3D), the other 7 values are not applicable and by default will be zero.

Whey are they applicable?

bsekachev:
@manasars @nmanovic @zhiltsov-max 

I've just noticed it does not dump any attributes in both pointclound and velodyne formats. 
Do you have any ideas if these formats support attributes, should we dump them anyhow?

bsekachev:
Issue with context menu and doublclick shape. Side views show the object wrong:

Peek 2021-06-08 12-12

After previous gif I fitted canvas and got following issue:
Screenshot from 2021-06-08 12-12-31

bsekachev:
Just pressed "play", then "stop" and see how many pointclouds from different images are drawn on the same canvas:

Peek 2021-06-08 12-34

bsekachev:
Peek 2021-06-08 13-31
Rotation canvas by left/right arrows conflicts with frame navigation. These keys are already used to search non-empty frame.

bsekachev:
We need to add a spin while loading and setup data. You can see it in 2D canvas and in the reference tool:

Screenshot from 2021-06-08 13-46-32

Screenshot from 2021-06-08 13-46-21

bsekachev:
I see a significant difference in how the camera works in the reference tool and your implementation. 
For example, in your implementation, it cannot go through the center (see the second gif). Also, I believe we should decrease the inertia effect. They are two things making camera navigation inconvenient. 

Peek 2021-06-08 13-541
Peek 2021-06-08 13-55

bsekachev:
Summary post:

 - [ ] [Spin is absent in 3D canvas](https://github.com/openvinotoolkit/cvat/pull/3234#issuecomment-856659027)
 - [ ] [Rotation canvas by left/right arrows conflicts with frame navigation. These keys are already used to search non-empty frame](https://github.com/openvinotoolkit/cvat/pull/3234#issuecomment-856651862)
 - [ ] [Multiple point clouds are drawn on the canvas after pressing play/pause. Context image also shifted a couple of times](https://github.com/openvinotoolkit/cvat/pull/3234#issuecomment-856619148)
 - [ ] [Inconvenient camera behaviour](https://github.com/openvinotoolkit/cvat/pull/3234#issuecomment-856663648)
 - [ ] [Object keeps being activated + exception in the console + wrong top/side/front views](https://github.com/openvinotoolkit/cvat/pull/3234#issuecomment-856603240)
 - [ ] [Attributes are not dumped](https://github.com/openvinotoolkit/cvat/pull/3234#issuecomment-856599520)
 - [ ] [Changes breaks one of cypress tests](https://github.com/openvinotoolkit/cvat/pull/3234#issuecomment-856526471)
 - [ ] Extra shortcuts in the list (are not applicable for 3D)
 - [ ] When paste holding ``ctrl``, an object can be pasted a lot of times
 - [ ] Changing text attribute is very lagging on 3D workspace
 - [ ] Since review more is not supported for 3D, "Issues" tab is extra in this workspace
 - [ ] Multiple context images are supported when create a task, but the server is able return only one and UI is able to show only one
 - [ ] [Issue with "pinned" property, shown on the gif](https://github.com/openvinotoolkit/cvat/pull/3234#issuecomment-856518168)

manasars:
> @manasars @nmanovic @zhiltsov-max
> 
> I've just noticed it does not dump any attributes in both pointclound and velodyne formats.
> Do you have any ideas if these formats support attributes, should we dump them anyhow?

@bsekachev @zhiltsov-max @nmanovic : The way attributes are fetched and stored in CVAT(Mutable, default, etc) is different from the supervisely format, so if we add specific to CVAT then when we import the Point cloud in the other application may fail.
Is this fine, kindly confirm.

nmanovic:
> > @manasars @nmanovic @zhiltsov-max
> > I've just noticed it does not dump any attributes in both pointclound and velodyne formats.
> > Do you have any ideas if these formats support attributes, should we dump them anyhow?
> 
> @bsekachev @zhiltsov-max @nmanovic : The way attributes are fetched and stored in CVAT(Mutable, default, etc) is different from the supervisely format, so if we add specific to CVAT then when we import the Point cloud in the other application may fail.
> Is this fine, kindly confirm.

Manasa, if a task doesn't have attributes, we don't have problems. Thus I think it is fine to dump attributes if we have them. At the same time I don't think that we should ask you to extend other formats. Thus if an original format doesn't support attributes, let's concentrate on other problems.

nmanovic:
@manasars , could you please look at https://docs.supervise.ly/data-organization/00_ann_format_navi/03_supervisely_format_tags? Does it look like an attribute in CVAT?

jayraj-solanki:
> @manasars , could you please look at https://docs.supervise.ly/data-organization/00_ann_format_navi/03_supervisely_format_tags? Does it look like an attribute in CVAT?

Hi @nmanovic 
Supervisely tags:
sup_tags

CVAT tags:
image

Here text,number attributes matches with Supervisely.
we can also consider "One of" attribute from Supervisely as it matches with the the Select type in CVAT

But for "radio","checkbox","Mutable" type of attribute in CVAT, there's no matching attribute in supervisely.
so if we import the Point cloud with  "radio","checkbox" or "Mutable" attribute in the other application may fail.
We wanted to know if we need to add only the matching attributes  with respect to both the applications.

bsekachev:
> > @manasars , could you please look at https://docs.supervise.ly/data-organization/00_ann_format_navi/03_supervisely_format_tags? Does it look like an attribute in CVAT?
> 
> Hi @nmanovic
> Supervisely tags:
> sup_tags
> 
> CVAT tags:
> image
> 
> Here text,number attributes matches with Supervisely.
> we can also consider "One of" attribute from Supervisely as it matches with the the Select type in CVAT
> 
> But for "radio","checkbox","Mutable" type of attribute in CVAT, there's no matching attribute in supervisely.
> so if we import the Point cloud with "radio","checkbox" or "Mutable" attribute in the other application may fail.
> We wanted to know if we need to add only the matching attributes with respect to both the applications.

Hi @jayraj-solanki 

"Radio" is the same like "Select", just another HTML element is used for convenience.
"Checkbox" can be considered like one of "True", "False" for example. 
Mutable attributes are possible only for tracks. CVAT 3D does not support tracks, so I think we can just ignore this flag.

cvat-ui/src/actions/annotation-actions.ts
```diff
@@ -575,6 +576,15 @@ export function activateObject(activatedStateID: number | null, activatedAttribu
...
 
+export function updateTabContentHeight(tabContentHeight: number): AnyAction {
+    return {
+        type: AnnotationActionTypes.UPDATE_TAB_CONTENT_HEIGHT,
+        payload: {
+            tabContentHeight,
+        },
+    };
+}
+
```
bsekachev, line 587: Code related with tab height was removed in develop

jayraj-solanki:
>Hi @bsekachev,
> Following are the updated review comments:
>
> * [x]  [Spin is absent in 3D canvas](https://github.com/openvinotoolkit/cvat/pull/3234#issuecomment-856659027)
> =>Fixed, when we have multiple frames and if it takes time to load, spinner is added in 3d view & context image 
> * [x]  [Rotation canvas by left/right arrows conflicts with frame navigation. These keys are already used to search non-empty frame](https://github.com/openvinotoolkit/cvat/pull/3234#issuecomment-856651862)
> =>Shift+arrows keys are now used for Camera tilt and rotate respectively
> * [x]  [Multiple point clouds are drawn on the canvas after pressing play/pause. Context image also shifted a couple of times](https://github.com/openvinotoolkit/cvat/pull/3234#issuecomment-856619148)
> =>We have increased the play instance delay time in 3D to 100ms, but since threejs requires more resourcs for computation, sometimes on the last frame the points collide
> * [x]  [Object keeps being activated + exception in the console + wrong top/side/front views](https://github.com/openvinotoolkit/cvat/pull/3234#issuecomment-856603240)
> * [x]  [Attributes are not dumped](https://github.com/openvinotoolkit/cvat/pull/3234#issuecomment-856599520)
> =>Fixed as a part of Datumaro code
> * [x]  [Changes breaks one of cypress tests](https://github.com/openvinotoolkit/cvat/pull/3234#issuecomment-856526471)
> =>Fixed as a part of Datumaro code
> * [x]  Since review more is not supported for 3D, "Issues" tab is extra in this workspace
> * [x]  Extra shortcuts in the list (are not applicable for 3D)
> * [x]  [Issue with "pinned" property, shown on the gif](https://github.com/openvinotoolkit/cvat/pull/3234#issuecomment-856518168)
> * [ ]  [Inconvenient camera behaviour](https://github.com/openvinotoolkit/cvat/pull/3234#issuecomment-856663648)
> => Supervisely is continously updating their features, and during the development of Milestone-3, 
we followed this design approach and it was similar to supervisely
> * [ ]  When paste holding `ctrl`, an object is expected to be pasted a lot of times
> =>Not fixed, Reason: The key controls are handled separately in 3D(using Threejs) , and so on click of Ctrl the screen operations are disabled(as per the change in M5 for object activation issue).
So if the user needs to copy paste several times, they can still do Ctrl + V multiple times to achieve the same.
> * [ ]  Changing text attribute is very lagging on 3D workspace
> =>Cannot be fixed, Reason : This issue occurs since on any key press the canvas sends an update to redux(as per the normal design) and in 3D since it uses Threejs and computational time is more unlike SVG, this causes a slight lag.
> * [ ]  Preview image is always the same image
> => Solution using open3d is a going to be heavy as discussed earlier, and it cannot be accomodated at this time
> * [ ]  Multiple context images are supported when create a task, but the server is able return only one and UI is able to show only one
> => During the initial phase it was discussed that we will allow multiple images to be supported for one pcd which might be useful for implementation in the future phases

cvat-ui/src/actions/annotation-actions.ts
manasars: Removed

dvkruchinin:
Hi @jayraj-solanki 
> > * [x]   [Changes breaks one of cypress tests](https://github.com/openvinotoolkit/cvat/pull/3234#issuecomment-856526471)
> >   =>Fixed as a part of Datumaro code

Still preproducing.

nmanovic:
> Rotation canvas by left/right arrows conflicts with frame navigation. These keys are already used to search non-empty frame
=>Shift+arrows keys are now used for Camera tilt and rotate respectively

- [ ] Need to update hints and F1 shortcut help.

nmanovic:
> Multiple point clouds are drawn on the canvas after pressing play/pause. Context image also shifted a couple of times
=>We have increased the play instance delay time in 3D to 100ms, but since threejs requires more resourcs for computation, sometimes on the last frame the points collide

- [ ] Why 100ms? Why not 1000ms? Could you please look for a more robust solution? I believe threejs should provide an event when a scene is rendered.

nmanovic:
> Changing text attribute is very lagging on 3D workspace
=>Cannot be fixed, Reason : This issue occurs since on any key press the canvas sends an update to redux(as per the normal design) and in 3D since it uses Threejs and computational time is more unlike SVG, this causes a slight lag.

The issue should be fixed. I see it is as an optimization problem. I'm updating a text box. I see no reason why the whole canvas should be redrawn.

nmanovic:
> When paste holding ctrl, an object is expected to be pasted a lot of times
=>Not fixed, Reason: The key controls are handled separately in 3D(using Threejs) , and so on click of Ctrl the screen operations are disabled(as per the change in M5 for object activation issue).
So if the user needs to copy paste several times, they can still do Ctrl + V multiple times to achieve the same.

- [ ] Could you please fix the behaviors? It doesn't important how ctrl is handled. Need to be consistent between 2D and 3D cases as much as possible. I see no reasons why it cannot be supported.

jayraj-solanki:
> Hi @nmanovic , @dvkruchinin , @bsekachev 
> * [x]  Still reproducing. Need to fix.
> * [x]  Need to update hints and F1 shortcut help.
> * [x]  The issue should be fixed. I see it is as an optimization problem. I'm updating a text box. I see no reason why the whole canvas should be redrawn.
> * [x]  Could you please fix the behaviors? It doesn't important how ctrl is handled. Need to be consistent between 2D and 3D cases as much as possible. I see no reasons why it cannot be supported.
> * [x]  Why 100ms? Why not 1000ms? Could you please look for a more robust solution? I believe threejs should provide an event when a scene is rendered.

Regarding failure of one unit test("Exception: Failed to find 'coco' dataset at '/tmp/cvat_1ln1_5h3o"), We ran the unit tests on OpenVino/develop branch and got the same exception, kindly confirm

dvkruchinin:
@jayraj-solanki 
I found the issue when interacting with the buttons to change the position of the point cloud. Alt + keys (U, I, O, J, K, L)
When I press on the keyboard shortcuts, I get an error in the browser console.
```
mousetrap.js:616 Uncaught TypeError: e is not a function
    at u (mousetrap.js:616)
    at b.t._handleKey (mousetrap.js:677)
    at b.handleKey (mousetrap.js:1008)
    at HTMLDocument.d (mousetrap.js:737)
u @ mousetrap.js:616
t._handleKey @ mousetrap.js:677
b.handleKey @ mousetrap.js:1008
d @ mousetrap.js:737
```

cvat/apps/dataset_manager/bindings.py
```diff
@@ -457,17 +483,42 @@ def match_frame_fuzzy(self, path):
...
         if is_video:
             ext = FrameProvider.VIDEO_FRAME_EXT
-        if include_images:
+
+        if dimension == DimensionType.DIM_3D:
+            def _make_image(image_id, **kwargs):
+                loader = osp.join(
+                    task_data.db_task.data.get_upload_dirname(), kwargs['path'])
+                related_images = []
+                image = Img.objects.get(id=image_id)
```
zhiltsov-max, line 506: Is this code going to make a db request per each image accessed?

cvat/apps/dataset_manager/bindings.py
```diff
@@ -490,17 +541,36 @@ def _make_image(i, **kwargs):
...
+                                               attributes={'frame': frame_data.frame})
+            elif dimension == DimensionType.DIM_3D:
+                attributes = {'frame': frame_data.frame}
+                if format_type == "point_cloud":
+                    attributes["name"] = self._user["name"]
+                    attributes["createdAt"] = self._user["createdAt"]
+                    attributes["updatedAt"] = self._user["updatedAt"]
+                    attributes["labels"] = []
+                    for label in frame_data.labels.values():
+                        attributes["labels"].append({"label_id": label.id, "name": label.name, "color": label.color})
```
zhiltsov-max, line 565: 1. If color information is not used in this format, there is no point in passing and storing it.
2. If it is needed and is not item-specific, it must be passed in categories (`MaskCategories`).

cvat/apps/dataset_manager/bindings.py
```diff
@@ -490,17 +541,36 @@ def _make_image(i, **kwargs):
...
+
+                dm_item = datumaro.DatasetItem(id=osp.split(frame_data.name)[-1],
+                                               annotations=dm_anno, pcd=dm_image[0], related_images=dm_image[1],
+                                               attributes=attributes)
+
             dm_items.append(dm_item)
 
-        self._items = dm_items
+        self._items = dm_items if dm_items else dm_item
```
zhiltsov-max, line 573:
```suggestion
        self._items = dm_items
```

nmanovic:
@manasars , there are some problems with tests

```
======================================================================
ERROR: test_tasks_upload (tests.test_cli.TestCLI)
----------------------------------------------------------------------
  File "/home/django/utils/cli/tests/test_cli.py", line 138, in test_tasks_upload
    self.cli.tasks_upload(1, 'COCO 1.0', path)
  File "/home/django/utils/cli/core/core.py", line 208, in tasks_upload
    response.raise_for_status()
  File "/opt/venv/lib/python3.8/site-packages/requests/models.py", line 941, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 500 Server Error: Internal Server Error for url: http://testserver/api/v1/tasks/1/annotations?format=COCO%201.0
```

manasars:
> @manasars , there are some problems with tests
> 
> ```
> ======================================================================
> ERROR: test_tasks_upload (tests.test_cli.TestCLI)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/home/django/utils/cli/tests/test_cli.py", line 138, in test_tasks_upload
>     self.cli.tasks_upload(1, 'COCO 1.0', path)
>   File "/home/django/utils/cli/core/core.py", line 208, in tasks_upload
>     response.raise_for_status()
>   File "/opt/venv/lib/python3.8/site-packages/requests/models.py", line 941, in raise_for_status
>     raise HTTPError(http_error_msg, response=self)
> requests.exceptions.HTTPError: 500 Server Error: Internal Server Error for url: http://testserver/api/v1/tasks/1/annotations?format=COCO%201.0
> ```

Hi @nmanovic ,

This test is failing even when we tried in CVAT cvat/develop with develop branch datumaro code as mentioned above.
https://github.com/openvinotoolkit/cvat/pull/3234#issuecomment-864748899

zhiltsov-max:
> This test is failing even when we tried in CVAT cvat/develop with develop branch datumaro code as mentioned above.

CVAT is only expected to work with the currently pinned version of Datumaro in `cvat/requirements/base.txt`.

dvkruchinin:
Hi @manasars,
There was also some strange behavior.
1. Create a cuboid
2. Ctrl+C on the cuboid
3. Ctrl+V
4. Without double-clicking on the perspective, go to another frame
The display of the cuboid to paste disappears. And on the frame where the cuboid was copied, it can no longer be copied. It is not possible to do this both through hot keys and through the sidebar "Make a copy".

aschernov:
Hi. Some issues were found:

1. Due to an annotated object is getting active after it was highlighted by mouse cursor, we deal with some freezes while moving the image when another object appears under the cursor. We tried to show it on the gif below:

rotate_freeze

2. An image blocks after pressing alt+i several times (6-8). Double mouse click fixes it:
alt+I
3. A context image in the upper right angle disappears after pressing ctrl+z or undo button. Need to reload a task to return it back:
context_image
4. Holding right mouse click uses for moving around an image and also for quick new cuboid drawing. It leads to inconvenient behavior while creating new cuboids in quick mode. Need to assign another button for adding cuboids, instead of right mouse click:
move_and_create
5. The escape button doesn't work after using quick cuboid adding with right mouse click.

jayraj-solanki:
Hi @nmanovic , @dvkruchinin , @bsekachev,@zhiltsov-max, @aschernov,
> Following are status of review comments:
>  * [x]  [ I found the issue when interacting with the buttons to change the position of the point cloud. Alt + keys (U, I, O, J, K, L) When I press on the keyboard shortcuts, I get an error in the browser console.](https://github.com/openvinotoolkit/cvat/pull/3234#issuecomment-865215685)
		Fixed
> * [x]   [The display of the cuboid to paste disappears. And on the frame where the cuboid was copied, it can no longer be copied. It is not possible to do this both through hot keys and through the sidebar "Make a copy".](https://github.com/openvinotoolkit/cvat/pull/3234#issuecomment-866830173)
		We cannot view the cuboid as 2D,but when we paste it, it will still have the reference
> * [x]   [Due to an annotated object is getting active after it was highlighted by mouse cursor, we deal with some freezes while moving the image when another object appears under the cursor.](https://github.com/openvinotoolkit/cvat/pull/3234#issuecomment-867406908)
		Modified as 2d, with Select tool, camera movement is disabled
> * [x]   [An image blocks after pressing alt+i several times (6-8).](https://github.com/openvinotoolkit/cvat/pull/3234#issuecomment-867406908)
		Fixed
> * [x]   [A context image in the upper right angle disappears after pressing ctrl+z or undo button. ](https://github.com/openvinotoolkit/cvat/pull/3234#issuecomment-867406908)
		Context image changes were done internally,but it's fixed
> * [x] [Holding right mouse click uses for moving around an image and also for quick new cuboid drawing. It leads to inconvenient behavior while creating new cuboids in quick mode. Need to assign another button for adding cuboids, instead of right mouse click](https://github.com/openvinotoolkit/cvat/pull/3234#issuecomment-867406908)
		Right mouse click conflict will not happen since now Select and Move contols are independent just like 2D
>  * [x] [The escape button doesn't work after using quick cuboid adding with right mouse click. ](https://github.com/openvinotoolkit/cvat/pull/3234#issuecomment-867406908)
		Fixed




@dvkruchinin, the test case 80(cuboid_make_copy) in canvas3d_functionality is failing, though it works correctly in UI, kindly help

cvat/apps/dataset_manager/bindings.py
jayraj-solanki: Yes, it will

cvat/apps/dataset_manager/bindings.py
jayraj-solanki: The color is specific to this attribute, required to be consistent for uploading in other apps

cvat/apps/dataset_manager/bindings.py
jayraj-solanki: Done suggested changes

dvkruchinin:
Hi @jayraj-solanki,
> the test case 80(cuboid_make_copy) in canvas3d_functionality is failing, though it works correctly in UI, kindly help

Unfortunately, I can't say exactly why this is happening. I have identified a pattern. Reproducing the scenario ```Copy a shape to an another frame after press "Ctrl+V" on the first frame.``` does not work on Linux. At the same time, it works in the develop branch.
I also checked it manually. On the current commit on Linux OS, it also does not work to copy.

jayraj-solanki:
> Hi @jayraj-solanki,
> 
> > the test case 80(cuboid_make_copy) in canvas3d_functionality is failing, though it works correctly in UI, kindly help
> 
> Unfortunately, I can't say exactly why this is happening. I have identified a pattern. Reproducing the scenario `Copy a shape to an another frame after press "Ctrl+V" on the first frame.` does not work on Linux. At the same time, it works in the develop branch.
> I also checked it manually. On the current commit on Linux OS, it also does not work to copy.

Hi  @dvkruchinin ,
In 3D, We cannot view the cuboid reference when navigating between frames because when we press "Ctrl+V", a new object instance is created to be placed in the particular pcd with the copied dimensions. So you can see the copied object after  "Ctrl+V" to create the copied cuboid and "Dbl click" to place in the frame.

In the below video, I've used "Ctrl+C" to copy and "Ctrl+V" to paste in next frame.


https://user-images.githubusercontent.com/25260417/123628134-633e7700-d830-11eb-892c-e0a6a1fbe646.mp4

dvkruchinin:
> In 3D, We cannot view the cuboid reference when navigating between frames because when we press "Ctrl+V", a new object instance is created to be placed in the particular pcd with the copied dimensions. So you can see the copied object after "Ctrl+V" to create the copied cuboid and "Dbl click" to place in the frame.
> 
> In the below video, I've used "Ctrl+C" to copy and "Ctrl+V" to paste in next frame.

This scenario is also checked in the test. ```Copy a cuboid to an another frame```.
If the fall of scenario ```Copy a shape to an another frame after press "Ctrl+V" on the first frame.``` is expected in this PR, then this scenario needs to be disabled.

manasars:
> > In 3D, We cannot view the cuboid reference when navigating between frames because when we press "Ctrl+V", a new object instance is created to be placed in the particular pcd with the copied dimensions. So you can see the copied object after "Ctrl+V" to create the copied cuboid and "Dbl click" to place in the frame.
> > In the below video, I've used "Ctrl+C" to copy and "Ctrl+V" to paste in next frame.
> 
> This scenario is also checked in the test. `Copy a cuboid to an another frame`.
> If the fall of scenario `Copy a shape to an another frame after press "Ctrl+V" on the first frame.` is expected in this PR, then this scenario needs to be disabled.

Hi @dvkruchinin : Yes you are right, we will have to disable this test since in 3D:
When any draw event is in progress, we cannot change the frame as the current object which is being dragged gets reset by the new frame from threejs. So the user can still use Ctrl+V to paste the object in any other frame.

cvat/requirements/base.txt
```diff
@@ -52,4 +52,4 @@ azure-storage-blob==12.8.1
 # --no-binary=pycocotools: workaround for binary incompatibility on numpy 1.20
 # of pycocotools and tensorflow 2.4.1
 # when pycocotools is installed by wheel in python 3.8+
-datumaro==0.1.9 --no-binary=datumaro --no-binary=pycocotools
+git+https://github.com/manasars/datumaro.git@cvat-3D-M6-pcd --no-binary=datumaro --no-binary=pycocotools
```
nmanovic, line 55: @manasars , could you please fix the link and point to the develop branch of Datumaro?

nmanovic:
@manasars , could you please appropriately update tests? They should be fixed or disabled with a corresponding comment.

- https://github.com/openvinotoolkit/cvat/pull/3234/checks?check_run_id=2928943230
- https://github.com/openvinotoolkit/cvat/pull/3234/checks?check_run_id=2928943633

nmanovic:
@manasars , our team has a number of automated tests for 3D feature. A lot of regression tests which were implemented after our team found problems with your PR. Could you please check that these tests work as well?

https://github.com/openvinotoolkit/cvat/pull/3286
https://github.com/openvinotoolkit/cvat/pull/3263

jayraj-solanki:
Hi @nmanovic, @zhiltsov-max 
1. Done Datomaru specific changes in cvat. 
2. For cypress(canvas3d_functionality), we have disabled one of the non-applicable test. https://github.com/openvinotoolkit/cvat/pull/3234#issuecomment-869603842

@zhiltsov-max, CI / Unit_testing (pull_request) are failing.
These are failing even in the develop branch before 3D changes were done as mentioned before https://github.com/openvinotoolkit/cvat/pull/3234#issuecomment-866621612, and we are not sure of the cause.
Can you please help us understand the root cause.

cvat/requirements/base.txt
jayraj-solanki: Done.

zhiltsov-max:
@jayraj-solanki,

> CI / Unit_testing (pull_request) are failing.

CVAT requires some updating to use the develop version of Datumaro. We'll apply them this week.

manasars:
> @jayraj-solanki,
> 
> > CI / Unit_testing (pull_request) are failing.
> 
> CVAT requires some updating to use the develop version of Datumaro. We'll apply them this week.

@zhiltsov-max : Thank you , so currently is there any other change expected as a part of Milestone-6, could you kindly confirm.

nmanovic:
@manasars , if the PR works as expected, all tests are passed, I'm OK to merge it. Could you please check that tests for 3D functionality works as expected to save time and accept the PR faster?

- [ ] https://github.com/openvinotoolkit/cvat/pull/3286
- [ ] https://github.com/openvinotoolkit/cvat/pull/3263

manasars:
> @manasars , if the PR works as expected, all tests are passed, I'm OK to merge it. Could you please check that tests for 3D functionality works as expected to save time and accept the PR faster?
> 
> * [ ]  [[Dependent] Server tests for CVAT-3D-M6 #3286](https://github.com/openvinotoolkit/cvat/pull/3286)
> * [ ]  [[Dependent] Cypress tests for CVAT-3D-M6. #3263](https://github.com/openvinotoolkit/cvat/pull/3263)

Hi @nmanovic ,

So in-order to check if only known 2D cases are failing and 3D test cases have passed, can we pull the latest cvat-3D-Milestone6 branch changes in the above mentioned PRs, can someone help us with that?

nmanovic:
@manasars , I believe you always can merge both PRs into your branch. I don't think you need help from our side.

manasars:
> @manasars , I believe you always can merge both PRs into your branch. I don't think you need help from our side.

@nmanovic , ok sure we shall do that then and confirm once we test.

jayraj-solanki:
Hi @zhiltsov-max,
In the latest datumaro code, the exported annotation in "SLY point cloud format", the annotation files are exported as  **filename.pcd.pcd.json**.Since filename contains **.pcd** twice, some unit tests are failing for point cloud (tests.test_rest_api_3D.Task3DTest)

image

So will this be changed in Datumaro or should we update the unit tests accordingly?

zhiltsov-max:
Probably, you don't need to pass `.pcd` in the `DatasetItem.id`. We don't have such problem in the format tests. https://github.com/openvinotoolkit/datumaro/blob/develop/tests/test_sly_pointcloud_format.py

nmanovic:
@manasars , do we have any other questions? Really need to complete the PR this week and merge.

manasars:
> @manasars , do we have any other questions? Really need to complete the PR this week and merge.

Hi @nmanovic , We do not have any further questions, as per Maxim's latest update we are modifying one portion of code and testing, we shall be pushing the changes in a while.

jayraj-solanki:
Hi @nmanovic,

We have fixed the cypress tests and all the tests have passed.
Kindly review and merge the PR.

cvat/apps/dataset_manager/bindings.py
```diff
@@ -490,14 +530,36 @@ def _make_image(i, **kwargs):
...
+                    attributes["createdAt"] = self._user["createdAt"]
+                    attributes["updatedAt"] = self._user["updatedAt"]
+                    attributes["labels"] = []
+                    index = 0
+                    for _, label in task_data.meta['task']['labels']:
+                        attributes["labels"].append({"label_id": index, "name": label["name"], "color": label["color"]})
+                        attributes["track_id"] = -1
+                        index += 1
+
+                dm_item = datumaro.DatasetItem(id=osp.split(frame_data.name)[-1].split('.')[0], image=None,
```
zhiltsov-max, line 559:
```suggestion
                dm_item = datumaro.DatasetItem(id=osp.splitext(frame_data.name)[0],
```

cvat/apps/dataset_manager/bindings.py
```diff
@@ -546,14 +623,18 @@ def convert_attrs(label, cvat_attrs):
                 a_value = cvat_attrs.get(a_name, a_desc['default_value'])
                 try:
                     if a_desc['input_type'] == AttributeType.NUMBER:
-                        a_value = float(a_value)
+                        if self._dimension == DimensionType.DIM_2D:
```
zhiltsov-max, line 626: How it depends on the task dimension?

cvat/apps/dataset_manager/bindings.py
```diff
@@ -513,19 +575,34 @@ def categories(self):
...
+                         "createdAt": cvat_anno.meta['task']['created'],
+                         "updatedAt": cvat_anno.meta['task']['updated']}
+        index = 0
         for _, label in cvat_anno.meta['task']['labels']:
+            attributes = []
             label_categories.add(label['name'])
+
             for _, attr in label['attributes']:
-                label_categories.attributes.add(attr['name'])
+                if dimension == DimensionType.DIM_3D:
```
zhiltsov-max, line 594: How it depends on the task dimension?

cvat/apps/dataset_manager/bindings.py
manasars: Hi @zhiltsov-max , the **osp.split(frame_data.name)[-1].split('.')[0]** returns the entire file path and not just the fileName.
We can leave it as above or change to : 
id=osp.splitext(osp.split(frame_data.name)[-1])[0]

zhiltsov-max: Why do you need to return only basename?

jayraj-solanki: hi @zhiltsov-max,
As per the suggest changes we are getting the folder structure as shown below, **"test_canvas3d/pointcloud/"** is  nested inside **ann, pointcloud and related images** folders before the actual data.

image

Hence we need only require the basename

cvat/apps/dataset_manager/bindings.py
jayraj-solanki: This check is not required now.
Reverted changes

cvat/apps/dataset_manager/bindings.py
jayraj-solanki: This check is not required now.
Reverted changes

cvat/apps/dataset_manager/bindings.py
zhiltsov-max: CVAT preserves relative image paths on export, so such behavoiur is probably expected. But it can be a problem of the specific test.

jayraj-solanki: @zhiltsov-max, If we make the suggested change _**_id=osp.splitext(frame_data.name)[0]_**_ then the folder structure downloaded on click of dump annotations manually is as above which is incorrect as per SOW and not compatible with supervisely.

manasars:
Hi @zhiltsov-max : I hope we are fine with all the review changes, kindly confirm.

nmanovic:
@manasars , did I miss something? I cannot rotate, zoom in/out, move the primary view using mouse. Is it expected?

manasars:
> @manasars , did I miss something? I cannot rotate, zoom in/out, move the primary view using mouse. Is it expected?

Hi @nmanovic : We can rotate, zoom in/out etc when we select **Move Controls** in the left side bar similar to 2D, this was changed a part of one of the review comment changes.

nmanovic:
@manasars , could you please point me on the review comment?

nmanovic:
@manasars , I see a problem with projections of points. If you look at the screenshot below, you will see that side and front views are not informative at all. At the same time in the alternative annotation tools, it is easy to see boundaries of these objects on side and front views.
From my perspective, in the patch CVAT tries to project all points on these views. In reality it is a bad idea. 

image

manasars:
> @manasars , could you please point me on the review comment?

Please find points 3 and 6 in the below link
https://github.com/openvinotoolkit/cvat/pull/3234#issuecomment-869423620

nmanovic:
> > @manasars , could you please point me on the review comment?
> 
> Please find points 3 and 6 in the below link
> [#3234 (comment)](https://github.com/openvinotoolkit/cvat/pull/3234#issuecomment-869423620)

Do you mean the comment below?

image

If it isn't a request from our team, I'm not OK with such fix. In CVAT 2D and other annotation tools it is possible to interact with a frame using mouse by default. In the reference tool the mode is also activated by default. I like the ability to press buttons, but in reality it is much faster to interact with a scene using mouse. Do you agree?

nmanovic:
@manasars , let me merge the PR, but I will wrote a separate mail with known problems which should be resolved.

nmanovic:
@manasars , I summarized found issues here: https://github.com/openvinotoolkit/cvat/issues/3438

It isn't all issues, but which I found during the final review today. I hope the list will not grow. Please review and comment.
To simplify next reviews and avoid the overhead on your side, I will merge the PR because all checks are passed. But I want to emphasize that it isn't an approval for milestone #6. The PR isn't completed and contains several critical issues which have to be fixed.

nmanovic (approved):

Issue #2944, opened 2021-03-12
[Bug] The "occluded" attribute does not take at upload WiderFace 1.0 annotations.


DmitriyOparin:
### My actions before raising this issue
- [x] Read/searched [the docs](https://github.com/opencv/cvat/tree/master#documentation)
- [x] Searched [past issues](/issues)

### Expected Behaviour

The "occluded" attribute equals True.

### Current Behaviour

The "occluded" attribute equals False.

### Possible Solution

Fix this bug.

### Steps to Reproduce (for bugs)

1.Create task with several images (type is png).
2.Draw 'Rectangle' (shape) in first frame.
3.Switch "occluded" to True.
4.Dump WiderFace 1.0 annotations.
5.Upload WiderFace 1.0 annotations.

### Your Environment

- Git hash commit (`git log -1`): ca975072110f8e3a7a18d96df9d35fcbaa5e7ef7
- Docker version `docker version` (e.g. Docker 17.0.05): N/A
- Are you using Docker Swarm or Kubernetes? No
- Operating System and version (e.g. Linux, Windows, MacOS): ubuntu 20.04
- Code example or link to GitHub repo or gist to reproduce problem:
- Other diagnostic information / logs:
    <details>
    <summary>Logs from `cvat` container</summary>
    </details>

### Next steps
You may [join our Gitter](https://gitter.im/opencv-cvat/public) channel for community support.

PR #3288, opened 2021-06-04
Add test for bug with attribute in WiderFace

yasakova-anastasia:
### Motivation and context

- Update Datumaro version to 0.1.9
- Add test for bug with attribute in WiderFace #2944 

### Checklist

### License

```python
#
```

### DIFF
--- a/cvat/apps/dataset_manager/tests/test_rest_api_formats.py
+++ b/cvat/apps/dataset_manager/tests/test_rest_api_formats.py
@@ -1,0 +1,233 @@
+# Copyright (C) 2021 Intel Corporation
+#
+# SPDX-License-Identifier: MIT
+
+import copy
+import json
+import os.path as osp
+import random
+from io import BytesIO
+
+from datumaro.components.dataset import Dataset
+from datumaro.util.test_utils import TestDir, compare_datasets
+from django.contrib.auth.models import Group, User
+from PIL import Image
+from rest_framework import status
+from rest_framework.test import APIClient, APITestCase
+
+from cvat.apps.dataset_manager.bindings import CvatTaskDataExtractor, TaskData
+from cvat.apps.dataset_manager.task import TaskAnnotation
+from cvat.apps.engine.models import Task
+
+tasks_path = osp.join(osp.dirname(__file__), 'assets', 'tasks.json')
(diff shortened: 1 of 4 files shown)


nmanovic:
@yasakova-anastasia , Looks like in different tests we have files with the same name. Could you please fix?

yasakova-anastasia:
@nmanovic , I thought these tests would be in dataset manager server tests. So these tests belong to the same files, they just need to be merged. How should it look in your opinion?

nmanovic:
> @nmanovic , I thought these tests would be in dataset manager server tests. So these tests belong to the same files, they just need to be merged. How should it look in your opinion?

I believe in many cases each test should have own tasks.json and annotations.json. If for a test it isn't important, we can use tasks.json and annotations.json from a common folder. In general I expect that if we add a test, it should not affect another test. Otherwise you have to check each time that previous tests cover fixed problems.

Again, it is a topic for redesign. I don't want to stop current PRs due to the issue.

cvat/apps/dataset_manager/tests/test_rest_api_formats.py
```diff
@@ -189,3 +217,49 @@ def test_api_v1_check_duplicated_polygon_points(self):
...
+    def test_api_v1_check_widerface_with_all_attributes(self):
+        test_name = self._testMethodName
+        dump_format_name = "WiderFace 1.0"
+        upload_format_name = "WiderFace 1.0"
+
+        for include_images in (False, True):
+            with self.subTest():
+                # create task with annotations
+                images = self._generate_task_images(3)
+                task = self._create_task(tasks["widerface with all attributes"], images)
```
nmanovic, line 230: In the future here it is better to use "backup" functionality to create a task. Also I don't like that all our tests cases in one file. Probably they should be spitted. What do you think?

nmanovic (approved):

cvat/apps/dataset_manager/tests/test_rest_api_formats.py
yasakova-anastasia: @nmanovic , yes, we probably need to have one file with the main functions for the tests and separate files for the tests.

PR #3298, opened 2021-06-06
fix json output of CLI command ls

dbcesar:
### Motivation and context

Hi, I am currently trying to parse the output of the command `cli.py ls --json`, but it breaks. I looked and it seems the current implementation returns the tasks as isolated json elements. That is fine when we have a single task, but when we have multiple tasks the command output is an invalid json. It looks like this:
```
{
   ...
}
{
   ...
}
```

This commit aims to fix this by returning the tasks as a Json array. I believe that would be the expected result when the user uses the --json flag. Let me know if I'm missing something here.   

### Checklist

### License

```python
#
```

### DIFF
--- a/utils/cli/core/core.py
+++ b/utils/cli/core/core.py
@@ -52,15 +52,17 @@ class CLI():
         response.raise_for_status()
         output = []
         page = 1
+        json_data_list = []
         while True:
             response_json = response.json()
             output += response_json['results']
             for r in response_json['results']:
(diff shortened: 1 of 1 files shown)


nmanovic (approved):
@dbcesar , I really think that it isn't good to output information using logs (especially for parsing). But it is a problem which exists already now. Let's accept the PR and improve CLI tools in the future. Thanks for the contribution.

Issue #3294, opened 2021-06-04
Implement better algorithm for changing labels (keep attributes if possible)


timurlenk07:
### My actions before raising this issue
- [*] Read/searched [the docs](https://github.com/opencv/cvat/tree/master#documentation)
- [*] Searched [past issues](/issues)

### Expected Behaviour
When changing the type of a label for another with identically set up attributes, the attributes remain the same.

### Current Behaviour
When changing the type of a label for another with identically set up attributes, the attributes are reset.

### Possible Solution
If label type is changed, check for attributes with identical names and types, and then set the value of the attribute of the new label to that of the old one.

### Steps to Reproduce (for bugs)

1. Create a task with at least two types of labels with identical attributes (ex. name "text" with type "Text")
2. Create a label on an image and set some value to the _text_ attribute
3. Change label class, and observe that the attribute is reset

### Context

This is a real hindrance for labelling data for OCR training, and there is already a feature request about it (#2755) but no progress on it for months. Also, it is not that "smart" of a feature, it's just a simple heuristic.

bsekachev:
It could be a feature request to write a smarter algorithm for changing labels. We would be happy if somebody from the community could prepare a patch. I expect this is client-part only (JavaScript, TypeScript) expertise is necessary.

nmanovic:
@timurlenk07 , thanks for your contribution. I will close the issue.

Issue #2755, opened 2021-02-02
[Feature request] Smart attribute keeper


victor-ab:
### My actions before raising this issue
- [x] Read/searched [the docs](https://github.com/opencv/cvat/tree/master#documentation)
- [x] Searched [past issues](/issues)

![image](https://user-images.githubusercontent.com/26576394/106546172-6fc0dc80-64e9-11eb-908c-911fbf0620c8.png)

### Expected Behaviour
If you change the current label to another label with the same attribute name, keep the attribute value.

### Current Behaviour
If you change the current label to another label, it cleans the attribute value even if both have the same attribute name.

### Possible Solution
I am not a CVAT specialist, sorry.

### Steps to Reproduce (for bugs)
1. change the current label to another label with the same attribute name

### Context
A very simple tweak could improve a lot the annotation experience for certain types of annotations, specially on documents.

timurlenk07:
I would also like this feature, it is an essential one for OCR annotations. Something simple would do the trick, like if the attribute name and type matches, keep the last value.

zhouyuangan:
keep watching for this feature...

nmanovic:
@timurlenk07 , thanks for your contribution. It looks like the issue was resolved long time ago. I will close the feature request.

https://github.com/openvinotoolkit/cvat/pull/3309

PR #3309, opened 2021-06-09
Added smart attribute keeping when changing label

timurlenk07:
### Motivation and context

Implements #2755 and #3294. Basically, when changing label class, attributes are retained **IF and only** there is an attribute with the same name and type, and the old value is still valid in the new configuration.

### How has this been tested?

Tested manually on rectangle labels with various attributes.

### Checklist

### License

```python
#
```

### DIFF
--- a/cvat-core/src/annotations-objects.js
+++ b/cvat-core/src/annotations-objects.js
@@ -248,6 +248,19 @@
             this.label = label;
             this.attributes = {};
             this.appendDefaultAttributes(label);
+
+            // Try to keep old attributes if name and inputType matches and old value is still valid
+            for (const attribute of redoLabel.attributes) {
+                for (const oldAttribute of undoLabel.attributes) {
+                    if (
+                        attribute.name === oldAttribute.name
+                        && attribute.inputType === oldAttribute.inputType
+                        && validateAttributeValue(undoAttributes[oldAttribute.id], attribute)
+                    ) {
+                        this.attributes[attribute.id] = undoAttributes[oldAttribute.id];
+                    }
+                }
+            }
             const redoAttributes = { ...this.attributes };
 
             this.history.do(
--- a/cvat-core/src/annotations-objects.js
+++ b/cvat-core/src/annotations-objects.js
@@ -249,12 +249,11 @@
             this.attributes = {};
             this.appendDefaultAttributes(label);
 
-            // Try to keep old attributes if name and inputType matches and old value is still valid
+            // Try to keep old attributes if name matches and old value is still valid
             for (const attribute of redoLabel.attributes) {
                 for (const oldAttribute of undoLabel.attributes) {
                     if (
                         attribute.name === oldAttribute.name
-                        && attribute.inputType === oldAttribute.inputType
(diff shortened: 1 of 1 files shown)


timurlenk07:
As I work with images, I did not know whether attribute mutability should be considered in the attribute keeping decision. Someone better informed could tell me, what would be the correct way to handle that?

cvat-core/src/annotations-objects.js
```diff
@@ -248,6 +248,19 @@
             this.label = label;
             this.attributes = {};
             this.appendDefaultAttributes(label);
+
+            // Try to keep old attributes if name and inputType matches and old value is still valid
+            for (const attribute of redoLabel.attributes) {
+                for (const oldAttribute of undoLabel.attributes) {
+                    if (
+                        attribute.name === oldAttribute.name
+                        && attribute.inputType === oldAttribute.inputType
```
bsekachev, line 257: input types "select" and "radio" are fully compatible
we can also check compatibility additionally between "text" and "radio' or "select"

bsekachev (commented):
Hi, sorry for delay in my response.
Generally works fine with shapes and tags, one minor comment about attribute types compatibility. Should we try to keep compatble attributes?

For tracks need to write similar code in Track::_saveLabel at the same file. Do you have an ability to do it? If not, it is okay, we can merge the PR as is. 

Also, please add a line to CONTRIBUTING.md and do ``npm version patch`` in ``cvat-core`` to update its version.

timurlenk07:
I could make the code to work similarly with tracks, though I won't have the possibility to test it.
Will do the line in CONTRIBUTING.md and the version patch, but I did not understand this question of yours:

> Generally works fine with shapes and tags, one minor comment about attribute types compatibility. Should we try to keep compatble attributes?

Could you explain that a bit? Or was it about what you've commented on the code?

bsekachev:
> Generally works fine with shapes and tags, one minor comment about attribute types compatibility. Should we try to keep compatble attributes?
 
Sure, I will try to rephrase.

As you probably know CVAT supports different types of attributes: CHECKBOX, SELECT, RADIO, TEXT and NUMBER.
Internally SELECT and RADIO are the same meaning (one of a list). So, they are compatible from this point of view and if attributes have the same name, we could try to keep it also regardless of different types.

Speaking about TEXT attributes, we definitely can convert SELECT and RADIO to TEXT with no any issues. And we can do a reverse thing if TEXT attribute value is a possible value for RADIO or SELECT attributes.

timurlenk07:
Now I get it. I will try to do these this week.
Should I rebase onto the latest master commit before doing `npm version patch`?

timurlenk07:
Now I only check if the attribute names are the same, and if the old value is still valid in the new attribute. This works even for checkboxes, though I am not sure if that is a desired behaviour.

bsekachev (approved):

bsekachev:
@dvkruchinin 

Probably we already have a test to change a label, can we update it to cover this case?

dvkruchinin:
> @dvkruchinin
> 
> Probably we already have a test to change a label, can we update it to cover this case?
> Both labels must have an attribute with compatible values.

Sure. I'll check and prepare a corresponding task.

Issue #2936, opened 2021-03-10
AI Tools performance issues


bsekachev:
### My actions before raising this issue
- [x] Read/searched [the docs](https://github.com/opencv/cvat/tree/master#documentation)
- [x] Searched [past issues](/issues)

**According to the feedback about AI tools, using these tools is ineffective.** 

- It happens because usually generated objects aren't precise and an annotator needs to adjust them (spending more time than for fully manual annotation).
- One more issues related with number of generated points. Sometimes they are not enough, sometimes they are extra.
- Some objects aren't annotated by the model 

<hr/>

- [ ] Probably we should check if these models work "full-power" 
- [ ] Future work should be directed to reducing number of corrections after automatic annotation
- [ ] It was noticed, than models work better on images with better quality, so need to check if we use all available quality
- [ ] Introducing a tracking step to accelerate the feature was proposed
- [ ] Interactors don't work well for thin objects and occluded objects. Need to investigate. 

### Interactor (f-BRS)

Annotation type | Objects | Time (h) | Spped (objects per hour)
-- | -- | -- | --
manual | 43 | 0.4 | 107.5
f-BRS | 43 | 0.41 | 104.88
f-BRS with correction | 43 | 0.96 | 44.79

_f-BRS 58.33% slower_

#### People
Annotation type | Objects | Time (h) | Spped (objects per hour)
-- | -- | -- | --
manual | 53 | 0.25 | 212.00
Mask RCNN | 53 | 0.015 | 3533.33
Mask RCNN with correction | 53 | 0.44 | 120.45

_Mask RCNN 43.18% slower_

#### Cars
Annotation type | Objects | Time (h) | Spped (objects per hour)
-- | -- | -- | --
manual | 44 | 0.25 | 176.00
Mask RCNN | 44 | 0.015 | 2933.33
Mask RCNN with correction | 44 | 0.5 | 88

_Mask RCNN 50% slower_

#### Common
Annotation type | Objects | Time (h) | Spped (objects per hour)
-- | -- | -- | --
manual | 97 | 0.5 | 194.00
Mask RCNN | 97 | 0.03 | 3233.33
Mask RCNN with correction | 97 | 0.94 | 103.19

_Mask RCNN 46.8% slower_

#### People
Annotation type | Objects | Time (h) | Spped (objects per hour)
-- | -- | -- | --
manual | 132 | 0.11 | 1200
YOLO | 132 | 0.008 | 16500
[...]

Issue #2515, opened 2020-12-01
CVAT AI Tools: dextr usage


aschernov:
### My actions before raising this issue
- [x] Read/searched [the docs](https://github.com/opencv/cvat/tree/master#documentation)
- [x] Searched [past issues](/issues)

### Current Behaviour

**1. Annotation speed** (manually annotation VS semi-automatic annotation using AI tools: dextr interactor
We made an experiment to compare manually polygon annotation speed and semi-automatic using the dextr interactor and got the following results:
|Annotation type| Total polygons| Time | Speed |
|---------------|:--------------:|:----:|:-----:|
| _manually_     | 1335 | 11,2 | 119,2 |
| _dextr_     | 1450 |   19,83 | 73,12 |

*dextr speed retard is **38,65%**

**2. DEXTR advantages and disadvantages**
**Advantages:**

- Accuracy regardless of zoom. It isn't necessary to use zoom to get a good annotation result:
<img width="175" alt="12" src="https://user-images.githubusercontent.com/45849884/100737939-eb1ce980-33e5-11eb-98a6-fcf02b8f87f6.png">(zoom)
<img width="175" alt="11" src="https://user-images.githubusercontent.com/45849884/100738006-0687f480-33e6-11eb-9869-9d51985803ca.png">(without zoom)

- Works well on round-shaped objects located on a contrast background like these ones:
<img width="300" alt="2" src="https://user-images.githubusercontent.com/45849884/100731703-cec87f00-33dc-11eb-95b7-fbf8cce62a0b.png">
<img width="300" alt="3" src="https://user-images.githubusercontent.com/45849884/100731709-cf611580-33dc-11eb-9603-ac1e016c8766.png">
<img width="300" alt="4" src="https://user-images.githubusercontent.com/45849884/100731710-cff9ac00-33dc-11eb-9794-d5520fb4c9b8.png">
<img width="300" alt="5" src="https://user-images.githubusercontent.com/45849884/100731714-d0924280-33dc-11eb-9b85-deece7b4adc5.png">
it took 5-10 seconds (put 4 points and wait a request to server) to annotate such objects using DEXTR

- Works well enough on irregularly-shaped objects like trees, vegetation, bushes:
<img width="150" alt="6" src="https://user-images.githubusercontent.com/45849884/100733390-29fb7100-33df-11eb-898c-1ebba4979b5a.png">(7 sec, some corrections were made on the left side)
[...]

PR #3417, opened 2021-07-14
Semi-automatic tools enhancements (Stage 0)

### LINKED ISSUE #3361: Improve UX for IOG serverless function (start with a bounding box)
Hi Guys,

I successfully updated CVAT and deployed IOG on my ubuntu machine (not windows anymore), remote server IP address, but it still gives the same error from months ago:
image
image
image

I also tried adding port 32001 (from the below issue) to the yaml and it gives the same error.
https://github.com/openvinotoolkit/cvat/issues/2828

Are you able to help?

bsekachev:
### Motivation and context
Resolved #3361
Related #2936
Related #2515

- Added implicit "Done" button when drawing any polyshapes
- Improved visibility for removable points
- Added correct  handling of IOG serverless on UI
- Redesigned approach how AI tools component interacts with cvat-core objects (it makes the implementation easier and makes us free of a couple of issues)

### Checklist

### License

```python
#
```


ActiveChooN (approved):

cvat-core/src/ml-model.js
```diff
@@ -17,7 +17,8 @@ class MLModel {
         this._params = {
             canvas: {
                 minPosVertices: data.min_pos_points,
-                enableNegVertices: true,
+                enableNegVertices: !['DEXTR'].includes(data.name),
```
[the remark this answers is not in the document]
bsekachev: We have a discussion about it before and there was a decision do not to do such config server-side.

ActiveChooN (dismissed):

klakhov (dismissed):

serverless/pytorch/shiyinzhang/iog/nuclio/model_handler.py
```diff
@@ -50,10 +50,10 @@ def handle(self, image, bbox, pos_points, neg_points, threshold):
             # extract a crop with padding from the image
             crop_padding = 30
             crop_bbox = [
-                max(bbox[0] - crop_padding, 0),
-                max(bbox[1] - crop_padding, 0),
-                min(bbox[2] + crop_padding, image.width - 1),
-                min(bbox[3] + crop_padding, image.height - 1)
+                max(bbox[0][0] - crop_padding, 0),
```
nmanovic, line 53: Is it unified with some other input parameters? Usually a bounding box is 4 coordinates [x0, y0, x1, y1]. I'm OK with any solution if there is a common ground behind the decision.

nmanovic:
@bsekachev , why did you decide to keep bounding box till the object is finished? Can I resize it? Please look at the interface below. Probably we need something similar.

img

serverless/pytorch/shiyinzhang/iog/nuclio/model_handler.py
nmanovic: Also it will be necessary to redeploy the function with the change. Again, I'm fine with that we we have strong reasons to unify different interfaces.

bsekachev:
> @bsekachev , why did you decide to keep bounding box till the object is finished? Can I resize it? Please look at the interface below. Probably we need something similar without the extra bbox during adding points. What do you think?

You can't resize it. I believe this bounding box does not disturb a user anyhow, maybe it even helps to remember what region was selected initially, but If you want it is removed, it can be done easily.

serverless/pytorch/shiyinzhang/iog/nuclio/model_handler.py
bsekachev: Screenshot from 2021-07-20 09-57-18

Done like for points. 
We anyway need to re-deploy all the functions after adding ``min_neg_points``

nmanovic:
> > @bsekachev , why did you decide to keep bounding box till the object is finished? Can I resize it? Please look at the interface below. Probably we need something similar without the extra bbox during adding points. What do you think?
> 
> You can't resize it. I believe this bounding box does not disturb a user anyhow, maybe it even helps to remember what region was selected initially, but If you want it is removed, it can be done easily.

I like the demo and interface from IOG github page. I will vote to remove the box.

nmanovic:
@bsekachev , are you going to unify interfaces to remove points with polygons in the future?

bsekachev:
> I like the demo and interface from IOG github page. I will vote to remove the box.

Done.

bsekachev:
> @bsekachev , are you going to unify interfaces to remove points with polygons in the future?

I am not sure they can be unified absolutely the same way because we must have negative points in the first case and we must keep dragging points in the second case. At least I think need to use an additional hotkey.

serverless/pytorch/shiyinzhang/iog/nuclio/function.yaml
```diff
@@ -7,6 +7,7 @@ metadata:
     spec:
     framework: pytorch
     min_pos_points: 1
+    min_neg_points: 0
```
nmanovic, line 10: It looks the method can have negative points:

```
Clicks representations: We use the same clicks representation as DEXTR[46] by centering a 2D Gaussian around
each click, creating two separate heatmaps for foreground
and background clicks. The resulting heatmaps are concatenated with the RGB input image to form a 5-channel
input for the network. Similar to [46], the bounding box is
first relaxed by several pixels to include context, followed
by cropping to focus on the object-of-interest (Figure 2(d)).
```

nmanovic:
```
Although our IOG approach requires only three clicks
to perform segmentation, our framework naturally supports interactive adding of new foreground and background
clicks for further refinement if the user is not satisfied with
the current segmentation output. To achieve this, we append a lightweight branch before the PSP module to accept the two-channel Gaussian heatmaps encoding all the
foreground and background clicks
```

bsekachev: Okay, it has in the implementation.
``min_neg_poinst: 0`` means that method supports negative points, but they are not obligatory.

nmanovic (approved):

bsekachev:
@TOsmanov  @aschernov 

Could you please update the documentation a bit?
1. Need to add a piece of info about "Done" button that works when drawing polyshapes or using interactors
Screenshot from 2021-07-22 00-25-30 (copy)

2. If we have documentation about how to remove points when using interactors, it should be updated also (cursor visualization has been updated and the point gets bigger when it is removable). 

Screenshot from 2021-07-22 00-27-07

If you need a sever with these changes, please contact @azhavoro

aschernov:
@bsekachev We'll update the documentation

bsekachev:
@dvkruchinin

Could you please update tests where we create polygons/polylines/points (need to test "Done" button)

PR #3450, opened 2021-07-22
Semi-automatic tools enhancements (Client-side points minimizer)

bsekachev:
### Motivation and context
Related #2936
Related #2515

New setting:
Screenshot from 2021-07-26 12-18-49

On the fly working:
Peek 2021-07-26 12-20

### How has this been tested?
Manual testing

### Checklist

### License

```python
#
```


bsekachev:
@TOsmanov @aschernov 

This is a feature you requested, please provide your feedback.
Also, need to update documentation when the PR is merged.


@dvkruchinin 
Could you prepare a test for the issue? At least we added a new setting (default accuracy) and need to check this setting can be saved and restored. It also should work for OpenCV intelligent scissors (I am not sure if we have a written test for the feature)

aschernov:
@bsekachev thank you, we'll take a look.

dvkruchinin:
> @dvkruchinin
> Could you prepare a test for the issue? At least we added a new setting (default accuracy) and need to check this setting can be saved and restored. It also should work for OpenCV intelligent scissors (I am not sure if we have a written test for the feature)

Sure. I'll prepare a test. Unfortunately, it will not be possible to do this for OpenCV yet, since the problem of losing the Content-Length header is not solved if OpenCV is loaded from the browser under the control of Cypress.

bsekachev:
@nmanovic Here you can see errors in tests we discussed last meeting

ActiveChooN (approved):

nmanovic:
@dmitriikhurtin , could you please look at problems with tests in the PR?

nmanovic:
@bsekachev ,

1. For an image I got less than 3 points. Probably need to restrict that.

image

2. From my point of view the "approximation accuracy" should be visible only after we have a polygon. Now it is visible as soon as we activate an interactor. 

3. Personally I will rename "approximation accuracy" into "Number of points" with 2 visible values: `low` and `high`.

4. Also for small epsilon I would prefer to have more values: 0.5, 1, 1.25 1.5, 2, 3, 4. In some cases the difference between number of points is too big.

cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/approximation-accuracy.tsx
```diff
@@ -0,0 +1,50 @@
...
+    onChange(value: number): void;
+}
+
+export function thresholdFromAccuracy(approxPolyAccuracy: number): number {
+    const maxAccuracy = 7;
+    const approxPolyMaxDistance = maxAccuracy - approxPolyAccuracy;
+    let threshold = 0;
+    if (approxPolyMaxDistance > 0) {
+        // 0.5 for 1, 1 for 2, 2 for 3, 4 for 4, ...
+        threshold = 2 ** (approxPolyMaxDistance - 2);
```
nmanovic, line 21: Probably I don't understand but the minimum value is 0.25 now. Right? Is it expected? Need to fix the comment or code.

```suggestion
        threshold = 2 ** (approxPolyMaxDistance - 1);
```

nmanovic:
@bsekachev , it will be great to apply the approxPolyDP function for existing polygons somehow in the future. For example, how to reduce the number of points for Mask RCNN?

nmanovic:
@bsekachev , looks like a natural idea to have "done" button directly on the slider. It is more visible. But I'm not sure.

cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/approximation-accuracy.tsx
bsekachev: Comment is not precise a bit, minimum value was intended is 0.25 for approxPolyMaxDistance 0.

bsekachev: 2 ** (1 - 2) = 0.5
2 ** (2 - 2) = 1
2 ** (3 - 2) = 2
2 ** (4 - 2) = 4

A separated condition if approxPolyMaxDistance === 0, the threshold is also 0 (no approximation)

dvkruchinin:
> @dvkruchinin , could you please look at problems with tests in the PR?

I can't reproduce the problem locally or in a fork. This could probably be related to internal problems of the github actions virtual environment.
I will soon prepare the necessary test and look at the passing of tests in the current repository.

bsekachev:
> For an image I got less than 3 points. Probably need to restrict that.

Looks like OpenCV can return it after approximation (event despite of "closed" flag equals to true).

I am not sure we should strict it because in this case we can only return original points, but it is not clear for a user (users shift slider to less points, but unexpectedly they get all the points).
But I would suggest to mark the shape as RED on the canvas. It's a kind of a signal that something is wrong with the shape. Even if a user submits such the shape, they get an exception from cvat-core (not enough points). 

> Personally I will rename "approximation accuracy" into "Number of points" with 2 visible values: low and high.
> Also for small epsilon I would prefer to have more values: 0.5, 1, 1.25 1.5, 2, 3, 4. In some cases the difference between number of points is too big.

Done

bsekachev:
> @bsekachev , looks like a natural idea to have "done" button directly on the slider. It is more visible. But I'm not sure.
@bsekachev , it will be great to apply the approxPolyDP function for existing polygons somehow in the future. For example, how to reduce the number of points for Mask RCNN?

These ideas are bound each other, because in current implementation "Done" button exists when drawing all polyshapes, but the slider does not. 

Right now I would suggest leaving as is beacuse I don't have ideas how it should be implemented in UI for all the cases (exactly when it should appear and disappear, how it should work with drawing (when we need Done button, but a user is not assumed to create extra points to be minimized), or with editing (when we could need minimize points after detectors, but done button is not supposed to be used), etc.

bsekachev:
> From my point of view the "approximation accuracy" should be visible only after we have a polygon. Now it is visible as soon as we activate an interactor.

@nmanovic Done

nmanovic:
Every integration with DEXTR leads to the problem below. Probably need to check the threshold value. 

```
Uncaught (in promise) RangeError: offset is out of bounds
    at Float32Array.set (<anonymous>)
    at Object.eval [as matFromArray] (eval at initialize (cvat-ui.b532816919397967c04d.min.js:435), <anonymous>:32:9059162)
    at Object.approxPoly (cvat-ui.b532816919397967c04d.min.js:435)
    at sH.approximateResponsePoints (cvat-ui.b532816919397967c04d.min.js:435)
    at sH.componentDidUpdate (cvat-ui.b532816919397967c04d.min.js:435)
    at os (cvat-ui.b532816919397967c04d.min.js:352)
    at pl (cvat-ui.b532816919397967c04d.min.js:352)
    at t.unstable_runWithPriority (cvat-ui.b532816919397967c04d.min.js:360)
    at Ba (cvat-ui.b532816919397967c04d.min.js:352)
    at fl (cvat-ui.b532816919397967c04d.min.js:352)
```

cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/approximation-accuracy.tsx
```diff
@@ -0,0 +1,76 @@
...
+    label: <strong>more</strong>,
+};
+
+export function thresholdFromAccuracy(approxPolyAccuracy: number): number {
+    const approxPolyMaxDistance = MAX_ACCURACY - approxPolyAccuracy;
+    let threshold = 0;
+    if (approxPolyMaxDistance > 0) {
+        if (approxPolyMaxDistance <= 8) {
+            // −2.75x+7y+1=0 linear made from two points (1; 0.25) and (8; 3)
+            threshold = (2.75 * approxPolyMaxDistance - 1) / 7;
```
nmanovic, line 38: Can the threshold be negative?

nmanovic: If approxPolyAccuracy == 13, it will be -1/7

bsekachev: No, approxPolyMaxDistance is an integer >= 1 and <= 8 in this piece of code.

bsekachev:
@nmanovic 

> Every interaction with DEXTR leads to the problem below. Probably need to check the threshold value.

Could you please provide more details (image, points, threshold), because I can't reproduce it.

cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/approximation-accuracy.tsx
bsekachev: if approxPolyAccuracy == 13, then  approxPolyMaxDistance == 0 and the condition in line 35 is false

nmanovic:
Use the video file: https://github.com/opencv/opencv/blob/master/samples/data/vtest.avi?raw=true, try to annotate people using DEXTR.

bsekachev:
@nmanovic 

Works as expected
ezgif-2-278bb7241d5d

bsekachev:
@nmanovic 

Maybe just give me points that DEXTR returns to client and be sure you pull the latest changes (I committed all the changes)

nmanovic:
@bsekachev , I will merge. But I cannot pull dots from extreme points (less and more).

nmanovic (approved):

PR #3468, opened 2021-07-27
Cypress test. Update case 10, 11. Added steps "Set start point", "Change direction".

dvkruchinin:
### Motivation and context
Coverage features.
Increase code coverage.

### Checklist

### License

```python
#
```


tests/cypress/integration/actions_objects2/case_10_polygon_shape_track_label_points.js
```diff
@@ -91,17 +91,48 @@ context('Actions on polygon', () => {
...
+
+        it('Draw a polygon shape, track with second label.', () => {
             cy.createPolygon(createPolygonShapeSwitchLabel);
             cy.createPolygon(createPolygonTrackSwitchLabel);
         });
+
+        it('Set start point.', () => {
+            let notFirtsPointID;
+            cy.get('#cvat_canvas_shape_4')
+                .trigger('mousemove', {scrollBehavior: false})
```
bsekachev, line 112: Global configuration for ``scrollBehavior`` in cypress.json is not an option?

https://docs.cypress.io/guides/references/configuration#cypress-json

tests/cypress/integration/actions_objects2/case_10_polygon_shape_track_label_points.js
```diff
@@ -91,17 +91,48 @@ context('Actions on polygon', () => {
...
+                .trigger('mouseover', {scrollBehavior: false})
+                .should('have.class', 'cvat_canvas_shape_activated');
+            cy.get('.svg_select_points').not('.cvat_canvas_first_poly_point').first().then((notFirtsPoint) => {
+                notFirtsPointID = notFirtsPoint.attr('id');
+            }).rightclick({scrollBehavior: false});
+            cy.get('.cvat-canvas-point-context-menu').contains('span', 'Set start point').click({scrollBehavior: false});
+            cy.get('.svg_select_points').filter('.cvat_canvas_first_poly_point').then((firtsPoint) => {
+                expect(Number(notFirtsPointID.match(/\d+$/))).to.be.equal(Number(firtsPoint.attr('id').match(/\d+$/)) - 10);
```
bsekachev, line 120: I believe better to rely on coordinates, rather then on ids

tests/cypress/integration/actions_objects2/case_11_polylines_shape_track_label_points.js
```diff
@@ -85,17 +85,33 @@ context('Actions on polylines', () => {
...
+            cy.get('#cvat_canvas_shape_4')
+                .trigger('mousemove', {scrollBehavior: false})
+                .trigger('mouseover', {scrollBehavior: false})
+                .should('have.class', 'cvat_canvas_shape_activated');
+            cy.get('.cvat_canvas_poly_direction').then((polyDirection) => {
+                polyDirectionAttrDataAngle = polyDirection.attr('data-angle');
+            }).click({scrollBehavior: false})
+            cy.get('.cvat_canvas_poly_direction').then((afterChangePolyDirection) => {
+                expect(polyDirectionAttrDataAngle).not.equal(afterChangePolyDirection.attr('data-angle'));
+            });
```
bsekachev, line 114: In case of polylines we could check that the last points became the first point after changing direction.

tests/cypress/integration/actions_objects2/case_10_polygon_shape_track_label_points.js
dvkruchinin: This value cannot be applied globally. Since there will be no scrolling in other cases. For example, when creating a task with Advanced configuration. In this case, Cypress will not scroll the screen down and the necessary fields will not be found.

bsekachev: I would say that the most often behaviour should be set by default globally. If we need scrolling in 10 tests and disable it in 100 tests, it should be disabled by default. The decision is up to you.

dvkruchinin: Yes, you're right. But I think it's better to do this as part of a separate task.

tests/cypress/integration/actions_objects2/case_10_polygon_shape_track_label_points.js
dvkruchinin: Reworked.

tests/cypress/integration/actions_objects2/case_11_polylines_shape_track_label_points.js
dvkruchinin: Reworked.

bsekachev (approved):

PR #3473, opened 2021-07-28
Semi-automatic tools enhancements (Non-blocking UI, tips)

### LINKED ISSUE #2515: CVAT AI Tools: dextr usage

bsekachev:
### Motivation and context

Related #2936
Resolved #2515

Peek 2021-08-02 12-25
Screenshot from 2021-08-02 12-26-25

### How has this been tested?
Manual testing

### Checklist

### License

```python
#
```


cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/interactor-tooltips.tsx
```diff
@@ -0,0 +1,74 @@
...
+        <div className='cvat-interactor-tip-container'>
+            <Paragraph>{desc}</Paragraph>
+            {tool ? (
+                <>
+                    <Paragraph>
+                        <Text>You can prevent server requests holding</Text>
+                        <Text strong>{' Ctrl '}</Text>
+                        <Text>key</Text>
+                    </Paragraph>
+                    {['f-BRS', 'IOG'].includes(tool) ? (
```
ActiveChooN, line 61: I guess it can be varied on the min_neg_points parameter, so the message will not be locked only for two models.

bsekachev: Yes, u are right

ActiveChooN (dismissed):

cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/interactor-tooltips.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+interface Props {
+    tool?: string;
+    withNegativePoints?: boolean;
+}
+
+function InteractorTooltips(props: Props): JSX.Element {
+    const { tool, withNegativePoints } = props;
+
+    const DEXTR_GIF = 'https://openvinotoolkit.github.io/cvat/images/dextr_example.gif';
```
nmanovic, line 18: Probably better to add that into function.yaml for corresponding serverless functions. UI should not know about specific serverless functions. What do you think?

bsekachev: @nmanovic 

Agree. Just in this case need to redeploy the interactors again

bsekachev: On the other hand, we should not store the full path on interactor (only image name maybe). To support fetching gif from specific documentation releases/develop.

nmanovic: Need to redeploy, but it is a small pay for a good design choice. Users will be able to add "documentation" for own DL models and UI will show it automatically.

I believe the interactor should be responsible for the image location. We can get it even from another site (e.g. https://raw.githubusercontent.com/shiyinzhang/Inside-Outside-Guidance/master/ims/IOG.gif). At the same time for an image from official documentation we can have a shortcut. Probably better to insert a link with a specific comment to avoid problems with versions: https://raw.githubusercontent.com/openvinotoolkit/cvat/0fbb19ae3846a017853d52e187f0ce149adced7d/site/content/en/images/dextr_example.gif

bsekachev: Done

bsekachev:
> @bsekachev , I like the experience, especially with DEXTR. But I'm missing a feature. When I'm annotating using DEXTR, it doesn't respect "selected opacity" parameter. In some cases it is really difficult to annotate if opacity is 0 during drawing.
> 
> image

Added feature: "Selected opacity" slider now defines opacity level of shapes being drawn (works when draw shapes, or work with interactors | trackers)

nmanovic (approved):

bsekachev:
@aschernov @TOsmanov 

Could you please update the user guide about using interactors according to new changes (non-blocking UI, Selected opacity slider)? Contact @azhavoro to get a CVAT instance with these changes. 

Also, we expect a feedback from the DA team about the previous PR (with points minimizer) and this PR, since they were implemented to satisfy your requests. 

@dvkruchinin 

Could you please prepare a test for:
- Added feature: "Selected opacity" slider now defines opacity level of shapes being drawn (works when drawing shapes)

aschernov:
@bsekachev , we will update the documentation.
Also, we will prepare a feedback about the implementations you mentioned.

aschernov:
@bsekachev , we made one more annotation experiment with another but similar annotation task  we used before, and got the following conclusions:
After test annotation, comparing the traditional method (manual) and the semi-automatic (DEXTR) method with the points minimizer, we got the following data:

Speed of manual annotation is faster than getting a polygon with a semi-automatic annotation tool + in most cases we have to correct it.
Examples of annotation with DEXTR with and without points minimizer:
In this example, there are inaccuracies in the result of work of the semi-automatic annotation tool. If reduce the number of points with the points minimizer, we will need to correct only 3 points.


Often, a semi-automatic annotation tool annotated an object badly. A lot of corrections are required, even if we set a lot of points. 


We can also use the minimizer to increase the number of points to get better polygon quality for objects that are well annotated with semi-automatic annotation tool, where no additional corrections are required, if the task requires it. It took 8 seconds to receive a response from the server, excluding the time to set the first 4 points. Annotating such objects could possibly take less time than manual annotating if the request to the server took less time.

Some complex objects require a lot of points. Using points minimizer in this case increased the time a bit for correcting the polygon. If the task does not require excessive accuracy, then using points minimizer will spend less time than without it.


In general, using points minimizer in many cases reduces the time for correcting the polygon created with semi-automatic annotation tool. But manual annotation is faster, mainly this may be related with a long wait for a response from the server. For simple objects, reducing the server response time would bring the speed annotation with a semi-automatic annotation tool closer to manual annotating speed.
Perhaps, absence of blocking UI has made the annotation with a semi-automatic annotation tool a little faster. Blocking UI didn't let additional viewing of the polygon while awaiting a response from the server. Working with a clear and large object, seeing some inaccuracies in the polygon, we can add additional points to those places that were not very well annotated with a semi-automatic annotation tool. We can also see an incorrectly placed point and we can remove it without waiting for a response from the server, this is convenient.

nmanovic:
@aschernov , can we try the same experiments on objects like persons? Each DL model is trained on a specific dataset. DEXTR was trained on Pascal or COCO. It means that the model didn't see classes which you try to annotate. This is why results are extremely bad (my guess). Probably it should be mentioned in documentation. When we are able to retrain the model on the fly, it makes sense to come back to experiments which were described in the comment above.

aschernov:
@nmanovic , sure. We'll prepare a corresponding test task and repeat the experiment. I'll let you know about the results.

PR #3617, opened 2021-08-28
Added the feature to Remove annotation objects in a specified range of frames

gudipudiramanakumar:
### Issue: #33 
The problem mentioned was to be able to remove all annotation objects in a given range of frames i.e. remove annotations on video from frame N till frame N+K.

### Solution

- Added a redirect to the Remove Annotations Modal to ask if the user wants to remove all the annotations or select the range of annotations to be removed in <code>AnnotationMenuComponent</code> 

- Added a new Modal <code>RemoveRangeConfirmComponent</code> that appears on clicking select range which allows the user to select the start frame and end frame of the range.

- On selecting the range and confirming the container <code>RemoveAnnotationsRangeContainer</code> calls the action <code>removeAnnotationsinRangeAsync</code>

- <code>The removeAnnotationsinRangeAsync</code> action removes the frames using the <code>removeObjectAsync</code> action which handles throwing errors

**Note: The locked objects will not be removed**

### Motivation and context
This fixes Issue: #33 and the workflow for removing annotations in bulk can be improved

### How has this been tested?

Tested locally by running the CVAT Application and the requested feature is working. All the Annotation Objects are being removed in the specified range of frames.

### Checklist

### License

```python
#
```


cvat-ui/src/actions/annotation-actions.ts
```diff
@@ -504,6 +508,62 @@ export function changePropagateFrames(frames: number): AnyAction {
...
+                states.forEach(async (state: any) => {
+                    await   dispatch(removeObjectAsync(sessionInstance,state,force));
+                });
+            }
+
+            dispatch({
+                type: AnnotationActionTypes.REMOVE_ANNOTATIONS_INRANGE_SUCCESS,
+                payload: {
+                    startFrame,
+                    endFrame,
```
bsekachev, line 532: Do you really need this payload? It was not changed in the function, so, I believe we do not need to pass it to reducers

cvat-ui/src/reducers/interfaces.ts
```diff
@@ -505,6 +505,11 @@ export interface AnnotationState {
         visible: boolean;
         data: any;
     };
+    removeinrange: {
+        sessionInstance: any | null;
+        startFrame: number;
+        endFrame: number;
+    };
```
bsekachev, line 512: I do not think that the state should be global. It is only used in one Modal window.

cvat-ui/src/reducers/annotation-reducer.ts
```diff
@@ -945,6 +950,39 @@ export default (state = defaultState, action: AnyAction): AnnotationState => {
...
+                },
+            };
+        }
+        case AnnotationActionTypes.REMOVE_ANNOTATIONS_INRANGE: {
+            const { sessionInstance } = action.payload;
+            return {
+                ...state,
+                removeinrange: {
+                    ...state.removeinrange,
+                    sessionInstance,
```
bsekachev, line 971: Why do we need this field in redux storage? It is already there. But in your case it is not necessary even to pass to ``removeAnnotationsinRangeAsync``. Just look how ``fetchAnnotationsAsync`` is implemented.

cvat-ui/src/actions/annotation-actions.ts
```diff
@@ -504,6 +508,62 @@ export function changePropagateFrames(frames: number): AnyAction {
...
 
+
+//To remove annotation objects present in given range of frames
+export function removeAnnotationsinRangeAsync(sessionInstance: any, startFrame: number, endFrame: number, force: boolean): ThunkAction{
+    return async (dispatch: ActionCreator<Dispatch>): Promise<void> => {
+        const { filters, showAllInterpolationTracks } = receiveAnnotationsParameters();
+        try {
+
+            for(let frame = startFrame; frame<=endFrame; frame++){
+                await dispatch(changeFrameAsync(frame));
```
bsekachev, line 519: Do we really need to navigate each frame here? In this case annotation removing can fetch huge amount of time (getting frames from the server, decoding them, etc).

cvat-ui/src/actions/annotation-actions.ts
```diff
@@ -504,6 +508,62 @@ export function changePropagateFrames(frames: number): AnyAction {
...
+
+                const states = await sessionInstance.annotations.get(frame, showAllInterpolationTracks, filters );
```
bsekachev, line 521: If ``showAllInterpolationTracks`` is true, it will remove all the tracks in the job, even if they are not presented in range. 
If filters is not empty, it will not remove some objects which are not filtered according to filters.

cvat-ui/src/actions/annotation-actions.ts
```diff
@@ -504,6 +508,62 @@ export function changePropagateFrames(frames: number): AnyAction {
...
+
+                states.forEach(async (state: any) => {
+                    await   dispatch(removeObjectAsync(sessionInstance,state,force));
```
bsekachev, line 524: Have you tested it with undo/redo? The objects probably can be restored with undo in this case, at the same time when remove all annotations, they can't be restored.

bsekachev:
Hi, @gudipudiramanakumar 

Sorry for the long response. Generally I like this patch, but I believe, it would be better to implement removing logic on ``cvat-core`` level, instead of removing all the objects via removeObjectAsync() (I has shown in comments above why it could be bad practice, what is more it has pure performance).

For example adding addition arguments to:
            await sessionInstance.annotations.clear(from, to);

And one more question is to how to deal with tracks. Shall we remove full track if it's part is in removing range? Or shall we just clear track's keyframes in this range? Maybe support both ways by additional arguments in clear() method?

cvat-ui/src/containers/annotation-page/top-bar/remove-range-confirm.tsx
```diff
@@ -0,0 +1,118 @@
...
+            return dispatch(await changeFrameAsync(toFrame));
+        },
+        cancel(): void {
+            dispatch(removeAnnotationsinRangeAction(null));
+        },
+    };
+}
+
+type Props = StateToProps & DispatchToProps;
+class RemoveAnnotationsRangeContainer extends React.PureComponent<Props> {
```
bsekachev, line 81: Definition of containers for components is outdated practice. Probably you can combine all the code in component file using useSelector(), useState(), and useDispatch() hooks

bsekachev:
I will close the pull request, because there is not any reply. Do not hesitate to reopen it if you plan to continue working

gudipudiramanakumar:
Hi, @bsekachev  I am willing to continue my work on this issue, could you please reopen this pull request.

bsekachev:
Sure

gudipudiramanakumar:
Ok, sure currently I got that working. But for tracks I am currently deleting the track object if the start frame is in the range, do you suggest deleting only thee keyframes.

bsekachev:
Sure, I will try to review it ASAP

bsekachev:
There is report in CI artifacts, but I am not sure you have access to it.

[eslint_report (1).zip](https://github.com/openvinotoolkit/cvat/files/7379264/eslint_report.1.zip)

> how to run the tests locally

```
cd tests && npm ci 
npm run cypress --headless --browser chrome # headless option (you can specify specific test with --spec option)
npm open cypress # gui
```

Issue #1616, opened 2020-05-29
Support upload data to tasks in multiple batches


LukeAI:
I'm trying to upload a few thousand images to a CVAT task using the APU, split into jobs of 100.

If I try to upload more than a certain threshold, I get HTTP 413 error.

If CVAT accepted multiple POSTs to _cvat_url/tasks/{task_id}/data_ then I would be able to do this.

Perhaps by default, each new POST could be created as a new job.

azhavoro:
Hi, try to change this settings https://github.com/opencv/cvat/blob/develop/cvat/settings/base.py#L405-L406

LukeAI:
thanks a lot! Can I ask: what is the purpose of baking in hard limits?

kaustubhharapanahalli:
@azhavoro can you make this a configurable parameter though a file that is in .gitignore so that we can keep CVAT to the latest version?

LukeAI:
personally, being able to set it in my docker-compose.override.yml would be really handy - so all of my custom cfg is in one place.

nmanovic:
@LukeAI , it will be supported by the PR: https://github.com/openvinotoolkit/cvat/pull/3692

PR #3692, opened 2021-09-20
Large files uploads

### LINKED ISSUE #956: Support uploading big files using a browser
I'm trying to upload video of size 2GB and webpage show me error just after few second 

Could not create the task
Error: Network Error.

and I've checked it works with videos of size 100mb

### LINKED ISSUE #2332: Uploading huge yolo dataset causes error 500
I have 11G zip yolo dataset. When I try to upload it through web interface, i get these errors:

```
Uncaught (in promise) TypeError: Cannot read property 'toString' of undefined
    at create-task-content.tsx:156
:8080/api/v1/jobs/5/annotations?format=YOLO%201.1:1 Failed to load resource: the server responded with a status of 500 (Internal Server Error)
cvat-app.tsx:217 Error: Request failed with status code 500. "Exception: Can't find image info for '/tmp/tmpnq5lma66/obj_train_data/frame_000000.png'\n".
e @ cvat-app.tsx:217
:8080/api/v1/tasks/6/data:1 Failed to load resource: the server responded with a status of 413 (Request Entity Too Large)
cvat-app.tsx:217 Error: Request failed with status code 413. "<!DOCTYPE HTML PUBLIC \"-//IETF//DTD HTML 2.0//EN\">\n<html><head>\n<title>413 Request Entity Too Large</title>\n</head><body>\n<h1>Request Entity Too Large</h1>\nThe requested resource does not allow request data with POST requests, or the amount of data provided in\nthe request exceeds the capacity limit.\n</body></html>\n".
```
how can we fix or workaround this?

klakhov:
### Motivation and context

Resolved #956
Resolved #2332
Resolved https://github.com/openvinotoolkit/cvat/issues/1616
Fixed large files uploading via webpage, implemented using chunk file uploads with upload progress bar.
Peek 2021-09-21 11-55

### How has this been tested?

Manually tested with large videos, archives, 20k images

### Checklist

### License

```python
#
```


cvat-core/src/server-proxy.js
```diff
@@ -595,6 +595,11 @@
                 }
 
                 const taskData = new FormData();
+
+                const clientFile = taskDataSpec.client_files[0];
+                const clientFiles = taskDataSpec.client_files;
```
bsekachev, line 600: Just newer syntax to get the first element of an array is:

const clientFiles = taskDataSpec.client_files;
const [clientFile] = clientFiles;

cvat-core/src/server-proxy.js
```diff
@@ -595,6 +595,11 @@
...
+                const uploadUsingChunks = clientFiles.length === 1;
```
bsekachev, line 601: I would suggest also checking if the file size more than chunkSize you use.

cvat-core/src/server-proxy.js
```diff
@@ -619,19 +624,91 @@
...
+                        } catch (errorData) {
+                            try {
+                                await deleteTask(response.data.id);
+                            } catch (_) {
+                                // ignore
+                            }
+                            throw generateError(errorData);
+                        }
                     }
+                    await upload();
```
bsekachev, line 656: Do you really need recursion here? Just I think it would look more natural in ``while``

cvat-core/src/server-proxy.js
```diff
@@ -619,19 +624,91 @@
...
-                        // ignore
+                onUpdate('The data are being uploaded to the server 0%');
+
+                async function chunkUpload(file) {
+                    const chunkSize = 1024 * 1024 * 100; // 100 mb
+                    const totalChunks = Math.ceil(file.size / chunkSize);
+                    let currentChunkNumber = 1;
+                    async function upload() {
+                        const prevChunkNumber = currentChunkNumber - 1;
+                        const fileChunk = file.slice(prevChunkNumber * chunkSize, Math.min(prevChunkNumber * chunkSize + chunkSize, file.size));
```
bsekachev, line 635:
```suggestion
                        const chunkOffset = prevChunkNumber * chunkSize;
                        const fileChunk = file.slice(chunkOffset, Math.min(chunkOffset + chunkSize, file.size));
```

cvat-core/src/server-proxy.js
```diff
@@ -619,19 +624,91 @@
...
+                        taskData.set('chunk_file_name', file.name);
+                        taskData.set('end_of_upload', currentChunkNumber >= totalChunks);
+                        taskData.set('use_chunk_upload', true);
+                        try {
+                            await Axios.post(`${backendAPI}/tasks/${response.data.id}/data`, taskData, {
+                                proxy: config.proxy,
+                            });
+                            onUpdate(`The data are being uploaded to the server ${Math.round((currentChunkNumber / totalChunks) * 100)}%`);
+                            currentChunkNumber++;
+                            if (currentChunkNumber <= totalChunks) await upload();
```
bsekachev, line 646: Do not need try inside if we wrap the function call (chunkUpload and bulkUpload).

cvat-core/src/server-proxy.js
```diff
@@ -619,19 +624,91 @@
...
+                        } catch (errorData) {
+                            try {
+                                await deleteTask(response.data.id);
+                            } catch (_) {
+                                // ignore
+                            }
+                            throw generateError(errorData);
+                        }
```
bsekachev, line 702: The same comment about try block. Better to put it outside.

cvat-core/src/server-proxy.js
```diff
@@ -619,19 +624,91 @@
...
+                        });
+                        taskData.set('end_of_upload', currentChunkNumber >= totalChunks);
+                        taskData.set('use_chunk_upload', false);
+                        try {
+                            await Axios.post(`${backendAPI}/tasks/${response.data.id}/data`, taskData, {
+                                proxy: config.proxy,
+                            });
+                            onUpdate(`The data are being uploaded to the server ${Math.round((currentChunkNumber / totalChunks) * 100)}%`);
+                            currentChunkNumber++;
+                            if (currentChunkNumber <= totalChunks) await upload();
```
bsekachev, line 694: Better to avoid recursion here

cvat-core/src/server-proxy.js
```diff
@@ -619,19 +624,91 @@
...
+
+                async function chunkUpload(file) {
+                    const chunkSize = 1024 * 1024 * 100; // 100 mb
+                    const totalChunks = Math.ceil(file.size / chunkSize);
+                    let currentChunkNumber = 1;
+                    async function upload() {
+                        const prevChunkNumber = currentChunkNumber - 1;
+                        const fileChunk = file.slice(prevChunkNumber * chunkSize, Math.min(prevChunkNumber * chunkSize + chunkSize, file.size));
+                        taskData.set('client_files[0]', fileChunk);
+                        taskData.set('chunk_file_name', file.name);
```
bsekachev, line 637: Can we get file name from the uploaded file after it is fully uploaded?

cvat-core/src/server-proxy.js
```diff
@@ -619,19 +624,91 @@
...
+                        taskData.set('end_of_upload', currentChunkNumber >= totalChunks);
+                        taskData.set('use_chunk_upload', true);
```
bsekachev, line 639: According to current design I think we should use query parameters instead of body arguments. 
?action=append (on the server can we understand what data client sent (chunk of a single file, or one file? If yes, we do not need different flags to chunks and bulks))
?action=submit (end of upload sent in a different request with no data)

cvat-core/src/server-proxy.js
```diff
@@ -619,19 +624,91 @@
...
+                        if (requestLimit - currentGroupSize >= file.size) {
+                            currentGroup.push(file);
+                            currentGroupSize += file.size;
+                        } else {
+                            fileGroups.push(currentGroup);
+                            currentGroupSize = file.size;
+                            currentGroup = [file];
+                        }
+                    }
+                    fileGroups.push(currentGroup);
```
bsekachev, line 674: Seems like it could be rewritten with reduce in more pleasant way :)

cvat-core/src/server-proxy.js
```diff
@@ -619,19 +624,91 @@
...
+                    const totalChunks = fileGroups.length;
```
bsekachev, line 675: Would rename to ``totalBulks``

cvat/apps/engine/views.py
```diff
@@ -616,10 +614,32 @@ def data(self, request, pk):
                     status=status.HTTP_400_BAD_REQUEST)
             serializer = DataSerializer(data=request.data)
             serializer.is_valid(raise_exception=True)
+            data = {k: v for k, v in serializer.validated_data.items()}
+
+            tmp_path = os.path.join('./data/tmp', pk)
```
bsekachev, line 619: What is current directory ``.``?  I think we shouldn't operate relative paths here.

cvat/apps/engine/views.py
```diff
@@ -616,10 +614,32 @@ def data(self, request, pk):
...
+            os.makedirs(tmp_path, exist_ok=True)
+            if data.get('use_chunk_upload'): # write chunk
+                with open(os.path.join(tmp_path, data.get('chunk_file_name')), 'ab+') as destination:
```
bsekachev, line 622: If ``chunk_file_name`` exists, ``use_chunk_upload`` also always true, right? 
I think ``use_chunk_upload`` is extra here.

cvat/apps/engine/views.py
```diff
@@ -616,10 +614,32 @@ def data(self, request, pk):
...
+                    destination.write(data.get('client_files')[0]['file'].read())
+            else: # write each file
+                for client_file in data.get('client_files'):
+                    with open(os.path.join(tmp_path, client_file['file'].name), 'ab+') as destination:
+                        destination.write(client_file['file'].read())
+            if 'end_of_upload' in request.data and not data.get('end_of_upload'):
+                return Response(data='Chunk is delivered', status=status.HTTP_202_ACCEPTED)
```
bsekachev, line 629: Do you need this response text ``Chunk is delivered``? I think status with code is enough

cvat-core/src/server-proxy.js
klakhov: Done.

cvat-core/src/server-proxy.js
klakhov: Done

cvat-core/src/server-proxy.js
klakhov: Done

cvat-core/src/server-proxy.js
klakhov: Done

cvat-core/src/server-proxy.js
klakhov: Done

cvat-core/src/server-proxy.js
klakhov: Done

cvat-core/src/server-proxy.js
klakhov: Done

cvat-core/src/server-proxy.js
klakhov: No, we cant get it from the file but there is a way to set up name to the chunk using `taskData.set(client_files[0], fileChunk, 'filename.mp4')`

cvat-core/src/server-proxy.js
klakhov: Done

cvat-core/src/server-proxy.js
klakhov: Done

cvat-core/src/server-proxy.js
klakhov: Done

cvat/apps/engine/views.py
klakhov: Changed to relative path

cvat/apps/engine/views.py
klakhov: Removed

cvat/apps/engine/views.py
klakhov: Removed

nmanovic:
Some resources and ideas:
- https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpuoverview.html
- https://github.com/juliomalegria/django-chunked-upload

cvat/apps/engine/views.py
```diff
@@ -616,10 +614,38 @@ def data(self, request, pk):
                     status=status.HTTP_400_BAD_REQUEST)
             serializer = DataSerializer(data=request.data)
             serializer.is_valid(raise_exception=True)
+            data = {k: v for k, v in serializer.validated_data.items()}
+
+            tmp_path = os.path.join(settings.TMP_DATA_ROOT, pk)
```
nmanovic, line 619: @klakhov , what is the reason to upload into a temporary directory? Why don't upload directly to upload_dir?

cvat/apps/engine/views.py
```diff
@@ -616,10 +614,38 @@ def data(self, request, pk):
...
+            os.makedirs(tmp_path, exist_ok=True)
```
nmanovic, line 620: @klakhov , What if several tasks are created in parallel? Can they override data of each other?

cvat/apps/engine/views.py
klakhov: The idea behind using tmp directory is that we are uploading files before creating task in database. So, if task is not created, we dont have access to upload_dir.
It is possible to change this logic to: create task, upload data to upload_dir, update data in database. Would it be the best choice?

cvat/apps/engine/views.py
klakhov: @nmanovic, There are no problems with that. There is separate tmp directory for each task in `data/tmp/<tid>`

bsekachev:
Tried this patch uploading several videos, images, and archives. Didn't meet any issues

cvat-core/src/server-proxy.js
```diff
@@ -618,9 +634,97 @@
...
 
-                onUpdate('The data are being uploaded to the server..');
+                onUpdate('The data are being uploaded to the server 0%');
+
+                async function chunkUpload(taskId, file) {
+                    const initChunkUpload = await Axios.post(`${backendAPI}/tasks/${response.data.id}/data`,
+                        taskData, {
+                            proxy: config.proxy,
+                            params: {
+                                action: 'init-chunk-upload',
```
nmanovic, line 644: @klakhov , why here `init-chunk-upload` action is a part of body and `submit` action is a query parameter?

nmanovic:
It looks like the feature is bigger than we thought. For example, it PR will break CLI. Need to fix that. Thus we need to implement the logic in Python and JavaScript. Thus I will recommend to use tus.io as @zhiltsov-max suggested.

1. JS client:  https://github.com/tus/tus-js-client
2. Py client: https://github.com/tus/tus-py-client

We don't have an official server version for Django or DRF. Thus I will recommend to look at django-tus under MIT license and implement our own version.
3. Py server: https://github.com/alican/django-tus

nmanovic:
@klakhov , I see that you have fixed only uploading data for tasks. I'm fine if you want to split the PR on several parts and implement huge file upload for other endpoints later.

nmanovic (changes requested):

cvat/apps/engine/mixins.py
```diff
@@ -0,0 +1,228 @@
...
+        'Cache-Control': 'no-store'
+    }
+    _file_id_regex = r'(?P<file_id>\b[0-9a-f]{8}\b-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-\b[0-9a-f]{12}\b)'
+
+    def _tus_response(self, extra_headers=None, *args, **kwargs):
+        response = HttpResponse(*args, **kwargs)
+        for key, value in self._base_tus_headers.items():
+            response.__setitem__(key, value)
+        if extra_headers:
+            for key, value in extra_headers.items():
```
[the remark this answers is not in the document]
klakhov: No, it won't work. HttpResponse and DRF Response store headers as(code from response.py constructor):
```
        if headers:
            for name, value in headers.items():
                self[name] = value
```

cvat/apps/engine/mixins.py
```diff
@@ -0,0 +1,228 @@
...
+        'Access-Control-Allow-Origin': "*",
+        'Access-Control-Allow-Methods': "PATCH,HEAD,GET,POST,OPTIONS",
+        'Access-Control-Expose-Headers': "Tus-Resumable,upload-length,upload-metadata,Location,Upload-Offset",
+        'Access-Control-Allow-Headers': "Tus-Resumable,upload-length,upload-metadata,Location,Upload-Offset,content-type",
+        'Cache-Control': 'no-store'
+    }
+    _file_id_regex = r'(?P<file_id>\b[0-9a-f]{8}\b-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-\b[0-9a-f]{12}\b)'
+
+    def _tus_response(self, extra_headers=None, *args, **kwargs):
+        response = HttpResponse(*args, **kwargs)
```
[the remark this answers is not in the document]
klakhov: Done

cvat/apps/engine/mixins.py
```diff
@@ -0,0 +1,228 @@
...
+        for key, value in self._base_tus_headers.items():
```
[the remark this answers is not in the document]
klakhov: No, it won't work. HttpResponse and DRF Response store headers as(code from response.py constructor):
```
        if headers:
            for name, value in headers.items():
                self[name] = value
```

cvat/apps/engine/mixins.py
```diff
@@ -0,0 +1,228 @@
...
+        return tus_file
+
+class TusChunk:
+    def __init__(self, request):
+        self.META = request.META
+        self.offset = int(request.META.get("HTTP_UPLOAD_OFFSET", 0))
+        self.size = int(request.META.get("CONTENT_LENGTH", 102400))
+        self.content = request.body
+
+class UploadMixin(object):
```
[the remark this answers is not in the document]
klakhov: Done.

cvat/apps/engine/mixins.py
```diff
@@ -0,0 +1,228 @@
...
+            file.write(chunk.content)
+        self.offset = cache.incr("tus-uploads/{}/offset".format(self.file_id), chunk.size)
+
+    def is_complete(self):
+        return self.offset == self.file_size
+
+    def rename(self):
+        file_id_path = os.path.join(self.upload_dir, self.file_id)
+        file_path = os.path.join(self.upload_dir, self.filename)
+        file_exists = os.path.lexists(os.path.join(self.upload_dir, self.filename))
```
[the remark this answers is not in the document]
klakhov: No, it cant. When tus upload is initialized filename is validated. there is validate_filename method:
```
    def validate_filename(self, filename):
        upload_dir = self.get_upload_dir()
        file_path = os.path.join(upload_dir, filename)
        return os.path.commonprefix((os.path.realpath(file_path), upload_dir)) == upload_dir
```

cvat/apps/engine/mixins.py
```diff
@@ -0,0 +1,228 @@
...
+        if file_exists:
+            raise FileExistsError
```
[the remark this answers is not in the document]
klakhov: Done.

cvat/apps/engine/mixins.py
```diff
@@ -0,0 +1,228 @@
...
+        self.offset = cache.get("tus-uploads/{}/offset".format(file_id))
+
+    def init_file(self):
+        file_path = os.path.join(self.upload_dir, self.file_id)
+        with open(file_path, 'wb') as file:
+            file.seek(self.file_size - 1)
+            file.write(b'\0')
+
+    def write_chunk(self, chunk):
+        file_path = os.path.join(self.upload_dir, self.file_id)
```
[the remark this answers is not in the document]
klakhov: Done.

cvat/apps/engine/mixins.py
```diff
@@ -0,0 +1,228 @@
...
+                    destination.write(client_file['file'].read())
+        return Response(status=status.HTTP_200_OK)
+
+    @action(detail=True, methods=['POST'], url_path='data/finish')
+    def finish(self, request, pk):
+        return self.upload_finished(request)
+
+    # override this to do stuff after upload
+    def upload_finished(self, request):
+        return Response(status=status.HTTP_200_OK)
```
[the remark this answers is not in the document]
klakhov: Done.

cvat/apps/engine/mixins.py
```diff
@@ -0,0 +1,228 @@
...
+        self.META = request.META
+        self.offset = int(request.META.get("HTTP_UPLOAD_OFFSET", 0))
+        self.size = int(request.META.get("CONTENT_LENGTH", 102400))
+        self.content = request.body
+
+class UploadMixin(object):
+    _tus_api_version = '1.0.0'
+    _tus_api_version_supported = ['1.0.0']
+    _tus_api_extensions = []
+    _tus_max_file_size = '26843545600' # 25gb
```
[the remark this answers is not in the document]
klakhov: There is no explanation of the limit, we can use any size here but tus requires to send max file size to the client. Moved it to settings

cvat/apps/engine/mixins.py
```diff
@@ -0,0 +1,228 @@
...
+
+        tus_file = TusFile(file_id, upload_dir)
+        tus_file.init_file()
+        return tus_file
+
+class TusChunk:
+    def __init__(self, request):
+        self.META = request.META
+        self.offset = int(request.META.get("HTTP_UPLOAD_OFFSET", 0))
+        self.size = int(request.META.get("CONTENT_LENGTH", 102400))
```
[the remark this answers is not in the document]
klakhov: I suppose here we need to use default chunk size. I've added it to settings/base.by as TUS_DEFAULT_CHUNK_SIZE

cvat-core/src/server-proxy.js
```diff
@@ -618,9 +634,97 @@
...
+                                size: file.size,
+                                file_name: file.name,
+                            },
+                        });
+                    const uploadId = initChunkUpload.data;
+                    const totalChunks = Math.ceil(file.size / chunkSize);
+                    let currentChunkNumber = 1;
```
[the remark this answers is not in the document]
klakhov: Changed to 0

cvat-core/src/server-proxy.js
```diff
@@ -618,9 +634,97 @@
...
+                    const tags = [];
+                    while (currentChunkNumber <= totalChunks) {
+                        const prevChunkNumber = currentChunkNumber - 1;
```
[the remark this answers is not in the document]
klakhov: This code was removed

cvat-core/src/server-proxy.js
```diff
@@ -618,9 +634,97 @@
...
+                                proxy: config.proxy,
+                                params: {
+                                    action: 'append',
+                                    upload_id: uploadId,
+                                },
+                            });
+                        const tag = tagResponse.data;
+                        tags.push({ chunk_number: currentChunkNumber, tag });
+                        taskData.delete('client_files[0]');
+                        const sentSize = Math.min(chunkOffset + chunkSize, file.size) - chunkOffset;
```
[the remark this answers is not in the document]
klakhov: This code was removed

cvat-core/src/server-proxy.js
```diff
@@ -618,9 +634,97 @@
...
+                        if (chunkSize - lastBulk.size >= file.size) {
+                            lastBulk.files.push(file);
+                            lastBulk.size += file.size;
+                        } else {
+                            fileGroups.push({ files: [file], size: file.size });
+                        }
+                        return fileGroups;
+                    }, [{ files: [], size: 0 }]);
+                    const totalBulks = fileBulks.length;
+                    let currentChunkNumber = 1;
```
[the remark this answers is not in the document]
klakhov: Changed to 0

cvat-core/src/server-proxy.js
```diff
@@ -618,9 +634,97 @@
...
+                            proxy: config.proxy,
+                        });
+                        for (let i = 0; i < fileBulks[currentChunkNumber - 1].files.length; i++) {
+                            taskData.delete(`client_files[${i}]`);
+                        }
+                        totalSentSize += fileBulks[currentChunkNumber - 1].size;
+                        currentChunkNumber++;
+                    }
+                }
+
```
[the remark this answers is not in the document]
klakhov: It's already done before chunk/bulk uploads

cvat/apps/engine/views.py
```diff
@@ -628,15 +627,62 @@ def jobs(self, request, pk):
...
+            elif action == 'submit' or not action:
+                uploaded_files = task_data.get_uploaded_files()
+                uploaded_files.extend(data.get('client_files'))
+                serializer.validated_data.update({'client_files': uploaded_files})
+
             db_data = serializer.save()
             db_task.data = db_data
             db_task.save()
-            data = {k:v for k, v in serializer.data.items()}
+            data = {k: v for k, v in serializer.data.items()}
```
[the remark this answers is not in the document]
klakhov: Done. (But we need to use `serilizer.validated_data` or an error is raised)

cvat-core/src/server-proxy.js
```diff
@@ -632,18 +649,95 @@
...
-                onUpdate('The data are being uploaded to the server..');
-                try {
-                    await Axios.post(`${backendAPI}/tasks/${response.data.id}/data`, taskData, {
-                        proxy: config.proxy,
+                onUpdate('The data are being uploaded to the server 0%');
+
+                async function chunkUpload(taskId, file) {
+                    return new Promise((resolve, reject) => {
+                        const upload = new tus.Upload(file, {
+                            endpoint: `${origin}/${backendAPI}/tasks/${taskId}/data/tus/`,
```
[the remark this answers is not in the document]
klakhov: We can remove `/tus/` and `/append/` endpoints with more complicated logic on `/data/` endpoint. But then we need to remove `/finish` too. What is the best way to do it? Check for amount of sent files?

cvat-core/src/server-proxy.js
```diff
@@ -632,18 +649,95 @@
...
+                                totalSentSize += file.size;
+                                resolve();
+                            },
+                        });
+                        upload.start();
                     });
+                }
+
+                async function bulkUpload(taskId, files) {
+                    const fileBulks = files.reduce((fileGroups, file) => {
```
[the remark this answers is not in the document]
klakhov: I don't think reduce looks complicated, here it helps to write less code.

cvat-core/src/server-proxy.js
```diff
@@ -632,18 +649,95 @@
...
+                    }, [{ files: [], size: 0 }]);
+                    const totalBulks = fileBulks.length;
+                    let currentChunkNumber = 1;
+                    while (currentChunkNumber <= totalBulks) {
+                        for (const [idx, element] of fileBulks[currentChunkNumber - 1].files.entries()) {
+                            taskData.append(`client_files[${idx}]`, element);
+                        }
+                        onUpdate(`The data are being uploaded to the server
+                                    ${((totalSentSize / totalSize) * 100).toFixed(2)}%`);
+                        await Axios.post(`${backendAPI}/tasks/${taskId}/data/append`, taskData, {
```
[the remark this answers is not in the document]
klakhov: Let's discuss amount endpoints in conversation about `/tus/` endpoint

cvat-core/src/server-proxy.js
```diff
@@ -632,18 +649,95 @@
...
+                            proxy: config.proxy,
+                        });
+                    for (const file of chunkFiles) {
+                        await chunkUpload(response.data.id, file);
+                    }
+                    if (bulkFiles.length > 0) {
+                        await bulkUpload(response.data.id, bulkFiles);
+                    }
+                    if (chunkFiles.length > 0 || bulkFiles.length > 0) {
+                        await Axios.post(`${backendAPI}/tasks/${response.data.id}/data/finish`,
```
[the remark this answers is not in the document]
klakhov: I don't see any comments with suggestions about it. We can do something based on amount of files or something else. Anyway, i would like to hear what do you think about it.

nmanovic:
@klakhov , 
- the known list of files can be a good heuristics. But in this case we cannot finish upload if by a reason it is not possible to send a file from the list.
- We can also use special acceptance headers (e.g. `Upload-Start` and `Upload-Finish`) like Tus does. It will be a part of our protocol. To support the legacy behaviors we can omit both headers and in the case the single request will be start and end at the same time.

What do you think?

klakhov:
@nmanovic 
I'm fine with that. Lets use headers.

klakhov:
@nmanovic Could you please take a look on updated pr. I've removed `/append`, `/tus`, `/finished` endpoints and used `Upload-Start`, `Upload-Finish`, `Upload-Multiple` headers to indicate request purpose

cvat/settings/base.py
```diff
@@ -482,3 +482,29 @@ def add_ssh_keys():
 
 USE_CACHE = True
 
+CORS_ALLOW_HEADERS = [
```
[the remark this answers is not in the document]
klakhov: Done.

nmanovic:
@klakhov , could you please resolve conflicts?

nmanovic (approved):
Great job!

Comments for the future:
1. I saw that in case of bulk upload we have the following operation for each file in python code: open, read, write, close. Probably we can just move them from a temporary location to upload directory. Probably it will be an optimization in the future.
2. Need to update CLI to support the new protocol
3. Need to add the protocol for all other upload operations (import task, import dataset, import project).

PR #3712, opened 2021-09-23
Manifest optimization

Marishka17:
### Checklist

### License

```python
#
```

### DIFF
--- a/utils/dataset_manifest/core.py
+++ b/utils/dataset_manifest/core.py
@@ -6,19 +6,49 @@ import av
 import json
 import os
 from abc import ABC, abstractmethod, abstractproperty
-from collections import OrderedDict
 from contextlib import closing
 from PIL import Image
+from tqdm import tqdm
 from .utils import md5_hash, rotate_image
+from tempfile import NamedTemporaryFile
 
 class VideoStreamReader:
-    def __init__(self, source_path):
-        self.source_path = source_path
-        self._key_frames = OrderedDict()
-        self.frames = 0
+    def __init__(self, source_path, chunk_size, force):
+        self._source_path = source_path
+        self._frames_number = None
+        self._force = force
+        self._upper_bound = 3 * chunk_size + 1
 
         with closing(av.open(self.source_path, mode='r')) as container:
-            self.width, self.height = self._get_frame_size(container)
+            video_stream = VideoStreamReader._get_video_stream(container)
+            isBreaked = False
+            for packet in container.demux(video_stream):
+                if isBreaked:
+                    break
+                for frame in packet.decode():
+                    # check type of first frame
+                    if not frame.pict_type.name == 'I':
+                        raise Exception('First frame is not key frame')
+
+                    # get video resolution
+                    if video_stream.metadata.get('rotate'):
+                        frame = av.VideoFrame().from_ndarray(
+                            rotate_image(
+                                frame.to_ndarray(format='bgr24'),
+                                360 - int(container.streams.video[0].metadata.get('rotate')),
+                            ),
+                            format ='bgr24',
+                        )
+                    self.height, self.width = (frame.height, frame.width)
+                    # not all videos contain information about numbers of frames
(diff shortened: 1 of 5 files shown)


utils/dataset_manifest/core.py
```diff
@@ -6,154 +6,192 @@
...
-            video_stream = self._get_video_stream(container)
-
-            for packet in container.demux(video_stream):
-                for frame in packet.decode():
-                    if not frame.pict_type.name == 'I':
-                        raise Exception('First frame is not key frame')
-                    return
+                if md5_hash(frame) != key_frame['md5'] or frame.pts != key_frame['pts']:
+                    return False
+                return True
```
nmanovic, line 72: Do we need to check only first frame in every packet?

utils/dataset_manifest/core.py
```diff
@@ -6,154 +6,192 @@
...
+        self._stop = value
+
+    @property
+    def step(self):
+        return self._step
+
+    @step.setter
+    def step(self, value):
+        assert isinstance(value, int)
+        self._step = value
```
nmanovic, line 190: Probably better just `self._step = int(value)`. Thus if value can be converted to int, it will be converted. Otherwise you will get an exception (e.g. `int(1)`, `int('1')`, `int(1.5)`)

utils/dataset_manifest/core.py
```diff
@@ -6,154 +6,192 @@
...
+        self._step = step
+
+    @property
+    def start(self):
+        return self._start
+
+    @start.setter
+    def start(self, value):
+        assert isinstance(value, int)
+        self._start = value
```
nmanovic, line 172: `self._start = int(value)`

nmanovic:
@Marishka17 , it will be great to have a line in CHANGELOG. Probably you can estimate some speed up which we should get in some scenarios. It will be helpful as well.

nmanovic:
@Marishka17 , need to add tqdm module into requirements

```
  File "./create.py", line 91, in <module>
    from dataset_manifest.core import VideoManifestManager, ImageManifestManager
  File "/mnt/data.ssd1tb/workspace/cvat/utils/dataset_manifest/__init__.py", line 4, in <module>
    from .core import VideoManifestManager, ImageManifestManager
  File "/mnt/data.ssd1tb/workspace/cvat/utils/dataset_manifest/core.py", line 13, in <module>
    from tqdm import tqdm
ModuleNotFoundError: No module named 'tqdm'
```

utils/dataset_manifest/core.py
```diff
@@ -355,53 +423,65 @@ def get_subset(self, subset_names):
...
+                'resolution': self._reader.resolution,
+                'length': len(self._reader),
+            },
+        }
+        for key, value in base_info.items():
+            json_item = json.dumps({key: value}, separators=(',', ':'))
+            file.write(f'{json_item}\n')
+
+    def _write_core_part(self, file):
+        for item in tqdm(self._reader,
```
nmanovic, line 453: It is better don't use tqdm inside a library. It is a good solution for create.py. What do you think?

Marishka17: I think that need to use `tqdm` where an iteration directly takes place. It seems to me not a very good solution to raise an iteration to a higher level. 
If we will use `tqdm` in create.py, we should change `manifest.create()` to `for _ in manifest.create(): pass` in all places where `manifest.create()` is used, or expand existing API. What do you think about it?

nmanovic: Does it spam CVAT server logs with unnecessary messages?

nmanovic: In general I don't advice a solution. I just want to say that the general rule is to use tqdm inside CLI not in a library.

nmanovic: You can pass an optional object into the `create` method. Thus CLI will pass tqdm like object. If the code is imported, the object will be None.

PR #3737, opened 2021-09-28
Tried to fix some tests

bsekachev:
### Motivation and context
In the patch I tried to solve some problems in Cypress tests (happen time to time), like: ``Timeout when finding cvat-something-popover-visible``
After removing ``-visible`` classes at all and updating tests, the root problem still persisted (animation didn't work correctly on CI [it is not clear where the problem is, in antd, its 3rdparty packages, cypress, or Chrome]. An element could be fully visible for 25 seconds, but it still has css property ``pointer-events: none``, which is normally disappears in milliseconds after the element is visible. It was solved by disabling animation for popover elements in the sidebar)

### How has this been tested?
Successful CI each time after re-running (4x restarts)

### Checklist

### License

```python
#
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Fixed JSON transform issues in network requests (<https://github.com/openvinotoolkit/cvat/pull/3706>)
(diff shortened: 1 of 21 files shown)


bsekachev:
@dvkruchinin 

Could you please look at the patch and left comments if any?

dvkruchinin:
> Could you please look at the patch and left comments if any?

@bsekachev 
I think this is a great solution.

PR #3832, opened 2021-10-27
Added rotated bounding boxes

### LINKED ISSUE #1128: Rotated bounding boxes
Thank you for the great effort, you are putting into this project :) There is, however, a feature I miss; rotated bounding boxes. Especially when objects are thin and diagonal, an ordinary bounding box fits poorly. Examples of such cases are shown here: [rotated bounding boxes](https://pythonawesome.com/rotated-mask-r-cnn-from-bounding-boxes-to-rotated-bounding-boxes/)

A way annotation could be carried out is by using the recently implemented "extreme clicking" combined with the [minimum bounding-box algorithem](https://github.com/cansik/LongLiveTheSquare)

bsekachev:
### Motivation and context
Resolved #1128

### How has this been tested?
Manual testing

**TODO:** 

- [x] Add a couple of jest tests
- [x] Add angle visualization
- [x] Fix migrations
- [x] Add default values in bindings
- [x] Add exporting as polygons (in case of mask), add exporting additional attribute in other cases

### Checklist

### License

```python
#
```


bsekachev:
@ActiveChooN @zhiltsov-max Could you please review this patch?

cvat/apps/dataset_manager/annotation.py
```diff
@@ -429,24 +429,39 @@ def _modify_unmached_object(obj, end_frame):
...
             copied["keyframe"] = False
             copied["frame"] = frame
+            if rotation is not None:
+                copied["rotation"] = rotation
             if points is not None:
                 copied["points"] = points
             return copied
 
+        def find_angle_diff(right_angle, left_abgle):
+            angle_diff = right_angle - left_abgle
```
zhiltsov-max, line 443:
```suggestion
        def find_angle_diff(right_angle, left_angle):
            angle_diff = right_angle - left_angle
```

cvat/apps/dataset_manager/bindings.py
```diff
@@ -508,10 +510,10 @@ def match_frame_fuzzy(self, path):
...
+    LabeledShape = NamedTuple('LabledShape', [('type', str), ('frame', int), ('label', str), ('points', List[float]), ('rotation', float), ('occluded', bool), ('attributes', List[InstanceLabelData.Attribute]), ('source', str), ('group', int), ('z_order', int), ('task_id', int)])
     LabeledShape.__new__.__defaults__ = (0,0)
     TrackedShape = NamedTuple('TrackedShape',
-        [('type', str), ('frame', int), ('points', List[float]), ('occluded', bool), ('outside', bool), ('keyframe', bool), ('attributes', List[InstanceLabelData.Attribute]), ('source', str), ('group', int), ('z_order', int), ('label', str), ('track_id', int)],
+        [('type', str), ('frame', int), ('points', List[float]), ('rotation', float), ('occluded', bool), ('outside', bool), ('keyframe', bool), ('attributes', List[InstanceLabelData.Attribute]), ('source', str), ('group', int), ('z_order', int), ('label', str), ('track_id', int)],
     )
     TrackedShape.__new__.__defaults__ = ('manual', 0, 0, None, 0)
```
zhiltsov-max, line 518: Consider adding a default value of 0.

zhiltsov-max:
Currently, there are no formats in Datumaro using such annotations. Maybe, we should convert them on exporting to polygons or masks in the corresponding formats?

nmanovic:
@bsekachev , I like the interface with a dot to rotate the box. Could we show the angle as a number when a user rotates the box? Probably it is necessary to precisely rotate bbox on 0, 90, 180, 270 angles. It can help if the user wants to reset the angle (don't want to delete box and redraw it again to remove).

bsekachev:
> Currently, there are no formats in Datumaro using such annotations. Maybe, we should convert them on exporting to polygons or masks in the corresponding formats?

@zhiltsov-max I think it make sense. But could you please point me where should I do it?

> Probably it is necessary to precisely rotate bbox on 0, 90, 180, 270 angles. It can help if the user wants to reset the angle (don't want to delete box and redraw it again to remove).

@nmanovic Possible to implement, but BTW we do not need to delete/draw again a box, since we have Shift+N (redraw one shape) feature.

nmanovic:
> > Currently, there are no formats in Datumaro using such annotations. Maybe, we should convert them on exporting to polygons or masks in the corresponding formats?
> 
> @zhiltsov-max I think it make sense. But could you please point me where should I do it?
> 
> > Probably it is necessary to precisely rotate bbox on 0, 90, 180, 270 angles. It can help if the user wants to reset the angle (don't want to delete box and redraw it again to remove).
> 
> @nmanovic Possible to implement, but BTW we do not need to delete/draw again a box, since we have Shift+N (redraw one shape) feature.

If we just show the angle during rotation (near box size), it will solve most of problems. Without the feature I don't know how to set a precise angle.

nmanovic:
```
INFO - 2021-11-03 07:00:49,339 - font_manager - generated new fontManager
Creating test database for alias 'default'...
CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph: (0043_auto_20211028_0945, 0043_auto_20211027_0718 in engine).
To fix them run 'python manage.py makemigrations --merge'
1
Error: Process completed with exit code 1.

```

zhiltsov-max:
> > Currently, there are no formats in Datumaro using such annotations. Maybe, we should convert them on exporting to polygons or masks in the corresponding formats?

> @zhiltsov-max I think it make sense. But could you please point me where should I do it?

I think, the most simple way to do this is to export these boxes as polygons in [this function](https://github.com/openvinotoolkit/cvat/blob/develop/cvat/apps/dataset_manager/bindings.py#L1088-L1169), because there is no other way to represent these annotations now in Datumaro, but note, that this will export them as polygons (or masks, where application) and import too (which can be unexpected for CVAT format, for example). I don't know if we need to introduce a rotation property of bbox in Datumaro.

Another option is exporting a `rotation` annotation attribute, which can be used by a format or a transform. Then, in most formats a transform can be added that converts bboxes with rotation attribute to polygons, [like this](https://github.com/openvinotoolkit/cvat/blob/develop/cvat/apps/dataset_manager/formats/icdar.py#L68-L85).

ActiveChooN:
> export these boxes as polygons

I don't think this looks like CVAT behavior expected by users. So, implementing a special attribute looks more attractive.

@bsekachev, overall code except comments looks great.

zhiltsov-max:
> > export these boxes as polygons

> I don't think this looks like CVAT behavior expected by users. So, implementing a special attribute looks more attractive.

Probably, we need to implement a transform on CVAT side to convert rotated boxes to polygons and use it in segmentation formats. Otherwise, export as boxes with an attribute.

cvat/apps/dataset_manager/bindings.py
```diff
@@ -1129,6 +1131,8 @@ def convert_attrs(label, cvat_attrs):
         anno_label = map_label(shape_obj.label)
         anno_attr = convert_attrs(shape_obj.label, shape_obj.attributes)
         anno_attr['occluded'] = shape_obj.occluded
+        if shape_obj.type == ShapeType.RECTANGLE:
+            anno_attr['rotation'] = shape_obj.rotation
```
bsekachev, line 1135: The main question is. Should we rewrite existing ``rotation`` attribute? (if it exists)

bsekachev:
@zhiltsov-max @ActiveChooN @nmanovic 

Done all mentioned changes.

cvat/apps/dataset_manager/bindings.py
zhiltsov-max: Maybe, we need to have a general policy about this, which also includes `occluded` and other things.

cvat/apps/dataset_manager/bindings.py
```diff
@@ -1149,9 +1153,23 @@ def convert_attrs(label, cvat_attrs):
...
+                def rotate_point(p):
+                    [x, y] = p
+                    rx = cx + math.cos(angle) * (x - cx) - math.sin(angle) * (y - cy)
+                    ry = cy + math.sin(angle) * (x - cx) + math.cos(angle) * (y - cy)
+                    return rx, ry
+
+                poly_points = list(chain.from_iterable(map(rotate_point, [(x0, y0), (x1, y0), (x1, y1), (x0, y1)])))
+                anno = datum_annotation.Polygon(poly_points,
+                    label=anno_label, attributes=anno_attr, group=anno_group,
+                    z_order=shape_obj.z_order)
```
zhiltsov-max, line 1168: We have several mask formats, which all should use this logic, but this function doesn't seem like a proper place for format checking. Consider creating an `ItemTransform`, which converts polygons with `rotation` attribute to rotated polygons - see example in https://github.com/openvinotoolkit/cvat/blob/develop/cvat/apps/dataset_manager/formats/icdar.py#L68-L85

cvat/apps/dataset_manager/bindings.py
bsekachev: ``occluded``, ``outside`` existed always with CVAT. But rotation does not and a existing tasks can include this attribute.

cvat/apps/dataset_manager/bindings.py
bsekachev: > but this function doesn't seem like a proper place for format checking

Before you said:
https://github.com/openvinotoolkit/cvat/blob/develop/cvat/apps/dataset_manager/bindings.py#L1088-L1169

bsekachev: @zhiltsov-max Reworked as an example.

But in this case we always need to rewrite attribute "rotation" (which may exist as it was said before). 
So, maybe adding ``rotation`` property to  datum_annotation._Shape is a solution?

And what other formats use mask? (You said not only ``Segmentation mask``)

zhiltsov-max: > And as I can see, there is other format check in this function: `if format_name == "sly_pointcloud"`:

Agree, it is one of the hacks introduced with 3d support. I hope, it will be refactored at some day. We've already made some progress in this and removed few other hacks.

> So, maybe adding rotation property to `datum_annotation._Shape` is a solution?

This would seem like it is a core functionality, which is supported everywhere (in all formats etc.), while it's not. Thus, we prefer keep CVAT properties as attributes.

> But in this case we always need to rewrite attribute `rotation` (which may exist as it was said before).

It is an open question to be resolved. Maybe we can check if such attribute is defined in the label configuration and rotate shapes during mask conversion only if there is no declared attribute? We're just inventing a "reasonable" solution here, because we speak about formats that don't support such option. Maybe, we should ask about this from the customer, who wanted this feature?

> And what other formats use mask? (You said not only `Segmentation mask`)

They are listed [here](https://openvinotoolkit.github.io/cvat/docs/manual/advanced/annotation-with-polygons/creating-mask/#creating-masks). You can also [search the codebase by `dataset.transform('polygons_to_masks')`](https://github.com/openvinotoolkit/cvat/search?q=dataset.transform%28%27polygons_to_masks%27%29) to be sure. I found 7 formats.

cvat/apps/dataset_manager/formats/mask.py
```diff
@@ -14,11 +18,36 @@
...
+class RotatedBoxesToPolygons(ItemTransform):
+    def _rotate_point(self, p, angle, cx, cy):
+        [x, y] = p
+        rx = cx + math.cos(angle) * (x - cx) - math.sin(angle) * (y - cy)
+        ry = cy + math.sin(angle) * (x - cx) + math.cos(angle) * (y - cy)
+        return rx, ry
+
+    def transform_item(self, item):
+        annotations = item.annotations
+        anns = [p for p in annotations if p.type == datum_annotation.AnnotationType.bbox and p.attributes['rotation']]
```
zhiltsov-max, line 30:
```suggestion
        annotations = item.annotations[:]
        anns = [p for p in annotations if p.type == datum_annotation.AnnotationType.bbox and 'rotation' in p.attributes]
```

cvat/apps/dataset_manager/bindings.py
bsekachev: Added the transformation to these 7 formats. Works fine except of MOTS, OpenImages (empty masks at least if look at them using only eyes, but the transformation is called), and ICDAR Segmentation: 

image

I suspect there are some format specific reasons of such behaviour, because the same results on develop branch.

cvat/apps/dataset_manager/formats/mask.py
bsekachev: 'rotation' always in p.attributes, but it can be 0. In this case we should not convert boxes to polygons.

cvat/apps/dataset_manager/bindings.py
zhiltsov-max: > MOTS, OpenImages (empty masks at least if look at them using only eyes, but the transformation is called)

It's ok, they have grayscale masks.

> I suspect there are some format specific reasons of such behaviour, because the same results on develop branch.

Yes, this format expects a string `color` attribute for each annotation.

@kirill-sizov , probably, this format deserves a better error message and, maybe, better handling of color info (autogeneration?).

nmanovic:
@bsekachev , somewhere the precision is lost. Always `N.0` is dumped like 72.0 even if in UI I see 71.6

nmanovic:
@bsekachev , for tracked boxes I see the same problem. All boxes has integer coordinates indeed.

nmanovic:
@bsekachev , also I tried to set the angle to 'zero'. It is just impossible. Need to redraw. Need to be able set a presice angle somehow. It can be important for angles like 0, 45, 90, ...

bsekachev:
>  also I tried to set the angle to 'zero'. It is just impossible.

@nmanovic 
I can try implementing the techique with Shift as I suggested before.

nmanovic:
> > also I tried to set the angle to 'zero'. It is just impossible.
> 
> @nmanovic I can try implementing the techique with Shift as I suggested before.

Please implement a standard option for a popular graphical editor. I believe if rotation isn't 0, it should be displayed as an attribute on the side panel as well. At least for occluded we have a button.

bsekachev:
@nmanovic 

Fixed precision. 
Implemented snapped rotation (with Shift step is 15 degrees). It is common approach (i.e. Photoshop).

> I believe if rotation isn't 0, it should be displayed as an attribute on the side panel as well

I believe on the side panel this information is totally useless (we don't display there x,y,height,width. Isn't it?
For occluded we have a button because there is no other way to switch it.

nmanovic:
> @nmanovic
> 
> Fixed precision. Implemented snapped rotation (with Shift step is 15 degrees). It is common approach (i.e. Photoshop).
> 
> > I believe if rotation isn't 0, it should be displayed as an attribute on the side panel as well
> 
> I believe on the side panel this information is totally useless (we don't display there x,y,height,width. Isn't it? For occluded we have a button because there is no other way to switch it.

I'm fine with the solution if it allows me to set precisely at least 0, 15, ... after I have rotated a box. Occluded and outside properties have corresponding shortcuts. In general it is really a good idea to have a "visual" way to setup an angle. The visual element should have a hint how to do the action using shortcuts. Also rotation is implemented as an attribute thus it is one more reason to have it on the side panel. Just believe me that it isn't useless. But I'm OK to way a feedback from our users before implementing that. It is up to you. Make a decision by yourself.

cvat/apps/dataset_manager/bindings.py
kirill-sizov: @zhiltsov-max, resolved it in Datumaro [PR 556](https://github.com/openvinotoolkit/datumaro/pull/556)

cvat/apps/engine/serializers.py
```diff
@@ -670,6 +670,7 @@ class ShapeSerializer(serializers.Serializer):
     type = serializers.ChoiceField(choices=models.ShapeType.choices())
     occluded = serializers.BooleanField()
     z_order = serializers.IntegerField(default=0)
+    rotation = serializers.FloatField(default=0, min_value=0, max_value=360)
```
nmanovic, line 673: @bsekachev , Is 360 a valid value?

nmanovic (approved):
LGTM

bsekachev:
@aschernov @TOsmanov 

Could you please update documentation when the patch is available for you on the internal server?

@dvkruchinin 

Could you please prepare some tests to cover the new code and features?

- Check that bounding boxes can be rotated
- Check interpolation, merging/splitting rotated shapes
- Check rotation with hold Shift button

TOsmanov:
> @aschernov @TOsmanov
> 
> Could you please update documentation when the patch is available for you on the internal server?

@bsekachev, Yes, sure

dvkruchinin:
> @dvkruchinin
> 
> Could you please prepare some tests to cover the new code and features?
> 
> * Check that bounding boxes can be rotated
> * Check interpolation, merging/splitting rotated shapes
> * Check rotation with hold Shift button

@bsekachev 
Sure. Tests will be prepared.

PR #3917, opened 2021-11-17
Handle attributes returned from nuclio detector

mikhail-treskin:
### Motivation and context
Currently CVAT ignores attributes info coming from nuclio detection function along with boxes. Such functionality is required for pipelined inference of several models e.g. detector->age/gender/pose recognition. 

### How has this been tested?
Tested with corresponding "..: test" targets from vscode configurations (logs may be attached if needed).
Manually checked that attributes returned from nuclio applied both in task auto annotation case and using AI tools in annotation mode for certain frame.   

### Checklist
- [ ] I have added description of my changes into [CHANGELOG](https://github.com/opencv/cvat/blob/develop/CHANGELOG.md) file **- TBD**
  https://github.com/opencv/cvat/blob/develop/README.md#documentation) accordingly - **not applicable**
- [ ] I have added tests to cover my changes - **changes can be tested with existing cases**

### License


mikhail-treskin:
@bsekachev @nmanovic could you please comment if cvat-ui version should be bumped in package.json in case of such change?

bsekachev:
@mikhail-treskin 

>  case of such change?

Hi, any case. Here, I think need to update patch version number (``npm version patch``).

cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/tools-control.tsx
```diff
@@ -1037,7 +1037,10 @@ export class ToolsControlComponent extends React.PureComponent<Props, State> {
                                     frame,
                                     occluded: false,
                                     source: 'auto',
-                                    attributes: {},
+                                    attributes: data.attributes.reduce(function(attrs_map: any, attr: any) {
+                                                                            attrs_map[attr.spec_id] = attr.value;
```
bsekachev, line 1041: as far as I understand, ``attr.spec_id`` is returned by nuclio function.
At the same time ``spec_id`` is the CVAT specific thing which additionally differs task by task. It means that nuclio function should not return these ids, only attribute names and matching should be implemented on CVAT side.

mikhail-treskin: Ok, it does make sense. And I guess the same is applicable for the changes in backend?

bsekachev: Yep. Matching better to be implemented on backend (for both methods, when call automatic annotation for one frame and for a whole task)

mikhail-treskin: The only one question which I have so far is how to avoid attribute names and spec ids matching on client side in case of one frame auto annotation. 

What  I could figure out so far regarding attributes adding, is that attrs format expected to be like `{spec_id: value}` and checked [here](https://github.com/openvinotoolkit/cvat/blob/develop/cvat-core/src/annotations-collection.js#L722). So in `runInference` method I can't create some temporary attribute basing only on name and value, and then in the backend match label+attr_name coming from frontend into spec_id.

so my current implementation of [runInference](https://github.com/openvinotoolkit/cvat/blob/1c038bb4e983b120cbdd6ec6e2583146c1958f59/cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/tools-control.tsx#L1026) looks like that:
``` typescript
runInference={async (task: any, model: Model, body: object) => {
                    try {
                        this.setState({ mode: 'detection', fetching: true });
                        const result = await core.lambda.call(task, model, { ...body, frame })
                        let mapping = {};
                        task.labels.forEach((label: any) => {
                            mapping[label.name] = {};
                            label.attributes.forEach((attr: any) => {
                                mapping[label.name][attr.name] = attr.id;
                            })
                        });
                        let i = 0;
                        const states = result.map(
                            (data: any): any =>
                                new core.classes.ObjectState({
                                    shapeType: data.type,
                                    label: task.labels.filter((label: any): boolean => label.name === data.label)[0],
                                    points: data.points,
                                    objectType: ObjectType.SHAPE,
                                    frame,
                                    occluded: false,
                                    source: 'auto',
                                    attributes: data.attributes.reduce(function(attrs_map: any, attr: any) {
                                        attrs_map[mapping[data.label][attr.name]] = attr.value;
                                        return attrs_map;
                                    }, {}),
                                    zOrder: curZOrder,
                                }),
                        );
```
Finally, my questions 😄 
* Is it ok to keep mapping of attr_name returned from nuclio to spec_id in the FE?
* If it's ok to keep mapping in the FE, which approach is more preferable, to build `mapping` object from the code on render stage or on inference stage?

nmanovic: @mikhail-treskin , could you please look how it is done for labels? For attributes we need to have mostly the same functionality. A serverless function should declare which attributes it supports (as it is done for labels). Based on the information, it is necessary map attributes per task.

Current patch cannot be applied as is to multiple tasks because spec_id will be different for different tasks.

Please let me know if you want to contribute the functionality. It can be quite complex if it is done in production quality. Also need to have a severless function which supports attributes. It will be an example and can be used for testing. I will move the patch into WIP for now.

mikhail-treskin: @nmanovic, yes the PR is definitely WIP6 just forgot to mark it properly.

Could you please have a look on changes in latest commit? as @bsekachev proposed previously i'm handling attributes returned from nuclio by name and then maps them into spec_id basing on db task data. 

> could you please look how it is done for labels? For attributes we need to have mostly the same functionality. A serverless function should declare which attributes it supports (as it is done for labels). Based on the information, it is necessary map attributes per task.

Does it presume that we also need to have functionality for attributes mapping in UI like it's done for labels?

> Also need to have a severless function which supports attributes

As an example can propose pipeline of few OMZ models like face_detection->emotion_recognition+age_gender_recognition (or only on of _recognition)

nmanovic: @mikhail-treskin , the updated patch looks much better. We are more than happy to accept any contribution from our community. Also I will be happy to understand how you are using CVAT in your company and which features are missing. Probably it can be a meeting. Let me know if you are ready to share the information.

An example with OMZ models will be great. Could you please add one or several of them to the PR as serverless functions? If you have any requests to these models, please ping @snosov1. Probably he can help somehow. He also is responsible for OTE: https://github.com/openvinotoolkit/training_extensions

It will be great to map attributes in UI in the future. But let's consider the feature as an optional recommendation. It should not block the PR. I will merge it without the functionality.

Better to decelerate attributes in yml config for a serverless function. Thus it is easy to understand what the function supports.

If you can extend the tutorial: https://openvinotoolkit.github.io/cvat/docs/manual/advanced/serverless-tutorial/, it will be great. But it is an optional recommendation as well.

mikhail-treskin: @nmanovic could you please have a look on serverless function? 
I'm not sure regarding face-detection model, tried to choose between 0206 and 0205, looks like 0206 more accurate but much slower than 0205, not sure what is more preferable in case of automatic annotation scenario.

serverless/common/openvino/model_loader.py
```diff
@@ -57,16 +50,23 @@ def infer(self, image, preprocessing=True):
...
+        inputs = self._prepare_inputs(image, preprocessing)
+        return self._net.start_async(request_id=request_id, inputs=inputs)
+
     def input_size(self):
         return self._input_layout[2:]
 
     @property
     def layers(self):
-        return self._network.layers
+        return self._network.layers
```
mikhail-treskin, line 72: @nmanovic @bsekachev btw just for your information, layers property of IENetwork was deprecated and removed in 2021.2, so maybe it makes sense to get rid of it on common ModelLoader and handle it locally in yolo function (the only one place where i found a usage)

mikhail-treskin: Here is an example of auto annotated image. age-gender looks pretty inaccurate, maybe to keep only emotions-recognition as a showcase?
Screenshot 2021-12-21 at 18 46 21

nmanovic:
@mikhail-treskin , is the PR ready for review?

mikhail-treskin:
> @mikhail-treskin , is the PR ready for review?

Yep, I guess so. I've referenced you and @bsekachev in some comments above

serverless/openvino/omz/intel/face-detection-0205/function.yaml
```diff
@@ -0,0 +1,61 @@
...
+  name: openvino-omz-face-detection-0205
+  namespace: cvat
+  annotations:
+    name: Attributed face detection
+    type: detector
+    framework: openvino
+    # attribute names have to be the same as in annotated task, otherwise values will be ignored
+    spec: |
+      [
+        { "id": 0, "name": "face", "attributes": ["age", "gender", "emotion"]}
```
nmanovic, line 11: @mikhail-treskin , thanks for the great PR. Could you please help us to polish the functionality? Each attribute should have a type and possible values. For example, when CVAT returns an attribute (e.g. GET /api/v1/tasks), it returns something like the structure below:

```
            "labels": [
                {
                    "id": 0,
                    "name": "age",
                    "attributes": [
                        {
                            "id": 0,
                            "name": "age",
                            "input_type": "number",
                            "default_value": "25",
                            "values": [
                                "0",
                                "150",
                                "1"
                            ]
                        }
                    ]
                }
            ],
```

Do you think it is valuable to have something like that here? Otherwise it is unclear which values each attribute has. Also it will be difficult to much them to a corresponding task. Basically, I should be able to take labels and attributes from the definition as is and create a task with them.

serverless/common/openvino/model_loader.py
nmanovic: @mikhail-treskin , I agree. I'm OK to upgrade the OpenVINO version and modify the code.

nmanovic:
@mikhail-treskin , could you please help us to fix eslint and pylint linters warnings for the PR?

I believe the PR is a great improvement. We definitely want to have it inside the develop branch. Could you please resolve mentioned issues and I will merge it.

mikhail-treskin:
> @mikhail-treskin , could you please help us to fix eslint and pylint linters warnings for the PR?
> 
> I believe the PR is a great improvement. We definitely want to have it inside the develop branch. Could you please resolve mentioned issues and I will merge it.

Yes, sure. Should be fixed in latest commit.

serverless/openvino/omz/intel/face-detection-0205/function.yaml
mikhail-treskin: Yes, I agree that possible values should be part of function config. 
But what about attribute type, i'm not really sure that it does make sense to store it in function config since attribute types is CVAT specific information. Maybe it make sense to add additional attributes verification on backend side, e.g. if certain attribute has type "Select" or "Radio" only one value can be accepted, and check if "Number" attribute coming from nuclio  is really convertible to number. Otherwise exception can be raised or fallback to default attribute value. 
What do you think about such an approach?

cvat/apps/lambda_manager/views.py
```diff
@@ -209,6 +219,11 @@ def invoke(self, db_task, data):
                 for item in response:
                     item["label"] = mapping.get(item["label"])
                 response = [item for item in response if item["label"]]
+            # TODO: Need to add attributes mapping similar to labels.
+            # Currently attribute is expicitely discarded if it is not decalred as supported in function config.
+            if self.attributes:
+                for item in response:
+                    item['attributes'] = [attr for attr in item.get("attributes", []) if attr['name'] in self.attributes[item['label']]]
```
mikhail-treskin, line 226: @nmanovic I came back to this code after some time and now I'm in doubt is this check really make sense.  Could you please advise?

Here I'm checking that attribute name returned from nuclio function present in `self.attributes` which is `spec` field from `function.yml` in fact (set on line 106). If attribute was not specified in function config it will be discarded. 
But it confuses me now, maybe better to check that attribute name and value returned from nuclio exist in  corresponding label in `db_task` data? 
Which behaviour will be preferable until attributes mapping doesn't implemented in UI?

serverless/openvino/omz/intel/face-detection-0205/function.yaml
nmanovic: @mikhail-treskin , `spec` section inside function.yaml is also CVAT specific. Basically I believe that CVAT labels and attributes definition and serverless labels and attributes definition should correspond to each other. Also in the future it is good to implement matching of attributes by name in UI (as it is implemented for labels). It will be necessary to understand the type of an attribute and possible values. I don't suggest implementing matching in UI for the PR. It is another story.

Proposal:
- Let's have the same definition for attributes as CVAT server support (see an example above). 
- If an attribute is supported by a serverless function, but it isn't supported by a corresponding label in CVAT, it should be ignored.
- If attributes are matched by name, it is necessary to see if we can cast it to a corresponding type. For example, any attribute can be converted to `text` type. At the same time `text` type cannot be converted to `select`. An attribute, which is matched by name but cannot be casted, should be ignored.

Thus we have a set of labels with attributes from a task. Also we have a number of labels with attributes from a serverless function. We match labels by name. After that we match attributes by name. After that we match attributes by type. During the process some labels and attributes will be filtered. The final list of labels with reduced number of attributes can be covered by our serverless function.
It is a typical case when a serverless function supports more labels and attributes when it is necessary. All of them should be ignored as it is implemented for labels.

cvat/apps/lambda_manager/views.py
```diff
@@ -98,6 +102,8 @@ def __init__(self, gateway, data):
                 "`{}` lambda function has non-unique labels".format(self.id),
                 code=status.HTTP_404_NOT_FOUND)
         self.labels = labels
+        # mapping of labels and corresponding supported attributes
+        self.attributes = {item['name']: item.get('attributes', []) for item in spec}
```
nmanovic, line 106: @mikhail-treskin , we need to keep attributes per label. For example, `person` label with `age`, `gender` attributes and `car` label with `model` and `type` attributes. We should have a separate list for each label. Not just `['age', 'gender', 'model', type'].

nmanovic, line 106: @mikhail-treskin , OK. The code looks good. Probably it can be a good idea to check that for the same label we don't have duplicated attributes (as in the code for labels above).

cvat/apps/lambda_manager/views.py
```diff
@@ -138,6 +144,10 @@ def to_dict(self):
             response.update({
                 'state': self.state
             })
+        if self.kind is LambdaType.DETECTOR:
+            response.update({
+                'attributes': self.attributes
```
nmanovic, line 149: @mikhail-treskin , Probably it is slightly better to return attributes together with labels.

mikhail-treskin: In general I agree but I was afraid that response structure changing  may broke compatibility between BE and FE in some cases which I'm not aware.

nmanovic:
@mikhail-treskin , thanks for all your time and contribution. Will you be able to continue improving the PR?

serverless/openvino/omz/intel/face-detection-0205/function.yaml
mikhail-treskin: @nmanovic Could you please have a look on logic of attributes filtering in lambda manager in latest [commit](https://github.com/openvinotoolkit/cvat/pull/3917/commits/0774a4cdaa6194a26563591edb4f80b8efad1a62)?

If it looks fine for you I'll implement same logic in FE in cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/tools-control.tsx

PR #4007, opened 2021-12-09
Fixed possible color collisions in the generated colormap

azhavoro:
### Checklist

### License

```python
#
```


cvat/apps/dataset_manager/formats/utils.py
```diff
@@ -62,18 +63,43 @@ def make_colormap(instance_data):
...
-def get_label_color(label_name, label_names):
+
+def generate_color(color, used_colors):
+    def tint_shade_color():
+        new_color = [0, 0, 0]
+        for added_color in (255, 0):
+            for factor in range(1, 10):
+                for c in range(0, 3):
+                    new_color[c] = int(color[c] + (added_color - color[c])*factor/10)
+                yield tuple(new_color)
```
nmanovic, line 74: yield tuple(map(lambda c: int(c + (added_color - c)* factor / 10), color))

cvat/apps/dataset_manager/formats/utils.py
```diff
@@ -62,18 +63,43 @@ def make_colormap(instance_data):
...
+
+    def get_unused_color():
```
nmanovic, line 73: I would recommend another algorithm to get unused colors:
- Sort all used_colors by x, y, z -> [(0, 0, 0), (10, 30, 22), (255, 0, 0)]
- Find 2 adjacent points with the biggest difference by one of coordinates: (10, 30, 22), (255, 0, 0). Because dx = 255 - 10 is the maximum and they are adjacent.
- Add a new point as ((x0 + x1) // 2, y0, z0). In our case (132, 30, 22).

cvat/apps/engine/migrations/0028_labelcolor.py
```diff
@@ -8,10 +8,10 @@ def alter_label_colors(apps, schema_editor):
 
     for task in Task.objects.all():
         labels = Label.objects.filter(task_id=task.id).order_by('id')
-        label_names = list()
+        label_names = {}
```
nmanovic, line 11: Do you really need here map name->color? Not just a set of colors?

cvat/apps/dataset_manager/formats/utils.py
```diff
@@ -61,19 +62,37 @@ def make_colormap(instance_data):
...
+        for added_color in (255, 0):
+            for factor in range(1, 10):
+                yield tuple(map(lambda c: int(c + (added_color - c) * factor / 10), color))
 
-def get_label_color(label_name, label_names):
+    def get_unused_color():
+        sorted_colors = sorted(used_colors)
+        max_dist = max(zip(sorted_colors, sorted_colors[1:]), key=lambda c_pair: c_pair[1][0] - c_pair[0][0])
+
+        return ((max_dist[0][0] + max_dist[1][0]) // 2, max_dist[0][1], max_dist[0][2])
```
kirill-sizov, line 75: Hi, @azhavoro, Is it the final version of this patch? If it is, then I have a question:
Will this solution work correctly if all used colors have a same R-value? For example in case:
```
used_colors = [(100, 0, 0), (100, 50, 0), (100, 0, 50)]
```
this code will return `(100, 0, 0)` but this color is already use. I'm not sure that such situation is very likely, but it seems to me that we can extend this algorithm to other components of the color as well.

azhavoro: Agree, fixed.

cvat/apps/dataset_manager/formats/utils.py
```diff
@@ -61,19 +63,40 @@ def make_colormap(instance_data):
...
+    label_colors = tuple(hex2rgb(c) for c in label_colors if c)
+    used_colors = set(itertools.chain(predefined.values(), label_colors))
     normalized_name = normalize_label(label_name)
 
     color = predefined.get(normalized_name, None)
-    name_hash = int.from_bytes(blake2s(normalized_name.encode(), digest_size=4).digest(), byteorder="big")
-    offset = name_hash + normalized_names.count(normalized_name)
-
     if color is None:
-        color = get_color_from_index(DEFAULT_COLORMAP_CAPACITY + offset)
```
ActiveChooN, line 75: this const is not used, so should be deleted

kirill-sizov (approved):
Looks good to me

nmanovic (approved):
LGTM

azhavoro:
@dvkruchinin Hi, could you prepare a test for this case?

dvkruchinin:
> @dvkruchinin Hi, could you prepare a test for this case?

Hi, sure. A test will be prepared.

PR #4039, opened 2021-12-16
Update workflows for linters.

dvkruchinin:
### Motivation and context

At the moment, when checking with linters, additional pages with a list of files in the pull request are not taken into account. This patch fixes this.

### Checklist

### License

```python
#
```

### DIFF
--- a/.github/workflows/bandit.yml
+++ b/.github/workflows/bandit.yml
@@ -7,17 +7,24 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: Run checks
+        env:
+          PER_PAGE: 100
+          PR_URL: "https://api.github.com/repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}"
+          COUNT_CHANGED_FILES: $(curl -s -X GET -G $PR_URL | jq .changed_files)
+          COUNT_PAGES: $((($COUNT_CHANGED_FILES / $PER_PAGE) + ($COUNT_CHANGED_FILES % $PER_PAGE > 0)))
         run: |
-          URL="https://api.github.com/repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/files"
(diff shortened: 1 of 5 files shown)


dvkruchinin:
For example: https://github.com/dvkruchinin/cvat/pull/229

.github/workflows/bandit.yml
```diff
@@ -8,26 +8,32 @@ jobs:
...
       - name: Run checks
         run: |
-          URL="https://api.github.com/repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/files"
-          PR_FILES=$(curl -s -X GET -G $URL | jq -r '.[] | select(.status != "removed") | .filename')
-          for files in $PR_FILES; do
-            extension="${files##*.}"
-            if [[ $extension == 'py' ]]; then
-              changed_files_bandit+=" ${files}"
-            fi
+          PER_PAGE=100
```
nmanovic, line 11: @dvkruchinin , will https://github.com/marketplace/actions/get-all-changed-files work?

dvkruchinin: I will check. Thank you.

dvkruchinin: @nmanovic 
Yes. It looks more interesting. But renamed files are not taken into account here. Do I need to include them in the check?

dvkruchinin: Files with the ```renamed``` status also need to be checked. Since if the file is changed and renamed, then in the API it has the status ```renamed```. And in the current situation, it will not be checked. I will finalize the PR.

nmanovic: @dvkruchinin , if a file is only renamed, it is fine to skip it. If it is renamed and modified and only in `renamed` group, need to address the case.

dvkruchinin: Unfortunately there is no ```renamed and modified``` status.

dvkruchinin: Added ```renamed``` files to check.

nmanovic (approved):
LGTM

PR #4062, opened 2021-12-22
Added support of ellipses

### LINKED ISSUE #579: Feature Request: Circle/Ellipse-shape annotation
Hi,

I am trying to label some circular objects. It would be nice if CVAT supports ellipse and/or circle annotation. 

The following shapes would be useful:
- Center circle (two points: center and point on circle)
- Three-point circle (three points on circle)
- Center ellipse (three points: center, long axis point, and short axis point)

Thank you!

bsekachev:
### Motivation and context
Resolved #579

ellipses

TODOs:
- [x] Export import to some formats (<ellipse .. /> in CVAT formats, and masks if the format supports them)
- [x] Add several API tests to cvat-core
- [x] Support new shape type in filters

### Checklist

### License

```python
#
```


nmanovic:
@bsekachev , should we keep the shape size/position unchanged if some part of the shape is outside of a frame? Otherwise it is impossible to annotate some objects which are particularly outside of a frame.

bsekachev:
@nmanovic 

For now for all shapes where it is possible we cut shape if it is out of image. Should we cut ellipses or not is it a question. Probably we should (to keep consistensy), but we also need to add an ability to configure this behaviour for all the shapes.

nmanovic:
> @nmanovic
> 
> For now for all shapes where it is possible we cut shape if it is out of image. Should we cut ellipses or not is it a question. Probably we should (to keep consistensy), but we also need to add an ability to configure this behavior for all the shapes.

What is the major difference with rotated bounding boxes? From my point of view an ellipse shape and a rotated bounding box  are very similar. I agree that it is better to configure the behavior but for some shapes like a rotated bounding box or an ellipse it is impossible to cut them without conversion to a polygon. Also could you please give me an example when the implemented behavior makes sense?

bsekachev:
@ActiveChooN @klakhov @zhiltsov-max 

Guys, could you please take a look at this patch?

cvat/apps/dataset_manager/formats/transformations.py
```diff
@@ -32,3 +35,16 @@ def transform_item(self, item):
                 z_order=ann.z_order))
 
         return item.wrap(annotations=annotations)
+
+class EllipsesToMasks(ItemTransform):
+    @staticmethod
+    def convert_ellipse(ellipse, img_h, img_w):
+        cx, cy, rightX, topY = ellipse.points
+        rx, ry = rightX - cx, cy - topY
```
zhiltsov-max, line 43:
```suggestion
        rx = rightX - cx
        ry = cy - topY
```
I don't recommend using multiple expressions in a line, because it makes reading worse. A matter of taste, though.

cvat/apps/dataset_manager/formats/transformations.py
```diff
@@ -32,3 +35,16 @@ def transform_item(self, item):
...
+        shape = (cx - rx, cy - ry, cx + rx, cy + ry)
+        mask = Image.new('1', (img_w, img_h))
+        drawer = ImageDraw.Draw(mask)
+        drawer.ellipse(shape, fill = 1)
```
zhiltsov-max, line 47: I suspect, opencv may be a bit faster, did you try to check it?

cvat/apps/dataset_manager/formats/transformations.py
```diff
@@ -32,3 +35,16 @@ def transform_item(self, item):
                 z_order=ann.z_order))
 
         return item.wrap(annotations=annotations)
+
+class EllipsesToMasks(ItemTransform):
```
zhiltsov-max, line 39:
```suggestion
class EllipsesToMasks:
```
There is no point in making it inherited now.

zhiltsov-max (dismissed):

ActiveChooN:
Hi,
Ellipse is not clipped for the size of the frame as you show in the gif (for all shapes), but in the develop it works fine
изображение
Will ellipses support rotation as bounding boxes? Looks like it can be helpful in some cases.

bsekachev:
@ActiveChooN 

> Ellipse is not clipped for the size of the frame as you show in the gif (not only ellipses btw), but in the develop it works fine

It was changed according to @nmanovic ideas. GIF just is not updated.

cvat/apps/dataset_manager/formats/transformations.py
bsekachev: I'll investigate this point

ActiveChooN (approved):
LGTM

nmanovic (changes requested):
@bsekachev , I will recommend to draw an ellipse as it is done by Photoshop and gimp by default. Looks like it is a better way for our scenarios as well. Draw an ellipse and try to annotate it using the current approach when an ellipse is drawing from a center. It will be difficult. Better to fit an ellipse into a virtual bounding box. Thus the interface can be similar to a bounding box but instead of a box, users will get an ellipse fitted into the box.

Need to say that graphical editors support both methods. The current behaviors is supported using ALT+CTRL. But I'm not sure that we need the second method. In real life it is very difficult to find a center of an object.

nmanovic:
@bsekachev , another idea which is similar for rotated bounding boxes. Rotated shapes better to annotate using 4 points.Thus you can draw a rotated bounding box or fit an ellipse into it.

BTW, how to rotate ellipse?

cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/draw-polyline-control.tsx
```diff
@@ -22,24 +22,20 @@ export interface Props {
...
-        {
-            className: 'cvat-draw-polyline-control cvat-active-canvas-control',
-            onClick: (): void => {
-                canvasInstance.draw({ enabled: false });
-            },
-        } :
-        {
-            className: 'cvat-draw-polyline-control',
-        };
+    const dynamcPopoverPros = isDrawing ? {
```
nmanovic, line 25: @bsekachev , `dynamc` typo?

klakhov:
@bsekachev great job! What do you think about quick way to draw a circle? Usually we can hold `shift` using ellipse tool to draw circles
circle

cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/draw-polyline-control.tsx
bsekachev: Yes, fixed

bsekachev:
@nmanovic 

> I will recommend to draw an ellipse as it is done by Photoshop

There was default way to draw ellipses provided by the svg library. Now I redesigned it according to photoshop default way of drawing. 

> BTW, how to rotate an ellipse?

Actually it was disabled because nobody asked us about rotation of ellipses. Now, If you wish, I enabled it. 

> another idea which is similar for rotated bounding boxes.

Agree, but it is a different feature where additional design stage is necessary for both bounding boxes and ellipses.

cvat/apps/dataset_manager/formats/transformations.py
bsekachev: Yes, OpenCV about 2.5 times faster, I updated the code.

bsekachev:
@klakhov 

That's interesting, but as far as I remember there were not such requests for bounding boxes for years. Maybe nobody needs it? And if we decide to implement the feature, it is probably better to do it for both bounding boxes and ellipses.

nmanovic:
> @klakhov
> 
> That's interesting, but as far as I remember there were not such requests for bounding boxes for years. Maybe nobody needs it? And if we decide to implement the feature, it is probably better to do it for both bounding boxes and ellipses.

@bsekachev , @klakhov ,

I agree that it is extremely rare case when you need circle or square. We are annotating real objects, not draw something on the canvas.

P.S. 'bounding box', 'ellipse' should have mostly the same implementation. The only difference is representation (how we are drawing it and export to a file).

nmanovic (approved):
LGTM

bsekachev:
Hi, @dvkruchinin 

Could you please add some tests to ellipses?
Drawing shapes, tracks / rotation / interpolation

dvkruchinin:
> Hi, @dvkruchinin
> 
> Could you please add some tests to ellipses? Drawing shapes, tracks / rotation / interpolation (I have already added a command to draw ellipses actually and updated test with statistics).

Hi @bsekachev,
Sure. A test will be prepared.

TOsmanov:
> @TOsmanov @aschernov
> 
> Guys, could you add a piece of documentation about the new supported type of shape?

Hi @bsekachev,
Yes, the documentation will be updated

mshz88:
please help me, why I can't find the ellipse shape in the web version? how to add the ellipse shape?

bsekachev:
@mshz88 

If you speak about cvat.org, the server was not updated yet. We are working on this, but it takes long time because of significant changes in CVAT. I hope the process will be finished the next week.

mshz88:
Thank you, @bsekachev CVAT is the best tool among others with many features. We are looking forward to seeing the ellipse shape in the updated version.

PR #4154, opened 2022-01-12
Fixed bug: tus failed to upload chunk at offset 0 over https

klakhov:
### Motivation and context

Currently if we use tus upload for cvat tasks over https error like this is displayed:

That happens because of incorrect protocol in chunk upload location returned by server. Client makes https request that is forwarded by traefik to django via http. Thats why django's `request.build_absolute_uri` returns location with http protocol.

There are several ways to let django know that client request was made over https:

- Add custom django configuration setting to use over https
- Add additional header to incoming request using [traefik middlewares](https://doc.traefik.io/traefik/middlewares/http/headers/) based on environment
- Modify `Location` header based on environment using [traefik plugin](https://github.com/tomMoulard/htransformation)
- Use django's [proxy_ssl_header setting](https://docs.djangoproject.com/en/2.2/ref/settings/#secure-proxy-ssl-header) as currently traefik provides  [X-Forwarded-Proto header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto) to determine original client protocol

I've used the last option. Based on `SECURE_PROXY_SSL_HEADER` setting, if `X-Forwarded-Proto` is correct, then django knows that request was sent via https.

### How has this been tested?

Deployed locally and set secure header to http, `X-Forwarded-Proto` was http, then django knew that connection was secure and returned https protocol location.

### Checklist

- [ ] ~~I have updated the [documentation](
- [ ] ~~I have linked related issues ([read github docs](

### License

```python
#
```


klakhov:
@azhavoro Could you look on these settings please? Is it correct that we don't need to change any traefik settings to keep `X-Forwarded-*` headers visible over https as we do with http?

CHANGELOG.md
```diff
@@ -174,6 +174,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Using combined data sources (directory and image) when create a task (<https://github.com/openvinotoolkit/cvat/pull/3424>)
 - Creating task with labels in project (<https://github.com/openvinotoolkit/cvat/pull/3454>)
 - Move task and autoannotation modals were invisible from project page (<https://github.com/openvinotoolkit/cvat/pull/3475>)
+- Fixed tus upload error over https (<https://github.com/openvinotoolkit/cvat/pull/4154>)
```
bsekachev, line 177: You put it to 1.5.0 release section

bsekachev: The closest unreleased 2.0.0

klakhov: Fixed

cvat/settings/base.py
```diff
@@ -515,3 +515,6 @@ def add_ssh_keys():
 TUS_MAX_FILE_SIZE = 26843545600 # 25gb
 TUS_DEFAULT_CHUNK_SIZE = 104857600  # 100 mb
 
+# used to make request secure if X-Forwarded-Proto: 'https' header is specified by our proxy
+# https://docs.djangoproject.com/en/2.2/ref/settings/#secure-proxy-ssl-header
```
azhavoro, line 519: I think the following link will also be helpful https://github.com/traefik/traefik/blob/master/docs/content/getting-started/faq.md#what-are-the-forwarded-headers-when-proxying-http-requests (or https://doc.traefik.io/traefik/getting-started/faq/#what-are-the-forwarded-headers-when-proxying-http-requests)

klakhov: Done

azhavoro (approved):

PR #4239, opened 2022-01-25
API versioning using accept header

Marishka17:
### Motivation and context

This PR contains a new API versioning scheme using [AcceptHeaderVersioning](https://www.django-rest-framework.org/api-guide/versioning/#acceptheaderversioning)
I've tried to implement an approach like [API versioning for HTTP REST interfaces (best practices)](https://gist.github.com/kirushik/cbe130d2bd4847020c6fe5e953e92618#use-versioned-vnd-content-types) (`application/vnd.cvat.v1.0+json`) but in this case, I couldn't fix unavailable drf-spectacular documentation. I guess It may be not supported in the current [implementation](https://github.dev/tfranzel/drf-spectacular/blob/803c5c12ee7e0ecfce79c4875ff2dd96e7e060bf/drf_spectacular/plumbing.py#L905-L906)
### Checklist

### License

```python
#
```


.github/workflows/schedule.yml
```diff
@@ -16,10 +16,10 @@ jobs:
           DJANGO_SU_NAME: "admin"
           DJANGO_SU_EMAIL: "admin@localhost.company"
           DJANGO_SU_PASSWORD: "12qwaszx"
-          API_ABOUT_PAGE: "localhost:8080/api/v1/server/about"
+          API_ABOUT_PAGE: "localhost:8080/api/server/about"
         run: |
           docker-compose -f docker-compose.yml -f docker-compose.dev.yml -f ./tests/docker-compose.email.yml -f tests/docker-compose.file_share.yml -f components/serverless/docker-compose.serverless.yml up -d --build
-          /bin/bash -c 'while [[ "$(curl -s -o /dev/null -w ''%{http_code}'' ${API_ABOUT_PAGE})" != "401" ]]; do sleep 5; done'
+          /bin/bash -c 'while [[ $(curl -s -o /dev/null -H "Accept: application/vnd.cvat+json; version=1.0" -w "%{http_code}" ${API_ABOUT_PAGE}) != "401" ]]; do sleep 5; done'
```
nmanovic, line 22: @Marishka17 , what do you think if by default we will use the latest version of API? Thus if we don't specify accept header, DRF will use the latest one. I know that the link https://github.com/interagent/http-api-design/blob/master/en/foundations/require-versioning-in-the-accepts-header.md says that it is a bad practice. But GitHub uses the strategy: https://docs.github.com/en/rest/overview/media-types

CHANGELOG.md
```diff
@@ -35,6 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Users don't have access to a task object anymore if they are assigneed only on some jobs of the task (<https://github.com/openvinotoolkit/cvat/pull/3788>)
 - Different resources (tasks, projects) are not visible anymore for all CVAT instance users by default (<https://github.com/openvinotoolkit/cvat/pull/3788>)
+- API versioning scheme: using accept header versioning instead of namespace versioning (<https://github.com/openvinotoolkit/cvat/pull/4239>)
```
nmanovic, line 38: did we use URL versioning? https://www.django-rest-framework.org/api-guide/versioning/#urlpathversioning

cvat/apps/dataset_manager/tests/test_formats.py
```diff
@@ -36,24 +37,30 @@ def generate_image_file(filename, size=(100, 100)):
     return f
 
 class ForceLogin:
-    def __init__(self, user, client):
+    def __init__(self, user, client, version=settings.BACKEND_VERSIONS.V1_0):
```
nmanovic, line 40: @Marishka17 , One more problem if we don't have a default version. I don't think that it is good idea to have BACKEND_VERSIONS.V1_0 and ACCEPT_HEADER_TEMPLATE in settings.

cvat/apps/engine/versioning.py
```diff
@@ -0,0 +1,26 @@
...
+from rest_framework.exceptions import NotAcceptable
+
+class CustomAcceptHeaderVersioning(AcceptHeaderVersioning):
+    """
+    GET /api/something/ HTTP/1.1
+    Host: localhost:8080
+    Accept: application/vnd.cvat+json; version=1.0
+    """
+
+    def determine_version(self, request, *args, **kwargs):
```
nmanovic, line 15: @Marishka17 , Do we need the method if we have a default version?

cvat/settings/base.py
```diff
@@ -154,9 +169,10 @@ def add_ssh_keys():
         # Don't try to use URLPathVersioning. It will give you /api/{version}
         # in path and '/api/docs' will not collapse similar items (flat list
         # of all possible methods isn't readable).
-        'rest_framework.versioning.NamespaceVersioning',
+        'cvat.apps.engine.versioning.CustomAcceptHeaderVersioning',
     # Need to add 'api-docs' here as a workaround for include_docs_urls.
-    'ALLOWED_VERSIONS': ('v1', 'api-docs'),
+    'ALLOWED_VERSIONS': (*BACKEND_VERSIONS.list(), 'api-docs', 'downloading'),
```
nmanovic, line 174: Let's try to remove `api-docs` and `downloading` from the list.

site/content/en/docs/contributing/rest-api-design.md
```diff
@@ -36,7 +36,7 @@ Common scheme for our REST API is `<VERB> [namespace] <objects> <id> <action>`.
 - Allow filtering, sorting, and pagination
 - Maintain good security practices
 - Cache data to improve performance
-- Versioning our APIs (e.g. `/api/v1`, `/api/v2`). It should be done when you
+- Versioning our APIs (e.g. `/api`, `/api/v2`). It should be done when you
```
nmanovic, line 39: @Marishka17 , need to update the statement. It looks wrong now (e.g., `/api/v2`)

tests/cypress/support/commands_organizations.js
```diff
@@ -31,9 +33,10 @@ Cypress.Commands.add('createOrganization', (organizationParams) => {
 Cypress.Commands.add('deleteOrganizations', (authResponse, otrganizationsToDelete) => {
     const authKey = authResponse.body.key;
     cy.request({
-        url: '/api/v1/organizations?page_size=all',
+        url: '/api/organizations?page_size=all',
         headers: {
             Authorization: `Token ${authKey}`,
+            Accept: acceptHeader,
```
nmanovic, line 39: @Marishka17 , @bsekachev , should UI work with the latest version (=default) of REST API always? Thus we can skip acceptHeader for tests and UI.

tests/rest_api/test_0001_users_api.py
```diff
@@ -12,12 +12,14 @@
 
 def test_non_admin_cannot_see_others():
     for username in ['dummy1', 'worker1', 'user1', 'business1']:
-        response = requests.get(config.get_api_url('users'), auth=(username, config.USER_PASS))
+        response = requests.get(config.get_api_url('users'), auth=(username, config.USER_PASS),
```
nmanovic, line 15: @Marishka17 , tests should test all versions. Thus if we have a default version, right now it will be enough. Let's discuss. Really want to find the right approach.

utils/cli/cli.py
```diff
@@ -34,6 +34,8 @@ def main():
     with requests.Session() as session:
         api = CVAT_API_V1('%s:%s' % (args.server_host, args.server_port), args.https)
         cli = CLI(session, api, args.auth)
+        #TODO add version indication
+        session.headers.update({'Accept': 'application/vnd.cvat+json; version=1.0'})
```
nmanovic, line 38: Again, probably CLI should work with the latest (default version) of API.

CHANGELOG.md
nmanovic: OK. It is namespace versioning: <i>to the client, this scheme is the same as URLPathVersioning. The only difference is how it is configured in your Django application, as it uses URL namespacing, instead of URL keyword arguments.</i>

.github/workflows/schedule.yml
Marishka17: @nmanovic ,  Also rest_framework [documentation](https://www.django-rest-framework.org/api-guide/versioning/) doesn't say that this is a bad practice, let's use the default version. In this case, I guess we can miss `accept header` in client/tests. I'll update it.

Marishka17:
@nmanovic , I've updated the API version to 2.0 as you said yesterday (I didn't do it before because I believed that we should have done tasks with server code redesigning and separating it to standalone applications).
Also, I've added a test for checking the version for `server/about` endpoint.
I've found several ways to version codebase (simply extend serializer, an approach like migrations, and others) and I guess we should discuss this moment with the team to choose the best approach for us. So It should be another task.

cvat/settings/base.py
```diff
@@ -151,12 +155,10 @@ def add_ssh_keys():
...
     ],
     'DEFAULT_VERSIONING_CLASS':
-        # Don't try to use URLPathVersioning. It will give you /api/{version}
-        # in path and '/api/docs' will not collapse similar items (flat list
-        # of all possible methods isn't readable).
-        'rest_framework.versioning.NamespaceVersioning',
-    # Need to add 'api-docs' here as a workaround for include_docs_urls.
-    'ALLOWED_VERSIONS': ('v1', 'api-docs'),
+        'rest_framework.versioning.AcceptHeaderVersioning',
+    'ALLOWED_VERSIONS': ('1.0', '2.0'),
```
nmanovic, line 159: @Marishka17 , do we support 1.0?

nmanovic (approved):
LGTM

Marishka17:
@nmanovic, Should we also [register](https://www.iana.org/form/media-types) vnd MIME type?

nmanovic:
@Marishka17 , we should. Need to contact OSPDT.

Issue #4268, opened 2022-02-01
Issue while using tool.


bhargav-sudo:
Hi, Facing this issue while using the CVAT tool.

![Screenshot from 2022-02-01 11-39-26](https://user-images.githubusercontent.com/57169077/151921635-41feb4da-ae55-4d42-a11b-4f57f8a2702a.png)

nmanovic:
@bhargav-sudo , could you please provide more details? I will close the issue. Try to fill the template. It is big, but information about CVAT version, your steps, how to reproduce is very critical to understand what is going on.

bhargav-sudo:
Hi @nmanovic 

My CVAT version. 

Server version : 1.7 
Core version : 3.17.0
Canvas version : 2.8.0
UI version : 1.25.0 

I installed through the documentation of CVAT and created user too. After doing all these things. When I opened the tool it says the above error.

rdutta1999:
I am facing the same issue. I am using the [Cvat web portal](https://cvat.org/) .

It was working fine yesterday, however, after I logged in today a few hours prior, I noticed multiple warnings (at the top-right of the page), all saying something along the lines of `reviewer object is removed`.

The layout of the Task details page has also changed.

- New categories in `Stage`: `validation`, `acceptance`.
- Gone is the `reviewer` column where we could add people to review the tasks.
- A new `State` column has been added.

Another important issue is, under `Assignee`, the old names are present but the respective individuals can't see the tasks on their Tasks page. I can't seem to assign other jobs to other people (other than myself). I type their username but I can't select their name (previously, typing the name caused them to appear in a dropdown sort of menu, and I could choose it).

azhavoro:
> I am facing the same issue. I am using the [Cvat web portal](https://cvat.org/) .
> 
> It was working fine yesterday, however, after I logged in today a few hours prior, I noticed multiple warnings (at the top-right of the page), all saying something along the lines of `reviewer object is removed`.
> 
> The layout of the Task details page has also changed.
> 
> * New categories in `Stage`: `validation`, `acceptance`.
> * Gone is the `reviewer` column where we could add people to review the tasks.
> * A new `State` column has been added.
> 
> Another important issue is, under `Assignee`, the old names are present but the respective individuals can't see the tasks on their Tasks page. I can't seem to assign other jobs to other people (other than myself). I type their username but I can't select their name (previously, typing the name caused them to appear in a dropdown sort of menu, and I could choose it).

[...]

PR #4271, opened 2022-02-01
Add a warning about restore DB

nmanovic:
An user tries to restore old DB backup for a new CVAT version:

https://github.com/openvinotoolkit/cvat/issues/4268#issuecomment-1026678433

### DIFF
--- a/site/content/en/docs/administration/advanced/backup_guide.md
+++ b/site/content/en/docs/administration/advanced/backup_guide.md
@@ -57,10 +57,10 @@ cvat_data.tar.bz2  cvat_db.tar.bz2  cvat_events.tar.bz2
 
 ## How to restore CVAT from backup
 
-__Warning: use exactly the same CVAT version to restore DB. Otherwise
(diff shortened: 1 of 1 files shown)


site/content/en/docs/administration/advanced/backup_guide.md
```diff
@@ -57,6 +57,11 @@ cvat_data.tar.bz2  cvat_db.tar.bz2  cvat_events.tar.bz2
 
 ## How to restore CVAT from backup
 
+__Warning: use exactly the same CVAT version to restore DB. Otherwise
+it will not work because between CVAT releases the layout of DB can be
+changed. You always can upgrade CVAT later. It will take care to migrate
+your data properly internally.__
+
```
nmanovic, line 64:
```suggestion
**Warning: use exactly the same CVAT version to restore DB. Otherwise
it will not work because between CVAT releases the layout of DB can be
changed. You always can upgrade CVAT later. It will take care to migrate
your data properly internally.**

```

PR #4295, opened 2022-02-06
REST API tests for GET jobs + GET jobs/ID/annotations

nmanovic:
### Checklist

### License

```python
#
```


nmanovic:
@kirill-sizov , I will merge the PR to unblock others. Please review it and share your comments next week.

tests/rest_api/test_0100_jobs.py
```diff
@@ -0,0 +1,170 @@
...
+            job_staff = get_job_staff(job, tasks, projects)
+            jid = str(job['id'])
+
+            for user in users:
+                if user['id'] in job_staff | org_staff:
+                    self._test_get_job_annotations_200(user['username'],
+                        jid, annotations['job'][jid], **kwargs)
+                else:
+                    self._test_get_job_annotations_403(user['username'],
+                        jid, **kwargs)
```
kirill-sizov, line 170: I see that this test can do more than one test for a similar test case, for example for test case (worker, not job staff) the function `_test_get_job_annotations_403` will be call more than once. I think from IAM testing point of view we have no reason in testing similar action for all users with same privilege/memberships. What do you think, @nmanovic?

nmanovic: @kirill-sizov , I agree. If you see an easy way to reduce redundancy, I'm OK with the approach.

tests/rest_api/test_0100_jobs.py
```diff
@@ -0,0 +1,170 @@
+# Copyright (C) 2021 Intel Corporation
+#
+# SPDX-License-Identifier: MIT
```
kirill-sizov, line 3: The name of this file led me to think why do we use numbers in filenames? If these numbers are not needed then I suggest to follow pattern `test_{resource_name}.py` or if we have some reasons to keep this numbers in filenames than it's better to describe (in README) a convention of assignment of this numbers.

PR #4319, opened 2022-02-10
Jobs page: advanced filtration and implemented sorting

bsekachev:
Resolve https://github.com/openvinotoolkit/cvat/issues/4385

### Motivation and context

**Sorting:**
image

**Quick filters:**
image

**Filter builder**
image

**Recent filters:**
image

@nmanovic  The implementation is buggy for now and not fully implemented yet, but it can send json logic requests to the server.
Example request (see query params):

```
filters: {"and":[{"in":[{"var":"state"},["in progress","new"]]},{"==":[{"var":"dimension"},"2d"]},{"<=":[5,{"var":"id"},500]},{"in":["Test task name",{"var":"task_name"}]},{"in":["Test project name",{"var":"project_name"}]}]}
```

- [x] keep state while fetching
- [x] colorify applied filters
- [x] implement server filtering
- [x] timedate before/after, setup timedate format
- [x] reduce code duplication
- [x] add sorting
- [x] test and debug

### Checklist

### License

```python
#
```


bsekachev:
@nmanovic 

Looks like everything works (I didn't test ``datetime`` yet, need to implement it on UI first).
But don't you think when filtering by ``assignee``, better to use username rather then user id? It is not very important from UI point of view, just my personal suggestion.

nmanovic:
> @nmanovic
> 
> Looks like everything works (I didn't test `datetime` yet, need to implement it on UI first). But don't you think when filtering by `assignee`, better to use username rather then user id? It is not very important from UI point of view, just my personal suggestion.

I agree, just need to map `assignee` to `assignee__username` lookup field. When you are ready with UI, I will improve the server part as well. Several comments:

- Better to rename the query parameter to `filter`
- Need to add a quick button to reset filters
- Is it possible to make all button on the filtering dialog visible from the beginning? An empty yellow box without any buttons looks confusing and agly.
- It will be better to choose a predefined filter and be able to edit it in the dialog. On your mockups the button was inside the popup dialog as far as I remember.
- Also the filter encoding isn't readable. But I don't think that we can fix that. Probably a good solution could be "deepObject", but it doesn't support nested objects: https://swagger.io/docs/specification/serialization/. Another solution is https://github.com/ljharb/qs but I don't see an implementation on python.

bsekachev:
> Better to rename the query parameter to filter

Ok. 

> Need to add a quick button to reset filters

Ok. 

> Is it possible to make all button on the filtering dialog visible from the beginning? An empty yellow box without any buttons looks confusing and agly.

Possible I think with our css setup, but the library does not provide strong API to modify default styles, so, it might be broken in case of significant changes. Anyway, I'll try. 

> It will be better to choose a predefined filter and be able to edit it in the dialog. On your mockups the button was inside the popup dialog as far as I remember.

I do not think a user should change predefined filters, because they are predefined..  And about the button, inside dialog on annotation view we have "Recent filters" list. From the meeting we discussed, that for predefined list, better to have them outside for quick access.

nmanovic:
@bsekachev , some nice ideas for UI: https://devdocs.magento.com/guides/v2.4/pattern-library/filters/data-table-filters/filtering.html

nmanovic:
> @nmanovic Look at new appearance (I've updated screenshots).

When the cross (clear filter) appear and disappear, UI elements are moving and it looks strange. I believe we need to following UI ideas in the link above. The interface looks professional, known, intuitive. Of course our interface in any case will be slightly different, but need to following the concept in general.

bsekachev:
> When the cross (clear filter) appear and disappear

Let's do the button visible always, but disable it when no filters applied. 

In the link you provided filters are simple set of conditions united with AND, in our case filters are more difficult, so many design solutions from there (like for example a line with active filters) is not applicable for us because in our case these lines are multilevel and unreadable. 

image
image

cvat/apps/engine/views.py
```diff
@@ -879,35 +883,85 @@ def dataset_export(self, request, pk):
...
 
     def get_queryset(self):
         queryset = super().get_queryset()
+
         if self.action == 'list':
             perm = JobPermission.create_list(self.request)
             queryset = perm.filter(queryset)
 
-        return queryset
+        order_by = self.request.query_params.get('order_by', 'id').split(',')
```
nmanovic, line 955: @bsekachev , let me look at the functionality. I believe we should have 3 separate ways for filters as described here: https://www.django-rest-framework.org/api-guide/filtering/#api-guide

- filter - a json complex filter condition
- sort - to sort by multiple fields
- search - simple search like `assignee:word ...`

I will prefer to remove the functionality to search by a field like `assignee=nmano` directly using query parameters. Probably the functionality can be included into `search` using special syntax.

bsekachev: > I will prefer to remove the functionality to search by a field like assignee=nmano directly using query parameters. Probably the functionality can be included into search using special syntax

Totally agree :)

nmanovic:
@bsekachev ,

- Indentation of the menu should not be by center. It looks strange. Now the indentation is (`center`, `center`, `left`). I propose to change it (`right`, `right`, `left`). Also I will vote to remove the `close` button and make the menu button with a state (open/close). Thus if you click on the button it will show the drop down menu and if you click one more time, it will close it.

image

nmanovic:
@bsekachev , for visualization of filters I will not recommend to use JsonLogic. It is good for communication between the server and client, but it is not human readable format. Better to use humanStringFormat or spelFormat (https://ukrbublik.github.io/react-awesome-query-builder/)

image

nmanovic:
@bsekachev , there is no fast way to recent the filter below. Indeed it exists but it isn't intuitive (need to click clear filters). I will vote to put apply button on the same row as the `recent` drop down menu and replace close by reset button.

image

nmanovic:
> @nmanovic
> 
> > I believe if recent filters are just a list without checkbox, it will be better. Also in the case close button will be unnecessary as well.
> 
> Do you mean a user is able to choose only one option?

Yes. In 99% of cases we need just one of recent filters. Not a combination of them. Otherwise how to join them using `or`, `and`, `not`?

bsekachev:
> Indentation of the menu should not be by center.

> to remove the close button

Done. 

>  there is no fast way to reset the filter below 

Done, but not exactly as u described. Apply and reset are located in one place:
image

This way it is implemented on annotation view.

bsekachev:
> I believe if recent filters are just a list without checkbox, it will be better. Also in the case close button will be unnecessary as well.

Done. 

>  Better to use humanStringFormat or spelFormat

Done. Used queryString format which is more readable. For now we can't use spelFormat because it was introduced starting from react-awesome-query-builder@5.0 and it looks like this release is underperformed, there are significant issues event with installing it.  We could find more 3rdparty converters, but looks like query string is also Ok (and again, it is used on annotation view).

image

nmanovic:
@bsekachev , need to remember the ordering. Now it is reset after I leave the job page.

nmanovic:
@bsekachev ,

I will propose to remove "Jobs" word, highlight "Jobs" in the header, align search bar to the left. 

image

Marishka17:
@kirill-sizov, Could you please take a look at small changes in REST API tests? In this PR job order was changed and I've updated assets and have added a step with filtering jobs with annotation with shapes (because without it tests failed).

kirill-sizov:
> @kirill-sizov, Could you please take a look at small changes in REST API tests? In this PR job order was changed and I've updated assets and have added a step with filtering jobs with annotation with shapes (because without it tests failed).

I looked, tests changes looks good to me.

Marishka17:
@bsekachev, I guess, filters for users also should be updated..

bsekachev:
@Marishka17 
We do not have filters for users as far as I know, only search.

bsekachev:
@ActiveChooN 

Could you please conduct a review?

nmanovic (approved):
LGTM

bsekachev:
@TOsmanov 

Could you please add/update documentation about current implementation of Jobs page?

TOsmanov:
> @TOsmanov
> 
> Could you please add/update documentation about current implementation of Jobs page?

@bsekachev , yes, sure

PR #4328, opened 2022-02-11
fix Literal error

ftconan:
### Checklist

### License

```python
#
```
1. ImportError: cannot import name 'Literal' from 'typing'
2. python verison: 3.7.3, macOS Monterey version: 12.1
```python
try:
    from typing import Literal
except ImportError:
    from typing_extensions import Literal
```
3. This should also work for all Python versions, given that typing_extensions is installed if you're using Python 3.7 or older.

### DIFF
--- a/cvat/apps/dataset_manager/bindings.py
+++ b/cvat/apps/dataset_manager/bindings.py
@@ -10,8 +10,12 @@ from attr import attrib, attrs
 from collections import namedtuple
 from types import SimpleNamespace
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


nmanovic:
@ftconan , thanks for your contribution! We are not going to support python 3.7. We are testing our code with python 3.8. Ubuntu image installs python 3.8. Even the fix is really tiny, I don't think that we should move in the direction. It is better to say in https://openvinotoolkit.github.io/cvat/docs/contributing/development-environment/ that CVAT server supports python >=3.8 only.

I will close the PR.

PR #4412, opened 2022-03-03
[WIP] allow to add URLs as remote source for task where they redirect to the actual resource location

japrescott:
### Motivation and context
When adding a remote source URL for a task and that URL makes a redirect (e.g first auths the user and then forwards it to where the resource is located), the file wont be downloaded and not imported. By allowing to follow redirects when trying to download the resource, this limitation is removed

### How has this been tested?
Ran it locally and tested it

### Checklist

### License

- [X] I have updated the license header for each file (see an example below)

```python
#
```

### DIFF
--- a/cvat/apps/engine/task.py
+++ b/cvat/apps/engine/task.py
@@ -257,7 +257,7 @@ def _download_data(urls, upload_dir):
         job.meta['status'] = '{} is being downloaded..'.format(url)
         job.save_meta()
 
-        response = requests.get(url, stream=True)
+        response = requests.get(url, stream=True, allow_redirects=True)
         if response.status_code == 200:
             response.raw.decode_content = True
             with open(os.path.join(upload_dir, name), 'wb') as output_file:


cvat/apps/engine/task.py
```diff
@@ -257,7 +257,7 @@ def _download_data(urls, upload_dir):
         job.meta['status'] = '{} is being downloaded..'.format(url)
         job.save_meta()
 
-        response = requests.get(url, stream=True)
+        response = requests.get(url, stream=True, allow_redirects=True)
```
nmanovic, line 260: @azhavoro , do you think it introduces any security risks? Will the code above validate redirected URL?

nmanovic, line 260: @azhavoro , do you think it introduces any security risks? Will the code above validate redirected URL?

azhavoro: There is SSRF volnureability, in case of original URL response retruns redirect to localhost for example. The function above is not follow to HTTP redirects and validate only original URL based on domain/IP address.

I think there are 2 option:
1. Handle all redirects manually
2. Don't change this and allow only "direct" links

@japrescott What do you think?

japrescott: @azhavoro Good catch with SSRF. But I think the vulnerability is small;
- if a user of CVAT wants to mess with CVAT internals, it can just do so directly by entering `http://localhost:8080/somewhere` already (no need for redirecting URLS)
- if a user trusts a bad website and adds a URL which would try to do something malicious, maximally a bad image could be injected into CVAT

Manually handling redirects does not mitigate SSRF and direct links are way to limiting (think Auth, JWT, etc)

In conclusion; I would allow these changes with `allow_redirects=True`.

azhavoro: @japrescott Hi, sorry for the late response

> @azhavoro Good catch with SSRF. But I think the vulnerability is small;

I don't think so if we are talking about publicly accessed intansers

> * if a user of CVAT wants to mess with CVAT internals, it can just do so directly by entering `http://localhost:8080/somewhere` already (no need for redirecting URLS)

Can you make a get request to all internal cvat services like elasticsearch this way? But if follow_redirect is enabled it's pretty easy

> * if a user trusts a bad website and adds a URL which would try to do something malicious, maximally a bad image could be injected into CVAT

Agree, but this case is under of scope of SSRF vulnerability.

So, I think `allow_redirects=True` by default it is not a good idea.

nmanovic:
@japrescott , due to security risks, I will recommend to implement the manual handling of redirects and validate every redirected URL.

nmanovic:
@japrescott , I will close the PR for now. Don't hesitate to re-open it if you want to fix it in more secure way.

Issue #4426, opened 2022-03-05
Cannot build docker image: could not find lib geos_c


evekeen:
Steps:
1. Run `docker-compose -f docker-compose.yml -f docker-compose.dev.yml build`
2. Run `docker-compose -f docker-compose.yml -f docker-compose.dev.yml up`

```
2022-03-05 18:06:02,650 DEBG 'rqworker_default_0' stderr output:
cvat        | Traceback (most recent call last):
cvat        |   File "/home/django/manage.py", line 21, in <module>
cvat        |     execute_from_command_line(sys.argv)
cvat        |   File "/opt/venv/lib/python3.8/site-packages/django/core/management/__init__.py", line 419, in execute_from_command_line
cvat        |     utility.execute()
cvat        |   File "/opt/venv/lib/python3.8/site-packages/django/core/management/__init__.py", line 413, in execute
cvat        |     self.fetch_command(subcommand).run_from_argv(self.argv)
cvat        |   File "/opt/venv/lib/python3.8/site-packages/django/core/management/base.py", line 354, in run_from_argv
cvat        |     self.execute(*args, **cmd_options)
cvat        |   File "/opt/venv/lib/python3.8/site-packages/django/core/management/base.py", line 393, in execute
cvat        |     self.check()
cvat        |   File "/opt/venv/lib/python3.8/site-packages/django/core/management/base.py", line 419, in check
cvat        |     all_issues = checks.run_checks(
cvat        |   File "/opt/venv/lib/python3.8/site-packages/django/core/checks/registry.py", line 76, in run_checks
cvat        |     new_errors = check(app_configs=app_configs, databases=databases)
cvat        |   File "/opt/venv/lib/python3.8/site-packages/django/core/checks/urls.py", line 13, in check_url_config
cvat        |     return check_resolver(resolver)
cvat        |   File "/opt/venv/lib/python3.8/site-packages/django/core/checks/urls.py", line 23, in check_resolver
cvat        |     return check_method()
cvat        |   File "/opt/venv/lib/python3.8/site-packages/django/urls/resolvers.py", line 416, in check
cvat        |     for pattern in self.url_patterns:
cvat        |   File "/opt/venv/lib/python3.8/site-packages/django/utils/functional.py", line 48, in __get__
cvat        |     res = instance.__dict__[self.name] = self.func(instance)
cvat        |   File "/opt/venv/lib/python3.8/site-packages/django/urls/resolvers.py", line 602, in url_patterns
[...]

PR #4451, opened 2022-03-13
Added missing geos dependency

evekeen:
Added missing dependency, otherwise it won't build on macos:
https://github.com/openvinotoolkit/cvat/issues/4426


nmanovic:
@evekeen , could you please add several lines with fixed issues into CHANGELOG.md?

nmanovic (approved):

Issue #4377, opened 2022-02-22
Image quality is worse after a backup and import of a task


evekeen:
Steps:
1. Backup a task
2. Import the task from the backup on the same instance
3. Compare image quality of 2 tasks

In the imported task quality is significantly worse
![imported](https://user-images.githubusercontent.com/3822656/155171106-27be55cf-a804-498f-b692-131f26417927.png)
![original](https://user-images.githubusercontent.com/3822656/155171114-c5a8bbd0-b563-4a93-9655-42a13f949ee1.png)

evekeen:
If anyone knows where the issue might be, let me know, I'd fix it myself

nmanovic:
@evekeen , you can try to look at [export_backup](https://github.com/openvinotoolkit/cvat/blob/e8f294f673ba04bec63b124033ce4cd4d34dfe90/cvat/apps/engine/views.py#L597) and [import_backup](https://github.com/openvinotoolkit/cvat/blob/e8f294f673ba04bec63b124033ce4cd4d34dfe90/cvat/apps/engine/views.py#L586)

You can setup the development environment using these instructions: https://openvinotoolkit.github.io/cvat/docs/contributing/

evekeen:
Could not setup the environment on mac m1.
However, it looks like the quality only affects the CVAT user interface. If you export dataset from the imported task - image quality is good.

mbirkhez:
For CLI tool, I know in core.py, the image quality is hard coded to be 70 which is why whenever you import or create tasks, the images are of lower resolution. I changed it in the code to fix it. Maybe this might help you?

migu0:
Same issue here. I import tasks through the interface and need 100% quality to detect prices edges. Happy to change the source code locally, where can I adjust this setting for importing tasks? Thank you.

migu0:
Found a workaround for this if you want to import a task at 100% image quality:

1. Create a new task, select 100% under "Advanced configuration" and select all images. Import images only.
2. Once task is created, click on 3 dots on the top right corner of the task and select "Upload annotation"
3. Select file that contains both images and annotations (not sure why images need to be included since they'll be discarded). Upload annotation and wait a bit until async task of importing annotations only has been completed.

evekeen:
I've removed the downscaling locally and built my version of cvat - pull requests approval here was not worth it.
You need this change: https://github.com/opencv/cvat/pull/4452/commits/52487a50166c976ea011fa475f1dde4558a15915

PR #4452, opened 2022-03-13
Removed downscaling, which made the image quality worse in the video …

evekeen:
A naive attempt to fix image quality for imported tasks: https://github.com/openvinotoolkit/cvat/issues/4377


nmanovic:
@Marishka17 , @azhavoro , please look at comment.

evekeen:
That's actually a nasty bug, which made the whole dataset look incorrect, since the work was done with down-scaled images, while exported dataset is not downscaled. And objects can be in different locations because of the quality loss

Marishka17:
@evekeen, Hello, thanks for the contribution. I'm not sure this is the correct solution.
I was able to reproduce the error when creating a task with a *.mov video file using the use_zip_chunks=True flag.
In this case, the `ZipCompressedChunkWriter` class is used to create a compressed chunk on the server.
If we create a backup of this task and import it, the `Mpeg4CompressedChunkWriter` will be used on the server.
I guess the problem with different quality is in this.

evekeen:
@Marishka17 how do you suggest to fix  that?

Marishka17:
As far as I remember the solution in [PR](https://github.com/openvinotoolkit/cvat/pull/3056)  was implemented without saving `use_zip_chunks` in `task.json`. When a task is creating [this](https://github.dev/openvinotoolkit/cvat/blob/b7e881f6c13491b600d19b82d89387ce0fbcf3aa/cvat/apps/engine/task.py#L457) check defines a chunk type, but in [buckup](https://github.dev/openvinotoolkit/cvat/blob/b7e881f6c13491b600d19b82d89387ce0fbcf3aa/cvat/apps/engine/backup.py#L507-L508) this option is missing.
 So I guess we need to save this metadata when creating a task backup.

Marishka17:
@azhavoro, Do you have other suggestions?

evekeen:
@Marishka17 I've just tried settings that option in my `task.json`, and it was ignored on import:
`WARNING cvat.server: the following keys are dropped {'use_zip_chunks'}`
Moreover, I see in the code we actually set `use_zip_chunks` programmatically [during the import](https://github.dev/openvinotoolkit/cvat/blob/b7e881f6c13491b600d19b82d89387ce0fbcf3aa/cvat/apps/engine/backup.py#L507)

evekeen:
I've supported the `use_zip_chunks` option.
Also, I suggest keeping the initial change of downscaling, since I don't see a reason why downscale

cvat/apps/engine/backup.py
```diff
@@ -504,8 +505,9 @@ def _write_data(zip_object):
         else:
             uploaded_files = _write_data(self._file)
 
-        data['use_zip_chunks'] = data.pop('chunk_type') == DataChoice.IMAGESET
+        use_zip_chunks = data.pop('use_zip_chunks', False) or data.pop('chunk_type') == DataChoice.IMAGESET
         data = data_serializer.data
+        data['use_zip_chunks'] = use_zip_chunks
```
Marishka17, line 510: Sorry for my mistake. But I've taken a look at this one more time and these lines will be enough to fix. 
```suggestion
        use_zip_chunks = data.pop('chunk_type') == DataChoice.IMAGESET
        data = data_serializer.data
        data['use_zip_chunks'] = use_zip_chunks
```

Marishka17, line 510: Sorry for my mistake. But I've taken a look at this one more time and these lines will be enough to fix. 
```suggestion
        use_zip_chunks = data.pop('chunk_type') == DataChoice.IMAGESET
        data = data_serializer.data
        data['use_zip_chunks'] = use_zip_chunks
```

evekeen: Why? I have `chunk_type=video` there

Marishka17: So, If you keep only this changes you have `chunk_type=video` only if you didn't select `use zip chunks`. If you selected `use_zip_chunks` option `chunk_type` is `imageset`. 
In which case do you have `chunk_type=video`? Could you please check one more time?

Marishka17:
Hi, @evekeen, thanks for the contribution! Are you still going to contribute to us? If so, could you please resolve conflicts in this PR?

Marishka17:
@azhavoro, Could you please take a look at the PR? I've resolved conflicts and updated it. PR should resolve issue #4377

azhavoro (approved):

nmanovic:
@evekeen , it looks like most parts of your PR has been merged. I see now changes in the PR. Thank you for the contribution.

PR #4485, opened 2022-03-17
Tus for project dataset

klakhov:
### Motivation and context

Continue TUS integration to provide chunk upload for large dataset files in projects

### Checklist

### License

```python
#
```


cvat-core/src/server-proxy.js
```diff
@@ -612,41 +612,64 @@
...
+                        new FormData(), {
+                            params,
+                            proxy: config.proxy,
+                            headers: { 'Upload-Finish': true },
+                        });
+                    return new Promise((resolve, reject) => {
+                        async function requestStatus() {
+                            try {
+                                const response = await Axios.get(`${url}?action=import_status`, {
+                                    params,
```
bsekachev, line 650:
```suggestion
                                const response = await Axios.get(url, {
                                    params: { ...params, action: 'import_status' },
```

cvat-core/src/server-proxy.js
```diff
@@ -612,41 +612,64 @@
...
+                                    proxy: config.proxy,
+                                });
+                                if (response.status === 202) {
+                                    if (onUpdate && response.data.message !== '') {
```
bsekachev, line 654:
```suggestion
                                    if (onUpdate && response.data.message) {
```

cvat-core/src/server-proxy.js
```diff
@@ -612,41 +612,64 @@
...
-                    Axios.post(`${url}?format=${format}`, formData, {
-                        proxy: config.proxy,
-                    }).then(() => {
                         setTimeout(requestStatus, 2000);
-                    }).catch((error) => {
-                        reject(generateError(error));
                     });
-                });
+                } catch (errorData) {
+                    generateError(errorData);
```
bsekachev, line 670: Probably generated error should be raised
And we should not return null in this case.

bsekachev:
Generally patch works for me except of 2 moments:

1. When I first time tried to import dataset, I get error with CSRF token (403), see gif above. I re-loginned and was not able to reproduce it anymore. So, maybe it was not related with this patch. Let's see if somebody meet the same issue. 
2. When error occurs, it does not show any errors, probably related with the error I commented in the code (generatedError is not raised)

error

cvat-core/src/server-proxy.js
klakhov: Will such lines be better?
```
  generateError(errorData);
  throw errorData;
```
If we dont throw or return something, eslint `consistent-return` error occurs

bsekachev: Yep, looks like we do the same in other server methods.

cvat-core/src/server-proxy.js
klakhov: Done

cvat-core/src/server-proxy.js
klakhov: Done

nmanovic (approved):

PR #4495, opened 2022-03-22
Fixed recreating shape does not use active cofiguration

### LINKED ISSUE #3768: 3D Cuboid annotation for 2D images is not working as expected

### Expected Behaviour
* Cuboids should be properly foreshortened from any perspective
* Using `Shift+N` should start with the previous configuration

### Current Behaviour
* Cuboid perspective can't be adjusted
* Using `Shift+N` starts a new cuboid in "From Rectangle" instead of "From 4 points" mode

https://ior.ad/7PtF
1. Just create a cuboid using 4 points
2. Use Shift+N to re-create it, but the mode changes automatically
3. Try adjusting the perspective. Doesn't work as well

### Context
* Impossible to label objects with different fore-shortenings
* Impossible to correct errors or to re-draw in track mode using 4 points

### Your Environment

* https://cvat.org

### Next steps
You may [join our Gitter](https://gitter.im/opencv-cvat/public) channel for community support.

ActiveChooN:
### Motivation and context

Fixed #3768

### Checklist

### License

```python
#
```


bsekachev:
@ActiveChooN 

It does not solve the issue as far as I understand.
For example:

1. Create one cuboid using 4 points
2. Create another cuboid using rectangle
3. Re-create the first cuboid (Shift + N), it uses rectangle again. 

The root problem is that drawn shapes does not remember configuration they were created with.

ActiveChooN:
@bsekachev 

> It does not solve the issue as far as I understand.

In my opinion, the shape should be created in the currently active mode and not in which it was created.

bsekachev:
@ActiveChooN 

I am not sure about it. Let me provide a small example:

1. You annotate something using objects from 5 points (according to specification these objects must have 5 points).
2. You annotate then other things using objects from 3 points (according to specification these objects must have 3 points).
3. You redraw something from 5 points, but its drawing process ends after the 3rd point, why? (What should a user to do to redraw correctly? Create another object using 5 points, delete it and then redraw this object. Why do not create another one? Because the object might be a track and a user probably wants to redraw this object only on this specific frame)

Maybe this case is about restrictions in label specification, but for now we do not have any. 

So, I believe, we should try to keep drawing method if possible. BTW, the similar issue with semi-automatic tools (serverless, OpenCV)

ActiveChooN:
@bsekachev, in case we have point-related objects (polygons, polylines) we can set `numberOfPoints` to object length. In the case of other parameters, I guess it's better to keep active methods.

Marishka17:
Hi, @ActiveChooN, thanks for the contribution:)
Are you still going to contribute to us? If so, could you please update PR?

nmanovic:
@bsekachev , could you please look at the PR? If it is OK, let's merge it.

bsekachev:
Let's close this pull request for now. Need to find more common solution which would work with interactors and intelligent scissors..

Issue #738, opened 2019-09-26
Problem with upload annotation  - landscape/portrait


sbanori:
Hello,
Maybe my problem is similar like in issue #733 but I will tell you about my case.
When we have photos, some of them can have bad orientation to make bboxes on them (1bokiem.jpg).

![image](https://user-images.githubusercontent.com/52743099/65676028-7a46a880-e04f-11e9-873f-a489f2e5fd50.png)

So we have two ways:
1. Rotate 90 degrees in any program to correct orientation (like photo 1ok.jpg) and import to CVAT's task.
2. Use a build in functionality to rotate photo.

And here is a bit problem with with second way.
1. I see bad orientation photo, I can't work on them so I use rotation in CVAT menu.
2. Now I can work, make BBoxes on rotated photo.
3. When I finished I saved changes, dump annotation  and closed task's window.
4. When I open again this task, the photo is again rotated on bad orientation, but BBoxes are rotated too, so ok. Here we don't have problem yet.
5. When I will now make new task, with the same photos, and I will use Upload annotation then everything works, I see rotated photo with rotated BBoxes so still everything is OK.
6. But now finally we have problem: When I will use way nr 1 (rotate photo to correct orientation - useful to make BBoxes on them) make a task from these photos and try to upload my annotation from the same photo but with bad-orientation, then I have error:
![image](https://user-images.githubusercontent.com/52743099/65676969-1329f380-e051-11e9-8ff9-758ad4e48327.png)
which means that rotating function in CVAT's window work only optically on this session, and it not saving the new orientation in photo data.

Is this works correct?

nmanovic:
Hi @sbanori ,

Basically it does work correctly. For the tool you changed photos (at least dimensions) and try to upload old annotations for new photos. It will not work. Even you are lucky and you don't have the error you will have incorrect annotations on these new photos (they are new just because you did some transformation with them out of the tool).

You have to update annotations as well (rotate bounding boxes as you did for images). It should be trivial but we don't have scripts which will do such transformation for you.

What can be done here?
- Save your preferences per task and load the task as you did in the past (we don't have the feature yet)
- If all photos are in the same "bad" orientation you can use option "rotate images on all frames" in settings (F2).
- We can provide a module which can do some transformation for you (we don't have the feature yet)

I will close the question for now. Don't hesitate to summit a feature request

dzanaga:
[...]

PR #4529, opened 2022-03-29
EXIF orientation support

### LINKED ISSUE #1059: Handle EXIF tag in JPG images
I've noticed that CVAT rotates relatively large images that are "Portrait" (taller than they are wide) 90 degrees left or right.

There are two existing issues that seem to be somewhat related to this:
https://github.com/opencv/cvat/issues/733
https://github.com/opencv/cvat/issues/738

But neither clearly explains the problem or provides the steps to reproduce the concern, so I figured I'd submit a new issue and document and explain everything clearly.

Steps to reproduce:

1) Download and unzip the attached `images.zip` at the bottom of this post, which contains 3 images I took outside my workplace this afternoon.  All 3 are 3024 wide x 4032 high.

2) Install the latest version of CVAT, then open Chrome and start CVAT.  Make a new task, name it anything you'd like.  Add a few labels, for example I added `car `and `fence`.  Drag the 3 unzipped images into the `Click or drag files to the area` area.  Wait for the files to be listed, then choose `Submit`.  Wait for CVAT to load in the images.

3) Go back to the main CVAT `Tasks` page, for the task you just made choose `Open`, then click on `Job #x` towards the bottom left and scroll through the images, all 3 will be rotated 90 degrees.

Here is a screenshot of what my screen looks like after following these steps:

image

An additional thing I've tried is to use the `rotation` buttons towards the bottom of the CVAT UI, which will rotate the images in the UI, and then I choose `Open Menu` -> `Save Work`.  This does temporarily make the images upright as they should be, however the problem is that if I leave the task and then open it again at a later time the images go back to being rotated 90 degrees.  I've also tried `Open Menu` -> `Settings` -> `Rotate All Images`, this does not help either.

A few other things I should mention:

1) CVAT only does this to relatively large images.  For example, I tried resizing the attached images all to 768 wide x 1024 high and then importing, and the images displayed in the correct orientation.

2) The 3 images in the attached set were all rotated 90 degrees left by CVAT, however sometimes CVAT will rotate large portrait images right.  I have not been able to determine any rhyme or reason as to when CVAT rotates left vs right, however I can say that left is more common.

This issue is a major concern for our company.  We can code around it with substantial effort, however CVAT really should not be altering our images.  Please address this concern as soon as possible.

[images.zip](https://github.com/opencv/cvat/files/4068786/images.zip)

--- Edit ---

It seems that in some cases this can be a known issue for browser-based applications.  Perhaps something along the lines of the suggestions provided in the highest rated answer to this SO question should be added to CVAT:

https://stackoverflow.com/questions/28251338/jquery-image-preview-exif-rotation-issue/28843763#28843763

Where/how something like that could be added to CVAT, I'm not sure.

ActiveChooN:
### Motivation and context

Resolve #1059, #738.

This PR adds ability to parse EXIF orientation information when media is uploaded and handle it on the client.

### How has this been tested?

Manually

### Checklist

### License

```python
#
```


cvat/apps/engine/models.py
```diff
@@ -223,6 +223,7 @@ class Image(models.Model):
     frame = models.PositiveIntegerField()
     width = models.PositiveIntegerField()
     height = models.PositiveIntegerField()
+    orientation = models.PositiveIntegerField(default=1)
```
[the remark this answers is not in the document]
ActiveChooN: Do you mean we should store rotated images on the server side? In that case, there would be a mismatch in the resolution of the raw data and compressed. I don't see a big problem here, but the whole idea behind this PR was to try to not modify the original data if it's possible.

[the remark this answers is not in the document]
bsekachev: Maybe more general solution is to store received EXIF information as a JSON in our database (as a part of meta information), instead of storing only orientation?

cvat/apps/engine/media_extractors.py
```diff
@@ -90,13 +90,26 @@ def _get_preview(obj):
         else:
             preview = obj
         preview.thumbnail(PREVIEW_SIZE)
+        orientation = preview._getexif().get(274, 1) if preview._getexif() else 1
```
bsekachev, line 93: Looks like _getexif() is a private method. Using such methods outside is not good practice. 
What these magic numbers (274, 1) means is not obvious.

cvat/apps/engine/media_extractors.py
```diff
@@ -90,13 +90,26 @@ def _get_preview(obj):
...
+        if orientation in [3, 4]:
```
bsekachev, line 94: The same comment about magic number, maybe to create a corresponding enumeration?

bsekachev:
Does this patch work for context images?

cvat/apps/engine/media_extractors.py
ActiveChooN: `_getexif()` is a convenient method to get EXIF information from PIL Image. 274 is the Orientation key, it can be gotten via 
```(python)
(orientation_key,_), = filter(lambda kv: kv[1]=='Orientation', ExifTags.TAGS.items())
```
But it looks messy. Probably I will add cont with some description.

cvat/apps/engine/models.py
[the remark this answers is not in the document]
ActiveChooN: I'm not sure about this. What kind of information can be helpful for us? I think we can move to that solution if we will have such a request to parse and use other EXIF tags. But EXIF usually stores a lot of information like thumbnail and photo instruments info which is quite useless for us but costs a lot of storage.

cvat/apps/engine/media_extractors.py
nmanovic: @ActiveChooN ,  I agree that it is not a good idea to use a private method. Please use the public API.

ActiveChooN:
@bsekachev 

> Does this patch work for context images?

Yeap

@bsekachev, @nmanovic, rewrote PR to prerotated images.

cvat/apps/engine/media_extractors.py
```diff
@@ -29,6 +29,20 @@
 from cvat.apps.engine.mime_types import mimetypes
 from utils.dataset_manifest import VideoManifestManager, ImageManifestManager
 
+ORIENTATION_EXIF_TAG = 274
+
+
+class ORIENTATION:
```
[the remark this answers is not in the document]
ActiveChooN: Should be `IntEnum` for propper comparison with ints. Fixed

cvat/apps/engine/media_extractors.py
```diff
@@ -173,7 +204,11 @@ def get_image_size(self, i):
                 properties = ValidateDimension.get_pcd_properties(f)
                 return int(properties["WIDTH"]),  int(properties["HEIGHT"])
         img = Image.open(self._source_path[i])
-        return img.width, img.height
+        width, height = img.width, img.height
+        orientation = img.getexif().get(ORIENTATION_EXIF_TAG, 1)
+        if orientation > 4:
+            width, height = height, width
```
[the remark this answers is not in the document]
ActiveChooN: Done.

azhavoro (dismissed):

bsekachev:
I tried this patch using the archive below. Looks like there are issues when images are mirrored 

[EXIF_Orientation_Samples.zip](https://github.com/openvinotoolkit/cvat/files/8396960/EXIF_Orientation_Samples.zip)

Windows image viewer is able to show such images correctly:
image

ActiveChooN:
> Looks like there are issues when images are mirrored

@bsekachev, fixed

bsekachev (approved):

cvat/apps/engine/media_extractors.py
```diff
@@ -62,6 +77,27 @@ def sort(images, sorting_method=SortingMethod.LEXICOGRAPHICAL, func=None):
     else:
         raise NotImplementedError()
 
+def image_size_within_orientation(img: Image):
+    orientation = img.getexif().get(ORIENTATION_EXIF_TAG, 1)
```
nmanovic, line 81:
```suggestion
    orientation = img.getexif().get(ORIENTATION_EXIF_TAG, ORIENTATION.NORMAL_HORIZONTAL)
```

utils/dataset_manifest/core.py
```diff
@@ -194,14 +194,18 @@ def __iter__(self):
             if idx in self.range_:
                 image = next(sources)
                 img = Image.open(image, mode='r')
+                orientation = img.getexif().get(274, 1)
```
nmanovic, line 197: @Marishka17 , probably you want to look at these changes and test them.

Marishka17 (approved):

utils/dataset_manifest/core.py
Marishka17: @nmanovic, Everything works fine.

nmanovic (approved):

PR #4543, opened 2022-04-04
Added paint brush tools

### LINKED ISSUE #1849: Paint Brush like to tool to fix the auto-annotated segmentation regions

### Expected Behaviour
A paint brush like tool for selecting segmentation region to correct automatic annotations. 

### Current Behaviour
The annotation points have to be moved around to fix the generated annotations.

### Context
The segmentation annotations that were generated, have points. If I need to fix the generated annotations, I will have to move the points around to fit the expected annotation region. Since a generated polygon has multiple points, it's hard to manage them. Having a paint brush like feature will help in fixing the segmentation annotations easy.

### LINKED ISSUE #4868: Empty gaps during semantic segmentation labelling
- [ X] Read/searched [the docs](https://github.com/cvat-ai/cvat/tree/master#documentation)
- [ X] Searched [past issues](/issues)

Good Day,
I am using DeepLabv3 Pytorch model for a semantic segmentation task and I would like to integrate this model to the backend of CVAT so that I can perform semi-automated labelling on new images. Let's just say the trained DeepLabv3 model gives an RGB-Mask output (same height and width as input). If I need to use this prediction in CVAT, I need to convert from RGB pixel-mask to polygon-mask. In order to find the polygons, we perform OpenCV's `findContours` function. But since these polygons are approximated, we'll have gaps in between them as shown below.
image

Sometimes these gaps could be so small that the human labeler might not notice them during validation. This is obviously a problem when training a new model with such labels. Semantic segmentation models expect all the pixels to have a class assigned to them. This is not the case here. 

My question is, does CVAT have any function that I could use in order to prevent these gaps from appearing when converting from pixel-based mask to polygon-based mask? If not, do you have any tips on how to avoid this issue?

Thanks a lot!

1. Do a forward propagation on a trained semantic segmentation model
1. Convert the model output (RGB-mask) into polygons (so that it is compatible with CVAT Annotation format)
1. Upload the annotations to CVAT and view the predicted-labels

### Context
Trying to generate the semi-automated labels for semantic segmentation without any "empty spaces". This is because the semantic segmentation models require every pixel to have one class assigned to them.

### Your Environment
- Operating System and version (e.g. Linux, Windows, MacOS): Windows

bsekachev:
### Motivation and context
Resolved #1849
Resolved #4868 

### Checklist

### License

```python
#
```


kkmehta03:
i tried checking out this branch but couldn't find the brush tool for some reason. 
i did 
```
docker-compose rm
docker-compose up
```
am i missing something here?

cvat-canvas/src/typescript/masksHandler.ts
```diff
@@ -0,0 +1,663 @@
...
+                imageDataToDataURL(imageBitmap, right - left + 1, bottom - top + 1,
+                    (dataURL: string) => new Promise((resolve) => {
+                        fabric.Image.fromURL(dataURL, (image: fabric.Image) => {
+                            try {
+                                image.selectable = false;
+                                image.evented = false;
+                                image.globalCompositeOperation = 'xor';
+                                image.opacity = 0.5;
+                                // image.left = this.latestMousePos.x - image.width / 2;
+                                // image.top = this.latestMousePos.y - image.height / 2;
```
klakhov, line 532: I see insertion works fine without those lines,needs to be removed?

cvat-core/src/session.ts
```diff
@@ -1939,819 +1946,811 @@ export class Task extends Session {
...
-
 buildDuplicatedAPI(Job.prototype);
 buildDuplicatedAPI(Task.prototype);
 
-Job.prototype.save.implementation = async function () {
-    if (this.id) {
-        const jobData = this._updateTrigger.getUpdated(this);
-        if (jobData.assignee) {
-            jobData.assignee = jobData.assignee.id;
+(async () => {
```
klakhov, line 1952: Why do we need such changes? I see `Property '...' does not exist on type 'Task'/'Job'` ts error anyway.

cvat-ui/src/components/annotation-page/canvas/brush-tools.tsx
```diff
@@ -0,0 +1,279 @@
...
+import useDraggable from './draggable-hoc';
+
+const DraggableArea = (
+    <div className='cvat-brush-tools-draggable-area'>
+        <Icon component={MoveIcon} />
+    </div>
+);
+
+const MIN_BRUSH_SIZE = 1;
+let polygonFinishingTipShown = false;
```
klakhov, line 35: `useState` for this tip?

klakhov (commented):
Great work!

I couldn't figure out how to use dragging feature, when I'm using this shortcut nothing happens 
ScreenRecorderProject11
Also I think tools on that panel lack corresponding tooltips

bsekachev:
@klakhov 

> I couldn't figure out how to use dragging feature, when I'm using this shortcut nothing happens

You can drag the panel using this element:
image

cvat-canvas/src/typescript/groupHandler.ts
```diff
@@ -185,23 +182,67 @@ export class GroupHandlerImpl implements GroupHandler {
...
     }
 
+    private appendToSelection(objectState: any): void {
+        const { clientID } = objectState;
+
+        const shape = this.canvas.select(`#cvat_canvas_shape_${clientID}`).first();
+        if (shape) {
+            if (objectState.shapeType === 'mask') {
+                const { points } = objectState;
+                const colorRGB = [139, 0, 139];
```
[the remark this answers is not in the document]
bsekachev: Yes, it is default group color defined in CSS files, but we can't apply CSS to change color of <image> element.

cvat-ui/src/components/annotation-page/canvas/brush-tools.tsx
bsekachev: ``useState`` is not necessary here, because it will be shown once the CVAT application is started. 
From my opinion, we should not show the tip again if for example a user reopened the job.  Don't you think so?

klakhov:
Okay, I see.
Also one more question: when I'm done working with mask it becomes almost invisible, should it really be this way? What if i want to create another mask along the border of the first one?
ezgif com-gif-maker

bsekachev:
@klakhov

> Also one more question

Hmm.. In general it depends on ``Shape opacity`` value.
And according to my idea it should increase automatically to good value, when a new mask created or appeared another way. 
What is your use-case?

klakhov:
Yep. Opacity slider solves this problem :smile:

klakhov:
image
For example if I want to have mask of the car and a lamp post. If opacity is low, I almost cant see car, and if I set it to high value I can see everything. I dont quite understand when opacity will change automatically except when I activate/deactivate mask.

bsekachev:
> Edit: I see after i finish working with mask, all masks have 30 opacity
 
Yes, it is expected. Buy my question is what have you done the first time when it was not changed automatically :) 
Just would like to know if there are unknown bugs

klakhov:
Cant reproduce it now. For some reason after switching branches, first mask that i've drawn  was not restoring it's opacity

klakhov:
Found one more problem, if I use `Plus` button to create new mask, brush dissapears.
masks-plus (1)

cvat/apps/dataset_manager/formats/pascal_voc.py
```diff
@@ -57,7 +57,8 @@ def _import(src_file, instance_data, load_data_callback=None):
                 shutil.move(f, anno_dir)
 
         dataset = Dataset.import_from(tmp_dir, 'voc', env=dm_env)
-        dataset.transform('masks_to_polygons')
+        if kwargs.get('conv_mask_to_poly', True):
```
zhiltsov-max, line 60: It doesn't look good to me to have a copy-pasted undocumented code fragment in almost all formats. Could you add some common class of function for this?

cvat/apps/dataset_manager/formats/transformations.py
```diff
@@ -36,6 +36,28 @@ def transform_item(self, item):
...
+        for rleCount in rle:
+            rleCount = math.trunc(rleCount)
+            while rleCount > 0:
+                x, y = offset % width, offset // width
+                mat[y + top][x + left] = value
+                rleCount -= 1
+                offset += 1
+            value = abs(value - 1)
+
+        rle = mask_utils.encode(np.asfortranarray(mat))
```
zhiltsov-max, line 57: There is an optimized version of mask to RLE conversion in Datumaro: https://github.com/cvat-ai/datumaro/blob/develop/datumaro/util/mask_tools.py#L155

cvat/apps/dataset_manager/tests/test_formats.py
```diff
@@ -962,6 +962,6 @@ def test_can_import_mots_annotations_with_splited_masks(self):
             task.update()
             task = self._create_task(task, images)
 
-            dm.task.import_task_annotations(task['id'], dataset_path, format_name)
+            dm.task.import_task_annotations(task['id'], dataset_path, format_name, True)
```
zhiltsov-max, line 965: Do you plan to test the masks somehow?

zhiltsov-max:
Tried the patch. Overall looks good. Some comments:
1. Probably, it a little bit unfair to call this new instrument "Draw a mask" from the data scientist perspective. Polygons do really work well as masks.
2. A context menu button or a separate tool to convert shapes to masks and back could be very useful (rasterize shape / vectorize mask)
3. Masks could be smoothed right after drawing
4. A mask mode to draw a filled mask could be useful (very close to the point 2 here)
5. Little mask gaps could be filled right after drawing
6. A mode to stick to the underlying edges could be useful (in polygons too. Maybe, it's more to polygons).

cvat/apps/dataset_manager/tests/test_formats.py
bsekachev: I, personally do not plan. I hope we will have some staff for that.

cvat/apps/dataset_manager/formats/pascal_voc.py
bsekachev: @zhiltsov-max 
Could you please provide an example how the code should look like? 
It is not clear what you expect.

bsekachev:
> Tried the patch. Overall looks good. Some comments:
> 
> 1. Probably, it a little bit unfair to call this new instrument "Draw a mask" from the data scientist perspective. Polygons do really work well as masks.
> 2. A context menu button or a separate tool to convert shapes to masks and back could be very useful (rasterize shape / vectorize mask)
> 3. Masks could be smoothed right after drawing
> 4. A mask mode to draw a filled mask could be useful (very close to the point 2 here)
> 5. Little mask gaps could be filled right after drawing
> 6. A mode to stick to the underlying edges could be useful (in polygons too. Maybe, it's more to polygons).

> A mode to stick to the underlying edges could be useful (in polygons too. Maybe, it's more to polygons).

It is implemented for polygons as automatic bordering mode. For masks we have ``Remove underlying pixels`` that allow to solve the use-case. 

I'm not sure I understand items 3-5, we can discuss them.

cvat/apps/dataset_manager/formats/transformations.py
bsekachev: image

Optimized?)

zhiltsov-max: Probably, it may depend on the hardware or library versions, because when I did this, it was like the opposite.

cvat/apps/dataset_manager/formats/pascal_voc.py
zhiltsov-max: It could look like a class:

```python
class MasksToPolygonsInImport:
    """
    Manages common logic for mask to polygons conversion in dataset import.
    This usecase is supposed for backward compatibility for the transition period.
    """

    @classmethod
    def declare_arg_names(cls):
        return ['conv_mask_to_poly']

    @classmethod
    def convert_dataset(cls, dataset, **kwargs) -> Dataset:
        if kwargs.get('conv_mask_to_poly', True):
            dataset.transform('masks_to_polygons')
        return dataset
```

cvat/apps/dataset_manager/formats/transformations.py
bsekachev: @zhiltsov-max 

I believe need to investigate this one. If for some users it works 10 time slower than for others.

I can provide my specs:

- Intel i7 12700K
- Python 3.8.10
- Ubuntu 20.04 (Run in VMWare Player 16)
- Pip freeze:

<details>
```
absl-py==1.3.0
asgiref==3.5.2
astroid==2.11.7
astunparse==1.6.3
attrs==21.4.0
av==8.0.2
azure-core==1.26.0
azure-storage-blob==12.13.0
boto3==1.17.61
botocore==1.20.112
cachetools==5.2.0
certifi==2022.9.24
cffi==1.15.1
charset-normalizer==2.0.12
click==7.1.2
contourpy==1.0.5
coreapi==2.3.3
coreschema==0.0.4
croniter==1.3.7
cryptography==38.0.1
cycler==0.11.0
datumaro==0.3
defusedxml==0.7.1
Deprecated==1.2.13
dill==0.3.5.1
diskcache==5.0.2
dj-pagination==2.5.0
dj-rest-auth==2.2.4
Django==3.2.15
django-allauth==0.50.0
django-appconf==1.0.4
django-auth-ldap==2.2.0
django-compressor==2.4
django-cors-headers==3.5.0
django-extensions==3.0.8
django-filter==2.4.0
django-rq==2.3.2
django-sendfile2==0.7.0
djangorestframework==3.12.4
dnspython==2.2.0
drf-spectacular==0.22.1
EasyProcess==0.3
entrypoint2==1.1
flatbuffers==22.9.24
fonttools==4.37.4
furl==2.1.0
gast==0.5.3
gitdb==4.0.9
GitPython==3.1.27
google-api-core==2.10.2
google-auth==2.12.0
google-auth-oauthlib==0.4.6
google-cloud-core==2.3.2
google-cloud-storage==1.42.0
google-crc32c==1.5.0
google-pasta==0.2.0
google-resumable-media==2.4.0
googleapis-common-protos==1.56.4
grpcio==1.49.1
h5py==3.6.0
idna==3.4
importlib-resources==5.10.0
inflection==0.5.1
isodate==0.6.1
isort==5.10.1
itypes==1.2.0
Jinja2==3.1.2
jmespath==0.10.0
jsonschema==4.16.0
keras==2.8.0
Keras-Preprocessing==1.1.2
kiwisolver==1.4.4
lazy-object-proxy==1.7.1
libclang==14.0.6
limits==2.7.0
lxml==4.9.1
Markdown==3.2.2
MarkupSafe==2.1.1
matplotlib==3.6.1
mccabe==0.7.0
mistune==2.0.4
msrest==0.7.1
natsort==8.0.0
networkx==2.8.7
nibabel==4.0.2
numpy==1.22.0
oauthlib==3.2.1
opencv-python==4.6.0.66
opencv-python-headless==4.5.5.62
opt-einsum==3.3.0
orderedmultidict==1.0.1
orjson==3.8.0
packaging==21.3
pandas==1.5.0
patool==1.12
pdf2image==1.14.0
Pillow==9.0.1
pkgutil-resolve-name==1.3.10
platformdirs==2.5.2
protobuf==4.21.7
pyasn1==0.4.8
pyasn1-modules==0.2.8
pycocotools==2.0.5
pycparser==2.21
Pygments==2.7.4
PyJWT==2.5.0
pylint==2.14.5
pylint-django==2.5.3
pylint-plugin-utils==0.7
pylogbeat==2.0.0
pyparsing==3.0.9
pyrsistent==0.18.1
python-dateutil==2.8.2
python-ldap==3.4.0
python-logstash-async==2.2.0
python3-openid==3.2.0
pytz==2020.1
pyunpack==0.2.1
PyYAML==6.0
rcssmin==1.0.6
redis==3.5.3
requests==2.26.0
requests-oauthlib==1.3.1
rjsmin==1.1.0
rope==0.17.0
rq==1.5.1
rq-scheduler==0.10.0
rsa==4.9
ruamel.yaml==0.17.21
ruamel.yaml.clib==0.2.6
rules==2.2
s3transfer==0.4.2
scipy==1.9.2
Shapely==1.7.1
six==1.16.0
smmap==5.0.0
snakeviz==2.1.0
sqlparse==0.4.2
tensorboard==2.8.0
tensorboard-data-server==0.6.1
tensorboard-plugin-wit==1.8.1
tensorboardX==2.5.1
tensorflow==2.8.1
tensorflow-estimator==2.8.0
tensorflow-io-gcs-filesystem==0.27.0
termcolor==2.0.1
tomli==2.0.1
tomlkit==0.11.5
tornado==6.2
types-cryptography==3.3.23
typing-extensions==4.4.0
uritemplate==4.1.1
urllib3==1.26.12
Werkzeug==1.0.1
wrapt==1.14.1
zipp==3.9.0
```
</details>

cvat/apps/dataset_manager/formats/pascal_voc.py
bsekachev: In this case we just incapsulate ``if`` logic.
Anyway need to modify each converter by calling ``convert_dataset``, right?

cvat/apps/dataset_manager/formats/transformations.py
zhiltsov-max: Could you please create an issue in the datumaro repo? I'm not sure we have resources for this right now, but it looks like an important task which should be done. For now, let's take your version.

zhiltsov-max:
> 3. Masks could be smoothed right after drawing

In some editors I've seen that the scribble the user draws (while a mouse button is pressed) is not being rasterized right away. Instead, while the button is pressed, the preview is shown, but when the button is released, the scribble drawn is being smoothed (probably, some kind of a Bezier curve is created to fit the scribble), and this new curve is what is actually resterized. The result is that the scribble is more smooth, i.e. has lesser sharp edges, but it may be slightly different from what the user has drawn.

> 4. A mask mode to draw a filled mask could be useful (very close to the point 2 here)

It is similar to what we understand as polygons. But if we draw a polygon and then convert it to a mask, we need to do extra clicks. With such mode where we can choose to draw a filled contour, the mask we get is already filled without extra actions. Otherwise, in the current mode, in many cases the user needs to manually fill the interior of the drawn mask by mouse, and this leads us to the next problem:

> 5. Little mask gaps could be filled right after drawing

If we are trying to fill a big contour by mouse scribbles, we often get numerous little gaps in the mask. This is especially observable if the brush size is small. And it may be hard or time-consuming to find and fix these little holes that should be filled in the resulting mask.

bsekachev:
@zhiltsov-max 

> It is similar to what we understand as polygons. But if we draw a polygon and then convert it to a mask, we need to do extra clicks. With such mode where we can choose to draw a filled contour, the mask we get is already filled without extra actions. Otherwise, in the current mode, in many cases the user needs to manually fill the interior of the drawn mask by mouse, and this leads us to the next problem:

As I understand what you are talking about is implemented already:
gfgfd


Speaking about others, I am not sure we should change user's input somehow, because it is manual tools. We should not decide for a users what they want.

zhiltsov-max:
Yes, this looks similar. Maybe I've overlooked some features.

cvat/apps/dataset_manager/formats/pascal_voc.py
zhiltsov-max: Yes, exactly.

nmanovic:
@bsekachev , several comments to improve UX:
- Need tooltips instead of notifications. Sometimes it is very difficult to understand what a function (button) does.
- In polygon tool some very important features are absent. For example, quick border drawing. If it is possible to unify drawing capabilities for polygons, it will be awesome.
- For brush it is necessary to hide the cursor. Otherwise I see some artifacts, which effect the usability.
- CTRL+1 and CTRL+2 doesn't change the label when I select the tool but before I click "draw mask" button.
- It is not very intuitive for new users to finish drawing a polygon by double click.
- Saving operation is extremely slow. I drew a complex mask, propagated it on 50 frames. It is saving results for several minutes. 
image

bsekachev:
> Need tooltips instead of notifications. Sometimes it is very difficult to understand what a function (button) does.
> For brush it is necessary to hide the cursor. Otherwise I see some artifacts, which effect the usability.

Done. 

> CTRL+1 and CTRL+2 doesn't change the label when I select the tool but before I click "draw mask" button.
> Saving operation is extremely slow. I drew a complex mask, propagated it on 50 frames. It is saving results for several minutes.

Will investigate this.. 

> In polygon tool some very important features are absent. For example, quick border drawing. If it is possible to unify drawing capabilities for polygons, it will be awesome.
> It is not very intuitive for new users to finish drawing a polygon by double click.

I will think if we can easily use drawing polygons implemented with SVG.js here. But it does not solve the second issue.

bsekachev:
@nmanovic 

> Saving operation is extremely slow. I drew a complex mask, propagated it on 50 frames. It is saving results for several minutes.

Should be much faster now.

kirill-sizov:
@bsekachev Do you see any reason why a [failed test](https://github.com/opencv/cvat/actions/runs/3310705375/jobs/5465410435) runs infinitely or it is deserve separate issue ticket?

bsekachev:
> @bsekachev Do you see any reason why a [failed test](https://github.com/opencv/cvat/actions/runs/3310705375/jobs/5465410435) runs infinitely or it deserves separate issue ticket?

@kirill-sizov I am not sure why they are infinitely, but now they are ok

nmanovic (approved):

PR #4633, opened 2022-05-12
Pin Python dependencies with pip-tools

### LINKED ISSUE #4618: [BUG] Building local images failing due to error in pip install of requirements "ImportError: cannot import name 'Feature' from 'setuptools'"
### Expected Behaviour
Running "docker-compose -f docker-compose.yml -f docker-compose.dev.yml build" succeeds

### Current Behaviour
Running the above command fails when running
"/bin/sh -c DATUMARO_HEADLESS=1 python3 -m pip install --no-cache-dir -r /tmp/requirements/${DJANGO_CONFIGURATION}.txt"

Error is as follows:

error

Run the above docker-compose command.  This error has happened on both a slightly outdated cvat repo that was pulled a few weeks ago, and also happened when I pulled down the latest repo.  It has happened on both my personal Windows laptop and a separate Ubuntu virtual machine, so I do not believe it has something to do with my local environment.

ngaloppo:
### Motivation and context

pip dependency resolver was taking extremely long due to complex version
resolution prior to this change (>6 hours on CI). This commit introduces
version pinning with pip-tools. A wrapper script to update the pinned
versions when dependencies are added, or when upgrades are needed, is
provided.

Fixes #4618 

### How has this been tested?

CI-nightly build phase seems to pass the error described in #4618. However, the full build still doesn't complete (>6h passed).

### Checklist

### License

```python
#
```


zhiltsov-max:
@ngaloppo, thank you for the PR! It seems that the issue linked was with a specific package, and it is resolved now. I'm not sure that we want this specific solution for the develop branch, but it can be a good way to make tagged releases and custom deployments more reproducible and stable against non-fixed dependency updates. @azhavoro, what do you think? Can we run such a script before releases?

zhiltsov-max:
Please also check and comment on https://github.com/opencv/cvat/issues/4270.

SpecLad:
I quite like `pip-compile`, and I would support integrating it. That said, I think a few tweaks need to be made to the current approach:

* We should not immediately drop all version specifiers for the `requirements.in` files. Rather, the initial `.in` files should just be renamed from the existing `.txt` files. This will help avoid unexpected changes when people regenerate the `.txt` files later. Later on, we can unpin/relax versions on a case-by-case basis.

* We should use `pip-compile-multi` to avoid duplication of requirements between the various `.txt` files. `pip-compile-multi` also automatically finds the files to process, so it'll let us drop the automation script.

nmanovic:
https://github.com/opencv/cvat/issues/5310

PR #4775, opened 2022-07-22
Enable running multiple docker-compose instances at the same time

JOT85:
### Motivation and context

We have a shared development machine which we host some internal services on. We would like to run multiple instances of CVAT through docker-compose however this currently fails for a few reasons. Firstly, instances have identical container names. Secondly, the more significant issue, is that the traefik instance defined in the docker compose was handling routing for everything, including containers from other docker-compose instances. This caused nothing to be accessible, and would also cause problems for any other applications defining traefik rules.

### Changes

To resolve these problems, I've changed the container names and added a `cvat.instance` label to each container. These names and instances are unique depending on the `CVAT_HOST` and `CVAT_EXT_PORT` pair. `CVAT_EXT_PORT` defines the port that the main cvat interface is exposed on. The name is suffixed with this pair, and the `cvat.instance` label is equal to this pair. The traefik instance then has a constrains which checks that label before trying to route for it.

I've added a simple explanation of how to set up multiple instances to the docs.

### Other changes

I've also commented out the port binding for opa by default. I don't think this needs to be exposed by default? That port would also have to be unique between instances if it wasn't commented by default.

### How has this been tested?

I've tested this on my personal system and the system we deploy development tools to.

### Checklist

### License

```python
#
```


Marishka17:
Hi, @JOT85, thanks for the contribution! Are you still going to contribute to us? If so, could you please resolve conflicts in this PR?

nmanovic:
@JOT85 , please use helm-chart to run multiple instances of CVAT on the same machine: https://opencv.github.io/cvat/docs/administration/advanced/k8s_deployment_with_helm/

We are not going to complicate deployment using docker-compose. Thank you for the contribution!

PR #4544, opened 2022-04-04
Fix dataset import for Datumaro format

kirill-sizov:
### Motivation and context

Fix #4410 

- Fixed project import and export for Datumaro, KITTI and VGGFace2 formats
- Added project import/export tests

### Checklist

- [x] I submit my changes into the `develop` branch
- [x] I have added a description of my changes into [CHANGELOG](https://github.com/opencv/cvat/blob/develop/CHANGELOG.md) file
- [x] I have updated the [documentation](
  https://github.com/opencv/cvat/blob/develop/README.md#documentation) accordingly
- [x] I have added tests to cover my changes
- [x] I have linked related issues ([read github docs](
  https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword))
- [x] I have increased versions of npm packages if it is necessary ([cvat-canvas](https://github.com/opencv/cvat/tree/develop/cvat-canvas#versioning),
[cvat-core](https://github.com/opencv/cvat/tree/develop/cvat-core#versioning), [cvat-data](https://github.com/opencv/cvat/tree/develop/cvat-data#versioning) and [cvat-ui](https://github.com/opencv/cvat/tree/develop/cvat-ui#versioning))

### License

- [x] I submit _my code changes_ under the same [MIT License](
  https://github.com/opencv/cvat/blob/develop/LICENSE) that covers the project.
  Feel free to contact the maintainers if that's a concern.
- [x] I have updated the license header for each file (see an example below)

```python
# Copyright (C) 2022 Intel Corporation
#
# SPDX-License-Identifier: MIT
```

PR #4861, opened 2022-08-26
Fix an issue with Imagenet dataset import

### LINKED ISSUE #4850: Import dataset of Imagenet format fail

### Expected Behaviour
Import Dataset successful.

### Current Behaviour

cannot import and feedback:
Could not import dataset to the [project 298](https://app.cvat.ai/projects/298)
Error: Request failed with status code 500. "ValueError: not enough values to unpack (expected 2, got 1)\n".

### Possible Solution
check code and detail logs from server side.

1. export dataset for image classification by ImageNet format. prepare dataset like: https://drive.google.com/file/d/1dj7CSoWi2MIIunYFlhHNoqeQnqjXFoZd/view?usp=sharing
1. import dataset into another project.

### Your Environment
app.cvat.ai
and localhost CVAT v2.1.0

Issue by @deryann

yasakova-anastasia:
### Motivation and context

- Fixed #4850 (The shared path to the files was found incorrectly)

### Checklist

### License


cvat/apps/dataset_manager/bindings.py
```diff
@@ -1643,8 +1643,6 @@ def load_dataset_data(project_annotation, dataset: Dataset, project_data):
                 dataset_files['media'] += \
                     list(map(lambda ri: ri.path, dataset_item.related_images))
 
-        shortes_path = min(dataset_files['media'], key=lambda x: len(Path(x).parts), default=None)
-        if shortes_path is not None:
-            dataset_files['data_root'] = str(Path(shortes_path).parent.absolute()) + osp.sep
+        dataset_files['data_root'] = osp.commonpath(dataset_files['media']) + osp.sep
```
zhiltsov-max, line 1646: I see 2 errors here:
- If there is only 1 image, its path will be used in full, including the filename
- if there are no images, the function will throw an exception

Could you also add a test?

yasakova-anastasia: Fixed these issues and added a test

zhiltsov-max (approved):
I was able to import the dataset. 

However, I've found another problem: if I import an ImageNet dataset, change a label on an image, export with images as ImageNet, and then import it, the dataset misses 1 image (the changed one). The image exists in the exported archive, but the file layout is like this:

```
/cat/cat_1.jpg
,,,
/dog/cat/cat_0.jpg
/dog/dog_0.jpg
...
```

tests/python/rest_api/test_projects.py
```diff
@@ -458,6 +458,23 @@ def test_admin_can_get_project_backup_and_create_project_by_backup(self, admin_u
...
+
+        tmp_file = io.BytesIO(response.data)
+        tmp_file.name = "dataset.zip"
+
+        import_data = {
+            "dataset_file": tmp_file,
+        }
+
+        self._test_import_project(username, project_id, format_name, import_data)
+
```
zhiltsov-max, line 477: Probably, this test should be removed when #4544 is merged.

cvat/apps/dataset_manager/bindings.py
```diff
@@ -1885,17 +1885,21 @@ def load_dataset_data(project_annotation, dataset: dm.Dataset, project_data):
             'data_root': dataset.data_path + osp.sep,
         }
 
+        data_root = None
         for dataset_item in subset_dataset:
             if dataset_item.image and dataset_item.image.has_data:
                 dataset_files['media'].append(dataset_item.image.path)
+                if data_root is None:
+                    data_root = osp.splitext(dataset_item.image.path)[0].replace(dataset_item.id, '')
```
SpecLad, line 1893: This could corrupt the path if its leading part happens to also contain the ID. You must make sure you're only removing the ID from the end of the path.

cvat/apps/dataset_manager/bindings.py
```diff
@@ -1885,17 +1885,21 @@ def load_dataset_data(project_annotation, dataset: dm.Dataset, project_data):
...
+                    data_root = osp.splitext(dataset_item.point_cloud)[0].replace(dataset_item.id, '')
             if isinstance(dataset_item.related_images, list):
                 dataset_files['media'] += \
                     list(map(lambda ri: ri.path, dataset_item.related_images))
 
-        shortes_path = min(dataset_files['media'], key=lambda x: len(Path(x).parts), default=None)
-        if shortes_path is not None:
-            dataset_files['data_root'] = str(Path(shortes_path).parent.absolute()) + osp.sep
+        if len(dataset_files['media']):
+            dataset_files['data_root'] = data_root
```
zhiltsov-max, line 1903: This solution doesn't seem correct. It just takes the first path, and it may not be the common prefix. Please check if [this function](https://docs.python.org/3/library/os.path.html#os.path.commonpath) can be useful here. Also, with this solution I'm still able to reproduce the error in a dataset with different directory levels.

cvat/apps/dataset_manager/bindings.py
zhiltsov-max: I'd also suggest to use just `osp.split` instead.

cvat/apps/dataset_manager/bindings.py
yasakova-anastasia: Fixed it. But nested folders are not possible for ImageNet at the moment.

tests/python/rest_api/test_projects.py
yasakova-anastasia: I think this test is needed. It refers to a specific issue. I agree that the updated test in the #4544 covers it. But the test as here is the most understandable and transparent (as it seems to me)

cvat/apps/dataset_manager/bindings.py
```diff
@@ -1885,17 +1885,24 @@ def load_dataset_data(project_annotation, dataset: dm.Dataset, project_data):
             'data_root': dataset.data_path + osp.sep,
         }
 
+        root_paths = set()
         for dataset_item in subset_dataset:
             if dataset_item.image and dataset_item.image.has_data:
                 dataset_files['media'].append(dataset_item.image.path)
+                data_root = dataset_item.image.path.rsplit(dataset_item.id, 2)
```
SpecLad, line 1892: Shouldn't `maxsplit` be 1 rather than 2?

yasakova-anastasia: You are right, thanks

zhiltsov-max:
I see the previous problem has gone, but I was lucky enough to find another one. If we export an Imagenet dataset in CVAT format and then import the dataset, there will be an error

```
File "cvat/cvat/apps/dataset_manager/bindings.py", line 1685, in import_dm_annotations
    match_dm_item(item, instance_data, root_hint=root_hint))
  File "cvat/cvat/apps/dataset_manager/bindings.py", line 1636, in match_dm_item
    raise CvatImportError("Could not match item id: "
cvat.apps.dataset_manager.bindings.CvatImportError: Could not match item id: 'cat/cat_0.jpg' with any task frame
```

I don't think this new problem should be fixed here, let's do it in another PR to unblock this.

cvat/apps/dataset_manager/bindings.py
```diff
@@ -1885,17 +1885,24 @@ def load_dataset_data(project_annotation, dataset: dm.Dataset, project_data):
...
+
             if isinstance(dataset_item.related_images, list):
                 dataset_files['media'] += \
                     list(map(lambda ri: ri.path, dataset_item.related_images))
 
-        shortes_path = min(dataset_files['media'], key=lambda x: len(Path(x).parts), default=None)
-        if shortes_path is not None:
-            dataset_files['data_root'] = str(Path(shortes_path).parent.absolute()) + osp.sep
+        if len(root_paths):
+            dataset_files['data_root'] = osp.commonpath(root_paths)
```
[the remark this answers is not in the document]
zhiltsov-max: Looking at how add_task is implemented, I think TensorFlows datasets, probably, cannot be imported, because images are packed into tfrecords.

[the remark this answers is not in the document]
yasakova-anastasia: > I think you might need to append `os.sep` here. In `add_task`, `data_root` is stripped from the left side of each path, and if it doesn't end with a slash, there will be a slash left at the start of the path, which will turn it into an absolute path.

Updated

SpecLad (approved):

PR #4980, opened 2022-09-20
Extend SDK layer 2 docs

zhiltsov-max:
### Motivation and context

- Added more explanations on how to use Core API objects

### Checklist

### License


site/content/en/docs/api_sdk/sdk/highlevel-api.md
```diff
@@ -61,3 +62,107 @@ with make_client(host="http://localhost") as client:
     task.update({'name': 'mytask'})
     task.remove()
 ```
+
+## Client
+
+The `cvat_sdk.core.client.Client` class manages session, provides connection management and
```
SpecLad, line 68:
```suggestion
The `cvat_sdk.core.client.Client` class manages sessions, provides connection management and
```

site/content/en/docs/api_sdk/sdk/highlevel-api.md
```diff
@@ -61,3 +62,107 @@ with make_client(host="http://localhost") as client:
...
+resource location operations. It is the staring point for using CVAT SDK.
```
SpecLad, line 69:
```suggestion
resource location operations. It is the starting point for using CVAT SDK.
```

site/content/en/docs/api_sdk/sdk/highlevel-api.md
```diff
@@ -61,3 +62,107 @@ with make_client(host="http://localhost") as client:
...
+
+An instance of `Client` can be created directly by calling the class constructor,
+or with the utility function `cvat_sdk.core.client.make_client()`, which can handle
+some configuration for you. The extended configuration for `Client` is performed with
+the `cvat_sdk.core.client.Config` class instances. A `Config` object can be passed to
+the `Client` constructor and then it can be accessed with the `Client.config` field.
```
SpecLad, line 75:
```suggestion
the `Client` constructor and then it can be accessed via the `Client.config` field.
```

site/content/en/docs/api_sdk/sdk/highlevel-api.md
```diff
@@ -61,3 +62,107 @@ with make_client(host="http://localhost") as client:
...
+`Client` objects implement [context manager protocol](https://docs.python.org/3/reference/datamodel.html#context-managers).
```
SpecLad, line 76:
```suggestion
`Client` objects implement the [context manager protocol](https://docs.python.org/3/reference/datamodel.html#context-managers).
```

This could also use a description of _how_ these objects implement the protocol, i.e., what happens on exit.

site/content/en/docs/api_sdk/sdk/highlevel-api.md
```diff
@@ -61,3 +62,107 @@ with make_client(host="http://localhost") as client:
...
+
+You can create `Client` this way:
```
SpecLad, line 78:
```suggestion
You can create a `Client` instance this way:
```

site/content/en/docs/api_sdk/sdk/highlevel-api.md
```diff
@@ -61,3 +62,107 @@ with make_client(host="http://localhost") as client:
...
+
+```python
+from cvat_sdk import make_client
+
+with make_client('localhost', port='8080', credentials=(user, password)) as client:
+    ...
+```
+
+Or, if you need to tweak `Client`, you can do this:
```
SpecLad, line 87:
```suggestion
Or, if you need to tweak the created `Client`, you can do this:
```

site/content/en/docs/api_sdk/sdk/highlevel-api.md
```diff
@@ -61,3 +62,107 @@ with make_client(host="http://localhost") as client:
...
+
+with Client('localhost:8080', config=config) as client:
+    ...
+```
+
+With `Client`, you can `login()` and `logout()`, and you can get Repository objects.
+
+## Entities and Repositories
+
+Entitites represent separate objects on the server. They provide read access to object fields
```
SpecLad, line 103:
```suggestion
Entities represent distinct objects on the server. They provide read access to object fields
```

site/content/en/docs/api_sdk/sdk/highlevel-api.md
```diff
@@ -61,3 +62,107 @@ with make_client(host="http://localhost") as client:
...
+
+You can create `Client` this way:
+
+```python
+from cvat_sdk import make_client
+
+with make_client('localhost', port='8080', credentials=(user, password)) as client:
+    ...
+```
+
```
SpecLad, line 86: It would be useful here to provide an equivalent example using `Client`, so that the reader can understand what `make_client` does for you.

site/content/en/docs/api_sdk/sdk/highlevel-api.md
```diff
@@ -61,3 +62,107 @@ with make_client(host="http://localhost") as client:
...
+from cvat_sdk import Config, Client
+
+config = Config()
+# set up some config fields ...
+
+with Client('localhost:8080', config=config) as client:
+    ...
+```
+
+With `Client`, you can `login()` and `logout()`, and you can get Repository objects.
```
SpecLad, line 99:
```suggestion
With a `Client` instance, you can call `login()` and `logout()`, and you can get Repository objects.
```

site/content/en/docs/api_sdk/sdk/highlevel-api.md
```diff
@@ -61,3 +62,107 @@ with make_client(host="http://localhost") as client:
...
+
+After calling these functions, we obtain local objects, representing their server counterparts.
+
+Object fields can be updated with the `update()` method:
+
+```python
+job.update({'stage': 'validation'})
+```
+
+The sever object will be updated and the local object will reflect the latest object state
```
SpecLad, line 139:
```suggestion
The server object will be updated and the local object will reflect the latest object state
```

site/content/en/docs/api_sdk/sdk/highlevel-api.md
```diff
@@ -61,3 +62,107 @@ with make_client(host="http://localhost") as client:
...
+after calling this operaiton.
+
+Note, that local objects may out of sync with their server counterparts for different reasons.
```
SpecLad, line 142:
```suggestion
Note that local objects may fall out of sync with their server counterparts for different reasons.
```

site/content/en/docs/api_sdk/sdk/highlevel-api.md
```diff
@@ -61,3 +62,107 @@ with make_client(host="http://localhost") as client:
...
+
+# update the server object with the first reference
+job_ref1.update(...)
+# job_ref2 is outdated now
+
+job_ref2.fetch()
+# job_ref2 is synced
+```
+
+The local changes in the object are discarded when `fetch()` is called.
```
SpecLad, line 158: This seems a bit contradictory, since:

1. You haven't explained how to make local changes, and
2. Earlier you said that objects provide read-only fields, which weakly implies that local changes aren't possible.

site/content/en/docs/api_sdk/sdk/highlevel-api.md
```diff
@@ -61,3 +62,107 @@ with make_client(host="http://localhost") as client:
...
+
+Finally, if you need to remove object from the server, you can use the `remove()` method.
```
SpecLad, line 160:
```suggestion
Finally, if you need to remove the object from the server, you can use the `remove()` method.
```

Also, whose method?

This should probably be supplemented with an example.

site/content/en/docs/api_sdk/sdk/highlevel-api.md
```diff
@@ -10,7 +10,8 @@ description: ''
 This layer provides high-level APIs, allowing easier access to server operations.
 API includes _Repositories_ and _Entities_. Repositories provide management
 operations for Entitites. Entitites represent separate objects on the server
-(e.g. tasks, jobs etc).
+(e.g. tasks, jobs etc). The key difference from the low-level API is that
```
SpecLad, line 13:
```suggestion
(e.g. tasks, jobs etc). A key difference from the low-level API is that
```

I don't think you could call it _the_ key difference when there are other important differences as well (like representing operations as methods rather than standalone functions).

cvat-sdk/cvat_sdk/core/proxies/model_proxy.py
```diff
@@ -182,6 +182,8 @@ def _export_update_fields(
     def fetch(self: Entity) -> Self:
         """
         Updates current object from the server
+
+        The local changes in the object are discarded when `fetch()` is called.
```
SpecLad, line 186:
```suggestion
        Any local changes in the object are discarded.
```

cvat-sdk/cvat_sdk/core/client.py
```diff
@@ -39,21 +43,30 @@ class Config:
...
         self.config = config or Config()
+        """Configuration for this object"""
+
         self.api_map = CVAT_API_V2(url)
+        """Handles server API URL interaction logic"""
+
         self.api_client = ApiClient(
             Configuration(host=self.api_map.host, verify_ssl=self.config.verify_ssl)
         )
+        """Provides low-level access to CVAT server"""
```
SpecLad, line 66:
```suggestion
        """Provides low-level access to the CVAT server"""
```

cvat-sdk/cvat_sdk/core/client.py
```diff
@@ -39,21 +43,30 @@ class Config:
 
 class Client:
     """
-    Manages session and configuration.
+    Manages session and configuration, provides resource location.
```
SpecLad, line 46: What is "resource location"? I don't think it's explained in the docs.

SpecLad (changes requested):

site/content/en/docs/api_sdk/sdk/highlevel-api.md
zhiltsov-max: Changed the wording

site/content/en/docs/api_sdk/sdk/highlevel-api.md
zhiltsov-max: Changed the wording

site/content/en/docs/api_sdk/sdk/highlevel-api.md
zhiltsov-max: Added

site/content/en/docs/api_sdk/sdk/highlevel-api.md
zhiltsov-max: Changed the wording

site/content/en/docs/api_sdk/sdk/highlevel-api.md
zhiltsov-max: Changed the wording

site/content/en/docs/api_sdk/sdk/highlevel-api.md
zhiltsov-max: I've added some explanations. I don't think it is a good idea to fix the actions done by this function in the documentation, because the code tends to change quickly.

site/content/en/docs/api_sdk/sdk/highlevel-api.md
zhiltsov-max: Changed the wording

site/content/en/docs/api_sdk/sdk/highlevel-api.md
zhiltsov-max: Changed the wording

site/content/en/docs/api_sdk/sdk/highlevel-api.md
zhiltsov-max: Added an example

site/content/en/docs/api_sdk/sdk/highlevel-api.md
zhiltsov-max: Class methods are not a big improvement over free functions, especially when we speak about this first layer, where apis and models are already grouped into classes. It's not that important in which order we put words, dots and dashes in the function call (`r = p_cls_foo(p, q)` vs `r = p.foo(q)` vs even `p_cls_foo(p, q, r)`). The one real benefit of this is the ability to inherit and overload, but its not something that I'd like to talk about in the docs. I've changed the wording.

cvat-sdk/cvat_sdk/core/proxies/model_proxy.py
zhiltsov-max: I like this new variant, applied

cvat-sdk/cvat_sdk/core/client.py
zhiltsov-max: Probably, it's something I'd like to see in future. Changed the wording

site/content/en/docs/api_sdk/sdk/highlevel-api.md
```diff
@@ -44,20 +48,177 @@ with make_client(host="http://localhost") as client:
...
+    client.login(('user', 'password'))
+    ...
+```
+
+You can specify server address both with and without the scheme. If the scheme is omitted,
+it will be deduced automatically.
+
+> The checks are performed in the following
+order: `https` (with the default port 8080), `http` (with the default port 80).
+In some cases it may lead to incorrect results - eg. you have 2 servers running on the
```
SpecLad, line 137:
```suggestion
In some cases it may lead to incorrect results - e.g. you have 2 servers running on the
```

site/content/en/docs/api_sdk/sdk/highlevel-api.md
```diff
@@ -44,20 +48,177 @@ with make_client(host="http://localhost") as client:
...
+can only see and modify their tasks and jobs, while an admin user can see all the tasks etc.
+
+## Entities and Repositories
+
+_Entities_ represent objects on the server. They provide read access to object fields
+and implement additional relevant operations, including both the general Read-Update-Delete and
+object-specific ones. The set of available general operations depends on the object type.
+
+_Repositories_ provide management operations for corresponding _Entities_. Typically, you don't
+need to create _Repository_ objects manually. To obtain a _Repository_ object, use the
```
SpecLad, line 156: Do you _ever_ need to create Repository objects manually? Seems like the "typically" could be removed.

site/content/en/docs/api_sdk/sdk/highlevel-api.md
```diff
@@ -8,25 +8,29 @@ description: ''
 ## Overview
 
 This layer provides high-level APIs, allowing easier access to server operations.
-API includes _Repositories_ and _Entities_. Repositories provide management
-operations for Entitites. Entitites represent separate objects on the server
-(e.g. tasks, jobs etc).
+API includes _Repositories_ and _Entities_. _Repositories_ provide management
+operations for _Entities_. _Entities_ represent objects on the server
```
SpecLad, line 12: I would suggest only using italics for the defining occurrences of the terms (once for each term). Otherwise it's going to be annoying to consistently italicize these terms everywhere - in fact, I already see them appear without italics at the bottom of the document. I don't think italics everywhere would help with readability either.

```suggestion
API includes Repositories and Entities. _Repositories_ provide management
operations for Entities. _Entities_ represent objects on the server
```

site/content/en/docs/api_sdk/sdk/highlevel-api.md
SpecLad: You didn't address the main issue here - it's still unclear how local changes might appear.

zhiltsov-max: I don't know. maybe the user changed some values occasionally, because full immutability is impossible in python. Maybe the user created the entity object manually from the response model data. Previously, there was a holistic idea of changing fields and sending updates separately, but currently it works bad, because response and request model fields doesn't map directly. From my point, "any" is good enough to cover these cases.

site/content/en/docs/api_sdk/sdk/highlevel-api.md
zhiltsov-max: Agree. I decided to leave them in 2 places, where the terms can be declared for a reader.

site/content/en/docs/api_sdk/sdk/highlevel-api.md
zhiltsov-max: Currently the answer is no, it's never needed. But it wasn't done on purpose.

site/content/en/docs/api_sdk/sdk/highlevel-api.md
SpecLad: We need to be consistent when describing how the user can interact with the API. If we mention local changes here, then that implies that it's possible to make local changes, and we need to describe that. Conversely, if it's _not_ possible to make local changes, then they should not be mentioned.

Note that by "possible" I mean "possible via supported, documented functionality". It goes without saying that in Python, you can pretty much modify anything if you're willing to go far enough. But the purpose of the documentation is to describe how users _should_ interact with the API, and so it only needs to cover situations that may arise from such legitimate use.

SpecLad (approved):

zhiltsov-max:
@nmanovic , the PR is ready for merging.

Issue #2987, opened 2021-03-21
Loading 16-bit tiff images


marcelschilling:
Hello, I tried to load tiff images using the online test version of cvat. However, it always failed when trying to upload a tif image. Uploading jpg or png was possible. Is tiff data format not supported? Thanks in advance. Best, Marcel

azhavoro:
Hi, I just successfully created a task with a tiff image, could you please share any image you tried to upload? Perhaps your images are using some type of compression or something else that is not supported.

marcelschilling:
Hi, thank you very much for your response. It tried it with another image (8 bit tiff rgb image) and it worked for me, too. However, using a 16 bit tiff image won't work. Maybe there is a issue/no compatibility to use 16 bit tiff images (maybe pil related)?

azhavoro:
@marcelschilling Hi,

unfortunately I cannot reproduce the issue with 16 bit depth image, could you please share any image that is not working for you? Thanks

I've tried with:
```
Format: TIFF (Tagged Image File Format)
  Mime type: image/tiff
  Class: DirectClass
  Geometry: 3072x2048+0+0
  Resolution: 72x72
  Print size: 42.6667x28.4444
  Units: PixelsPerInch
  Colorspace: sRGB
  Type: TrueColor
  Endianness: MSB
  Depth: 16-bit
  Channel depth:
    Red: 16-bit
    Green: 16-bit
    Blue: 16-bit
```

azhavoro:
@marcelschilling I'll close the issue due to no response, please don't hesitate to reopen if the issue is still actual for you.

gcamfer:
I have the same issue.
I think the problem is when there is only one channel, EXIF info:
```
Filename - THM_00078_00070.tif
ImageWidth - 640
ImageLength - 512
BitsPerSample - 
Compression - 1 (None)
Make - senseFly
Model - senseFlyThermal
StripOffset - 3638
SamplesPerPixel - 1
StripByteCount - 655360
ExifOffset - 3478
DateTimeOriginal - 2014:10:23 10:33:49

GPS information: - 
GPSLatitudeRef - N
GPSLatitude - 46.535397  0  0 (46.535397)
GPSLongitudeRef - E
GPSLongitude - 6.55555  0  0 (6.555550)
GPSAltitude - 528.10 m
```

I download the images from sensefly web:
https://senseflycom.s3.amazonaws.com/datasets/thermomap-solarpanel/Geotagged%20images.rar

[...]

PR #5005, opened 2022-09-27
Loading 16-bit tiff images

MyNameIsFu:
Pull Request regarding Issue #2987 

PIL.Image conversion from I;16 to L or RGB are unsuccessful as for now.
See the corresponding Issue in the Pillow GitHub (Opened 2018, so no changes to be expected)
https://github.com/python-pillow/Pillow/issues/3011

The proposed changes at least fix this issue for the mode 'I;16' and delivers a possible solution for other modes (eg. I;16B/L/N).

This results in a correct calculation of the preview thumbnail and the actual image, the annotation will be performed on.

We have used this solution on our own dataset and created annotations accordingly.

### Checklist
- [x] I have increased versions of npm packages if it is necessary ([cvat-canvas](https://github.com/cvat-ai/cvat/tree/develop/cvat-canvas#versioning),

### License


nmanovic:
@MyNameIsFu , thanks for the contribution. Could you please add a line into CHANGELOG.md?

MyNameIsFu:
I added a comment. Is there anything I have to change regarding the failed pipeline?

kirill-sizov:
Hi, @MyNameIsFu, I checked your PR and met with problem: I can't create an issue with an image. This error is also the cause of the failure of the entire our testing pipeline.


I attach the image, so you can reproduce this problem.
image

MyNameIsFu:
@kirill-sizov The attached image seems to be cropped. Can you upload it again?

kirill-sizov:
I've uploaded the right image, this is a cropped small image that I have a problem with.

MyNameIsFu:
Ah sorry, I'll debug it and push the changes :+1:

MyNameIsFu:
@kirill-sizov 
I created a new Project for the test image and 2 separate tasks (from .png and from .zip). Both are created successfully and i can annotate and save correctly. The image you provided is in png format and open in from PIL.Image as mode 'rgba'. My code changes explicitly take effect on mode == 'I;16' only. Are you sure th
Screenshot from 2022-09-29 15-11-20
is is the correct image that throws the errors?

kirill-sizov:
@MyNameIsFu, I got the following error:
Screenshot from 2022-09-29 19-07-15

And here is logs from `cvat_worker_default` container [log.zip](https://github.com/opencv/cvat/files/9676443/log.zip)

MyNameIsFu:
Okay I looked into the Error, the CI Pipeline and the Tests and None of my changes should have any effect on the above error. I'm not even able to reproduce the Issue with CVAT. Nevertheless I tried to reproduce the Error without CVAT and instead with pure PIL and it appears to be a problem with implicit image conversion from RGBA to RGB (PNG to JPEG format).
Just as a reminder, I only touched the code regarding single-channel Images and can't think of a reason why it should have an impact on the 3 and 4 Channel images.

I'll take another look at the error and why it might differently behave on the Pipeline and my System. I also pushed Pillow version 9.2.0, just out of curiousity.

nmanovic:
@MyNameIsFu , thanks for the contribution once again. If tests are OK, I will merge the PR. In general I hope to reimplement the solution in the future. Actually we need to have 16-bit image in UI to change contract and brightness dynamically. Even it sounds simple, the actual implementation is going to be non-trivial.

cvat/apps/engine/media_extractors.py
```diff
@@ -5,6 +5,7 @@
 import os
 import tempfile
 import shutil
+from typing import List
 import zipfile
```
nmanovic, line 9:
```suggestion
import zipfile
```

nmanovic (approved):

PR #5058, opened 2022-10-07
cvat-sdk: Fix creating tasks with non-local files

### LINKED ISSUE #4962: CLI create task error with sorting_method
I am trying to upload tasks via command line interface using the following command:

```
(cvat) user@machine:~/cvat python cvat-cli/src/cvat_cli/__main__.py --auth some_user:some_password \
--server-host some_server create test_cli \
share //some_folder/some_file.mp4 --chunk_size 8 \
--completion_verification_period 5 --frame_step 360 \
--project_id 1 --lfs --overlap 0 --segment_size 100 \
--use_zip_chunks --use_cache
```

I have tried this with the script in the cvat repo and installing cvat-cli using pip:

```
(cvat) user@machine:~/cvat cvat-cli --auth some_user:some_password \
--server-host some_server create test_cli \
share //some_folder/some_file.mp4 --chunk_size 8 \
--completion_verification_period 5 --frame_step 360 \
--project_id 1 --lfs --overlap 0 --segment_size 100 \
--use_zip_chunks --use_cache
```
Get the same error using both cases.

### Expected Behaviour

 I would expect the task to be created successfully.

### Current Behaviour

I get the following output using both methods:

```
0.00B [00:00, ?B/s][2022-09-16 11:08:49] INFO: Created task ID: 29 NAME: test_cli
[2022-09-16 11:08:49] CRITICAL: Status Code: 400
Reason: Bad Request
HTTP response headers: HTTPHeaderDict({'Allow': 'GET, POST, HEAD, OPTIONS', 'Content-Length': '69', 'Content-Type': 'application/vnd.cvat+json', 'Date': 'Fri, 16 Sep 2022 09:08:49 GMT', 'Referrer-Policy': 'same-origin', 'Server': 'Apache', 'Vary': 'Accept,Origin', 'X-Content-Type-Options': 'nosniff', 'X-Frame-Options': 'DENY'})
HTTP response body: {"sorting_method":["\"\"lexicographical\"\" is not a valid choice."]}
```

### Possible Solution

I have used tcpdump to capture traffic and used wireshark to see what's going on with the HTTP stream. Since the problem is in that it is not recognising the sorting method when using the CLI i have tried and captured traffic using the WUI and the CLI. The difference is merely that when using the CLI, the are extra quotation marks added to the string value of sorting_method:
CVAT_HTTP_WUI
CVAT_HTTP_CLI

I am not very familiar with HTTP protocols and how cvat server handles these requests, but I am assuming that removing the quotation marks from the sorting method would suffice.

Thanks in advance and please let me know if there's any extra information you need (this is my firts issue)

SpecLad:
Forcing the `Content-Type` for the `upload_data` API call to `multipart/form-data` does not work, because the current logic for
converting Python values to parts (`ApiClient._convert_body_to_post_params`) does not encode them in a way that Django REST Framework can understand (it JSON-encodes each part).

Fortunately, we don't actually need to do that, since when we create a task with non-local files, we don't need to upload any files, and so we can just post the original JSON, so do just that.

I couldn't add a test for the remote image case, because CVAT rejects all URLs with non-public IP addressses. However, I did test this case manually.

### Motivation and context

Fixes #4962.

### How has this been tested?

SDK unit tests + manual testing for remote files.

### Checklist

- ~[ ] I have increased versions of npm packages if it is necessary ([cvat-canvas](https://github.com/cvat-ai/cvat/tree/develop/cvat-canvas#versioning),

### License


zhiltsov-max (changes requested):
The implementation looks OK to me. However, I'd like to ask for few other changes:
1. The `tests/share` is better be called `tests/shared_drive` or similar. Otherwise, it looks like it's just a directory with shared data for tests (which is also needed, in perspective)
2. Probably, we need to add a similar test in the rest_api tests layer, there is no such test yet.
3. Probably, we should check task creation with more params. Maybe just add more/all params in the existing test, or add a new one.

SpecLad:
> 1. The tests/share is better be called tests/shared_drive or similar. Otherwise, it looks like it's just a directory with shared data for tests (which is also needed, in perspective)

Sure, I can do that.

> 2. Probably, we need to add a similar test in the rest_api tests layer, there is no such test yet.
> 3. Probably, we should check task creation with more params. Maybe just add more/all params in the existing test, or add a new one.

I think that goes beyond the scope of this PR. The tests I added are sufficient to make sure that the bugfix works; adding other tests would be a separate task.

zhiltsov-max (approved):

azhavoro:
This PR broke e2e test `case_107_connected_file_share`
```js
CypressError: `cy.exec('mv cypress/integration/actions_tasks3/assets/case_107/image_case_107_1.png cypress/integration/actions_tasks3/assets/case_107/image_case_107_1.png.bk')` failed because the command exited with a non-zero code.

Information about the failure:
Code: 1

Stderr:
mv: cannot stat 'cypress/integration/actions_tasks3/assets/case_107/image_case_107_1.png': No such file or directory
```

SpecLad:
My bad; didn't see that usage. I'll try to fix it.

PR #5062, opened 2022-10-09
Refactor server enums

zhiltsov-max:
### Motivation and context

There are lots of copy-paste around enums in the server implementation. This PR introduces common bases that take the common code an reduce boilerplate.

- Added `StrEnum` and `DjangoEnum` base classes for enums
- Refactored most enums on the server

### Checklist

### License


zhiltsov-max:
@kirill-sizov, please fix the pylint warnings in the develop branch:
```
************* Module organizations.views
cvat/apps/organizations/views.py:194:4: W0221: Variadics removed in overridden 'InvitationViewSet.perform_create' method (arguments-differ)
************* Module engine.views
cvat/apps/engine/views.py:309:4: W0221: Variadics removed in overridden 'ProjectViewSet.perform_create' method (arguments-differ)
cvat/apps/engine/views.py:822:4: W0221: Variadics removed in overridden 'TaskViewSet.perform_create' method (arguments-differ)
cvat/apps/engine/views.py:1741:4: W0221: Variadics removed in overridden 'IssueViewSet.perform_create' method (arguments-differ)
cvat/apps/engine/views.py:1816:4: W0221: Variadics removed in overridden 'CommentViewSet.perform_create' method (arguments-differ)

```

nmanovic:
@SpecLad , could you please review?

SpecLad:
Why not use the Django built-in `models.TextChoices` class? Seems like it does roughly the same thing as these new classes.

zhiltsov-max:
Probably it was added after most of the models. Ok. I've finally found a confirmation in the docs that TextChoices doesn't introduce new DB tables. So yes, probably it could be used instead. I've tried to use it, and it creates an extra migration for a typical declaration of members. Not sure if we want this.

SpecLad:
> I've tried to use it, and it creates an extra migration for a typical declaration of members.

What does the migration do?

zhiltsov-max:
Django enums add a display value for members. In our cases we just return the member name, but in Django enums it will be generated from the names into normal text, so it is almost always different. We can provide our own definitions as the second value:
```python
class YearInSchool(models.TextChoices):
        FRESHMAN = 'FR', _('Freshman')
```
Or try to override the default and reproduce our current behavior.

kirill-sizov:
> @kirill-sizov, please fix the pylint warnings in the develop branch:
> 
> ```
> ************* Module organizations.views
> cvat/apps/organizations/views.py:194:4: W0221: Variadics removed in overridden 'InvitationViewSet.perform_create' method (arguments-differ)
> ************* Module engine.views
> cvat/apps/engine/views.py:309:4: W0221: Variadics removed in overridden 'ProjectViewSet.perform_create' method (arguments-differ)
> cvat/apps/engine/views.py:822:4: W0221: Variadics removed in overridden 'TaskViewSet.perform_create' method (arguments-differ)
> cvat/apps/engine/views.py:1741:4: W0221: Variadics removed in overridden 'IssueViewSet.perform_create' method (arguments-differ)
> cvat/apps/engine/views.py:1816:4: W0221: Variadics removed in overridden 'CommentViewSet.perform_create' method (arguments-differ)
> ```

Yes, I will fix it in separate PR

SpecLad:
I think in most cases the Django auto-generated display value will be fine, and we can override it when it isn't.

But you didn't answer the question. I don't think the display values get saved to the DB, so what does the migration do?

zhiltsov-max:
@SpecLad, having `DimensionType` inherited from Django `TextField` we get this migration:

```python
field=models.CharField(choices=[('3d', 'Dim 3D'), ('2d', 'Dim 2D')], default='2d', max_length=2),
```

for the same enum declaration. It's not clear to me if we need these changes in DB for a simple refactoring, nor do I know if we can use these new values anywhere. I think, switching to `TextChoices` is a good idea, but the default label values need to be changed back to the existing values.

cvat/apps/engine/backup.py
```diff
@@ -45,7 +44,7 @@
 
 from cvat.apps.engine.location import StorageType, get_location_configuration
 
-class Version(Enum):
+class Version(DjangoEnum, StrEnum):
```
SpecLad, line 47: I don't think this is needed. This class is not used in models.

cvat/apps/engine/cloud_provider.py
```diff
@@ -25,29 +26,18 @@
 
 from cvat.apps.engine.log import slogger
 from cvat.apps.engine.models import CredentialsTypeChoice, CloudProviderChoice
+from cvat.apps.engine.utils import DjangoEnum, StrEnum
 
-class Status(str, Enum):
+class Status(DjangoEnum, StrEnum):
```
SpecLad, line 31: Likewise, both this and `Permissions` aren't used with Django models.

cvat/apps/engine/location.py
```diff
@@ -1,14 +1,14 @@
-from enum import Enum
+# Copyright (C) 2022 CVAT.ai Corporation
+#
+# SPDX-License-Identifier: MIT
 
 from cvat.apps.engine.models import Location
+from cvat.apps.engine.utils import DjangoEnum, StrEnum
 
-class StorageType(str, Enum):
+class StorageType(DjangoEnum, StrEnum):
```
SpecLad, line 8: Same here...

cvat/apps/engine/utils.py
```diff
@@ -133,3 +135,54 @@ def configure_dependent_job(queue, rq_id, rq_func, db_storage, filename, key):
             job_id=rq_job_id_download_file
         )
     return rq_job_download_file
+
+class StrEnum(str, Enum):
+    """
+    A enum with string items.
```
SpecLad, line 141:
```suggestion
    An enum with string items.
```

cvat/apps/engine/views.py
```diff
@@ -608,18 +608,24 @@ def _get_rq_response(queue, job_id):
...
+        PREVIEW = 'preview'
+        CONTEXT_IMAGE = 'context_image'
+
+    class ImageQualityChoices(DjangoEnum, StrEnum):
+        COMPRESSED = 'compressed'
+        ORIGINAL = 'original'
 
-        if not data_type or data_type not in possible_data_type_values:
+    def __init__(self, data_type, data_num, data_quality, task_dim):
+        if not data_type or data_type not in self.DataTypeChoices.values():
```
SpecLad, line 623:
```suggestion
        if data_type not in self.DataTypeChoices.values():
```

cvat/apps/organizations/views.py
```diff
@@ -31,11 +31,6 @@
         responses={
             '200': OrganizationReadSerializer(many=True),
         }),
-    update=extend_schema(
-        summary='Method updates an organization by id',
-        responses={
-            '200': OrganizationWriteSerializer,
-        }),
```
SpecLad, line 38: This doesn't seem relevant to this PR.

cvat/settings/base.py
```diff
@@ -562,6 +562,8 @@ def add_ssh_keys():
         'JobStage': 'cvat.apps.engine.models.StageChoice',
         'StorageType': 'cvat.apps.engine.models.StorageChoice',
         'SortingMethod': 'cvat.apps.engine.models.SortingMethod',
+        'WebhookType': 'cvat.apps.webhooks.models.WebhookTypeChoice',
+        'WebhookContentType': 'cvat.apps.webhooks.models.WebhookContentTypeChoice',
```
SpecLad, line 566: Can't we rename the classes themselves? Why do it here?

SpecLad (changes requested):

cvat/apps/engine/backup.py
zhiltsov-max: The reason for these changes is to replace the copy-pasted code for enum methods by a single implementation.

cvat/settings/base.py
zhiltsov-max: This is done here, because this component is responsible for such renames for better look of the generated schema. The class on the server can be renamed independently, if it is needed.

cvat/apps/organizations/views.py
zhiltsov-max: Moved to another PR

cvat/apps/engine/backup.py
SpecLad: But there is no copy-pasted code here.

SpecLad: ... and in some of the other cases I pointed out, it would be sufficient to inherit from `StrEnum` alone.

zhiltsov-max: Ok, I simplified/restored a couple of enums, which don't use additional methods now.

cvat/apps/engine/cloud_provider.py
zhiltsov-max: Both Status and Permissions have extra methods, that I would like to unify.

cvat/apps/engine/location.py
zhiltsov-max: Ok, removed DjangoEnum

SpecLad (approved):

PR #4863, opened 2022-08-26
Add webhooks

kirill-sizov:
### Motivation and context
- [x] Change event naming template from `<resource>_<action>` to `<aciton>:<resource>`

### Checklist

- [ ] I submit my changes into the `develop` branch
- [ ] I have added a description of my changes into [CHANGELOG](https://github.com/cvat-ai/cvat/blob/develop/CHANGELOG.md) file
- [ ] I have updated the [documentation](
  https://github.com/cvat-ai/cvat/blob/develop/README.md#documentation) accordingly
- [ ] I have added tests to cover my changes
- [ ] I have linked related issues ([read github docs](
  https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword))
- [ ] I have increased versions of npm packages if it is necessary ([cvat-canvas](https://github.com/cvat-ai/cvat/tree/develop/cvat-canvas#versioning),
[cvat-core](https://github.com/cvat-ai/cvat/tree/develop/cvat-core#versioning), [cvat-data](https://github.com/cvat-ai/cvat/tree/develop/cvat-data#versioning) and [cvat-ui](https://github.com/cvat-ai/cvat/tree/develop/cvat-ui#versioning))

### License

- [ ] I submit _my code changes_ under the same [MIT License](
  https://github.com/cvat-ai/cvat/blob/develop/LICENSE) that covers the project.
  Feel free to contact the maintainers if that's a concern.

PR #5078, opened 2022-10-11
Add missing migration for webhooks

zhiltsov-max:
### Motivation and context

It seems that a migration was done in #4863 , but it was not added to the repository. This PR adds the missing migration, which is generated by `python manage.py makemigrations`.

- Changed the webhook model `status_code` field type

### Checklist

### License


nmanovic (approved):

kirill-sizov:
Added migration quite similar with [this migration](https://github.com/opencv/cvat/blob/develop/cvat/apps/webhooks/migrations/0002_alter_webhookdelivery_status_code.py).

And the reason why the lists of available values differ is because of the version of Python being used. Because for different python version `http.HTTPStatus` can differ.

So I think we should make `WebhookDelivery.status_code` independent from Python version.

And for now I see only one way to do it: hardcode all possible status codes. @zhiltsov-max what do you think about it?

zhiltsov-max:
I think we can just store an integer without extra restrictions.

Issue #4179, opened 2022-01-14
Unexpected behaviour with task `meta`


Sigura:
### My actions before raising this issue
- [x] Read/searched [the docs](https://github.com/opencv/cvat/tree/master#documentation)
- [x] Searched [past issues](/issues)

### Steps to Reproduce (for bugs)
1. Make `manifest.jsonl` with own order (not abc) on aws (2D images)
2. Add `manifest.jsonl` to cloud storage
2. Create tasks with same `manifest.jsonl`
3. In task creation form files in expected order (as in `manifest.jsonl`)
4. After creation file names show in abc order in "player"
5. But files shown as it is in `manifest.jsonl` file

Export to COCO mess too – file names and order.

This endpoint shown files in expected order:
`cloudstorages/{store_id}/content?manifest_path=manifest.jsonl`

But `tasks/{new-task-id}/data/meta` return abc order. (looks like abc)

I checked folder `/home/django/data/storages/{store_id}/{name}/manifest.jsonl` it has expected order

### Possible Solution

Let users make own order in `manifest.jsonl` and use it as it is

And restore it from manifest or form `request.data['server_files']`

Steps:

1. Remove all unexpected `sorted` from code

Found related:
 * [task.py#L281](https://github.com/openvinotoolkit/cvat/blob/develop/cvat/apps/engine/task.py#L281)
 * [task.py#L129](https://github.com/openvinotoolkit/cvat/blob/develop/cvat/apps/engine/task.py#L129)
 * [media_extractors.py#L126](https://github.com/openvinotoolkit/cvat/blob/develop/cvat/apps/engine/media_extractors.py#L126)

2. Fix order after `serializer` creation

~~and I don't know why `data['server_files']` income reversed?!!~~

now I know why it may come in any order, because find the root of this evil

https://github.com/openvinotoolkit/cvat/blob/develop/cvat/apps/engine/views.py#L630

Because looks like serializer does not guaranties income order, is not it?

But it fixable there – just rebuild it if from `request.data['server_files']`.

### Your Environment
```
Author: Maria Khrustaleva <maya17grd@gmail.com>
Date:   Wed Dec 1 16:57:31 2021 +0300

    Remove key_file_path field (#3959)

    * Remove key_file_path field
    * fix
    * Update comparison
```

Sigura:
I found cache
```
> docker exec -u root -it cvat_db sh
> psql -U root -d cvat
[...]

PR #5083, opened 2022-10-11
Fix predefined sorting for task data

### LINKED ISSUE #5061: cvat-cli `--sorting-method predefined` results in seemingly random order of images
- [x ] Read/searched [the docs](https://github.com/cvat-ai/cvat/tree/master#documentation)
- [ x] Searched [past issues](/issues)

using the current master (v.2.2.0) and cvat-cli for create task and sorting method predefined will result in a seemingly random order of images in the task.
Using version 2.1.0 works as expected but only sometimes.
(creating many tasks with a script)

zhiltsov-max:
### Motivation and context

Fixes #5061, #4179

- Added a way to declare custom file ordering for the local task data uploads via TUS protocol
- Added an option to use a manifest to support the `predefined` sorting method
  - This file is required for the `predefined` sorting mode with image archives
- Fixed file ordering when tasks are created from SDK or CLI in the `predefined` sorting mode
- Added more tests for task data uploading API

The uploading protocol is implemented:

The user specifies `sorting_method=predefined` if the task creation request. Then the data is uploaded.

1. Client files uploading
1.1. The files are uploaded as separate files (using the TUS protocol) or grouped files (using the `Upload-Multiple` requests). 
1.2. The `Upload-Finish` request comes (or its unlabeled legacy equivalent). The new optional field can be supplied: `upload_file_order` - a list of strings. It allows to override the input file order, if necessary, and is only valid with the `predefined` sorting method specified.
1.2.1. If the field is empty or missing, the client files in the data requests are considered ordered.
1.2.2. If the field is not empty, a list containing the file list in the required order is expected in the `upload_file_order` field.
1.2.2.1. If there are `client_files` in the request, the files are sorted
1.2.2.2. If file lists mismatch, an explanatory error is raised.

2. Data processing
2.1. At this point, all `*_files` are considered ordered as requested.
2.2. Require a metafile for zip uploads with predefined sorting. The file is expected to accompany the uploaded zip file, not to be inside of the archive.
2.3. If there is a metafile in the input data, files are ordered after the metafile.
2.3.1. If the data is extracted from cloud, only the specified subset of the files is kept in the manifest.
2.3.2. If the upload data doesn't exist in the metafile, an error is raised.
2.3.3. A `job_file_mapping` has higher priority than metafile, if specified.

### Checklist

### License


zhiltsov-max:
Problems:
1. If we send data with TUS, the data is stored in a directory, and then it is returned unordered. It can also be sent unordered. Then, when we try to get the list of the uploaded files, we get the files in an unspecified order (https://github.com/opencv/cvat/issues/5061).

2. If we send an archive, the data is returned as the archive reports, i.e. the file order is unspecified.


Possible solutions:
1. Require a metadata file when `sorting_method=predefined`
  
**REJECTED**: Too obligatory for users. We want to avoid using metafiles until really necessary

2. 
  2.1. Require an upload spec (a request) for TUS uploads
     2.1.1. Introduce a new optional request body in the `Upload-Start` request. It has 2 fields: `ordered` (boolean, default=False), `files` (list of strings, optional)
       2.1.1.1. If `ordered=True`, the files are considered ordered. Write the input ordering into a special metafile (e.g. `upload_dir/__upload_auto_file_ordering__`) and update it on each input file request.
       2.1.1.2. If `ordered=False` and "files" is not empty, the "files" field is expected to contain the file list in the required order. A temporary metafile (e.g. `upload_dir/__upload_custom_file_ordering__`) is generated from the input data.
       2.1.1.3. If "ordered"=False and "files" is empty or missing, the files are considered unordered. No extra actions are done.
       2.1.1.4. The files are being uploaded as separate files and grouped files. If the input file has a reserved name, fail the request with an error. If a temporary metafile for auto ordering is found in the upload location, the file is updated.
       2.1.1.5. The `Upload-Finished` request comes.
       2.1.1.5.1. If no metafiles found in the upload location, do not report any error. We don't raise errors to support existing cases where a manifest file is sent in the uploaded data, but is not supposed for sorting.
       2.1.1.5.2. Sort the input files (the `client_files` field) using any of the temporary metafiles found.

  2.2. Require a metafile for zip uploads with predefined sorting. The file is expected to accompany the uploaded zip file, not to be inside of the archive.

**BEING TESTED**

@nmanovic , @SpecLad , @azhavoro  - do you have any comments about the new solution?

SpecLad:
Is the `ordered` field necessary? Seems like the file list should be enough.

zhiltsov-max:
> Is the ordered field necessary? Seems like the file list should be enough.

It's hard to say, probably no. We could encode `ordered=True` into the default value for the array, but it looks like there are some clients that depend on the existing behavior (e.g. that the files are either in the direct or reversed order), and I'm not sure if it is good to break their code. The existing behavior is different - the files are returned as `os.listdir()` returns (which is unspecified, but in practice it gives 2 variants for some reason).

zhiltsov-max:
I decided to remove the backward compatibility mode. It turned out that Django was sorting the `client_files` field lexicographically in all cases (with and without TUS), so the `predefined` sorting was never really working for local uploads. The old behavior can be reproduced with just the `lexicographical` sorting method now. For `predefined`, I fixed behavior in single-request uploads, and I implemented a way to optionally specify the input file order. I decided not to require this just for TUS to be consistent with the simple, single-request version.

cvat-sdk/cvat_sdk/core/uploading.py
```diff
@@ -341,43 +345,99 @@ def upload_files(
...
+        if not candidates:
+            return None
+        elif len(candidates) != 1:
+            raise TooManyManifests(candidates=candidates)
+        else:
+            return candidates[0]
+
+    @staticmethod
+    def _default_upload_metainfo_filename() -> str:
+        return "manifest.jsonl"
```
SpecLad, line 414: This seems like it should be a constant.

cvat-sdk/cvat_sdk/core/uploading.py
```diff
@@ -341,43 +345,99 @@ def upload_files(
...
+
+    def _generate_upload_metainfo_file(self, resources: Sequence[str]) -> StringIO:
```
SpecLad, line 416: The `StringIO` object seems unnecessary, as the result of this function gets immediately written into a real file anyway. So you could just take the file as an argument and write there.

cvat-sdk/cvat_sdk/core/uploading.py
```diff
@@ -341,43 +345,99 @@ def upload_files(
...
+        contents = StringIO()
+        self._write_base_information(contents)
+
+        records = []
+        for file in resources:
+            name, ext = osp.splitext(osp.basename(file))
+            record = {"name": name, "extension": ext}
+            records.append(record)
```
SpecLad, line 424: Why collect the records into a list first? You could just write each of them out immediately.

CHANGELOG.md
```diff
@@ -41,7 +41,7 @@ non-ascii paths while adding files from "Connected file share" (issue #4428)
   (<https://github.com/opencv/cvat/pull/5057>)
 
 ### Security
-- TDB
+- Fixed an arbitrary server file write vulnerability (<https://github.com/opencv/cvat/pull/5083>)
```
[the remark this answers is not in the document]
zhiltsov-max: Removed. I couldn't prove it is exploitable.

cvat-cli/src/cvat_cli/parser.py
```diff
@@ -122,7 +122,7 @@ def make_cmdline_parser() -> argparse.ArgumentParser:
     task_create_parser.add_argument(
         "--completion_verification_period",
         dest="status_check_period",
-        default=20,
+        default=5,
```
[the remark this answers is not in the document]
zhiltsov-max: It improves user experience, but doesn't deserves a separate PR.

cvat-sdk/cvat_sdk/core/uploading.py
```diff
@@ -341,43 +345,99 @@ def upload_files(
...
-                meta={"filename": osp.basename(filename)},
-                pbar=pbar,
-                logger=self._client.logger.debug,
-            )
+            self._tus_finish_upload(url, fields=kwargs)
 
-        self._tus_finish_upload(url, fields=kwargs)
+    def _has_upload_metainfo_file(self, resources: Sequence[str]) -> bool:
+        file = self._find_upload_metainfo_file(resources)
+        return file and osp.isfile(file)
```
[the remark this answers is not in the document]
zhiltsov-max: Not actual anymore

cvat-sdk/cvat_sdk/core/uploading.py
zhiltsov-max: Not actual anymore

cvat-sdk/cvat_sdk/core/uploading.py
zhiltsov-max: Not actual anymore

cvat-sdk/cvat_sdk/core/uploading.py
zhiltsov-max: Not actual anymore

zhiltsov-max:
Seems ready for review

site/content/en/docs/manual/advanced/dataset_manifest.md
```diff
@@ -11,11 +11,62 @@ description: This section on [GitHub](https://github.com/cvat-ai/cvat/tree/devel
 
 <!--lint disable heading-style-->
 
-### Steps before use
+## Overview
```
SpecLad, line 14: I think the page title should also be adjusted, because it's not just about the command line anymore.

site/content/en/docs/manual/advanced/dataset_manifest.md
```diff
@@ -11,11 +11,62 @@ description: This section on [GitHub](https://github.com/cvat-ai/cvat/tree/devel
...
+of network traffic used and speed up the task creation process. However, they can also
+be used in other cases.
+
+A manifest file is a text file in the JSONL format.
+
+## How and when to use manifest files
+
+Manifest files can be used in the following cases:
+- A video file or a set of images is used as the data source and
+  the caching mode is enabled (`use_cache`). [Read mode](/docs/manual/advanced/data_on_fly/)
```
SpecLad, line 31: What is `use_cache`? It's not clear from this text. The linked page doesn't explain it either.

site/content/en/docs/manual/advanced/dataset_manifest.md
```diff
@@ -11,11 +11,62 @@ description: This section on [GitHub](https://github.com/cvat-ai/cvat/tree/devel
...
+- The data is located in a cloud storage. [Read more](/docs/manual/basics/cloud-storages/)
+- The `predefined` file sorting method is specified. [Read more](/docs/manual/basics/creating_an_annotation_task/#sorting-method)
+
+### The predefined sorting method
+
+Independently of the file source used, when the `predefined`
+sorting method is selected, the source files will be ordered according
+to the `.jsonl` manifest file, if it is found in the input list of files.
+For archives (e.g. `.zip`), a manifest file (`*.jsonl`) required when using
```
SpecLad, line 40:
```suggestion
For archives (e.g. `.zip`), a manifest file (`*.jsonl`) is required when using
```

utils/dataset_manifest/core.py
```diff
@@ -348,20 +364,19 @@ class _ManifestManager(ABC):
     }
 
     def _json_item_is_valid(self, **state):
-        for item in self._requared_item_attributes:
+        for item in self._required_item_attributes:
             if state.get(item, None) is None:
-                raise Exception(f"Invalid '{self.manifest.name} file structure': '{item}' is required, but not found")
+                raise MissingFieldError(
+                    f"Invalid '{self.manifest.name} file structure': "
```
SpecLad, line 370:
```suggestion
                    f"Invalid '{self.manifest.name}' file structure: "
```

utils/dataset_manifest/core.py
```diff
@@ -487,24 +502,21 @@ def _write_core_part(self, file, _tqdm):
...
-    # pylint: disable=arguments-differ
-    @_set_index
-    def create(self, _tqdm=None):
+    def create(self, *, _tqdm=None): # pylint: disable=arguments-differ
         """ Creating and saving a manifest file """
         if not len(self._reader):
-            with NamedTemporaryFile(mode='w', delete=False)as tmp_file:
-                self._write_core_part(tmp_file, _tqdm)
-            temp = tmp_file.name
+            tmp_file = StringIO()
```
SpecLad, line 508: Why do we need a `tmp_file` at all? Why not just write directly to `manifest_file`?

utils/dataset_manifest/core.py
```diff
@@ -707,27 +720,27 @@ def validators(self):
...
+        #     raise InvalidFieldValueError('Incorrect height field')
 
 def is_manifest(full_manifest_path):
-    return _is_video_manifest(full_manifest_path) or \
-        _is_dataset_manifest(full_manifest_path)
+    return is_video_manifest(full_manifest_path) or \
+        is_dataset_manifest(full_manifest_path)
 
-def _is_video_manifest(full_manifest_path):
+def is_video_manifest(full_manifest_path):
```
SpecLad, line 740: This seems like an unnecessary change. `_is_video_manifest` isn't used from other modules, so it doesn't need to be public.

utils/dataset_manifest/errors.py
```diff
@@ -0,0 +1,23 @@
...
+    pass
+
+class InvalidManifestFieldError(InvalidManifestError):
+    pass
+
+class MissingFieldError(InvalidManifestFieldError):
+    pass
+
+class InvalidFieldValueError(InvalidManifestFieldError):
+    pass
```
SpecLad, line 23: These fine-grained exception types seem unnecessary. They're not handled any differently from the base type, and this is not a library, so there's no external user who could handle them differently.

SpecLad (changes requested):

utils/dataset_manifest/core.py
```diff
@@ -148,18 +150,20 @@ def __iter__(self):
...
-                use_image_hash=False,
+                *args,
                 start = 0,
                 step = 1,
                 stop = None,
-                *args,
+                meta=None,
+                sorting_method=SortingMethod.PREDEFINED,
+                use_image_hash=False,
+                with_image_info=True,
```
SpecLad, line 160: Is this argument actually supplied anywhere? I don't see it.

zhiltsov-max: It is used in tests primarily. Probably, it should be added to CLI to be really useful.

utils/dataset_manifest/core.py
```diff
@@ -273,6 +279,14 @@ def name(self):
         return os.path.basename(self._path) if not self._upload_dir \
             else os.path.relpath(self._path, self._upload_dir)
 
+    @abstractmethod
```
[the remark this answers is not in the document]
zhiltsov-max: Seems to be so. Removed

utils/dataset_manifest/core.py
zhiltsov-max: I'm not sure. Probable, the reason is to avoid creating the file and the removing it in case of an error.

utils/dataset_manifest/core.py
```diff
@@ -24,14 +26,14 @@ def __init__(self, source_path, chunk_size, force):
 
         with closing(av.open(self.source_path, mode='r')) as container:
             video_stream = VideoStreamReader._get_video_stream(container)
-            isBreaked = False
+            stop = False
```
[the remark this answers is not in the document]
zhiltsov-max: Fixed

utils/dataset_manifest/core.py
```diff
@@ -148,18 +150,20 @@ def __iter__(self):
 class DatasetImagesReader:
     def __init__(self,
                 sources,
-                meta=None,
-                sorting_method=SortingMethod.PREDEFINED,
-                use_image_hash=False,
+                *args,
```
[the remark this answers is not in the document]
zhiltsov-max: Fixed

utils/dataset_manifest/core.py
```diff
@@ -334,11 +348,13 @@ def __getitem__(self, number):
     def __len__(self):
         return len(self._index)
 
+    def __iter__(self):
+        yield from self._index.values()
```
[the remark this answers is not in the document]
zhiltsov-max: > because self._index can be loaded from a JSON file, the order of entries in which can be arbitrary

It can, but it should not be a problem in realistic scenarios.

utils/dataset_manifest/core.py
```diff
@@ -334,11 +348,13 @@ def __getitem__(self, number):
...
+
 def _set_index(func):
     def wrapper(self, *args, **kwargs):
-        func(self, *args,  **kwargs)
-        if self._create_index:
-            self.set_index()
+        self.set_index()
+        return func(self, *args,  **kwargs)
```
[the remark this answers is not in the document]
zhiltsov-max: Removed

utils/dataset_manifest/core.py
zhiltsov-max: It can be used in the server, which uses this module as a package.

SpecLad: But _is_ it used in the server? Doesn't seem like it.

utils/dataset_manifest/core.py
SpecLad: I think I figured it out. It's because `VideoStreamReader.__len__` can return `None` until you iterate over it once, and `_write_core_part` needs to know the final length.

utils/dataset_manifest/core.py
SpecLad: I agree.

utils/dataset_manifest/core.py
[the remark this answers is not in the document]
SpecLad: >  It doesn't use __len__ and requires __getitem__ to throw StopIteration.

`IndexError` also works (and makes more sense for `__getitem__`).

> For me, it looks extra.

What does that mean?

utils/dataset_manifest/core.py
zhiltsov-max: 2 other functions (`is_dataset_manifest` and `is_manifest`) are used now. While it is not used currently, it looks natural to have them all public.

site/content/en/docs/manual/advanced/dataset_manifest.md
```diff
@@ -126,3 +171,14 @@ A manifest file contains some intuitive information and some specific like:
 {"name":"image2","extension":".jpg","width":183,"height":275,"meta":{"related_images":[]},"checksum":"4b4eefd03cc6a45c1c068b98477fb639"}
 {"name":"image3","extension":".jpg","width":301,"height":167,"meta":{"related_images":[]},"checksum":"0e454a6f4a13d56c82890c98be063663"}
 ```
+
+If the images are available to the server at the time of task creation,
+extra fields can be omitted:
```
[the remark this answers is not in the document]
zhiltsov-max: Currently no, it can be added as discussed below.

site/content/en/docs/manual/advanced/dataset_manifest.md
```diff
@@ -11,11 +11,62 @@ description: This section on [GitHub](https://github.com/cvat-ai/cvat/tree/devel
...
+    --entrypoint '/usr/bin/bash' \
+    cvat/server \
+    -c 'python -m pip install -r utils/dataset_manifest/requirements.txt && python utils/dataset_manifest/create.py --output-dir /local /local/<path/to/sources>
+```
+
+Make sure to adapt the command to your file locations.
+
+### Use the script directly
+
+**Ubuntu:20.04**
```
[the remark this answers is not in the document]
zhiltsov-max: Changed

site/content/en/docs/manual/advanced/dataset_manifest.md
zhiltsov-max: Changed

site/content/en/docs/manual/advanced/dataset_manifest.md
zhiltsov-max: Updated, I tried to make it a bit clearer.

site/content/en/docs/manual/advanced/dataset_manifest.md
```diff
@@ -11,11 +11,62 @@ description: This section on [GitHub](https://github.com/cvat-ai/cvat/tree/devel
 
 <!--lint disable heading-style-->
 
-### Steps before use
+## Overview
 
-When used separately from Computer Vision Annotation Tool(CVAT), the required dependencies must be installed
+When a new task is created in CVAT, we need to tell where to take the dataset sources.
```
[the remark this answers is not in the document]
zhiltsov-max: Updated

utils/dataset_manifest/core.py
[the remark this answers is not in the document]
zhiltsov-max: > Maybe it shouldn't, but it still could. I think it's worthwhile to defend against rare occurrences.

The index file should only be created by us from the given manifest file. It is always expected to be loaded from a file. If it is invalid for some reason, it means there is either a bug or the user (attacker) tries to hack something.

> `IndexError` also works

Ok, I need to try this option.

cvat/apps/engine/mixins.py
```diff
@@ -93,9 +94,27 @@ def __init__(self, request):
...
+    a. A single Data request
+    b.1. An Upload-Start request
+    b.2.a. The regular TUS protocol requests (Upload-Length + Chunks)
+    b.2.b. Upload-Multiple requests
+    b.3. An Upload-Finished request
+
+    Requests:
+    - Data - POST, no extra headers or 'Upload-Start' + 'Upload-Finish' headers
+    - Upload-Start - POST, has an 'Upload-Start' header
+    - Upload-Length - HEAD, has an 'Upload-Length' header (read the TUS protocol)
```
SpecLad, line 112:
```suggestion
    - Upload-Length - POST, has an 'Upload-Length' header (read the TUS protocol)
```

cvat/apps/engine/mixins.py
```diff
@@ -93,9 +94,27 @@ def __init__(self, request):
...
+    - Chunk - PATCH (read the TUS protocol)
```
SpecLad, line 113:
```suggestion
    - Chunk - HEAD/PATCH (read the TUS protocol)
```

cvat/apps/engine/mixins.py
```diff
@@ -217,32 +236,55 @@ def append_tus_chunk(self, request, file_id):
                                       extra_headers={'Upload-Offset': tus_file.offset,
                                                      'Upload-Filename': tus_file.filename})
 
-    def validate_filename(self, filename):
+    def validate_uploaded_file_name(self, filename: str) -> bool:
+        """
+        Checks the file name to be valid.
```
SpecLad, line 241: It would be useful to explain what it means for a file name to be valid.

cvat/apps/engine/mixins.py
```diff
@@ -217,32 +236,55 @@ def append_tus_chunk(self, request, file_id):
                                       extra_headers={'Upload-Offset': tus_file.offset,
                                                      'Upload-Filename': tus_file.filename})
 
-    def validate_filename(self, filename):
+    def validate_uploaded_file_name(self, filename: str) -> bool:
```
SpecLad, line 239: As long as you're renaming stuff, I think it would be better to call this something like `is_valid_uploaded_file_name`, to make it clear that it returns a boolean value.

cvat/apps/engine/mixins.py
```diff
@@ -146,21 +165,21 @@ def upload_data(self, request):
...
-            return self.init_tus_upload(request)
+            return self.init_file_upload(request)
         elif bulk_file_upload:
-            return self.append(request)
+            return self.append_files(request)
         else: # backward compatibility case - no upload headers were found
             return self.upload_finished(request)
 
-    def init_tus_upload(self, request):
+    def init_file_upload(self, request):
```
SpecLad, line 176: Why rename this? I think the old name is better, it shows that this is handling for the case where the file is uploaded with TUS. Same for `append_tus_chunk`.

cvat/apps/engine/mixins.py
```diff
@@ -217,32 +236,55 @@ def append_tus_chunk(self, request, file_id):
...
+
+        client_files = self._get_request_client_files(request)
         if client_files:
             upload_dir = self.get_upload_dir()
             for client_file in client_files:
-                with open(os.path.join(upload_dir, client_file['file'].name), 'ab+') as destination:
+                filename = client_file['file'].name
+                if not self.validate_uploaded_file_name(filename):
+                    return Response(status=status.HTTP_400_BAD_REQUEST,
+                        data=f"File name {filename} is not allowed")
```
SpecLad, line 270:
```suggestion
                        data=f"File name {filename} is not allowed", content_type="text/plain")
```

cvat/apps/engine/models.py
```diff
@@ -409,6 +409,7 @@ class ClientFile(models.Model):
     class Meta:
         default_permissions = ()
         unique_together = ("data", "file")
+        # note that Django will sort the files by filenames automatically
```
SpecLad, line 412: Is this documented somewhere?

cvat/apps/engine/serializers.py
```diff
@@ -1002,6 +1002,19 @@ class LogEventSerializer(serializers.Serializer):
 class AnnotationFileSerializer(serializers.Serializer):
     annotation_file = serializers.FileField()
 
+class TusUploadingMetaSerializer(serializers.Serializer):
+    """
+    Allows to specify TUS file uploading details.
+
+    To state that the input files are sent in the correct order,
+    pass an empty list of files in the 'files' field. A list of
+    strings is expected to be a list file names in the required order.
```
SpecLad, line 1011:
```suggestion
    strings is expected to be a list of file names in the required order.
```

cvat/apps/engine/serializers.py
```diff
@@ -1002,6 +1002,19 @@ class LogEventSerializer(serializers.Serializer):
...
+    """
+
+    files = serializers.ListField(
+        child=serializers.CharField(max_length=1024), default=[], allow_empty=True)
```
SpecLad, line 1015:
```suggestion
        child=serializers.CharField(max_length=1024), default=list, allow_empty=True)
```

To avoid sharing the same list between multiple instances.

SpecLad:
FYI, I think you can remove "[Dependent]" from the title.

cvat/apps/engine/mixins.py
zhiltsov-max: I wanted to hide the fact that TUS is used from the class API. From my point, it should be an insignificant implementation detail.

cvat/apps/engine/models.py
zhiltsov-max: I couldn't find neither the reason why it happens, nor mentions in documentation. It looks like it's just a detail of the implementation in Django.

cvat/apps/engine/mixins.py
SpecLad: It can't be an implementation detail, because these methods can only be called with request objects that match the requirements of the TUS protocol. So I think they're pretty closely linked with TUS.

cvat/apps/engine/models.py
SpecLad: Then you shouldn't claim it in such definitive terms. You could note that Django may rearrange the files (although so far I don't see why it's relevant).

cvat/apps/engine/task.py
```diff
@@ -294,6 +297,57 @@ def _download_data(urls, upload_dir):
...
+    """
+
+    if is_dataset_manifest(path):
+        return ImageManifestManager(path)
+    elif not is_manifest(path):
+        raise InvalidManifestError(
+            "Can't recognize a manifest file in "
+            "the uploaded file '{}'".format(os.path.basename(path))
+        )
+    return None
```
SpecLad, line 317: The annotated result type is `ImageManifestManager`, and looking at this function's usage, a `None` value is not expected, so `None` shouldn't be returned.

Dockerfile
```diff
@@ -48,6 +48,8 @@ ENV PATH="/opt/venv/bin:${PATH}"
 RUN python3 -m pip install --no-cache-dir -U pip==22.0.2 setuptools==60.6.0 wheel==0.37.1
 COPY cvat/requirements/ /tmp/requirements/
 RUN DATUMARO_HEADLESS=1 python3 -m pip install --no-cache-dir -r /tmp/requirements/${DJANGO_CONFIGURATION}.txt
+COPY utils/dataset_manifest/ /tmp/dataset_manifest/
+RUN python3 -m pip install --no-cache-dir -r /tmp/dataset_manifest/requirements.txt
```
SpecLad, line 52: It would be useful to merge this `pip install` with the previous one, so that if there are any conflicts between the two requirements files, pip would complain about them.

cvat/apps/engine/views.py
```diff
@@ -851,16 +859,179 @@ def jobs(self, request, pk):
...
+
+        with open(path) as f:
+            return [line.strip() for line in f]
+
+    def _prepare_upload_metafile_entry(self, filename: str) -> str:
+        filename = osp.normpath(filename.strip())
+        upload_dir = self.get_upload_dir()
+        if filename.startswith(upload_dir):
+            filename = osp.relpath(filename, upload_dir)
+        return filename.lstrip("/.")
```
SpecLad, line 907: What is this meant to do? If a file name starts with a dot, then this will remove it, which seems like a bug.

cvat/apps/engine/views.py
```diff
@@ -851,16 +859,179 @@ def jobs(self, request, pk):
...
+    def _read_tus_upload_meta_file(self, path: str) -> Sequence[str]:
+        """
+        Reads the input file and returns its contents.
+        """
+
+        with open(path) as f:
+            return [line.strip() for line in f]
+
+    def _prepare_upload_metafile_entry(self, filename: str) -> str:
+        filename = osp.normpath(filename.strip())
```
SpecLad, line 903: Why is there a `strip` here?

cvat/apps/engine/views.py
```diff
@@ -851,16 +859,179 @@ def jobs(self, request, pk):
...
+            # file uploading can be restarted. In such case we'll
+            # just reuse the first position of this file
+            pass
+
+    class _InvalidMetafileError(Exception):
+        """
+        Indicates an invalid upload metafile found
+        """
+
+    def _sort_uploaded_files(self, uploaded_files: List[Dict[str, Any]]) -> List[str]:
```
SpecLad, line 930: The type of `uploaded_files` doesn't seem right, given the usage - shouldn't it be `List[str]`?

cvat/apps/engine/views.py
```diff
@@ -851,16 +859,179 @@ def jobs(self, request, pk):
...
+            if osp.isfile(metafile):
+                self._maybe_append_tus_upload_metafile_entry(metafile, response['Upload-Filename'])
+
+        return response
+
+    # UploadMixin method
+    def append_files(self, request):
+        client_files = self._get_request_client_files(request)
+        if self._is_data_uploading() and client_files:
+            metafile = osp.join(self.get_upload_dir(), self._get_tus_input_ordering_metafile_name())
```
SpecLad, line 987: As I understand, the client files are unordered - but then why would we record their order in the manifest? It's going to be meaningless. I think this method of uploading should just be disallowed when the sorting order is predefined and there was no initial file list.

cvat/apps/engine/views.py
```diff
@@ -851,16 +859,179 @@ def jobs(self, request, pk):
...
+        return [
+            self._get_tus_input_ordering_metafile_name(),
+            self._get_tus_custom_ordering_metafile_name(),
+        ]
+
+    # UploadMixin method
+    def validate_uploaded_file_name(self, filename: str) -> bool:
+        return (
+            super().validate_uploaded_file_name(filename) and
+            osp.basename(filename) not in self._get_tus_metafile_names()
```
SpecLad, line 891: IMO, we should just reserve some subset of filenames (e.g. all names beginning with `__cvat`) for internal use, and reject them in the base class's implementation of `validate_uploaded_file_name`. That way, we won't risk colliding with a file name already being used by someone if we need to add more special names later.

cvat/apps/engine/views.py
```diff
@@ -928,11 +1128,58 @@ def upload_finished(self, request):
...
+        description=textwrap.dedent("""\
+            Allows to upload data to a task.
+            Supports the TUS open file uploading protocol (https://tus.io/).
+
+            Implements the following protocols:
+            a. A single Data request
+            b.1. An Upload-Start request
+            b.2.a. The regular TUS protocol requests (Upload-Length + Chunks)
+            b.2.b. Upload-Multiple requests
+            b.3. An Upload-Finished request
```
SpecLad, line 1141: It's not clear from the description that some of the list levels represent alternatives, while some represent sequential requests. I think it would be more instructive to spell it out, e.g.:

-----
To use this endpoint, send either a single Data request, or the following sequence of requests:

1. An Upload-Start request.
2. One or more of Upload-Length or Upload-Multiple request. For each `Upload-Length` request send, complete the upload using one or more `Chunk` requests.
3. An Upload-Finish request.
-----

It might also be worth mentioning that `Chunk` requests are send not to this endpoint, but to a file-specific one.

cvat/apps/engine/views.py
```diff
@@ -928,11 +1128,58 @@ def upload_finished(self, request):
...
+
+            Requests:
+            - Data - POST, no extra headers or 'Upload-Start' + 'Upload-Finish' headers.
```
SpecLad, line 1144:
```suggestion
            - Data - POST, no extra headers.
```

Seems redundant.

cvat/apps/engine/views.py
```diff
@@ -928,11 +1128,58 @@ def upload_finished(self, request):
...
+              Contains Data in the body.
```
SpecLad, line 1145:
```suggestion
              Contains data in the body.
```

cvat/apps/engine/views.py
```diff
@@ -928,11 +1128,58 @@ def upload_finished(self, request):
...
+            pass an empty list of files in the 'files' field. If the files are sent unordered,
+            the ordered file list is expected in the 'files' field. It must be a list of string
+            file names, relatively to the dataset root.
+            Example:
+            files = [
+                cats/cat_1.jpg,
+                dogs/dog2.jpg,
+                image_3.png,
+                ...
+            ]
```
SpecLad, line 1165: Shouldn't it be something like:

```json
{
    "files": [
        "cats/cat_1.jpg",
        "dogs/dog2.jpg",
        "image_3.png",
        ...
    ]
}
```

instead? We're talking about the request body here, which is presumably going to be in JSON.

cvat/apps/engine/views.py
```diff
@@ -928,11 +1128,58 @@ def upload_finished(self, request):
...
+            - Data - POST, no extra headers or 'Upload-Start' + 'Upload-Finish' headers.
+              Contains Data in the body.
+            - Upload-Start - POST, has an 'Upload-Start' header.
+              Can contain upload metainfo in the body.
+            - Upload-Length - HEAD, has an 'Upload-Length' header (read the TUS protocol)
+            - Chunk - PATCH (read the TUS protocol).
+            - Upload-Finish - POST, has an 'Upload-Finish' header. Can contain Data in the body.
+            - Upload-Multiple - POST, has a 'Upload-Multiple' header. Contains Data in the body.
+
+            The 'Upload-Start' request allows to specify file order for TUS requests.
```
SpecLad, line 1153: Wouldn't it also work for non-TUS (Upload-Multiple) requests?

cvat/apps/engine/views.py
```diff
@@ -928,11 +1128,58 @@ def upload_finished(self, request):
...
+            b.2.b. Upload-Multiple requests
+            b.3. An Upload-Finished request
+
+            Requests:
+            - Data - POST, no extra headers or 'Upload-Start' + 'Upload-Finish' headers.
+              Contains Data in the body.
+            - Upload-Start - POST, has an 'Upload-Start' header.
+              Can contain upload metainfo in the body.
+            - Upload-Length - HEAD, has an 'Upload-Length' header (read the TUS protocol)
+            - Chunk - PATCH (read the TUS protocol).
```
SpecLad, line 1149:
```suggestion
            - Upload-Length - POST, has an 'Upload-Length' header (see the TUS specification)
            - Chunk - HEAD/PATCH (see the TUS specification).
```

cvat/apps/engine/views.py
```diff
@@ -928,11 +1128,58 @@ def upload_finished(self, request):
...
+            - Upload-Finish - POST, has an 'Upload-Finish' header. Can contain Data in the body.
+            - Upload-Multiple - POST, has a 'Upload-Multiple' header. Contains Data in the body.
+
+            The 'Upload-Start' request allows to specify file order for TUS requests.
+            This may be needed because files in these requests and the requests themselves
+            can come (or be sent) unordered. To state that the input files are sent ordered,
+            pass an empty list of files in the 'files' field. If the files are sent unordered,
+            the ordered file list is expected in the 'files' field. It must be a list of string
+            file names, relatively to the dataset root.
```
SpecLad, line 1158:
```suggestion
            file paths, relative to the dataset root.
```

cvat/apps/engine/views.py
```diff
@@ -928,11 +1128,58 @@ def upload_finished(self, request):
...
+                dogs/dog2.jpg,
+                image_3.png,
+                ...
+            ]
+
+            Independently of the file declaration field used
+            ('client_files', 'server_files', etc.), when the 'predefined'
+            sorting method is selected, the uploaded files will be ordered according
+            to the '.jsonl' manifest file, if it is found in the list of files.
+            For archives (e.g. '.zip'), a manifest file ('*.jsonl') required when using
```
SpecLad, line 1171:
```suggestion
            For archives (e.g. '.zip'), a manifest file ('*.jsonl') is required when using
```

cvat/apps/engine/views.py
```diff
@@ -928,11 +1128,58 @@ def upload_finished(self, request):
...
+            the 'predefined' file ordering. Such file must be provided next to the archive
+            in the list of files. Read more about manifest files here:
+            https://opencv.github.io/cvat/docs/manual/advanced/dataset_manifest/
+
+            After all data is sent, the operation status can be retrieved in the /status endpoint.
```
SpecLad, line 1176:
```suggestion
            After all data is sent, the operation status can be retrieved via the /status endpoint.
```

cvat/apps/engine/views.py
```diff
@@ -886,19 +1057,48 @@ def upload_finished(self, request):
...
             data['copy_data'] = serializer.validated_data['copy_data']
+
+            if data['client_files']:
+                # Django will sort client_files by filename automatically, which conflicts
+                # with the "predefined" sorting method. We need to restore the upload file
+                # order. This is not needed for other *_files fields, because they are
+                # not being sorted.
+                try:
+                    if data['sorting_method'] == models.SortingMethod.PREDEFINED:
+                        data['client_files'] = self._sort_uploaded_files(data['client_files'])
```
OutSorcerer, line 1098: Should sorting also be applied to `server_files`?

Since we have issues with file order when creating a task with files on a connected file share while using a POST request to `/api/tasks/{id}/data` API, where files are passed as `server_files`.

zhiltsov-max: Hi, thank you for the comment. In all cases except client_files with TUS uploading, we pass the ordered list of files in the request. Could you please describe the use case and the problem in more details?

OutSorcerer: Hi! We use REST API of CVAT to programmatically create CVAT tasks. Our files are mostly stored on a mounted file share, so we pass them as `server_files`. 

The order of files of the resulting CVAT task does not match the original order in a list of  `server_files` passed to a POST request to `/api/tasks/{id}/data`. 

I launched CVAT from `zm/fix-predefined-sorting` to test if it helps in our case, and unfortunately the order was not maintained.

I added a debug print directly before line 1091 of `views.py` (`if data['client_files']:`) and it shows that the order of `server_files` is violated somewhere before this point.

zhiltsov-max: Thank you for reporting, I'll investigate this problem. Could you please attach the request you're sending so I could reproduce it?

OutSorcerer: I now added one more debug print of `request.data` after line 1056 of `views.py` (`elif self.action == 'data':`) and the order of `server_files` was correct at this point, so something changes it between lines 1056 and 1091, e.g. deserialization.

OutSorcerer: >Thank you for reporting, I'll investigate this problem. Could you please attach the request you're sending so I could reproduce it?

Sure, here is the body of a POST request to `api/tasks/{id}/data`:
https://gist.github.com/OutSorcerer/c485498a06a6d3ac44df42de75cdc5fe

It is quite long, but I was not able to reproduce it with small amount of files. I replaced the customer name in file names with "customer-name" string, this is the only difference.

Here is an example of sorting at line 1091 at `data['server_files']` (it displays 2 characters before the dot in a file name, originally files where sorted in reverse order by these values, but CVAT sorts them mostly in ascending):

https://gist.github.com/OutSorcerer/a1351628ef57c5fb7dc9d8262093271a

zhiltsov-max: Ok, thanks. One possible reason is DB saving/loading (in `serializer.save()`) - I've found that for some reason Django sorts `client_files`, and this may be true for other file lists as well.

OutSorcerer: Looks like it is because PostgreSQL [does not maintain insertion order in general](https://dba.stackexchange.com/questions/95822/does-postgres-preserve-insertion-order-of-records). Django has [order-with-respect-to](https://docs.djangoproject.com/en/4.1/ref/models/options/#order-with-respect-to) for cases like this.

cvat/apps/engine/mixins.py
zhiltsov-max: Probably it's an implementation detail/

cvat/apps/engine/mixins.py
zhiltsov-max: Ok, probably you're right.

cvat/apps/engine/views.py
```diff
@@ -851,16 +859,179 @@ def jobs(self, request, pk):
...
+        return "__tus_input_order__"
+
+    def _get_tus_custom_ordering_metafile_name(self) -> str:
+        return "__tus_custom_order__"
+
+    def _get_tus_metafile_names(self) -> List[str]:
+        return [
+            self._get_tus_input_ordering_metafile_name(),
+            self._get_tus_custom_ordering_metafile_name(),
+        ]
```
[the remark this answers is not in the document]
zhiltsov-max: Done

site/content/en/docs/manual/advanced/dataset_manifest.md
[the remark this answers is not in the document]
zhiltsov-max: I decided to remove support for this to simplify testing and usage.

cvat/apps/engine/views.py
zhiltsov-max: I've added code to restore DB insertion order in the DB fetching results. Thank you for your help.

utils/dataset_manifest/core.py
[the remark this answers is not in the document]
zhiltsov-max: Done

utils/dataset_manifest/core.py
zhiltsov-max: The argument was removed to simplify things.

cvat/apps/engine/media_extractors.py
```diff
@@ -365,12 +365,18 @@ def get_zip_filename(self):
...
+            prefix = self.extract_dir or os.path.dirname(self._zip_source.filename)
+            return os.path.join(prefix, self._source_path[i])
         else: # necessary for mime_type definition
             return self._source_path[i]
 
     def reconcile(self, source_files, step=1, start=0, stop=None, dimension=DimensionType.DIM_2D, sorting_method=None):
+        if source_files:
+            # file list is expected to be a processed output of self.get_path()
+            # which returns files with the output directory prefix
+            prefix = self.extract_dir or os.path.dirname(self._zip_source.filename or '')
```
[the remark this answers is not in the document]
zhiltsov-max: Done

cvat/apps/engine/views.py
zhiltsov-max: Fixed. This is a validation for safety, because input filenames are user-provided. Moved to serializer.

cvat/apps/engine/views.py
zhiltsov-max: The same as with dots.

cvat/apps/engine/views.py
zhiltsov-max: Fixed

cvat/apps/engine/views.py
zhiltsov-max: Client files are always ordered, but in some cases the order is different from what is needed. Because of this there are 2 modes of uploading with TUS: with the input order and with the custom order specified explicitly. By default, the input order is used, so there are no extra actions needed from the user to utilize the predefined sorting method. If the user wishes to send files unordered and then sort them on the server side - it will require sending the desired filename list in a special request (the Upload-Start).

cvat/apps/engine/views.py
zhiltsov-max: Done.

cvat/apps/engine/views.py
zhiltsov-max: These are 2 different options nevertheless.

cvat/apps/engine/views.py
zhiltsov-max: Applied

cvat/apps/engine/views.py
zhiltsov-max: Ok, done

zhiltsov-max:
The implemented protocol is in the PR description. Ready for review.

cvat/apps/engine/views.py
```diff
@@ -797,16 +800,179 @@ def perform_create(self, serializer, **kwargs):
...
+        if not upload_metafile:
+            raise FileNotFoundError(
+                "Uploaded file order is not specified. "
+                "Please make sure the Upload-Start request has been sent."
+            )
+
+        expected_files = self._read_tus_upload_meta_file(osp.join(upload_dir, upload_metafile))
+        expected_files.append(upload_metafile)
+
+        uploaded_file_names = { f: f for f in uploaded_files }
```
SpecLad, line 895: Why is this a dict? Seems like it could just as easily be a set.

cvat/apps/engine/serializers.py
```diff
@@ -726,6 +748,13 @@ def validate_job_file_mapping(self, value):
 
         return value
 
+    def validate_tus_file_order(self, value):
```
SpecLad, line 751: This function is not used anywhere.

And I don't think it should be - as long as you verify that the list of filenames matches the actually uploaded filenames, it will follow that there are no suspicious paths in it.

cvat-sdk/cvat_sdk/core/uploading.py
```diff
@@ -335,7 +336,15 @@ def upload_files(
         if pbar is not None:
             pbar.start(total_size, desc="Uploading data")
 
-        self._tus_start_upload(url)
+        upload_info = None
+        if str(kwargs.get("sorting_method")).lower() == "predefined":
+            # Request file ordering, because we reorder files to send more efficiently
+            upload_info = {
+                "tus_file_order": [p.name for p in resources],
```
SpecLad, line 343: Why is `tus_file_order` sent with the first request? It seems like the API would be simpler if it was sent with the last request, where all the other settings are. Do we need the file order before the upload is complete?

SpecLad (changes requested):

cvat/apps/engine/views.py
SpecLad: That's better, but now the question is, why is the field called `tus_file_order` and not just `file_order`?

cvat-sdk/cvat_sdk/core/uploading.py
zhiltsov-max: The idea is that the first request distinguishes between recording the input order and a custom order. In principle, the required order can be sent later.

cvat/apps/engine/views.py
zhiltsov-max: The idea was to relate to the whole "CVAT-TUS" protocol, but it may be confused. Probably, it can be renamed to `upload_file_order`, because `file_order` can also be confusing - it can mean it is always applied, while it's not supposed for this.

tests/python/rest_api/test_tasks.py
```diff
@@ -534,9 +530,7 @@ def test_can_create_task_with_sorting_method(self):
 
         # Besides testing that the sorting method is applied, this also checks for
         # regressions of <https://github.com/opencv/cvat/issues/4962>.
```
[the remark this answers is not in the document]
zhiltsov-max: I moved the comment to the CLI test, because it does just this.

tests/python/sdk/test_tasks.py
```diff
@@ -127,6 +138,27 @@ def test_can_create_task_with_local_data(self):
...
+        self, fxt_new_task_without_data: Task
+    ):
+        task = fxt_new_task_without_data
+
+        task_files = generate_image_files(6)
+        for i, f in enumerate(task_files):
+            fname = self.tmp_path / osp.basename(f.name)
+            with fname.open("wb") as fd:
+                fd.write(f.getvalue())
+                task_files[i] = str(fname)
```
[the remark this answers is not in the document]
zhiltsov-max: Fixed

utils/dataset_manifest/core.py
```diff
@@ -487,24 +493,21 @@ def _write_core_part(self, file, _tqdm):
...
-                self._write_core_part(tmp_file, _tqdm)
-            temp = tmp_file.name
+            tmp_file = StringIO()
+            self._write_core_part(tmp_file, _tqdm)
             with open(self._manifest.path, 'w') as manifest_file:
                 self._write_base_information(manifest_file)
-                with open(temp, 'r') as tmp_file:
-                    manifest_file.write(tmp_file.read())
-            os.remove(temp)
+                manifest_file.write(tmp_file.read())
```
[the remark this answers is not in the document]
zhiltsov-max: Fixed

cvat-sdk/cvat_sdk/core/uploading.py
SpecLad: IMO, it would be better if the order was sent in the last request instead of the first:

* The mental model for the user is simpler: all settings go in the final request. The client code becomes simpler as well. There's no `image_quality` duplication.
* The server code should become simpler as well. There's no need to save the custom file order to disk -> `_TUS_CUSTOM_ORDERING_FILE_NAME` and everything related to it can be removed. `upload_started` can be reduced to just `self._init_tus_input_ordering()`. Checks for whether `_TUS_INPUT_ORDERING_FILE_NAME` exists can be removed. The big comment in `DataSerializer` is no longer relevant.

The only downside I see is that you will end up needlessly saving the input order in the cases where a custom order is supplied, but like... you're already doing that in all cases when the sorting is _not_ predefined, so does it really matter if we do it in those cases as well? It shouldn't be cost-prohibitive to just record the order of input files.

cvat/apps/engine/views.py
```diff
@@ -887,23 +959,36 @@ def upload_finished(self, request):
             task_data = self._object.data
             serializer = DataSerializer(task_data, data=request.data)
             serializer.is_valid(raise_exception=True)
-            data = dict(serializer.validated_data.items())
-            uploaded_files = task_data.get_uploaded_files()
```
SpecLad, line 891: This appears to be the only usage of `get_uploaded_files`, so it can probably be deleted.

cvat/apps/engine/views.py
```diff
@@ -887,23 +959,36 @@ def upload_finished(self, request):
...
-            uploaded_files.extend(data.get('client_files'))
-            serializer.validated_data.update({'client_files': uploaded_files})
```
SpecLad, line 893: I'm confused about how this is supposed to work now. Won't a `client_files` field in the `Upload-Finish` request now overwrite this field in the database, causing all previously-recorded files to be forgotten?

zhiltsov-max: No, the only thing being done in the serializer is that the new rows are added in the DB, so no overwrites happen. Now, we just record everything the user sent on reach request, and optionally sort the data fetched from the DB during the last request processing.

cvat/apps/engine/views.py
zhiltsov-max: Yes, done

zhiltsov-max: Yes, done

PR #5157, opened 2022-10-23
Projects page lazy loading

chepuhasasha:
# Realization of "lazy loading" of project cards.

updated package: `cvat-ui`

The current approach to navigating project card pages loads data every time the page is opened, causing lots of React renders, API queries and increasing time to interactivity state. 

[Preview old behavior.](https://postimg.cc/Ln87FrWb)

The project card script has been redesigned to "infinite scrolling".  Which allows us to reduce the number of requests to the API by at least half, significantly reduce the number of React re-renderers and reduce the time to the state of interactivity. 

[New behavior preview.](https://postimg.cc/NyBjgGwc)

### Checklist
- [ ] ~I submit my changes into the `develop` branch~
- [ ] ~I have added a description of my changes into [CHANGELOG](https://github.com/cvat-ai/cvat/blob/develop/CHANGELOG.md) file~
- [ ] ~I have updated the [documentation](
  https://github.com/cvat-ai/cvat/blob/develop/README.md#documentation) accordingly~
- [ ] ~I have added tests to cover my changes~
- [ ] ~I have linked related issues ([read github docs](
  https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword))~
- [ ] ~I have increased versions of npm packages if it is necessary ([cvat-canvas](https://github.com/cvat-ai/cvat/tree/develop/cvat-canvas#versioning),
[cvat-core](https://github.com/cvat-ai/cvat/tree/develop/cvat-core#versioning), [cvat-data](https://github.com/cvat-ai/cvat/tree/develop/cvat-data#versioning) and [cvat-ui](https://github.com/cvat-ai/cvat/tree/develop/cvat-ui#versioning))~

### License

### DIFF
--- a/cvat-ui/src/actions/projects-actions.ts
+++ b/cvat-ui/src/actions/projects-actions.ts
@@ -17,6 +17,7 @@ const cvat = getCore();
 export enum ProjectsActionTypes {
     UPDATE_PROJECTS_GETTING_QUERY = 'UPDATE_PROJECTS_GETTING_QUERY',
     GET_PROJECTS = 'GET_PROJECTS',
+    GET_PROJECTS_LAZY = 'GET_PROJECTS_LAZY',
     GET_PROJECTS_SUCCESS = 'GET_PROJECTS_SUCCESS',
     GET_PROJECTS_FAILED = 'GET_PROJECTS_FAILED',
     CREATE_PROJECT = 'CREATE_PROJECT',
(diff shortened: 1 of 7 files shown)


bsekachev:
Hi @chepuhasasha 

Thank you for your proposal. After a quick review I have some comments about it:

- If we have less than 4 projects:
image

- If I try to apply a filter o sorting, UI fails:
image

And the reason here not only because something is broken on client, these settings are applied server-side during extracting data from the database, so, they don't work at all now

- After creating a new project we can not see it without updating the whole application

- The same issue, if somebody else created a project from another client

- Similar issue if some projects have been updated from another client, we will not see this update without reloading a page

General comment is that we actually have had infinite scrolling in the past, but we have received a feedback from our customers: it was inconvenient and didn't work well always, so, we decided to use pagination

bsekachev:
I will close the PR because of the problems I mentioned. 
Anyway, thank you.

PR #5159, opened 2022-10-24
Test task QA (MK)

caramba-v1:
Mikhail Krastelev test task


tests/cypress.json
```diff
@@ -4,15 +4,16 @@
...
   "env": {
     "user": "admin",
     "email": "admin@localhost.company",
     "password": "12qwaszx"
   },
   "testFiles": [
     "auth_page.js",
-    "skeletons_pipeline.js",
+    "skeleton/skeletons_pipeline.js",
+    "skeleton/skeletons_import_export.js",
```
bsekachev, line 16:
```suggestion
    "skeleton/*.js",
```

tests/cypress/integration/skeleton/skeletons_import_export.js
```diff
@@ -0,0 +1,109 @@
...
+context('Importing and exporting skeletons', () => {
+    const fileToUpload = 'mounted_file_share/svg/upload.svg';
+    const { downloadsFolder } = require('../../../cypress.json');
+
+    before(() => {
+        cy.task('createFolder', downloadsFolder);
+    });
+
+    beforeEach(() => {
+        cy.loginHeadless();
```
bsekachev, line 16: Not sure we need to do login before each test

tests/cypress/integration/skeleton/skeletons_import_export.js
```diff
@@ -0,0 +1,109 @@
...
+                        clientY: rect.y + height * yOffset,
+                        button: 0,
+                        bubbles: true,
+                    }));
+                });
+            });
+        };
+
+        it('Upload svg file', () => {
+            cy.visit('/tasks/create');
```
bsekachev, line 77: This line probably should be a part of ``beforeEach``

tests/cypress/integration/skeleton/skeletons_import_export.js
```diff
@@ -0,0 +1,109 @@
...
+                const parser = new DOMParser();
+                const xml = parser.parseFromString(contentWithRoot, 'image/svg+xml');
+                /* eslint-disable-next-line */
+                expect(xml.querySelector('parsererror'), 'SVG parsing error').to.be.null;
+                const circles = xml.getElementsByTagName('circle');
+                const lines = xml.getElementsByTagName('line');
+                console.log({
+                    circle: circles.length,
+                    line: lines.length,
+                });
```
bsekachev, line 37: Probably need to remove

tests/cypress/integration/skeleton/skeletons_import_export.js
```diff
@@ -0,0 +1,109 @@
...
+                cy.get('.cvat-skeleton-configurator-svg').within(() => {
+                    cy.get('circle').should('have.length', svgElements.circle);
+                    if (svgElements.line) {
+                        cy.get('line').should('have.length', svgElements.line);
+                    } else {
+                        cy.get('line').should('not.exist');
+                    }
+                });
+            });
+        };
```
bsekachev, line 56: In general would be better to recursively compare documents structure, I believe. Not just number of elements.

tests/cypress/integration/skeleton/skeletons_import_export.js
```diff
@@ -0,0 +1,109 @@
...
+            ];
+            drawCanvasPoints(points);
+
+            cy.task('listFiles', downloadsFolder).then((before) => {
+                cy.get('.cvat-skeleton-configurator-svg-buttons > button').click();
+                cy.task('listFiles', downloadsFolder).then((after) => {
+                    expect(after.length).to.be.eq(before.length + 1);
+                    return (after.filter((f) => !before.includes(f)))[0];
+                }).then((fileName) => {
+                    cy.log(fileName);
```
bsekachev, line 103: Probably should be removed

tests/cypress/plugins/index.js
```diff
@@ -31,6 +32,27 @@ module.exports = (on, config) => {
         },
     });
     on('task', { isFileExist });
+    on('task', {
+        createFolder: (folder) => {
+            if (!fs.existsSync(folder)) {
+                fs.mkdirSync(folder);
+            }
+            return null;
```
bsekachev, line 40: Just curios about why we return ``null`` every time.

tests/cypress/plugins/index.js
```diff
@@ -31,6 +32,27 @@ module.exports = (on, config) => {
...
+        },
+    });
+    on('task', {
+        cleanFolder: (folder) => {
```
bsekachev, line 44: Will we have some dangerous situation if folder contains ``..``?

bsekachev:
@caramba-v1 

Thank you for the contribution.
In general, I like the patch. Only have some minor comments and questions.

tests/cypress/plugins/index.js
caramba-v1: A task must not return undefined, so returning null is quite a common practice as I see in the [docs](https://docs.cypress.io/api/commands/task#Syntax)

tests/cypress/integration/skeleton/skeletons_import_export.js
caramba-v1: Cypress intentionally clears cookies between tests to reset the state. Not sure how it works in other tests yet

PR #5160, opened 2022-10-24
Add tests for export job dataset

Marishka17:
### Motivation and context

Added REST API tests to export job datasets & annotations and validate their structure.
### Checklist

~~- [ ] I have increased versions of npm packages if it is necessary ([cvat-canvas](https://github.com/cvat-ai/cvat/tree/develop/cvat-canvas#versioning),

### License


tests/python/rest_api/test_jobs.py
```diff
@@ -510,3 +515,87 @@ def test_can_export_annotations(self, admin_user: str, jobs_with_shapes: List):
...
+        task_size = tasks[task_id]["size"]
+
+        response = self._export_annotations(username, jid, format="COCO 1.0")
+        assert response.data
+        temp_anno_file_name = None
+        with NamedTemporaryFile(
+            mode="w+b", prefix="cvat", delete=False, suffix=".zip"
+        ) as temp_anno_file:
+            temp_anno_file_name = temp_anno_file.name
+            temp_anno_file.write(response.data)
```
zhiltsov-max, line 540: What is the purpose of actual writing files on disk? Probably, it will be better to write the data to `BytesIO` or directly to `ZipFile`.

tests/python/rest_api/test_jobs.py
```diff
@@ -510,3 +515,87 @@ def test_can_export_annotations(self, admin_user: str, jobs_with_shapes: List):
...
+        # NOTE: data step is not stored in assets, default = 1
+        assert job_data["stop_frame"] - job_data["start_frame"] + 1 == len(
+            exported_annotations["images"]
+        )
+
+        assert task_size > len(exported_annotations["images"])
+        os.remove(temp_anno_file_name)
+
+    @pytest.mark.parametrize("username, jid", [("admin1", 14)])
+    def test_check_exported_cvat_dataset_structure(
```
zhiltsov-max, line 555: Why only these two formats were tested?

cvat/apps/dataset_manager/task.py
```diff
@@ -70,9 +70,18 @@ def _merge_table_rows(rows, keys_for_merge, field_id):
     return list(merged_rows.values())
 
 class JobAnnotation:
-    def __init__(self, pk):
-        self.db_job = models.Job.objects.select_related('segment__task') \
-            .select_for_update().get(id=pk)
+    def __init__(self, pk, is_prefetched=False):
```
zhiltsov-max, line 73: This new argument is unclear to me - when and why it should be used? I don't see how the fact the data is (was?) prefetched affects this class.

Marishka17: It also affects the sorting of images in the annotation file in cvat format. (We can get sorted sequence from db or sort in code, but I've chosen this approach because we use it for tasks)

zhiltsov-max: I don't see how this answers the questions above.


> We can get sorted sequence from db or sort in code, but I've chosen this approach because we use it for tasks

I see that sorting is performed in the newly added code when frames are fetched from DB. The DB itself doesn't impose any sorting.

tests/python/rest_api/test_jobs.py
Marishka17: > Why only these two formats were tested?

I believe that in these tests 2 formats will be enough because tests should check that the annotation file contains only job annotations (the dataset contains only job data) and not the entire task. If this is true for our format, then the same for the others.

Issue #5096, opened 2022-10-12
Create a package for utils/dataset_manifest 


zhiltsov-max:
### My actions before raising this issue
- [ ] Read/searched [the docs](https://github.com/cvat-ai/cvat/tree/master#documentation)
- [ ] Searched [past issues](/issues)

### Expected Behaviour

I suggest to create a separate package for this component, and add it as a dependency. It will also simplify installation for users (by not requiring them to download the full repo. Probably it should also be integrated to CLI as a command.

### Current Behaviour

Currently, this component is a set of scripts. However, it is used in the server implementation, it is supposed to be installed by users, and it is going to be used in cvat-cli.

nmanovic:
@zhiltsov-max , do you think we can include the generation of a manifest file into CLI?

PR #5172, opened 2022-10-26
Fix create manifest command

Marishka17:
### Checklist

### License

### DIFF
--- a/site/content/en/docs/manual/advanced/dataset_manifest.md
+++ b/site/content/en/docs/manual/advanced/dataset_manifest.md
@@ -60,8 +60,7 @@ optional arguments:
 ### Alternative way to use with cvat/server
 
 ```bash
-docker run -it --entrypoint python3 -v /path/to/host/data/:/path/inside/container/:rw cvat/server
-utils/dataset_manifest/create.py --output-dir /path/to/manifest/directory/ /path/to/data/
+docker run -it -u root --entrypoint bash -v /path/to/host/data/:/path/inside/container/:rw cvat/server -c "pip3 install -r utils/dataset_manifest/requirements.txt && python3 utils/dataset_manifest/create.py --output-dir /path/to/manifest/directory/ /path/to/data/"
 ```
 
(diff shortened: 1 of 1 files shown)


nmanovic (approved):

site/content/en/docs/manual/advanced/dataset_manifest.md
```diff
@@ -60,8 +60,7 @@ optional arguments:
 ### Alternative way to use with cvat/server
 
 ```bash
-docker run -it --entrypoint python3 -v /path/to/host/data/:/path/inside/container/:rw cvat/server
-utils/dataset_manifest/create.py --output-dir /path/to/manifest/directory/ /path/to/data/
+docker run -it -u root --entrypoint bash -v /path/to/host/data/:/path/inside/container/:rw cvat/server -c "pip3 install -r utils/dataset_manifest/requirements.txt && python3 utils/dataset_manifest/create.py --output-dir /path/to/manifest/directory/ /path/to/data/"
```
zhiltsov-max, line 63: Do I understand correctly that this solution offers to download and install packages every time?

Marishka17: Yes, and also I know that it is not a good idea to run pip with sudo/docker with root. But it is the simplest way for the user. (I mean that from my point of view it is not comfortable for the user to create some docker file that inherits from our server image and then run the command. Another way is adding tqdm module to server requirements but it doesn't suit us. And maybe the last variant is to install in code). So, how do you suggest improving this one?

zhiltsov-max: My ideas are #5096 or, at least, modifying the server container to have this environment installed, if this module is supposed to be used this way. Otherwise, it's just a hack, and its hard to be called usable. Just imagine a user that tries to find out the correct way to invoke the command - so is is expected to be executed multiple times in a row - and each time he needs to wait for downloading and installation. The user also needs internet connection to run this command.

PR #5147, opened 2022-10-20
CVAT authentication

Marishka17:
### Motivation and context

PR contains changes to support social authentication with google and github providers. Also several bugs with related field were fixed. Resolved issue #4820. There is now possible to request another mail with email confirmation link.
Resolved issue #4832 with attempt to confirm email one more time with the same confirmation link.
I guess this issue #4837 can also be closed because now we can send another confirmation email.
### How has this been tested?

Manually
### Checklist

- [x] I submit my changes into the `develop` branch
- [x] I have added a description of my changes into [CHANGELOG](https://github.com/cvat-ai/cvat/blob/develop/CHANGELOG.md) file
- [ ] I have updated the [documentation](
  https://github.com/cvat-ai/cvat/blob/develop/README.md#documentation) accordingly
- [ ] I have added tests to cover my changes
- [x] I have linked related issues ([read github docs](
  https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword))
- [ ] I have increased versions of npm packages if it is necessary ([cvat-canvas](https://github.com/cvat-ai/cvat/tree/develop/cvat-canvas#versioning),
[cvat-core](https://github.com/cvat-ai/cvat/tree/develop/cvat-core#versioning), [cvat-data](https://github.com/cvat-ai/cvat/tree/develop/cvat-data#versioning) and [cvat-ui](https://github.com/cvat-ai/cvat/tree/develop/cvat-ui#versioning))

### License

- [x] I submit _my code changes_ under the same [MIT License](
  https://github.com/cvat-ai/cvat/blob/develop/LICENSE) that covers the project.
  Feel free to contact the maintainers if that's a concern.

PR #5181, opened 2022-10-28
New authentication UI

### LINKED ISSUE #5166: UI Bug: If email the first part contains ``.`` during registration, it causes the issue

### Expected Behaviour
Need to think about

### Current Behaviour
- Automatically assigneed username also contains ``.``
- If we try to edit it, it leads to impossibility to submit the format
- 
### Possible Solution
Need to think about

test

### Your Environment

- Git hash commit (`git log -1`):
- Docker version `docker version` (e.g. Docker 17.0.05):
- Are you using Docker Swarm or Kubernetes?
- Operating System and version (e.g. Linux, Windows, MacOS):
- Code example or link to GitHub repo or gist to reproduce problem:
- Other diagnostic information / logs:
    <details>
    <summary>Logs from `cvat` container</summary>
    </details>

klakhov:
Dependent on #5147
Resolved #5166 

### Checklist

### License


bsekachev:
Changing form size looks very strange.. Cursor is not over the field anymore if I enter something 
test

bsekachev:
The issue I submitted before for the previous form still persists in a new one

test 1

bsekachev:
Should blue highlighting look this way? 
image

bsekachev:
- Pressing "Enter" in password field, I expect it make login request Now it does not do anything. 

- Reset password form has not good looking highlighting:
image

Again, pressing Enter I expect the form to be submitted. BTW, the same about registration form when focus in the latest field.

bsekachev:
When window width < 1200px, the form becomes following (takes all the width):

image

I believe that 1200px is huge value and we should minimize form maybe when windows width less then 800px.. But it is discussable.

cvat-core/src/api-implementation.ts
```diff
@@ -63,17 +63,15 @@ const config = require('./config');
             firstName,
             lastName,
             email,
-            password1,
-            password2,
+            password,
```
bsekachev, line 66: Need to update cvat-core major version..

cvat-ui/src/base.scss
```diff
@@ -13,6 +13,8 @@ $layout-lg-grid-color: rgba(0, 0, 0, 0.15);
 
 $header-color: #d8d8d8;
 $text-color: #303030;
+$text-color-2: white;
```
bsekachev, line 16: Need to update header

cvat-ui/src/components/signing-common/styles.scss
```diff
@@ -0,0 +1,296 @@
...
+                    border: none;
+                }
+
+                &::after {
+                    position: absolute;
+                    content: "";
+                    height: $grid-unit-size*1.5;
+                    width: $grid-unit-size*1.5;
+                    border-radius: $grid-unit-size;
+                    top: $grid-unit-size*0.4;
```
bsekachev, line 238: We should use ``grid-unit-size`` only with integer values. It was the core idea of introduction it.

cvat-ui/src/components/email-confirmation-pages/email-verification-sent.tsx
```diff
@@ -0,0 +1,32 @@
...
+ * Component for displaying message that email should be verified
+ */
+
+export default function EmailVerificationSentPage(): JSX.Element {
+    return (
+        <Layout>
+            <Content>
+                <Row justify='center' align='middle' id='email-verification-sent-page-container'>
+                    <Col>
+                        <h1>Please, confirm your email</h1>
```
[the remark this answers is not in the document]
Marishka17: @bsekachev, Now it is implemented so: when a user with an unverified email tries to log in to CVAT (and gets to this page), a new confirmation email will be sent to him.

cvat-ui/src/base.scss
klakhov: I belive its ok. When we gonna implement new ui for user homepage and something like that, probably we will use such declarations

cvat-ui/src/components/signing-common/styles.scss
klakhov: In this case we have `toggle` which is quite small to `$grid-unit-size` so its hard to use just integers. I tried to make it a bit bigger so there is less floats here

kirill-sizov:
@klakhov there is already not necessary to run full check workflow, all our tests are running in main workflow (as it was before full check)

cvat-ui/src/base.scss
bsekachev: I would suggest do not try to guess what we will do in the future..

cvat-core/src/api-implementation.ts
bsekachev: Still not updated

bsekachev:
Input field still tries to escape from the initial position when I entering something, it does not look well IMHO.. 

example

Please, look also how "Next" button goes out of the form and then returns back

bsekachev:
Why can't we have a username with dots? They are allowed in django https://docs.djangoproject.com/en/4.1/ref/contrib/auth/#django.contrib.auth.models.User.username

klakhov:
@bsekachev I belive It's better now (adjusted size a bit).
jumpingform2

dotinusename

cvat-ui/src/base.scss
klakhov: Ok. I've moved those declarations in corresponding file.

cvat-ui/src/components/signing-common/styles.scss
klakhov: What I mean here that `$grid-unit-size: 8px` is quite large for such thing like toogle. So in that case using something like `$grid-unit-size * 0.5` should be fine

cvat-ui/src/components/signing-common/styles.scss
```diff
@@ -129,7 +142,7 @@
     border-radius: $grid-unit-size * 4;
     background: $background-color-1;
     padding: $grid-unit-size * 6;
-    height: 710px;
+    height: 648px;
```
bsekachev, line 145: $grid-unit-size * 81

cvat-ui/src/components/signing-common/styles.scss
```diff
@@ -84,7 +97,7 @@
     .ant-typography {
         text-transform: uppercase;
         color: $border-color-1;
-        font-size: $grid-unit-size * 1.25;
+        font-size: $grid-unit-size * 1.5;
```
bsekachev, line 100: for fonts ``grid-unit-size`` is not necessary

cvat-ui/src/components/signing-common/styles.scss
klakhov: done

PR #5189, opened 2022-10-28
Fix ESLint errors in cvat-core/src/server-proxy.ts

SpecLad:
Specifically:

* `padded-blocks`
* `@typescript-eslint/no-var-requires`
* `@typescript-eslint/return-await`

### Motivation and context

As far as I know, "go to definition" in VS Code only works with ES6-style imports. Since I was examining code in this file, I updated the imports to make it easier. I fixed the other ESLint errors, because otherwise it wouldn't let me commit the code. 🙂 

### How has this been tested?

I'm relying on CI for this.

### Checklist

- [x] I have increased versions of npm packages if it is necessary ([cvat-canvas](https://github.com/cvat-ai/cvat/tree/develop/cvat-canvas#versioning),

### License

### DIFF
--- a/cvat-core/package.json
+++ b/cvat-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "cvat-core",
-  "version": "7.0.1",
+  "version": "7.0.2",
   "description": "Part of Computer Vision Tool which presents an interface for client-side integration",
   "main": "src/api.ts",
   "scripts": {
--- a/cvat-core/src/server-proxy.ts
+++ b/cvat-core/src/server-proxy.ts
@@ -3,9 +3,16 @@
 //
 // SPDX-License-Identifier: MIT
 
-import { isEmail } from './common';
-import { StorageLocation, WebhookSourceType } from './enums';
+import FormData from 'form-data';
+import store from 'store';
+import Axios from 'axios';
+import * as tus from 'tus-js-client';
 import { Storage } from './storage';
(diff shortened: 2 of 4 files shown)


cvat-core/src/server-proxy.ts
```diff
@@ -742,7 +755,7 @@ class ServerProxy {
                 }
             }
             try {
-                return wait();
+                return await wait();
```
bsekachev, line 758: We do not need ``await`` when return result of an async function from an async function..

cvat-core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "cvat-core",
-  "version": "7.0.1",
+  "version": "7.0.2",
```
bsekachev, line 3: If it is only refactoring, we do not need to update version number

cvat-core/src/server-proxy.ts
SpecLad: In this case we do need `await`. Without it, the function will immediately return the promise it gets from `wait`, and the `catch` block will never be entered.

cvat-core/package.json
SpecLad: The `return await` change is technically a behavior change.

SpecLad:
I had to fix a bug in two other files for the tests to pass, and _those_ files fail ESLint. I'm not going to fix them (at least for now), so the ESLint build job will be red.

PR #5312, opened 2022-11-18
PostgresSQL DB v15 and health check endpoint

azhavoro:
### Checklist

### License


nmanovic:
@azhavoro , I think a little bit and I propose another solution. Could you please think and give me your opinion?

Let's have a health check endpoint for CVAT. Thus we can use it in our UI and show that the server is not ready yet. Of course, we don't know the reason why, but we can show possible reasons and recommendations, a link to our documentation. I believe it is a step in the right direction.

I will recommend to integrate https://github.com/revsys/django-health-check app. Also we can write small custom plugins and detect some popular problems. For example, OPA isn't running. The problem already can be shown in UI.

kirill-sizov:
Probably we can try to speed up our tests by simplifying [`restore.sql` script](https://github.com/opencv/cvat/blob/8e6bbffb1aed41c37261515134345644832d28ea/tests/python/shared/assets/cvat_db/restore.sql):

I would suggest:
```sql
DROP DATABASE IF EXISTS :to WITH FORCE;
```

instead of current:
```sql
SELECT pg_terminate_backend(pg_stat_activity.pid)
FROM pg_stat_activity
WHERE pg_stat_activity.datname = 'cvat' AND pid <> pg_backend_pid();

DROP DATABASE IF EXISTS :to;
```

cvat-ui/src/components/cvat-app.tsx
```diff
@@ -121,7 +127,7 @@ class CVATApplication extends React.PureComponent<CVATAppProps & RouteComponentP
             customWaViewHit(_location.pathname, _location.search, _location.hash);
         });
 
-        verifyAuthorized();
+        loadBackendHealth();
```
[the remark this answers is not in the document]
azhavoro: How would it be better to set the progress state from the cvat-core function? Callback?

site/content/en/docs/administration/advanced/upgrade_guide.md
```diff
@@ -9,6 +9,9 @@ description: 'Instructions for upgrading CVAT deployed with docker compose'
 
 ## Upgrade guide
 
+Note: updating CVAT from version 2.3.0 to version 2.4.0 requires additional manual actions with database data due to
```
nmanovic, line 12: from version 2.2.0 to version 2.3.0

nmanovic:
@azhavoro , healthcheck django package provides a command health_check. It can be used to implement https://docs.docker.com/compose/compose-file/compose-file-v3/#healthcheck. Thus we will have a status for cvat_server.

site/content/en/docs/administration/advanced/upgrade_guide.md
```diff
@@ -69,3 +72,43 @@ docker pull cvat/ui:v2.1.0
...
+   [installation guide](/docs/administration/basics/installation/#how-to-get-cvat-source-code) for details.
+
+1. Start database container only:
+   ```shell
+   docker-compose up -d cvat_db
+   ```
+
+1. Import PostgreSQL dump into new DB container:
+   ```shell
+   docker exec -i cvat_db psql -d postgres < cvat.db.dump
```
nmanovic, line 108: Probably need to run the command in quiet mode.

nmanovic:
Need to document the command as well.

image

nmanovic:
Need to improve the initial page

image

nmanovic:
No need to write the header one more time in the body. Also at the end there is no a space between `the` and `Upgrade Guide`. 

image

docker-compose.yml
```diff
@@ -155,13 +155,14 @@ services:
     labels:
       - traefik.enable=true
       - traefik.http.services.cvat-ui.loadbalancer.server.port=80
-      - traefik.http.routers.cvat-ui.rule=Host(`${CVAT_HOST:-localhost}`)
+      - traefik.http.routers.cvat-ui.rule=Host(`${CVAT_HOST:-localhost}`) &&
+        ! PathPrefix(`/api/`, `/git/`, `/opencv/`, `/static/`, `/admin`, `/documentation/`, `/django-rq`)
```
nmanovic, line 159: @azhavoro , why do we need the line now? How is it connected with the patch?

azhavoro: @nmanovic I think it's better to remove the `HEALTHCHECK` command from the Dockerfile for this PR, this currently leads to routing issues related to Traefik container health check behavior https://github.com/traefik/traefik/issues/7732. At the moment I don't find a suitable workaround, so I propose to create a separate issue about adding `HEALTHCHECK` instruction to CVAT server docker image.

nmanovic: OK. Let's do that in this case.

nmanovic (approved):
LGTM

PR #5356, opened 2022-11-24
SDK: add a high-level method to download task data chunks

SpecLad:
### Motivation and context

I will need it for the PyTorch data adapter.

### Checklist

### License

### DIFF
--- a/cvat-sdk/cvat_sdk/core/proxies/tasks.py
+++ b/cvat-sdk/cvat_sdk/core/proxies/tasks.py
@@ -9,9 +9,10 @@ import json
 import mimetypes
 import os
 import os.path as osp
+import shutil
 from enum import Enum
 from time import sleep
-from typing import Any, Dict, List, Optional, Sequence
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Sequence
 
 from PIL import Image
 
@@ -31,6 +32,9 @@ from cvat_sdk.core.proxies.model_proxy import (
 )
 from cvat_sdk.core.uploading import AnnotationUploader, DataUploader, Uploader
 from cvat_sdk.core.utils import filter_dict
+
+if TYPE_CHECKING:
+    from _typeshed import SupportsWrite
 
 
 class ResourceType(Enum):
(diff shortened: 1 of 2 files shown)


cvat-sdk/cvat_sdk/core/proxies/tasks.py
```diff
@@ -153,6 +157,23 @@ def get_preview(
...
+    ) -> None:
+        params = {}
+        if quality:
+            params["quality"] = quality
+        (_, response) = self.api.retrieve_data(
+            self.id, number=chunk_id, **params, type="chunk", _parse_response=False
+        )
+
+        with response:
+            shutil.copyfileobj(response, output_file)
```
zhiltsov-max, line 175: 1. Is it better than simply returning a file object (e.g. like in the other download* functions), bytes, or the response itself?
2. The fact that the response is a zip file could be documented, the contents may change depending on the task type. As a starting solution its ok, but providing a high-level chunk reader in future could be of great help for users.

SpecLad: > Is it better than simply returning a file object (e.g. like in the other download* functions), bytes, or the response itself?

I considered other options, but I found them unsatisfactory:

1. `bytes`: chunks can be large (I think arbitrarily large, given that chunk size is a settable parameter), so I would like to avoid loading them into memory all at once.
2. `BytesIO`: that's just `bytes` with extra steps. A user can also easily obtain the data in a `BytesIO` object with the current approach by passing that object as the output file argument.
3. Returning the response itself: this would be workable, but annotating the return type would be tricky. I don't want to straight up declare that the function returns an `urllib3.HTTPResponse` (that would be exposing implementation details), and AFAIK there's no type annotation in the standard library or typeshed that would represent "a context manager with a `read()` method". I _could_ define a custom protocol, but that seems like too much complexity for unclear benefit. Compared to the current approach, this would let the user process partially downloaded chunks, but at least in the case of ZIP chunks, that's not very useful, because you need a seekable stream to read a ZIP file.

SpecLad: > The fact that the response is a zip file could be documented, the contents may change depending on the task type.

Can't chunks also be video files in some cases?

> As a starting solution it's ok, but providing a high-level chunk reader in future could be of great help for users.

I agree, but that'll probably be part of the dataset API.

zhiltsov-max: > Can't chunks also be video files in some cases?

It seems they can, but it only increases the need in describing the function.

SpecLad: I can describe the _function_, but I don't think defining what a chunk is should be a part of that description. This is a general CVAT concept, and it should be explained in the CVAT manual (e.g. in the glossary).

zhiltsov-max (approved):

nmanovic (approved):

PR #5225, opened 2022-11-03
Better reporting for user limits

zhiltsov-max:
### Motivation and context

- Added explanatory messages for actions denied for user limits
- Fixed few rules and checks
- Upgraded OPA version

### How has this been tested?

Unit tests

### Checklist

- [ ] I submit my changes into the `develop` branch
- [ ] I have added a description of my changes into [CHANGELOG](https://github.com/cvat-ai/cvat/blob/develop/CHANGELOG.md) file
- [ ] I have updated the [documentation](
  https://github.com/cvat-ai/cvat/blob/develop/README.md#documentation) accordingly
- [ ] I have added tests to cover my changes
- [ ] I have linked related issues ([read github docs](
  https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword))
- [ ] I have increased versions of npm packages if it is necessary ([cvat-canvas](https://github.com/cvat-ai/cvat/tree/develop/cvat-canvas#versioning),
[cvat-core](https://github.com/cvat-ai/cvat/tree/develop/cvat-core#versioning), [cvat-data](https://github.com/cvat-ai/cvat/tree/develop/cvat-data#versioning) and [cvat-ui](https://github.com/cvat-ai/cvat/tree/develop/cvat-ui#versioning))

### License

- [ ] I submit _my code changes_ under the same [MIT License](
  https://github.com/cvat-ai/cvat/blob/develop/LICENSE) that covers the project.
  Feel free to contact the maintainers if that's a concern.

PR #5364, opened 2022-11-28
[WIP] [Dependent] Extend limit manager

kirill-sizov:
### Motivation and context
Extend limit manager that was introduced in #5225 

### Checklist

### License


cvat/apps/limit_manager/models.py
```diff
@@ -0,0 +1,42 @@
...
+from django.contrib.auth.models import User
+from cvat.apps.organizations.models import Organization
+
+
+class ConsumableCapability(models.TextChoices):
+    TASK_CREATE = auto()
+    PROJECT_CREATE = auto()
+    ORG_CREATE = auto()
+    CLOUD_STORAGE_CREATE = auto()
+    WEBHOOK_CREATE = auto()
```
zhiltsov-max, line 17: Seems to be unused

cvat/apps/limit_manager/models.py
```diff
@@ -0,0 +1,42 @@
...
+# TO-DO:
+#  - add validation user__isnull ^ org__isnull
+class Limitation(models.Model):
+    user = models.ForeignKey(User, on_delete=models.CASCADE, null=True)
+    org = models.ForeignKey(Organization, on_delete=models.CASCADE, null=True)
+
+    # List of limitations:
+    #   - null means unlimited
+    #   - blank means that certain limit not applicable
+    memberships = models.PositiveIntegerField(default=3, null=True, blank=True)
```
zhiltsov-max, line 29: Please declare/update and use the default values in the `settings/base.py` and in `settings/testing.py`.

cvat/apps/iam/permissions.py
```diff
@@ -21,10 +21,12 @@
...
 from cvat.apps.engine.models import Project, Task, Job, Issue
 from cvat.apps.limit_manager.core.limits import (CapabilityContext, LimitManager,
-    Limits, OrgCloudStoragesContext, OrgTasksContext, ProjectWebhooksContext,
+    Limits, OrgCloudStoragesContext, OrgMembersContext, OrgTasksContext, OrgProjectWebhooksContext,
     OrgCommonWebhooksContext,
     TasksInOrgProjectContext, TasksInUserSandboxProjectContext, UserOrgsContext,
-    UserSandboxCloudStoragesContext, UserSandboxTasksContext)
+    UserSandboxCloudStoragesContext, UserSandboxProjectWebhooksContext,
+    UserSandboxTasksContext,
+    OrgLambdaCallOfflineContext, UserSandboxLambdaCallOfflineContext)
```
zhiltsov-max, line 29: Could you please change the import to `from cvat.apps.limit_manager.core import limits` so it was more clear? The list will only grow in future.

cvat/apps/limit_manager/models.py
kirill-sizov: Done

bsekachev (approved):

nmanovic:
@klakhov , could you please resolve the conflicts?

cvat-ui/src/components/header/header.tsx
```diff
@@ -33,7 +33,7 @@ import Select from 'antd/lib/select';
 import { getCore } from 'cvat-core-wrapper';
 import config from 'config';
 
-import { CVATLogo } from 'icons';
+import { CVATLogo, UpgradeIcon } from 'icons';
```
nmanovic, line 36: I believe `Upgrade` work and icon looks unclear. Probably `Switch to Pro plan` with `$` icon should look better. Do you agree? Probably need to talk with Breeze.

cvat/settings/base.py
```diff
@@ -264,15 +265,40 @@ def add_ssh_keys():
...
-    "ORG_CLOUD_STORAGES": 10,
-    "ORG_COMMON_WEBHOOKS": 20,
+    "ORG_CLOUD_STORAGES": 1,
+    "ORG_COMMON_WEBHOOKS": 10,
+    "ORG_PROJECT_WEBHOOKS": 10,
+    "ORG_LAMBDA_CALL_OFFLINE": 0,
+}
+
+
+ORG_LIMITS_MAPPING = {
```
nmanovic, line 283: @kirill-sizov , could you please explain why we need the mapping in settings?

tests/python/rest_api/test_cloud_storages.py
```diff
@@ -125,6 +125,13 @@ class TestPostCloudStorage:
         }
     ]
 
+    @pytest.fixture(autouse=True)
```
nmanovic, line 128: Please add a comment here. Need to explain why we decided to add the fixture.

cvat/settings/base.py
kirill-sizov: We have limit names that used by iam, and we also have limit names that we display to user.
And so since we have two different names for each limit we need to define mapping between them.

cvat/apps/engine/views.py
```diff
@@ -2003,6 +2005,16 @@ def self(self, request):
         serializer = serializer_class(request.user, context={ "request": request })
         return Response(serializer.data)
 
+    @action(detail=True, methods=['GET', 'PATCH'])
```
nmanovic, line 2008: First of all, it is the wrong location for the method. I will recommend moving the code into limit_manager. Also the 'limits' endpoind is better than 'limitations' from my perspective.

nmanovic: /api/limits?user_id=id
/api/limits?org_id=id
/api/limits?project_id=id

cvat/apps/iam/rules/limitations.rego
```diff
@@ -0,0 +1,34 @@
+package limitations
```
nmanovic, line 1: Let's rename the file into limits.rego

nmanovic: OK. I see that we already have limits. It is an indication, that something wrong.

cvat/apps/limit_manager/migrations/0005_limitation_type.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 3.2.16 on 2022-12-28 11:22
```
nmanovic, line 1: Can we have 1 migration instead of 5?

nmanovic: Need to experiment with upgrading from the previous revision after that.

cvat/apps/limit_manager/models.py
```diff
@@ -2,10 +2,43 @@
...
+from enum import Enum
 from django.db import models
+from django.contrib.auth.models import User
+from cvat.apps.organizations.models import Organization
 
-import cvat.apps.limit_manager.core.limits as core
+class LimitationTypeChoice(str, Enum):
+    DEFAULT = 'default'
+    PAID = 'paid'
+    MANUALLY = 'manually'
```
nmanovic, line 13: CUSTOM

cvat/apps/limit_manager/models.py
```diff
@@ -2,10 +2,43 @@
...
 
-class Limits(core.Limits, models.TextChoices):
-    pass
+    def __str__(self):
+        return self.value
+
+# TO-DO:
+#  - add validation user__isnull ^ org__isnull
+class Limitation(models.Model):
+    user = models.ForeignKey(User, on_delete=models.CASCADE, null=True)
```
nmanovic, line 25: Can one user or org have multiple limits?

cvat/apps/limit_manager/models.py
```diff
@@ -2,10 +2,43 @@
...
+    org = models.ForeignKey(Organization, on_delete=models.CASCADE, null=True)
+    type = models.CharField(max_length=32, choices=LimitationTypeChoice.choices(),
+        default=LimitationTypeChoice.DEFAULT)
+
+    # List of limitations:
+    #   - null means unlimited
+    #   - blank means that certain limit not applicable
+    memberships = models.PositiveIntegerField(null=True, blank=True)
```
nmanovic, line 33: I'm not sure but probably one [JsonField](https://docs.djangoproject.com/en/4.1/ref/models/fields/#django.db.models.JSONField) can be a better choice to keep all limits below.

cvat/apps/limit_manager/serializers.py
```diff
@@ -0,0 +1,60 @@
...
+from cvat.apps.organizations.serializers import OrganizationReadSerializer
+from cvat.apps.organizations.models import Organization
+
+from .models import Limitation
+
+org_limitation_fields = tuple(settings.ORG_LIMITS_MAPPING.values())
+user_limitation_fields = tuple(settings.USER_LIMITS_MAPPING.values())
+
+
+class UserLimitationReadSerializer(ModelSerializer):
```
nmanovic, line 15: I believe we should have 1 serialize and 2 tables in DB. One for users and another one for organizations.

cvat/apps/engine/views.py
zhiltsov-max: I agree with both points.

zhiltsov-max: Actually, I don't see big reasons in this specific endpoint now, because it looks too narrow for our current usecases. Probably, there are no cases where we would like to get or edit 1 single limitation. Getting or editing just a single limitation is ineffective, because it would require a separate request for each operation. My suggestion is to consider 1 user or 1 org limits as just a single structure. A user or an org is not going to have an arbitrary number of limitations, which populated from somewhere else (unlike projects or users, for example).

zhiltsov-max: Actually, I don't see big reasons in this specific endpoint now, because it looks too narrow for our current usecases. Probably, there are no cases where we would like to get or edit 1 single limitation. Getting or editing just a single limitation is ineffective, because it would require a separate request for each operation. My suggestion is to consider 1 user or 1 org limits as just a single structure. A user or an org are not going to have an arbitrary number of limitations, which populated from somewhere else (unlike projects or users, for example). An example of such could be an option to control org limits on a user basis, which is not going to be supported in the near future.

zhiltsov-max: One thing is that all of these "list" requests will return at most 1 structure. Thus, it can be better to add:

`/users/limits` (GET, PATCH)
`/orgs/limits` (GET, PATCH)

or

`/limits/user/{id}` (GET, PATCH)
`/limits/org/{id}` (GET, PATCH)

zhiltsov-max:
Do we still need this PR?

kirill-sizov:
@zhiltsov-max We will close it when we move our deployment to the develop branch

PR #5416, opened 2022-12-02
Fix REST API tests

Marishka17:
### Motivation and context

Fixed REST API tests after merging #5408 and #5396 to develop


tests/python/shared/fixtures/data.py
```diff
@@ -357,8 +357,17 @@ def find(jobs):
...
-        return list(filter(lambda t: annotations["task"][str(t["id"])]["shapes"], tasks))
+    def find(tasks, skip_tasks=None):
+        def _is_task_with_shapes(t):
+            if (
+                skip_tasks is not None
+                and t["id"] in skip_tasks
+                or not annotations["task"][str(t["id"])]["shapes"]
+            ):
+                return False
+            return True
```
kirill-sizov, line 368: Can we use such solution?
```suggestion
    def find(tasks):
        def _is_task_with_shapes(t):
            if (
                not t.get("data")
                or not annotations["task"][str(t["id"])]["shapes"]
            ):
                return False
            return True
```

Marishka17: ok, thanks

kirill-sizov:
@bsekachev Are these e2e tests unstable? Or are they related with this PR?

bsekachev:
@kirill-sizov 

case_82_canvas3d_functionality_cuboid_opacity_outlined_borders.js

This one was fixed: https://github.com/opencv/cvat/pull/5409#issuecomment-1334340257

case_118_multi_tasks.js 

Was quite stable before, not sure why it is failed here.

kirill-sizov (approved):
LGTM

nmanovic (approved):

PR #5417, opened 2022-12-02
SDK: Add an adapter layer that presents a CVAT task as a torchvision dataset

SpecLad:
### How has this been tested?

Unit tests; custom demo script.

### Checklist

### License

### DIFF
--- a/cvat-sdk/cvat_sdk/pytorch/__init__.py
+++ b/cvat-sdk/cvat_sdk/pytorch/__init__.py
@@ -252,7 +252,6 @@ _SUPPORTED_SHAPE_TYPES = frozenset(["rectangle", "polygon", "polyline", "points"
 @attrs.frozen
 class ExtractBoundingBoxes:
     include_shape_types: FrozenSet[str] = attrs.field(
-        default=_SUPPORTED_SHAPE_TYPES,
         converter=frozenset,
         validator=attrs.validators.deep_iterable(attrs.validators.in_(_SUPPORTED_SHAPE_TYPES)),
         kw_only=True,
@@ -265,9 +264,6 @@ class ExtractBoundingBoxes:
         for shape in target.annotations.shapes:
             if shape.type.value not in self.include_shape_types:
                 continue
-
-            if shape.type.value not in _SUPPORTED_SHAPE_TYPES:
-                raise UnsupportedDatasetError(f"Unsupported shape type {shape.type.value!r}")
 
             if shape.rotation != 0:
                 raise UnsupportedDatasetError("Rotated shapes are not supported")
--- a/cvat-sdk/cvat_sdk/pytorch/__init__.py
+++ b/cvat-sdk/cvat_sdk/pytorch/__init__.py
@@ -45,17 +45,52 @@ class UnsupportedDatasetError(cvat_sdk.core.exceptions.CvatSdkException):
 
 @attrs.frozen
 class FrameAnnotations:
+    """
+    Contains annotations that pertain to a single frame.
+    """
+
     tags: List[LabeledImage] = attrs.Factory(list)
     shapes: List[LabeledShape] = attrs.Factory(list)
 
 
 @attrs.frozen
 class Target:
+    """
+    Non-image data for a dataset sample.
+    """
+
     annotations: FrameAnnotations
+    """Annotations for the frame corresponding to the sample."""
+
     label_id_to_index: Mapping[int, int]
+    """
+    A mapping from label_id values in `LabeledImage` and `LabeledShape` objects
+    to an index in the range [0, num_labels), where num_labels is the number of labels
+    defined in the task. This mapping is consistent across all samples for a given task.
+    """
 
 
 class TaskVisionDataset(torchvision.datasets.VisionDataset):
+    """
+    Represents a task on a CVAT server as a PyTorch Dataset.
+
+    This dataset contains one sample for each frame in the task, in the same
+    order as the frames are in the task. Deleted frames are omitted.
+    Before transforms are applied, each sample is a tuple of
+    (image, target), where:
+
+    * image is a `PIL.Image.Image` object for the corresponding frame.
+    * target is a `Target` object containing annotations for the frame.
+
+    This class caches all data and annotations for the task on the local file system
+    during construction. If the task is updated on the server, the cache is updated.
+
+    Limitations:
+
+    * Only tasks with image (not video) data are supported at the moment.
+    * Track annotations are currently not accessible.
+    """
+
     def __init__(
         self,
         client: cvat_sdk.core.Client,
@@ -65,6 +100,15 @@ class TaskVisionDataset(torchvision.datasets.VisionDataset):
         transform: Optional[Callable] = None,
         target_transform: Optional[Callable] = None,
     ) -> None:
+        """
+        Creates a dataset corresponding to the task with ID `task_id` on the
+        server that `client` is connected to.
+
+        `transforms`, `transform` and `target_transforms` are optional transformation
+        functions; see the documentation for `torchvision.datasets.VisionDataset` for
+        more information.
+        """
+
         self._logger = client.logger
 
         self._logger.info(f"Fetching task {task_id}...")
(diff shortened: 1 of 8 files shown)


cvat-sdk/cvat_sdk/pytorch/__init__.py
```diff
@@ -0,0 +1,280 @@
...
+    shapes: List[LabeledShape] = attrs.Factory(list)
+
+
+@attrs.frozen
+class Target:
+    annotations: FrameAnnotations
+    label_id_to_index: Mapping[int, int]
+
+
+class TaskDataset(torchvision.datasets.VisionDataset):
```
nmanovic, line 58: If we don't inherit the class, which problems we will have?

nmanovic: OK. Probably extras_require is the right approach here.

nmanovic: But at the same time, if we want to test the feature, we need to install torch and torchvision explicitly, but indeed they don't need us. What do think?

cvat-sdk/cvat_sdk/pytorch/__init__.py
```diff
@@ -0,0 +1,280 @@
...
+
+import cvat_sdk.core
+import cvat_sdk.core.exceptions
+from cvat_sdk.api_client.model_utils import to_json
+from cvat_sdk.core.utils import atomic_writer
+from cvat_sdk.models import DataMetaRead, LabeledData, LabeledImage, LabeledShape, TaskRead
+
+_ModelType = TypeVar("_ModelType")
+
+_CACHE_DIR = Path(appdirs.user_cache_dir("CVAT_SDK", "CVAT.ai"))
```
nmanovic, line 38:
```suggestion
_CACHE_DIR = Path(appdirs.user_cache_dir("cvat-sdk", "CVAT.ai"))
```

cvat-sdk/cvat_sdk/pytorch/__init__.py
nmanovic, line 58: Where can I find ProjectDataset?

SpecLad: There is none at the moment; however, one could be built easily using [`ConcatDataset`](https://pytorch.org/docs/stable/data.html#torch.utils.data.ConcatDataset).

Frankly, I'm not sure if it would even make sense to create a `Dataset` out of the entire project - such a dataset would contain all of the subsets, so it wouldn't be suitable for either training or testing.

SpecLad: > If we don't inherit the class, which problems we will have?

That depends on what the alternative is. The PyTorch [`DataLoader`](https://pytorch.org/docs/stable/data.html#torch.utils.data.DataLoader) class requires the dataset to be an instance of `torch.utils.data.Dataset`, so we would at least have to inherit from that.

We could inherit directly from `Dataset`, rather than `VisionDataset` (in fact, for a long time, that _was_ what my prototype did). The `VisionDataset` class provides dataset transform support that helps the class behave more like other torchvision datasets; however we could, in principle, reimplement that ourselves. I eventually decided to use `VisionDataset` because IMO, if you're working with vision datasets in PyTorch, you'll want to use torchvision anyway for its [image processing transforms](http://pytorch.org/vision/stable/transforms.html).

We could implement multiple classes (e.g. `TaskDataset` inheriting from `Dataset` and `TaskVisionDataset` inheriting from `VisionDataset`) and multiple sets of extras (with torchvision and without), but that just seems like overcomplicating things.

> But at the same time, if we want to test the feature, we need to install torch and torchvision explicitly, but indeed we don't need them. What do think?

I don't understand the question. Naturally we would have to install PyTorch and torchvision to test integration with them.

cvat-sdk/cvat_sdk/pytorch/__init__.py
SpecLad: Applied.

SpecLad:
Some notes I made while developing this:

There's an issue with CVAT's data model that makes it somewhat difficult to translate CVAT tasks to datasets. In a dataset, a label is usually represented by a numeric index, but there's no corresponding concept in CVAT. CVAT labels have IDs, which are numeric, but they are not controllable by the user, and different projects will have different label IDs, even if their sets of labels are the same. Therefore, I don't think CVAT label IDs are suitable for use as indexes when working with neural networks.

In this PR, I assign indexes by enumerating all labels in order of ascending ID, which is at least somewhat controllable by the user (you can enter the labels in the correct order), but still not amazing. There's no way to determine from the CVAT UI which labels will have which indexes, and no way to affect the indexes once they've been created.

IMO, for better interoperability with deep learning frameworks, CVAT should assign each label a numeric index, and let the user customize it. Then we can use those assignments in the PyTorch adapter (and other adapters, should we make them).

Another way to help the user in this regard would be to let them specify a custom label-id-to-index mapping when instantiating the adapter class.

-----

The way I extract the Nth frame in a chunk is by taking the Nth entry in the ZIP file's info list, which is the same method that the UI uses. It's not great, though; we're relying on the zip producer library that the server uses to record the files in the catalog in the order that they were specified, which doesn't seem especially safe to me. I don't think the order of files in the catalog is meant to be part of ZIP's data model.

IMO, it would be better to record some sort of manifest in each chunk file that would give the member name for each frame index. Alternatively, this manifest could be part of the data metadata (i.e. the result of `GET /api/tasks/<id>/data/meta`).

-----

I haven't yet implemented some features that could be useful in the adapter layer. Some of them are already documented in the docstrings (as "limitations"). Some others are:

* Accessing images from an external directory (rather than downloading chunks). Could be useful if the task uses files from a mounted share, and the user has also mounted that share. Shouldn't be too difficult to implement; the file names that need to be accessed are part of the data metadata that we already download (to check for deleted frames).

* Limiting the dataset to one or several jobs that are part of the task. To support this, we need to a) filter out frames that are not in those jobs, and b) query the job annotations instead of task annotations and combine them ourselves. I assume CVAT combines annotations on overlapping frames by just concatenating the corresponding arrays, but I would have to check to be sure.

nmanovic (approved):
OK. It looks good to me as an initial version. Let's us merge it.

SpecLad:
Oh, I forgot another note:

I'm currently wiping the entire task from the cache when a change in the `updated_date` is detected. This includes the data chunks, which might seem rather imprudent, given that CVAT (as far as I know) does not let the user modify the data once it has been uploaded, and it would be much more efficient to keep the chunks when wiping out the cache.

This is just because I wanted to protect against the scenario where a future version of CVAT _does_ introduce a way to alter task data, and a user with an old version of the SDK receives stale data, because that version never purges data from the cache.

Ideally, there should be a way to independently verify how up-to-date each chunk is, so that we can redownload the absolutely minimum amount of information needed. I was thinking we could use conditional HTTP queries (AKA `If-Modified-Since`) to redownload chunks if the `updated_date` of a task does not match the last modified timestamp of the local chunk. If the server replies that the chunk hasn't changed, we simply bump the last modified date of the local chunk file. This would be fairly efficient (still need one HTTP query per chunk, but no actual data retransmission in the common case) and more future-proof.

Of course, this would need the server to support `If-Modified-Since` queries. I don't know if it currently does, or how difficult it would be to implement. However, given that we offload file transfers to Django (via django-sendfile2), I wouldn't be surprised if we got conditional queries for free.

zhiltsov-max:
> [no way to specify label index in CVAT]
> Another way to help the user in this regard would be to let them specify a custom label-id-to-index mapping when instantiating the adapter class

Personally, I like this variant, since it is specific to the training. At least, as an option. On the other hand, the ability to specify it in CVAT directly would make the process more transparent and provide a single point of configuration for this. Many datasets rely on the label order.

> [all the cache is cleaned] This is just because I wanted to protect against the scenario where a future version of CVAT does introduce a way to alter task data

Probably, it's better to keep cached chunks. CVAT doesn't allow to change data once it's uploaded, and there are no expected changes in this logic. The data can be changed, however, if we use a shared folder and the caching is enabled. The data can also be changed in a project. Per chunk update reporting sounds good to me, but it's probably quite far from what we have now.

> [projects are not implemented, because it is difficult to use them]

I think, we can allow to select a subset to work with.

SpecLad:
> Another way to help the user in this regard would be to let them specify a custom label-id-to-index mapping when instantiating the adapter class.

This is now implemented in #5455.

PR #5458, opened 2022-12-13
tests: fix long shutdown times for the webhook_receiver container

SpecLad:
Currently, `server.py` runs as PID 1, which means that it won't be terminated by a `SIGTERM` signal unless it explicitly handles it (which it doesn't). So when Docker tries to shut the container down, it sends the server a `SIGTERM`, which gets ignored, and then sits there for 10 seconds before sending it a `SIGKILL`.

To work around this, enable the built-in Docker init program, which forwards signals to the Python server. Since the Python server is no longer PID 1, `SIGTERM` will now shut it down immediately.

The `init` option is supported starting from the Compose format version 3.7, so bump the version.

### Motivation and context

I noticed that `webhook_receiver` takes the longest of all test containers to shut down, even though it's a trivial server.

### How has this been tested?

`time docker container stop test_webhook_receiver_1`

### Checklist

### License

### DIFF
--- a/tests/docker-compose.test_servers.yml
+++ b/tests/docker-compose.test_servers.yml
@@ -1,10 +1,11 @@
-version: '3.3'
+version: '3.7'
 
 services:
   webhook_receiver:
     image: python:3.9-slim
     restart: always
     command: python3 /tmp/server.py
+    init: true
     env_file:
       - ./tests/python/webhook_receiver/.env
     expose:


kirill-sizov (approved):
LGTM

SpecLad:
@nmanovic Mind merging this?

nmanovic (approved):
LGTM, I believe it is the simplest fix for the issue.

PR #5484, opened 2022-12-17
Limit test execution time

zhiltsov-max:
### Motivation and context

This should protect us from unnoticed hangs, which happen sometimes and waste CI and developer time.
Currently, the limit is 15s per test, which should be enough in normal conditions.

### Checklist

### License

### DIFF
--- a/tests/python/pytest.ini
+++ b/tests/python/pytest.ini
@@ -1,2 +1,6 @@
 [pytest]
+required_plugins = pytest-timeout
 addopts = --verbose --capture=tee-sys
+# We expect no regular individual test to run too long
+# can be overridden for specific tests with a test decorator
+timeout = 15
--- a/tests/python/requirements.txt
+++ b/tests/python/requirements.txt
@@ -1,4 +1,5 @@
 pytest==6.2.5
+pytest-timeout==2.1.0
 requests==2.26.0
 deepdiff==5.6.0
 boto3==1.17.61


nmanovic:
@zhiltsov-max , before making the change, we need to understand the current statistics for our tests. What is the average and maximum time in 10 runs? How does it depend on a PC (because we don't control that)

nmanovic:
@zhiltsov-max , the change can easily can make our tests unstable. I will prefer to run tests till the first fail. It will save time. What do you think?

zhiltsov-max:
> the change can easily can make our tests unstable. I will prefer to run tests till the first fail. It will save time. What do you think?

I like the idea. It seems, however, that the library worked this way some time ago, but then they decided to fail a test instead.

> before making the change, we need to understand the current statistics for our tests. What is the average and maximum time in 10 runs? How does it depend on a PC (because we don't control that)

I think we should choose a meaningful limit, which will deliver comfortable and adequate development experience. I think times up to 1 minute are good enough, because in general, we don't expect tests to work for a long time, and if it is expected for some tests, we can state this explicitly. The only reason for the limit to exist it to protect the developers from unexpected test hangs. Probably, we can extend the timeout up to 1 or 5 minutes, and try to fail the suite in the case of a problem.

azhavoro (approved):

nmanovic (approved):

PR #5498, opened 2022-12-21
Simplify upload data for task

Marishka17:
### Motivation and context

It's possible to specify only the manifest file and filename pattern for creating task with cloud storage data. 
The special characters supported now for the pattern are `*`, `?`, `[seq]`, `[!seq]`.
Please see [here](https://github.com/opencv/cvat/blob/8898a8b2647514dd6f3f6ce83745b1ca8ef72bce/tests/python/rest_api/test_tasks.py#L686) for some examples of how to use this functionality.
### How has this been tested?

Manually, added REST API test
### Checklist

~~- [ ] I have increased versions of npm packages if it is necessary ([cvat-canvas](https://github.com/cvat-ai/cvat/tree/develop/cvat-canvas#versioning),

### License


cvat/apps/engine/serializers.py
```diff
@@ -391,11 +392,25 @@ def validate_chunk_size(self, value):
             raise serializers.ValidationError('Chunk size must be a positive integer')
         return value
 
+    def validate_pattern(self, value):
+        import string
+        supported_wildcards = set(["*", "[", "]", "."])
+        no_escapted_symbols = set(['!', '"', '%', "'", ',', '/', ':', ';', '<', '=', '>', '@', "`", "_"])
+        legal_special_chars = supported_wildcards.union(no_escapted_symbols)
+        for c in set(string.punctuation) - legal_special_chars:
```
zhiltsov-max, line 400:
```suggestion
        non_escaped_symbols = set(['!', '"', '%', "'", ',', '/', ':', ';', '<', '=', '>', '@', "`", "_"])
        allowed_special_chars = supported_wildcards.union(non_escaped_symbols)
        for c in set(string.punctuation) - allowed_special_chars:
```

tests/python/rest_api/test_tasks.py
```diff
@@ -15,9 +18,17 @@
...
 
 from .utils import export_dataset
 
+sys.path.append(
+    os.path.join(
+        os.path.dirname(os.path.dirname(os.path.abspath(__file__))).split("tests")[0], "utils"
+    )
+)
+from dataset_manifest import ImageManifestManager
+
```
zhiltsov-max, line 31: I suggest to do one of the following:
- create a manifest via the server docker image as per documentation, the image must be installed anyway to run the tests
- create a separate package from the dataset_manifest, require it in the test deps (#5096)
- require dataset_manifest deps in the test deps and add it to the tests PATH

tests/python/rest_api/test_tasks.py
```diff
@@ -15,9 +18,17 @@
 
 from shared.utils.config import get_method, make_api_client, patch_method
 from shared.utils.helpers import generate_image_files
+from shared.utils.s3 import make_client
```
zhiltsov-max, line 21: I suggest to import this s3 helper as a package, this will make calls like `make_client()` much more clear: `s3.make_client()`. We already have several SDK and test helpers called `make_client`, it's simple to confuse them.

zhiltsov-max (changes requested):
Could you please add:
- PR description
- Any documentation about this feature

cvat/apps/engine/serializers.py
```diff
@@ -371,12 +371,13 @@ class DataSerializer(WriteOnceMixin, serializers.ModelSerializer):
     use_cache = serializers.BooleanField(default=False)
     copy_data = serializers.BooleanField(default=False)
     cloud_storage_id = serializers.IntegerField(write_only=True, allow_null=True, required=False)
+    pattern = serializers.CharField(allow_null=True, required=False)
```
zhiltsov-max, line 374:
```suggestion
    filename_pattern = serializers.CharField(allow_null=True, required=False)
```

cvat/apps/engine/serializers.py
```diff
@@ -391,11 +392,25 @@ def validate_chunk_size(self, value):
             raise serializers.ValidationError('Chunk size must be a positive integer')
         return value
 
+    def validate_pattern(self, value):
+        import string
```
zhiltsov-max, line 396: What is the point of importing the package here? How it is better importing in the file beginning?

Marishka17: I forgot to move it, thanks

cvat/apps/engine/task.py
```diff
@@ -300,69 +328,76 @@ def _create_thread(db_task, data, isBackupRestore=False, isDatasetImport=False):
...
-    media, task_mode = _validate_data(media, manifest_files)
-
-    if data['server_files']:
-        if db_data.storage == models.StorageChoice.LOCAL:
-            _copy_data_from_source(data['server_files'], upload_dir, data.get('server_files_path'))
-        elif db_data.storage == models.StorageChoice.SHARE:
-            upload_dir = settings.SHARE_ROOT
-
+    # find and validate manifest file
+    manifest_files = _count_manifest_files(data)
```
zhiltsov-max, line 338:
```suggestion
    manifest_files = _find_and_validate_manifest_files(data)
```

cvat/apps/engine/task.py
```diff
@@ -300,69 +328,76 @@ def _create_thread(db_task, data, isBackupRestore=False, isDatasetImport=False):
...
     manifest_root = None
-    if db_data.storage in {models.StorageChoice.LOCAL, models.StorageChoice.SHARE}:
+
+    # we sould also handle this case because files from the share source have not been downloaded yet
```
zhiltsov-max, line 341:
```suggestion
    # we should also handle this case because files from the share source have not been downloaded yet
```

cvat/apps/engine/task.py
```diff
@@ -300,69 +328,76 @@ def _create_thread(db_task, data, isBackupRestore=False, isDatasetImport=False):
...
-                file_name = properties['name']
-                properties['name'] = os.path.join(cloud_storage_manifest_prefix, file_name)
-                return properties
-            content = list(map(_add_prefix, raw_content))
+        cloud_storage_manifest_prefix = os.path.dirname(manifest_file)
+
+    # update list with server files if task creation approach with pattern and manifest file is used
+    if is_data_in_cloud and data['pattern']:
+        if 1 != len(data['server_files']):
+            raise ValidationError('Using a pattern is only supported with a manifest, but others files were found')
```
zhiltsov-max, line 369: The message gives very little troubleshooting information.

tests/python/shared/utils/helpers.py
```diff
@@ -18,10 +18,10 @@ def generate_image_file(filename="image.png", size=(50, 50), color=(0, 0, 0)):
     return f
 
 
-def generate_image_files(count) -> List[BytesIO]:
+def generate_image_files(count, prefixes=None) -> List[BytesIO]:
     images = []
     for i in range(count):
-        image = generate_image_file(f"{i}.jpeg", color=(i, i, i))
+        image = generate_image_file(f"{prefixes[i] if prefixes else ''}{i}.jpeg", color=(i, i, i))
```
zhiltsov-max, line 24:
```suggestion
        prefix = prefixes[i] if prefixes else ''
        image = generate_image_file(f"{prefix}{i}.jpeg", color=(i, i, i))
```

utils/dataset_manifest/core.py
```diff
@@ -195,8 +195,11 @@ def __iter__(self):
                 image = next(sources)
                 img = Image.open(image, mode='r')
                 orientation = img.getexif().get(274, 1)
-                img_name = os.path.relpath(image, self._data_dir) if self._data_dir \
-                    else os.path.basename(image)
+                if isinstance(image, str):
+                    img_name = os.path.relpath(image, self._data_dir) if self._data_dir \
+                        else os.path.basename(image)
+                else:
```
zhiltsov-max, line 201: Which type this value is supposed to have?

cvat/apps/engine/task.py
Marishka17: I added this comment on a block of code (validation process is in `_validate_manifest` function)

tests/python/rest_api/test_tasks.py
```diff
@@ -674,6 +685,77 @@ def test_create_task_with_cloud_storage_files(
...
+        for image in images:
+            s3_client.create_file(
+                data=image,
+                bucket=cloud_storage["resource"],
+                filename=f"{'test/sub/' if sub_dir else ''}{image.name}",
+            )
+
+        with NamedTemporaryFile(mode="wb+", suffix="manifest.jsonl") as tmp_file:
+            tmp_name = tmp_file.name
+            tmp_manifest = ImageManifestManager(manifest_path=tmp_name)
```
zhiltsov-max, line 720:
```suggestion
            tmp_manifest = ImageManifestManager(manifest_path=tmp_file.name)
```

tests/python/rest_api/test_tasks.py
```diff
@@ -674,6 +685,77 @@ def test_create_task_with_cloud_storage_files(
...
+            task_id = self._test_create_task(
+                self._USERNAME, task_spec, data_spec, content_type="application/json", org=org
+            )
+
+            with make_api_client(self._USERNAME) as api_client:
+                (task, response) = api_client.tasks_api.retrieve(task_id, org=org)
+                assert response.status == HTTPStatus.OK
+                assert task.size == task_size
+        else:
+            self._test_cannot_create_task(self._USERNAME, task_spec, data_spec)
```
zhiltsov-max, line 757: It would be better to also check the failure reason.

Marishka17: Fixed

tests/python/rest_api/test_tasks.py
Marishka17: Not relevant anymore

utils/dataset_manifest/core.py
Marishka17: BytesIO, but not relevant anymore

cvat/apps/engine/task.py
Marishka17: Fixed

tests/python/rest_api/test_tasks.py
```diff
@@ -674,6 +685,77 @@ def test_create_task_with_cloud_storage_files(
...
+        s3_client = make_client()
+
+        cloud_storage = cloud_storages[cloud_storage_id]
+
+        for image in images:
+            s3_client.create_file(
+                data=image,
+                bucket=cloud_storage["resource"],
+                filename=f"{'test/sub/' if sub_dir else ''}{image.name}",
+            )
```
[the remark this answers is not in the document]
Marishka17: Fixed

tests/python/rest_api/test_tasks.py
Marishka17: Fixed

tests/python/rest_api/test_tasks.py
Marishka17: The first option has been implemented

cvat/apps/engine/serializers.py
```diff
@@ -391,11 +393,24 @@ def validate_chunk_size(self, value):
             raise serializers.ValidationError('Chunk size must be a positive integer')
         return value
 
+    def validate_filename_pattern(self, value):
+        supported_wildcards = set(["*", "[", "]", "."])
```
zhiltsov-max, line 397: If I understand correctly that the dot is for 1 any character, then it probably should be changed to `?` in wildcards and then changed to `.` for regex.

tests/python/rest_api/test_tasks.py
```diff
@@ -674,6 +679,107 @@ def test_create_task_with_cloud_storage_files(
...
+
+            with make_api_client(self._USERNAME) as api_client:
+                (task, response) = api_client.tasks_api.retrieve(task_id, org=org)
+                assert response.status == HTTPStatus.OK
+                assert task.size == task_size
+        else:
+            status = self._test_cannot_create_task(self._USERNAME, task_spec, data_spec)
+            assert "No media data found" in status.message
+
+        with ExitStack() as stack:
```
zhiltsov-max, line 772: The callbacks should be added right after the addition, otherwise the function may exit before this code is executed. I suggest to use the [`request`](https://docs.pytest.org/en/6.2.x/fixture.html#adding-finalizers-directly) fixture to make code clearer. An example is [here](https://github.com/opencv/cvat/blob/7d32a7d857a14063bb0be91f8545ff79fce06b6f/tests/python/rest_api/test_limits.py#L526-L536).

cvat/apps/engine/serializers.py
```diff
@@ -396,6 +397,7 @@ def validate(self, attrs):
         if 'start_frame' in attrs and 'stop_frame' in attrs \
             and attrs['start_frame'] > attrs['stop_frame']:
             raise serializers.ValidationError('Stop frame must be more or equal start frame')
+
```
zhiltsov-max, line 400: Probably, we still need to validate the pattern expression.

Marishka17: @zhiltsov-max, Could you please provide pattern example when we need to handle it? (If we use regex I agree that validation is needed). I guess It is not necessary anymore.

zhiltsov-max: My guess is that we could check for invalid characters in the pattern and produce a meaningful error message in case of problems. It's up to you to decide.

zhiltsov-max (approved):
I'd like to have any documentation on the feature, otherwise it's almost like "dead code". We could also add a CLI option for this feature. It can be done in other PRs.

tests/python/shared/utils/helpers.py
```diff
@@ -18,10 +18,11 @@ def generate_image_file(filename="image.png", size=(50, 50), color=(0, 0, 0)):
     return f
 
 
-def generate_image_files(count) -> List[BytesIO]:
+def generate_image_files(count, prefixes=None) -> List[BytesIO]:
     images = []
     for i in range(count):
-        image = generate_image_file(f"{i}.jpeg", color=(i, i, i))
+        prefix = prefixes[i] if prefixes else ""
```
nmanovic, line 24: It works only because len(prefixes) == count in the code. Probably it is a bug. Let's fix in your next PR.

PR #5519, opened 2022-12-27
test task - regression testing to export tasks

AparinAA:
Added test to https://github.com/opencv/cvat/pull/5396

### DIFF
--- a/tests/cypress/integration/regTest_for_export_task/innerExportTask.js
+++ b/tests/cypress/integration/regTest_for_export_task/innerExportTask.js
@@ -1,0 +1,138 @@
+// Start writing your Cypress tests below!
+// If you're unfamiliar with how Cypress works,
+// check out the link below and learn how to write your first test:
+// https://on.cypress.io/writing-first-test
+
+/// <reference types="cypress" />
+
+Cypress.Commands.add('exportInnerTask', ({
+    taskName,
+    archiveCustomName,
+    order,
+    imageClick = false,
+}) => {
+    const orderId = Number.isNaN(parseInt(order, 10)) ? 0 : parseInt(order, 10);
+    const archFileNameFin = `${orderId}_${archiveCustomName}_${taskName}`;
+    cy.get('.cvat-task-page-actions-button')
+        .trigger('mouseover');
+    cy.get('.ant-dropdown')
+        .not('.ant-dropdown-hidden')
+        .within(() => {
+            cy.contains('[role="menuitem"]', new RegExp('^Export task dataset$')).click();
+        });
+    cy.get('.cvat-modal-export-task')
+        .should('be.visible')
+        .find('.cvat-modal-export-select')
+        .click();
+    cy.get('.cvat-modal-export-option-item')
+        .eq(orderId)
+        .click();
+    cy.get('.cvat-modal-export-task')
+        .find('.cvat-modal-export-filename-input')
+        .type(archFileNameFin);
+
+    if (imageClick) {
+        cy.get('.cvat-modal-export-save-images').click();
+    }
+
+    cy.get('.cvat-modal-export-task').contains('button', 'OK').click();
+
+    cy.get('.cvat-notification-notice-export-task-start').should('be.visible');
+    cy.closeNotification('.cvat-notification-notice-export-task-start');
+    cy.getDownloadFileName().then((file) => {
+        cy.verifyDownload(file);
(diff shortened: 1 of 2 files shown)


nmanovic:
@bsekachev , could you please look at the PR?

tests/cypress/integration/regTest_for_export_task/innerExportTask.js
```diff
@@ -0,0 +1,138 @@
...
+        cy.saveJob();
+        cy.goToTaskList();
+    });
+
+    after(() => {
+        cy.goToTaskList();
+        cy.deleteTask(taskName);
+    });
+
+    describe(`Testing "${taskName}"`, () => {
```
bsekachev, line 113: Better to be more specific here. It is not clear what we are testing. Using string template is not the best option here because it makes code more difficult to read and understand.

tests/cypress/integration/regTest_for_export_task/innerExportTask.js
```diff
@@ -0,0 +1,138 @@
...
+        it('Open a task', () => {
+            cy.openTask(taskName);
+        });
```
bsekachev, line 116: Should be a part of ``before`` probably, because we don't test opening a task here

tests/cypress/integration/regTest_for_export_task/innerExportTask.js
```diff
@@ -0,0 +1,138 @@
+// Start writing your Cypress tests below!
+// If you're unfamiliar with how Cypress works,
+// check out the link below and learn how to write your first test:
+// https://on.cypress.io/writing-first-test
```
bsekachev, line 4: We need to put a lincese header here instead of default Cypress text:

// Copyright (C) 2022 CVAT.ai Corporation
//

tests/cypress/integration/regTest_for_export_task/innerExportTask.js
```diff
@@ -0,0 +1,138 @@
...
+
+/// <reference types="cypress" />
+
+Cypress.Commands.add('exportInnerTask', ({
```
bsekachev, line 8: There is a dedicated files for Cypress commands definition:
```
tests/cypress/support/*.js
```

tests/cypress/integration/regTest_for_export_task/innerExportTask.js
```diff
@@ -0,0 +1,138 @@
...
+    taskName,
+    archiveCustomName,
+    order,
+    imageClick = false,
+}) => {
+    const orderId = Number.isNaN(parseInt(order, 10)) ? 0 : parseInt(order, 10);
```
bsekachev, line 14:
```suggestion
    const orderId = +order || 0; 
```

But generally, if we need an integer here, passing a string to this command does not make sense

tests/cypress/integration/regTest_for_export_task/innerExportTask.js
```diff
@@ -0,0 +1,138 @@
...
+    const height = 800;
+    const posX = 10;
+    const posY = 10;
+    const color = 'gray';
+    const archiveName = `${imageFileName}.zip`;
+    const archivePath = `cypress/fixtures/${archiveName}`;
+    const imagesFolder = `cypress/fixtures/${imageFileName}`;
+    const directoryToArchive = imagesFolder;
+    const newLabelName = 'person';
+    const exportOrders = [0, 1, 2];
```
bsekachev, line 67: Choosing an export format by its order is not robust option I think. Order can be changed easily. 
Better to choose by name

tests/cypress/integration/regTest_for_export_task/outerExportTask.js
```diff
@@ -0,0 +1,138 @@
+// Start writing your Cypress tests below!
```
bsekachev, line 1: 1. Please check comments from the previous file, many of them applicable here
2. A lot of code dublication, I would prefer rework it somehow to reduce it. Maybe to combine these two tests into the one.

bsekachev:
Looks like the test has not been run on CI because of configuration of ``tests/cypress.json`` and probably some configuration in ``.github/workflows``.  Please, correct me if I am wrong.

nmanovic:
@AparinAA , thanks for PR. I will close it.

AparinAA:
@nmanovic 

OK. I can fix the wrong if needed

PR #5521, opened 2022-12-28
Fix vulnerability

Marishka17:


nmanovic:
@Marishka17 , could you please add a line into CHANGELOG.md? Need to update the security section.

PR #5522, opened 2022-12-28
multiple fixes for helm chart

Keramblock:
Hi, if I want to use ingress from your helm chart - that does not necessary means that i want to use traefik, so I moved option to allow traefik to separate place


Keramblock:
Hi, @azhavoro, could you please help me with helm chart?

Right now I see, that you have created there 3 deployments, all of them are using same pvc for different stuff like:
https://github.com/opencv/cvat/blob/develop/helm-chart/templates/cvat_backend/worker_default/deployment.yml#L162
https://github.com/opencv/cvat/blob/develop/helm-chart/templates/cvat_backend/worker_low/deployment.yml#L162
https://github.com/opencv/cvat/blob/develop/helm-chart/templates/cvat_backend/server/deployment.yml#L207

That is lead to broken state in k8s with default cloud RWO storage mode because of the conflict:
image
image

So is there any reason to use pvc like that? Do different serve

Keramblock:
Also for what they are currently used, Right now I dont see any info about them in main or helm readme?

Keramblock:
@nmanovic or mb you know the answer? Could you please help me with this issues because as far as i can see - helm chart is not usable as is and need to be fixed =(

Keramblock:
also fixed broken opa version

azhavoro:
> Also for what they are currently used? Right now I dont see any info about them in main or helm readme =(

@Keramblock Hi, sorry for the late response
`server`, `worker_default` and `worker_low` are use the same data and in general case a RWX storage like NFS is needed here (or you have to use affinity to assing all cvat pods for a single node). I agree that this should have been described in the readme at least. Currently I have only one idea how to improve it - use intermediate NFS server that uses RWO storage and create corresponding NFS backed PV and PVC that can be used by many CVAT pods, but I'm not sure about performanse in this case. What do you think?

Keramblock:
@azhavoro RWX usually absent in default cloud k8s setup(GKE/AKS/etc) so imo we should avoid using it, we could use one RWO with readonly flags for multiple reads, if that is suitable for this services. 
Otherwise this chart will not be useful for large portion(my estimate is >95%) of kubernetes clusters worldwide.

Keramblock:
Hi, @azhavoro I see, that this PR is effectively dead and I cant dedicate to CVAT helm enough time right now, but I do need some changes and I believe, that changes are necessary for whole community. Could I create issues instead, so you/someone else could fix them?

Keramblock:
This PR is to outdated, so I am closing it.

PR #5556, opened 2023-01-06
Improve error messages when limits reached

kirill-sizov:
### Checklist

- [ ] I submit my changes into the `develop` branch
- [ ] I have added a description of my changes into [CHANGELOG](https://github.com/cvat-ai/cvat/blob/develop/CHANGELOG.md) file
- [ ] I have updated the [documentation](
  https://github.com/cvat-ai/cvat/blob/develop/README.md#documentation) accordingly
- [ ] I have added tests to cover my changes
- [ ] I have linked related issues ([read github docs](
  https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword))
- [ ] I have increased versions of npm packages if it is necessary ([cvat-canvas](https://github.com/cvat-ai/cvat/tree/develop/cvat-canvas#versioning),
[cvat-core](https://github.com/cvat-ai/cvat/tree/develop/cvat-core#versioning), [cvat-data](https://github.com/cvat-ai/cvat/tree/develop/cvat-data#versioning) and [cvat-ui](https://github.com/cvat-ai/cvat/tree/develop/cvat-ui#versioning))

### License

- [ ] I submit _my code changes_ under the same [MIT License](
  https://github.com/cvat-ai/cvat/blob/develop/LICENSE) that covers the project.
  Feel free to contact the maintainers if that's a concern.

PR #5612, opened 2023-01-20
Fix a CI issue that causes unit tests to run against the wrong version of CVAT

SpecLad:
There seems to be a bug somewhere in the Docker ecosystem (it's probably either Docker Compose, Docker Buildx or BuildKit) that causes `docker compose build` to ignore base images that are already present in the system, and instead fetch them from Docker Hub, if there's a custom Buildx builder configured. There's a bug report here:
<https://github.com/docker/compose/issues/9939>.

This bug means that when the build pipeline builds the `cvat_ci` image, it's based on the latest release of `cvat/server` from Docker Hub instead of the version that we just built. Consequently, we run the unit tests against that release instead of the development version.

Fortunately, we don't actually need to set up a Buildx builder in most jobs (including the `unit_testing` job), so just don't do that.

Also, use `cvat/server:local` as the base image in `Dockerfile.ci`. This will prevent a similar bug from reoccurring in the future, since the `local` tag should never be uploaded to Docker Hub.

### Motivation and context

I saw a unit test that should fail due to a regression since 2.3.0, yet it still succeeds on CI. This should prevent cases like that.

### How has this been tested?

I only changed the build pipelines, so I can't test this myself. 🙂 

### Checklist

### License

### DIFF
--- a/Dockerfile.ci
+++ b/Dockerfile.ci
@@ -1,4 +1,4 @@
-FROM cvat/server
+FROM cvat/server:local
 
 ENV DJANGO_CONFIGURATION=testing
 USER root
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -135,9 +135,6 @@ jobs:
         with:
           python-version: '3.8'
 
-      - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v2
-
       - name: Download CVAT server image
         uses: actions/download-artifact@v3
         with:
@@ -194,9 +191,6 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-
-      - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v2
 
       - name: Download CVAT server image
         uses: actions/download-artifact@v3
@@ -264,9 +258,6 @@ jobs:
                 'canvas3d_functionality_2', 'issues_prs', 'issues_prs2', 'masks', 'skeletons']
     steps:
       - uses: actions/checkout@v3
-
-      - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v2
 
       - uses: actions/setup-node@v3
         with:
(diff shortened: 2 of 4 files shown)


SpecLad:
Note: I'm expecting the unit tests to start failing because of this fix, so this PR shouldn't be merged just yet.

SpecLad:
@kirill-sizov Could you take a look at the remaining test failures? I believe they are caused by #5556.

Dockerfile.ci
```diff
@@ -1,4 +1,4 @@
-FROM cvat/server
+FROM cvat/server:local
```
kirill-sizov, line 1: This solution is fine for our CI workflows, but it can confuse developers who runs unit tests locally.
I would suggest such solution:
```docker
ARG CVAT_VERSION
FROM cvat/server:${CVAT_VERSION}
```
And add CVAT_VERSION argument into the docker-compose.ci.yml
```YAML
  cvat_ci:
    image: cvat_ci
    network_mode: host
    build:
      context: .
      dockerfile: Dockerfile.ci
      args:
        - CVAT_VERSION=${CVAT_VERSION:-latest}
```

But maybe it's not really important, @SpecLad What do you think?

SpecLad: > This solution is fine for our CI workflows, but it can confuse developers who run unit tests locally.

But isn't this file specifically _for_ CI workflows? It's even called `Dockerfile.ci`. 🙂 

> I would suggest such solution:

Yeah, I thought of doing something like that. I didn't do it for a couple of reasons:

1. It didn't seem especially useful (as per above).
2. It makes regressions more likely, because if `CVAT_VERSION` it not added to the environment (for whatever reason), then it'll silently fall back to `:latest` again.

IMO, we should move in the other direction and make it so that developers use the `:local` tag by default. That would eliminate all confusion between published and development versions of the images. But that goes beyond fixing the immediate issue, so I didn't want to spend time on it.

kirill-sizov (approved):
LGTM

nmanovic (approved):
LGTM

PR #5632, opened 2023-01-26
Clean up data when a project is deleted.

nmanovic:
Fix https://github.com/opencv/cvat/issues/5595

### DIFF
--- a/cvat/apps/engine/signals.py
+++ b/cvat/apps/engine/signals.py
@@ -3,21 +3,17 @@
 # SPDX-License-Identifier: MIT
 import shutil
 
+from django.contrib.auth.models import User
 from django.db.models.signals import post_delete, post_save
 from django.dispatch import receiver
-from django.contrib.auth.models import User
 
-from .models import (
-    Data,
-    Job,
-    StatusChoice,
-    Task,
-    Profile,
-)
+from .models import (CloudStorage, Data, Job, Profile, Project,
+    StatusChoice, Task)
 
 
-@receiver(post_save, sender=Job, dispatch_uid="update_task_status")
-def update_task_status(instance, **kwargs):
+@receiver(post_save, sender=Job,
+    dispatch_uid=__name__ + ".update_task_status_handler")
+def __update_task_status_handler(instance, **kwargs):
     db_task = instance.segment.task
     db_jobs = list(Job.objects.filter(segment__task_id=db_task.id))
     status = StatusChoice.COMPLETED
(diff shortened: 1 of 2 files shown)


cvat/apps/engine/signals.py
```diff
@@ -3,21 +3,17 @@
...
-    Profile,
-)
+from .models import (CloudStorage, Data, Job, Profile, Project,
+    StatusChoice, Task)
 
 
-@receiver(post_save, sender=Job, dispatch_uid="update_task_status")
-def update_task_status(instance, **kwargs):
+@receiver(post_save, sender=Job,
+    dispatch_uid=__name__ + ".update_task_status_handler")
```
zhiltsov-max, line 15: Where do these names come from? I suggest to declare they somewhere or provide a function to access them.

nmanovic: Directories are always different. They have ID of objects from DB inside. It is not possible.

cvat/apps/engine/signals.py
```diff
@@ -3,21 +3,17 @@
...
-    Task,
-    Profile,
-)
+from .models import (CloudStorage, Data, Job, Profile, Project,
+    StatusChoice, Task)
 
 
-@receiver(post_save, sender=Job, dispatch_uid="update_task_status")
-def update_task_status(instance, **kwargs):
+@receiver(post_save, sender=Job,
```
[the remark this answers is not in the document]
nmanovic: Directories are always different. They have ID of objects from DB inside. It is not possible.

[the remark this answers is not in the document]
zhiltsov-max: As I remember, owning entity id (e.g. task, project) and data ids are different. There are tests in the [PR](https://github.com/opencv/cvat/pull/5083/files#diff-890920c4573bb488ff08dd5351c6d43e2a22bfbcccd5ee61fa53e9f37657bb46R56-R59) that found a case of collision (in a purely testing scenario yet). Can it become a problem in this solution?

cvat/apps/engine/signals.py
zhiltsov-max: Okay. As I understand, this id is supposed to be unique and should not be used from anywhere. Maybe it would be better to supply an UUID then?

nmanovic: I'm not sure it will be better than the proposed solution.

cvat/apps/engine/signals.py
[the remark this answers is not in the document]
nmanovic: @zhiltsov-max , could you please give me more information about the problem you mean? Basically the code below removes directories:
- For Jobs: /home/django/data/jobs/<id>
- For Tasks: /home/django/data/tasks/<id>
- For Projects: /home/django/data/projects/<id>
- For Data: /home/django/data/data/<id>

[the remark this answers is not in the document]
zhiltsov-max: From the linked fragment of code:
```
# The parent class clears DB changes, and it can lead to under-cleaned task data,
# which can affect other tests.
# This situation is not expected to happen on a real server, because
# cache keys include Data object ids, which cannot be reused or freed.
```
Thus, if:
1. A new task is created
2. Task data is uploaded
3. The task removed
4. A new task is created
5. The second task data is uploaded, the files have overlapping names and the new data directory id is the same as the first task had, and the cache is used

then the second task can look to the same cache keys as the first task had and receive matching frames from the first task data, instead of the new data. AFAIK, this should not happen on real servers, because data ids always grow.

cvat/apps/engine/signals.py
```diff
@@ -30,18 +26,45 @@ def update_task_status(instance, **kwargs):
...
     if not hasattr(instance, 'profile'):
         profile = Profile()
         profile.user = instance
         profile.save()
 
-@receiver(post_delete, sender=Task, dispatch_uid="delete_task_files_on_delete_task")
-def delete_task_files_on_delete_task(instance, **kwargs):
+@receiver(post_delete, sender=Project,
+    dispatch_uid=__name__ + ".delete_project_handler")
+def __delete_project_handler(instance, **kwargs):
```
SpecLad, line 39: These handlers are named inconsistently:

* `__delete_project_handler` - handles project deletion.
* `__create_profile_handler` - handles _user_ creation by creating a profile.

They should all be named either for what they handle or for what they do.

cvat/apps/engine/signals.py
```diff
@@ -3,21 +3,17 @@
...
-)
+from .models import (CloudStorage, Data, Job, Profile, Project,
+    StatusChoice, Task)
 
 
-@receiver(post_save, sender=Job, dispatch_uid="update_task_status")
-def update_task_status(instance, **kwargs):
+@receiver(post_save, sender=Job,
+    dispatch_uid=__name__ + ".update_task_status_handler")
+def __update_task_status_handler(instance, **kwargs):
```
SpecLad, line 16: I would not recommend to use names beginning with two underscores, because such names can be subject to Python's name mangling feature. One underscore is sufficient.

cvat/apps/engine/signals.py
```diff
@@ -30,18 +26,45 @@ def update_task_status(instance, **kwargs):
...
+            db_project.save()
+    except Project.DoesNotExist:
+        pass # probably the project has been deleted
 
-@receiver(post_delete, sender=Data, dispatch_uid="delete_data_files_on_delete_data")
-def delete_data_files_on_delete_data(instance, **kwargs):
+@receiver(post_delete, sender=Job,
+    dispatch_uid=__name__ + ".delete_job_handler")
+def __delete_job_handler(instance, **kwargs):
+    shutil.rmtree(instance.get_dirname(), ignore_errors=True)
```
SpecLad, line 59: I think `ignore_errors=True` is a bad idea, since that way we won't learn about any problems with deleting the directory. It would be better to pass an `onerror` callback that would log the error.

cvat/apps/engine/signals.py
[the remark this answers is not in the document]
SpecLad: Perhaps we can guard against this by checking that when we create a data directory, it does not already exist?

cvat/apps/engine/signals.py
nmanovic: It will be even good if it will be mangled. The name isn't important. It should not be used anywhere in the code.

cvat/apps/engine/signals.py
nmanovic: @SpecLad , in general I agree with the proposal. But our logging system right now depends on the DB objects and files in deleted directories. Such change can lead to unexpected problems. Let's wait while we have a new logging system from Andrey. It should write events directly into ClickHouse DB.

SpecLad: Okay, but can you at least add a TODO explaning this?

cvat/apps/engine/signals.py
nmanovic: renamed to `__save_job_handler`

cvat/apps/engine/signals.py
nmanovic: I have added TODO comment

nmanovic:
@zhiltsov-max , @SpecLad , do you have any other comments?

SpecLad (approved):

PR #5635, opened 2023-01-27
New models UI

klakhov:
### Checklist

### License


cvat-core/src/ml-model.ts
```diff
@@ -104,8 +114,106 @@ export default class MLModel {
...
+        },
+    },
+});
+
+Object.defineProperties(MLModel.prototype.delete, {
+    implementation: {
+        writable: false,
+        enumerable: false,
+        value: async function implementation(): Promise<void> {
+            if (Number.isInteger(this.id) && this.deletable) {
```
bsekachev, line 184: ``id`` is specified like a string in typings

bsekachev: SerializedModel

cvat-core/src/lambda-manager.ts
```diff
@@ -73,32 +92,59 @@ class LambdaManager {
...
         const timeoutCallback = async (): Promise<void> => {
             try {
                 this.listening[requestID].timeout = null;
-                const response = await serverProxy.lambda.status(requestID);
+                let response = null;
+                if (provider === ModelProviders.CVAT) {
+                    response = await serverProxy.lambda.status(requestID);
+                } else {
+                    response = await serverProxy.functions.status(requestID);
+                }
```
bsekachev, line 147: Can we reduce code dublication for this logic? I found 4 places.
For example we could return ``serverProxy.lambda`` or ``serverProxy.functions`` depending on ``provider`` value

cvat-core/src/lambda-manager.ts
```diff
@@ -18,18 +24,26 @@ class LambdaManager {
...
-        if (Array.isArray(this.cachedList)) {
-            return [...this.cachedList];
-        }
-
-        const result = await serverProxy.lambda.list();
+        let lambdaFunctions = [];
+        let functions = [];
+        try {
+            // lambda.list returns error if list is empty, but we should work with models anyway
+            lambdaFunctions = await serverProxy.lambda.list();
```
bsekachev, line 31: 1. why we have error in case with empty list? Do you mean 404?
2. it ignores all the errors, not sure it is correct
3. better to handle on server-proxy because it looks like server-specific thing

cvat-ui/src/actions/models-actions.ts
```diff
@@ -64,34 +91,87 @@ export const modelsActions = {
...
+        dispatch(modelsActions.getModels(query));
+
+        const filteredQuery = { ...query };
+        if (query) {
+            for (const key of Object.keys(query)) {
+                if ((filteredQuery as Indexable)[key] === null) {
+                    delete (filteredQuery as Indexable)[key];
+                }
+            }
+        }
```
bsekachev, line 126: Can we write a function to remove null values from objects and use it for all the pages (projects, tasks, models, etc..)?

cvat-ui/src/actions/models-actions.ts
```diff
@@ -64,34 +91,87 @@ export const modelsActions = {
...
+    return async function (dispatch) {
+        const model = new cvat.classes.MLModel(modelData);
+
+        dispatch(modelsActions.createModel());
+        try {
+            const createdModel = await model.save();
+            dispatch(modelsActions.createModelSuccess(createdModel));
+        } catch (error) {
+            dispatch(modelsActions.createModelFailed(error));
+            throw error;
```
bsekachev, line 147: What is the reason to throw the error after dispatching error action?

cvat-ui/src/actions/models-actions.ts
```diff
@@ -64,34 +91,87 @@ export const modelsActions = {
...
+
+export function deleteModelAsync(model: MLModel): ThunkAction {
+    return async function (dispatch) {
+        dispatch(modelsActions.deleteModel(model));
+        try {
+            await model.delete();
+            dispatch(modelsActions.deleteModelSuccess(model.id));
+        } catch (error) {
+            dispatch(modelsActions.deleteModelFailed(model.id, error));
+            throw error;
```
bsekachev, line 160: The same question here

cvat-ui/src/actions/models-actions.ts
bsekachev, line 147: What is the reason to throw the error after dispatching error action?

cvat-ui/src/actions/models-actions.ts
bsekachev, line 160: The same question here

cvat-ui/src/reducers/notifications-reducer.ts
```diff
@@ -803,6 +803,36 @@ export default function (state = defaultState, action: AnyAction): Notifications
...
+        }
+        case ModelsActionTypes.DELETE_MODEL_FAILED: {
+            return {
+                ...state,
+                errors: {
+                    ...state.errors,
+                    models: {
+                        ...state.errors.models,
+                        canceling: {
+                            message: 'Could not delete model',
```
bsekachev, line 829: Let's pass id or name here to be more specific (in case when user deletes several models)

cvat-ui/src/reducers/models-reducer.ts
```diff
@@ -15,27 +17,52 @@ const defaultState: ModelsState = {
...
+    query: {
+        page: 1,
+        id: null,
+        search: null,
+        filter: null,
+        sort: null,
+    },
+    providers: {
+        fetching: false,
+        current: [],
```
bsekachev, line 34: let's rename to just ``list`` or something like this, because ``current`` is better to be used with paginatable resources

cvat-ui/src/reducers/models-reducer.ts
```diff
@@ -15,27 +17,52 @@ const defaultState: ModelsState = {
...
         }
         case ModelsActionTypes.GET_MODELS_SUCCESS: {
             return {
                 ...state,
-                interactors: action.payload.models.filter((model: Model) => ['interactor'].includes(model.type)),
-                detectors: action.payload.models.filter((model: Model) => ['detector'].includes(model.type)),
-                trackers: action.payload.models.filter((model: Model) => ['tracker'].includes(model.type)),
-                reid: action.payload.models.filter((model: Model) => ['reid'].includes(model.type)),
+                interactors: action.payload.models.filter((model: MLModel) => (
+                    [ModelType.INTERACTOR].includes(model.type)
```
bsekachev, line 51: Why not use ``===`` here?

bsekachev: and below

cvat-ui/src/reducers/models-reducer.ts
```diff
@@ -47,6 +74,52 @@ export default function (state = defaultState, action: ModelsActions | AuthActio
...
+                    ...mutual,
+                    trackers: [...state.trackers, action.payload.model],
+                };
+            }
+            if ([ModelType.REID].includes(action.payload.model.type)) {
+                return {
+                    ...mutual,
+                    trackers: [...state.reid, action.payload.model],
+                };
+            }
```
bsekachev, line 117: Why we put ``reid`` to ``trackers``?

cvat-ui/src/reducers/models-reducer.ts
```diff
@@ -47,6 +74,52 @@ export default function (state = defaultState, action: ModelsActions | AuthActio
...
+                ...state,
+                fetching: false,
+            };
+        }
+        case ModelsActionTypes.CREATE_MODEL_SUCCESS: {
+            const mutual = {
+                ...state,
+                fetching: false,
+            };
+            if ([ModelType.INTERACTOR].includes(action.payload.model.type)) {
```
bsekachev, line 94: Fell like all this code can be reduced, like:

```
return {
 ...mutual,
[`${action.payload.model.type}s`]: [...`${action.payload.model.type}s`, action.payload.model]  
}
```

But this DOES NOT work for reid. Just a kind of idea to think about it.

cvat-ui/src/reducers/models-reducer.ts
```diff
@@ -47,6 +74,52 @@ export default function (state = defaultState, action: ModelsActions | AuthActio
...
+                    trackers: [...state.trackers, action.payload.model],
+                };
+            }
+            if ([ModelType.REID].includes(action.payload.model.type)) {
+                return {
+                    ...mutual,
+                    trackers: [...state.reid, action.payload.model],
+                };
+            }
+            return {
```
bsekachev, line 118: We lost classifiers?

cvat-core/src/ml-model.ts
```diff
@@ -26,19 +29,26 @@ interface ModelTip {
...
-    type: ModelType;
+    id?: string;
+    name?: string;
+    labels?: string[];
+    version?: number;
+    attributes?: Record<string, ModelAttribute>;
+    framework?: string;
+    description?: string;
+    kind?: ModelType;
+    type?: string;
```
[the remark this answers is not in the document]
klakhov: Yes, `kind` is our original `type` (detector, tracker, etc). And `type` now is field that is provided by model provider, for example: `instance-segmentation`, `object-detection`.

cvat-ui/src/actions/models-actions.ts
klakhov: We need it in create model form so we can catch it, and success message is not displayed

klakhov:
@nmanovic PR is green, I've fixed all known issues.

nmanovic (approved):

PR #5646, opened 2023-01-31
Audit logs

azhavoro:
### Checklist

### License


bsekachev:
Probably something does not work for ``job`` query parameter. 

image

I would suggest to rename ``organization`` parameter to ``org`` (like for other REST API endpoints)

bsekachev:
For load:job event we have job and task in payload (as task id and job id).  For other events we have them like a part of the body. I believe they should be in the body also for load:job event. 

image

Organization is always null, but I work in an organization. This information also can be received from POST request query parameter. This way it would work also for server-side create events for resources (like tasks, projects, etc).

cvat/apps/events/views.py
```diff
@@ -0,0 +1,142 @@
+# Copyright (C) 2018-2022 Intel Corporation
+#
+# SPDX-License-Identifier: MIT
```
bsekachev, line 3: Not Intel anymore (: 
The same in multiple files.

cvat/apps/events/export.py
```diff
@@ -0,0 +1,155 @@
...
+
+def export(request, queue_name):
+    action = request.query_params.get('action', None)
+    filename = request.query_params.get('filename', None)
+
+    query_params = {
+        'organization': request.query_params.get('organization', None),
+        'project': request.query_params.get('project', None),
+        'task': request.query_params.get('task', None),
+        'job': request.query_params.get('jid', None),
```
bsekachev, line 92: ``jid`` or ``job``?

.github/workflows/main.yml
```diff
@@ -53,10 +53,10 @@ jobs:
           mkdir /tmp/cvat_ui
           mkdir /tmp/cvat_sdk
 
+      # TO-DO: add cache-from
```
nmanovic, line 56: @azhavoro , should we revert the change?

components/analytics/clickhouse/init.sh
```diff
@@ -0,0 +1,47 @@
...
+
+CLICKHOUSE_DB="${CLICKHOUSE_DB:-cvat}";
+CLICKHOUSE_USER="${CLICKHOUSE_USER:-user}";
+CLICKHOUSE_PASSWORD="${CLICKHOUSE_PASSWORD:-password}";
+
+cat <<EOT > /etc/clickhouse-server/users.d/user.xml
+<yandex>
+  <!-- Docs: <https://clickhouse.tech/docs/en/operations/settings/settings_users/> -->
+  <users>
+    <${CLICKHOUSE_USER}>
```
nmanovic, line 11: Probably it is necessary to check that CLICKHOUSE_USER value without spaces and special symbols.

cvat-core/src/enums.ts
```diff
@@ -76,37 +76,36 @@ export enum Source {
 }
 
 export enum LogType {
```
nmanovic, line 78: I don't see `create:task`, `create:project`, `sort:task`, `filter:task`. Are we going to handle such events on UI?

cvat-core/src/log.ts
```diff
@@ -10,7 +10,7 @@ import { ArgumentError } from './exceptions';
 
 export class Log {
```
nmanovic, line 11: EventLogger

cvat-core/src/log.ts
```diff
@@ -137,74 +148,45 @@ class LogWithObjectsInfo extends Log {
...
-        ) {
-            const message = `
-                The field "working_time" is required for ${this.type} log. It must be a number not less than 0
-            `;
-            throw new ArgumentError(message);
-        }
-    }
-}
-
 class LogWithExceptionInfo extends Log {
```
nmanovic, line 151: Do we really need a separate class? I believe it can be decided based on LogType (maybe EventType) value.

cvat-core/src/logger-storage.ts
```diff
@@ -186,30 +168,28 @@ Object.defineProperties(LoggerStorage.prototype.save, {
...
-            const userActivityLog = logFactory(LogType.sendUserActivity, logPayload);
-            collectionToSend.push(userActivityLog);
-
             try {
                 this.saving = true;
-                await serverProxy.logs.save(collectionToSend.map((log) => log.dump()));
+                const sendTimestamp = new Date();
+                await serverProxy.logs.save({
+                    events: collectionToSend.map((log) => log.dump()),
+                    send_timestamp: sendTimestamp.toISOString(),
```
nmanovic, line 187: send_timestamp -> timestamp

cvat-core/src/server-proxy.ts
```diff
@@ -1888,7 +1888,7 @@ async function saveLogs(logs) {
     const { backendAPI } = config;
```
nmanovic, line 1888: Better to rename the function as well. In general we don't send logs now. We send events which can be exported and processed.

docker-compose.yml
```diff
@@ -227,11 +236,96 @@ services:
...
+      cvat:
+        aliases:
+          - grafana
+
 volumes:
   cvat_db:
   cvat_data:
   cvat_keys:
   cvat_logs:
+  cvat_log_db:
```
nmanovic, line 328: cvat_logs -> cvat_events. Let's rename all volumes (logs -> events)
cvat_events_db

site/content/en/docs/administration/advanced/analytics.md
```diff
@@ -13,110 +13,18 @@ description: 'Instructions for deployment and customization of Analytics. This s
 
 ![](/images/image097.jpg)
 
-It is possible to proxy annotation logs from client to ELK. To do that run the following command below:
+It is possible to proxy annotation logs from the UI to the Clickhouse database and use Grafana for visualization.
```
nmanovic, line 16: logs -> events

docker-compose.yml
```diff
@@ -204,8 +208,13 @@ services:
     ports:
       - 8080:8080
       - 8090:8090
+    environment:
+      CVAT_HOST: ${CVAT_HOST:-localhost}
+      DJANGO_LOG_VIEWER_HOST: grafana
```
nmanovic, line 213: CVAT_EVENT_VIEWER_HOST

docker-compose.yml
```diff
@@ -227,11 +236,96 @@ services:
       - --set=bundles.cvat.polling.min_delay_seconds=5
       - --set=bundles.cvat.polling.max_delay_seconds=15
 
+  clickhouse:
```
nmanovic, line 239: Need to have a well-defined pattern to name services. Now some of them have "cvat_" prefix, and some of them don't.  I don't see the logic.

docker-compose.dev.yml
```diff
@@ -67,6 +67,22 @@ services:
         socks_proxy:
       dockerfile: Dockerfile.ui
 
+  cvat_redis:
+    ports:
+      - '6379:6379'
```
nmanovic, line 72: `docker-compose -f docker-compose.yml -f docker-compose.dev.yml   up -d --build` doesn't work for me anymore because I have redis on my computer. What is a reason to export the port?

nmanovic: OK. Probably it is fine. But I don't like that in the dev file we open so many ports. It can lead to conflicts.
I have started thinking that we need analytics.yml for all new services.

nmanovic: I will recommend not opening the port, because if a user has running redis on the host, it will lead to problems.

nmanovic (commented):
If I press on the Grafana icon, I can see the login page. Is it possible to disable?

nmanovic (commented):
Need to have a solution for home dashboard. The current solution isn't user-friendly. 

https://github.com/grafana/grafana/issues/10266#issuecomment-673519617

nmanovic:
@azhavoro , I can not see payload of events
image

nmanovic:
@azhavoro , I don't understand why create:annotations event has obj_name = job, obj_id = 1, etc.
image

nmanovic:
I don't think we need send:task_info from the client. I don't understand why it provides "working time". Full statistics about the job should be extracted from the server audit logs.  

```
{"frame count": 124, "track count": 1, "object count": 1, "box count": 1, "polygon count": 0, "polyline count": 0, "points count": 0, "cuboids count": 0, "tag count": 0, "client_id": "581764", "is_active": true, "working_time": 470}
```

nmanovic:
What problems I see below:

- Server audit logs doesn't have create:job event. We should have all events. We should be able to track every change of the server state.
- I don't press fit image
- It is unclear which object I drew
- Payload for draw:object has working time.

image

nmanovic:
I will recommend don't put update:annotations and delete:annotations events if they don't have any value in the context.
Also, instead of create:annotations, I would put create:bbox and into the count field I would put the number of bounding boxes. 

image

nmanovic:
need to understand how to use obj_name, obj_id, obj_val fields properly. They should not duplicate information which we already have.

nmanovic:
For a user we should have both: id and its username

nmanovic:
update:assignee should have obj_name = user, obj_id = assignee_id, obj_val = username (or something like that)

image

nmanovic:
When a task is deleted, we don't have events delete:job in audit logs.

nmanovic:
I don't see any information about a created organization

nmanovic:
SQL query for "all events" should be something like

```
SELECT * 
FROM events 
WHERE $__timeFilter(timestamp)
ORDER BY timestamp DESC
```

remove unnecessary filters, also it is important to have DESC. Because otherwise after update you will see the first page every time.

cvat/apps/engine/log.py
```diff
@@ -171,6 +171,8 @@ class ServerLogManager(_AggregateLogManager):
```
nmanovic, line 171: Кстати, я не думаю, что нам нужно писать логи в файлы, как мы это делали (см выше). Теперь все логи должны быть в ClickHouse

cvat/apps/engine/mixins.py
```diff
@@ -339,6 +340,7 @@ class CreateModelMixin(mixins.CreateModelMixin):
     def perform_create(self, serializer, **kwargs):
         serializer.save(**kwargs)
         signal_create.send(self, instance=serializer.instance)
+        event_signal_create.send(self, instance=serializer.instance)
```
nmanovic, line 343: We already have multiple problems in WebHooks app due to own signals. Please look at the issue: https://github.com/opencv/cvat/issues/5631

Use Django DB signals only! They should be enough to track how the server state is changed in most cases.

cvat/apps/engine/task.py
```diff
@@ -162,6 +163,7 @@ def _save_task_to_db(db_task: models.Task, *, job_file_mapping: Optional[JobFile
 
         db_job = models.Job(segment=db_segment)
         db_job.save()
+        event_signal_create.send(models.Job, instance=db_job)
```
nmanovic, line 166: Use Django DB signals only: https://docs.djangoproject.com/en/4.1/ref/signals/#module-django.db.models.signals

cvat/apps/engine/views.py
```diff
@@ -1540,6 +1488,7 @@ def annotations(self, request, pk):
             if serializer.is_valid(raise_exception=True):
                 try:
                     data = dm.task.patch_job_data(pk, serializer.data, action)
+                    event_signal_annotations_patch.send(self, instance=self._object, annotations=data, action=action)
```
nmanovic, line 1491: Don't use own signals

cvat/apps/events/apps.py
```diff
@@ -0,0 +1,12 @@
+# Copyright (C) 2023 CVAT.ai Corporation
+#
+# SPDX-License-Identifier: MIT
+
+from django.apps import AppConfig
+
+
+class LogViewerConfig(AppConfig):
```
nmanovic, line 8: Need to rename the app. Think about consistency in the code.

cvat/apps/events/formatter.py
```diff
@@ -0,0 +1,14 @@
+# Copyright (C) 2023 CVAT.ai Corporation
+#
+# SPDX-License-Identifier: MIT
+
+import logging
+
+class CvatFormatter(logging.Formatter):
```
nmanovic, line 7: class EventFormatter

Could you please write a comment about why we need it? It looks dummy.

cvat/apps/iam/permissions.py
```diff
@@ -335,7 +335,8 @@ class ServerPermission(OpenPolicyAgentPermission):
     class Scopes(StrEnum):
         VIEW = 'view'
         SEND_EXCEPTION = 'send:exception'
-        SEND_LOGS = 'send:logs'
+        SEND_EVENTS = 'send:events'
+        LIST_EVENTS = 'list:events'
```
nmanovic, line 339: Probably DUMP_EVENTS = 'dump:events' is a better name.

cvat/apps/iam/rules/server.rego
```diff
@@ -30,10 +30,15 @@ allow {
...
 }
 
 allow {
     input.scope == utils.LIST_CONTENT
     utils.has_perm(utils.WORKER)
 }
+
+allow {
+    input.scope == utils.LIST_EVENTS
+    utils.has_perm(utils.BUSINESS)
```
nmanovic, line 43: Why business?

Which code is responsible for filtration? Can I get all events? IAM should decide which events I can get. For example, if I a maintainer of an organization, I can see all events of the organization. At the same time, if I'm a worker, I can see only my own events.

cvat/requirements/base.txt
```diff
@@ -53,3 +53,4 @@ dnspython==2.2.0
 setuptools==65.5.1
 django-health-check==3.17.0
 psutil==5.9.4
+clickhouse-connect==0.5.2
```
nmanovic, line 56: 0.5.7

cvat/apps/events/views.py
```diff
@@ -0,0 +1,151 @@
...
+            OpenApiParameter('query-id', location=OpenApiParameter.QUERY, type=OpenApiTypes.STR, required=False,
+                description="ID of query request that need to check or download"),
+        ],
+        responses={
+            '200': OpenApiResponse(description='Download of file started'),
+            '201': OpenApiResponse(description='CSV log file is ready for downloading'),
+            '202': OpenApiResponse(description='Creating a CSV log file has been started'),
+        })
+    @action(detail=False, methods=['GET', 'POST'])
+    def events(request):
```
nmanovic, line 67: Why do we need the action in EventsViewSet? Why don't use 'create' and 'list' methods?

cvat/apps/events/views.py
```diff
@@ -0,0 +1,151 @@
...
+            )
+
+    @staticmethod
+    @extend_schema(summary='Method saves an exception from a client on the server',
+        description='Sends logs to the ELK if it is connected',
+        request=EventSerializer, responses={
+            '201': ClientEventsSerializer,
+        })
+    @action(detail=False, methods=['POST'], serializer_class=EventSerializer)
+    def exception(request):
```
nmanovic, line 123: Do we really want to have a separate endpoint for exceptions? From my perspective is a regular event.

cvat/apps/events/views.py
```diff
@@ -0,0 +1,151 @@
...
+    @action(detail=False, methods=['GET', 'POST'])
+    def events(request):
+        if request.method == 'POST':
+            serializer = ClientEventsSerializer(data=request.data)
+            serializer.is_valid(raise_exception=True)
+
+            org = request.iam_context['organization']
+            org_id = org.id if org else None
+
+            send_time = datetime_parser.isoparse(serializer.data["send_timestamp"])
```
nmanovic, line 75: Can we adjust timestamps inside the serializer? What are pros and cons? I believe it is a part of de-serialization logic.

cvat/apps/events/views.py
```diff
@@ -0,0 +1,151 @@
...
+                        event['payload'] = json.dumps(payload)
+
+                last_timestamp = timestamp
+                event['timestamp'] = str((timestamp + time_correction).timestamp())
+                event['source'] = 'client'
+                event['organization'] = org_id
+                message = JSONRenderer().render(event).decode('UTF-8')
+                jid = event.get("job")
+                tid = event.get("task")
+                if jid:
```
nmanovic, line 100: I don't think that we need clogger at all. What do you think?

cvat/apps/events/export.py
```diff
@@ -0,0 +1,155 @@
...
+        'project': request.query_params.get('project', None),
+        'task': request.query_params.get('task', None),
+        'job': request.query_params.get('job', None),
+        'user': request.query_params.get('user', None),
+        'from': request.query_params.get('from', None),
+        'to': request.query_params.get('to', None),
+    }
+
+    if query_params['from']:
+        query_params['from'] = parser.parse(query_params['from']).timestamp()
```
nmanovic, line 99: Can we have an exception here? Should we protect ourselves from a wrong user input?

cvat/apps/events/views.py
nmanovic: Also, we need to think about the interface for long operations. For example, how to get the list of all operations which are in progress now? Let's say we reload UI, we don't know query_id anymore. How to download the CSV file?

cvat/apps/events/export.py
```diff
@@ -0,0 +1,155 @@
...
+        return output_filename
+    except Exception:
+        log_exception(slogger.glob)
+        raise
+
+def export(request, queue_name):
+    action = request.query_params.get('action', None)
+    filename = request.query_params.get('filename', None)
+
+    query_params = {
```
nmanovic, line 88: From my perspective, we should implement the functionality using DjangoFilterBackend: https://www.django-rest-framework.org/api-guide/filtering/

Please work with Maxim to understand the best solution here.
Look at the PR: https://github.com/opencv/cvat/pull/5575

docker-compose.yml
```diff
@@ -227,11 +236,96 @@ services:
...
+    environment:
+      - CLICKHOUSE_DB=cvat
+      - CLICKHOUSE_USER=user
+      - CLICKHOUSE_PASSWORD=user
+    networks:
+      cvat:
+        aliases:
+          - clickhouse
+    volumes:
+      - ./components/analytics/clickhouse/init.sh:/docker-entrypoint-initdb.d/init.sh:ro
```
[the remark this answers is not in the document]
azhavoro: discussed offline

docker-compose.yml
azhavoro: add `cvat_` prefix

docker-compose.dev.yml
azhavoro: removed

cvat/apps/engine/log.py
azhavoro: I perefere to keep server logs for first iteration

cvat/apps/iam/rules/server.rego
azhavoro: Implemented by @kirill-sizov

cvat/apps/events/views.py
azhavoro: deleted

cvat/apps/events/export.py
azhavoro: Discussed with Maxim: DjangoFilterBackend doesn't provide any advantages here due to there is not a django queryset.

cvat-ui/src/utils/controls-logger.ts
```diff
@@ -0,0 +1,58 @@
...
+        if (!target) {
+            return;
+        }
+
+        let element = (target as HTMLElement);
+        let toRecord = this.isEventToBeRecorded(element, classFilter);
+
+        const logData = {
+            text: element.innerText,
+            classes: element.className,
```
bsekachev, line 32: Need to think what kind of information we are going to record about pressed buttons. Maybe  need to add descriptive class name to all the buttons. Now not everything has it and it is not always clear what button has been pressed.

cvat-ui/src/utils/controls-logger.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    public initSave(): void {
+        if (this.#savingInterval) return;
+        this.#savingInterval = setInterval(() => {
+            core.logger.save();
+        }, CONTROLS_LOGS_INTERVAL) as unknown as number;
+    }
+
+    private isEventToBeRecorded(node: HTMLElement, filter: string[]): boolean {
+        return filter.some((cssClass: string) => node.classList.contains(cssClass));
```
bsekachev, line 54: For now the implementation is allright, but in the future we probably need to generalize it somehow

bsekachev (approved):

cvat/apps/events/signals.py
```diff
@@ -0,0 +1,83 @@
...
+    CloudStorage,
+    Issue,
+    Comment,
+    Label,
+)
+
+from .handlers import handle_update, handle_create, handle_delete
+from .event import EventScopeChoice, event_scope
+
+@receiver(pre_save, sender=Organization)
```
nmanovic, line 24: should we specify dispatch_uuid here for every receiver?

cvat/apps/events/serializers.py
```diff
@@ -0,0 +1,61 @@
...
+    obj_id = serializers.IntegerField(required=False, allow_null=True)
+    obj_val = serializers.CharField(required=False, allow_null=True)
+    source = serializers.CharField(required=False, allow_null=True)
+    timestamp = serializers.DateTimeField(required=True)
+    count = serializers.IntegerField(required=False, allow_null=True)
+    duration = serializers.IntegerField(required=False, default=0)
+    project_id = serializers.IntegerField(required=False, allow_null=True)
+    task_id = serializers.IntegerField(required=False, allow_null=True)
+    job_id = serializers.IntegerField(required=False, allow_null=True)
+    user_id = serializers.IntegerField(required=False, allow_null=True)
```
nmanovic, line 23: Let's add user_name and user_email. Thus it will be easy to look for a user in ClickHouse and Grafana.

cvat/apps/events/serializers.py
```diff
@@ -0,0 +1,61 @@
...
+    org_id = serializers.IntegerField(required=False, allow_null=True)
```
nmanovic, line 24: Let's add org_slug

nmanovic (commented):
Please look for `kibana` word and cleanup

components/analytics/grafana/dashboards/Untitled-1.sql
```diff
@@ -0,0 +1,28 @@
+
```
nmanovic, line 1: Is it possible to rename the file?

cvat-core/src/session-implementation.ts
```diff
@@ -795,7 +804,15 @@ export function implementTask(Task) {
     };
 
     Task.prototype.logger.log.implementation = async function (logType, payload, wait) {
-        const result = await loggerStorage.log(logType, { ...payload, task_id: this.id }, wait);
+        const result = await loggerStorage.log(
+            logType,
+            {
+                ...payload,
+                project: this.projectId,
```
nmanovic, line 811: project_id?

nmanovic: task_id?

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -522,25 +523,32 @@ def setUp(self):
...
-            "job_id": 1,
-            "proj_id": 2,
-            "client_id": 12321235123,
-            "message": "just test message",
-            "filename": "http://localhost/my_file.js",
-            "line": 1,
-            "column": 1,
-            "stack": ""
+            "events": [{
+            "scope": "exception",
```
nmanovic, line 527: probably the name should be unified with UI.

components/analytics/grafana/dashboards/Untitled-1.sql
azhavoro: this is tmp file, I will remove it

cvat/apps/events/export.py
```diff
@@ -0,0 +1,166 @@
...
+        ) as client:
+            result = client.query(query, parameters=parameters)
+
+        with open(output_filename, 'w', encoding='UTF8') as f:
+            writer = csv.writer(f)
+            writer.writerow(result.column_names)
+            writer.writerows(result.result_rows)
+
+        archive_ctime = os.path.getctime(output_filename)
+        scheduler = django_rq.get_scheduler(settings.CVAT_QUEUES.IMPORT_DATA.value)
```
nmanovic, line 69: Why IMPORT_DATA queue?

cvat/settings/base.py
```diff
@@ -217,6 +218,7 @@ def add_ssh_keys():
     # FIXME
     # 'corsheaders.middleware.CorsPostCsrfMiddleware',
     'django.contrib.auth.middleware.AuthenticationMiddleware',
+    'crum.CurrentRequestUserMiddleware',
```
nmanovic, line 221: We will need to remove the extension. It isn't well supported. Let's think how to do that.

cvat/apps/events/handlers.py
```diff
@@ -0,0 +1,378 @@
...
+    if isinstance(instance, User):
+        return instance
+
+    if isinstance(instance, Job):
+        return instance.segment.task.owner
+
+    return get_current_user()
+
+def user_id(instance):
+    current_user = _get_current_user(instance)
```
nmanovic, line 68: Why do we need to call _get_current_user? We handle a request. The request has an authentication user. Why should we get a user from job.segment.task.owner?

nmanovic (approved):
Let's merge the PR. I will propose to discuss some tricky moments on Monday.

PR #5657, opened 2023-02-01
UI change frame optimization

PMazarovich:
### Motivation and context

Currently user have to delete current frame and enter needed one, then press enter.
With this change, when user click on InputNumber, this field automatically clears and user don't need to do it manually.

### How has this been tested?

It was tested manually

### Checklist

### License

### DIFF
--- a/cvat-ui/src/components/annotation-page/top-bar/player-navigation.tsx
+++ b/cvat-ui/src/components/annotation-page/top-bar/player-navigation.tsx
@@ -129,7 +129,7 @@ function PlayerNavigation(props: Props): JSX.Element {
                                 setFrameInputValue(Math.floor(clamp(+value, startFrame, stopFrame)));
                             }
                         }}
-                        onClick={()=> inputFrameRef.current?.select()}
+                        onFocus={() => inputFrameRef.current?.select()}
                         onBlur={() => {
                             onInputChange(frameInputValue);
                         }}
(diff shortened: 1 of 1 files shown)


klakhov:
@PMazarovich Thanks for contribution! Can you please fix ESLint issues and add a line to [changelog](https://github.com/opencv/cvat/blob/develop/CHANGELOG.md)?

klakhov:
Also I think it would be good to add restoring frame number if new frame number is not provided by user.

bsekachev:
@PMazarovich 

Could you please clarify motivation of this pull request?
Currently, when I delete a frame, it automatically goes to the new one:
test 2

If this checkbox in settings is disabled:
image

PMazarovich:
@klakhov , I'll try to fix this
@bsekachev here is some screenshots of how this is used. It is not about deleted frames.
In second screenshot you may see, that InputNumber automatically clears, when user clicks on it. This is more convenient than deleting current frame number before inserting needed one
frst
scnd
thrd

bsekachev:
@PMazarovich 

Okay, I see. Thank you for clarification.
But I would propose another solution because I do not really like this logic with ``null`` target frame. What's more it brings some not logic things like on this demo (we are on frame 10, click in, click out, start going next/previous and it setups 1, 2, etc.)

test 3. 

Finally this solution does not work with hotkey (~) which is used to focus the input field for a frame number.

I would suggest **when users focuses the element, CVAT selects its text automatically**, using built-in methods, for example:
image

Where ``temp2`` is an input element.  Of course it just raw proposal, need to wrap it somehow in React application. 

In the interface, we see like this when focused:
image

PMazarovich:
@bsekachev , @klakhov requested changes are in place

bsekachev:
@PMazarovich 

Great, thank you for the contribution!

bsekachev (approved):

PR #5662, opened 2023-02-02
Remove collections in endpoints

zhiltsov-max:
### Motivation and context

This PR is focused on server performance optimization for requests with collections.
- Collections with user-controlled size are removed from server output and replaced 
  with structures providing links to the matching paginated endpoints and summary info:
  - /api/projects/{id}.tasks -> /api/tasks?project_id={id}
  - /api/tasks/{id}.segments -> /api/jobs?task_id={id}: .start_frame, .stop_frame
  - /api/jobs/{id}.issues -> /api/issues?job_id={id}
  - /api/issues/{id}.comments -> /api/comments?issue_id={id}
  - /api/projects | tasks | jobs/{id}.labels -> /api/labels? project_id|task_id|job_id={id}
- Added Label-related endpoints:
  - /api/labels (GET)
  - /api/labels/{id} (GET, PATCH, DELETE)
    - PATCH and DELETE both cannot be used for sublabels
- Added and updated tests
  - Removed unused test server data
  - Added several new cases with skeletons and ownerships
- Fixed possible problems with duplicate labels appearing on renaming

### Checklist

### License


Marishka17:
/api/labels?task_id=13
```
{
  "count": 18,
  "next": "http://localhost:7000/api/labels?page=2&task_id=13",
  "previous": null,
  "results": [
    {
      "id": 57,
      "name": "person",
      "color": "#c06060",
      "attributes": [],
      "type": "skeleton",
      "sublabels": [
        {
          "id": 58,
          "name": "1",
          "color": "#d12345",
          "attributes": [],
          "type": "points",
          "has_parent": true
        },
      ],
      "task_id": 13,
      "parent_id": null,
      "has_parent": false,
      "svg": "some"
    },
    {
      "id": 58,
      "name": "1",
      "color": "#d12345",
      "attributes": [],
      "type": "points",
      "sublabels": [],
      "task_id": 13,
      "parent_id": 57,
      "has_parent": true
    }
  ]
}
```
In this response information about sublabels was repeated 2 times inside the parent label and outside.
Also, why do we need to include task_id or project_id in response in each label?

zhiltsov-max:
> In this response information about sublabels was repeated 2 times inside the parent label and outside.

It's a good point. Probably, it will be better to remove this.

> Also, why do we need to include task_id or project_id in response in each label?

How would you like to change this and is there need to change this? The request cheap enough, and the query is just a list with filters.

cvat/apps/engine/serializers.py
```diff
@@ -807,20 +996,21 @@ def validate(self, attrs):
...
 class ProjectReadSerializer(serializers.ModelSerializer):
-    labels = LabelSerializer(many=True, source='label_set', partial=True, default=[], read_only=True)
     owner = BasicUserSerializer(required=False, read_only=True)
     assignee = BasicUserSerializer(allow_null=True, required=False, read_only=True)
     task_subsets = serializers.ListField(child=serializers.CharField(), required=False, read_only=True)
     dimension = serializers.CharField(max_length=16, required=False, read_only=True, allow_null=True)
     target_storage = StorageSerializer(required=False, allow_null=True, read_only=True)
     source_storage = StorageSerializer(required=False, allow_null=True, read_only=True)
-    tasks = HyperlinkedModelViewSerializer(models.Task, filter_key='project_id')
+    tasks = CollectionSummarySerializer(models.Task, url_filter_key='project_id')
```
Marishka17, line 1005: Maybe need also to define `TasksSummarySerializer` here and in other places where `CollectionSummarySerializer` is used. (just for visibility)
image

cvat/apps/engine/views.py
```diff
@@ -1867,6 +1869,121 @@ def get_serializer_class(self):
...
+        parameters=[
+            # These filters are implemented differently from others
+            OpenApiParameter('job_id', description='A simple equality filter for job id'),
+            OpenApiParameter('task_id', description='A simple equality filter for task id'),
+        ],
+        responses={
+            '200': LabelSerializer(many=True),
+        }),
+    partial_update=extend_schema(
+        summary='Methods does a partial update of chosen fields in an label',
```
Marishka17, line 1891: I guess it would be better to add information about how this endpoint can be used to delete label (when `deleted` is specified)

cvat/apps/engine/serializers.py
```diff
@@ -134,66 +179,124 @@ class Meta:
...
+
+        super().__init__(*args, **kwargs)
+
     def validate(self, attrs):
+        if self._local:
+            if attrs.get('deleted'):
+                raise serializers.ValidationError(
+                    'Labels cannot be deleted by updating in this endpoint. '
+                    'Please use the DELETE method instead.'
+                )
```
Marishka17, line 227:
```suggestion
        if self._local and attrs.get('deleted'):
            raise serializers.ValidationError(
                'Labels cannot be deleted by updating in this endpoint. '
                'Please use the DELETE method instead.'
            )
```

cvat/apps/engine/serializers.py
```diff
@@ -134,66 +179,124 @@ class Meta:
...
     def to_representation(self, instance):
         label = super().to_representation(instance)
         if label['type'] == str(models.LabelType.SKELETON):
             label['svg'] = instance.skeleton.svg
+
+        # Clean mutually exclusive fields
+        if not label.get('task_id'):
+            label.pop('task_id', None)
+        if not label.get('project_id'):
+            label.pop('project_id', None)
```
Marishka17, line 208:
```suggestion
        for field in self.__class__.Meta.extra_kwargs.keys():
            if not label.get(field):
                label.pop(field)
```

cvat/apps/engine/views.py
```diff
@@ -1867,6 +1869,121 @@ def get_serializer_class(self):
...
+        'project__organization'
+    ).all()
+
+    # NOTE: This filter works incorrectly for this view
+    # it requires task__organization OR project__organization check.
+    # Thus, we rely on permission-based filtering
+    iam_organization_field = None
+
+    search_fields = ('name', 'parent')
+    filter_fields = list(search_fields) + ['id', 'project_id', 'type', 'color', 'parent_id']
```
Marishka17, line 1924: Do we really need color in filter?

zhiltsov-max: Why not? It can be useful to find the label needed by color.

cvat/apps/engine/serializers.py
zhiltsov-max: Done

cvat/apps/engine/views.py
zhiltsov-max: I've added it to the field description.

cvat/apps/engine/serializers.py
zhiltsov-max: Done

zhiltsov-max:
@Marishka17,

> In this response information about sublabels was repeated 2 times inside the parent label and outside.

Probably, it can be done later. Currently it's unlikely to become a problem.

nmanovic:
> @Marishka17,
> 
> > In this response information about sublabels was repeated 2 times inside the parent label and outside.
> 
> Probably, the field can be removed later. Currently it's unlikely to become a problem.

I don't think that a sublabel should be in output 2 times. Looks like a bug for me.

Marishka17:
- @bsekachev, Small bug (I'm not sure whether is it related to this PR). The link is incorrect
image

- @bsekachev I have several labels in the task and I try to remove some via UI.
I see the first `DELETE` request to `labels/id`. But after that I see `PATCH` request to `tasks/tid` with payload `labels: []` Why do we need it?
image
image

- @bsekachev, @zhiltsov-max  If I first remove sublabel via `DELETE` labels/id and then open the annotation view I cannot annotate with this skeleton (the request exception with message: `"ObjectState must be provided correct Label, got wrong value undefined"` will be sent)
But If I first annotate some object with the skeleton and then remove sublabel via REST API then I cannot open the annotation view.
image

cvat/apps/engine/serializers.py
```diff
@@ -134,66 +195,124 @@ class Meta:
...
+        super().__init__(*args, **kwargs)
+
     def validate(self, attrs):
+        if self._local and attrs.get('deleted'):
+            raise serializers.ValidationError(
+                'Labels cannot be deleted by updating in this endpoint. '
+                'Please use the DELETE method instead.'
+            )
+
         if attrs.get('deleted') and attrs.get('id') is None:
```
Marishka17, line 245: @zhiltsov-max, FYI: not related to this PR but..
If I try to remove the label via `PATCH tasks/tid` request with the body
```
{
  "labels": [
    {
      "id": 101,
      "deleted": true
    }
  ]
}
``` 
image

cvat-core/src/api-implementation.ts
```diff
@@ -233,15 +236,21 @@ export default function implementAPI(cvat) {
...
         const tasks = await Promise.all(tasksData.map(async (taskItem) => {
-            // Temporary workaround for UI
-            // Fixme: too much requests on tasks page
-            let jobs = { results: [] };
             if ('id' in filter) {
-                jobs = await serverProxy.jobs.get({
+                // When request task by ID we also need to add labels and jobs to work with them
+                const labels = await serverProxy.labels.get({ task_id: taskItem.id });
+                const jobs = await serverProxy.jobs.get({
                     filter: JSON.stringify({ and: [{ '==': [{ var: 'task_id' }, taskItem.id] }] }),
```
nmanovic, line 243: Probably better to use a simple filter like `?task_id=taskItem.id`

cvat-core/src/session-implementation.ts
```diff
@@ -402,29 +377,47 @@ export function implementTask(Task) {
...
+
+            // leave only new labels to create them via project PATCH request
+            taskData.labels = (taskData.labels || [])
+                .filter((label: SerializedLabel) => !Number.isInteger(label.id)).map((el) => el.toJSON());
+
+            const serializedTask = await serverProxy.tasks.save(this.id, taskData);
+            const labels = await serverProxy.labels.get({ task_id: serializedTask.id });
             const jobs = await serverProxy.jobs.get({
-                filter: JSON.stringify({ and: [{ '==': [{ var: 'task_id' }, data.id] }] }),
+                filter: JSON.stringify({ and: [{ '==': [{ var: 'task_id' }, serializedTask.id] }] }),
```
nmanovic, line 411: Again, better to use a simple filter.

bsekachev:
@Marishka17 

> If I first remove sublabel via DELETE labels/id and then open the annotation view I cannot annotate with this skeleton (the request exception with message: "ObjectState must be provided correct Label, got wrong value undefined" will be sent)
But If I first annotate some object with the skeleton and then remove sublabel via REST API then I cannot open the annotation view.

If you removed only sublabel of a skeleton (not the skeleton itself), it makes labels specification broken. Server should not allow to do this. UI crash in this case is expected because existing annotations can not be linked with any of labels.

bsekachev:
Other comments are fixed

cvat/apps/engine/serializers.py
zhiltsov-max: This seems to require another PR to fix correctly. I've added a workaround for this. Probably, this will be resolved in #5700. We need to remove validate_labels calls in project/task serializers and try to use DB constraints instead. They seem to be broken in the case of labels for some reason.

cvat/apps/engine/views.py
```diff
@@ -1802,6 +1810,146 @@ def get_serializer_class(self):
     def perform_create(self, serializer, **kwargs):
         super().perform_create(serializer, owner=self.request.user)
 
+
+@extend_schema(tags=['labels'])
+@extend_schema_view(
+    retrieve=extend_schema(
+        summary='Method returns details of an label',
```
Marishka17, line 1817:
```suggestion
        summary='Method returns details of a label',
```

cvat/apps/engine/views.py
```diff
@@ -1802,6 +1810,146 @@ def get_serializer_class(self):
...
+            # These filters are implemented differently from others
+            OpenApiParameter('job_id', description='A simple equality filter for job id'),
+            OpenApiParameter('task_id', description='A simple equality filter for task id'),
+            OpenApiParameter('project_id', description='A simple equality filter for project id'),
+        ],
+        responses={
+            '200': LabelSerializer(many=True),
+        }),
+    partial_update=extend_schema(
+        summary='Methods does a partial update of chosen fields in an label'
```
Marishka17, line 1833:
```suggestion
        summary='Methods does a partial update of chosen fields in a label'
```

cvat/apps/engine/serializers.py
```diff
@@ -134,66 +195,129 @@ class Meta:
...
     sublabels = SublabelSerializer(many=True, required=False)
     svg = serializers.CharField(allow_blank=True, required=False)
+    has_parent = serializers.BooleanField(read_only=True, source='has_parent_label', required=False)
 
     class Meta:
         model = models.Label
-        fields = ('id', 'name', 'color', 'attributes', 'deleted', 'type', 'svg', 'sublabels', 'has_parent')
+        fields = (
+            'id', 'name', 'color', 'attributes', 'deleted', 'type', 'svg',
+            'sublabels', 'project_id', 'task_id', 'parent_id', 'has_parent'
```
Marishka17, line 208: As we discussed need to keep only one field `has_parent` or `parent_id` in the next PR (probably parent_id more informative)

Marishka17 (approved):

PR #5700, opened 2023-02-14
Fix /api/tasks leads to 500 server error

yasakova-anastasia:
### Motivation and context

- Fixed re-deleting of skeleton sublabels (attempt to remove already removed sublabels)

Screenshot from 2023-02-13 10-00-03

- Updated Label model (added `project` to `unique_together`)
- Added migration to remove or rename wrong labels

### How has this been tested?

REST API tests

### Checklist

### License


cvat/apps/engine/serializers.py
```diff
@@ -687,12 +687,13 @@ def update_labels(labels, parent_label=None):
...
-                        svg = svg.replace(f'data-label-name="{db_sublabel.name}"', f'data-label-id="{db_sublabel.id}"')
-                    db_skeleton = models.Skeleton.objects.create(root=db_label, svg=svg)
-                    logger.info(f'label:update Skeleton id:{db_skeleton.id} for label_id:{db_label.id}')
+                if not label.get('deleted'):
+                    update_labels(sublabels, parent_label=db_label)
+                    if label.get('id') is None and db_label.type == str(models.LabelType.SKELETON):
+                        for db_sublabel in list(db_label.sublabels.all()):
+                            svg = svg.replace(f'data-label-name="{db_sublabel.name}"', f'data-label-id="{db_sublabel.id}"')
+                        db_skeleton = models.Skeleton.objects.create(root=db_label, svg=svg)
+                        logger.info(f'label:update Skeleton id:{db_skeleton.id} for label_id:{db_label.id}')
```
Marishka17, line 696: Now we have 2 functions `update_labels` in `TaskWriteSerializer` and `ProjectWriteSerializer` and in one function was this problem. Can we remove code duplication to avoid such problems?

zhiltsov-max: This will be done in #5662

cvat/apps/engine/serializers.py
```diff
@@ -148,52 +151,94 @@ def to_representation(self, instance):
...
         return label
 
+    def __init__(self, *args, **kwargs):
+        self._local = kwargs.pop('local', False)
+        """
+        Indicates that the operation is called from the dedicated ViewSet
+        and not from the parent entity, i.e. a project or task.
+        """
+
+        super().__init__(*args, **kwargs)
```
zhiltsov-max, line 161: Please remove parts about the `_local` variable - they are not relevant to this PR.

cvat/apps/engine/serializers.py
```diff
@@ -148,52 +151,94 @@ def to_representation(self, instance):
...
+
     def validate(self, attrs):
+        if self._local:
+            if attrs.get('deleted'):
+                raise serializers.ValidationError(
+                    'Labels cannot be deleted by updating in this endpoint. '
+                    'Please use the DELETE method instead.'
+                )
+
```
zhiltsov-max, line 170:
```suggestion
```

cvat/apps/engine/serializers.py
```diff
@@ -207,8 +252,138 @@ def update_instance(validated_data, parent_instance, parent_label=None):
...
+        elif isinstance(instance.task, models.Task):
+            parent_serializer = TaskWriteSerializer(
+                instance=instance.task, data=parent_query, partial=True,
+            )
+
+        parent_serializer.is_valid(raise_exception=True)
+        parent_serializer.save()
+
+        self.instance = models.Label.objects.get(pk=instance.pk)
+        return self.instance
```
zhiltsov-max, line 384:
```suggestion
```

cvat/apps/engine/serializers.py
yasakova-anastasia: Removed

yasakova-anastasia:
> Do we need to check anything in the existing DB or add a migration to fix existing labels?

I added a migration that removes duplicate labels and skeleton labels without skeletons.

cvat/apps/engine/models.py
```diff
@@ -514,7 +514,8 @@ def has_parent_label(self):
 
     class Meta:
         default_permissions = ()
-        unique_together = ('task', 'name', 'parent')
+        unique_together = (('task', 'name', 'parent'),
+                           ('project', 'name', 'parent'))
```
zhiltsov-max, line 518: During experimenting, I've noticed that this constraint doesn't work for some reason. I've tried the following variants:

```
unique_together = (('task', 'name', 'parent'), ('project', 'name', 'parent'))
unique_together = (('task_id', 'name', 'parent_id'), ('project_id', 'name', 'parent_id'))
unique_together = ('project', 'task', 'name', 'parent')
unique_together = ('project_id', 'task_id', 'name', 'parent_id')
unique_together = ('name', 'parent')
unique_together = ('name', 'parent_id')
constraints = [models.UniqueConstraint(fields=...all the same..., name=...)]

unique_together = ('name', ) # only this worked
```
Only the simplest one worked in my case. If `validate_labels()` checks are removed in `Task/Project- WriteSerializer`, users are able to create several equal labels. This problem doesn't appear in other cases where `unique_together` used. SO tells it can be because of the DB used, but both with MySQL and Postgres the problem appears.

The test case (remember to remove `validate_labels`):
POST /api/projects

```
{
  "name": "test_duplicate_labels",
  "labels": [{"name": "cat"}, {"name": "cat"}, {"name": "cat"}]
}
```
-> OK, 3 same labels in a project.

Marishka17: I guess it can happen because we have 3 ForeignKey which can be null. To solve the problem with UniqueConstraint, all permutations should be specified (but it's too hard, may be we can use `UniqueTogetherValidator`).
```
constraints = [
    models.UniqueConstraint(
        name='project_name_parent_unique',
        fields=('project', 'name', 'parent'),
        condition=Q(task__isnull=True)
    ),
    ...
    models.UniqueConstraint(
        name='name_unique',
        fields=('name',),
        condition=Q(project__isnull=True) & Q(task__isnull=True) & Q(parent__isnull=True)
    )
]
```

zhiltsov-max: Is there any mentioning in the documentation about problems with `null`? 
For instance, this tuple should work normally even with nulls:

```
unique_together = ('project', 'task', 'name', 'parent')
```

because it makes unique hashes for each label.

Marishka17: In SQL null is not equal to anything - not even to another null

zhiltsov-max: Okay, probably, this resolves the question then.

zhiltsov-max:
@yasakova-anastasia, please remove code that is obsolete now: 
https://github.com/opencv/cvat/blob/develop/cvat/apps/engine/models.py#L523-L540
https://github.com/opencv/cvat/blob/develop/cvat/apps/engine/serializers.py#L993-L1002
https://github.com/opencv/cvat/blob/develop/cvat/apps/engine/serializers.py#L1093-L1103

The checks are not completely valid, and this logic should be enforced by the DB now.

cvat/apps/engine/serializers.py
```diff
@@ -287,12 +287,15 @@
...
-            )
+            try:
+                db_label = models.Label.objects.create(
+                    name=validated_data.get('name'),
+                    type=validated_data.get('type'),
+                    parent=parent_label,
+                    **parent_info
+                )
+            except IntegrityError as exc:
+                raise exceptions.ValidationError(str(exc)) from exc
```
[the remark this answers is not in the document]
zhiltsov-max: @yasakova-anastasia, probably, now these warnings are actually valid. I suggest to check for the message and forward only if the message is safe. Please also remove or reuse the `InvalidLabel` exception. Another option is that we override the model create/save methods and add this try-catch to reraise `InvalidLabel`. This second variant can even provide better encapsulation.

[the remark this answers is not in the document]
yasakova-anastasia: @zhiltsov-max, I overridden the `save` method and added the `create` method. I moved the try-catch logic to these methods. And now an error message about the uniqueness of labels is displayed instead of an `IntegrityError` message. Could you please review again?

cvat/apps/engine/models.py
```diff
@@ -520,6 +519,24 @@ def __str__(self):
...
         return bool(self.parent)
 
+    def save(self, force_insert=False, force_update=False, using=None,
+             update_fields=None):
+        try:
+            super().save(force_insert, force_update, using, update_fields)
+        except ValueError as exc:
+            raise exceptions.ValidationError(str(exc)) from exc
+        except IntegrityError:
+            raise exceptions.ValidationError("All label names must be unique")
```
zhiltsov-max, line 529:
```suggestion
    def save(self, force_insert=False, force_update=False, using=None,
             update_fields=None):
        try:
            super().save(force_insert, force_update, using, update_fields)
        except IntegrityError as exc:
            raise InvalidLabel("All label names must be unique") from exc
```

I suggest to change both methods this way:
- This will encapsulate model logic about constraints in this class (InvalidLabel will need to be catched from the calling side)
- ValueError here means the data saved was incorrect, so it's an internal serializer error, not a user error

cvat/apps/engine/migrations/0064_delete_wrong_labels.py
```diff
@@ -0,0 +1,39 @@
...
+    migration_name = os.path.splitext(os.path.basename(__file__))[0]
+    with get_migration_logger(migration_name) as log:
+        log.info('\nDeleting skeleton Labels without skeletons...')
+
+        Label = apps.get_model('engine', 'Label')
+        for label in Label.objects.all():
+            if label.type == "skeleton" and not hasattr(label, "skeleton"):
+                label.delete()
+
+        log.info('\nDeleting duplicate Labels...')
```
zhiltsov-max, line 16: Can such tasks be annotated already? If so, I suggest to rename the duplicate labels instead.

zhiltsov-max (commented):
Please add `@transaction.atomic` to the `update_label()` too.

cvat/apps/engine/models.py
yasakova-anastasia: Updated

cvat/apps/engine/migrations/0064_delete_wrong_labels.py
yasakova-anastasia: The labels may have already been used, I added a rename for duplicates. I verified that it is not possible to add a skeletal annotation with repeated nested labels, so I suggest deleting such skeletons.

yasakova-anastasia:
> Please add `@transaction.atomic` to the `update_label()` too.

Added.

I noticed that if the project creation fails with a unique labels error, the project is still created (without labels). I'm not sure this is correct. I think in this case the project should not be created. So I suggest adding `@transaction.atomic to the create and update methods for the task and project serializers.

zhiltsov-max:
This looks like the issue (https://github.com/opencv/cvat/issues/5750). Yes, this definitely looks invalid to me. I agree, we should add it, and also check if we can benefit from using [`ATOMIC_REQUESTS`](https://docs.djangoproject.com/en/dev/topics/db/transactions/#tying-transactions-to-http-requests) (it's stated that it can reduce performance).

yasakova-anastasia:
I added `@transaction.atomic`. I think we can replace this with `ATOMIC_REQUESTS` in another PR.

zhiltsov-max (approved):

cvat/apps/engine/migrations/0064_delete_wrong_labels.py
zhiltsov-max: Could you please also rename the file, the function, and update the log message?

nmanovic (approved):

Issue #5722, opened 2023-02-17
Can't install 2.3.0 helm because of missing dependency


gecube:
Hi!

I am using flux for installing your nice CVAT installation.

To do it I am using something like:

```yaml
---
apiVersion: v1
kind: Namespace
metadata:
  name: cvat
---
apiVersion: source.toolkit.fluxcd.io/v1beta2
kind: GitRepository
metadata:
  name: cvat
  namespace: cvat
spec:
  interval: 5m0s
  url: https://github.com/opencv/cvat
  ref:
    tag: v2.3.0
---
apiVersion: helm.toolkit.fluxcd.io/v2beta1
kind: HelmRelease
metadata:
  name: cvat
  namespace: cvat
spec:
  chart:
    spec:
      chart: ./helm-chart
      sourceRef:
        kind: GitRepository
        name: cvat
  targetNamespace: cvat
  releaseName: cvat
  interval: 5m0s
  timeout: 10m
  values:
    postgresql:  
      global:
        storageClass: gp2
      secret:
        password: ****
        postgres_password: ****
        replication_password: ****
    ingress:
      enabled: true
      annotations:
        cert-manager.io/cluster-issuer: letsencrypt-http01-nginx-issuer
        nginx.ingress.kubernetes.io/proxy-body-size: 1024m
      className: nginx
      hosts:
        - host: *****
          paths:
            - path: /api
              pathType: "Prefix"
              service:
                name: backend-service
                port: 8080
            - path: /admin
              pathType: "Prefix"
              service:
                name: backend-service
                port: 8080
            - path: /static
              pathType: "Prefix"
              service:
                name: backend-service
                port: 8080
            - path: /django-rq
              pathType: "Prefix"
              service:
                name: backend-service
                port: 8080
            - path: /git
              pathType: "Prefix"
              service:
                name: backend-service
                port: 8080
            - path: /opencv
              pathType: "Prefix"
              service:
                name: backend-service
                port: 8080
            - path: /profiler
              pathType: "Prefix"
              service:
                name: backend-service
                port: 8080
            - path : /
              pathType: "Prefix"
              service:
                name: frontend-service
                port: 80
      tls:
        - hosts:
          - ****
          secretName: ingress-tls-cvat
```

It is perfectly the same as to use helm install directly.
And I am gotting the error:

```
[...]

PR #5732, opened 2023-02-18
Helm: fix postgres helm chart version

gecube:
This PR is intended to solve issue the #5722


nmanovic:
@gecube , thanks for the contribution! I will close the PR because we should not modify the original release v2.3.0. Could you please prepare the patch for develop branch?

PR #5760, opened 2023-02-23
Allow typed simple filters

zhiltsov-max:
### Motivation and context

This PR implements typed query params for simple filters, making SDK interface clearer.

### Checklist

### License

### DIFF
--- a/cvat/apps/engine/filters.py
+++ b/cvat/apps/engine/filters.py
@@ -7,8 +7,10 @@ from typing import Any, Dict, Iterator, Optional
 from functools import reduce
 import operator
 import json
+import warnings
 
 from django_filters import FilterSet
+from django_filters import filters as djf
 from django_filters.filterset import BaseFilterSet
 from django_filters.rest_framework import DjangoFilterBackend
 from django.db.models import Q
@@ -283,7 +285,7 @@ class SimpleFilter(DjangoFilterBackend):
 
     def get_filterset_class(self, view, queryset=None):
         lookup_fields = self.get_lookup_fields(view)
-        if not lookup_fields or not queryset:
+        if not lookup_fields or queryset is None:
             return None
 
         MetaBase = getattr(self.filterset_base, 'Meta', object)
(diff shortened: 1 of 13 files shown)


cvat/apps/engine/filters.py
```diff
@@ -306,33 +308,38 @@ def get_lookup_fields(self, view):
...
-
-        lookup_fields = self.get_lookup_fields(view)
+    def get_schema_operation_parameters(self, view):
+        try:
+            queryset = view.get_queryset()
+        except Exception:
+            queryset = None
+            warnings.warn(
+                "{} is not compatible with schema generation".format(view.__class__)
+            )
```
SpecLad, line 318: Are there any views left that are incompatible with schema generation? Why couldn't they be adapted?

cvat/apps/engine/filters.py
```diff
@@ -306,33 +308,38 @@ def get_lookup_fields(self, view):
...
         parameters = []
-        for field_name in lookup_fields:
-            parameters.append({
+        for field_name, field in filterset_class.base_filters.items():
+            if isinstance(field, djf.BooleanFilter):
+                parameter_schema = { 'type': 'boolean' }
+            elif isinstance(field, (djf.NumberFilter, djf.ModelChoiceFilter)):
+                parameter_schema = { 'type': 'integer' }
+            else:
+                parameter_schema = { 'type': 'string' }
```
SpecLad, line 332: Perhaps we should check that the field is actually a string here? If we're doing typing, it would be useful to check that the type we're applying is appropriate.

cvat/apps/engine/filters.py
zhiltsov-max: Probably, there are no such views and we can tighten this requirement.

cvat/apps/engine/filters.py
zhiltsov-max: Probably, these types are everything we need to support in simple filters. Basically, all fields must be serialized into strings, but we do an exception for numbers. Which other types would you like to support? To my knowledge, there are only dates that could be supported, but there are no such filters now. What is possible is that the model pk is not an integer, this is a good point.

SpecLad: My point is not that we should support other types, but that when we declare a parameter to be a string, we should check that it actually is a string. If you add an assert for that, then in the future, when someone adds a filter of a type that hasn't been used before, they will be notified that they need to update the schema generation too.

zhiltsov-max: Updated

SpecLad (approved):

PR #5803, opened 2023-03-02
Remove authentication token usage from the UI

SpecLad:
### Motivation and context

From what I can see, the token confers no benefits on top of the session cookie (which we are already using), and it has drawbacks:

* Logging out in one session deletes the token from the DB, which effectively invalidates all other sessions too.

* It's accessible from JavaScript, so it could be stolen by an XSS attack (in comparison, the session cookie is HTTP-only).

* It requires custom code, whereas cookies work automatically.

The only annoying thing about using session cookies is that they require CSRF protection, but Axios handles that for us. We only need to do it manually for TUS uploads.

Move the origin concatenation to `chunkUpload` to make sure that it cannot be used to send our CSRF token to some other origin.

### How has this been tested?

I manually checked that the UI still works, including TUS uploads.

### Checklist

### License


SpecLad:
CC @klakhov

SpecLad:
One other disadvantage of the authentication token is that it lacks any lifetime mechanism, so it lasts forever. Session cookies, on the other hand, have lifetimes both on the server (implemented by Django) and on the client.

PR #5818, opened 2023-03-07
Add tests for getting swagger docs

kirill-sizov:
### Checklist

### License

### DIFF
--- a/tests/python/rest_api/test_server.py
+++ b/tests/python/rest_api/test_server.py
@@ -6,8 +6,7 @@
 from http import HTTPStatus
 
 import pytest
-
-from shared.utils.config import make_api_client, get_method
+from shared.utils.config import get_method, make_api_client
 
 
 @pytest.mark.usefixtures("restore_db_per_class")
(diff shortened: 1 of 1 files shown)


tests/python/rest_api/test_server.py
```diff
@@ -36,3 +36,10 @@ def test_can_get_schema_unauthorized(self):
 
             assert response.status == HTTPStatus.OK
             assert data
+
+
+@pytest.mark.usefixtures("restore_db_per_class")
+class TestGetSwaggerDocs:
+    def test_can_get_swagger_docs(self):
+        response = get_method("admin1", "/swagger")
```
Marishka17, line 44: I guess you should check the `/schema` endpoint because the request to `/swagger` will pass with 200 status.

kirill-sizov: Test for `/schema` endpoint already implemented in this file

Marishka17: Ok, but I thought that this test should check if "normal" documentation is rendered. If our schema on the server is not valid, then this test will pass... Why do we need a such test in this case?

kirill-sizov: We need this test such case when schema built successfully, but swagger documentation built with error

kirill-sizov: Yes, you right

PR #5830, opened 2023-03-09
Make traefik optional with traefik.enabled value

### LINKED ISSUE #5733: Can't disable traefik ingress controller without removing ingress object
Hi!

We have the next use case: we are using k8s with already installed nginx ingress controller. In fact it can be any ingress controller of our choice. As well - it may be already installed traefik. Right now I can't disable installation of traefik embedded into CVAT helm chart because in that case Ingress object won't be created. 

https://github.com/opencv/cvat/blob/27224f25967d5e13ade686c4dc7f1bd363415d74/helm-chart/Chart.yaml#L61
https://github.com/opencv/cvat/blob/27224f25967d5e13ade686c4dc7f1bd363415d74/helm-chart/values.yaml#L289

### The expected behaviour:

to have two options in helm chart:

1. disable of Ingress object itself
2. disable of embedded Traefik ingress controller in case when customer doesn't want it.

My proposal is to make another enabled option under traefik block in values.yaml:

```yaml
traefik:
  enabled: true # default value if you want to install Traefik by default.
  logs:
    general:
      level: INFO
  service:
    externalIPs:
      # - "192.168.49.2"
```

and to change traefik block in Chart.yaml:

```yaml
  - name: traefik
    version: 10.24.0
    condition: traefik.enabled
``` 

The change is trivial, so I may create a PR for it.

ruckc:
### Motivation and context
Closes #5733. If you have an existing kubernetes cluster with an existing ingress controller, CVAT may not work properly.  This makes adds a value, `traefik.enabled` which can toggle traefik's subchart installation, instead of using `ingress.enabled`.

### How has this been tested?
Tested my branch in my kubernetes environment where I had an existing traefik installation that was conflicting with the cvat provided one.  The error was `0/1 nodes are available: 1 node(s) didn't have free ports for the requested pod ports.`

### Checklist

- [X] I have linked related issues (see [GitHub docs](
- [X] I have increased versions of npm packages if it is necessary

### License


ruckc:
@nmanovic / @azhavoro - Can we get this merged... its frustrating having to live on a branch

PR #5889, opened 2023-03-20
Fix most discrepancies between the different ways of generating the API schema

SpecLad:
### Motivation and context

Currently, if you try to generate the API schema with the VS Code launch configuration, it'll remove the `minimum` and `maximum` attributes from the definition of some of the integer fields. This is because those attributes come from the configured database engine in use. The "canonical" version of the schema is generated from the Docker image, which uses the production configuration, and thus PostgreSQL; while using the VS Code launch configuration will default to the development configuration, and thus SQLite.

Fix it by aligning the configurations. Since ultimately, the schema is for the production version of the server, override the configuration to production for the VS Code launch configuration.

Note that there's one more discrepancy that remains; that of the API version number. I haven't figured out the best way to fix that yet.

### How has this been tested?

By pressing F5 🙂 

### Checklist

### License

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -393,7 +393,7 @@
             ],
             "django": true,
             "cwd": "${workspaceFolder}",
-            "env": {},
+            "env": {"DJANGO_CONFIGURATION": "production"},
             "console": "internalConsole"
         },
         {
--- a/site/content/en/docs/contributing/development-environment.md
+++ b/site/content/en/docs/contributing/development-environment.md
@@ -11,7 +11,7 @@ description: 'Installing a development environment for different operating syste
   Ubuntu 18.04
 
   ```bash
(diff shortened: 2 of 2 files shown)


SpecLad:
CC @Marishka17  @zhiltsov-max

zhiltsov-max:
Probably, this should not depend on the database used, but I agree with this as a quick fix. However, there seem to be a problem with dependencies in the development environment because Apache is not installed.

zhiltsov-max:
No, it fails to compile a package because Apache is required.

nmanovic (dismissed):

SpecLad:
> No, it fails to compile a package because Apache is required. I understand that it needs to be installed, but I suggest to add such notes to the development environment configuration steps then.

Fair point. I did it.

PR #5892, opened 2023-03-21
Open resource links from any organization/sandbox

klakhov:
### Motivation and context
**Server part:**
- [x] Update swagger docs
- [x] For object requests: define iam_context according to the obj.organization_id
- [x] For list and create requests: iam_context is defined according to the query parameters `org_id` or `org` (slug)
- [x] Fix tests

**UI part:**
- [x] Change organization when link leads to resource in different organization 
- [x] Fix tests
### Checklist

### License


bsekachev (approved):

tests/python/rest_api/test_resource_import_export.py
```diff
@@ -268,20 +268,17 @@ def test_import_resource_from_cloud_storage_with_default_location(
...
         [
-            ("projects", "annotations"),
-            ("projects", "dataset"),
-            ("projects", "backup"),
-            ("tasks", "annotations"),
-            ("tasks", "dataset"),
-            ("tasks", "backup"),
-            ("jobs", "annotations"),
-            ("jobs", "dataset"),
+            ("projects", "dataset", HTTPStatus.NOT_FOUND),
```
nmanovic, line 273: Why the number of use cases was reduced?

cvat/apps/iam/permissions.py
```diff
@@ -50,6 +50,53 @@ class PermissionResult:
...
+            try:
+                return Organization.objects.get(id=organization_id)
+            except Organization.DoesNotExist:
+                return None
+        return None
+
+    return request.iam_context["organization"]
+
+def get_privilege(request):
+    # TO-DO: refactor it
```
nmanovic, line 69: Do you want to refactor in the PR?

tests/python/rest_api/test_resource_import_export.py
kirill-sizov: These tests don't make sense because we don't have endpoints for importing job datasets, importing project annotations, or importing task datasets, so I removed them.

kirill-sizov: These tests don't make sense, because we don't have endpoints for importing job datasets, importing task datasets, or importing project annotations

cvat/apps/iam/filters.py
```diff
@@ -25,7 +25,18 @@ def filter_queryset(self, request, queryset, view):
         # filter isn't necessary but it is an extra check that we show only
         # objects inside an organization if the request in context of the
         # organization.
-        visibility = request.iam_context['visibility']
+        visibility = None
```
nmanovic, line 28: Could you please check if we need OrganizationFilterBackend at all?

cvat/apps/iam/views.py
```diff
@@ -86,7 +67,7 @@ def __init__(self, get_response):
     def __call__(self, request):
```
nmanovic, line 67: Do we need the middleware at all? Should we delete it?

tests/python/rest_api/test_jobs.py
```diff
@@ -83,7 +83,7 @@ def test_admin_get_job(self, jobs, tasks, org):
         for job in jobs[:8]:
             self._test_get_job_200("admin2", job["id"], job, **kwargs)
 
-    @pytest.mark.parametrize("org_id", ["", None, 1, 2])
+    @pytest.mark.parametrize("org_id", [""])
```
nmanovic, line 86: Please use the same set of orgs here.

nmanovic: Need to be sure that our test system covers primary use cases connected with changes in the PR.

kirill-sizov: Fixed

cvat/apps/iam/views.py
kirill-sizov: This middleware allows us to quickly access specified org_id in requests and also efficiently perform "create" actions like [here](https://github.com/opencv/cvat/blob/89e1b21b5b4d6a684c2d1ff1491ab48ac4a5797a/cvat/apps/engine/views.py#L265).

cvat/apps/iam/filters.py
kirill-sizov: We still need this filter, since this functionality cannot be covered by SimpleFilter. The reason of it the following:
to show valid lists in UI we want to return all non-org objects for example by requesting GET /tasks and only org-related objects with GET /tasks?org=org. And OrganizationFilterBackend filter does this.

nmanovic:
@kirill-sizov , job cannot be found if I'm not inside an organization. Is it expected?

tests/python/shared/assets/cvat_db/data.json
```diff
@@ -525,6 +525,14 @@
     "expire_date": "2022-10-31T17:09:16.915Z"
   }
 },
+{
```
nmanovic, line 528: Do we need sessions in DB?

tests/python/sdk/test_client.py
```diff
@@ -235,11 +235,10 @@ def test_organization_contexts(admin_user: str):
         client.projects.retrieve(personal_project.id)
         client.projects.retrieve(org_project.id)
 
-        # only the personal project should be visible in the personal workspace
+        # retrieve personal and org projects by id
         client.organization_slug = ""
         client.projects.retrieve(personal_project.id)
-        with pytest.raises(NotFoundException):
-            client.projects.retrieve(org_project.id)
+        client.projects.retrieve(org_project.id)
```
nmanovic, line 241: Need tests for all objects: projects, tasks, jobs, cloud storages, etc.

kirill-sizov:
> @kirill-sizov , job cannot be found if I'm not inside an organization. Is it expected?

Yes, this is a bug. The server returns 404 for the request `GET jobs/{id}/data/meta?org`. So I fixed it. Now query parameter will be ignored for detailed requests. Such ignoring also promotes backward compatibility (if such users already use our API with using `org` parameters this patch won't break their pipelines)

cvat/schema.yml
```diff
@@ -349,10 +229,6 @@ paths:
       operationId: cloudstorages_list
       summary: Returns a paginated list of storages
       parameters:
-      - in: header
-        name: X-Organization
-        schema:
-          type: string
```
SpecLad, line 355: Why is the header removed here?

cvat/schema.yml
```diff
@@ -385,7 +261,7 @@ paths:
       - in: query
         name: org_id
         schema:
-          type: integer
+          type: string
```
SpecLad, line 264: This doesn't seem right. `org_id` should be an integer.

cvat/schema.yml
kirill-sizov: Fixed

cvat/schema.yml
kirill-sizov: Fixed

kirill-sizov:
@klakhov Could you please check [this](https://github.com/opencv/cvat/actions/runs/4862625399/jobs/8669526195?pr=5892) test it seems that this step `Logout, the third user login. The user does not see the project, the task. The user can't open the job using direct link.` is not valid anymore?

SpecLad:
I think there are a few more API endpoints that don't need the `X-Organization`/`org`/`org_id` parameters:

* `events_create`
* `organizations_list`
* `users_list`
* `webhooks_list_deliveries`

Could you double-check and remove those parameters if they are indeed unnecessary?

tests/python/sdk/test_client.py
```diff
@@ -235,17 +234,15 @@ def test_organization_contexts(admin_user: str):
...
 
-        # only the organizational project should be visible in the organization
+        # org context doesn't make sense for detailed request
         client.organization_slug = org.slug
         client.projects.retrieve(org_project.id)
-        with pytest.raises(NotFoundException):
-            client.projects.retrieve(personal_project.id)
+        client.projects.retrieve(personal_project.id)
```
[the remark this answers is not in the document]
kirill-sizov: This test check that we can retrieve even if org_slug query parameter is specified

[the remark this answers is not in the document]
SpecLad: Okay, I suppose there is some merit to a test like this.

Still, we should test that setting the org context affects what it's supposed to affect (such as list results).

kirill-sizov:
> I think there are a few more API endpoints that don't need the `X-Organization`/`org`/`org_id` parameters:
> 
> * `events_create`
> * `organizations_list`
> * `users_list`
> * `webhooks_list_deliveries`
> 
> Could you double-check and remove those parameters if they are indeed unnecessary?

Yes, thanks for your attention. I have fixed this by applying such strategy: if view is not detailed and view has iam_organization_field then this endpoints will have `org` parameters.  

`org` for endpoint `users_list` helps to get all members in organization
`org` for endpoints `events_create` needs for IAM (as for others POST endpoints)

tests/python/sdk/test_client.py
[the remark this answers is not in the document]
kirill-sizov: Such tests are implemented in the `tests/python/rest_api` directory for each endpoint

kirill-sizov:
> 
> > `org` for endpoints `events_create` needs for IAM (as for others POST endpoints)
> 
> Why? I don't think events belong to organizations.

Events associated with organizations. We manage access to events based on which organization particular events belong.
It's a unified approach, currently, we use the org_id parameter for POST requests for each object that has an `organization` attribute. I see no reason why we should use a different strategy for events.

SpecLad (commented):
Sorry, I just saw that I had a bunch of pending comments that I forgot to publish...

tests/python/sdk/test_client.py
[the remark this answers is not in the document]
kirill-sizov: Added SDK tests that check the possibility of organization filtering for main endpoints

cvat/apps/iam/filters.py
nmanovic: Why cannot SimpleFilter cover the use case? Could you please explain?

kirill-sizov: It requires additional refactoring of SimpleFilter in order to support not only query parameters but only headers ("X-Organization")

tests/python/sdk/test_client.py
```diff
@@ -235,17 +234,56 @@ def test_organization_contexts(admin_user: str):
...
+        org = client.organizations.create(models.OrganizationWriteRequest(slug="testorg"))
+
+        # create a project and task in sandbox
+        client.organization_slug = None
+        client.projects.create(models.ProjectWriteRequest(name="personal_project"))
+        client.tasks.create_from_data(
+            spec={"name": "personal_task", "labels": [{"name": "a"}]}, resources=[fxt_image_file]
+        )
+
+        # create a project and in the organization
```
SpecLad, line 260:
```suggestion
        # create a project and task in the organization
```

tests/python/sdk/test_client.py
```diff
@@ -235,17 +234,56 @@ def test_organization_contexts(admin_user: str):
...
+        )
+
+        # return only non-org objects if org parameter is empty
+        client.organization_slug = ""
+        projects, tasks, jobs = client.projects.list(), client.tasks.list(), client.jobs.list()
+
+        assert len(projects) == len(tasks) == len(jobs) == 1
+        assert projects[0].organization == tasks[0].organization == jobs[0].organization == None
+
+        # return only all objects if org parameter wasn't presented
```
SpecLad, line 274:
```suggestion
        # return all objects if org parameter wasn't presented
```

CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ without use_cache option (<https://github.com/opencv/cvat/pull/6074>)
 - Support task creation with cloud storage data and without use_cache option (<https://github.com/opencv/cvat/pull/6074>)
 
 ### Changed
+- Resource links are opened from any organization/sandbox if available for user (<https://github.com/opencv/cvat/pull/5892>)
+- Running SAM masks decoder on frontend (<https://github.com/opencv/cvat/pull/6019>)
```
nmanovic, line 18: @kirill-sizov , probably the line is wrong. Could you please fix?

nmanovic:
I got the error below when I requested a task.

```
Could not receive the requested project from the server
Error: Request failed with status code 403. {"detail":"You do not have permission to perform this action."}.
```

nmanovic:
@kirill-sizov , /api/jobs request doesn't work in debug at least. Could you please look?

Screenshot from 2023-05-23 20-36-04

kirill-sizov:
@nmanovic I fixed the performance problems and with getting jobs. But confused about this issue:

> I got the error below when I requested a task.
> 
> ```
> Could not receive the requested project from the server
> Error: Request failed with status code 403. {"detail":"You do not have permission to perform this action."}.
> ```

I have checked accessing tasks with different parameters of privileges and organization roles (this is also covered by our test) and couldn't reproduce this problem. Could please clarify how did you get this error?

Did you get this error in the development instance or in docker?
Does this task belong to the organization or project?
Which privilege does the requester have?
Which organization role does the requester have?
Does the requester assign to the task or corresponding project?

nmanovic (approved):

cvat/apps/iam/filters.py
```diff
@@ -2,32 +2,62 @@
...
-            return queryset
+
+        return queryset
+
+    def get_schema_operation_parameters(self, view):
+        if not view.iam_organization_field or view.detail:
+            return []
+
+        return [
+            {
```
zhiltsov-max, line 45: Shouldn't these values be aligned with the `ORGANIZATION_OPEN_API_PARAMETERS` somehow? The code can easily diverge.

cvat/apps/engine/schema.py
```diff
@@ -228,4 +229,27 @@ class CloudStorageReadSerializerExtension(_CloudStorageSerializerExtension):
 class CloudStorageWriteSerializerExtension(_CloudStorageSerializerExtension):
     target_class = 'cvat.apps.engine.serializers.CloudStorageWriteSerializer'
 
+ORGANIZATION_OPEN_API_PARAMETERS = [
```
zhiltsov-max, line 232: How and when this should be used? Is it for backward compatibility only?

cvat/apps/iam/filters.py
```diff
@@ -2,32 +2,62 @@
...
 
     def filter_queryset(self, request, queryset, view):
-        # Rego rules should filter objects correctly (see filter rule). The
-        # filter isn't necessary but it is an extra check that we show only
-        # objects inside an organization if the request in context of the
-        # organization.
-        visibility = request.iam_context['visibility']
-        if visibility and view.iam_organization_field:
+        # Filter works only for "list" requests and allows to return
+        # only non-organization objects if org isn't specified
```
zhiltsov-max, line 16: Probably, these parameters should be documented somehow, especially the (past?) logic about org=''/org=None etc.

cvat/apps/iam/schema.py
```diff
@@ -107,35 +78,6 @@ def get_security_definition(self, auto_schema):
         return [sessionid_schema, csrftoken_schema]
 
 class CustomAutoSchema(AutoSchema):
```
zhiltsov-max, line 80: This class can be moved to engine if there are no IAM specifics added. The only reason it is here is that we can't define it in the engine and import here.

cvat/apps/events/handlers.py
```diff
@@ -58,10 +58,7 @@ def organization_id(instance):
         return instance.id
 
     try:
-        oid = getattr(instance, "organization_id", None)
-        if oid is None:
-            return instance.get_organization_id()
-        return oid
+        return getattr(instance, "organization_id", None)
```
zhiltsov-max, line 61: Should this attr be enforced on the objects somehow? It's easy to miss it and it won't be noticed.

zhiltsov-max (commented):
Do you know if we can do anything about places like this https://github.com/opencv/cvat/blob/develop/cvat/apps/engine/views.py#L1832 ?

zhiltsov-max (commented):
Probably, such tests should be updated too https://github.com/opencv/cvat/blob/develop/tests/python/rest_api/test_labels.py#L249-L253

cvat/apps/engine/schema.py
kirill-sizov: We use this in "schema" decorators for views like [here](https://github.com/opencv/cvat/blob/ed3dbe8f9d711a9894b51977b3ad11f4ec89f374/cvat/apps/engine/views.py#L1724-L1730). But yes I agree probably it would be more logical to move it in iam

cvat/apps/iam/schema.py
kirill-sizov: Moving this class in apps/engine/schema.py still leads to error due to [this line](https://github.com/opencv/cvat/blob/ed3dbe8f9d711a9894b51977b3ad11f4ec89f374/cvat/apps/engine/__init__.py#L7)

cvat/apps/events/handlers.py
kirill-sizov: Currently objects that have foreign key to Organization has `organization_id` attribute, for those models that don't have explicit relation with Organization has `get_organization_id` attribute. This logic is used in few places of code: events, webhooks, permissions. Currently I don't know how this can be done without defining organization_id in each model

zhiltsov-max:
> But I would suggest to postpone this idea until we have other such objects besides labels.

There is a couple more in the Honey Pot feature endpoints. Probably, an option to provide a list or even a mapping would be useful, similarly to what is done for other list filters.

PR #5909, opened 2023-03-22
Fixed resources import

Marishka17:
### Motivation and context

Fixed:
- wrong location of tmp file when importing job annotations
- ```Traceback (most recent call last):
    File "/home/maya/Documents/cvat/.env/lib/python3.8/site-packages/rq/worker.py", line 795, in work
      self.execute_job(job, queue)
    File "/home/maya/Documents/cvat/cvat/rqworker.py", line 37, in execute_job
      return self.perform_job(*args, **kwargs)
    File "/home/maya/Documents/cvat/.env/lib/python3.8/site-packages/rq/worker.py", line 1389, in perform_job
      self.handle_exception(job, *exc_info)
    File "/home/maya/Documents/cvat/.env/lib/python3.8/site-packages/rq/worker.py", line 1438, in handle_exception
      fallthrough = handler(job, *exc_info)
    File "/home/maya/Documents/cvat/cvat/apps/engine/views.py", line 2233, in rq_exception_handler
      rq_job.exc_info = "".join(
  AttributeError: can't set attribute
  ```

Resolves https://github.com/opencv/cvat/issues/5773
Resolves https://github.com/opencv/cvat/issues/5563
- root causes of the issues: 
  - [x] the annotation file was uploaded to the server by tus protocol and rq job was created but no one next requests for checking status were not made. (e.g. user closed the browser tab)
  - [x] the annotation file was uploaded to the server by tus protocol but rq job has not yet been created (e.g cvat instance restarted)
  - [x] tasks/projects creation from backups with the same name at the same time by different users
### How has this been tested?
<!-- Please describe in detail how you tested your changes.
Include details of your testing environment, and the tests you ran to
see how your change affects other areas of the code, etc. -->
Manually
### Checklist
<!-- Go over all the following points, and put an `x` in all the boxes that apply.
If an item isn't applicable for some reason, then ~~explicitly strikethrough~~ the whole
line. If you don't do that, GitHub will show incorrect progress for the pull request.

### License


zhiltsov-max:
@Marishka17, AFAIK, the problem is also actual for interrupted backup uploads.

Marishka17:
> AFAIK, the problem is also actual for interrupted backup uploads.

Yes, I know, this PR should fix that too.

SpecLad:
You should add "Fixes #xxx" tags to the description so that those issues can be closed automatically when this PR is merged.

nmanovic:
@Marishka17 , could you please resolve conflicts?

cvat/apps/engine/views.py
```diff
@@ -301,34 +305,34 @@ def perform_create(self, serializer, **kwargs):
...
                 import_func=_import_project_dataset,
                 rq_func=dm.project.import_dataset_as_project,
-                rq_id=rq_id,
+                rq_id_template=self.IMPORT_RQ_ID_TEMPLATE
             )
         else:
             action = request.query_params.get("action", "").lower()
             if action in ("import_status",):
                 queue = django_rq.get_queue(settings.CVAT_QUEUES.IMPORT_DATA.value)
+                rq_id = request.query_params.get('rq_id')
```
SpecLad, line 329: Shouldn't `rq_id` be validated somehow? What's to prevent a user from checking someone else's request status?

tests/python/rest_api/test_tasks.py
```diff
@@ -1378,3 +1380,58 @@ def test_move_task_from_one_project_to_another_with_attributes(self, task_id, pr
...
+        ).format(id=task_id)
+        uploader = Uploader(self.client)
+
+        response = uploader.upload_file(
+            url, filename, meta=params, query_params=params, logger=self.client.logger.debug
+        )
+
+        rq_id = json.loads(response.data)["rq_id"]
+        assert rq_id
+        sleep(3)
```
SpecLad, line 1433: This seems suspicious. Why sleep here? Why for 3 seconds?

cvat-sdk/cvat_sdk/core/uploading.py
```diff
@@ -273,17 +277,20 @@ def upload_file_and_wait(
...
+        assert rq_id, "The rq_id was not found in the response"
+
         self._wait_for_completion(
             url,
             success_status=201,
             positive_statuses=[202],
             status_check_period=status_check_period,
             query_params=params,
-            method="POST",
+            method="PUT",
```
SpecLad, line 293: Why? It seems wrong, why would this use `PUT` when the initial request was `POST`?

cvat/apps/engine/mixins.py
```diff
@@ -170,7 +173,28 @@ def init_tus_upload(self, request):
...
+                # we need to create unique temp dir here because
+                # users can try to import backups with the same name at the same time
+                unique_tmp_dir = Path(mkdtemp(prefix='cvat-', suffix='.backup', dir=self.get_upload_dir()))
+                filename = os.path.join(unique_tmp_dir.name, filename)
+                metadata['filename'] = filename
+            file_path = os.path.join(self.get_upload_dir(), filename)
+            file_exists = os.path.lexists(file_path)
+
+            if file_exists:
+                # check whether the rw_job is in progress or has been finished/failed
```
SpecLad, line 187:
```suggestion
                # check whether the rq_job is in progress or has been finished/failed
```

cvat/apps/engine/mixins.py
```diff
@@ -170,7 +173,28 @@ def init_tus_upload(self, request):
             if message_id:
                 metadata["message_id"] = base64.b64decode(message_id)
 
-            file_exists = os.path.lexists(os.path.join(self.get_upload_dir(), filename))
+            import_type = request.path.strip('/').split('/')[-1]
+            if import_type == 'backup':
+                # we need to create unique temp dir here because
+                # users can try to import backups with the same name at the same time
+                unique_tmp_dir = Path(mkdtemp(prefix='cvat-', suffix='.backup', dir=self.get_upload_dir()))
+                filename = os.path.join(unique_tmp_dir.name, filename)
```
SpecLad, line 181: Instead of creating a unique directory, why not create a unique file name? The effect will be the same, but it'll be simpler, and you won't need custom removal logic.

cvat-sdk/cvat_sdk/core/uploading.py
```diff
@@ -89,6 +90,7 @@ def create_url(self):
         headers["upload-length"] = str(self.file_size)
         headers["upload-metadata"] = ",".join(self.encode_metadata())
         resp = self._api_client.rest_client.POST(self.client.url, headers=headers)
+        self.metadata["real_filename"] = resp.headers.get("Upload-Filename")
```
SpecLad, line 93: IMO, it's not appropriate to use `self.metadata` to store this. `self.metadata` is the stuff that is passed via the `Upload-Metadata` header, which this is not. Just use a custom attribute.

cvat/apps/engine/views.py
```diff
@@ -495,9 +499,17 @@ def export_backup(self, request, pk=None):
             resource_type_field_name=None
         ),
         responses={
-            '201': OpenApiResponse(description='The project has been imported'), # or better specify {id: project_id}
-            '202': OpenApiResponse(description='Importing a backup file has been started'),
+            '201': OpenApiResponse(description='The project has been imported'),
+            '202': OpenApiResponse(RqIdSerializer, description='Importing a backup file has been started'),
         })
+        # TODO: for some reason the code generated from schema with different serializers
```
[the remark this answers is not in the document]
Marishka17: Yes, code that will be generated by openapi generator

cvat/apps/engine/views.py
```diff
@@ -2243,8 +2259,12 @@ def _download_file_from_bucket(db_storage, filename, key):
     with open(filename, 'wb+') as f:
         f.write(data.getbuffer())
 
-def _import_annotations(request, rq_id, rq_func, db_obj, format_name,
+def _import_annotations(request, rq_id_template, rq_func, db_obj, format_name,
                         filename=None, location_conf=None, conv_mask_to_poly=True):
+
+    def _is_there_past_non_deleted_job():
+        return rq_job and request.method == 'POST' and (rq_job.is_finished or rq_job.is_failed)
```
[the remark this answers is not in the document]
Marishka17: It makes the condition clearer

cvat/apps/engine/views.py
```diff
@@ -2322,6 +2349,8 @@ def _import_annotations(request, rq_id, rq_func, db_obj, format_name,
                 exc_info = exc_info.replace(import_error_prefix + ': ', '')
                 return Response(data=exc_info,
                     status=status.HTTP_400_BAD_REQUEST)
+            elif "Failed to find 'coco' dataset" in exc_info:
```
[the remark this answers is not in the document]
Marishka17: I don't think that such a way is good, I will prefer to handle the specific exceptions from Datumaro, but in the current implementation if some dataset has an incorrect structure then Datumaro raises just Exception(...). And in such cases, a user should receive a response with 400 status, not a server error.

tests/python/rest_api/test_tasks.py
Marishka17: This test consists in checking that if a rq_job was created when uploading the annotations, but no request was made to check the status of the rq_job, then the resources should be released and the user could upload the data again. 3 seconds is enough for uploading annotations and resource cleanup.

cvat-sdk/cvat_sdk/core/uploading.py
Marishka17: Because the POST request means that we will create the rq_job (only such an approach will allow us to determine the presence of zombie rq_jobs). We have the same logic on the UI client with PUT request.

cvat/apps/engine/views.py
```diff
@@ -480,7 +484,7 @@ def annotations(self, request, pk):
     def export_backup(self, request, pk=None):
         return self.serialize(request, backup.export)
 
-    @extend_schema(summary='Methods create a project from a backup',
+    @extend_schema(methods=['POST'], summary='Methods create a project from a backup',
```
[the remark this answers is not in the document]
Marishka17: It makes the documentation clearer for me. I can remove if you insist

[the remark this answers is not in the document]
SpecLad: I guess it's fine.

cvat/apps/engine/views.py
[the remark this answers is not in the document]
SpecLad: You should clarify that in the comment.

cvat/apps/engine/serializers.py
```diff
@@ -595,6 +595,9 @@ class RqStatusSerializer(serializers.Serializer):
     message = serializers.CharField(allow_blank=True, default="")
     progress = serializers.FloatField(max_value=100, default=0)
 
+class RqIdSerializer(serializers.Serializer):
+    rq_id = serializers.CharField()
```
zhiltsov-max, line 599: The parameter should not be called `rq_`, because it can uncover implementation details to a potential attacker. I suggest to change it to `upload_id` or `queue_id` instead. It should also be limited to some size like 40 symbols. It will be even better if the value is a hashed string.

SpecLad: > The parameter should not be called rq_, because it can uncover implementation details to a potential attacker.

The source is open - all details are already out there.

I would prefer if it was called `request_id`, but `rq_id` is already used in other places, and I'd rather avoid gratuitous changes.

cvat/apps/engine/views.py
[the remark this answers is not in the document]
zhiltsov-max: Probably, it's not completely correct. Datumaro should throw an error like [DatasetNotFoundError](https://github.com/cvat-ai/datumaro/blob/develop/datumaro/components/errors.py#L274), or [DatasetImportError](https://github.com/cvat-ai/datumaro/blob/develop/datumaro/components/errors.py#L198), or, at least, [DatumaroError](https://github.com/cvat-ai/datumaro/blob/develop/datumaro/components/errors.py#L17). If you get some other type here, probably the error is being repackaged somewhere.

cvat/apps/engine/views.py
```diff
@@ -39,6 +39,7 @@
 
 import cvat.apps.dataset_manager as dm
 import cvat.apps.dataset_manager.views  # pylint: disable=unused-import
+from cvat.apps.dataset_manager.views import IMPORT_CACHE_FAILED_TTL, IMPORT_CACHE_SUCCESS_TTL
```
zhiltsov-max, line 42: Why these variables are defined in the dataset manager? They seem nothing to do with datasets, and they are also relevant to backups. All of this looks like engine configuration parameters, so I suggest to place them either in settings or in the engine config.

tests/python/rest_api/test_projects.py
```diff
@@ -461,11 +461,13 @@ def _test_import_project(self, username, project_id, format_name, data):
...
             )
             assert response.status == HTTPStatus.ACCEPTED
+            rq_id = json.loads(response.data).get("rq_id")
+            assert rq_id, "The rq_id was not found in the response"
 
             while True:
                 # TODO: It's better be refactored to a separate endpoint to get request status
                 (_, response) = api_client.projects_api.retrieve_dataset(
-                    project_id, action="import_status"
+                    project_id, action="import_status", rq_id=rq_id
```
zhiltsov-max, line 470: BTW, the change seems to be breaking. Why won't we just implement a new uniform API for file uploading and status checks? A protocol like yours, but more generic:

1. POST `/backup`, `/annotations`, `/dataset` -> 202 + upload_id
2. Optionally, CVAT-TUS protocol requests
3. GET `/status` to get request status

Surely, need to think more about this, it's just an idea.

cvat/apps/dataset_manager/tests/test_formats.py
```diff
@@ -923,8 +923,8 @@ def _test_can_import_annotations(self, task, import_format):
             expected_ann = TaskAnnotation(task["id"])
             expected_ann.init_from_db()
 
-            dm.task.import_task_annotations(task["id"],
-                file_path, import_format, True)
+            dm.util.remove_resources(dm.task.import_task_annotations, should_be_deleted=False)(file_path, task["id"],
```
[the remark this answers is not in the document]
Marishka17: FYI, no, the source file won't be removed in this case, but I refactored the code

cvat/apps/engine/views.py
Marishka17: Fixed

tests/python/rest_api/test_tasks.py
Marishka17: Fixed to use the time based on the previous import

tests/python/rest_api/test_projects.py
Marishka17: Need to think about it, but I'm not sure this approach should be implemented in this PR. I guess it will require a significant redesign.

.github/workflows/helm.yml
```diff
@@ -78,8 +78,27 @@ jobs:
...
+          do
+            docker logs ${container_name} &> ${LOGS_DIR}/${container_name}.log
+          done
+
+    - name: Uploading "cvat" container logs as an artifact
+      if: failure() && steps.run_tests.conclusion == 'failure'
+      uses: actions/upload-artifact@v3.1.2
+      with:
+        name: rest_api_container_logs
+        path: "${{ github.workspace }}/testing"
```
zhiltsov-max, line 104: I'm not sure these changes are relevant to the PR, could you please check?

Marishka17: I've added it because I didn't understand why the helm tests failed just by tests log (but after adding it helm tests passed). 
I can extract this to another PR, but I guess that we should have the assets with server logs for helm tests too.

cvat/apps/dataset_manager/project.py
```diff
@@ -160,7 +160,7 @@ def data(self) -> dict:
         raise NotImplementedError()
 
 @transaction.atomic
-def import_dataset_as_project(project_id, dataset_file, format_name, conv_mask_to_poly):
+def import_dataset_as_project(src_file, project_id, format_name, conv_mask_to_poly):
```
zhiltsov-max, line 163: What is the point of this change?

Marishka17: Because all import functions (`import_dataset_as_project`, `import_task_annotations`, `import_job_annotations`, 
 including backups import `_import_project`, `_import_task`) will be called from another function `import_resource_with_clean_up_after` that will remove the uploaded file.

tests/python/rest_api/test_tasks.py
```diff
@@ -1518,3 +1521,64 @@ def test_can_report_correct_completed_jobs_count(tasks, jobs, admin_user):
...
+            assert response.status == HTTPStatus.OK
+            annotations = json.loads(response.data)["shapes"]
+            assert len(annotations) > 0
+
+    def _delete_annotations(self, task_id):
+        with make_api_client(self.user) as api_client:
+            (_, response) = api_client.tasks_api.destroy_annotations(id=task_id)
+            assert response.status == HTTPStatus.NO_CONTENT
+
+    @pytest.mark.parametrize("task_id", [20])
```
zhiltsov-max, line 1552: Why a specific id is needed here?

cvat/apps/engine/views.py
```diff
@@ -309,34 +312,41 @@ def perform_create(self, serializer, **kwargs):
...
             )
         else:
             action = request.query_params.get("action", "").lower()
             if action in ("import_status",):
                 queue = django_rq.get_queue(settings.CVAT_QUEUES.IMPORT_DATA.value)
+                rq_id = request.query_params.get('rq_id')
+                if not rq_id:
+                    return Response('The rq_id param should be specified in the query parameters', status=status.HTTP_400_BAD_REQUEST)
+
+                # check that used has access to current rq_job
```
zhiltsov-max, line 340:
```suggestion
                # check that the user has access to the current rq_job
```

cvat/apps/engine/views.py
```diff
@@ -492,9 +502,12 @@ def export_backup(self, request, pk=None):
             serializers=[ProjectFileSerializer, OpenApiTypes.NONE],
             resource_type_field_name=None
         ),
+        # TODO: for some reason the code generated by the openapi generator from schema with different serializers
+        # contains only one serializer, need to fix that.
```
zhiltsov-max, line 506: Probably, then new protocol should be described in the documentation somewhere. At least, in the endpoint description.

tests/python/rest_api/test_tasks.py
```diff
@@ -1518,3 +1521,64 @@ def test_can_report_correct_completed_jobs_count(tasks, jobs, admin_user):
 
         task, _ = api_client.tasks_api.retrieve(task["id"])
         assert task.jobs.completed == 1
+
+
+class TestImportTaskAnnotations:
```
zhiltsov-max, line 1526: Shouldn't there be tests for:
- 2 parallel uploads, if it is supported now
- a successful uploading after an interrupted upload
- the file cleanup after successful and unsuccessful uploads (I don't think we need to store unfinished files longer than, e.g., 24 hours)

cvat/apps/engine/views.py
```diff
@@ -2256,24 +2284,29 @@ def _import_annotations(request, rq_id, rq_func, db_obj, format_name,
...
 
         av_scan_paths(filename)
         meta = {
             'tmp_file': filename,
-            'tmp_file_descriptor': fd,
         }
         rq_job = queue.enqueue_call(
-            func=rq_func,
-            args=(db_obj.pk, filename, format_name, conv_mask_to_poly),
+            func=import_resource_with_clean_up_after,
```
zhiltsov-max, line 2296: Probably, this function can be used as a decorator on the functions sent to the queue, this may simplify the code.

tests/python/rest_api/test_tasks.py
Marishka17: > the file cleanup after successful and unsuccessful uploads (I don't think we need to store unfinished files longer than, e.g., 24 hours)

What do you mean by `no need to store unfinished files longer than, e.g., 24 hours`? Now all uploaded files will be deleted in any case (it doesn't matter how the job was finished) immediately after completion.

cvat/apps/engine/views.py
zhiltsov-max: That's interesting, I didn't know RQ doesn't support decorators. Ok, then there is no other way.

tests/python/rest_api/test_tasks.py
zhiltsov-max: After completion of what? The uncompleted uploads (e.g., a half of the file(-s) was uploaded, then the browser was closed) will never trigger the job.

SpecLad:
> You should add "Fixes #xxx" tags to the description so that those issues can be closed automatically when this PR is merged.

Ping?

cvat/apps/engine/views.py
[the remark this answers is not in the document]
SpecLad: You could accomplish that with a comment above the condition. That would be equally clear, and not require the reader to jump back and forth in the code.

cvat/apps/engine/backup.py
```diff
@@ -905,7 +929,7 @@ def _import(importer, request, queue, rq_id, Serializer, file_field_name, locati
                 serializer = Serializer(data=request.data)
                 serializer.is_valid(raise_exception=True)
                 payload_file = serializer.validated_data[file_field_name]
-                fd, filename = mkstemp(prefix='cvat_', dir=settings.TMP_FILES_ROOT)
+                filename = NamedTemporaryFile(prefix='cvat_', dir=settings.TMP_FILES_ROOT, delete=False).name
                 with open(filename, 'wb+') as f:
```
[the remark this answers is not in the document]
SpecLad: There's some similar code in `cvat/apps/engine/views.py`, you should update that too.

cvat/apps/engine/backup.py
```diff
@@ -49,6 +55,23 @@ class Version(Enum):
     V1 = '1.0'
 
 
+IMPORT_CACHE_FAILED_TTL = timedelta(hours=10).total_seconds()
+IMPORT_CACHE_SUCCESS_TTL = timedelta(hours=1).total_seconds()
```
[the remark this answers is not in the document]
SpecLad: I don't think that matters. Logically, these values are time intervals, so `timedelta` is the appropriate type for them. You can convert them to the appropriate unit at the point of use.

cvat-sdk/cvat_sdk/core/uploading.py
```diff
@@ -273,17 +279,19 @@ def upload_file_and_wait(
     ):
         url = self._client.api_map.make_endpoint_url(endpoint.path, kwsub=url_params)
         params = {"format": format_name, "filename": filename.name}
-        self.upload_file(
+        response = self.upload_file(
             url, filename, pbar=pbar, query_params=params, meta={"filename": params["filename"]}
         )
 
+        assert json.loads(response.data).get("rq_id"), "The rq_id was not found in the response"
```
SpecLad, line 286: I don't understand this. Shouldn't the `rq_id` be used somehow? Why would we check that it's there and then not use it?

cvat/apps/engine/views.py
```diff
@@ -2286,6 +2319,8 @@ def _import_annotations(request, rq_id, rq_func, db_obj, format_name,
                 exc_info = exc_info.replace(import_error_prefix + ': ', '')
                 return Response(data=exc_info,
                     status=status.HTTP_400_BAD_REQUEST)
+            elif any([i in exc_info for i in ("DatumaroError", "DatasetImportError", "DatasetNotFoundError")]):
```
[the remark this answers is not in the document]
Marishka17: These exceptions are not wrapped into `CvatImportError` because they occur in `Dataset.import_from` and using the method as I can see is not handled by `DatumaroError`/`DatasetImportError`/`DatasetNotFoundError` anywhere. ([1](https://github.com/opencv/cvat/blob/8ad485be560e6bbf6acedd55f359e7eca85433d3/cvat/apps/dataset_manager/formats/coco.py#LL30C30-L30C30), [2](https://github.com/opencv/cvat/blob/8ad485be560e6bbf6acedd55f359e7eca85433d3/cvat/apps/dataset_manager/formats/cvat.py#L1442)...).

This bug was also fixed.

zhiltsov-max:
Please add a reference https://github.com/OpenAPITools/openapi-generator/issues/6126 in the endpoint comments with multiple return types.

.github/workflows/helm.yml
Marishka17: @zhiltsov-max, I've removed it because it was wrong and  I just saw that this has already been implemented in [PR](https://github.com/opencv/cvat/pull/6013)

tests/python/rest_api/test_tasks.py
```diff
@@ -1634,3 +1638,110 @@ def test_can_report_correct_completed_jobs_count(tasks, jobs, admin_user):
...
+        url = self.client.api_map.make_endpoint_url(
+            self.client.api_client.tasks_api.create_annotations_endpoint.path
+        ).format(id=task_id)
+
+        uploader = Uploader(self.client)
+        uploader._tus_start_upload(url, query_params=params)
+        uploader._upload_file_data_with_tus(
+            url, filename, meta=params, logger=self.client.logger.debug
+        )
+        sleep(60)
```
SpecLad, line 1736: 60 seconds is a really long time to delay the test suite by. Is there no way to make it shorter? Maybe you could do the temporary file check every few seconds and end the test as soon as it passes?

tests/python/rest_api/test_tasks.py
```diff
@@ -1634,3 +1638,110 @@ def test_can_report_correct_completed_jobs_count(tasks, jobs, admin_user):
...
+        try:
+            result = subprocess.check_output(
+                f'docker exec test_cvat_server_1 bash -c "ls data/tasks/{task_id}/tmp | wc -l"',
```
SpecLad, line 1739: It should be possible to do the same thing in Kubernetes and avoid having this test be Docker-exclusive. See the `kube_exec_cvat` function.

supervisord/utils.conf
```diff
@@ -41,3 +41,12 @@ command=%(ENV_HOME)s/wait-for-it.sh %(ENV_CVAT_REDIS_HOST)s:6379 -t 0 -- bash -i
     "
 environment=VECTOR_EVENT_HANDLER="SynchronousLogstashHandler"
 numprocs=1
+
+[program:rqworker_cleaning]
```
SpecLad, line 45: I think you also need to update the Helm chart to run the new worker.

Marishka17: We have also `rqworker_notifications` in utils and I don't see it in helm (maybe I missed something).
@azhavoro, Could you please help me resolve comments related to helm?

azhavoro: I think we should to be consistent and not add a new worker pod to the helm if the worker is running inside the utils container when deploying with docker-compose.

cvat/settings/base.py
```diff
@@ -667,3 +674,7 @@ class CVAT_QUEUES(Enum):
 }
 
 BUCKET_CONTENT_MAX_PAGE_SIZE =  500
+
+IMPORT_CACHE_FAILED_TTL = timedelta(days=90)
+IMPORT_CACHE_SUCCESS_TTL = timedelta(hours=1)
+RUN_CLEAN_IMPORT_CACHE_FUNC_AFTER = timedelta(hours=2)
```
zhiltsov-max, line 680: The chose name would better suit for a boolean value, I suggest to rename into something like `UPLOADED_FILES_TTL`, `IMPORT_DATA_TTL`, `IMPORT_CACHE_CLEAN_DELAY` etc., supposing the value is about time.

cvat/settings/base.py
```diff
@@ -667,3 +674,7 @@ class CVAT_QUEUES(Enum):
 }
 
 BUCKET_CONTENT_MAX_PAGE_SIZE =  500
+
+IMPORT_CACHE_FAILED_TTL = timedelta(days=90)
```
zhiltsov-max, line 678: Please comment why such a long time is needed. Maybe, it should only be done on the platform, while the opensource version can have lesser time?

docker-compose.yml
```diff
@@ -77,6 +77,7 @@ services:
       DJANGO_LOG_SERVER_HOST: vector
       DJANGO_LOG_SERVER_PORT: 80
       no_proxy: clickhouse,grafana,vector,nuclio,opa,${no_proxy:-}
+      NUMPROCS: 1
```
zhiltsov-max, line 80: Probably, this change is from a local debug configuration and should not be committed?

tests/python/rest_api/test_tasks.py
```diff
@@ -1729,3 +1733,110 @@ def test_can_report_correct_completed_jobs_count(tasks, jobs, admin_user):
...
+            )
+
+        sleep(required_time)
+        if successful_upload:
+            self._check_annotations(task_id)
+            self._delete_annotations(task_id)
+        task.import_annotations(self.format, filename)
+        self._check_annotations(task_id)
+
+    @pytest.mark.timeout(70)
```
zhiltsov-max, line 1812: Please add an option to set the cleanup job delay with an env var, then define an extra env var for the test containers (example: [settings](https://github.com/opencv/cvat/blob/fa34b767325e7b92a6ab9e9385ef8ee39556b510/cvat/apps/quality_control/default_settings.py#L7), [use](https://github.com/opencv/cvat/blob/fa34b767325e7b92a6ab9e9385ef8ee39556b510/tests/docker-compose.quality_settings.yml#L7)). We totally don't want unit tests running more than a minute, unless they are in a special test suite.

cvat/settings/base.py
Marishka17: The [default](https://python-rq.org/docs/) value is 1 year, I guess that we need to set some optimal value to not keep it too long, but still be able to see the details of failed job

docker-compose.yml
Marishka17: No, we have such environment variables for other containers and I guess that passing this env into utils container is also useful (as you can see it is used now for configuring the number of processes for rqworker)

supervisord/utils.conf
SpecLad: Okay, my bad. I thought this was a new supervisord config.

SpecLad (approved):

SpecLad:
Please resolve merge conflicts and add a changelog entry.

PR #5926, opened 2023-03-24
Reduce deletion time of task and projects

azhavoro:
Before:
before
before_silk
After:
after
after_silk

### Checklist

### License

### DIFF
--- a/cvat/apps/engine/models.py
+++ b/cvat/apps/engine/models.py
@@ -326,6 +326,37 @@ class Project(models.Model):
     def get_log_path(self):
         return os.path.join(self.get_project_logs_dirname(), "project.log")
 
+    def delete(self, using=None, keep_parents=False):
+        from cvat.apps.events.signals import resource_delete
+        from cvat.apps.events.utils import clear_cache
+        try:
+            db_tasks = Task.objects.filter(
+                project_id=self.id
+            ).prefetch_related(
+                'project',
+                'segment_set__job_set__issues__comments'
+            )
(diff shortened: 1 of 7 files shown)


cvat/apps/engine/views.py
```diff
@@ -256,6 +256,12 @@ def perform_create(self, serializer, **kwargs):
             organization=self.request.iam_context['organization']
         )
 
+    def perform_destroy(self, instance):
```
nmanovic, line 259: I will not recommend implementing the logic this way. It is a workaround.

Need to optimize handers of signals. Manually call post_delete for tasks from post_delete for projects with all necessary prefetches. The same for post_delete jobs, etc. Need to use local memcache to check that the signal handler has been called already. The optimization should be trivial. It will work for all objects (not only for jobs)

nmanovic:
@azhavoro , could you please look at helm tests?

cvat/apps/engine/models.py
```diff
@@ -391,6 +422,33 @@ def get_tmp_dirname(self):
     def __str__(self):
         return self.name
 
+    def delete(self, using=None, keep_parents=False):
+        from cvat.apps.events.signals import resource_delete
+        from cvat.apps.events.utils import clear_cache
+        try:
```
nmanovic, line 428: I will recommend filling the cache in pre_delete and flushing it in post_delete. Thus we can be sure that we generate events only after an object is deleted. 

1. pre_delete -> store all required information inside the cache (you can prepare ready to use event here)
2. post_delete -> check that you have the required information inside the cache and flush it, otherwise handle the signal as usual.

PR #5944, opened 2023-03-27
Fix formatting script

zhiltsov-max:
### Motivation and context

Fixes a mistake in the script.

### Checklist

### License


dev/format_python_code.sh
```diff
@@ -6,10 +6,10 @@ PYTHON="${PYTHON:-python3}"
 BLACK="${PYTHON} -m black"
 ISORT="${PYTHON} -m isort"
 
-${BLACK} --version >/dev/null 2>&1 && {
+[ ${BLACK} --version >/dev/null 2>&1 ] && {
```
SpecLad, line 9: This doesn't look right. You can't just put arbitrary commands in `[ ]`:

```console
$ [ python -m black --version ]
-bash: [: syntax error: `-m' unexpected
```

zhiltsov-max: Does any of the versions work in your case?

SpecLad: No, but that's because the operator is wrong. It should be `||`, not `&&`.

Although for clarity, I would advise to rewrite it as:

```bash
if ! ${BLACK} --version >/dev/null 2>&1; then
    ....
fi
```

zhiltsov-max: > It should be ||

Ah, yes. Thanks. Done.

PR #5961, opened 2023-03-31
Remove the tus chunk endpoints from the schema

SpecLad:
### Motivation and context

As explained in the comment in `_tus_chunk_action`, I don't think we need explicit documentation for these endpoints. Hiding them makes the documentation a bit cleaner, especially because these endpoints were indistinguishable from one another.

It also removes the corresponding SDK code, which is good, since clients should follow the URL returned by the creation endpoint, not invoke the chunk endpoints directly.

### How has this been tested?

CI.

### Checklist

### License

### DIFF
--- a/cvat/apps/engine/views.py
+++ b/cvat/apps/engine/views.py
@@ -51,7 +51,6 @@ from cvat.apps.engine.models import (
     CloudProviderChoice, Location
 )
 from cvat.apps.engine.models import CloudStorage as CloudStorageModel
-from cvat.apps.engine.parsers import TusUploadParser
 from cvat.apps.engine.serializers import (
     AboutSerializer, AnnotationFileSerializer, BasicUserSerializer,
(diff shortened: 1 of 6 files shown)


cvat/apps/engine/views.py
```diff
@@ -83,6 +83,26 @@
 
 _UPLOAD_PARSER_CLASSES = api_settings.DEFAULT_PARSER_CLASSES + [MultiPartParser]
 
+def _tus_chunk_action(*, detail: bool, suffix_base: str):
```
zhiltsov-max, line 86: I suggest putting this function to `view_utils.py`

SpecLad: Alright, done.

zhiltsov-max (approved):

cvat/apps/engine/views.py
SpecLad: This caused an import loop, so I moved some stuff around to fix it.

PR #5962, opened 2023-04-01
Optimized jobs request

nmanovic:


nmanovic:
@SpecLad , could you please look?

cvat/apps/engine/serializers.py
```diff
@@ -156,10 +150,36 @@ class TasksSummarySerializer(_CollectionSummarySerializer):
 class CommentsSummarySerializer(_CollectionSummarySerializer):
     pass
 
+class BasicSummarySerializer(serializers.Serializer):
+    url = serializers.URLField(read_only=True)
+    count = serializers.IntegerField(read_only=True)
 
-class IssuesSummarySerializer(_CollectionSummarySerializer):
-    pass
+    def to_representation(self, instance):
```
SpecLad, line 157: I don't think you need to override `to_representation` here. You can replace the fields with `SerializerMethodField`s, and that should cause the methods to be invoked automatically.

cvat/apps/engine/serializers.py
```diff
@@ -156,10 +150,36 @@ class TasksSummarySerializer(_CollectionSummarySerializer):
...
+        }
+
+class LabelsSummarySerializer(BasicSummarySerializer):
+    def get_url(self, request, instance):
+        filter_key = instance.__class__.__name__.lower() + '_id'
+        return reverse('label-list', request=request,
+            query_params={ filter_key: instance.id })
+
+    def get_count(self, instance):
+        return getattr(instance, 'task_labels_count', 0) + getattr(instance, 'proj_labels_count', 0)
```
SpecLad, line 174: Under what conditions will the attributes be missing?

cvat/apps/engine/serializers.py
```diff
@@ -1059,15 +1083,14 @@ class ProjectReadSerializer(serializers.ModelSerializer):
...
+    labels = LabelsSummarySerializer(source='*')
 
     class Meta:
         model = models.Project
         fields = ('url', 'id', 'name', 'owner', 'assignee',
             'bug_tracker', 'task_subsets', 'created_date', 'updated_date', 'status',
             'dimension', 'organization', 'target_storage', 'source_storage',
-            'tasks', 'labels',
-        )
+            'tasks', 'labels')
```
SpecLad, line 1093: Unnecessary change.

cvat/apps/engine/views.py
```diff
@@ -1234,13 +1242,12 @@ def preview(self, request, pk):
...
-        'segment__task__project__label_set__sublabels__attributespec_set',
-        'segment__task__label_set__attributespec_set',
-        'segment__task__project__label_set__attributespec_set'
-    ).all()
+    queryset = Job.objects.select_related('assignee', 'segment__task__data',
+        'segment__task__project').annotate(Count('issues', distinct=True),
+        task_labels_count=Count('segment__task__label',
+            filter=Q(segment__task__label__parent__isnull=True), distinct=True),
+        proj_labels_count=Count('segment__task__project__label',
+            filter=Q(segment__task__project__label__parent__isnull=True), distinct=True)).all()
```
SpecLad, line 1250: Could you reformat this like in `ProjectViewSet`? It's hard to understand which arguments are for what.

cvat/apps/engine/serializers.py
nmanovic: I like the proposal but the method has common logic. It handles the case when the request is None. I will prefer to leave it as is. Otherwise, I have to repeat it 4 times.

PR #5974, opened 2023-04-04
Tried to optimize annotation fetching

bsekachev:
### Checklist

### License


cvat/apps/engine/mixins.py
```diff
@@ -279,8 +279,7 @@ def export_annotations(self, request, db_obj, export_func, callback, get_data=No
 
         data = get_data(self._object.pk)
         serializer = LabeledDataSerializer(data=data)
-        if serializer.is_valid(raise_exception=True):
-            return Response(serializer.data)
+        return Response(serializer.initial_data)
```
bsekachev, line 282: ``get_data`` is passed as an argument

It is a minor problem because in general it does not guarantee us that it receives "safe" data from the database..

cvat/apps/dataset_manager/task.py
```diff
@@ -453,8 +453,25 @@ def _init_shapes_from_db(self):
         for shape_id, shape_elements in elements.items():
             shapes[shape_id].elements = shape_elements
 
-        serializer = serializers.LabeledShapeSerializer(list(shapes.values()), many=True)
-        self.ir_data.shapes = serializer.data
```
bsekachev, line 457: Unnessesary converstion to serializer and back, from performance point of view it would be better to handle raw data, received from the database, keeping output

cvat/apps/dataset_manager/task.py
```diff
@@ -453,8 +453,25 @@ def _init_shapes_from_db(self):
...
+                    del value[field]
+
+            if value['elements']:
+                for element in value['elements']:
+                    convert_shape(element)
+
+            return value
+
+        values = [convert_shape(value) for value in shapes.values()]
+        self.ir_data.shapes = values
```
bsekachev, line 474: now the function "returns" generator, might be a problem?

cvat/apps/dataset_manager/task.py
```diff
@@ -399,8 +408,15 @@ def _init_tags_from_db(self):
             self._extend_attributes(db_tag.labeledimageattributeval_set,
                 self.db_attributes[db_tag.label_id]["all"].values())
 
-        serializer = serializers.LabeledImageSerializer(db_tags, many=True)
-        self.ir_data.tags = serializer.data
+        def convert_tag(tag):
+            tag['attributes'] = tag['labeledshapeattributeval_set']
+            del tag['labeledshapeattributeval_set']
```
bsekachev, line 413:
```suggestion
            tag['attributes'] = tag['labeledimageattributeval_set']
            del tag['labeledimageattributeval_set']
```

cvat/apps/dataset_manager/task.py
```diff
@@ -546,8 +575,28 @@ def _init_tracks_from_db(self):
...
+            del track['trackedshape_set']
+            track['attributes'] = track['labeledtrackattributeval_set']
+            del track['labeledtrackattributeval_set']
+            _remove_none(track)
+            for attr in track['attributes']:
+                del attr['id']
+
+            for shape in track['shapes']:
+                _remove_none(shape)
+                shape['attributes'] = shape['trackedshapeattributeval_set']
```
bsekachev, line 589:
```suggestion
                shape['attributes'] = shape['trackedshapeattributeval_set']
                del shape['trackedshapeattributeval_set']
```

nmanovic:
@bsekachev , can we pack all your changes into https://www.django-rest-framework.org/api-guide/serializers/#baseserializer? Inherit LabeledShapeSeiralizer from BaseSerializer and implement all logic here? Will it work?

cvat/apps/dataset_manager/task.py
```diff
@@ -512,6 +513,7 @@ def _init_tracks_from_db(self):
         tracks = {}
         elements = {}
         for db_track in db_tracks:
+            db_track['elements'] = []
```
bsekachev, line 516:
```suggestion
```

cvat/apps/engine/serializers.py
```diff
@@ -1207,6 +1206,65 @@ class SubLabeledShapeSerializer(ShapeSerializer, AnnotationSerializer):
 class LabeledShapeSerializer(SubLabeledShapeSerializer):
     elements = SubLabeledShapeSerializer(many=True, required=False)
 
+class LabeledImageSerializerFromDB(serializers.BaseSerializer):
```
SpecLad, line 1209: Do we need this to be a serializer at all? Given the usage, it seems like it would be simpler to just use a plain function.

cvat/apps/engine/serializers.py
```diff
@@ -1207,6 +1206,65 @@ class SubLabeledShapeSerializer(ShapeSerializer, AnnotationSerializer):
...
+    def to_representation(self, instance):
+        def convert_tag(tag):
+            tag['attributes'] = [OrderedDict(attr) for attr in tag['labeledimageattributeval_set']]
```
SpecLad, line 1212: Why is `OrderedDict` used here?

cvat/apps/engine/serializers.py
```diff
@@ -1207,6 +1206,65 @@ class SubLabeledShapeSerializer(ShapeSerializer, AnnotationSerializer):
...
+            del tag['labeledimageattributeval_set']
+            if 'parent' in tag:
+                del tag['parent']
+            for attr in tag['attributes']:
+                del attr['id']
+
+            return OrderedDict(tag)
```
SpecLad, line 1219: This function seems easy to misuse, because it both modifies the old value _and_ returns a new one. Normally, I would expect it to do only one of those.

Why not have the function construct a new dictionary with the needed keys and return it? It would be easier to understand, and also safer, because you won't be able to expose unwanted fields by forgetting to delete them.

cvat/apps/engine/serializers.py
bsekachev: To convert these values from ``dotdict`` type back to OrderedDict
Otherwise tests fail because it calls redefined ``__eq__`` function from ``dotdict`` class which is not supposed to be used there.

cvat/apps/engine/serializers.py
SpecLad: Why is it better? Can you summarize the discussion?

cvat/apps/engine/serializers.py
SpecLad: Makes sense. Would it be possible to use regular `dict` instead? Or would that also break the tests? It doesn't seem like we need `OrderedDict` otherwise.

bsekachev: We do not want to break API, the serializer returns OrderedDict now. It is why we use also OrderedDict

cvat/apps/engine/serializers.py
bsekachev: Because it is in general serialization logic, and we keep it in the serializer. 
But very basic, which works quick:

 https://www.django-rest-framework.org/api-guide/serializers/#baseserializer

cvat/apps/engine/serializers.py
bsekachev: Makes sense

cvat/apps/engine/backup.py
```diff
@@ -386,8 +386,7 @@ def serialize_annotations():
             for db_job_id in db_job_ids:
                 annotations = dm.task.get_job_data(db_job_id)
                 annotations_serializer = LabeledDataSerializer(data=annotations)
-                annotations_serializer.is_valid(raise_exception=True)
-                job_annotations.append(self._prepare_annotations(annotations_serializer.data, self._label_mapping))
+                job_annotations.append(self._prepare_annotations(annotations_serializer.initial_data, self._label_mapping))
```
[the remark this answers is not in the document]
bsekachev: You are right, aborted

cvat/apps/engine/serializers.py
```diff
@@ -1207,6 +1206,72 @@ class SubLabeledShapeSerializer(ShapeSerializer, AnnotationSerializer):
 class LabeledShapeSerializer(SubLabeledShapeSerializer):
     elements = SubLabeledShapeSerializer(many=True, required=False)
```
SpecLad, line 1208: It would be helpful to add a comment explaining why these custom serializers are used.

cvat/apps/engine/serializers.py
```diff
@@ -1207,6 +1206,72 @@ class SubLabeledShapeSerializer(ShapeSerializer, AnnotationSerializer):
...
+            attr_keys = ['spec_id', 'value']
+            keys = [
+                'id', 'label_id', 'type', 'frame', 'group',
+                'source', 'occluded', 'outside', 'z_order',
+                'rotation', 'points',
+            ]
+            result = OrderedDict([(key, shape[key]) for key in keys])
+            result['attributes'] = [
+                OrderedDict([(key, attr[key]) for key in attr_keys]) for attr in shape['labeledshapeattributeval_set']
+            ]
```
SpecLad, line 1236: There's a lot of duplication between the three classes. I think some common functions could be extracted here so that this code would become something like this:

```python
result = _convert_annotation(shape, specific_keys=['type', 'occluded', 'outside', 'z_order', 'rotation', 'points'])
result['attributes'] = _convert_attributes(shape['labeledshapeattributeval_set'])
```

cvat/apps/engine/serializers.py
```diff
@@ -1207,6 +1206,72 @@ class SubLabeledShapeSerializer(ShapeSerializer, AnnotationSerializer):
...
+                'id', 'label_id', 'frame', 'group', 'source',
+            ]
+            shape_keys = [
+                'id', 'type', 'frame', 'occluded', 'outside', 'z_order',
+                'rotation', 'points', 'trackedshapeattributeval_set',
+            ]
+            result = OrderedDict([(key, track[key]) for key in keys])
+            result['shapes'] = [
+                OrderedDict([(key, shape[key]) for key in shape_keys]) for shape in track['trackedshape_set']
+            ]
```
SpecLad, line 1257: It would be useful to add a `convert_tracked_shape` function that would do this conversion. That way, all shape-related code would be together (right now, it's split over this and the `for shape` loop below, two discontinuous blocks).

You could also be able to reuse the hypothetical common function that I mentioned in the other comment.

PR #5975, opened 2023-04-04
Fix write once fields

zhiltsov-max:
### Motivation and context

This PR fixes a problem with WriteOnceSerializer that it affected only the fields declared in the `Meta.fields`, while the manually declared fields were left unaffected. This PR also fixes several issues with invalid field declarations.

### Checklist

### License

### DIFF
--- a/cvat/apps/engine/serializers.py
+++ b/cvat/apps/engine/serializers.py
@@ -46,36 +46,23 @@ class WriteOnceMixin:
     Inspired by http://stackoverflow.com/a/37487134/627411.
     """
 
-    def get_extra_kwargs(self):
-        extra_kwargs = super().get_extra_kwargs()
-
-        # We're only interested in PATCH/PUT.
-        if 'update' in getattr(self.context.get('view'), 'action', ''):
-            extra_kwargs = self._set_write_once_fields(extra_kwargs)
-
-        return extra_kwargs
-
     def get_fields(self):
         fields = super().get_fields()
 
-        extra_kwargs = self.get_extra_kwargs()
-
-        for field_name, field_extra_kwargs in extra_kwargs.items():
-            field = fields.get(field_name)
-            read_only = field_extra_kwargs.get('read_only')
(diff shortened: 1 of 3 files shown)


cvat/schema.yml
```diff
@@ -8338,9 +8338,6 @@ components:
         assignee:
           type: integer
           nullable: true
-        message:
-          type: string
-          minLength: 1
```
SpecLad, line 8343: This seems odd to me. Should we not allow editing issue messages? Why not?

cvat/apps/engine/serializers.py
```diff
@@ -27,6 +27,70 @@
...
+    def get_extra_kwargs(self):
+        extra_kwargs = super().get_extra_kwargs()
+
+        # We're only interested in PATCH/PUT.
+        if 'update' in getattr(self.context.get('view'), 'action', ''):
+            extra_kwargs = self._set_write_once_fields(extra_kwargs)
+
+        return extra_kwargs
+
+    def get_fields(self):
```
SpecLad, line 58: Is there any point in overriding both `get_extra_kwargs` and `get_fields`? Seems like you could just remove the intermediate step (setting the `read_only` kwarg) and just use the `write_once_fields` setting directly in this function.

cvat/schema.yml
zhiltsov-max: Because they must be edited via the first comment. The field doesn't work.

cvat/apps/engine/serializers.py
zhiltsov-max: Technically, leaving just the `get_extra_kwargs` would be ideal, but it only affects the fields from the model, not the manually defined ones.

SpecLad: It would, but since it doesn't work - by design - I don't think we should persist with the kwargs approach.

SpecLad (approved):

PR #5991, opened 2023-04-06
Fix version checks in SDK

zhiltsov-max:
### Motivation and context

- Added missing compatibility with micro- and smaller releases

### How has this been tested?

Unit test

### Checklist

### License

### DIFF
--- a/cvat-sdk/cvat_sdk/core/client.py
+++ b/cvat-sdk/cvat_sdk/core/client.py
@@ -253,25 +253,27 @@ class Client:
                 raise IncompatibleVersionException(msg)
             return
 
-        sdk_version = pv.Version(VERSION)
-
-        # We only check base version match. Micro releases and fixes do not affect
-        # API compatibility in general.
-        if all(
-            server_version.base_version != sv.base_version for sv in self.SUPPORTED_SERVER_VERSIONS
+        if not any(
+            self._is_version_compatible(server_version, supported_version)
+            for supported_version in self.SUPPORTED_SERVER_VERSIONS
         ):
             msg = (
                 "Server version '%s' is not compatible with SDK version '%s'. "
                 "Some SDK functions may not work properly with this server. "
                 "You can continue using this SDK, or you can "
                 "try to update with 'pip install cvat-sdk'."
-            ) % (server_version, sdk_version)
+            ) % (server_version, pv.Version(VERSION))
             self.logger.warning(msg)
             if fail_if_unsupported:
                 raise IncompatibleVersionException(msg)
 
+    def _is_version_compatible(self, current: pv.Version, target: pv.Version) -> bool:
+        # Check for (major, minor) compatibility.
+        # Micro releases and fixes do not affect API compatibility in general.
+        upper_bound = pv.Version(f"{target.epoch}{target.major}.{target.minor + 1}")
+        return target <= current < upper_bound
+
     def get_server_version(self) -> pv.Version:
-        # TODO: allow to use this endpoint unauthorized
         (about, _) = self.api_client.server_api.retrieve_about()
         return pv.Version(about.version)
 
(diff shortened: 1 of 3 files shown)


cvat-sdk/cvat_sdk/core/client.py
```diff
@@ -253,25 +253,27 @@ def check_server_version(self, fail_if_unsupported: Optional[bool] = None) -> No
...
+            ) % (server_version, pv.Version(VERSION))
             self.logger.warning(msg)
             if fail_if_unsupported:
                 raise IncompatibleVersionException(msg)
 
+    def _is_version_compatible(self, current: pv.Version, target: pv.Version) -> bool:
+        # Check for (major, minor) compatibility.
+        # Micro releases and fixes do not affect API compatibility in general.
+        upper_bound = pv.Version(f"{target.epoch}{target.major}.{target.minor + 1}")
+        return target <= current < upper_bound
```
SpecLad, line 274: I think this is slightly wrong, because it will calculate that 2.5.0a1 is compatible with 2.4.0. I would suggest something like this instead:

```
return current in packaging.specifiers.SpecifierSet(f"~= {target.epoch}!{target.major}.{target.minor}.{target.micro}")
```

tests/python/sdk/test_client.py
```diff
@@ -160,6 +160,49 @@ def mocked_version(_):
...
+    [
+        ("3.2", "2"),
+        ("2", "2.1"),
+        ("2.1", "2.1"),
+        ("2.1.1", "2.1"),
+        ("2.2", "2.1"),
+        ("2.2", "2.3"),
+        ("2.1.0.dev123", "2.1.post2"),
+        ("1!1.3", "2.1"),
+        ("1!1.1.dev12", "1!1.1"),
```
SpecLad, line 174: There should be some cases that test multiple supported versions. There are multiple versions listed in the actual code, after all.

tests/python/sdk/test_client.py
```diff
@@ -160,6 +160,49 @@ def mocked_version(_):
...
+    server_version: str,
+    supported_version: str,
+):
+    logger, _ = fxt_logger
+
+    supported_v = pv.Version(supported_version)
+    server_v = pv.Version(server_version)
+
+    # Currently, it is ~=, as defined in https://peps.python.org/pep-0440/
+    expected_true = (
```
SpecLad, line 189: This should be part of the test parameters as a hardcoded value. Otherwise, you're just testing your code against a copy of itself.

Also, it's called `expected_true`, but... true what? You never actually check for a true value in this test.

cvat-sdk/cvat_sdk/core/client.py
zhiltsov-max: Yes, it's a nice catch.

tests/python/sdk/test_client.py
zhiltsov-max: > This should be part of the test parameters as a hardcoded value

It's becomes very hard to maintain quickly. Maybe it can work, if not touched often.

> Also, it's called expected_true, but... true what? You never actually check for a true value in this test.

"True" here is that there are no exceptions raised.

tests/python/sdk/test_client.py
zhiltsov-max: Ok, makes sense.

cvat-sdk/cvat_sdk/core/client.py
SpecLad: Since the version on the right side comes from the SDK, I wouldn't expect it to be a prerelease. And what's wrong with the 3rd case?

SpecLad (approved):

cvat-sdk/cvat_sdk/core/client.py
zhiltsov-max: I think, given what we do now (add the next release version to support the public instance, and once we publish it, we update SDK), it makes sense to try to limit the next version to its pre-release variant, so the old SDK will stop working automatically, once the new server release is deployed.

E.g.:
public server: 2.5.devX
sdk: v2.4 supporting >=2.4, <2.5

public server: 2.5
sdk v2.4 stops working

PR #5994, opened 2023-04-07
Make mild improvements to the main Dockerfile

SpecLad:
### Motivation and context

This applies several minor improvements that together shave off a little bit of build time (on my machine, it goes down from 14:10 to 13:20) and a little bit of size of the final image (likely just a few megabytes, so I didn't measure it exactly).

I also have a bigger improvement in the works, but it's a lot more complicated (and might not work out in the end), so I'd like to get these out of the way first.

### How has this been tested?

Copious amounts of `docker build .`. I haven't really tested if the resulting image still works, so I'm relying on CI for this.

### Checklist

### License

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,10 +2,6 @@ ARG PIP_VERSION=22.0.2
 
 FROM ubuntu:20.04 as build-image
 
-ARG http_proxy
-ARG https_proxy
-ARG no_proxy="nuclio,${no_proxy}"
-ARG socks_proxy
 ARG DJANGO_CONFIGURATION="production"
 
 RUN apt-get update && \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,26 +1,24 @@
+ARG PIP_VERSION=22.0.2
+
 FROM ubuntu:20.04 as build-image
 
-ARG http_proxy
-ARG https_proxy
-ARG no_proxy="nuclio,${no_proxy}"
-ARG socks_proxy
```
zhiltsov-max, line 6: Why proxy settings were removed?

SpecLad: `http_proxy`, `https_proxy` and `no_proxy` are all [predefined](https://docs.docker.com/engine/reference/builder/#predefined-args) in Docker, while `socks_proxy` is not a standard variable and won't do anything unless it's used explicitly - which, in the build stage, it isn't.

azhavoro (approved):

PR #6039, opened 2023-04-18
Honey pot UI

zhiltsov-max:
### Motivation and context

- Added UI for the Honey Pot feature

### Checklist

### Information about comments
- [x] Fixed aligns of preview and assignee on task page
- [x] Added margin in the end of job list on task page
- [x] Solved/Unsolved issues for Job are kept on place, we can think about removing them in PR with analytics page
image
- [x] Copy Job list button was restored to it's place
- [x] Added borders and border highlight like on task card for job card
- [x] Changed "0%" to "<1%"
image
- [x] Submit and Submit and Open button. I think we should keep those buttons for consistency. In any case user can decide by himself what he wants, just create and go back or immediately go to GT job
- [x]  About GT job issues and participating in computing completion progress
These jobs still have to be annotated and reviewed by someone in the annotation team, it makes sense to treat their progress equally with other jobs.
- [x] Space in submit form.
Personally, I think without it, form looks too short and page becomes empty.
- [x] Frame intercesction (as all data on Quality page) is extracted from Quality report, so currently we wont show it immidiately
- [x] Tracks are now shown only on frames included in GT job
- [x] Object menu in object sidebar is removed for GT annotation 
- [x] Sorting for conflcits/quality is fixed
- [x] Frame intersection `N/A` and no ability to download the report means the report doesn't exist yet. We need to wait for it to be computed
- [x] With fixed problem with tracks, there won't be any annotations on jobs with 0% intersection
- [x] 2d cuboids and tags wont supported in this version
- [x] Automatically go to the next frame in tag annotation mode
- [x] Find empty frame navigation in GT job
- [x] Delete frame in GT job
Currently deleted frames are marked for a task, so deleting frame in GT deletes it everywhere. It really doesn't make sence to support deleting frames in GT jobs, So lets disable that option for now
- ~~[ ] Broken top margin in the frame filter configuration dialog in the annotation page~~

 

### License


cvat/schema.yml
```diff
@@ -8441,6 +8834,79 @@ components:
...
+        task_id:
+          type: integer
+          nullable: true
+          readOnly: true
+        parent_id:
+          type: integer
+          nullable: true
+          readOnly: true
+        target:
+          $ref: '#/components/schemas/QualityReportTarget'
```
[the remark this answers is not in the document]
zhiltsov-max: The reason for the target field is the following:

```
GET /quality_reports?task_id=X -> task & nested job reports
GET /quality_reports?task_id=X&target=job -> nested job reports
```

cvat/schema.yml
```diff
@@ -1830,6 +1830,14 @@ paths:
         description: A simple equality filter for the task_name field
         schema:
           type: string
+      - name: type
+        in: query
+        description: A simple equality filter for the type field
+        schema:
+          type: string
+          enum:
+          - normal
```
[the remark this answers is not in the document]
zhiltsov-max: Yeah, "annotation" sounds good to me.

cvat/schema.yml
```diff
@@ -7223,6 +7548,54 @@ components:
...
+            minimum: 0
+          description: |
+            The list of frame ids. Applicable only to the manual frame selection
+      required:
+      - task_id
+      - type
+    JobWriteTypeEnum:
+      enum:
+      - ground_truth
+      type: string
```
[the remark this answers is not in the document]
zhiltsov-max: Okay, done.

cvat/schema.yml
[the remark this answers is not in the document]
SpecLad: Okay, that makes sense.

.vscode/launch.json
```diff
@@ -84,6 +84,26 @@
             ],
             "justMyCode": false,
         },
+        {
+            "name": "REST API tests: Attach to RQ quality rerports worker",
```
SpecLad, line 88:
```suggestion
            "name": "REST API tests: Attach to RQ quality reports worker",
```

bsekachev:
@klakhov 

May we align task preview a little bit?
image


And stick **Assignee** field to the right:
image


I would suggest adding a little margin at the end of this page:
image


Do we need this "Solved"/"Unsolved" issues here? Looks like now this is information for "analytics" page. 
image


With a little vertical gap (8px) it looks better IMHO: 
image


In this PR we lost this button. It is expected? 
image


Need to add border color as for tasks on project page:
image


Last updated filter does not work as expected. It does not show anything. 

image
image


**"0%"** -> **"< 1 %"**
image


Do we need both "Submit" and "Submit and open" buttons if we only can create on Job?
image

Any issues for GT job are unexpected:
image


- It is not convenient to put GT jobs to "Completed", "Acceptance". First need to put it to "Acceptance" on task page, then open the job and in annotation menu set "Completed". If we first set "Completed", then "Acceptance", "Completed" will be reset to "New".

bsekachev:
On annotation view it shows "Missing annotation" only on one frame. On others it does not, but it is still missing (all the objects are tracks)
image
image

This menu should not exist for GT annotations:
image

I am not sure, that sorting by "Conflicts" and by "Quality" works:
image
image

bsekachev:
What is difference between frame intersection N/A and 0%?
image
image

How is that possible that jobs with 0.0% intersection frames have GT annotations?

Yes, all annotations are tracks. But we annotated GT annotations on random frames. For example on frame 10 and 100.
And we can't guarantee that on frame 50 the interpolated object is correct. But it is considered as GT in job with frame 50. 

What's more I have a job with one frame **300** (the latest frame of the task).
But the latest frame in GT job is 294. 

So, I still have GT annotations on frame 300 where I didn't annotate anything.

bsekachev:
On GT job I do not have any masks annotated. Mask is a shape only, it can not be even interpolated. 

image

zhiltsov-max:
@bsekachev, 

> Any issues for GT job are unexpected

> I do not think GT jobs should participate in computing completion progress

These jobs still have to be annotated and reviewed by someone in the annotation team, it makes sense to treat their progress equally with other jobs.

> It is not convenient to put GT jobs to "Completed", "Acceptance". First need to put it to "Acceptance" on task page, then open the job and in annotation menu set "Completed".

There is the button for this in the job actions menu:


> How is that possible that jobs with 0.0% intersection frames have GT annotations?

Can it be a percent truncation problem? Maybe, there's just 1 common frame giving something like 0.04, which is being truncated by the UI.

> Yes, all annotations are tracks. But we annotated GT annotations on random frames. For example on frame 10 and 100. And we can't guarantee that on frame 50 the interpolated object is correct. But it is considered as GT in job with frame 50.

How would you suggest to resolve the problem? This is the same for frame step.

bsekachev:
One more bug in tag annotation mode with enabled option "Automatically go to the next frame"
image

It opens frame absent in GT job.

bsekachev:
It does not detect missing tags.
Is it expected?

image
image

cvat-canvas/src/scss/canvas.scss
```diff
@@ -148,6 +148,46 @@ polyline.cvat_canvas_shape_splitting {
     stroke-dasharray: 5;
 }
 
+.cvat_canvas_ground_truth {
+    stroke-dasharray: 1;
+}
+
+.cvat_canvas_conflicted {
+    stroke: #ff4800 !important;
```
bsekachev, line 156: May we avoid using !important?

cvat-core/src/quality-report.ts
```diff
@@ -0,0 +1,140 @@
...
+            id: undefined,
+            parent_id: undefined,
+            task_id: undefined,
+            job_id: undefined,
+            target: '',
+            gt_last_updated: undefined,
+            summary: undefined,
+            created_date: undefined,
+        };
+
```
bsekachev, line 80: Let's use class ``#privateProperties`` in new classes
Current approach is outdated

cvat-core/src/enums.ts
```diff
@@ -34,6 +34,11 @@ export enum JobState {
     REJECTED = 'rejected',
 }
 
+export enum JobType {
+    ANNOTATION = 'annotation',
```
bsekachev, line 38: I would suggest renaming to "REGULAR"

cvat-core/src/quality-conflict.ts
```diff
@@ -0,0 +1,162 @@
...
+    MISMATCHING = 'mismatching_label',
+    MISSING = 'missing_annotation',
+}
+
+export enum ConflictSeverity {
+    ERROR = 'error',
+    WARNING = 'warning',
+}
+
+export interface RawQualityConflictData {
```
bsekachev, line 16: Let's use ``Serialized`` instead of ``Raw`` everywhere

bsekachev:
``Find empty frame`` feature navigates to not existing frames in GT job:
image

bsekachev:
- [ ] I removed frame 20 in GT job and it is removed in a regular job

cvat-core/src/annotations-collection.ts
```diff
@@ -703,6 +703,14 @@ export default class Collection {
                 continue;
             }
 
+            if (filter) {
+                if (filter.jobID) {
+                    if (object.jobID && (object.jobID !== filter.jobID)) {
```
bsekachev, line 708: ``object.jobID && object.jobID !== filter?.jobID``

cvat-core/src/annotations-objects.ts
```diff
@@ -65,12 +65,14 @@ class Annotation {
...
     protected groupColors: Record<number, string>;
     public serverID: number | null;
+    public jobID: number;
     protected parentID: number | null;
     protected dimension: DimensionType;
     public group: number;
     public label: Label;
     public frame: number;
     private _removed: boolean;
+    protected isGroundTruth: boolean;
```
bsekachev, line 75: Do we really need both jobID and isGroundTruth?

cvat-core/src/annotations.ts
```diff
@@ -29,12 +29,42 @@ function getCache(sessionType) {
     throw new ScriptingError(`Unknown session type was received ${sessionType}`);
 }
 
-async function getAnnotationsFromServer(session) {
+function processGroundTruthAnnotations(rawAnnotations, groundTruthAnnotations) {
+    groundTruthAnnotations.shapes.forEach((annotation) => { annotation.is_gt = true; });
+    groundTruthAnnotations.tracks.forEach((annotation) => { annotation.is_gt = true; });
+    groundTruthAnnotations.tags.forEach((annotation) => { annotation.is_gt = true; });
```
bsekachev, line 35: To reduce code, let's concat and then apply forEach

cvat-core/src/annotations.ts
```diff
@@ -29,12 +29,42 @@ function getCache(sessionType) {
...
+    const result = {
+        shapes: [...rawAnnotations.shapes, ...groundTruthAnnotations.shapes],
```
bsekachev, line 37: I would recommend using concat & slice instead of spread operator. It works much faster.

Try two pieces of code:

```
[...Array(10000000), ...Array(10000000)]
```

and

```
Array(10000000).slice(0).concat(Array(10000000).slice(0))
```

cvat-core/src/annotations.ts
```diff
@@ -29,12 +29,42 @@ function getCache(sessionType) {
...
+
+function addJobId(rawAnnotations, jobID) {
+    rawAnnotations.shapes.forEach((annotation) => { annotation.job_id = jobID; });
+    rawAnnotations.tracks.forEach((annotation) => { annotation.job_id = jobID; });
+    rawAnnotations.tags.forEach((annotation) => { annotation.job_id = jobID; });
+    const result = {
+        shapes: [...rawAnnotations.shapes],
+        tracks: [...rawAnnotations.tracks],
+        tags: [...rawAnnotations.tags],
+    };
```
bsekachev, line 52: The same comments about concat and slice.

cvat-core/src/enums.ts
zhiltsov-max: It's not totally clear, what "regular" means. Originally, it was "normal", which had from the same problem.

zhiltsov-max:
@bsekachev, tags and cuboids are not supported yet.

cvat-core/src/annotations.ts
klakhov: Done

cvat-core/src/annotations-objects.ts
klakhov: Yes, we can compare jobID to GT job id all the time, but I think using such parameter makes everything much simplier

cvat-ui/src/reducers/notifications-reducer.ts
```diff
@@ -1533,7 +1541,40 @@ export default function (state = defaultState, action: AnyAction): Notifications
                         fetching: {
                             message: 'Could not fetch a list of jobs',
                             reason: action.payload.error.toString(),
-                            className: 'cvat-notification-notice-update-organization-membership-failed',
+                            className: 'cvat-notification-notice-update-get-jobs-failed',
```
bsekachev, line 1544: update-get-jobs?

cvat-ui/src/reducers/jobs-reducer.ts
```diff
@@ -89,6 +92,54 @@ export default (state: JobsState = defaultState, action: JobsActions): JobsState
...
+
+            return {
+                ...state,
+                activities: {
+                    ...state.activities,
+                    deletes: {
+                        ...deletes,
+                    },
+                },
+            };
```
bsekachev, line 109: Avoid making state object mutable

```suggestion
            return {
                ...state,
                activities: {
                    ...state.activities,
                    deletes: {
                        ...deletes,
                        [jobID]: false,
                    },
                },
            };
```

cvat-ui/src/reducers/jobs-reducer.ts
```diff
@@ -89,6 +92,54 @@ export default (state: JobsState = defaultState, action: JobsActions): JobsState
...
+        }
+        case JobsActionTypes.DELETE_JOB_SUCCESS: {
+            const { jobID } = action.payload;
+            const { deletes } = state.activities;
+
+            deletes[jobID] = true;
```
bsekachev, line 115: The same

cvat-ui/src/reducers/jobs-reducer.ts
```diff
@@ -89,6 +92,54 @@ export default (state: JobsState = defaultState, action: JobsActions): JobsState
...
+                        ...deletes,
+                    },
+                },
+            };
+        }
+        case JobsActionTypes.DELETE_JOB_FAILED: {
+            const { jobID } = action.payload;
+            const { deletes } = state.activities;
+
+            delete deletes[jobID];
```
bsekachev, line 131: The same, please use omit() from lodash

bsekachev, line 131: The same, please use omit() from lodash

cvat-ui/src/containers/annotation-page/top-bar/top-bar.tsx
```diff
@@ -440,21 +442,41 @@ class AnnotationTopBarContainer extends React.PureComponent<Props, State> {
...
             if (playing) {
                 onSwitchPlay(false);
             }
-            this.changeFrame(value);
+            const newFrame = await jobInstance.frames.search(
+                { notDeleted: !showDeletedFrames },
+                Math.min(jobInstance.stopFrame, value),
+                jobInstance.stopFrame,
+            );
+            console.log(frameNumber, value, newFrame);
```
bsekachev, line 476: extra code

cvat-ui/src/containers/annotation-page/standard-workspace/objects-side-bar/objects-list.tsx
```diff
@@ -69,23 +75,30 @@ function mapStateToProps(state: CombinedState): StateToProps {
...
     let statesLocked = true;
-
-    objectStates.forEach((objectState: ObjectState) => {
+    const filteredStates = filterAnnotations(objectStates, {
+        statesSources,
+        frame: frameNumber,
+        groundTruthJobFramesMeta,
+        workspace,
+    });
+    filteredStates.forEach((objectState: ObjectState) => {
```
bsekachev, line 101: Potential performance issue
``mapStateToProps`` called each time when global state is updated.
Since it returns filteredStates being a different object each time, it will cause extra rerenders

cvat-ui/src/components/task-analytics-page/task-analytics-page.tsx
```diff
@@ -0,0 +1,131 @@
...
+    const receieveTask = (): void => {
+        if (Number.isInteger(id)) {
+            core.tasks.get({ id })
+                .then(([task]: Task[]) => {
+                    if (task) {
+                        setTaskInstance(task);
+                    }
+                }).catch((error: Error) => {
+                    notification.error({
+                        message: 'Could not receive the requested project from the server',
```
bsekachev, line 40:
```suggestion
                        message: 'Could not receive the requested task from the server',
```

cvat-ui/src/components/task-analytics-page/task-analytics-page.tsx
```diff
@@ -0,0 +1,131 @@
...
+        } else {
+            notification.error({
+                message: 'Could not receive the requested task from the server',
+                description: `Requested task id "${id}" is not valid`,
+            });
+            setFetchingTask(false);
+        }
+    };
+
+    const onJobUpdate = (job: Job): void => {
```
bsekachev, line 55: useCallback?

cvat-ui/src/components/task-analytics-page/quality/job-list.tsx
```diff
@@ -0,0 +1,270 @@
+// Copyright (C) 2020-2022 Intel Corporation
```
bsekachev, line 1:
```suggestion
```

cvat-ui/src/components/task-analytics-page/quality/job-list.tsx
```diff
@@ -0,0 +1,270 @@
...
+        useSelector((state: CombinedState) => state.analytics.quality.jobsReports);
+
+    useEffect(() => {
+        const jobsReportsMap: Record<number, QualityReport> = {};
+        for (const job of jobs) {
+            const report = jobReportsFromState.find((_report: QualityReport) => _report.jobId === job.id);
+            if (report) jobsReportsMap[job.id] = report;
+        }
+        setJobsReports(jobsReportsMap);
+    }, [renderedJobs, jobReportsFromState]);
```
bsekachev, line 50:
```suggestion
    }, [renderedJobs, jobReportsFromState, jobs]);
```

cvat-ui/src/components/task-analytics-page/quality/common.ts
```diff
@@ -0,0 +1,19 @@
+export function percent(a?: number, b?: number): string | number {
```
bsekachev, line 1: missed license header

cvat-ui/src/components/task-analytics-page/quality/common.ts
```diff
@@ -0,0 +1,19 @@
...
+    if (typeof a !== 'undefined' && Number.isFinite(a) && b) {
+        return `${Number((a / b) * 100).toFixed(1)}%`;
+    }
+    return 'N/A';
+}
+
+const THRESHOLD = 5000;
```
bsekachev, line 8: better to move to 'consts'

cvat-ui/src/containers/annotation-page/standard-workspace/objects-side-bar/objects-list.tsx
klakhov: Fixed that with storing filteredStates in state and changing it only if objectStates changed

cvat-ui/src/components/job-item/job-item.tsx
```diff
@@ -0,0 +1,249 @@
...
+                                                {JobStage.ANNOTATION}
+                                            </Select.Option>
+                                            <Select.Option value={JobStage.VALIDATION}>
+                                                {JobStage.VALIDATION}
+                                            </Select.Option>
+                                            <Select.Option value={JobStage.ACCEPTANCE}>
+                                                {JobStage.ACCEPTANCE}
+                                            </Select.Option>
+                                        </Select>
+
```
bsekachev, line 188:
```suggestion
```

cvat-ui/src/components/job-item/job-actions-menu.tsx
```diff
@@ -0,0 +1,101 @@
...
+import Menu from 'antd/lib/menu';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { MenuInfo } from 'rc-menu/lib/interface';
+import { exportActions } from 'actions/export-actions';
+
+import {
+    Job, JobStage, JobType, getCore,
+} from 'cvat-core-wrapper';
+import { deleteJobAsync } from 'actions/jobs-actions';
+import Modal from 'antd/lib/modal';
```
bsekachev, line 16: let's keep 3rdparty imports at the top

cvat-ui/src/components/job-item/job-actions-menu.tsx
```diff
@@ -0,0 +1,101 @@
...
+            onOk: () => {
+                dispatch(deleteJobAsync(job));
+            },
+            okButtonProps: {
+                type: 'primary',
+                danger: true,
+            },
+            okText: 'Delete',
+        });
+    }, [id]);
```
bsekachev, line 47: why not rely on `job``?

cvat-ui/src/components/create-job-page/create-job-page.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+import './styles.scss';
+
+import React, { useEffect, useState } from 'react';
+import { Row, Col } from 'antd/lib/grid';
+import { useParams } from 'react-router';
+import Text from 'antd/lib/typography/Text';
+import Spin from 'antd/lib/spin';
+import { Task } from 'reducers';
+import { notification } from 'antd';
```
bsekachev, line 13:
```suggestion
import notification  from 'antd/lib/notification';
```

cvat-ui/src/components/create-job-page/create-job-page.tsx
```diff
@@ -0,0 +1,71 @@
...
+                        setTaskInstance(task);
+                    }
+                }).catch((error: Error) => {
+                    notification.error({
+                        message: 'Could not fetch requested task from the server',
+                        description: error.toString(),
+                    });
+                }).finally(() => {
+                    setFetchingTask(false);
+                });
```
bsekachev, line 38: After any async requests in hooks need to be sure that the component is still mounted before updating its state, otherwise it leades to errors to the browser console (changing state of unmounted component). 

You can use custom hook for that:
```
import { useIsMounted } from '@root/utils/hooks';
const isMounted = useIsMounted()
if (isMounted()) { ...  updateState .... }
```

bsekachev: Please, check the same in other functional components

cvat-ui/src/containers/annotation-page/standard-workspace/objects-side-bar/objects-list.tsx
```diff
@@ -184,27 +203,50 @@ class ObjectsListContainer extends React.PureComponent<Props, State> {
...
+
+    public componentDidUpdate(prevProps: Props): void {
+        const {
+            statesSources, objectStates,
+        } = this.props;
+        const { objectStates: stateObjects } = this.state;
+        if (objectStates !== stateObjects ||
+            statesSources !== prevProps.statesSources
+        ) {
+            this.updateObjects();
```
bsekachev, line 223: Probably ``getDerivedStateFromProps`` is exactly what you need here.

cvat-ui/src/components/annotation-page/tag-annotation-workspace/frame-tags.tsx
```diff
@@ -44,18 +47,26 @@ function mapDispatchToProps(dispatch: ThunkDispatch<CombinedState, {}, Action>):
...
     useEffect(() => {
-        setFrameTags(states.filter((objectState: any): boolean => objectState.objectType === ObjectType.TAG));
-    }, [states]);
+        setFrameTags(
+            states.filter(
+                (objectState: ObjectState): boolean => (
+                    objectState.objectType === ObjectType.TAG &&
+                    (!objectState.jobID || statesSources.includes(objectState.jobID))
+                ),
+            ),
```
bsekachev, line 67: Can we use filterAnnotations here?

cvat-ui/src/components/annotation-page/standard-workspace/objects-side-bar/objects-list-header.tsx
```diff
@@ -91,7 +113,12 @@ function ObjectListHeader(props: Props): JSX.Element {
                         <HideAllSwitcher {...props} />
                     </>
                 )}
-                <CollapseAllSwitcher {...props} />
+                <Row>
+                    {readonly && (
+                        <GTSwitcher {...props} />
+                    )}
```
bsekachev, line 119: It is not clear how readonly should affect GTSwitcher
They have different semantic meaning
If you wanted to show GTSwitcher only in review mode, better to use workspace variable

cvat-ui/src/components/annotation-page/review/issues-aggregator.tsx
```diff
@@ -50,7 +53,7 @@ export default function IssueAggregatorComponent(): JSX.Element | null {
         }
 
         return () => {};
-    }, [canvasInstance]);
+    }, [canvasInstance, ready]);
```
bsekachev, line 56: should we use ``ready`` in the body?

like: ``canvasInstance instanceof Canvas and ready`` ?

cvat-ui/src/components/annotation-page/review/conflicts-aggregator.tsx
```diff
@@ -0,0 +1,125 @@
+// Copyright (C) 2023 CVAT.ai Corporation
```
bsekachev, line 1: Almost full dublicate of ``issues-aggregator.tsx``. 
I believe better to add the loop ``for (const conflict of mapping) {`` to there

cvat-ui/src/containers/annotation-page/standard-workspace/objects-side-bar/objects-list.tsx
klakhov: I think its better to use DidUpdate, cuz we also need to know if statesSources changed

cvat-ui/src/components/annotation-page/review/issues-aggregator.tsx
klakhov: Added

cvat-ui/src/components/annotation-page/review/conflicts-aggregator.tsx
klakhov: Moved all logic to issues-aggregator

bsekachev:
@klakhov 

> Submit and Submit and Open button. I think we should keep those buttons for consistency. In any case user can decide by himself what he wants, just create and go back or immediately go to GT job

There is no reason to stay on "Create GT job" page after GT job is already created because several GT jobs can not be created.
Two options I see: open GT job immediately, return to analytics page.

cvat-core/src/quality-settings.ts
```diff
@@ -0,0 +1,229 @@
...
+            low_overlap_threshold: undefined,
+            compare_line_orientation: undefined,
+            line_orientation_threshold: undefined,
+            compare_groups: undefined,
+            group_match_threshold: undefined,
+            check_covered_annotations: undefined,
+            object_visibility_threshold: undefined,
+            panoptic_comparison: undefined,
+            compare_attributes: undefined,
+        };
```
bsekachev, line 57: outdated code?

cvat-core/src/quality-settings.ts
```diff
@@ -0,0 +1,229 @@
...
+
+        Object.defineProperties(
+            this,
+            Object.freeze({
+                id: {
+                    get: () => data.id,
+                },
+
+            }),
+        );
```
bsekachev, line 88: outdated code?

cvat-core/src/server-proxy.ts
```diff
@@ -2167,6 +2200,75 @@ async function receiveWebhookEvents(type: WebhookSourceType): Promise<string[]>
     }
 }
 
+async function getQualityReports(filter): Promise<SerializedQualityReportData[]> {
+    const params = enableOrganization();
```
bsekachev, line 2204: ``enableOrganization`` works automatically. 
It is not necessary to add it explicitly in simple requests.

It is necessary for example when creating a task, or exporting a dataset (where organization can be changed in theory during the process)

cvat-core/src/server-proxy.ts
```diff
@@ -2167,6 +2200,75 @@ async function receiveWebhookEvents(type: WebhookSourceType): Promise<string[]>
...
+
+    try {
+        const response = await Axios.get(`${backendAPI}/quality/settings`, {
+            params: {
+                ...params,
+                task_id: taskID,
+            },
+        });
+        const settings = response.data.results[0];
+        return settings;
```
bsekachev, line 2249:
```suggestion
        return response.data.results[0];
```

klakhov:
@bsekachev 
> There is no reason to stay on "Create GT job" page after GT job is already created because several GT jobs can not be created. Two options I see: open GT job immediately, return to analytics page.

Ok, I see. Let's stick to the first option

bsekachev:
Okay, let's solve conflicts wait for Andrey review and I believe we can merge it

zhiltsov-max:
- [ ] It seems the report computation time info is missing (it was under the downloading button)
image

- [ ] Job actions menu needs to be fixed: the 3 to "Go To ..." buttons have no use outside the annotation window, probably need to add the black border around the menu
image

- [ ] Job frame filter has invalid top margin (it's a regression against develop)
image

klakhov:
> * [x]  It seems the report computation time info is missing (it was under the downloading button)
Fixed
> * [ ]  Job actions menu needs to be fixed: the 3 to "Go To ..." buttons have no use outside the annotation window, probably need to add the black border around the menu
This menu is mutual for Jobs cards(on task and jobs pages), those buttons are used on Jobs Page mostly. I dont think we need 2 different menus
> * [x]  Job frame filter has invalid top margin (it's a regression against develop)
Fixed

azhavoro (approved):

zhiltsov-max:
> This menu is mutual for Jobs cards(on task and jobs pages), those buttons are used on Jobs Page mostly. I dont think we need 2 different menus

Ok, but job menus on the task page have the same problem - the "Go To" buttons are not needed there.

PR #6074, opened 2023-04-25
Improve task creation with cloud storage and share data

### LINKED ISSUE #6037: Task creation impossible in WebUI when local share contains a big hierarchy of folders with images

1. Create a folder X with at least 500 folders inside (1000 would be better), each containing some images.
1. Start new CVAT instance v2.4.2 (with HTTPS), with folder X connected as local share.
1. Create new annotation project, inside it - new annotation task.
1. Select "Connected file share", "check" the folder "X", observe ~500 requests fired for each sub-folder.
2. "Un-check" one subfolder, another ~500 requests will be fired again for each sub-folder still "checked".
3. When the requests are not finished, try to "Submit and Open" the task - it will refuse with an error, that the task should contain at least one file.

### Expected Behaviour

The FileManager **should not** request full information about each sub-folder (via separate request for each sub-folder) each time I check or uncheck one of the sub-folders. It should make these requests only when I click "Submit".

### Current Behaviour

FileManager requests full information about each sub-folder each time that I check/uncheck one of the subfolders. While the requests are sent (1 request per sub-folder at ~10RPS speed), task can't be submitted - it fails with error "task should contain at least one file".

### Possible Solution

Stop FileManager from triggering the GetFiles each time I check or uncheck a sub-folder in "Connected file share".

### Context

If your folder structure is big (~1000 sub-folders in each folder), and you need to uncheck manually at least 10 subfolders, the situation becomes unbearable: thousands of requests are sent with stable frequency of ~10 RPS. The task can't be submitted for creation until all thousands of requests are finished, which makes this GUI virtually impossible to use.

This worked well in CVAT v2.2, and judging by Git Blame, it seems that broken behavior was induced by commit 1db247aedd.

### Your Environment

- Git hash commit (`git log -1`): b4549f42ee1c
- Docker version `docker version` (e.g. Docker 17.0.05): 23.0.3
- Are you using Docker Swarm or Kubernetes? No, own instance in-cloud
- Operating System and version (e.g. Linux, Windows, MacOS): Ubuntu 22.04.2 LTS

Attached are examples "how many requests are fired". Picture 1 - before unchecking one sub-folder, Picture 2 - after unchecking one subfolder.
Screenshot-20230417223806-1502x831
Screenshot-20230417224707-1502x831

Marishka17:
### Motivation and context
Resolved #6037
Related https://github.com/opencv/cvat/issues/4400
Related https://github.com/opencv/cvat/issues/6028

image

#### This PR contains several improvements in working with cloud storage/share data:
- [x] Manifest file is optional on UI
- [x] Redefined and unified the logic to work with both cloud storage and share point data
- [x] Supported task creation with real cloud storage data with `use_cache`
- [x] Supported task creation with real cloud storage data (all data types supported by the server by default, including `video`) without `use_cache`
- [x] Supported task creation with directories from cloud storage
- [x] Supported files and directories that should be excluded from selected with cloud storage and share point data (server only)
- [x] Added new version of REST API `cloudstorages/{id}/content` endpoint

### How has this been tested?

Added REST API tests for cloud storage content and task creation actions.
### Checklist

- [x] Improve tree with cloud storage files
- [x] Add examples to the documentation

### License


cvat-ui/src/components/file-manager/file-manager.tsx
```diff
@@ -67,30 +55,37 @@ export class FileManager extends React.PureComponent<Props, State> {
...
+            files: {
+                ...files,
+                cloudStorage: keys.filter((item) => !item.endsWith('.jsonl')),
+            },
+        });
+        onUploadCloudStorageFiles(keys);
+    };
+
+    private handleUploadSharedStorageFiles = (
+        shareFiles: { key: string, type: RemoteFileType, mime_type: string }[],
```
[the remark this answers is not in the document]
bsekachev: Done

cvat-ui/src/components/file-manager/file-manager.tsx
```diff
@@ -152,78 +146,12 @@ export class FileManager extends React.PureComponent<Props, State> {
...
-                            <Text strong>
-                                <a href={SHARE_MOUNT_GUIDE_URL}> mounted </a>
-                            </Text>
-                            share before you built CVAT and the shared storage contains files
-                        </Paragraph>
-                    </div>
-                )}
+                <RemoteBrowser
+                    resource='share'
+                    onSelectFiles={this.handleUploadSharedStorageFiles}
```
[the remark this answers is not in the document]
bsekachev: Done

cvat-ui/src/components/create-task-page/create-task-content.tsx
```diff
@@ -316,17 +315,18 @@ class CreateTaskContent extends React.PureComponent<Props & RouteComponentProps,
...
     }[]): void => {
+        let filteredFiles = shareFiles;
         const { many } = this.props;
         const { files } = this.state;
 
         let uploadFileErrorMessage = '';
 
         if (!many && shareFiles.length > 1) {
-            uploadFileErrorMessage = shareFiles.every((it) => it.mime_type === 'image') ?
+            uploadFileErrorMessage = shareFiles.every((it) => ['image', 'DIR'].includes(it.mime_type)) ?
```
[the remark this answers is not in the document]
bsekachev: Done

cvat/schema.yml
```diff
@@ -746,6 +746,7 @@ paths:
   /api/cloudstorages/{id}/content:
     get:
       operationId: cloudstorages_retrieve_content
+      description: This method was deprecated. Please use the new version of API.
```
SpecLad, line 749: You should name the new version.

cvat/schema.yml
```diff
@@ -790,6 +792,70 @@ paths:
...
+      parameters:
+      - in: header
+        name: X-Organization
+        schema:
+          type: string
+      - in: query
+        name: delimiter
+        schema:
+          type: string
+        description: The character you use to group keys
```
SpecLad, line 808: I don't understand this. What are keys? The word "key" doesn't occur anywhere else in this description.

Marishka17: `key` is the official term of the file on the cloud bucket. Most parts of CS providers use this term.

cvat/apps/engine/task.py
```diff
@@ -61,11 +62,44 @@ class SegmentsParams(NamedTuple):
...
+):
     job = rq.get_current_job()
     job.meta['status'] = 'Data are being copied from source..'
     job.save_meta()
 
+    # filter data from files/directories that should be excluded
+    if data_to_be_excluded:
+        files_to_be_excluded, directories_to_be_excluded = [], []
+        for f in data_to_be_excluded:
+            if f.endswith(os.path.sep):
```
SpecLad, line 79: `os.path.sep` should not be used here. `data_to_be_excluded` is supplied by the client, and the client doesn't know what path separator is used on the server (nor should it be concerned with it).

Also, it seems unnecessary to even do this check at all. Can't we exclude all files that are either in `data_to_be_excluded` or whose ancestor is in there? We don't need to know whether each individual entry in `data_to_be_excluded` is a file or a directory.

cvat/apps/engine/task.py
```diff
@@ -299,7 +345,6 @@ def _validate_manifest(manifests, root_dir, is_in_cloud, db_cloud_storage, data_
                 raise ValidationError("Manifest file can be uploaded only if 'Use cache' option is also selected")
             return manifest_file
         raise ValidationError('Invalid manifest was uploaded')
-    return None
```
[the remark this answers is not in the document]
Marishka17: In python, every function returns something. If it doesn't have a return statement, then it returns None. This matches the declared type.

cvat/apps/engine/views.py
```diff
@@ -2105,7 +2110,82 @@ def content(self, request, pk):
             # need to update index
             manifest.set_index()
             manifest_files = [os.path.join(manifest_prefix, f) for f in manifest.data]
-            return Response(data=manifest_files, content_type="text/plain")
+            return Response(
+                data=manifest_files,
+                content_type='text/plain',
+                headers={'Deprecation': datetime(2023, 7, 1, tzinfo=timezone.get_current_timezone()).strftime('%a, %d %B %Y %H:%M:%S %Z')}
```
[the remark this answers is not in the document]
Marishka17: I just wanted to emphasize in every possible way that this endpoint will be deprecated. Yes, I assumed this date, because I don't know in advance the date of the next release. But I can't indicate in the `Deprecation` header that this endpoint will be deprecated in the next release due to specification.
So, do you recommend removing it at all?

cvat/schema.yml
```diff
@@ -790,6 +792,70 @@ paths:
                 items:
                   type: string
           description: A manifest content
+  /api/cloudstorages/{id}/content-v2:
```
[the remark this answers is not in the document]
Marishka17: Maxim is going to use slashes in API (quality_reports -> quality/reports), but I'm not sure that I should use it too.

cvat/apps/engine/task.py
Marishka17: Fixed, thanks

site/content/en/docs/api_sdk/sdk/lowlevel-api.md
```diff
@@ -483,3 +483,70 @@ while True:
...
+
+Here you can find the examples how to get the bucket content with 2 versions.
+
+```python
+import os
+from pprint import pprint
+
+from cvat_sdk.api_client import ApiClient, Configuration
+
+# get the content of the bucket using old version of REST API Endpoint
```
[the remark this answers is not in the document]
Marishka17: I've removed this example

site/content/en/docs/api_sdk/sdk/lowlevel-api.md
```diff
@@ -483,3 +483,70 @@ while True:
 with open('output_file', 'wb') as output_file:
     output_file.write(response.data)
 ```
+### Different versions of API Endpoints
+#### The cloudstorages/id/content REST API Endpoint
+
+> **Warning:** The GET /cloudstorages/id/content REST API Endpoint is deprecated.
+> We recommend that you use `retrieve_content_v2` method that matches to revised API when using SDK.
+> For backward compatibility, we continue to support the prior version of this API until version 2.5.0 is released.
```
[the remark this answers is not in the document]
Marishka17: I've updated it and added a small note to the API reference

cvat/schema.yml
```diff
@@ -6977,9 +7065,9 @@ components:
           $ref: '#/components/schemas/FileInfoTypeEnum'
         mime_type:
           type: string
+          nullable: true
```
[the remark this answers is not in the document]
Marishka17: You're right, I removed this change and added mime type for cloud storage data too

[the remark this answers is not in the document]
Marishka17: @bsekachev, FYI 
Could you please update the UI when you get the chance?

cvat/apps/engine/cache.py
```diff
@@ -58,11 +64,19 @@ def get_local_preview_with_mime(self, frame_number, db_data):
...
 
-    def get_cloud_preview_with_mime(self, db_storage):
-        item = self._get_or_set_cache_item(
-            key=f'cloudstorage_{db_storage.id}_preview',
-            create_function=lambda: self._prepare_cloud_preview(db_storage)
-        )
+    def get_cloud_preview_with_mime(
+        self,
+        db_storage,
+        just_check: bool = False
```
azhavoro, line 70: From my point of view, separate `get` method is better

cvat/apps/engine/models.py
```diff
@@ -861,6 +861,10 @@ def get_specific_attributes(self):
     def get_key_file_path(self):
         return os.path.join(self.get_storage_dirname(), 'key.json')
 
+    @property
+    def has_at_least_one_manifest(self) -> bool:
```
azhavoro, line 865: `manifest_count(self)` looks more logical

cvat/apps/engine/serializers.py
```diff
@@ -715,6 +716,13 @@ class DataSerializer(serializers.ModelSerializer):
         help_text="Uploaded files")
     server_files = ServerFileSerializer(many=True, default=[],
         help_text="Paths to files from a file share mounted on the server, or from a cloud storage")
+    server_files_exclude = serializers.ListField(required=False, default=[],
+        child=serializers.CharField(max_length=1024),
```
azhavoro, line 720: Doest it mean that path cannot be more than 1024 symbols? Why 1024 exactly?

cvat/apps/engine/serializers.py
```diff
@@ -795,6 +803,12 @@ def validate(self, attrs):
             and attrs['start_frame'] > attrs['stop_frame']:
             raise serializers.ValidationError('Stop frame must be more or equal start frame')
 
+        if (
+            (server_files := attrs.get('server_files'))
+            and attrs.get('cloud_storage_id')
+            and len(list(filter(lambda x: not x['file'].endswith('.jsonl'), server_files))) > settings.CLOUD_STORAGE_MAX_FILES_COUNT
```
azhavoro, line 809:
```suggestion
            and sum(1 for f in server_files if f['file'].endswith('.jsonl'))
            and len(list(filter(lambda x: not x['file'].endswith('.jsonl'), server_files))) > settings.CLOUD_STORAGE_MAX_FILES_COUNT
```

cvat/apps/engine/task.py
```diff
@@ -61,11 +62,37 @@ class SegmentsParams(NamedTuple):
...
     job.meta['status'] = 'Data are being copied from source..'
     job.save_meta()
 
+    # filter data from files/directories that should be excluded
+    if data_to_be_excluded:
+
+        expanded_server_files = []
+
+        for f in server_files:
+            path = Path(server_dir or settings.SHARE_ROOT) / Path(os.path.normpath(f))
```
azhavoro, line 81:
```suggestion
            path = Path(server_dir or settings.SHARE_ROOT) / Path(f).resolve()
```

cvat/apps/engine/task.py
```diff
@@ -61,11 +62,37 @@ class SegmentsParams(NamedTuple):
...
+            if not path.is_dir():
+                expanded_server_files.append(f)
+            else:
+                for root, _, files in os.walk(str(path)):
+                    if files:
+                        expanded_server_files.extend([os.path.join(f, os.path.relpath(root, str(path)), i) for i in files])
```
azhavoro, line 87: I think it would be better to rewrite this with `Path` as well

cvat/apps/engine/task.py
```diff
@@ -61,11 +62,37 @@ class SegmentsParams(NamedTuple):
...
+
+        server_files = expanded_server_files
+
+        server_files = list(filter(
+            lambda x: x not in data_to_be_excluded and f'{Path(x).parent}/' not in data_to_be_excluded,
```
azhavoro, line 92: please use Path functionality to compare filename paths

cvat/apps/engine/views.py
```diff
@@ -156,7 +157,8 @@ def share(request):
                         "mime_type": entry_mime_type,
                     })
 
-            serializer = FileInfoSerializer(many=True, data=data)
+            # return directories at the top of the list
+            serializer = FileInfoSerializer(many=True, data=sorted(sorted(data, key=lambda x: x['name']), key=lambda x: x['type']))
```
azhavoro, line 161: Can this be sorted by (x['name'], x['type']) tuple key?

utils/dataset_manifest/core.py
```diff
@@ -147,16 +150,18 @@ def __iter__(self):
 
 class DatasetImagesReader:
     def __init__(self,
```
azhavoro, line 152: Could you please add typing for all arguments? For now it's looking a little bit strange....

cvat/apps/engine/serializers.py
Marishka17: I used the same limit as for server_files

cvat/apps/engine/models.py
azhavoro: > We don't need the information about the number of attached manifests anywhere. We need to know if is there at least one.

If you return count value, the semantics of the function will not change, you can use it as before `if xxx.manifest_count(): ...` or just compare it with an int value.

azhavoro: If return count value we get a bit more universal method. For example, in the future I will need information about the total number of manifests. But I agree this is a insignificant comment, and the method doesnt affect the functionality of pr

cvat/apps/engine/cache.py
Marishka17: Updated

cvat/apps/engine/serializers.py
Marishka17: Thanks, changed to `sum(1 for f in server_files if not f['file'].endswith('.jsonl'))`

cvat/apps/engine/views.py
Marishka17: Thanks, updated to (x['type'], x['name'])

cvat/apps/engine/serializers.py
```diff
@@ -715,6 +716,13 @@ class DataSerializer(serializers.ModelSerializer):
         help_text="Uploaded files")
     server_files = ServerFileSerializer(many=True, default=[],
         help_text="Paths to files from a file share mounted on the server, or from a cloud storage")
+    server_files_exclude = serializers.ListField(required=False, default=[],
+        child=serializers.CharField(max_length=1024),
+        help_text=textwrap.dedent("""\
+            Paths to files and directories from a file share mounted on the server, or from a cloud storage
+            that should be excluded from the directories specified in server_files.
+        """)
```
zhiltsov-max, line 724: Please specify if it is compatible with 
- a single manifest file in the input list
- a single zip file contents
- the filename_pattern parameter

cvat/schema.yml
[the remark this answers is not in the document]
SpecLad: I don't think slashes would make sense here, there's no hierarchy.

If Maxim is to get rid of underscores, then we can settle on using dashes.

cvat/apps/engine/task.py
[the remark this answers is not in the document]
SpecLad: I know, but from a maintainability standpoint, if a function is supposed to return a value, then it should end with a return statement, otherwise it looks like you just forgot to return something.

https://peps.python.org/pep-0008/#programming-recommendations

cvat/apps/engine/views.py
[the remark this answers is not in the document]
Marishka17: Oh, thank you, for some reason I missed it when I looked at the spec

cvat/apps/engine/views.py
```diff
@@ -2105,7 +2112,84 @@ def content(self, request, pk):
...
+            prefix = request.query_params.get('prefix')
+            page_size = request.query_params.get('page_size', str(settings.BUCKET_CONTENT_MAX_PAGE_SIZE))
+            if not page_size.isnumeric():
+                return HttpResponseBadRequest('Wrong value for page_size was found')
+            page_size = min(int(page_size), settings.BUCKET_CONTENT_MAX_PAGE_SIZE)
+
+            # make api identical to share api
+            if prefix and prefix.startswith('/'):
+                prefix = prefix[1:]
+            delimiter = request.query_params.get('delimiter', '/')
```
SpecLad, line 2166: Do we actually need this parameter?

First, I don't see it being used anywhere in the UI.

Second, I think we would need extensive tests for a case where a storage bucket uses a delimiter other than '/', as I expect that a lot of things might break in this case (basically, any path manipulation code related to cloud paths is liable to break). I don't see any such tests right now.

Third, even if we do decide to support such a case, I think it would make more sense for the delimiter to be a property of the cloud storage object, rather than an operation parameter.

cvat/apps/engine/task.py
```diff
@@ -434,18 +492,77 @@ def _create_thread(
...
+                            dirs.append(f['name'])
+                    # we check the limit of files on each iteration to reduce the number of possible requests to the bucket
+                    if (len(additional_files) + len(dirs) + number_of_files) > settings.CLOUD_STORAGE_MAX_FILES_COUNT:
+                        raise ValidationError(
+                            'The maximum number of the cloud storage attached files '
+                            f'is {settings.CLOUD_STORAGE_MAX_FILES_COUNT}')
+            data['server_files'].extend(additional_files)
+            del additional_files
+
+            if data_to_be_excluded := data.get('server_files_exclude'):
```
SpecLad, line 544:
```suggestion
            if server_files_exclude := data.get('server_files_exclude'):
```

cvat/apps/engine/task.py
```diff
@@ -61,11 +62,37 @@ class SegmentsParams(NamedTuple):
...
+                expanded_server_files.append(f)
+            else:
+                for root, _, files in os.walk(str(path)):
+                    if files:
+                        expanded_server_files.extend([str(Path(f) / Path(root).relative_to(path) / Path(i)) for i in files])
+
+        server_files = expanded_server_files
+
+        server_files = list(filter(
+            lambda x: x not in data_to_be_excluded and f'{Path(x).parent}/' not in data_to_be_excluded,
```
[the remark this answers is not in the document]
bsekachev: @SpecLad 

> Moreover, x is a path in the local syntax, whereas data_to_be_excluded always contains Unix paths.

Actually, the problem exists not only for ``data_to_be_excluded``, but probably also for ``server_files``. And for browsing shared storage, because client always uses and returns ''/'' as a separator. 

Since it is not a new in this patch, and looks like requiring significant efforts to resolve, I would suggest to create an issue and consider it as a different tasks. BTW, I personally don't remember someone claims about it.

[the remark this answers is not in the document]
Marishka17: > I'm not sure this is correct, as x is (I think?) an absolute path, while data_to_be_excluded contains paths relative to the share root. Moreover, x is a path in the local syntax, whereas data_to_be_excluded always contains Unix paths.

No, x is a relative path.

> It's not sufficient to only check Path(x).parent, since a path might be excluded by its grandparent (or great-grandparent, etc...).

Fixed

> I don't think we need to require excluded items that are directories to end with a slash. Is there any benefit from doing so?

Yes, in this case, they are easier to handle and differentiate.

cvat/apps/engine/task.py
```diff
@@ -61,11 +62,31 @@ class SegmentsParams(NamedTuple):
...
     job = rq.get_current_job()
     job.meta['status'] = 'Data are being copied from source..'
     job.save_meta()
 
+    # filter data from files/directories that should be excluded
+    if server_files_exclude:
+        for f in server_files:
+            path = Path(server_dir or settings.SHARE_ROOT) / Path(f)
+            if not path.is_dir():
+                server_files.append(f)
```
SpecLad, line 80: `server_files` should not be modified, as it is a) an input parameter, and b) the list you're iterating over.

cvat/apps/engine/task.py
```diff
@@ -61,11 +62,31 @@ class SegmentsParams(NamedTuple):
...
+            else:
+                server_files.remove(f)
+                server_files.extend([str(f / i.relative_to(path)) for i in path.glob('**/*') if i.is_file()])
+
+        server_files = list(filter(
+            lambda x: x not in server_files_exclude and all([f'{Path(i)}/' not in server_files_exclude for i in Path(x).parents]),
```
SpecLad, line 86: The `Path` cast is unnecessary, as `i` is already a path.

cvat/apps/engine/task.py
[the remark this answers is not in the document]
SpecLad: > Yes, in this case, they are easier to handle and differentiate.

Are they, though? All I can see is that you have to append an extra slash every time you want to check if a directory is in `server_files_exclude`.

cvat/apps/engine/task.py
```diff
@@ -61,12 +62,34 @@ class SegmentsParams(NamedTuple):
...
 
-    for path in server_files:
+    filtered_server_files = server_files.copy()
+
+    # filter data from files/directories that should be excluded
+    if server_files_exclude:
+        for f in server_files:
+            path = Path(server_dir or settings.SHARE_ROOT) / Path(f)
+            if not path.is_dir():
+                filtered_server_files.append(f)
```
SpecLad, line 82: This seems to add a duplicate element to `filtered_server_files` (since `filtered_server_files` is initialized from `server_files.copy()`, `f` must already be in it).

cvat/apps/engine/task.py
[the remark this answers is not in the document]
Marishka17: I guess that it will help on the client side when the list with all files inside the directory will be unchecked and added to `server_files_exclude` as dir. The one approach should be used for `server_files` and `server_files_exclude`.
And the directory in the server_files should be ended with a slash because share directory can be handled without slash but cloud storage "directory" without slash is just a prefix, not directory.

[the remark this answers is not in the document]
SpecLad: Okay, that makes sense.

SpecLad (approved):

bsekachev:
@lubin-liu 

Roughly saying each request to S3 costs money. In case of using manifest, when you browse the cloud storage from CVAT, this is for free, because it uses data from manifest.

bsekachev:
And browsing itself works faster

PR #6075, opened 2023-04-26
Fix skeleton tracks

kirill-sizov:
### Motivation and context
Currently, we don't have validation of incoming annotations, as a result, there is exist some cases when annotations successfully saved in DB, but it's impossible to export them. In order to successfully export a dataset with a skeleton track it's required that each track satisfy the following condition:
` {frame number of track} == {frame number of parent track} == {frame number of the first shape of the track}`

This PR adds an additional step during saving annotation in DB. This additional step check that all these there "frame numbers" are equal and try to automatically fix it's not true.

### Checklist

~- [ ] I have linked related issues (see [GitHub docs](
  https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword))~

### License


cvat/apps/dataset_manager/task.py
```diff
@@ -115,47 +116,87 @@ def __init__(self, pk, is_prefetched=False):
     def reset(self):
         self.ir_data.reset()
 
+    def _validate_attribute_for_existence(self, db_attr_val, label_id, attr_type):
+        if db_attr_val.spec_id not in self.db_attributes[label_id][attr_type]:
+            raise AttributeError("spec_id `{}` is invalid".format(db_attr_val.spec_id))
```
zhiltsov-max, line 121: Shouldn't it become a Validation error at some point?

cvat/apps/dataset_manager/task.py
```diff
@@ -115,47 +116,87 @@ def __init__(self, pk, is_prefetched=False):
...
 
             for track in tracks:
                 track_attributes = track.pop("attributes", [])
                 shapes = track.pop("shapes")
                 elements = track.pop("elements", [])
                 db_track = models.LabeledTrack(job=self.db_job, parent=parent_track, **track)
-                if db_track.label_id not in self.db_labels:
-                    raise AttributeError("label_id `{}` is invalid".format(db_track.label_id))
+
+                self._validate_label_for_existence(db_track.label_id)
```
zhiltsov-max, line 171: If the checks are performed on saving, how the existing saved problems will be handled?

tests/python/rest_api/test_tasks.py
```diff
@@ -466,6 +466,94 @@ def test_can_export_task_with_several_jobs(self, admin_user, tid, format_name):
...
+                }
+            ],
+        }
+        response = patch_method(
+            admin_user, f"tasks/{tid}/annotations", annotations, action="update"
+        )
+        assert response.status_code == HTTPStatus.OK
+
+        response = self._test_export_task(admin_user, tid, format="COCO Keypoints 1.0")
+        assert response.data
```
zhiltsov-max, line 555: 1. If there are errors in the annotations that are ignored - probably, the results written should be checked for this
2. If there are errors that should not be ignored - probably, it should be checked in the response status

cvat/apps/dataset_manager/task.py
kirill-sizov: This is not a new check, I just moved this code in a separate function. We don't have existing saved problems related to this check.

kirill-sizov: If you mean another check that I added in this PR, you are right the existing save problems still exist. But in this case, it's possible to back up the task and upload it again. For sure we can write migration, but such migration will take a really long time.

zhiltsov-max: > If you mean another check that I added in this PR, you are right the existing save problems still exist. 

Ok, probably I've missed this.

> For sure we can write migration, but such migration will take a really long time.

Ok, but I thought about adding some checks during the export or track conversion to shapes instead. What do you thinks about this option?

kirill-sizov: Yes, performing this check during the export also will resolve problem, but personally I don't like the idea of storing invalid annotations in db

nmanovic:
@kirill-sizov , why did cvat_data.tar.bz2 grow in 3 times?

kirill-sizov:
> @kirill-sizov , why did cvat_data.tar.bz2 grow in 3 times?

Fixed

nmanovic (approved):

PR #6082, opened 2023-04-27
Fix issue azure.core.exceptions.ResourceExistsError: The specified blob already exists.

Marishka17:
### Motivation and context

The problem is that If a user tries to export annotations to the Azure container and the file with such name already exists, then an exception will be raised by default. This is not suitable for us, because our side should not be affected by this and we should have the same behavior for different CS providers.
### How has this been tested?

Manually
### Checklist

### License

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - TDB
 
 ### Fixed
-- TDB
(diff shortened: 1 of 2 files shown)


nmanovic:
@Marishka17 , could you please add a line into CHANGELOG.md? Also it will be great to know more about the problem.

Marishka17:
@nmanovic, The problem is that If a user tries to export annotations to the Azure container and the file with such name already exists, then an exception will be raised by default. This is not suitable for us, because our side should not be affected by this and we should have the same behavior for different CS providers.

PR #6108, opened 2023-05-05
Fix navigation when tracking

elinep:
Include 'navigationBlocked' flag in AnnotationTopBarContainer so that a refresh through componentDidUpdate()->play() is triggered when a tool (ex: opencv tracker) stops blocking navigation with switchNavigationBlocked(true)

Change-Id: I11db2cb5eefb2f41f6bd865f0d9bb0ca76603fee

### Motivation and context
Navigation stalls when tracking tool are activated (see https://github.com/opencv/cvat/issues/5960)

### Checklist

### License


bsekachev:
Hello, I do not think this logic with passing ``navigationBlocked`` to the component only to update it is good approach.  It is too implicit. 

Also removing ``onSwitchPlay(false);`` breaks other cases.

 Probably we need to find another solution. Anyway, thank you.

elinep:
> Hello, I do not think this logic with passing `navigationBlocked` to the component only to update it is good approach. It is too implicit.

I agree.

> Also removing `onSwitchPlay(false);` breaks other cases.

Could you be more specific about these cases ? I could not think/find any. 

Thank you for your time.

PR #6137, opened 2023-05-12
multiple fixes to make helm work

### LINKED ISSUE #5733: Can't disable traefik ingress controller without removing ingress object

### LINKED ISSUE #6043: [Helm] multiples fixes needed
Hi, I cannot create another PR right now, or update current one, but I decided at least to write down my chart concerns in case someone else could fix that:

1. Treafik is hardcoded to values right now and that is bad, because it is not standard package and people could use other software as ingress(nginx/istio/ambassador/etc)
2. Previously there was a way to specify psql db creds via secret, that was for some reason removed in new versions. That is important because some people(including me) using Terraform to create db in cloud and provirde k8s secret to pass creds to cvat
3. Right now I see, that there is 5-6 deployments of cvat deployed by default - are you sure, that I need all of them just to start cvat on my cluster? My guess that at least part of them should be opt-in and switched off by default. I see that compose with https uses only one!
4.  Right now there is only one PVC for all of the deployments, I think that will not work from the box for almost anybody, who tried to use this chart

### LINKED ISSUE #6096: Helm Chart Volume Configuration "ReadWriteOnce" Doesn't Work with Multiple Pods
- [X] Read/searched [the docs](https://github.com/cvat-ai/cvat/tree/master#documentation)
- [X] Searched [past issues](/issues)

When using the Helm Chart, the storage.yml ReadWriteOnce option under the backend folder structure doesn't work in my K8s environment. If you have deployments across multiple nodes, ReadWriteOnce will form a race condition causing only one backend pod to start successfully. ReadWriteOnce by definition only allows one node to mount - so how is this supposed to work? 

1. Attempt to deploy the helm chart to Kubernetes with the default storage

### Expected Behaviour
1. The defaults for the helm chart should be able to launch CVAT in K8s with the minimal amount of modification. 
### Possible Solution
1. Changing storage.yml to ReadWriteMany 
2. Put all backend deployments into the same node (maybe there's a reason this isn't designed this way though)

Keramblock:
### Motivation and context

Right now helm chart is broken and not usable at least in my environment, I trying to fix it to make it work
content: 

1. Moved test-related values to another values.file to separate it from default config
2. fixed issue with multiple caches in same RWX volume, which prevents db migration to start
3. Removed hardcoded mandatory traefik ingress usage
4. Added confugurable default storage option to chart

### How has this been tested?

We test it on our AKS with RWX volume

### Checklist

### License

closes #6043 
closes #6096 
closes #5733


Keramblock:
@mdacoca if you could also help with the review and helm tests - that would be great =)

azhavoro:
@Keramblock Hi, Thanks for the contributing! I will test the PR this week.

Keramblock:
@azhavoro have you had a chance to take a look?

helm-chart/values.yaml
```diff
@@ -91,6 +91,9 @@ cvat:
             name: http
     defaultStorage:
         enabled: true
+        storageClassName: default
+        accessModes:
+         - ReadWriteOnce
```
DreamingRaven, line 96: Correct me if I am wrong, I was testing the cvat helm chart out today, and I believe this should probably be ReadWriteMany as at-least the original chart before your changes would spin up multiple pods seeking to access the same volume by PVC name. I dont know how cvat acts when multiple pods can read and write at once, but without this set to ReadWriteMany only one of the pods like utils or server will get this volume and the rest will be waiting for it indefinitely since they cant bind to it any more.

Keramblock: Hi, I have limited knowledge of what is happening internally, so I left it as it is and I'm waiting for input from @azhavoro or @mdacoca to see if it's okay to change it by default.

In my opinion, the current chart may not work as it is, and we might need to change the default access mode. However, I'm almost certain that it would not help because the default storage class usually does not support this access mode either.

P.S. Thank you very much for reviewing! I appreciate it.

DreamingRaven: Ah I see, for me I back my kubernetes cluster with rook so I have a default storage class that supports RWX but yeah I get what you mean, there does not seem to be a good solution for everyone but I think RWO is broken for everyone.

Also on a related note I might not assume the users default storage class, In my case I use [rook](https://rook.io/) which gives me these:
```
NAME                        PROVISIONER                     RECLAIMPOLICY   VOLUMEBINDINGMODE   ALLOWVOLUMEEXPANSION   AGE
ceph-block                  rook-ceph.rbd.csi.ceph.com      Delete          Immediate           true                   26h
ceph-bucket                 rook-ceph.ceph.rook.io/bucket   Delete          Immediate           false                  26h
ceph-filesystem (default)   rook-ceph.cephfs.csi.ceph.com   Delete          Immediate           true                   26h
```
Im not 100% certain but I believe if in the PVC you dont specify the storage class that it will auto populate it with whatever the cluster has as a default no matter what it is called. This might be more universal if you put this behind and if condition so if the user sets a storageClassName it will use it, and if not it will leave it blank in the CRD for kubernetes to fill it in. Something like {{ if eq .Values.cvat.backend.defaultStorage.storageClassName "" }} maybe or {{ if empty .Values.cvat.backend.defaultStorage.storageClassName }} (this was only quick off the top of my head may need editing the condition to invert it).

Also I wanted to say its no problem, I have a vested interest in this too so I would love to see your PR be merged as it works a lot better than the current one which is a bit broken in a few places.

Keramblock: Thanks a lot, added)

DreamingRaven: That's beautiful thanks for the changes!!! This now has made ArgoCD happier as well since it understands the value is now auto-filled by kube. Thanks a lot for your additions, super appreciated!
cvat
I will keep testing the chart for you and let you know if there are any other minor changes but it all looks good currently.

helm-chart/values.yaml
```diff
@@ -24,7 +24,7 @@ cvat:
       additionalVolumeMounts: []
     worker:
       export:
-        replicas: 2
+        replicas: 1
```
azhavoro, line 27: Why did you change the default value? From my point of view, it is better to have a similar number of workers for k8s and docker-compose

helm-chart/values.yaml
```diff
@@ -34,7 +34,7 @@ cvat:
         additionalVolumes: []
         additionalVolumeMounts: []
       import:
-        replicas: 2
+        replicas: 1
```
azhavoro, line 37: the same comment

helm-chart/values.yaml
```diff
@@ -225,7 +228,7 @@ nuclio:
 #      password: somepass
 
 analytics:
-  enabled: true
+  enabled: false
```
azhavoro, line 231: Analytics is currently enabled by default, please revert the change

helm-chart/values.yaml
```diff
@@ -319,62 +322,58 @@ grafana:
...
-            name: backend-service
-            port: 8080
-        - path : /
-          pathType: "Prefix"
-          service:
-            name: frontend-service
-            port: 80
+  enabled: false
+#  enabled: true
+#  annotations:
```
azhavoro, line 327: `ingress.enabled = true` does nothing, why rules should be commented by you opinion?

azhavoro (approved):

helm-chart/values.yaml
Keramblock: will do

helm-chart/values.yaml
Keramblock: Because there are some lists in that config and they are hard to override with user values files.
For example "paths" list  here will be very hard to override for me to use another Ingress provider like nginx, or ambassador, or gke lb

Keramblock: my config just for example: 
image

Keramblock: also ingress.enblaed is used here, for example: 
изображение

Keramblock: also ingress.enabled is used here: 
изображение

helm-chart/values.yaml
eremeevfd: We can drop a comment why do we have 2 replicas by default maybe? Just for people to know what it is and why they should or shouldn't change it

helm-chart/values.yaml
eremeevfd: @azhavoro @Keramblock Maybe we should add ImplementationSpecific to default values?

eremeevfd: @Keramblock Could you please add comment to commented values in ingress section, smth like "In case you need an ingress, don't forget to uncomment paths, or write them manually in your values"

Keramblock: @azhavoro also I moved traefik configuration to tests.values.yml just in case you will need it for your internal stuff, so that change will not interrupt you workflow.

eremeevfd: @azhavoro The problem with default paths with predefined host "cvat.local" that if you specify your own host and other configs for paths, then you get:
```yaml
hosts:
  - host: cvat.local
     paths:
       - path: /api
   - host: my-cvat-host.example.com
      paths:
        - path: /api
```
It is a bit ugly and contains unused values, I think it'd be better to preserve default paths maybe and allow to specify host (as basic config)

eremeevfd: Also, according to this docs: https://docs.nginx.com/nginx-ingress-controller/configuration/ingress-resources/basic-configuration/#new-features-available-in-kubernetes-118-and-above
ImplementationSpecific is a default value, so we can just omit it here

Keramblock: @eremeevfd default values is for cvat.local host. I doubt that will be usefull for other users.

Keramblock: > @azhavoro @Keramblock Maybe we should add ImplementationSpecific to default values?

default values is for cvat.local host. I doubt that will be usefull for other users.

helm-chart/values.yaml
Keramblock: done

eremeevfd (approved):

nmanovic:
@Keramblock , thanks for the contribution! Could you please fix remark linter? You use `[helm]` prefix in CHANGELOG.md. It makes the linter unhappy. You need to escape `[ ]` or remove the prefix.

Keramblock:
> @Keramblock , thanks for the contribution! Could you please fix remark linter? You use `[helm]` prefix in CHANGELOG.md. It makes the linter unhappy. You need to escape `[ ]` or remove the prefix.

sure, will fix that in 10 minutes

Keramblock:
Done

Keramblock:
@nmanovic @azhavoro @mdacoca could you please take a look?

Keramblock:
@nmanovic @azhavoro @mdacoca am I doing something wrong? I trying to help, but it looks like no one is interested in that.

azhavoro:
> @nmanovic @azhavoro @mdacoca am I doing something wrong? I trying to help, but it looks like no one is interested in that.

@Keramblock Hi, sorry for the delay, this is great PR, but the helm tests are fading, I'm going to take a look, but I'm a little busy right now.

DreamingRaven (approved):
allgood
I wish I could help here at all, but for what its worth, everything seems to be working on my end!
I hope they get around to this soon!

Keramblock:
Thanks a lot!

Keramblock:
@azhavoro any news?

nmanovic:
@Keramblock , could you please resolve the current conflicts? Please also look at problems with the linter.

Keramblock:
@nmanovic  I tried to fix helm checks, but it sadly didnt succeed in that with following error:
` Command failed: kubectl exec release-5598438317-1-backend-server-8568c4654b-ztdkz -- tar --strip 3 -xjf /tmp/cvat_data.tar.bz2 -C /home/django/data/.

Error message: Defaulted container "cvat-app-backend-server-container" out of: cvat-app-backend-server-container, user-data-permission-fix (init)
tar: cache: Cannot utime: Operation not permitted
tar: cache: Cannot change mode to rwxr-xr-x: Operation not permitted`
I have limited knowledge of github ci and your CI infrastructure, so I need someones help with that.

Keramblock:
I think I found source of problems - looks like that is old values file, that for some reason located outside of chart dir, that was not updated. I will try to fix it

Keramblock:
That looks like last PR to develop just broke docker build

kirill-sizov:
> That looks like last PR to develop just broke docker build

Hi, @Keramblock Actually that [last PR](https://github.com/opencv/cvat/pull/6514) isn't related to Helm tests at all. It seems that it was a temporary connection problem since as you can see in the [log](https://github.com/opencv/cvat/actions/runs/5599098812/jobs/10239618997) builder wasn't able to download the library from GitHub.

I've restarted the workflow for this PR let's see how it goes

Keramblock:
Thanks a lot for the help!

azhavoro:
@Keramblock Hi, there is a patch that fixes our tests
[fix.zip](https://github.com/opencv/cvat/files/12142849/fix.zip)

Keramblock:
> @Keramblock Hi, there is a patch that fixes our tests [fix.zip](https://github.com/opencv/cvat/files/12142849/fix.zip)

@azhavoro thanks a lot! will commit that today.

Keramblock:
@azhavoro  I think it is ready for review

azhavoro (approved):

Keramblock:
@azhavoro @nmanovic @kirill-sizov  thanks a lot for the help!

PR #6129, opened 2023-05-11
Modernize OpenVINO-based Nuclio functions and allow them to run on Kubernetes

SpecLad:
### Motivation and context

Currently, OpenVINO-based functions assume that a local directory will be mounted into the container. In Kubernetes, that isn't possible, so implement an alternate approach: create a separate base image and inherit the function image from it.

In addition, implement some modernizations:

* Upgrade the version of OpenVINO to the latest (2022.3). Make the necessary updates to the code. Note that 2022.1 introduced an entirely new inference API, but I haven't switched to it yet to minimize changes.

* Use the runtime version of the Docker image as the base instead of the dev version. This significantly reduces the size of the final image (by ~3GB).

* Replace the `faster_rcnn_inception_v2_coco` model with `faster_rcnn_inception_resnet_v2_atrous_coco`, as the former has been
  removed from OMZ.

* Ditto with `person-reidentification-retail-0300` -> `0277`.

* The IRs used in the DEXTR function are not supported by OpenVINO anymore (format too old), so rewrite the build process to create them from the original code/weights instead.

### How has this been tested?

I manually tried each affected function to make sure they still work.

### Checklist

- [x] I submit my changes into the `develop` branch
- [ ] I have added a description of my changes into the [CHANGELOG](https://github.com/opencv/cvat/blob/develop/CHANGELOG.md) file
- ~~[ ] I have updated the documentation accordingly~~
- ~~[ ] I have added tests to cover my changes~~
- ~~[ ] I have linked related issues (see [GitHub docs](
  https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword))~~
- ~~[ ] I have increased versions of npm packages if it is necessary
  ([cvat-canvas](https://github.com/opencv/cvat/tree/develop/cvat-canvas#versioning),
  [cvat-core](https://github.com/opencv/cvat/tree/develop/cvat-core#versioning),
  [cvat-data](https://github.com/opencv/cvat/tree/develop/cvat-data#versioning) and
  [cvat-ui](https://github.com/opencv/cvat/tree/develop/cvat-ui#versioning))~~

### License

- [x] I submit _my code changes_ under the same [MIT License](
  https://github.com/opencv/cvat/blob/develop/LICENSE) that covers the project.
  Feel free to contact the maintainers if that's a concern.

PR #6100, opened 2023-05-03
Ensure that Nuclio function names are valid Kubernetes service names

SpecLad:
### Motivation and context

When deployed to Kubernetes, Nuclio creates a Kubernetes service for each function,
and service names can't contain underscores or periods.

### Checklist

- [x] I submit my changes into the `develop` branch
- ~~[ ] I have added a description of my changes into the [CHANGELOG](https://github.com/opencv/cvat/blob/develop/CHANGELOG.md) file~~
- ~~[ ] I have updated the documentation accordingly~~
- ~~[ ] I have added tests to cover my changes~~
- ~~[ ] I have linked related issues (see [GitHub docs](
  https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword))~~
- ~~[ ] I have increased versions of npm packages if it is necessary
  ([cvat-canvas](https://github.com/opencv/cvat/tree/develop/cvat-canvas#versioning),
  [cvat-core](https://github.com/opencv/cvat/tree/develop/cvat-core#versioning),
  [cvat-data](https://github.com/opencv/cvat/tree/develop/cvat-data#versioning) and
  [cvat-ui](https://github.com/opencv/cvat/tree/develop/cvat-ui#versioning))~~

### License

- [x] I submit _my code changes_ under the same [MIT License](
  https://github.com/opencv/cvat/blob/develop/LICENSE) that covers the project.
  Feel free to contact the maintainers if that's a concern.

PR #6140, opened 2023-05-12
Make serverless function naming and Docker image naming consistent

SpecLad:
For function names, take the relative path, lowercase, join with dashes and replace underscores with dashes.

For image names, take the relative path, lowercase and join with dots.

In a couple cases, rename the function directory instead of fixing the config.

### Motivation and context

All these inconsistencies are driving me batty. Derived names should be derived the same way everywhere in order to be predictable.

### Checklist

### License

- [x] ] I submit _my code changes_ under the same [MIT License](


SpecLad:
The idea is originally from #6129, but I fixed all functions this time, not just OpenVINO ones.

Here's the script I used to detect mismatches:

```python
import yaml

from pathlib import Path

for config_path in Path('serverless').glob("**/*.yaml"):
    config_rel = config_path.relative_to('serverless')

    assert config_rel.parent.name == 'nuclio'

    parts = list(config_rel.parents[1].parts)

    if parts[0] == 'pytorch':
        parts[0] = 'pth'
    elif parts[0] == 'tensorflow':
        parts[0] = 'tf'

    with config_path.open('rb') as config_file:
        config = yaml.safe_load(config_file)

    expected_name = '-'.join(part.replace('_', '-').lower() for part in parts)
    actual_name = config['metadata']['name']

    if actual_name != expected_name:
        print(f'{config_path}: name mismatch: expected {expected_name!r} actual {actual_name!r}')

    expected_image_name = 'cvat.' + '.'.join(part.lower() for part in parts)
    actual_image_name = config['spec']['build']['image']

    if actual_image_name != expected_image_name:
        print(f'{config_path}: image name mismatch: expected {expected_image_name!r} actual {actual_image_name!r}')
```

Probably should add something like this to CI, but no time for that right now.

bsekachev:
Quick look at our documentation shows that we still have some references using previous naming. 
For example:

image

Also, If we rename SAM, we need to update our SAM article: https://www.cvat.ai/post/facebook-segment-anything-model-in-cvat (better to collaborate with Maria to know if we published it somewhere else).

And just thoughts, this patch will result our users have several deployed models with the same name in CVAT. Maybe it is not a big problem.

SpecLad:
> Quick look at our documentation shows that we still have some references using previous naming.

I believe those are all referring to the TensorFlow version (which hasn't changed).

> Also, If we rename SAM, we need to update our SAM article

Yeah, that's a fair point. I don't want to deal with it right now, so I just reverted the SAM changes. I might return to it later.

> And just thoughts, this patch will result our users have several deployed models with the same name in CVAT. Maybe it is not a big problem.

I don't think it's a big problem (especially given that I already updated a whole bunch of names in #6100), but just in case, I added a changelog entry describing the name changes.

nmanovic (approved):

PR #6150, opened 2023-05-15
Fix the CPU HRNet Nuclio function

SpecLad:
### Motivation and context

It was broken for two reasons:

* Due to some changes on the PyTorch website, the old way of installing the PyTorch packages now installs the ROCm version rather than the CPU version (and it doesn't work doe to a missing dependency).

* The newest version of NumPy doesn't work with HRNet due to the latter's usage of `np.int`.

Fix these problems, and in addition, rework the build recipe to avoid installing unneeded packages. Altogether, the changes massively shrink the Docker image size (from ~14 GB to ~2 GB).

I didn't update the GPU version, because a) the first issue doesn't affect it, b) the second issue is already fixed in it, and c) I don't have a GPU to test it on.

### How has this been tested?

Manual testing with CVAT.

### Checklist

### License


bsekachev (approved):
LGTM

PR #6195, opened 2023-05-24
Switch to uvicorn

azhavoro:
### Checklist

### License


Dockerfile
```diff
@@ -172,10 +172,12 @@ RUN if [ "${CVAT_DEBUG_ENABLED}" = 'yes' ]; then \
     fi
 
 # Install and initialize CVAT, copy all necessary files
+COPY --chown=${USER} cvat/nginx.conf /etc/nginx/nginx.conf
+COPY --chown=${USER} backend_entrypoint.sh /
```
SpecLad, line 176: Why is this in `/` and not `$HOME`, like the others?

Dockerfile
```diff
@@ -172,10 +172,12 @@ RUN if [ "${CVAT_DEBUG_ENABLED}" = 'yes' ]; then \
     fi
 
 # Install and initialize CVAT, copy all necessary files
+COPY --chown=${USER} cvat/nginx.conf /etc/nginx/nginx.conf
```
SpecLad, line 175: This should probably be owned by root, if it's in `/etc`.

Dockerfile
```diff
@@ -164,6 +162,8 @@ RUN --mount=type=bind,from=build-image,source=/tmp/wheelhouse,target=/mnt/wheelh
 ENV NUMPROCS=1
 COPY --from=build-image-av /opt/ffmpeg/lib /usr/lib
 
+RUN chown -R ${USER} /var/lib/nginx /var/log/nginx
```
SpecLad, line 165: `/var/log/nginx` shouldn't need to be modified, since you overrode the log file locations to be in `/home/django`. I would also suggest to override the location of whatever it is that is stored to `/var/lib/nginx`, to be stored in `/home/django` too, for consistency.

cvat/nginx.conf
```diff
@@ -0,0 +1,80 @@
...
+      gzip_http_version 1.1;
+      gzip_types
+          text/html
+          application/javascript
+          application/x-javascript
+          text/javascript
+          application/wasm
+          image/x-icon;
+
+      root /home/django;
```
SpecLad, line 52: I'm not familiar with nginx configuration, but doesn't that mean that _everything_ under `/root/django` will be served, including source code?

Dockerfile
```diff
@@ -5,7 +5,6 @@ FROM ${BASE_IMAGE} as build-image-base
 
 RUN apt-get update && \
     DEBIAN_FRONTEND=noninteractive apt-get --no-install-recommends install -yq \
-        apache2-dev \
```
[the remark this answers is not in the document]
azhavoro: We don't use apache for dev env

Dockerfile
azhavoro: fixed

cvat/nginx.conf
azhavoro: Shouldn't be true because 'location' is not '\' here but changed to the `alias` directive instead of `root`

Dockerfile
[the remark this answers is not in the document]
SpecLad: FYI, the reason is #5889 (although I should really revert it now that we're using PostgreSQL in testing...).

cvat/nginx.conf
SpecLad: Okay, I see.

SpecLad (approved):

PR #6204, opened 2023-05-26
Honey pot server

zhiltsov-max:
### Motivation and context

The server part of #6039 

- Added support for Ground Truth jobs in a task
- Added support for job creation and removal (only Ground Truth jobs can be created or removed in a task)
- Added a component to autocompute quality metrics for a task
- Added tests
- Fixed https://github.com/opencv/cvat/issues/5971 (both parts - the outside problem and the manifest problem, the manifest part fix is also available in #6216)

### Checklist

### License


azhavoro:
@zhiltsov-max could you please update `cvat/schema.yml` ?

zhiltsov-max:
@azhavoro, it is actual, do you have any problems with it?

azhavoro:
> @azhavoro, it is actual, do you have any problems with it?


https://github.com/opencv/cvat/actions/runs/5113276424/jobs/9192480263?pr=6204

```
Loaded image: cvat/server:latest
diff --git a/cvat/schema.yml b/cvat/schema-expected.yml
index c087[21](https://github.com/opencv/cvat/actions/runs/5113276424/jobs/9192480263?pr=6204#step:10:22)2..6f8aa18 100644
--- a/cvat/schema.yml
+++ b/cvat/schema-expected.yml
@@ -3694,6 +3694,7 @@ paths:
           - br
           - bs
           - ca
+          - ckb
           - cs
           - cy
           - da
@@ -3746,6 +3747,7 @@ paths:
           - ml
           - mn
           - mr
+          - ms
           - my
           - nb
           - ne
@@ -5821,6 +58[23](https://github.com/opencv/cvat/actions/runs/5113276424/jobs/9192480263?pr=6204#step:10:24),7 @@ components:
         chunk_size:
           type: integer
           readOnly: true
+          nullable: true
         size:
           type: integer
           readOnly: true
@@ -6977,6 +6980,7 @@ components:
           type: string
           format: date-time
           readOnly: true
+          nullable: true
         role:
           allOf:
           - $ref: '#/components/schemas/RoleEnum'
@@ -87[64](https://github.com/opencv/cvat/actions/runs/5113276424/jobs/9192480263?pr=6204#step:10:65),6 +87[68](https://github.com/opencv/cvat/actions/runs/5113276424/jobs/9192480263?pr=6204#step:10:69),7 @@ components:
         status_code:
           type: integer
           readOnly: true
+          nullable: true
         redelivery:
           type: boolean
           readOnly: true

API schema has changed! Please update cvat/schema.yml:

  docker run --rm --entrypoint /bin/bash cvat/server:dev \
    -c 'python manage.py spectacular' > cvat/schema.yml
Error: Process completed with exit code 1.
```

azhavoro:
1. I see a 404 error message every time I open the `view analytics` page.
   image
1. Last update mismatch. Does this mean that Job update time and GT update time are not the same? Have I missed smth?
   image
   image
1. Create a task -> create a GT job -> annotate GT job and set `acceptance` stage. N/A quality in this case, this is expected?
   image
1. The number of objects from the GT job is present in `Job Info` statistics. it looks strange...
1. The scenario is : `Create job -> create GT job -> annotate GT job with 1 object and set the `acceptance` stage -> Open job and annotate the same object but with a different label -> save annotations -> wait for quality report. I repeated this scenario twice and the report was only created for the first task. For task 2 I still have N/A quality, but Job and GT job are annotated... 
 1. UI: it seems that some GT Job frames are not displaying correctly: I only see 4 frames out of 17 correctly and a black image for the others. I don't know the exact steps to reproduce the issue, but this has happened to me several times. 
1. "frame_share_percent": 0.07331378299120235, which is not actually a percentage.
1. Name of the downloaded report can contain task_id and job_id. Currently it is just `quality-report-{some_number}`.

zhiltsov-max:
@azhavoro, 

> 1. I see a 404 error message every time I open the view analytics page.

I suppose, it's because the UI has not been updated yet to use the latest changes in the server API (the place where the task quality settings are obtained). At this point, it is expected. @klakhov 

> 2. Last update mismatch. Does this mean that Job update time and GT update time are not the same? Have I missed smth?

Hmm, looks like another UI issue. I can see in my setup that there is the /api/jobs response with some task times, and they do not correspond to what is shown in the UI. @klakhov 

> 3. Create a task -> create a GT job -> annotate GT job and set acceptance stage. N/A quality in this case, this is expected?
> The scenario is ...

Please try to move the gt job into the **Completed** state as well:
image

Probably, we need to update the help message. @klakhov 

> 4. The number of objects from the GT job is present in Job Info statistics. it looks strange...

I'm not sure I understand the problem correctly, there is no such field in the API responses. Where I can find it?

> 6. UI: it seems that some GT Job frames are not displaying correctly: I only see 4 frames out of 17 correctly and a black image for the others. I don't know the exact steps to reproduce the issue, but this has happened to me several times.

The black frames are expected to be in the places of the excluded frames in the job segment, currently such frames can be reached by direct navigation to the frame id in the player:
image

Or by using the Play button in the player in the GT job. It should be fixed in UI (@klakhov). Could you please check that only the valid frames are displayed and none of the included frames are black? How many GT frames was specified for the GT job?

> 7. "frame_share_percent": 0.07331378299120235, which is not actually a percentage.

Please explain it in more detail. Do you suggest to multiply it by 100?

> 8. Name of the downloaded report can contain task_id and job_id. Currently it is just quality-report-{some_number}.

@klakhov, please give the downloaded file a more descriptive name.

.vscode/launch.json
```diff
@@ -199,21 +219,44 @@
             "console": "internalConsole"
         },
         {
-            "name": "server: RQ - scheduler",
+            "name": "server: RQ - quality reports",
```
kirill-sizov, line 222: Why we don't add new workers in `server: debug` vscode compound?

kirill-sizov (commented):
Here is my small comments after I viewed about half of the PR

cvat/apps/quality_control/views.py
```diff
@@ -0,0 +1,333 @@
...
+from cvat.apps.iam.permissions import (
+    AnnotationConflictPermission, QualityReportPermission, QualitySettingPermission
+)
+
+
+@extend_schema(tags=["quality"])
+@extend_schema_view(
+    list=extend_schema(
+        summary='Method returns a paginated list of annotation conflicts',
+        parameters=[
```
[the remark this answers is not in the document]
zhiltsov-max: They are in the schema already even without this, so I removed extra declarations. I don't know if it's supposed by the IAM code, however.

cvat/apps/quality_control/signals.py
```diff
@@ -0,0 +1,60 @@
...
+        tasks.append(instance.segment.task)
+    elif isinstance(instance, Annotation):
+        tasks.append(instance.job.segment.task)
+    elif isinstance(instance, QualitySettings):
+        tasks.append(instance.task)
+    else:
+        assert False
+
+    for task in tasks:
+        qc.QualityReportUpdateManager().schedule_quality_autoupdate_job(task)
```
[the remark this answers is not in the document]
zhiltsov-max: Yes, that's true.

cvat/apps/quality_control/quality_reports.py
```diff
@@ -0,0 +1,2497 @@
...
+        # - job names are stable between potential writers
+        # - if multiple simultaneous writes can happen, the objects written must be the same
+        # - once a job is created, it can only be updated by the scheduler and the handling worker
+
+        if not self._should_update(task):
+            return
+
+        now = timezone.now()
+        delay = self._get_quality_check_job_delay()
+        next_job_time = now.utcnow() + delay
```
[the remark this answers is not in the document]
zhiltsov-max: I agree about using the same naming throughout the codebase, though I don't know if it can change anything in this particular case.

cvat/apps/quality_control/quality_reports.py
```diff
@@ -0,0 +1,2497 @@
...
+        dm.AnnotationType.mask,
+        dm.AnnotationType.polygon,
+        dm.AnnotationType.polyline,
+        dm.AnnotationType.skeleton,
+    ]
+
+    compare_attributes: bool = True
+    "Enables or disables attribute checks"
+
+    ignored_attributes: List[str] = []
```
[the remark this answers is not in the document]
zhiltsov-max: This is not supposed to be controlled by the user yet. It can be moved there once it is supported by the server.

cvat/apps/quality_control/serializers.py
```diff
@@ -0,0 +1,144 @@
...
+        }.items():
+            extra_kwargs.setdefault(field_name, {}).setdefault(
+                'help_text', textwrap.dedent(help_text.lstrip('\n'))
+            )
+
+    def validate(self, attrs):
+        for k, v in attrs.items():
+            if k.endswith('_threshold') or k in ['oks_sigma', 'line_thickness']:
+                if not 0 <= v <= 1:
+                    raise serializers.ValidationError(f"{k} must be in the range [0; 1]")
```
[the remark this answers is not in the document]
zhiltsov-max: I can add such type, but I don't see how this resolves the problem with forgetting, since the serializer fields are mostly added in the meta. The current solution checks for all `*_threshold` fields, so new fields will automatically be covered, if the follow the naming.

.vscode/launch.json
zhiltsov-max: Yes, I can add it.

cvat/apps/quality_control/views.py
```diff
@@ -0,0 +1,333 @@
...
+    list=extend_schema(
+        summary='Method returns a paginated list of quality settings',
+        responses={
+            '200': QualitySettingsSerializer(many=True),
+        }),
+    retrieve=extend_schema(
+        summary='Method returns details of quality settings',
+        responses={
+            '200': QualitySettingsSerializer,
+        }),
```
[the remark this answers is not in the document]
zhiltsov-max: Added

cvat/apps/quality_control/quality_reports.py
```diff
@@ -0,0 +1,2497 @@
...
+        if task.dimension != DimensionType.DIM_2D:
+            raise self.QualityReportsNotAvailable(
+                "Quality reports are only supported in 2d tasks"
+            )
+
+        gt_job = task.gt_job
+        if gt_job is None or not (
+            gt_job.stage == StageChoice.ACCEPTANCE and gt_job.status == StatusChoice.COMPLETED
+        ):
+            return self.QualityReportsNotAvailable(
```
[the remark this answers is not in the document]
zhiltsov-max: Fixed

cvat/apps/quality_control/models.py
```diff
@@ -0,0 +1,225 @@
...
+        return {k: v for k, v in default_settings.items() if k in existing_fields}
+
+    def to_dict(self):
+        return model_to_dict(self)
+
+    @property
+    def organization_id(self):
+        if task := self.task:
+            return getattr(task.organization, 'id', None)
+        return None
```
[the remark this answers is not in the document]
zhiltsov-max: Updated

cvat/apps/iam/rules/conflicts.rego
```diff
@@ -0,0 +1,111 @@
...
+        {"report__job__segment__task__owner_id": user.id},
+        {"report__job__segment__task__assignee_id": user.id}, "|",
+        {"report__job__segment__task__project__owner_id": user.id}, "|",
+        {"report__job__segment__task__project__assignee_id": user.id}, "|",
+        {"report__task__owner_id": user.id}, "|",
+        {"report__task__assignee_id": user.id}, "|",
+        {"report__task__project__owner_id": user.id}, "|",
+        {"report__task__project__assignee_id": user.id}, "|",
+    ]
+}
```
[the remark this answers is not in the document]
zhiltsov-max: Added

cvat/apps/quality_control/views.py
```diff
@@ -0,0 +1,339 @@
...
+    }
+    ordering_fields = list(filter_fields)
+    ordering = '-id'
+    serializer_class = AnnotationConflictSerializer
+
+    def get_queryset(self):
+        queryset = super().get_queryset()
+
+        if self.action == 'list':
+            self.check_permissions(self.request)
```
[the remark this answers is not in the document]
zhiltsov-max: I haven't found any permission check calls in the default implementation on the list endpoint.

cvat/apps/quality_control/default_settings.py
```diff
@@ -0,0 +1,8 @@
+# Copyright (C) 2023 Intel Corporation
+#
+# SPDX-License-Identifier: MIT
+
+import os
+
+QUALITY_CHECK_JOB_DELAY = int(os.getenv('CVAT_QUALITY_CHECK_JOB_DELAY', 15 * 60))
+"The delay before the next quality check job is queued, in seconds"
```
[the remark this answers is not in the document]
zhiltsov-max: If we define settings only in the global settings override file (such as `base.py` etc.):
- the code will fail at some point in runtime if any settings are not defined (the application code will depend on the global configuration)
- developers won't be able to determine the list of required settings unless by reading all the code (including the programmatic way, if needed)
- we won't be able to provide default values for settings (or there is going to be special treatment for undefined settings in the implementation, which may become duplicated) or modify the set of settings without breaking something

I wanted to avoid this. The default values still can be overridden in base.py and other settings files, and it is done for the sever api tests.

rqscheduler.py
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env python
+
+# This script adds access to the Django env and settings in the default rqscheduler
+# implementation. This is required for correct work with CVAT queue settings and
+# their access options such as login and password.
+
+# Required to initialize Django settings correctly
+from cvat.wsgi import application  # pylint: disable=unused-import
```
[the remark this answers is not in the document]
zhiltsov-max: Fixed

cvat/apps/quality_control/views.py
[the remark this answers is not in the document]
kirill-sizov: > Permissions in REST framework are always defined as a list of permission classes. Before running the main body of the > view each permission in the list is checked [link](https://www.django-rest-framework.org/api-guide/permissions/)

This statement matches what I see when I call list methods in debug mode (I get into PolicyEnforcer then I visit get_queryset method)

cvat/apps/quality_control/models.py
```diff
@@ -0,0 +1,225 @@
...
+        return report.comparison_summary
+
+    def get_task(self) -> Task:
+        if self.task is not None:
+            return self.task
+        else:
+            return self.job.segment.task
+
+    def get_json_report(self) -> str:
+        return self.data
```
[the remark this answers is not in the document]
zhiltsov-max: It's for better encapsulation. The field may contain arbitrary data, and the way the data is stored can also be changed, but the client code won't depend on this.

cvat/apps/quality_control/views.py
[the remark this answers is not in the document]
zhiltsov-max: Okay, I will remove such checks then

kirill-sizov (approved):

cvat/schema.yml
```diff
@@ -5127,6 +5672,43 @@ components:
...
+      - tag
+      - track
+      - rectangle
+      - polygon
+      - polyline
+      - points
+      - ellipse
+      - cuboid
+      - mask
+      - skeleton
```
SpecLad, line 5699: This seems to largely overlap with `ShapeType`. Perhaps it would be better to replace the shape types in this enum with a single `shape` member, and add an additional `shape_type` field (of type `ShapeType`) to `AnnotationId`?

cvat/schema.yml
```diff
@@ -5918,6 +6510,22 @@ components:
...
       - width
+    FrameSelectionMethodEnum:
+      enum:
+      - random_uniform
+      - manual
+      type: string
+      description: |-
+        * `random_uniform` - RANDOM_UNIFORM
+        * `manual` - MANUAL
+    ImportanceEnum:
```
SpecLad, line 6521: FWIW, I think this type of value is usually called "severity". Consider using this term.

cvat/schema.yml
```diff
@@ -6097,6 +6705,9 @@ components:
         mode:
           type: string
           readOnly: true
+        frame_count:
```
SpecLad, line 6708: There should probably be documentation on how this relates to `start_frame` and `stop_frame`.

cvat/schema.yml
```diff
@@ -6152,6 +6771,53 @@ components:
...
+          $ref: '#/components/schemas/JobStage'
+        state:
+          $ref: '#/components/schemas/OperationStatus'
+        type:
+          $ref: '#/components/schemas/JobType'
+        task_id:
+          type: integer
+        frame_selection_method:
+          $ref: '#/components/schemas/FrameSelectionMethodEnum'
+        count:
```
SpecLad, line 6798: This should probably be called `frame_count`, or it's a bit unclear.

cvat/schema.yml
```diff
@@ -6152,6 +6771,53 @@ components:
...
+          description: |
+            The number of frames included in the job.
+            Applicable only to the random frame selection
+        seed:
+          type: integer
+          minimum: 0
+          description: |
+            The seed value for the random number generator.
+            The same value will produce the same frame sets.
+            Applicable only to the random frame selection
```
SpecLad, line 6810: This should document what happens when the seed is unspecified.

cvat/schema.yml
```diff
@@ -7140,6 +7866,66 @@ components:
...
+          format: double
+          description: |
+            Thickness of polylines, relatively to the (image area) ^ 0.5.
+            The distance to the boundary around the GT line,
+            inside of which the checked line points should be
+        low_overlap_threshold:
+          type: number
+          format: double
+          description: Used for distinction between strong / weak (low_overlap) matches
+        oriented_lines:
```
SpecLad, line 7895:
```suggestion
        orient_lines:
```

cvat/schema.yml
```diff
@@ -7140,6 +7866,66 @@ components:
...
+          format: double
+          description: |
+            Minimal visible area percent of the spatial annotations (polygons, masks)
+            for reporting covered annotations,
+            useful with the 'check_covered_annotations' option
+        panoptic_comparison:
+          type: boolean
+          description: Use only the visible part of the masks and polygons in comparisons
+        compare_attributes:
+          type: boolean
```
SpecLad, line 7928: Description?

cvat/schema.yml
```diff
@@ -7140,6 +7866,66 @@ components:
...
+          description: Used for distinction between strong / weak (low_overlap) matches
+        oriented_lines:
+          type: boolean
+          description: Indicates that polylines have direction
+        line_orientation_threshold:
+          type: number
+          format: double
+          description: |
+            The minimal gain in the GT IoU between the given and reversed line directions
+            to consider the line inverted. Only useful with the 'oriented_lines' parameter
```
SpecLad, line 7903: The associated boolean parameter is described differently for each of `line_orientation_threshold`, `group_match_threshold` and `object_visibility_threshold`. Please use the same format everywhere.

cvat/schema.yml
```diff
@@ -7385,6 +8171,170 @@ components:
...
+          type: number
+          format: double
+      required:
+      - iou_threshold
+    QualityReportSummary:
+      type: object
+      properties:
+        frame_count:
+          type: integer
+        frame_share_percent:
```
SpecLad, line 8235: Percentages are IMO strictly a display format. Raw data should use fractions (from 0 to 1).

cvat/schema.yml
```diff
@@ -7385,6 +8171,170 @@ components:
...
+      - target
+    QualityReportCreateRequest:
+      type: object
+      properties:
+        task_id:
+          type: integer
+          writeOnly: true
+      required:
+      - task_id
+    QualityReportParameters:
```
SpecLad, line 8222: It seems odd to have both parameters _and_ settings. What is the relation between the two?

cvat/schema.yml
zhiltsov-max: What would you like to explain in the description?

cvat/schema.yml
```diff
@@ -4231,7 +4732,6 @@ paths:
           multipart/form-data:
             schema:
               $ref: '#/components/schemas/DataRequest'
-        required: true
```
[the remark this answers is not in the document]
zhiltsov-max: Because there is the `Upload-Start`, which doesn't expect any body.

cvat/schema.yml
zhiltsov-max: The parameters are different, I don't see how the same format can be used without hurting understanding.

cvat/schema.yml
zhiltsov-max: The parameter indicates that the line direction is important in the task. The suggested name would mean that lines are being changed somehow.

cvat/schema.yml
zhiltsov-max: Ah, if your suggestion was about this part, then ok.

cvat/schema.yml
zhiltsov-max: This is the actual frame count in the job, what description would you like to add? Probably, I'd defer it till all fields are described.

cvat/schema.yml
zhiltsov-max: Updated

cvat/schema.yml
zhiltsov-max: Added

cvat/schema.yml
zhiltsov-max: Added

cvat/schema.yml
zhiltsov-max: Updated

cvat/schema.yml
zhiltsov-max: Changed

cvat/schema.yml
```diff
@@ -3347,6 +3426,428 @@ paths:
...
+          content:
+            application/vnd.cvat+json:
+              schema:
+                $ref: '#/components/schemas/QualityReport'
+          description: ''
+        '202':
+          content:
+            application/vnd.cvat+json:
+              schema:
+                type: string
```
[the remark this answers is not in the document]
zhiltsov-max: Added

cvat/schema.yml
```diff
@@ -3347,6 +3426,428 @@ paths:
...
+          type: integer
+      - in: query
+        name: report_id
+        schema:
+          type: integer
+        description: A simple equality filter for report id
+      - name: search
+        required: false
+        in: query
+        description: 'A search term. Available search_fields: []'
```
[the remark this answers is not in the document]
zhiltsov-max: Fixed

cvat/schema.yml
zhiltsov-max: Done

cvat/apps/quality_control/quality_reports.py
```diff
@@ -2500,6 +2502,10 @@ def prepare_report_for_downloading(db_report: models.QualityReport, *, host: str
                     f"&serverID={ann_id['obj_id']}"
                 )
 
+    serialized_data["comparison_summary"]["frame_share_percent"] = (
+        serialized_data["comparison_summary"].pop("frame_share") * 100
+    )
```
SpecLad, line 2507: Why do this? I don't see why the report couldn't also contain the fraction.

zhiltsov-max: I talked to the annotators, they said they better like percents. Having both seems extraneous

cvat/schema.yml
SpecLad: Sounds good.

cvat/schema.yml
```diff
@@ -3347,6 +3426,428 @@ paths:
...
+  /api/quality/settings/{id}:
+    get:
+      operationId: quality_retrieve_settings
+      summary: Method returns details of quality settings
+      parameters:
+      - in: path
+        name: id
+        schema:
+          type: integer
+        description: A unique integer value identifying this quality settings.
```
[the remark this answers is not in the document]
SpecLad: I don't see it.

cvat/schema.yml
SpecLad: > not every parameter can be controlled

If something can't be controlled, how is it a parameter? It seems to me that being controllable is inherent to the definition of a parameter.

> and not everything controlled is supposed to be related to reports or quality estimation process.

I don't understand this either; we're talking about `QualitySettings`, shouldn't everything in there affect the quality estimation process?

cvat/schema.yml
zhiltsov-max: Why it must be equal to that number? There can also be skipped frames in the task configuration and only specific frames in the job segment. It's just the actual number of frames.

SpecLad: Alright.

cvat/schema.yml
zhiltsov-max: Quality control settings can include such parameters as report computation period or default output format, which do not affect any quality metrics. I suppose settings may be split into more specific sections if such parameters are introduced. I think, we can remove the report parameters from the response, as we did not find any good use for them in UI yet.

SpecLad: > I think, we can remove the report computation parameters from the response, as we did not find any good use for them in UI yet.

That's fine by me.

But FWIW, I think we shouldn't intermix settings that affect the contents of the report with other settings. Ideally, the former should be aggregated into a dedicated structure that can then be reused in the quality report.

cvat/schema.yml
zhiltsov-max: I've split it into 2 fields - `type` and `shape_type`.

cvat/apps/quality_control/quality_reports.py
zhiltsov-max: Added

cvat/schema.yml
zhiltsov-max: Removed parameters from reports summary

cvat/schema.yml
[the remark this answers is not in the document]
SpecLad: This is still incorrect; there's also now a duplicate `id` parameter listed.

SpecLad (approved):

nmanovic:
@zhiltsov-max , could you please release conflits in CHANGELOG.md?

nmanovic:
Could you please check that code coverage works for the new worker? Please work with @yasakova-anastasia if it is necessary.

PR #6242, opened 2023-06-05
Fix REST API tests (import with timeout)

yasakova-anastasia:
### Checklist

### License


azhavoro:
It seems that the reason for the this test fails is default value of the check interval for the scheduler. I'm ok with this as a quick fix, but it looks better if we use a lower interval value for testing purposes only in the future without increasing the test timeout.

https://github.com/rq/rq-scheduler/blob/master/rq_scheduler/scheduler.py#L31
 https://github.com/rq/django-rq/blob/master/django_rq/queues.py#L329

yasakova-anastasia:
@azhavoro, could you please review again?

cvat/settings/testing_rest.py
```diff
@@ -11,4 +11,6 @@
     "django.contrib.auth.hashers.MD5PasswordHasher",
 ]
 
-IMPORT_CACHE_CLEAN_DELAY = timedelta(seconds=30)
+IMPORT_CACHE_CLEAN_DELAY = timedelta(seconds=1)
```
Marishka17, line 14: Why did you change it to 1 second? It means that the cleaning job will start after 1 second after the file upload was finished, not imported. I guess that the tests are passed now because the interval is more than IMPORT_CACHE_CLEAN_DELAY. Logically it looks wrong to me.

tests/python/rest_api/test_tasks.py
```diff
@@ -1825,7 +1825,7 @@ def test_check_import_cache_after_previous_interrupted_upload(self, tasks_with_s
             url, filename, meta=params, logger=self.client.logger.debug
         )
         number_of_files = 1
-        sleep(30)  # wait when the cleaning job from rq worker will be started
+        sleep(5)  # wait when the cleaning job from rq worker will be started
```
Marishka17, line 1828: Why do we need to sleep only 5 seconds here?
How about changing the number of iterations in `for _ in range(12):` too?

azhavoro: I suppose this value should be the same as `RQ_SCHEDULER_CHECK_INTERVAL + IMPORT_CACHE_CLEAN_DELAY`, it also seems that we can reduce `RQ_SCHEDULER_CHECK_INTERVAL` to 5-10s

cvat/settings/testing_rest.py
azhavoro: I think we can set 15s here for two reasons:
1. we don't currently have huge test data to import
2. 15 is the standard timeout for test cases

tests/python/rest_api/test_tasks.py
azhavoro: @Marishka17 please correct me if I'm wrong

cvat/settings/testing_rest.py
yasakova-anastasia: Set 15s for IMPORT_CACHE_CLEAN_DELAY and 10s for scheduler interval.

Marishka17 (approved):

Marishka17:
@yasakova-anastasia, Could you please resolve conflicts?

azhavoro:
Unfortunately, the test still fails: https://github.com/opencv/cvat/actions/runs/5240951371/jobs/9472316010
I'm starting to think that the problem might not be the timeout setting...

cvat/apps/dataset_manager/views.py
```diff
@@ -85,7 +85,10 @@ def export(dst_format, project_id=None, task_id=None, job_id=None, server_url=No
                 os.replace(temp_file, output_path)
 
             archive_ctime = osp.getctime(output_path)
-            scheduler = django_rq.get_scheduler(settings.CVAT_QUEUES.EXPORT_DATA.value)
+            scheduler = django_rq.get_scheduler(
+                settings.CVAT_QUEUES.EXPORT_DATA.value,
+                interval=settings.DJANGO_RQ_SCHEDULER_INTERVAL
```
zhiltsov-max, line 90: Are you sure the interval should be set here and not in the scheduler startup parameters?

PR #6276, opened 2023-06-09
Auto-annotation support for jobs

zhiltsov-max:
### Motivation and context

Depends on #6204 

- Added an option to run auto-annotation on jobs
- Added API schema definitions for offline lambda call endpoints

### Checklist

### License


azhavoro (approved):

cvat/schema.yml
```diff
@@ -6540,6 +6560,98 @@ components:
...
+          type: string
+          format: date-time
+        started:
+          type: string
+          format: date-time
+        ended:
+          type: string
+          format: date-time
+        exc_info:
+          type: string
```
SpecLad, line 6586: I think these should be marked nullable (I see that they are set to null in the response to the initial create request).

cvat/schema.yml
```diff
@@ -6540,6 +6560,98 @@ components:
...
+        id:
+          type: string
+          description: The name of the function
+        task:
+          type: integer
+          description: The id of the task
+        job:
+          type: integer
+          description: The id of the job
+        threshold:
```
SpecLad, line 6607: This should be marked nullable, as well.

cvat/schema.yml
zhiltsov-max: I couldn't find where it happens for `exc_info`, but there are other fields that can potentially be null, so I marked them.

cvat/schema.yml
zhiltsov-max: Done

SpecLad (approved):

PR #6362, opened 2023-06-22
Implement more comprehensive SSRF mitigation

SpecLad:
### Motivation and context

The current mitigation approach (resolving the IP address and checking if it's in the private range) is insufficient for a few reasons:

* It is susceptible to DNS rebinding (an attacker-controlled DNS name resolving to a public IP address when queried during the check, and to a private IP address afterwards).

* It is susceptible to redirect-based attacks (a server with a public address redirecting to a server with a private address).

* It is only applied when downloading remote files of tasks (and is not easily reusable).

Replace it with an approach based on smokescreen, a proxy that blocks connections to private IP addresses. In addition, use this proxy for webhooks, since they also make requests to untrusted URLs.

The benefits of smokescreen are as follows:

* It's not susceptible to the problems listed above.

* It's configurable, so system administrators can allow certain private IP ranges if necessary. This configurability is exposed via the `SMOKESCREEN_OPTS` environment variable.

* It doesn't require much code to use.

The drawbacks of smokescreen are:

* It's not as clear when the request fails due to smokescreen (compared to manual IP validation). To compensate, make the error message in `_download_data` more verbose.

* The smokescreen project seems to be in early development (judging by the 0.0.x version numbers). Still, Stripe itself uses it, so it should be good enough. It's also not very convenient to set up (on account of not providing binaries), so disable it in development environments.

Keep the scheme check from `_validate_url`. I don't think this check prevents any attacks (as requests only supports http/https to begin with), but it provides a friendly error message in case the user tries to use an unsupported scheme.

### How has this been tested?

Manual testing.

### Checklist

### License


azhavoro:
LGTM

azhavoro (approved):

PR #6367, opened 2023-06-23
Project quality

### LINKED ISSUE #6591: Typo in interface
- [ ] Read/searched [the docs](https://github.com/cvat-ai/cvat/tree/master#documentation)
- [ ] Searched [past issues](/issues)

"Createda"

Screenshot from 2023-07-31 15-29-36

### Your Environment

- Git hash commit (`git log -1`):
- Docker version `docker version` (e.g. Docker 17.0.05):
- Are you using Docker Swarm or Kubernetes?
- Operating System and version (e.g. Linux, Windows, MacOS):
- Code example or link to GitHub repo or gist to reproduce problem:
- Other diagnostic information / logs:
    <details>
    <summary>Logs from `cvat` container</summary>
    </details>

zhiltsov-max:
### Motivation and context

- Added project quality estimation
- Added project quality UI
- Added tests
- Refactored task quality UI
- Fixed task quality settings access in org for workers
- Fixed gt job status checks before and after task report computation
- Task quality settings now can be initialized manually
- Default quality settings' values now available in API schema
- Fixed GT job creation for the whole task size
- Added missing task/project info in job/task quality report responses
- Fixes #6591
- Fixed that only the first job page had quality results displayed in UI
- Fixed the error with invalid data in GT job chunks

TBD:
- [ ] Add instance-agnostic segmentation comparison (per class?) / auto-grouping for masks in segmentation comparisons
- [ ] Add class-specific options
- [ ] Fix invalid accuracy formula used (needs 2*TP/(DS+GT))
- [ ] Add "included annotation types" switch

### Checklist

### License


cvat/apps/quality_control/quality_reports.py
```diff
@@ -2480,8 +2495,347 @@ def _save_reports(self, *, task_report: Dict, job_reports: List[Dict]) -> models
...
     def _get_task_quality_params(self, task: Task) -> Optional[ComparisonParameters]:
-        quality_params, _ = models.QualitySettings.objects.get_or_create(task=task)
-        return ComparisonParameters.from_dict(quality_params.to_dict())
+        if task.project:
+            try:
+                quality_params = task.project.quality_settings
+            except models.QualitySettings.DoesNotExist:
+                quality_params = None
+        else:
+            quality_params, _ = models.QualitySettings.objects.get_or_create(task=task)
```
SpecLad, line 2504: The logic here seems inconsistent.

* If the task is within a project, then we use the settings from the project. If there are no such settings, we return `None` (which prevents report generation).
* If the task is not within a project, then we use the settings from the task. If there are no such settings, we create default ones.

Why not use default settings in the first case too?

zhiltsov-max: > Why not use default settings in the first case too?

Technically, we could, but it's about the server load and existing projects. Basically, I'd prefer to remove creation of default settings for tasks too, but in the case of GT jobs, there is an explicit action of GT job creation, so there is the opportunity to create settings (during the job creation or the first computation, as it is now). In the case of projects there was no such obvious explicit action, so I decided to make creation of quality settings for project to be so. Other variant - add a migration for existing projects to add settings automatically. I don't like it for several reasons. Just adding default settings as you suggest won't work well too, because it can lead to unexpected results and unnecessary server load for unused functionality in existing projects (as there are many things that can trigger report computation).

bsekachev:
Wrong number for **Total frames.** There are only 300. Probably overlapping frames are counted also here. 
image

Clicking these links causes UI to fail:
image
image

For a while this table is empty:
image
After some time it getting filled in. I would propose to add loading indication for better UX. 

Tried to open this link:
image
Leads to following (correct project id is ``1``. Probably task id is considered as project id): 
image

On project quality page initialization we see loading bar (before it gets response from the server) and then see ready metrics.
On task quality page initialization we first see ``N/A`` (before it gets response from the server) and then see ready metrics.
I would propose to unify this behavior, I personally like the first one (as it works for projects)
image
image

cvat-core/src/api-implementation.ts
```diff
@@ -390,19 +405,56 @@ export default function implementAPI(cvat) {
...
+        conflicts.count = conflictsData.count;
+        return conflicts;
+    };
 
-        return reportsData.map((conflict) => new QualityConflict({ ...conflict }));
+    cvat.analytics.quality.settings.get.implementation = async (filter: any) => {
+        interface FilterParams {
+            id?: number, taskId?: number, projectId?: number
+        }
+        const { id, taskId, projectId }: FilterParams = filter;
```
bsekachev, line 418:
```suggestion
    cvat.analytics.quality.settings.get.implementation = async (
    filter: { id?: number, taskId?: number, projectId?: number }) => {
        const { id, taskId, projectId } = filter;
```

cvat-core/src/api-implementation.ts
```diff
@@ -390,19 +405,56 @@ export default function implementAPI(cvat) {
...
+        if (settings) {
+            return new QualitySettings({ ...settings });
+        }
+        return null;
     };
 
-    cvat.analytics.quality.settings.get.implementation = async (taskID: number) => {
-        const settings = await serverProxy.analytics.quality.settings.get(taskID);
-        return new QualitySettings({ ...settings });
+    cvat.analytics.quality.settings.update.implementation = async (settingsId: number, values: any) => {
```
bsekachev, line 427: May me avoid ``any`` in arguments and return values?

cvat-core/src/api-implementation.ts
```diff
@@ -390,19 +405,56 @@ export default function implementAPI(cvat) {
...
+        const settings = await serverProxy.analytics.quality.settings.update(settingsId, values);
+
+        if (settings) {
+            return new QualitySettings({ ...settings });
+        }
```
bsekachev, line 432: Always true

cvat-core/src/api-implementation.ts
```diff
@@ -390,19 +405,56 @@ export default function implementAPI(cvat) {
...
 
-    cvat.frames.getMeta.implementation = async (type, id) => {
-        const result = await serverProxy.frames.getMeta(type, id);
-        return new FramesMetaData({ ...result });
+    cvat.analytics.quality.settings.create.implementation = async (values: any) => {
+        const settings = await serverProxy.analytics.quality.settings.create(values);
+
+        if (settings) {
+            return new QualitySettings({ ...settings });
+        }
```
bsekachev, line 441: Always true

cvat-core/src/api.ts
```diff
@@ -295,13 +295,30 @@ function build() {
...
                         return result;
                     },
                 },
             },
         },
+        scheme: {
+            async get() {
+                return PluginRegistry.apiWrapper(cvat.scheme.get);
+            },
+        },
```
bsekachev, line 321: It should be in server namespace.

```
cvat.server.scheme.get()
```

Or just ``cvat.server.scheme()`. I  do not really expect that client can do something else with the server  scheme

cvat-core/src/api-implementation.ts
```diff
@@ -390,19 +405,56 @@ export default function implementAPI(cvat) {
...
+        return null;
+    };
+
+    cvat.analytics.quality.settings.defaults.implementation = async () => {
+        const scheme = await cvat.scheme.get();
```
bsekachev, line 446: From the inside of cvat-core, better to call implementation method directly.
I do not think such call should trigger plugins if there are any

So, just ``await cvat.scheme.get.implementation()`` or ``serverProxy.scheme.get()``.

cvat-core/src/server-response-types.ts
```diff
@@ -71,7 +72,9 @@ export interface SerializedProject {
     status: ProjectStatus;
 }
 
-export type TasksFilter = ProjectsFilter & { ordering?: string; }; // TODO: Need to clarify how "ordering" is used
+export type TasksFilter = ProjectsFilter & {
```
bsekachev, line 75: Why did you remove this comment? It is still not clear..

cvat-core/src/server-response-types.ts
```diff
@@ -31,6 +31,7 @@ export interface FunctionsResponseBody {
 
 export interface ProjectsFilter {
     page?: number;
+    pageSize?: number | 'all';
```
bsekachev, line 34: Probably need to fix type JobsFilter below:
```
export type JobsFilter = Omit<ProjectsFilter & {
    task_id?: number;
}, 'pageSize'>;
```

cvat-core/src/api-implementation.ts
bsekachev, line 427: According to current approach, update should be a method of QualitySettings class

cvat-core/src/api-implementation.ts
```diff
@@ -390,19 +405,56 @@ export default function implementAPI(cvat) {
...
+        if (settings) {
+            return new QualitySettings({ ...settings });
+        }
+        return null;
     };
 
-    cvat.frames.getMeta.implementation = async (type, id) => {
-        const result = await serverProxy.frames.getMeta(type, id);
-        return new FramesMetaData({ ...result });
+    cvat.analytics.quality.settings.create.implementation = async (values: any) => {
```
bsekachev, line 436:
```suggestion
    cvat.analytics.quality.settings.create.implementation = async (values: SerializedQualitySettingsData): Promise<QualitySettings> => {
```

cvat-core/src/quality-settings.ts
```diff
@@ -60,7 +63,11 @@ export default class QualitySettings {
     }
 
     get task(): number {
-        return this.#task;
+        return this.#task_id;
+    }
+
+    get project(): number {
```
bsekachev, line 69: Looks like number | null is correct for both task and project getters
```suggestion
    get project(): number | null {
```

cvat-core/src/quality-settings.ts
```diff
@@ -182,15 +191,31 @@ export default class QualitySettings {
...
+            },
+        }),
+        create: Object.freeze({
+            writable: false,
+            enumerable: false,
+            value: async function implementation() {
+                const result = await serverProxy.analytics.quality.settings.create(this.toJSON());
+                return result;
+            },
+        }),
```
bsekachev, line 219: Basically everywhere we have method ``save`` which perform both create and update depending on the object exist or not

cvat-core/src/quality-settings.ts
```diff
@@ -182,15 +191,31 @@ export default class QualitySettings {
...
-    },
-});
+Object.defineProperties(
+    QualitySettings.prototype,
+    Object.freeze({
+        save: Object.freeze({
+            writable: false,
+            enumerable: false,
+            value: async function implementation() {
+                const result = await serverProxy.analytics.quality.settings.update(this.id, this.toJSON());
```
bsekachev, line 208: How is it different from ``cvat.analytics.quality.settings.update.implementation`` declared in ``api-implementation``?

cvat-core/src/quality-settings.ts
```diff
@@ -182,15 +191,31 @@ export default class QualitySettings {
...
+                return new QualitySettings(result);
+            },
+        }),
+        create: Object.freeze({
+            writable: false,
+            enumerable: false,
+            value: async function implementation() {
+                const result = await serverProxy.analytics.quality.settings.create(this.toJSON());
```
bsekachev, line 216: How is it different from ``cvat.analytics.quality.settings.create.implementation`` declared in ``api-implementation``?

cvat-core/src/api-implementation.ts
```diff
@@ -34,6 +34,10 @@ import QualitySettings from './quality-settings';
 import { FramesMetaData } from './frames';
 import AnalyticsReport from './analytics-report';
 
+function isPageSize(value: any) {
```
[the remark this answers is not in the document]
zhiltsov-max: Not sure it's right, other checking functions used accept `any`, and it makes sense to check `any`, because we want to validate an arbitrary passed value.

cvat-core/src/api-implementation.ts
```diff
@@ -34,6 +34,10 @@ import QualitySettings from './quality-settings';
...
+    return isInteger(value) || value === 'all';
```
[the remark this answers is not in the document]
zhiltsov-max: It does. It seems a good solution in the current circumstances - otherwise we would probably need to use a loop to fetch all the data, which has bad performance.

cvat-core/src/api-implementation.ts
```diff
@@ -390,19 +405,56 @@ export default function implementAPI(cvat) {
             };
         }
 
-        const reportsData = await serverProxy.analytics.quality.conflicts(updatedParams);
+        const conflictsData = await serverProxy.analytics.quality.conflicts(updatedParams);
```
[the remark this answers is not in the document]
zhiltsov-max: 1 - potentially, 2 - yes. It's a regular list endpoint.

cvat-core/src/server-response-types.ts
zhiltsov-max: Do you mean, it's not clear in the UI? From the server side it's quite simple and straightforward.

cvat-core/src/api-implementation.ts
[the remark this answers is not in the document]
bsekachev: This is what type we expect used to static linting.
Code changes what actually what passed in runtime.

If we pass any, we already do not see type errors by static analizer

cvat-core/src/api-implementation.ts
zhiltsov-max: Updated

cvat-core/src/api-implementation.ts
zhiltsov-max: Updated

cvat-core/src/api.ts
zhiltsov-max: Updated

cvat-core/src/api-implementation.ts
zhiltsov-max: Updated

cvat-core/src/api-implementation.ts
zhiltsov-max: Updated

cvat-core/src/api-implementation.ts
zhiltsov-max: Sure. Updated.

cvat-core/src/api-implementation.ts
```diff
@@ -362,24 +368,33 @@ export default function implementAPI(cvat) {
...
+            updatedParams.task_id = filter.taskId;
         }
         if ('jobId' in filter) {
-            updatedParams = {
-                job_id: filter.jobId,
-                sort: '-created_date',
-                target: filter.target,
-            };
+            updatedParams.job_id = filter.jobId;
+        }
```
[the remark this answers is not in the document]
zhiltsov-max: Updated.

cvat-core/src/api-implementation.ts
```diff
@@ -362,24 +368,33 @@ export default function implementAPI(cvat) {
     };
 
     cvat.analytics.quality.reports.implementation = async (filter) => {
-        let updatedParams: Record<string, string> = {};
+        const updatedParams: Record<string, string> = {};
```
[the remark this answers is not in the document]
zhiltsov-max: Updated.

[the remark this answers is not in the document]
zhiltsov-max: Updated.

cvat-core/src/api-implementation.ts
bsekachev: Removed

cvat-core/src/quality-settings.ts
bsekachev: Updated

cvat-core/src/quality-settings.ts
bsekachev: Removed extra API

cvat-core/src/quality-settings.ts
bsekachev: Removed extra API

cvat/apps/quality_control/models.py
```diff
@@ -222,4 +310,26 @@ def to_dict(self):
...
+            return getattr(self.project.organization, "id", None)
+        return None
+
+    @transaction.atomic
+    def save(self, *args, **kwargs) -> None:
+        try:
+            return super().save(*args, **kwargs)
+        except IntegrityError as ex:
+            message = str(ex)
+            if "duplicate key value violates unique constraint" in message:
```
SpecLad, line 325: This doesn't seem very robust. I think It would be better to query the database to see if the given quality parameters really already exist.

cvat/apps/quality_control/quality_reports.py
```diff
@@ -251,17 +253,17 @@ class ComparisonReportAnnotationsSummary(_Serializable):
     total_count: int
     ds_count: int
     gt_count: int
-    confusion_matrix: ConfusionMatrix
+    confusion_matrix: Optional[ConfusionMatrix]
 
-    @property
+    @cached_property
```
SpecLad, line 258: I think to cache safely, you should make the class frozen.

cvat/apps/quality_control/quality_reports.py
```diff
@@ -322,6 +339,16 @@ def accumulate(self, other: ComparisonReportAnnotationShapeSummary):
     def _fields_dict(self, *, include_properties: Optional[List[str]] = None) -> dict:
         return super()._fields_dict(include_properties=include_properties or ["accuracy"])
 
+    _CACHED_FIELDS = ["accuracy", "precision", "recall"]
```
SpecLad, line 342:
```suggestion
    _CACHED_FIELDS = ["accuracy"]
```

cvat/apps/quality_control/quality_reports.py
```diff
@@ -417,20 +462,40 @@ def _fields_dict(self, *, include_properties: Optional[List[str]] = None) -> dic
             ]
         )
 
+    _CACHED_FIELDS = ["frame_count", "mean_conflict_count", "frame_share"]
+
+    def __init__(self, *args, **kwargs):
+        # these fields are optional, but can be computed on access
+        for field_name in self._CACHED_FIELDS:
+            if field_name in kwargs:
+                setattr(self, field_name, kwargs.pop(field_name))
```
SpecLad, line 471: This block could probably be factored out.

cvat/apps/quality_control/serializers.py
```diff
@@ -134,10 +157,37 @@ class Meta:
                 "help_text", textwrap.dedent(help_text.lstrip("\n"))
             )
 
+    def get_extra_kwargs(self):
+        from .models import QualitySettings
```
SpecLad, line 161: Aren't `models` already imported at the top?

cvat/apps/quality_control/default_settings.py
```diff
@@ -6,3 +6,6 @@
 
 QUALITY_CHECK_JOB_DELAY = int(os.getenv("CVAT_QUALITY_CHECK_JOB_DELAY", 15 * 60))
 "The delay before the next quality check job is queued, in seconds"
+
+PROJECT_QUALITY_CHECK_JOB_DELAY = int(os.getenv("CVAT_PROJECT_QUALITY_CHECK_JOB_DELAY", 5 * 60))
```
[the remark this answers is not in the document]
zhiltsov-max: Project reporting is executed after the task reports. It's significantly cheaper to compute, unlike task reports. In case of projects, this delay is mainly to accumulate several task reports and not compute a new report on each task update. In the case of tasks, it's mainly to reduce server load because of hard computations.

cvat/apps/quality_control/models.py
```diff
@@ -76,26 +77,38 @@ class QualityReport(models.Model):
...
     )
-
-    parent = models.ForeignKey(
-        "self", on_delete=models.CASCADE, related_name="children", null=True, blank=True
+    project = models.ForeignKey(
+        Project, on_delete=models.CASCADE, related_name="quality_reports", null=True, blank=True
     )
+
+    # job reports should all have a single parent report
+    # task reports may have none, or be shared between several project reports
```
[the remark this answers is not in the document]
zhiltsov-max: It could also be implemented for jobs, but I didn't do this to simplify logic. In the case of projects, a task can be unchanged from the previous project report computation. In this case, the previous task report is used in the new project report.

cvat/apps/quality_control/models.py
zhiltsov-max: The error should come from the DB, what do you mean?

cvat/schema.yml
```diff
@@ -8880,17 +8953,11 @@ components:
       - error_count
       - frame_count
       - frame_share
+      - frames_with_errors
       - gt_count
+      - total_frames
       - valid_count
       - warning_count
-    QualityReportTarget:
```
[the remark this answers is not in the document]
zhiltsov-max: It seems drf-spectacular will map all the enums with the same members into 1 type. I think since we just launched the feature, it's safe enough to change API type names.

bsekachev (approved):

cvat/apps/quality_control/default_settings.py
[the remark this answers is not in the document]
SpecLad: Okay. In that case, I suggest that you update the documentation string to mention that the delay starts after task reports.

Also, you should update the docstring for the preceding setting to note that it applies to tasks (I assume the setting itself cannot be renamed due to backwards compatibility).

cvat/apps/quality_control/models.py
SpecLad: I mean that you have no guarantee that the error will contain this wording in the future.

cvat/schema.yml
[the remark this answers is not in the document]
SpecLad: This seems like dangerous behavior. Suppose we add another possible target for analytics, but not quality (or vice versa) - will the enum split back into two again? Is there no way to ask drf-spectacular to not merge enums?

rqscheduler.py
```diff
@@ -4,10 +4,33 @@
...
+    parser.add_argument('--clear', action='store_true', help="Clean up all jobs from all queues")
+    args, _ = parser.parse_known_args()
+
+    if args.clear:
+        connection = Redis(args.host, args.port, args.db, args.password)
+        for queue in Queue.all(connection=connection):
+            for job in queue.jobs:
+                job.cancel()
+
+        sys.exit(0)
```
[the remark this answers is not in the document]
zhiltsov-max: I don't think creating an extra script just for this is a good idea. There is the existing `burst` parameter, which is quite similar in behavior, so I decided to put this queue management code here.

cvat/schema.yml
```diff
@@ -3828,7 +3843,7 @@ paths:
         required: false
         in: query
         description: 'A filter term. Available filter_fields: [''id'', ''job_id'',
-          ''created_date'', ''gt_last_updated'', ''target_last_updated'', ''parent_id'']'
```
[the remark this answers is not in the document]
SpecLad: Maybe it's worth noting in the docs? It's a bit unusual.

rqscheduler.py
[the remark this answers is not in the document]
zhiltsov-max: > They are not similar in behavior. rqscheduler works with scheduled tasks, whereas your code works with queued tasks.

Generally, this is what schedulers do - they manage queues and task order, the specific queue is a detail.

> But besides that, it just doesn't fit the description of the script.

Do you mean the comment on top of the file? Ok, I'll update the description.

> (Incidentally, I'm not sure why we even have this script at all. Seems like it could just be replaced with manage.py rqscheduler.)

This is in the description. Without the application import line, the original implementation cannot work with queue parameters from the django application config. The `manage.py rqscheduler` has different parameters from the original `rqscheduler.py`, and these parameters don't allow us to have 1 scheduler for all queues.

cvat/schema.yml
```diff
@@ -5947,7 +5998,7 @@ components:
           type: integer
         project_id:
           type: integer
-    AnalyticsReportTargetEnum:
+    AnalyticsReportTarget:
```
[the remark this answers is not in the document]
zhiltsov-max: Updated

cvat/apps/iam/permissions.py
```diff
@@ -1929,21 +1959,68 @@ def create(cls, request, view, obj, iam_context):
...
+                            try:
+                                project = Project.objects.get(id=project_id)
+                            except Project.DoesNotExist:
+                                raise ValidationError(f"Project {project_id} does not exist")
+
+                        if (task_id := request.data.get('task_id')):
+                            try:
+                                task = Task.objects.get(id=task_id)
+                            except Task.DoesNotExist:
+                                raise ValidationError(f"Task {task_id} does not exist")
```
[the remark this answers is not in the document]
zhiltsov-max: It's checked in the serializer, this logic doesn't belong to permissions.

cvat/apps/quality_control/models.py
zhiltsov-max: If it changes in future, it should be catched by a corresponding test.

rqscheduler.py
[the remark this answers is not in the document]
SpecLad: > Generally, this is what schedulers do - they manage queues and task order, the specific queue is a detail.

I'm not talking generally, I'm talking specifically about RQ Scheduler. The purpose of `rqscheduler.main()` is to find scheduled tasks and enqueue them. Whereas your code finds _enqueued_ tasks (which might not have even come from the scheduler) and deletes them. There's no relation.

Your code doesn't even use the `rq_scheduler` module. That seems to me like a sign that it shouldn't be in `rqscheduler.py`.

> The manage.py rqscheduler has different parameters from the original rqscheduler.py, and these parameters don't allow us to have 1 scheduler for all queues.

I see... that's pretty annoying.

Frankly, it seems like a goof by django-rq, because I don't think `Scheduler.run` even uses the queue that was passed in in the constructor. It processes all scheduled jobs regardless of queue. I think I'll file an issue for django-rq.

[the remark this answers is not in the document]
SpecLad: > Frankly, it seems like a goof by django-rq, because I don't think Scheduler.run even uses the queue that was passed in in the constructor. It processes all scheduled jobs regardless of queue. I think I'll file an issue for django-rq.

Oh, I see why django-rq does this. It's because it allows you to configure different servers per queue (rq/django-rq#28), so you need to know the name of the queue to connect to the appropriate server. Less of a goof and more of a design error.

Except... hold on. If you need to know the name of the queue to connect to the appropriate server, then how could this script have worked in the first place? How would merely importing `cvat.asgi` help it connect?

cvat/schema.yml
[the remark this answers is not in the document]
zhiltsov-max: Restored old names

cvat/apps/quality_control/signals.py
```diff
@@ -21,38 +24,100 @@
...
 
+@receiver(
+    post_save,
+    sender=QualityReport,
+    dispatch_uid=__name__ + ".save_task_quality_report-update_project_quality_metrics",
+)
+def _on_task_report_save__update_project_quality_metrics(instance, created, **kwargs):
+    project = None
+
+    if isinstance(instance, QualityReport):
```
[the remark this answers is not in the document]
zhiltsov-max: Updated

cvat/apps/quality_control/quality_reports.py
zhiltsov-max: Updated

cvat/apps/quality_control/quality_reports.py
zhiltsov-max: Updated

cvat/apps/quality_control/default_settings.py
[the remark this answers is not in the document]
zhiltsov-max: Added

cvat/apps/iam/permissions.py
```diff
@@ -1929,21 +1959,68 @@ def create(cls, request, view, obj, iam_context):
...
+                        elif scope in [Scopes.UPDATE, Scopes.CREATE]:
+                            project_scope = ProjectPermission.Scopes.UPDATE_DESC
+                        else:
+                            assert False
+
+                        parent_perm = ProjectPermission.create_base_perm(
+                            request, view,
+                            iam_context=get_iam_context(request, project),
+                            scope=project_scope, obj=project
+                        )
```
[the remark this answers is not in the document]
zhiltsov-max: I don't see a good way to do this now.

cvat/apps/iam/permissions.py
```diff
@@ -1834,7 +1853,7 @@ def create(cls, request, view, obj, iam_context):
 
     def __init__(self, **kwargs):
         if 'job_owner_id' in kwargs:
-            self.job_owner_id = int(kwargs.pop('job_owner_id'))
+            self.rq_job_owner_id = int(kwargs.pop('job_owner_id'))
```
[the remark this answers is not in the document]
zhiltsov-max: Updated

cvat/apps/engine/serializers.py
```diff
@@ -628,10 +628,10 @@ def create(self, validated_data):
             frame_selection_method = validated_data.pop("frame_selection_method", None)
             if frame_selection_method == models.JobFrameSelectionMethod.RANDOM_UNIFORM:
                 frame_count = validated_data.pop("frame_count")
-                if size <= frame_count:
+                if size < frame_count:
                     raise serializers.ValidationError(
-                        f"The number of frames requested ({frame_count}) must be lesser than "
-                        f"the number of the task frames ({size})"
+                        f"The number of frames requested ({frame_count}) must be lesser than or "
```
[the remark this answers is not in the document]
zhiltsov-max: Updated

cvat/schema.yml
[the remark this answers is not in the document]
zhiltsov-max: There are several such places, filters can support different sets of fields depending on the situation.

cvat/apps/quality_control/default_settings.py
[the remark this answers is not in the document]
SpecLad: Not quite - you added what the delay is before, but you didn't add what it's after.

cvat/apps/quality_control/models.py
SpecLad: Is there a corresponding test? I couldn't find one.

cvat/apps/quality_control/quality_reports.py
```diff
@@ -79,8 +81,72 @@ def from_dict(cls, d: dict):
         raise NotImplementedError("Must be implemented in the subclass")
 
 
-@define(kw_only=True)
-class AnnotationId(_Serializable):
+class _ReportNode(_Serializable):
+    _CHECK_SETATTR_FIELD_NAME = "_check_setattr"
+
+    _CACHED_FIELDS: Optional[List[str]] = None
```
SpecLad, line 87: I don't quite understand; it seems that the code depends on this attribute being initialized in each derived class, but you removed all `_CACHED_FIELDS` assignments?

cvat/schema.yml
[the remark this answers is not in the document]
SpecLad: Looks like there's still a name change: `AnalyticsReportTargetEnum` -> `AnalyticsTargetEnum`.

cvat/apps/quality_control/quality_reports.py
zhiltsov-max: The list of cached fields can be defined explicitly in each sublclass, but it's being derived automatically from the class declaration now in all cases in the `_find_cached_fields()` classmethod.

cvat/apps/quality_control/default_settings.py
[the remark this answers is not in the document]
zhiltsov-max: It's computed after task quality reports, but not only after them, because sometimes there are no tasks in the projects. I don't see how this information would be useful here.

cvat/apps/quality_control/models.py
zhiltsov-max: Added a dedicated test for this, but we don't have such tests in general.

tests/extras/clear_rq.py
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env python
+
+import argparse
+import os
+
+# Required to initialize Django settings correctly
+from cvat.asgi import application  # pylint: disable=unused-import
+
```
SpecLad, line 8: This shouldn't be needed; this script doesn't use Django.

cvat/apps/quality_control/quality_reports.py
SpecLad: I see. But in that case, why even have the `_CACHED_FIELDS` attribute?

PR #6371, opened 2023-06-26
Project/Task/Job Analytics

klakhov:
### Checklist

### License


cvat-ui/src/components/analytics-page/analytics-performance.tsx
```diff
@@ -0,0 +1,222 @@
...
+    'rgba(53, 162, 235, 0.5)',
+    'rgba(170, 83, 85, 0.5)',
+    'rgba(44, 70, 94, 0.5)',
+    'rgba(28, 66, 98, 0.5)',
+];
+
+function AnalyticsOverview(props: Props): JSX.Element | null {
+    const { report, onTimePeriodChange } = props;
+
+    // TODO make it more expressive
```
bsekachev, line 42: What does it mean?)

cvat-ui/src/components/analytics-page/analytics-performance.tsx
```diff
@@ -0,0 +1,222 @@
...
+                            labels={dateLabels}
+                            title={entry.title}
+                            key={name}
+                        />
+                    ),
+                    key: name,
+                });
+            }
+            default: {
+                throw Error(`View type ${entry.defaultView} is not supported`);
```
bsekachev, line 170:
```suggestion
                throw new Error(`View type ${entry.defaultView} is not supported`);
```

bsekachev: Probably UI should not fail if there is a not supported view. Warning would be enough, I believe

cvat-ui/src/components/analytics-page/analytics-performance.tsx
```diff
@@ -0,0 +1,222 @@
...
+                                }
+                                return {
+                                    value: 0,
+                                    date: left.date,
+                                };
+                            });
+                            delete dataseries[leftName];
+                            delete dataseries[rightName];
+                        }
+                    }
```
bsekachev, line 122: I would suggest keeping any computation logic in cvat-core module

cvat-ui/src/components/analytics-page/analytics-performance.tsx
```diff
@@ -0,0 +1,222 @@
...
+                            case '*': {
+                                operator = (left: number, right: number) => left * right;
+                                break;
+                            }
+                            case '/': {
+                                operator = (left: number, right: number) => (right !== 0 ? left / right : 0);
+                                break;
+                            }
+                            default: {
+                                throw Error(`Operator type ${transform.binary.operator} is not supported`);
```
bsekachev, line 99:
```suggestion
                                throw new Error(`Operator type ${transform.binary.operator} is not supported`);
```

The same, should we fail the whole UI in case of not supported operator?

cvat/schema.yml
```diff
@@ -8888,6 +9040,31 @@ components:
...
         * `random` - RANDOM
+    StatisticsItem:
+      type: object
+      properties:
+        title:
+          type: string
+        description:
+          type: string
+        granularity:
+          $ref: '#/components/schemas/GranularityEnum'
```
zhiltsov-max, line 9051: Probably, these enum fields should allow a null value instead of having a dedicated enum member for this. I'm not sure if none is a valid member for such enum.

cvat/schema.yml
```diff
@@ -8888,6 +9040,31 @@ components:
         * `natural` - NATURAL
         * `predefined` - PREDEFINED
         * `random` - RANDOM
+    StatisticsItem:
```
zhiltsov-max, line 9043: I recommend adding some help info for `granularity`, `default_view`, `dataseries` and `transformations` fields to clarify about their contents for API users.

cvat/schema.yml
```diff
@@ -12,6 +12,106 @@ info:
...
+      - in: query
+        name: job_id
+        schema:
+          type: integer
+        description: Specify job ID
+      - in: query
+        name: project_id
+        schema:
+          type: integer
+        description: Specify project ID
```
zhiltsov-max, line 75: I suggest moving these 2 parameters to the request body. It looks natural to have them there since they describe the request object.

cvat/schema.yml
```diff
@@ -12,6 +12,106 @@ info:
...
+      - analytics
+      security:
+      - sessionAuth: []
+        csrfAuth: []
+        tokenAuth: []
+      - signatureAuth: []
+      - basicAuth: []
+      responses:
+        '200':
+          description: '...'
```
zhiltsov-max, line 58: Probably, the response description is missing.

cvat/schema.yml
```diff
@@ -12,6 +12,106 @@ info:
     name: MIT License
     url: https://en.wikipedia.org/wiki/MIT_License
 paths:
+  /api/analytics/reports:
+    get:
+      operationId: analytics_list_reports
```
zhiltsov-max, line 17: Probably, this operation should be called `analytics_get_report`.

cvat/apps/iam/permissions.py
```diff
@@ -2018,3 +2026,44 @@ def has_object_permission(self, request, view, obj):
...
+    def create(cls, request, view, obj, iam_context):
+        permissions = []
+        if view.basename == 'analytics_reports':
+            scopes = cls.get_scopes(request, view, obj)
+            for scope in scopes:
+                self = cls.create_base_perm(request, view, scope, iam_context, obj)
+                permissions.append(self)
+
+            if job_id := request.query_params.get('job_id', None):
+                perm = JobPermission.create_scope_view(iam_context, int(job_id))
```
zhiltsov-max, line 2045: The parent context may miss the object org information, so you may need to recreate the context using the job/task/project object.

cvat/apps/analytics_report/report/primary_metrics/objects.py
```diff
@@ -0,0 +1,49 @@
...
+#
+# SPDX-License-Identifier: MIT
+
+from cvat.apps.analytics_report.report.primary_metrics.imetric import IPrimaryMetric
+
+class JobObjects(IPrimaryMetric):
+    _title = "Objects"
+    _description = "Metric shows number of added/changed/deleted objects."
+    _default_view = "histogram"
+    _query = "SELECT toStartOfDay(timestamp) as day, sum(JSONLength(JSONExtractString(payload, {object_type:String}))) as s FROM events WHERE scope = {scope:String} AND job_id = {job_id:UInt64} GROUP BY day ORDER BY day ASC"
```
zhiltsov-max, line 11: Please add a comment why a custom SQL is used here.

cvat/apps/analytics_report/__init__.py
zhiltsov-max, line 1: Please enable code formatting for this component. It can be done by adding a file like [this](https://github.com/opencv/cvat/blob/develop/cvat/apps/quality_control/pyproject.toml) and corresponding lines in [this file](https://github.com/opencv/cvat/blob/develop/dev/format_python_code.sh), [this](https://github.com/opencv/cvat/blob/develop/.github/workflows/black.yml) and [this](https://github.com/opencv/cvat/blob/develop/.github/workflows/isort.yml).

cvat/apps/analytics_report/default_settings.py
```diff
@@ -0,0 +1,8 @@
+# Copyright (C) 2023 Intel Corporation
```
zhiltsov-max, line 1:
```suggestion
# Copyright (C) 2023 CVAT.ai Corporation
```

cvat/apps/analytics_report/report/create.py
```diff
@@ -0,0 +1,319 @@
...
+            intervals = max(1, 1 + (now - last_update_time) // delay)
+            next_update_time = last_update_time + delay * intervals
+            queue_job_id = self._make_regular_queue_job_id(obj, next_update_time)
+
+        return queue_job_id
+
+    class AnalyticsReportsNotAvailable(Exception):
+        pass
+
+    def schedule_analytics_report_autoupdate_job(self, *, job=None, task=None, project=None):
```
zhiltsov-max, line 117: Probably, there should be a check that only 1 arg is passed.

cvat/apps/analytics_report/report/create.py
```diff
@@ -0,0 +1,319 @@
...
+        if rq_job and not self.is_custom_analytics_check_job(rq_job):
+            rq_job = None
+
+        return rq_job
+
+    def is_custom_analytics_check_job(self, rq_job) -> bool:
+        return rq_job.meta.get("job_type") == self._RQ_CUSTOM_ANALYTICS_CHECK_JOB_TYPE
+
+    @classmethod
+    def _check_job_analytics(cls, *, cvat_job_id: int=None, cvat_task_id: int=None, cvat_project_id: int=None) -> int:
```
zhiltsov-max, line 186: Probably, a better name for this function (and the class, actually) should be chosen, because it mentions "job" in the name, while computes reports for tasks and projects as well.

cvat/apps/analytics_report/report/create.py
```diff
@@ -0,0 +1,319 @@
...
+        return {
+            "title": statistics_object.title,
+            "description": statistics_object.description,
+            "granularity": statistics_object.granularity,
+            "default_view": statistics_object.default_view,
+            "transformations": statistics_object.transformations,
+            "dataseries": statistics_object.calculate(),
+        }
+
+    def _compute_report_for_job(self, db_job):
```
zhiltsov-max, line 208:
```suggestion
    def _compute_report_for_job(self, db_job: Job):
```

cvat/apps/analytics_report/report/create.py
```diff
@@ -0,0 +1,319 @@
...
+            total_annotation_speed = JobTotalAnnotationSpeed(db_job, primary_statistics=statistics["annotation_speed"])
+            total_object_count = JobTotalObjectCount(db_job, primary_statistics=statistics["annotation_speed"])
+
+            statistics["total_annotation_speed"] = self._get_statistics_entry(total_annotation_speed)
+            statistics["total_object_count"] = self._get_statistics_entry(total_object_count)
+
+            db_report.statistics = statistics
+            db_report.save()
+
+        return db_report
```
zhiltsov-max, line 241: Probably, this function should return an integer id or other simple object (or maybe it should do the calling function instead). RQ will need to serialize the object (likely via pickle) to save the result.

cvat/apps/analytics_report/report/create.py
zhiltsov-max, line 186: The return value type doesn't seem to be followed. For instance, `_compute_report_for_job` returns a whole report and it's being returned from this function then.

cvat/apps/analytics_report/report/create.py
```diff
@@ -0,0 +1,319 @@
...
+        }
+
+    def _compute_report_for_job(self, db_job):
+        db_report = getattr(db_job, "analytics_report", None)
+        was_created = False
+        if db_report is None:
+            db_report = AnalyticsReport.objects.create(
+                job_id=db_job.id,
+                statistics={})
+            db_report.save()
```
zhiltsov-max, line 215: This `save()` call seems to be extra, `objects.create()` should have called the DB already.

cvat/apps/analytics_report/report/create.py
```diff
@@ -0,0 +1,319 @@
...
+            "default_view": statistics_object.default_view,
+            "transformations": statistics_object.transformations,
+            "dataseries": statistics_object.calculate(),
+        }
+
+    def _compute_report_for_job(self, db_job):
+        db_report = getattr(db_job, "analytics_report", None)
+        was_created = False
+        if db_report is None:
+            db_report = AnalyticsReport.objects.create(
```
zhiltsov-max, line 212: This code is subject to race conditions (2 parallel analytics computation calls are possible). I suggest wrapping this block in a `with transaction.atomic()` section.

cvat/apps/analytics_report/report/create.py
```diff
@@ -0,0 +1,319 @@
...
+                job_id=db_job.id,
+                statistics={})
+            db_report.save()
+            was_created = True
+            db_job.analytics_report = db_report
+
+        # recalculate the report if it is not relevant
```
zhiltsov-max, line 219:
```suggestion
        # recalculate the report if there is no report or the existing one is outdated
```

cvat/apps/analytics_report/report/get.py
```diff
@@ -0,0 +1,127 @@
...
+        query_params["end_date"] = datetime.min
+
+    job_id = query_params.get('job_id', None)
+    task_id = query_params.get('task_id', None)
+    project_id = query_params.get('project_id', None)
+
+    if job_id is None and task_id is None and project_id is None:
+        raise serializers.ValidationError("No any job, task or project specified")
+
+    if [job_id, task_id, project_id].count(True) > 1:
```
zhiltsov-max, line 110: Probably, this line works differently than expected:
```
[1, 2, 0].count(True)
>>> 1
```

```suggestion
    if sum(map(bool, [job_id, task_id, project_id])).count(True) > 1:
```

```suggestion
    if not (bool(job_id) ^ bool(task_id) ^ bool(project_id)):
```

cvat/apps/analytics_report/report/get.py
```diff
@@ -0,0 +1,127 @@
...
+                df["date"] = parser.parse(df["datetime"]).date()
+                del df["datetime"]
+    return statistics
+
+def _get_object_report(obj_model, pk, start_date, end_date):
+    db_obj = obj_model.objects.get(pk=pk)
+    try:
+        db_analytics_report = db_obj.analytics_report
+    except AnalyticsReport.DoesNotExist:
+        raise
```
zhiltsov-max, line 38: Probably, there is no point in this try-catch block. I think it would be reasonable to convert doesn't exist errors (both from reports and `obj`) to `rest_framework.exceptions.NotFoundNotFound` for the response.

cvat-ui/package.json
zhiltsov-max, line 1: If there are no reports, only an error is shown in UI. Given the reports will only be computed on job/task/project updates, many users are likely to think the feature is not working. I suggest adding some more user-friendly information, such as:
- empty charts with no data 
- a message about how to obtain anything useful of the page

image

docker-compose.yml
```diff
@@ -202,6 +202,29 @@ services:
...
+      no_proxy: clickhouse,grafana,vector,nuclio,opa,${no_proxy:-}
+      NUMPROCS: 2
+    command: -c supervisord/worker.analytics_reports.conf
+    volumes:
+      - cvat_data:/home/django/data
+      - cvat_keys:/home/django/keys
+      - cvat_logs:/home/django/logs
+    networks:
+      - cvat
+
```
zhiltsov-max, line 227: I suggest adding 

in `docker-compose.dev.yml`:
```
  cvat_worker_analytics_reports:
    environment:
      CVAT_ANALYTICS_CHECK_JOB_DELAY: 5
      # For debugging, make sure to set 1 process
      # Due to the supervisord specifics, the extra processes will fail and
      # after few attempts supervisord will give up restarting, leaving only 1 process
      NUMPROCS: 1
      CVAT_DEBUG_ENABLED: '${CVAT_DEBUG_ENABLED:-no}'
      CVAT_DEBUG_PORT: '9095'
      COVERAGE_PROCESS_START:
    ports:
      - '9095:9095'
```

in `launch.json`:
```
{
            "name": "REST API tests: Attach to RQ analytics reports worker",
            "type": "python",
            "request": "attach",
            "connect": {
                "host": "127.0.0.1",
                "port": 9095
            },
            "pathMappings": [
                {
                    "localRoot": "${workspaceFolder}",
                    "remoteRoot": "/home/django/"
                },
                {
                    "localRoot": "${workspaceFolder}/.env",
                    "remoteRoot": "/opt/venv",
                }
            ],
            "justMyCode": false,
        },
```

cvat/apps/analytics_report/__init__.py
zhiltsov-max, line 1: I've got an error when tried to compute a report for an old task (by updating a job):

```
  File "/opt/venv/lib/python3.10/site-packages/rq/worker.py", line 936, in perform_job
    rv = job.perform()
  File "/opt/venv/lib/python3.10/site-packages/rq/job.py", line 684, in perform
  File "/opt/venv/lib/python3.10/site-packages/rq/job.py", line 690, in _execute
  File "/home/django/cvat/apps/analytics_report/report/create.py", line 195, in _check_job_analytics
    return cls()._compute_report_for_project(db_project)
  File "/home/django/cvat/apps/analytics_report/report/create.py", line 291, in _compute_report_for_project
    self._compute_report_for_task(db_task)
  File "/home/django/cvat/apps/analytics_report/report/create.py", line 257, in _compute_report_for_task
    job_reports.append(self._compute_report_for_job(db_job))
  File "/home/django/cvat/apps/analytics_report/report/create.py", line 235, in _compute_report_for_job
    statistics["total_annotation_speed"] = self._get_statistics_entry(total_annotation_speed)
  File "/home/django/cvat/apps/analytics_report/report/create.py", line 205, in _get_statistics_entry
    "dataseries": statistics_object.calculate(),
  File "/home/django/cvat/apps/analytics_report/report/derived_metrics/total_annotation_speed.py", line 19, in calculate
    total_wt += ds[1]["value"]
TypeError: unsupported operand type(s) for +=: 'int' and 'NoneType'
```

cvat/apps/analytics_report/report/primary_metrics/imetric.py
```diff
@@ -0,0 +1,48 @@
+# Copyright (C) 2023 CVAT.ai Corporation
+#
+# SPDX-License-Identifier: MIT
+
+from cvat.apps.analytics_report.report.primary_metrics.utils import make_clickhouse_query
+from datetime import datetime, timezone
+
+class IPrimaryMetric():
```
zhiltsov-max, line 8: I suppose all these classes prefixed with `I` to reflect they describe an interface, but they describe abstract classes instead by providing an implementation and defining the fields. If you only wanted to declare fields, it will be better to change declarations to be like this:

```python
class MyClass:
    field: type
```

This will define just an interface with no implementation. For abstract base classes it would be meaningful to add the [`ABCMeta` metaclass](https://docs.python.org/3.10/library/abc.html) (so the class methods that are supposed for redefining were required to be redefined in subclasses):

```
from abc import ABCMeta, abstractmethod

class MyClass(metaclass=ABCMeta):
    @abstractmethod
    def foo(self):
        ... # must be redefined in a subclass
```

cvat/apps/analytics_report/report/primary_metrics/imetric.py
```diff
@@ -0,0 +1,48 @@
...
+    _title = None
+    _description = None
+    _query = None
+    _granularity = None
+    _default_view = None
+    _transformations = []
```
zhiltsov-max, line 14: Such definition will define class fields, not instance fields. I'm not sure if it was supposed.

cvat-ui/package.json
zhiltsov-max, line 1: image

-> "Last year"

zhiltsov-max, line 1: This value is probably extrapolated (the task is just created and I've added just a few annotations):

image

Probably, it would be more natural to show the number of objects created without extrapolating it for the whole hour.

cvat/apps/analytics_report/report/create.py
```diff
@@ -0,0 +1,486 @@
...
+            next_update_time = last_update_time + delay * intervals
+            queue_job_id = self._make_regular_queue_job_id(obj, next_update_time)
+
+        return queue_job_id
+
+    class AnalyticsReportsNotAvailable(Exception):
+        pass
+
+    def schedule_analytics_report_autoupdate_job(self, *, job=None, task=None, project=None):
+        if sum(map(bool, (job, task, project))) > 1:
```
zhiltsov-max, line 132:
```suggestion
        assert sum(map(bool, (job, task, project))) == 1
```

This error looks more like an implementation error than a problem-specific one.

cvat/apps/analytics_report/report/create.py
```diff
@@ -0,0 +1,486 @@
...
+    ) -> bool:
+        if cvat_job_id is not None:
+            queryset = Job.objects.select_related("analytics_report")
+            with transaction.atomic():
+                # The Job could have been deleted during scheduling
+                try:
+                    db_job = queryset.get(pk=cvat_job_id)
+                    db_report = cls._get_analytics_report(db_job)
+                except Job.DoesNotExist:
+                    return False
```
zhiltsov-max, line 231:
```suggestion
                try:
                    db_job = queryset.get(pk=cvat_job_id)
                except Job.DoesNotExist:
                    return False
                
                db_report = cls._get_analytics_report(db_job)
```

cvat/apps/analytics_report/report/create.py
```diff
@@ -0,0 +1,486 @@
...
+                    db_report = cls._get_analytics_report(db_task)
+                except Task.DoesNotExist:
+                    return False
+
+            db_report, job_reports = cls()._compute_report_for_task(
+                db_task=db_task, db_report=db_report
+            )
+
+            with transaction.atomic():
+                # The job could have been deleted during processing
```
zhiltsov-max, line 271:
```suggestion
                # The task could have been deleted during processing
```

cvat/apps/analytics_report/report/create.py
```diff
@@ -0,0 +1,486 @@
...
+        elif cvat_project_id is not None:
+            queryset = Project.objects.select_related("analytics_report").prefetch_related(
+                "tasks__segment_set__job_set"
+            )
+            with transaction.atomic():
+                try:
+                    db_project = queryset.get(pk=cvat_project_id)
+                    db_report = cls._get_analytics_report(db_project)
+                except Project.DoesNotExist:
+                    return False
```
zhiltsov-max, line 310:
```suggestion
            with transaction.atomic():
                try:
                    db_project = queryset.get(pk=cvat_project_id)
                except Project.DoesNotExist:
                    return False
                    
                db_report = cls._get_analytics_report(db_project)
```

cvat/apps/analytics_report/report/create.py
```diff
@@ -0,0 +1,486 @@
...
+        elif cvat_task_id is not None:
+            queryset = Task.objects.select_related("analytics_report").prefetch_related(
+                "segment_set__job_set"
+            )
+            with transaction.atomic():
+                try:
+                    db_task = queryset.get(pk=cvat_task_id)
+                    db_report = cls._get_analytics_report(db_task)
+                except Task.DoesNotExist:
+                    return False
```
zhiltsov-max, line 264:
```suggestion
                try:
                    db_task = queryset.get(pk=cvat_task_id)
                except Task.DoesNotExist:
                    return False
                    
                db_report = cls._get_analytics_report(db_task)
```

zhiltsov-max (commented):
Probably, it makes sense to add the last report update time on the analytics page.

cvat/schema.yml
```diff
@@ -12,6 +12,100 @@ info:
...
+        tokenAuth: []
+      - signatureAuth: []
+      - basicAuth: []
+      responses:
+        '201':
+          content:
+            application/vnd.cvat+json:
+              schema:
+                $ref: '#/components/schemas/RqId'
+          description: A analytics report request has been computed
```
SpecLad, line 96: If this is the response when the report has been created, shouldn't it have an empty body (or a body with some information about the report), rather than an `RqId`?

cvat/schema.yml
```diff
@@ -8884,6 +9037,34 @@ components:
...
+        description:
+          type: string
+        granularity:
+          nullable: true
+          oneOf:
+          - $ref: '#/components/schemas/GranularityEnum'
+          - $ref: '#/components/schemas/NullEnum'
+        default_view:
+          $ref: '#/components/schemas/DefaultViewEnum'
+        dataseries:
```
SpecLad, line 9054: Shouldn't it be two words (i.e. `data_series`)?

cvat/schema.yml
```diff
@@ -8884,6 +9037,34 @@ components:
...
+          type: object
+          additionalProperties: {}
+        transformations:
+          type: array
+          items:
+            type: object
+            additionalProperties: {}
```
SpecLad, line 9061: Is there no way to provide more static typing information here?

cvat/schema.yml
```diff
@@ -8884,6 +9037,34 @@ components:
         * `natural` - NATURAL
         * `predefined` - PREDEFINED
         * `random` - RANDOM
+    StatisticsItem:
```
SpecLad, line 9040: IMO, "Metric" would be a more descriptive name for this structure.

cvat/schema.yml
```diff
@@ -5826,6 +5920,47 @@ components:
...
+        job_id:
+          type: integer
+        task_id:
+          type: integer
+        project_id:
+          type: integer
+        statistics:
+          type: object
+          additionalProperties:
+            $ref: '#/components/schemas/StatisticsItem'
```
SpecLad, line 5940: It's unclear why this is an object rather than an array. What is it indexed by?

bsekachev:
Shell we add

```
 ports:
      - '8123:8123'
```

To dev compose file for clickhouse?

bsekachev:
I only created 7 objects in a task. It shows me 8 updated: 
image

bsekachev:
- If I use "Remove annotations" button from annotation menu, it does not show any removed objects in statistics
- Try this sequence: create a task, draw several objects, create analytics report manually, go to analytics page. Everything is fine. Then go to the task again, remove one object, save, create one more analytics report manually and go to analytics page again. It will show created objects - 1 were updated, but they weren't touched after creation. 
- Agree with Maxim, better to see report computation time
- Shell we consider such actions like "Uploading annotations" when computing annotations speed? 
image
- I would propose improving header message, using something more descriptive than task/project name with a hyperlink to a target resource, for example: 
image
- ID for project and for task are on different places: 
image
image

bsekachev:
It shows different order of components for a project where were not any analytics reports computed and for a task where they were computed. Not critical, but maybe by default we should keep components order? Apply some kind of sorting?
image
image

cvat-canvas/src/typescript/shared.ts
```diff
@@ -41,7 +41,7 @@ export interface DrawnState {
     occluded?: boolean;
     hidden?: boolean;
     lock: boolean;
-    source: 'AUTO' | 'SEMI-AUTO' | 'MANUAL';
+    source: 'AUTO' | 'SEMI-AUTO' | 'MANUAL' | 'FILE';
```
bsekachev, line 44: I believe we need to update ``computeNewSource`` function in ``cvat-core/src/annotations-objects.ts``. 
It is responsible for getting new source after an object was modified somehow.

cvat/schema.yml
```diff
@@ -7596,6 +7758,36 @@ components:
...
+        granularity:
+          nullable: true
+          oneOf:
+          - $ref: '#/components/schemas/GranularityEnum'
+          - $ref: '#/components/schemas/NullEnum'
+        default_view:
+          $ref: '#/components/schemas/DefaultViewEnum'
+        data_series:
+          type: object
+          additionalProperties: {}
```
SpecLad, line 7779: What about typing information here?

cvat/schema.yml
```diff
@@ -6089,6 +6220,19 @@ components:
...
+      type: object
+      properties:
+        left:
+          type: string
+          nullable: true
+        operator:
+          $ref: '#/components/schemas/OperatorEnum'
+        right:
+          type: string
+          nullable: true
```
SpecLad, line 6233: It's not clear what the "left" and "right" strings are supposed to be. This could use some documentation.

bsekachev (approved):

zhiltsov-max (approved):

PR #6394, opened 2023-06-28
Fix overly broad exception conversion in `LambdaFunction.invoke`

SpecLad:
### Motivation and context

The intent of the `try`/`except` statement is to catch accesses to missing members of the `data` dictionary, but due to the large amount of code in the `try` block, it may end up catching entirely unrelated `KeyError`s. Those unrelated `KeyError`s should not be converted to `ValidationError`s, since they might not have anything to do with input validation, and the conversion will make it harder to debug these exceptions.

An example of these misapplied conversions is a recent bug where a `KeyError` was coming from inside `_get_image` (fixed by f6420eb7).

To fix this, make sure to only catch `KeyError`s emitted by accesses to `data`.

### How has this been tested?

I manually checked that a) invoking functions still works, and b) omitting required parameters still results in a 400 error.

### Checklist

### License


zhiltsov-max:
Please check if this needs a new test in this application tests.

SpecLad:
> Please check if this needs a new test in this application tests.

I don't think this change is testable, because it only makes a difference when there's an internal error in CVAT - which isn't supposed to happen.

nmanovic (approved):

PR #6455, opened 2023-07-11
Keep model input data unchanged in SDK

zhiltsov-max:
### Motivation and context

Nested containers may be modified during the model input data parsing in class constructors. This can lead to subtle memory errors, which are very hard to find. In CVAT, this helps to avoid unexpected problems in tests, such as one test affecting another one by subtly changing test assets.

### How has this been tested?

Unit tests

### Checklist

### License


cvat-sdk/gen/templates/openapi-generator/model_utils.mustache
```diff
@@ -1311,6 +1311,8 @@ def validate_and_convert_types(input_value, required_types_mixed, path_to_item,
         # for this type, there are not more inner variables left to look at
         return input_value
     if isinstance(input_value, list):
+        input_value = copy(input_value) # avoid changing the input value
```
SpecLad, line 1314: I think it would be better to do something like `output_value = [None] * len(input_value)`, and then adjust the following code accordingly. That way, it'll be clear that we are not, in fact, returning the input value.

Also, you can move this below the first `if` block, since there's no need to create a new list if the original is empty.

zhiltsov-max: > [Add `output_value = `]

Ok, makes sense.

> Also, you can move this below the first if block, since there's no need to create a new list if the original is empty.

I don't agree with this. This is python, for sure, but I'd like to avoid using and storing the input objects if they are mutable. If we store an empty list from the input data, it can lead to something like this:

```
inputs = { 'field': [] }
m1 = model(**inputs)

inputs["field"].append(42)
m2 = model(**inputs)

m1 == m2
```

I'm not sure this is desired.

SpecLad: Okay, that's a good point. I thought `validate_and_convert_types` was only used for transient values (e.g. when sending a request), but if it's used in model constructors too, then it makes sense to copy the input data.

SpecLad (approved):

PR #6456, opened 2023-07-11
Honey Pot fixes

zhiltsov-max:
### Motivation and context

Fixed:

UI:
- [x] opening skeleton point RMB menu leads to UI fail
- [x] annotation conflicts are accumulated when changing frames, if the annotation has multiple conflicts
- [x] improve OKS sigma description

Server:
- [x] average mean iou can be negative

### Checklist

### License


cvat-ui/src/reducers/review-reducer.ts
```diff
@@ -131,8 +131,28 @@ export default function (state: ReviewState = defaultState, action: any): Review
...
+                                if (prop === 'description') {
+                                    return description;
+                                }
+
+                                // By default, it looks like Reflect.get(target, prop, receiver)
+                                // which has a different value of `this`
+                                const value = target[prop];
+                                if (value instanceof Function) {
+                                    return function (...args) {
+                                        return value.apply(this === receiver ? target : this, args);
```
bsekachev, line 148: Can we just use ``target`` context here?

cvat-ui/src/reducers/review-reducer.ts
```diff
@@ -131,8 +131,28 @@ export default function (state: ReviewState = defaultState, action: any): Review
...
+                                    };
+                                }
+                                return value;
```
bsekachev, line 151: Not sure what is problem with using Reflect here.
In this example context is preserved:
```js
(() => {
    const obj = { a: 5, test: function () { return this.a } };
    const proxy = new Proxy(obj, {
        get(target, prop, receiver) {
            if (prop === 'def') { return 'default' }
            return Reflect.get(target, prop, receiver);
        }
    })
    console.log(proxy.def) // default
    console.log(proxy.test()) // 5
})()
```

PR #6483, opened 2023-07-14
[Dependent] Client side lambdas

SpecLad:
Depends on #6428.

### Checklist

### License


cvat-cli/src/cvat_cli/parser.py
```diff
@@ -369,6 +369,21 @@ def make_cmdline_parser() -> argparse.ArgumentParser:
         help="time interval between checks if archive processing was finished, in seconds",
     )
 
+    #######################################################################
+    # Auto-annotate
+    #######################################################################
+    auto_annotate_task_parser = task_subparser.add_parser(
+        "auto-annotate", description="Automatically annotate a CVAT task."
```
zhiltsov-max, line 376: Probably, it should be clarified, that this command doesn't just execute a task auto-annotation on the remote server, but runs a local script instead (though, a command for the first would also be nice to have as an option).

tests/python/cli/test_cli.py
```diff
@@ -302,3 +302,16 @@ def test_can_control_organization_context(self):
...
+
+    def test_auto_annotate(self, fxt_new_task: Task):
+        annotations = fxt_new_task.get_annotations()
+        assert not annotations.shapes
+
+        self.run_cli(
+            "auto-annotate",
+            str(fxt_new_task.id),
+            f"--function={__package__}.example_function",
+        )
```
zhiltsov-max, line 314: I suggest adding several other tests for different file placement, because we don't have a package when we call it from the actual CLI (in the tests we cheat a little).

tests/python/cli/example_function.py
```diff
@@ -0,0 +1,18 @@
+# Copyright (C) 2023 CVAT.ai Corporation
```
zhiltsov-max, line 1: I suppose this file can be used as an example by many people, so I suggest improving the style and naming to spread good behaviors. I'd suggest to add type annotations to the function. I'm not sure we should really use the same low level API models, since we don't really request anything here, and changes there may easily break user code for no reason. Maybe we can just define no-op inherited classes just to expose them in this API as the first step to decoupling.

tests/python/cli/example_function.py
```diff
@@ -0,0 +1,18 @@
...
+#
+# SPDX-License-Identifier: MIT
+
+import cvat_sdk.auto_annotation as cvataa
+import cvat_sdk.models as models
+
+spec = cvataa.DetectionFunctionSpec(
```
zhiltsov-max, line 8: Maybe *manifest* is a more common word for this.

tests/python/cli/example_function.py
```diff
@@ -0,0 +1,18 @@
...
+    labels=[
+        models.PatchedLabelRequest(name="car", id=0),
+    ]
+)
+
+
+def detect(context, image):
```
zhiltsov-max, line 15: Not everyone is happy with just image bytes as the input. Probably, it would be nice to include the current frame info in the context or function arguments. It also seems reasonable to add a progress callback.

cvat-sdk/cvat_sdk/auto_annotation/interface.py
```diff
@@ -0,0 +1,33 @@
...
+
+import attrs
+import PIL.Image
+from typing_extensions import Protocol
+
+import cvat_sdk.models as models
+
+
+@attrs.frozen(kw_only=True)
+class DetectionFunctionSpec:
```
nmanovic, line 15: From the user's perspective, I would not recommend using PatchedLabelRequest and LabeledShapeRequest. They are counterintuitive. 

I will recommend defining alternative naming. Something like Label or Shape.

Another approach is to define some high-level interfaces like LabelingSpec and LabeledFrame and use them.

In both cases, the interface will be clear and attractive.

Also, even I can read comments about DetectionFunctionContext, I'm not sure that we need it. How are you going to use it in the future?



defining the "Label" interface in HL SDK. Something like Sequence[Label] would be better (PatchedLabelRequest -> Label). Another approach is to provide LabelingSpec in SDK and define it as a sequence of labels.

The same comment I can leave about LabeledShapeRequest. It is better to have just Shape. Let's try to improve our high-level interface for users. We can probably define LabeledFrame interface or something like that

tests/python/cli/example_function.py
nmanovic, line 1: Let's try to add a real example based on https://github.com/ultralytics/ultralytics. As you suggested, we can create a python module with a detection function that is based on ultralytics and use it to explain how to write such functions. Also, we can use it for a demonstration to our users.
Without a way to run the functionality out of the box, it will demotivate 90% of regular users.

tests/python/cli/test_cli.py
```diff
@@ -302,3 +302,16 @@ def test_can_control_organization_context(self):
...
         assert org_task_id in all_task_ids
+
+    def test_auto_annotate(self, fxt_new_task: Task):
+        annotations = fxt_new_task.get_annotations()
+        assert not annotations.shapes
+
+        self.run_cli(
+            "auto-annotate",
+            str(fxt_new_task.id),
+            f"--function={__package__}.example_function",
```
nmanovic, line 313: Again, from a regular user perspective, I will recommend supporting at least several interfaces. I'm not sure that we have to force users to know what a Python module is. At the same time, many users know what is a directory. Do you think we can redesign the interface and use widely known concepts amount CLI users?

cvat-sdk/cvat_sdk/auto_annotation/driver.py
```diff
@@ -0,0 +1,226 @@
...
+    shapes = []
+
+    context = DetectionFunctionContext()
+
+    for sample in dataset.samples:
+        frame_shapes = function.detect(context, sample.media.load_image())
+        mapper.validate_and_remap(frame_shapes, sample.frame_index)
+        shapes.extend(frame_shapes)
+
+    client.logger.info("Uploading annotations to task %d", task_id)
```
nmanovic, line 223: Here we force users to wait till all frames are annotated. In CLI we should be able to show progress and also I believe it will be useful to have a configuration option to upload data in batches. What do you think?

cvat-cli/src/cvat_cli/parser.py
```diff
@@ -369,6 +369,21 @@ def make_cmdline_parser() -> argparse.ArgumentParser:
...
+    )
+    auto_annotate_task_parser.add_argument("task_id", type=int, help="task ID")
+    auto_annotate_task_parser.add_argument(
+        "--function",
+        help="name of module to use as the function",
+        metavar='MODULE',
+        required=True,
+        dest="function_module",
+    )
```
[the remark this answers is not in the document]
nmanovic: Need to discuss this. From one point of view, some of these parameters can be convenient:
- job_id(s) - probably we should be able to annotate jobs as well. I agree. But I don't think that we need the argument here
- clear_existing - better to have just a separate command to remove annotation. Otherwise we can get very complicated interface.
- script file path - OK, instead of a module name
- strict - agree and maybe it should be a default mode. Thus we avoid many complains, that a function doesn't annotate a task. I met the problem today.
- skip-errors - I'm not sure we want to skip any errors.

cvat-cli/src/cvat_cli/parser.py
SpecLad: Yeah, that's a fair point. I'm thinking of maybe renaming this command to `auto-annotate-local` to better underscore the distinction (and perhaps `auto-annotate` can run server-side auto-annotation instead).

cvat-cli/src/cvat_cli/parser.py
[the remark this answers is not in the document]
SpecLad: My thoughts:

> `job_id(s)`

I think ideally the CLI should have separate sets of task and job based commands (e.g. `cvat-cli task list` / `cvat-cli job list`). IIUC, job auto-annotation won't require you to use any task-based APIs, so it would make more sense to implement it as a job subcommand (`cvat-cli job auto-annotate`).

However, that would be beyond the scope of this PR.

> `clear_existing` / `append`

I agree with Maxim that this should be an option. Moreover, the default should probably be "append". Currently, the command clears existing annotations, but that's mainly just because it's easier. Defaulting to `append` will avoid accidental data loss.

@nmanovic I think we _should_ have a separate command to clear annotations, but it's not very convenient to run a separate command every time you want to reannotate, so I think an option for this command would be useful to have.

> script file path

I'm thinking of adding a `--function-file` option as an alternative to `--function` (and maybe renaming the latter to `--function-module`). Instead of importing, this will load the file as a string and compile it. It will require that the file has no relative imports, though.

> an extra parameter to for module root directory

I don't know what this means.

> `strict`

Yeah, fair point. I also agree with Nikita that this should be the default.

> `skip-errors`

I get the idea, but then what do you do with frames that were skipped? There's no easy way to find them afterwards, and even if we could find them, we can't rerun the function on just those frames.

We could consider some solution for this, but IMO it goes beyond the current scope.

cvat-sdk/cvat_sdk/auto_annotation/interface.py
SpecLad: I think I'll do something of a middle ground. I'll add factory functions with nice names and nice interfaces that construct the same old `*Request` objects. That way, the function implementation can look nice and clean, but we're not introducing any new concepts (and the users can still construct the low-level request objects if the factory functions are insufficient for some reason).

> Also, even I can read comments about DetectionFunctionContext, I'm not sure that we need it. How are you going to use it in the future?

See the thread directly above this one.

cvat-sdk/cvat_sdk/auto_annotation/driver.py
SpecLad: Regarding progress: yes we should.

Regarding uploading data in batches: I think notionally it could be useful, but in practice there are problems:

* I'm not sure if the API supports partial uploads like this.
* If the annotation fails mid-way, there's no way to restart it from where it left off (and it it doesn't fail mid-way, then what's the point in uploading in batches?).

tests/python/cli/example_function.py
SpecLad: > I suppose this file can be used as an example by many people, so I suggest improving the style and naming to spread good behaviors. I'd suggest to add type annotations to the function.

I can add types, sure. I'm not sure if this is a good example to follow, though... I think a YOLOv8-based function that @nmanovic suggested in another thread would be a better example.

>  I'm not sure we should really use the same low level API models, since we don't really request anything here, and changes there may easily break user code for no reason.

I mean... that's why we shouldn't make such changes. Naturally, breaking changes cause breakage.

cvat-cli/src/cvat_cli/parser.py
zhiltsov-max: The current name is good enough (though I'd like it to be 1 word somehow), we can just extend the description.

tests/python/cli/test_cli.py
zhiltsov-max: - A test that we can execute a function from a file/module in the current directory
- A test that we can execute a function from a file/module outside the current directory subtree, and which is not installed in the current system or python environment (e.g. `cvat-cli auto-annotate --function ../../my_cvat_model.py ...`

tests/python/cli/example_function.py
SpecLad: Type annotations added.

cvat-sdk/cvat_sdk/auto_annotation/interface.py
SpecLad: > I'll add factory functions with nice names and nice interfaces

I did it.

tests/python/cli/example_function.py
SpecLad: Added.

tests/python/cli/test_cli.py
SpecLad: I added an alternate option that takes a file path.

cvat-cli/src/cvat_cli/parser.py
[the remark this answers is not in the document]
zhiltsov-max: > Instead of importing, this will load the file as a string and compile it. It will require that the file has no relative imports, though.

> > an extra parameter to for module root directory

> I don't know what this means.

I've given an example in the test comments. Basically, it's needed both for relative imports and for fully-qualified imports to specify where to look for the import path (like `PYTHONPATH` does). Without this, it's hard to use scripts relying on a customized framework (e.g. mmdetection), lying somewhere near the cvat adapter script, or just outside the current directory:

```bash
# .../some/dir/mymodel/mmdetection/...
# .../some/dir/mymodel/cvat_adapter.py

cvat-cli auto-annotate --module-root '.../some/dir/' --module 'mymodel.cvat_adapter' ...
```

> > skip-errors

> I get the idea, but then what do you do with frames that were skipped? There's no easy way to find them afterwards, and even if we could find them, we can't rerun the function on just those frames.

I thought about things like OOM errors on some frames. Not sure how often it happens in real applications.

Actually, the implementation allows execution on specific frames (except for tracks), as the function is executed on separate frames, so it's only a question of API/CLI. I also see an option to execute the model only on some part of the task a useful feature, though the CLI use may be complicated in this case. Probably, we can leave this for future updates.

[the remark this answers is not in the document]
SpecLad: > I've given an example in the test comments. Basically, it's needed both for relative imports and for fully-qualified imports to specify where to look for the import path (like PYTHONPATH does). 

So why not just use `PYTHONPATH`?

> Actually, the implementation allows execution on specific frames (except for tracks), as the function is executed on separate frames, so it's only a question of API/CLI. 

It's true that we can run the function on specific frames, but how do we know which frames to use (in the context of recovery from a partially successful auto-annotation run?

> I also see an option to execute the model only on some part of the task a useful feature, though the CLI use may be complicated in this case. Probably, we can leave this for future updates.

I agree.

tests/python/cli/test_cli.py
SpecLad: I don't see the point in these tests to be honest. We don't do anything unusual with the import mechanics. If Python can import a module, then so can we (and vice versa).

cvat-cli/src/cvat_cli/parser.py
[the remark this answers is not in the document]
zhiltsov-max: > So why not just use PYTHONPATH?

Probably, it's an implementation detail, and this obliges users to learn environment variables.

> how do we know which frames to use (in the context of recovery from a partially successful auto-annotation run?

It could be reported somehow, e.g. in the logs.

cvat-cli/src/cvat_cli/parser.py
SpecLad: Description extended.

cvat-cli/src/cvat_cli/parser.py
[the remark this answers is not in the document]
SpecLad: I added `--clear-existing` and `--allow-unmatched-labels`. Correspondingly, I changed the defaults to not erase old annotations, and abort when a function declares labels not configured in the task.

> Probably, it's an implementation detail, and this obliges users to learn environment variables.

The user would have to learn something either way, and at least they might already know about `PYTHONPATH`.

tests/python/cli/example_function.py
SpecLad: I added the frame name to the context.

cvat-sdk/cvat_sdk/auto_annotation/driver.py
SpecLad: I added progress reporting.

tests/python/cli/example_function.py
```diff
@@ -0,0 +1,23 @@
...
+    ],
+)
+
+
+def detect(
+    context: cvataa.DetectionFunctionContext, image: PIL.Image.Image
+) -> List[models.LabeledShapeRequest]:
+    return [
+        cvataa.rectangle(0, [1, 2, 3, 4]),
+    ]
```
zhiltsov-max, line 23: How about changing the interface to be the following:
- The function does not return anything
- There is a function to add an annotation for a frame (e.g. `context.add_rectangle()`)

This way:
- the users won't need to know the return annotation types (there can be another function for the users who know)
- we can change the function API without breaking it
- we can manage annotation storage (and optimize it, potentially)

zhiltsov-max (approved):

tests/python/cli/example_function.py
SpecLad: It is a tempting proposal. 🤔 I'm just not sure if I'll have time to implement it; it's a pretty big change.

PR #6501, opened 2023-07-18
Close av contaner after use

azhavoro:
### Motivation and context

Test: run YOLO v7 on 350 frames on video task 
before:
wo_fix
after:
with_fix

full records:
https://asciinema.org/a/kIp3QqsMd3MNs5Ft4LCq97vp7
https://asciinema.org/a/qpwdvrbET6CBM8k5iNc6sU2Ff

### Checklist

### License


cvat/apps/engine/media_extractors.py
```diff
@@ -429,74 +429,71 @@ def _has_frame(self, i):
 
         return False
 
-    def _decode(self, container):
-        frame_num = 0
-        for packet in container.demux():
-            if packet.stream.type == 'video':
+    def __iter__(self):
+        with self._get_av_container() as container:
```
SpecLad, line 433: I should note that this doesn't completely fix the problem, because the container will only be closed if the generator reaches the end or if there's an exception raised within it. If the generator is simply abandoned by the calling code before it reaches the end, then it'll never gain control back, so it'll never close the container.

So a more thorough fix would be to track down all places where this `__iter__` is called and add calls to `generator.close`.

I'm not saying you should do it in _this_ PR (the problem has already existed before this PR), but it should be done at some point, otherwise there can still be subtle memory leaks.

azhavoro: Ok, let's do it in another PR

SpecLad (approved):

PR #6556, opened 2023-07-24
Revamp the progress reporting API

SpecLad:
### Motivation and context

There are several problems with how progress reporting is handled in the SDK, both on the interface and implementation level:

* The user is supposed to know, for a given function, which units it will report progress in. This is unnecessary coupling and prevents us from switching to different units, or to have several progress bars using different units.

* To create a TqdmProgressReporter, you have to create a tqdm instance, which immediately draws a progress bar. This works poorly if the function prints any log messages before the progress actually starts.

* There's no easy way to automatically call `finish` on a progress bar, so some functions don't (for example, `Downloader.download_file`). This can cause unexpected output, since tqdm will refresh the progress bar in a background thread (possibly after we've already printed something else).

* `iter` is basically broken, because it divides by `period`, which is 0 in all current implementations.

* Even ignoring that, it's hard to use correctly, because you need to manually call `finish` in case of an exception.

* `split` is not implemented and not used.

* `StreamWithProgress.seek` assumes that the progress bar is at 0 at the start, and so does `ProgressReporter.iter`. The former also works incorrectly if the second argument is not `SEEK_SET`.

Fix these problems by doing the following:

* Add a new `start2` method which accepts arbitrary tqdm parameters. The default implementation calls `start`, so that if a user has implemented the original interface, it'll keep working.

* Add a `TqdmProgressReporter2` that accepts tqdm parameters instead of a tqdm instance, and only creates an instance after `start2` is called. Use it where `TqdmProgressReporter` was used before. The old `TqdmProgressReporter` is kept for compatibility, but it doesn't support any `start2` arguments other than those supported by the original `start`.

* Add a `task` context manager, which automatically calls `start2` and `finish`. Use it everywhere instead of explicit `start`/`finish` calls. Remove `start`/`finish` calls from `StreamWithProgress` and `iter`.

* Implement basic assertions to ensure that `start2` and `finish` are used correctly.

* Remove `period` and `split`.

* Rewrite `StreamWithProgress.seek` and `ProgressReporter.iter` to use relative progress reports.

These changes should be backwards compatible for users who pass predefined or custom progress reporters into SDK functions. They are not backwards compatible for users who try to use progress reporters directly (e.g. calling `start`/`finish`). I don't consider that a significant issue, since the purpose of the `ProgressReporter` interface is for the user to get progress information from the SDK, not for them to use it in their own code.

Originally developed for #6483.

### Checklist

### License

### DIFF
--- a/cvat-sdk/cvat_sdk/core/uploading.py
+++ b/cvat-sdk/cvat_sdk/core/uploading.py
@@ -7,7 +7,7 @@ from __future__ import annotations
 import json
 import os
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Sequence, Tuple
+from typing import TYPE_CHECKING, Any, ContextManager, Dict, List, Optional, Sequence, Tuple
 
 import requests
 import urllib3
@@ -180,12 +180,24 @@ class Uploader:
         # query params are used only in the extra messages
         assert meta["filename"]
 
+        if pbar is None:
+            pbar = NullProgressReporter()
+
+        file_size = filename.stat().st_size
+
         self._tus_start_upload(url, query_params=query_params)
-        real_filename = self._upload_file_data_with_tus(
-            url=url, filename=filename, meta=meta, pbar=pbar, logger=logger
-        )
+        with self._uploading_task(pbar, file_size):
+            real_filename = self._upload_file_data_with_tus(
+                url=url, filename=filename, meta=meta, pbar=pbar, logger=logger
+            )
         query_params["filename"] = real_filename
         return self._tus_finish_upload(url, query_params=query_params, fields=fields)
+
+    @staticmethod
+    def _uploading_task(pbar: ProgressReporter, total_size: int) -> ContextManager[None]:
+        return pbar.task(
+            total=total_size, desc="Uploading data", unit_scale=True, unit="B", unit_divisor=1024
+        )
 
     def _wait_for_completion(
         self,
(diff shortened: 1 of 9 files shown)


cvat-sdk/cvat_sdk/core/uploading.py
```diff
@@ -180,13 +180,25 @@ def upload_file(
...
+            real_filename = self._upload_file_data_with_tus(
+                url=url, filename=filename, meta=meta, pbar=pbar, logger=logger
+            )
         query_params["filename"] = real_filename
         return self._tus_finish_upload(url, query_params=query_params, fields=fields)
 
+    @staticmethod
+    def _uploading_task(pbar: ProgressReporter, total_size: int) -> ContextManager[None]:
+        return pbar.task(
+            total=total_size, desc="Uploading data", unit_scale=True, unit="B", unit_divisor=1024
```
zhiltsov-max, line 199: I see several problems with this change:

- It supposes that the progress reporter is tqdm
- There is no way for the user to implement their reporter correctly, because the parameters used in SDK are not specified anywhere, they are just passed, and they are similar everywhere, hopefully
- The suffix `2` is quite cryptic and tells nothing about the differences, probably it can be replaced with `Lazy` or `Deferred`

SpecLad: > * The implementation supposes that the progress reporter is tqdm
> * The tqdm has quite a broad list of parameters, it will be hard to implement a custom reporter correctly

That's not quite true, because a reporter doesn't have to support every parameter that tqdm supports. It can simply use whatever parameters it recognizes, and ignore the rest. A barebones reporter doesn't have to use _any_ information other than `total` (maybe not even `total`...).

That said, I think I can improve this by explicitly listing the parameters that we actually use right now, and adding a note saying that any kwargs should be ignored. That way, users have more guidance on which parameters they should pay attention to, and there's a way for us to introduce non-tqdm parameters in the future if we need them.

> The suffix `2` is quite cryptic and tells nothing about the differences, probably it can be replaced with Lazy or Deferred

The only thing that the `2` is supposed to communicate is that it's better than 1. Which, to me, seems like the main thing that users need to know in this case. If you really want to, I can use `Deferred` instead.

SpecLad: > Sounds good. I think it's the right direction.

Okay, I did it.

> I won't claim that deferred is the best variant, but 2 feels like "TODO" for me - like if we couldn't find a good name and just left it for future.

Mind you, the good name is `TqdmProgressReporter` - I just couldn't use it because it was already taken. 🙂 

Well, I renamed it to `DeferredTqdmProgressReporter` anyway.

zhiltsov-max (approved):

PR #6663, opened 2023-08-12
Removed extra endpoints from the server which used only to get client assets

bsekachev:
### Checklist

### License


cvat-ui/webpack.config.js
```diff
@@ -209,6 +209,14 @@ module.exports = (env) => {
                         from: '../node_modules/onnxruntime-web/dist/*.wasm',
                         to  : 'assets/[name][ext]',
                     },
+                    {
+                        from: 'src/assets/opencv*.js',
+                        to  : 'assets/opencv.js',
```
klakhov, line 214: What do you think about creating separate directory for opencv in assets? It would be a mess if will keep adding different files without any managment

klakhov (commented):
While looking at assests directory I've found [excessive file](https://github.com/opencv/cvat/blob/develop/cvat-ui/src/assets/rotate-icon%20(copy).svg). Its not used anywhere, I belive we should delete it.

Should [this line](https://github.com/opencv/cvat/blob/develop/docker-compose.yml#L54) also be modified if we remove `opencv` app?

PR #6694, opened 2023-08-17
Debugging E2E tests

SpecLad:
Ignore this.


bsekachev:
Tried to fix couple of cypress tests here #6693
Not really sure why they fail sometimes, only some ideas

PR #6702, opened 2023-08-19
Improved task creation pipeline

bsekachev:
### Motivation and context
**Instead of this:**

image

image

**Show this:**

Снимок экрана 2023-08-19 115926

image

### Checklist

### License


cvat-core/src/server-proxy.ts
```diff
@@ -1065,48 +1067,89 @@ async function restoreProject(storage: Storage, file: File | string) {
...
+        return listenToCreateCallbacks[id].promise;
+    }
+
+    const promise = new Promise<SerializedTask>((resolve, reject) => {
+        const { backendAPI } = config;
+        const params = enableOrganization();
+        async function checkStatus(): Promise<void> {
+            try {
+                const response = await Axios.get(`${backendAPI}/tasks/${id}/status`, { params });
+                if (['Queued', 'Started'].includes(response.data.state)) {
```
klakhov, line 1090: Why dont we use `RQStatus` enum here? Should server return status in lowercase?

cvat-core/src/session-implementation.ts
```diff
@@ -523,6 +523,17 @@ export function implementTask(Task) {
         });
     };
 
+    Task.prototype.listenToCreate.implementation = async function (
+        onUpdate: (state: RQStatus, progress: number, message: string) => void = () => {},
+    ): Promise<TaskClass> {
+        if (Number.isInteger(this.id) && this.size === 0) {
```
klakhov, line 529: Task size can be != 0 with id=null?

cvat-core/src/server-proxy.ts
bsekachev: > Why dont we use RQStatus enum here? 

Exactly because server returns statuses not in lower case. 
I do not want to change API scheme

klakhov: But I suppose the statuses returned from server are the ones in RQStatus, we can cast it to lowercase on UI

bsekachev: Done

klakhov (approved):

PR #6773, opened 2023-08-31
Rework chunk cache

azhavoro:
### Checklist

### License


SpecLad:
Please rebase, as this conflicts with #6761.

docker-compose.yml
```diff
@@ -18,8 +18,16 @@ services:
 
   cvat_redis:
     container_name: cvat_redis
-    image: redis:7.0-alpine
+    image: eqalpha/keydb:x86_64_v6.3.2
     restart: always
+    command: [
+      "keydb-server",
+      "/etc/keydb/keydb.conf",
+      "--storage-provider", "flash", "/data/flash",
```
SpecLad, line 26: Isn't FLASH support an enterprise-only feature? It says so here: <https://docs.keydb.dev/docs/config-file/>.

docker-compose.yml
```diff
@@ -18,8 +18,16 @@ services:
...
+      "--maxmemory", "5G",
+      "--maxmemory-policy", "allkeys-lfu"]
```
SpecLad, line 28:
```suggestion
      "--maxmemory-policy", "allkeys-lfu",
    ]
```

helm-chart/Chart.yaml
```diff
@@ -59,3 +54,8 @@ dependencies:
     version: 10.24.0
     repository: https://helm.traefik.io/traefik
     condition: traefik.enabled
+
+  - name: keydb
+    version: 0.48.0
+    repository: https://enapter.github.io/charts/
```
SpecLad, line 60: I see that the [Enapter/charts](https://github.com/Enapter/charts) repository has been archived not too long ago. Perhaps we shouldn't rely on it?

SpecLad (changes requested):

docker-compose.yml
azhavoro: https://docs.keydb.dev/blog/2020/01/20/blog-post/
Flash is an available feature for the open source version: https://github.com/Snapchat/KeyDB#building-keydb and is enabled by default for provided docker images https://github.com/Snapchat/KeyDB/blob/main/pkg/docker/Dockerfile#L56.
I don't know why the example config still has this note in it

helm-chart/Chart.yaml
azhavoro: Unfortunately there are no alternatives available, this repository is mentioned in the documentation

helm-chart/values.yaml
```diff
@@ -226,19 +226,23 @@ postgresql:
...
-    password: cvat_redis
-  cluster:
-    enabled: false
+    name: cvat-keydb-secret
+    password: cvat_keydb
+  multiMaster: false
+  activeReplicas: false
+  nodes: 1
+  configExtraArgs:
+  - storage-provider: ["flash", "/data/flash"]
```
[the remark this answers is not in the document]
azhavoro: https://github.com/Enapter/charts/blob/master/keydb/templates/sts.yaml#L129
persistentVolume is enabled by default

docker-compose.yml
```diff
@@ -18,8 +18,16 @@ services:
...
+    volumes:
+      - cvat_cache_db:/data
```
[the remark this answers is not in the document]
azhavoro: I would prefer to use the default location for the following reasons:
1. Same location as in the helm configuration
2. No need to fix permissions for the new folder (`/data` already has keydb as owner)

cvat/settings/testing_rest.py
```diff
@@ -24,3 +24,5 @@
 HEALTH_CHECK = {
     'DISK_USAGE_MAX': 100,  # percent
 }
+
+CACHES["media"]["TIMEOUT"] = 1
```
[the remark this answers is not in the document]
azhavoro: added

cvat/settings/base.py
```diff
@@ -40,6 +41,10 @@
 ALLOWED_HOSTS = os.environ.get('ALLOWED_HOSTS', 'localhost,127.0.0.1').split(',')
 INTERNAL_IPS = ['127.0.0.1']
 
+REDIS_HOST = os.getenv('CVAT_REDIS_HOST', 'localhost')
+REDIS_PORT = os.getenv('CVAT_REDIS_PORT', 6379)
+REDIS_PASSWORD = urllib.parse.quote(os.getenv('CVAT_REDIS_PASSWORD', ''))
```
SpecLad, line 46: The password itself shouldn't be quoted; the quoting is only needed for embedding it into a URL.

cvat/settings/base.py
```diff
@@ -40,6 +41,10 @@
 ALLOWED_HOSTS = os.environ.get('ALLOWED_HOSTS', 'localhost,127.0.0.1').split(',')
 INTERNAL_IPS = ['127.0.0.1']
 
+REDIS_HOST = os.getenv('CVAT_REDIS_HOST', 'localhost')
```
SpecLad, line 44: Please use lowercase; uppercase makes it look like a setting, which it isn't.

cvat/settings/testing_rest.py
```diff
@@ -24,3 +24,11 @@
 HEALTH_CHECK = {
     'DISK_USAGE_MAX': 100,  # percent
 }
+
+# We don't really need to keep the cache for tests for a long time,
+# as this may cause some tests to fail due to the presence of data
+# in the cache when it is not expected.
+# We used to reset the cache every time we reset the data fixtures.
```
SpecLad, line 31: Can we do an equivalent of that now?

The timeout-based approach seems iffy, since the behavior is different from the production configuration (and I don't think we can guarantee that it would take more than one second to start the next test).

docker-compose.yml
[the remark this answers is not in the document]
SpecLad: Yeah, okay.

helm-chart/Chart.yaml
SpecLad: Documentation or not, I don't think we should rely on a know dead project, especially for a key dependency like this. What are we going to do if we need to update the KeyDB version?

The alternative is to add custom templates for the KeyDB resources.

azhavoro: How often do we change the version of postgresql or redis dependencies? We can use the repo until we really need to change something. Also, I definitely vote for not including keydb chart in CVAT repository and just forking the repo to cvat organization. Forking and changing a single line in Chart.yaml would take a bit of time and could be done anytime it is really needed. At the moment I don't see any forcing factors to do this.

docker-compose.yml
```diff
@@ -18,8 +18,17 @@ services:
...
     container_name: cvat_redis
-    image: redis:7.0-alpine
+    image: eqalpha/keydb:x86_64_v6.3.2
     restart: always
+    command: [
+      "keydb-server",
+      "/etc/keydb/keydb.conf",
+      "--storage-provider", "flash", "/data/flash",
+      "--maxmemory", "5G",
+      "--maxmemory-policy", "allkeys-lfu",
```
SpecLad, line 28: We're using the same KeyDB instance for both asynchronous tasks and caching. For cached objects, this policy makes sense, but wouldn't it cause problems for task queues? We could start randomly losing background tasks if their DB entries get evicted.

azhavoro: https://docs.keydb.dev/docs/flash:

> With KeyDB on FLASH you can set maxmemory to the amount of DRAM you want allocated without the fear of losing > data when exceeded. Instead the data is removed from DRAM but remains in FLASH where it is not lost and can still be > served.

SpecLad: Ah, alright. That's reassuring.

SpecLad (approved):

PR #6798, opened 2023-09-04
Removed separate loggers for each task and project

nmanovic:
We have a memory leak in the logging system. The patch doesn't have the aim to redesign the logging system. Only to fix the memory leak and redirect all messages into cvat_server.log file. In the future it is better to think which error messages are really helpful and send them to ClickHouse instead of a local file. It will allow to avoid any problems with multi processing. Also we have huge amount of logs inside ClickHouse. I don't see any reason to have some of them on the file system.

### DIFF
--- a/cvat/apps/engine/log.py
+++ b/cvat/apps/engine/log.py
@@ -5,40 +5,10 @@
 import logging
 import sys
 import os.path as osp
-from typing import Dict
 from contextlib import contextmanager
-
-from attr import define, field
 from django.conf import settings
 
-from cvat.settings.base import LOGGING
-from .models import Job, Task, Project, CloudStorage
-
-def _get_project(pid):
-    try:
-        return Project.objects.get(pk=pid)
-    except Exception:
-        raise Exception('{} key must be a project identifier'.format(pid))
-
-def _get_task(tid):
-    try:
-        return Task.objects.get(pk=tid)
-    except Exception:
-        raise Exception('{} key must be a task identifier'.format(tid))
-
-def _get_job(jid):
-    try:
-        return Job.objects.select_related("segment__task").get(id=jid)
-    except Exception:
-        raise Exception('{} key must be a job identifier'.format(jid))
-
-def _get_storage(storage_id):
-    try:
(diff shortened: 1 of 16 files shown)


cvat/apps/engine/log.py
```diff
@@ -49,108 +19,8 @@ def get_logger(logger_name, log_file):
...
-
-@define(slots=False)
-class ServerLogManager(_AggregateLogManager):
-    project = field(factory=ProjectLoggerStorage)
-    task = field(factory=TaskLoggerStorage)
-    job = field(factory=JobLoggerStorage)
-    cloud_storage = field(factory=CloudSourceLoggerStorage)
-    glob = field(factory=lambda: logging.getLogger('cvat.server'))
-
-slogger = ServerLogManager()
```
SpecLad, line 140: IMO, something like `slogger` would still be useful to ensure consistent formatting for common context data and to shorten repetitive code. I think something like this would be a good replacement:

```python
class _LoggerAdapterMapping:
    def __init__(self, logger: logging.Logger, object_type: str) -> None:
        self._logger = logger
        self._object_type = object_type

    def __getitem__(self, id_: int) -> logging.LoggerAdapter:
        return logging.LoggerAdapter(self._logger, {"cvat_context": f"{self._object_type}.id={id_}"})

class ServerLogManager:
    def __init__(self, logger_name: str) -> None:
        self.glob = logging.getLogger(logger_name)
        self.project = _LoggerAdapterMapping(self._logger, "Project")
        # ditto for task, job, etc...
```

In the individual files you would then just replace

```python
from cvat.apps.engine.log import slogger
```

with

```
from cvat.apps.engine.log import ServerLogManager

```

and everything else would work as before.

You would also need a custom formatter class:

```python
class CvatContextFormatter(logging.Formatter):
    def format(self, record: logging.LogRecord) -> str:
        s = super().format(record)

        if context := getattr(record, "cvat_context", None):
            s = f'[{context}] {s}'
        
        return s
```

and to update the `LOGGING` configuration key to use it.

nmanovic: I will close the PR and open the new one.

PR #6800, opened 2023-09-04
Do not trigger hotkeys when modals are opened

### LINKED ISSUE #6788: Bug: When in Review Mode, Writing and Issue and Auto Save
So I am on a Job, in review mode, Selected an area to write an issue. 

If while I am writing the issue the auto save is activated, it navigates to a random window...

bsekachev:
### Motivation and context
Resolved #6788 + better UX because now you can use focus and enter/space/tab on modal windows 

### Checklist

### License

### DIFF
--- a/cvat-ui/src/utils/mousetrap-react.tsx
+++ b/cvat-ui/src/utils/mousetrap-react.tsx
@@ -52,6 +52,26 @@ export default function GlobalHotKeys(props: Props): JSX.Element {
     return children || <></>;
 }
 
+Mousetrap.prototype.stopCallback = function (e: KeyboardEvent, element: Element): boolean {
+    // if the element has the class "mousetrap" then no need to stop
+    if ((` ${(element as HTMLElement).className} `).indexOf(' mousetrap ') > -1) {
+        return false;
+    }
+
+    // stop when modals are opened
+    const someModalsOpened = Array.from(
+        window.document.getElementsByClassName('ant-modal'),
+    ).some((el) => (el as HTMLElement).style.display !== 'none');
+    if (someModalsOpened) {
+        return true;
+    }
+
+    // stop for input, select, and textarea
+    return element.tagName === 'INPUT' ||
+        element.tagName === 'SELECT' ||
+        element.tagName === 'TEXTAREA';
+};
+
 export function getApplicationKeyMap(): KeyMap {
     return {
         ...applicationKeyMap,


cvat-ui/src/utils/mousetrap-react.tsx
```diff
@@ -52,6 +52,26 @@ export default function GlobalHotKeys(props: Props): JSX.Element {
     return children || <></>;
 }
 
+Mousetrap.prototype.stopCallback = function (e: KeyboardEvent, element: Element, combo: string): boolean {
+    // if the element has the class "mousetrap" then no need to stop
+    if ((` ${(element as HTMLElement).className} `).indexOf(' mousetrap ') > -1) {
+        return false;
+    }
```
klakhov, line 59: What elements have "mousetrap" class?
Do we really need those spaces?
```suggestion
    // if the element has the class "mousetrap" then no need to stop
    if ((`${(element as HTMLElement).className}`).indexOf(' mousetrap ') > -1) {
        return false;
    }
```

klakhov (commented):
I think we need to add closing settings modal when opening a shortcut model. Now I can open two windows with `f1`/`f2` and it looks strange 
settings-shortcuts

cvat-ui/src/utils/mousetrap-react.tsx
bsekachev: @klakhov 

This is a part of default implementation (can be found here: https://craig.is/killing/mice)
According to it, elements having "mousetrap" class trigger shortcuts handlers (but we do not actually use such elements).
Spaces are necessary to fix case when "mousetrap" class is in the beginning, or at the end.

bsekachev: Removed, since not used

bsekachev:
> I think we need to add closing settings modal when opening a shortcut model. Now I can open two windows with f1/f2 and it looks strange

Now user not able to open both settings and shortcuts.

klakhov:
I think we need to move settings modals a bit higher, now Im getting scrollbar on FullHD screen
settings-shortcuts2

bsekachev:
@klakhov 

Did not you scale the browser? For me on full hd it is allright. 
Or maybe you have system scale 125%?

bsekachev:
@klakhov 

BTW in your previous comment it was fine :)

bsekachev:
@klakhov 

Moved scrolling into the table. Now it should work better even for smaller screens:

image

cvat-ui/src/components/header/styles.scss
```diff
@@ -94,3 +94,10 @@
 .cvat-modal-organization-selector {
     width: 100%;
 }
+
+.cvat-shortcuts-modal-window-table {
+    .ant-table {
+        max-height: $grid-unit-size * 70;
+        overflow-y: auto;
+    }
+}
```
klakhov, line 103:
```suggestion
}

```

klakhov (approved):

PR #6801, opened 2023-09-04
Reset source/target storage if related cloud storage has been deleted

Marishka17:
### Motivation and context

Resolved https://github.com/opencv/cvat/issues/6781
### How has this been tested?

Manually
### Checklist

### License


nmanovic:
@Marishka17 , Why we have cloud_storage_id here https://github.com/opencv/cvat/blob/8c8cf70b1d392ad6b8ec9288fe4423cf85762647/cvat/apps/engine/models.py#L1151? Why don't we have a Foreign Key?

Marishka17:
@azhavoro or @SpecLad, Could you please take a look at the PR?

cvat/apps/engine/serializers.py
```diff
@@ -1980,63 +1982,55 @@ class Meta:
...
     for i in storages:
-        storage_conf = validated_data.get(i)
-        if storage_conf:
-            cloud_storage_id = storage_conf.get('cloud_storage_id')
-            if cloud_storage_id:
-                _validate_existence_of_cloud_storage(cloud_storage_id)
+        if storage_conf := validated_data.get(i):
+            if cloud_storage_id := storage_conf.get('cloud_storage_id'):
+                try:
+                    _ = models.CloudStorage.objects.get(id=cloud_storage_id)
```
SpecLad, line 2026: How about using `.filter(id=cloud_storage_id).exists()` to avoid fetching the actual object?

cvat/apps/engine/location.py
```diff
@@ -14,31 +14,25 @@ class StorageType(str, Enum):
...
-            location_conf['location'] = Location.LOCAL
-        else:
-            location_conf['location'] = storage.location
-            sid = storage.cloud_storage_id
-            if sid:
-                location_conf['storage_id'] = sid
+        location_conf.update({
+            'location': storage.location if storage is not None else Location.LOCAL,
+            **({'storage_id': storage.cloud_storage.id} if getattr(storage, 'cloud_storage', None) else {})
+        })
```
SpecLad, line 31: This seems like a downgrade compared to the previous version of the code. It's now much harder to understand what's happening (and I don't think the logic is actually any different?).

Also, you should use `storage.cloud_storage_id` instead of `storage.cloud_storage.id`, as the latter will fetch the cloud storage record from the database, even through you don't need it here.

cvat/apps/engine/serializers.py
```diff
@@ -1030,6 +1030,8 @@ def _create_files(self, instance, files):
                 )
 
 class StorageSerializer(serializers.ModelSerializer):
+    cloud_storage_id = serializers.IntegerField(required=False, allow_null=True)
+
     class Meta:
         model = models.Storage
         fields = ('id', 'location', 'cloud_storage_id')
```
SpecLad, line 1037: `cloud_storage_id` should probably be removed from here, since it's not a field anymore.

SpecLad (changes requested):

cvat/apps/engine/serializers.py
Marishka17: I don't think so.
https://www.django-rest-framework.org/api-guide/serializers/#specifying-fields-explicitly
```
AssertionError: The field 'cloud_storage_id' was declared on serializer StorageSerializer, but has not been included in the 'fields' option.
```

cvat/apps/engine/migrations/0076_remove_storage_cloud_storage_id_and_more.py
```diff
@@ -0,0 +1,27 @@
...
+        migrations.AddField(
+            model_name="storage",
+            name="cloud_storage",
+            field=models.ForeignKey(
+                null=True,
+                on_delete=django.db.models.deletion.CASCADE,
+                related_name="+",
+                to="engine.cloudstorage",
+            ),
+        ),
```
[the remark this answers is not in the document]
Marishka17: Oh, sorry, I was in a hurry and didn't think about that... Thank you

cvat/apps/engine/migrations/0076_remove_storage_cloud_storage_id_and_more.py
```diff
@@ -0,0 +1,44 @@
...
+    db_storages_to_update = []
+
+    for db_storage in db_storages:
+        if not CloudStorage.objects.filter(pk=db_storage.cloud_storage_id).exists():
+            db_storage.cloud_storage_id = None
+            db_storage.location = Location.LOCAL
+            db_storages_to_update.append(db_storage)
+
+    if db_storages_to_update:
+        Storage.objects.bulk_update(db_storages_to_update, fields=['cloud_storage_id', 'location'], batch_size=500)
```
SpecLad, line 20: A couple questions about this:

1. I notice that here you're updating the invalid `Storage` objects by resetting the `location` and `cloud_storage_id` fields. But that's not what happens with any future `Storage` objects whose corresponding `CloudStorage` is deleted - since you used `on_delete=CASCADE`, such `Storage` objects will be deleted entirely. To maintain consistency, wouldn't it make sense to delete the `Storage` objects here, as well?

2. I think it should be possible to do this deletion entirely within the database. Something like this should work: `Storage.objects.filter(Exists(CloudStorage.objects.filter(pk=OuterRef('cloud_storage_id')))).delete()`. This should be simpler and more efficient. You could also replace `delete` with `update` if you disagree with item 1.

cvat/apps/engine/location.py
Marishka17: Okay, I rolled back the previous version if you like it better.

cvat/apps/engine/migrations/0076_remove_storage_cloud_storage_id_and_more.py
Marishka17: 1. Yes, you are right
2. Thanks

cvat/apps/engine/serializers.py
```diff
@@ -1980,63 +1982,64 @@ class Meta:
...
-        new_location = new_conf.get('location')
-        storage_instance.location = new_location or storage_instance.location
-        storage_instance.cloud_storage_id = new_conf.get('cloud_storage_id', \
-            storage_instance.cloud_storage_id if not new_location else None)
+        if not cloud_storage_id and new_location == models.Location.CLOUD_STORAGE:
+            raise serializers.ValidationError('Cloud storage was selected as location but its id was not specified')
+        elif cloud_storage_id and new_location and new_location != models.Location.CLOUD_STORAGE:
+            raise serializers.ValidationError(f"It is not allowed to specify '{new_location}' location together with cloud storage id")
+        elif cloud_storage_id and not new_location and getattr(storage_instance, 'location', None) != models.Location.CLOUD_STORAGE:
+            raise serializers.ValidationError(f"The configuration of {storage_type} is not full")
```
SpecLad, line 2007: This could use some refactoring to remove duplicate conditions, e.g.:

```py
if cloud storage_id:
    ...
else:
    ...
```

cvat/apps/engine/serializers.py
```diff
@@ -1980,63 +1982,64 @@ class Meta:
...
+        new_conf = validated_data.pop(storage_type, None)
 
         if not new_conf:
             continue
 
+        cloud_storage = None
         cloud_storage_id = new_conf.get('cloud_storage_id')
-        if cloud_storage_id:
-            _validate_existence_of_cloud_storage(cloud_storage_id)
+        new_location = new_conf.get('location')
```
SpecLad, line 1997: If this is called `new_location`, then the other variable ought to be called `new_cloud_storage_id`. Otherwise it's a bit unclear what's old and what's new.

cvat/apps/engine/serializers.py
```diff
@@ -1980,63 +1982,64 @@ class Meta:
...
+        elif cloud_storage_id and new_location and new_location != models.Location.CLOUD_STORAGE:
+            raise serializers.ValidationError(f"It is not allowed to specify '{new_location}' location together with cloud storage id")
+        elif cloud_storage_id and not new_location and getattr(storage_instance, 'location', None) != models.Location.CLOUD_STORAGE:
+            raise serializers.ValidationError(f"The configuration of {storage_type} is not full")
 
-        cloud_storage_id = storage_instance.cloud_storage_id
         if cloud_storage_id:
             try:
-                _ = models.CloudStorage.objects.get(id=cloud_storage_id)
+                cloud_storage = models.CloudStorage.objects.get(id=cloud_storage_id)
```
SpecLad, line 2011: I don't think this variable is actually needed, since you can use `cloud_storage_id` when updating the `Storage` object (and you already do that when you create a new `Storage` object, so might as well be consistent).

cvat/apps/engine/serializers.py
```diff
@@ -1980,63 +1982,64 @@ class Meta:
...
-            storage_instance.save()
-            setattr(instance, storage, storage_instance)
-            continue
+        storage_instance = getattr(instance, storage_type)
 
-        new_location = new_conf.get('location')
-        storage_instance.location = new_location or storage_instance.location
-        storage_instance.cloud_storage_id = new_conf.get('cloud_storage_id', \
-            storage_instance.cloud_storage_id if not new_location else None)
+        if not cloud_storage_id and new_location == models.Location.CLOUD_STORAGE:
```
SpecLad, line 2002: What about the case where `cloud_storage_id` is `None`, current location is `CLOUD_STORAGE`, and `new_location` is unspecified? Seems like that should be prohibited too.

SpecLad (approved):

PR #6802, opened 2023-09-04
Remove the DJANGO_CONFIGURATION environment variable

SpecLad:
### Motivation and context

There are a few issues with it:

* It doesn't work for the Enterprise version, because its settings module is not located under `cvat.settings`.

* The name `DJANGO_CONFIGURATION` suggests that it's defined by Django, but it isn't. It's a CVAT-specific variable. There's also a 3rd-party module that uses this variable (django-configurations), but with different semantics, which could cause confusion.

* It's just redundant to have two variables that effectively do the same thing (`DJANGO_CONFIGURATION` and `DJANGO_SETTINGS_MODULE`).

Besides its runtime effect, `DJANGO_CONFIGURATION` is also used in the Dockerfile to select the packages to install. I could see that still being useful, so replace it with a build-only variable named `CVAT_CONFIGURATION`.

### How has this been tested?

I built the Docker image with various values of `CVAT_CONFIGURATION`.

### Checklist

### License

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -73,11 +73,11 @@ COPY utils/dataset_manifest/requirements.txt /tmp/utils/dataset_manifest/require
 # Exclude av from the requirements file
 RUN sed -i '/^av==/d' /tmp/utils/dataset_manifest/requirements.txt
 
-ARG DJANGO_CONFIGURATION="production"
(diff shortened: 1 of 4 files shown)


nmanovic (approved):
LGTM

PR #6804, opened 2023-09-05
Removed separate loggers for each task and project

nmanovic:
We have a memory leak in the logging system. The patch doesn't have the aim to redesign the logging system. Only to fix the memory leak and redirect all messages into cvat_server.log file. In the future it is better to think which error messages are really helpful and send them to ClickHouse instead of a local file. It will allow to avoid any problems with multi processing. Also we have huge amount of logs inside ClickHouse. I don't see any reason to have some of them on the file system.


nmanovic:
@SpecLad , do you have any other comments?

SpecLad (approved):

PR #6753, opened 2023-08-25
Fixed fail for masks & skeletons with centered text

bsekachev:
### Motivation and context
Resolved #6743

### Checklist

- [x] I submit my changes into the `develop` branch
- [x] I have added a description of my changes into the [CHANGELOG](https://github.com/opencv/cvat/blob/develop/CHANGELOG.md) file
- [ ] I have updated the documentation accordingly
- [ ] I have added tests to cover my changes
- [ ] I have linked related issues (see [GitHub docs](
  https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword))
- [x] I have increased versions of npm packages if it is necessary
  ([cvat-canvas](https://github.com/opencv/cvat/tree/develop/cvat-canvas#versioning),
  [cvat-core](https://github.com/opencv/cvat/tree/develop/cvat-core#versioning),
  [cvat-data](https://github.com/opencv/cvat/tree/develop/cvat-data#versioning) and
  [cvat-ui](https://github.com/opencv/cvat/tree/develop/cvat-ui#versioning))

### License

- [x] I submit _my code changes_ under the same [MIT License](
  https://github.com/opencv/cvat/blob/develop/LICENSE) that covers the project.
  Feel free to contact the maintainers if that's a concern.

Issue #6743, opened 2023-08-24
Activating a mask object with centered-text alignment causes error


bsekachev:
### My actions before raising this issue
- [x] Read/searched [the docs](https://github.com/cvat-ai/cvat/tree/master#documentation)
- [x] Searched [past issues](/issues)

### Steps to Reproduce (for bugs)

1. Enable centered aligned text
![image](https://github.com/opencv/cvat/assets/40690378/d22fcf0d-d512-4d91-b215-710f3fb382ff)
2. Hover a mask
![image](https://github.com/opencv/cvat/assets/40690378/e9030b25-7604-4777-8162-9fa5b017f98b)

PR #6828, opened 2023-09-07
Add test for mouseover mask/skeleton

klakhov:
### Motivation and context

Test for #6743, #6753

### Checklist

### License


klakhov:
@bsekachev 
Could you look on this?

tests/cypress/support/const.js
```diff
@@ -33,6 +33,17 @@ export const multiAttrParams = {
...
+    name: 'skeleton',
+    type: 'shape',
+    points: [
+        { x: 0.55, y: 0.15 },
+        { x: 0.20, y: 0.35 },
+        { x: 0.43, y: 0.55 },
+        { x: 0.63, y: 0.38 },
+        { x: 0.27, y: 0.15 },
+    ],
+};
```
bsekachev, line 46: I do not think we want to add skeleton to each test, it increases running time and quite useless in the most of cases

bsekachev: And if you would like to improve the tests here, I believe creating a task "by default' should be headless, it is much faster and this is good cypress pattern. 
Of course we still need some tests to create tasks via interface, but they should be a different tests that checks exactly tasks creating pipeline.

tests/cypress/support/commands_masks.js
```diff
@@ -0,0 +1,65 @@
+// Copyright (C) 2023 CVAT.ai Corporation
+//
+// SPDX-License-Identifier: MIT
+
+Cypress.Commands.add('drawMask', (instructions) => {
```
bsekachev, line 5: You created a dedicated commands file for masks, I do not mind, but we need some generalized approach here.
Why do not we have in this case dedicated files for skeletons, rectangles, etc?

tests/cypress/e2e/actions_tasks2/case_111_settings_text_size_position_label_content.js
```diff
@@ -10,6 +10,7 @@ import {
     labelName,
     attrName,
     textDefaultValue,
+    skeleton,
```
bsekachev, line 13: I already wrote the suggestion to create in headless by default, but it is optional. 
The minimum changes would be to create a task in headless mode within this test

tests/cypress/support/commands_masks.js
klakhov: I've noticed that our main `commands.js` file gets way too big with adding such functions. What do you think about moving shapes commands in separate `commands_shapes.js`?

bsekachev: Need to think better and probably it is a thing for a separate PR. Because now we have spaghetti here, for example command to draw a 3D cuboid shape is located in ``commands_canvas3d``

klakhov: Okay, lets keep that improvement for separate pr

tests/cypress/support/const.js
klakhov: Removed that. 
Ive spent a couple of hours trying to headless-create task with .zip client files, but had no luck(strangely uploading .png files works well). 
Lets keep everythin as it is here for now, and keep that improvement as separate PR as well

bsekachev (approved):

PR #6901, opened 2023-09-26
Inviting unregistered users by email

klakhov:
### Checklist

### License


cvat/apps/organizations/serializers.py
```diff
@@ -127,3 +146,35 @@ class Meta:
         model = Membership
         fields = ['id', 'user', 'organization', 'is_active', 'joined_date', 'role']
         read_only_fields = ['user', 'organization', 'is_active', 'joined_date']
+
+class AcceptInvitationSerializer(RegisterSerializer):
+    def validate_username(self, username):
+        return username
```
azhavoro, line 152: Why this metod is needed?

cvat/apps/organizations/serializers.py
```diff
@@ -127,3 +146,35 @@ class Meta:
...
+
+    def validate_email(self, email):
+        return email
```
azhavoro, line 155: the same comment as above

cvat/apps/organizations/views.py
```diff
@@ -205,11 +209,39 @@ def perform_create(self, serializer):
         serializer.save(
             owner=self.request.user,
             key=get_random_string(length=64),
-            organization=self.request.iam_context['organization']
+            organization=self.request.iam_context['organization'],
+            request=self.request
```
azhavoro, line 213:
```suggestion
            request=self.request,
```

cvat/apps/organizations/serializers.py
```diff
@@ -127,3 +146,35 @@ class Meta:
...
+            'username': self.validated_data.get('username', ''),
+            'password1': self.validated_data.get('password1', ''),
+            'firstname': self.validated_data.get('firstname', ''),
+            'lastname': self.validated_data.get('lastname', ''),
+        }
+
+    def save(self, request, invitation):
+        self.cleaned_data = self.get_cleaned_data()
+        user = invitation.membership.user
+        if "password1" in self.cleaned_data:
```
azhavoro, line 168: It seems this condition is always true

cvat/apps/organizations/models.py
```diff
@@ -53,18 +58,43 @@ class Meta:
...
+        expiration_date = self.sent_date + timedelta(
+            days=settings.ORG_INVITATION_EXPIRY,
+        )
+        return expiration_date <= timezone.now()
+
+    @property
+    def organization_slug(self):
+        return self.membership.organization.slug
+
+    def send(self, request):
```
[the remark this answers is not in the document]
klakhov: Currently we get 500 error from server if thats a case. Should we handle it explicitly?

cvat/apps/organizations/serializers.py
klakhov: We need to override `validate_email` because `RegisterSerializer` expects it to be unique, but we already have create user with such email

cvat-core/src/api-implementation.ts
```diff
@@ -106,6 +106,28 @@ export default function implementAPI(cvat) {
...
+            firstName,
+            lastName,
+            email,
+            password,
+            userConfirmations,
+            key,
+        );
+
+        return orgSlug;
+    };
```
bsekachev, line 129: Not sure that ``server`` is good namespace here.
What do you think about ``cvat.organizations.acceptInvitation``?

cvat-core/src/organization.ts
```diff
@@ -234,24 +319,9 @@ Object.defineProperties(Organization.prototype.members, {
...
-                                membership.invitation = null;
-                            });
-                    }
-
-                    return Promise.resolve();
-                }),
-            );
+            result.results = result.results.map((rawMembership: SerializedMembershipData) => new Membership(
+                rawMembership,
+            ));
```
bsekachev, line 324: Instead of redefining ``result.results`` better to create a new identificator. 

const memberships = .... 
memberships.count = result.count;
return memberships;

cvat/apps/organizations/serializers.py
azhavoro: It's clear

cvat/schema.yml
```diff
@@ -1531,6 +1531,64 @@ paths:
       responses:
         '204':
           description: The invitation has been deleted
+  /api/invitations/{key}/accept:
+    post:
+      operationId: invitations_create_accept
```
SpecLad, line 1536: Why is this called `invitations_create_accept`? Doesn't seem like it would create an invitation.

cvat/schema.yml
```diff
@@ -1531,6 +1531,64 @@ paths:
...
+          type: string
+        description: A unique value identifying this invitation.
+        required: true
+      tags:
+      - invitations
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/InvitationWriteRequest'
```
SpecLad, line 1550: Why does this accept a request body? You shouldn't need to supply anything to accept an invitation (except the identifier of the invitation, which is already passed via the URL path).

cvat/schema.yml
```diff
@@ -1531,6 +1531,64 @@ paths:
...
+        required: true
+      security:
+      - {}
+      responses:
+        '200':
+          content:
+            application/vnd.cvat+json:
+              schema:
+                $ref: '#/components/schemas/InvitationWrite'
```
SpecLad, line 1559: I'm not sure why this needs to return a response body, but even if it does, shouldn't it be `InvitationRead` rather than `InvitationWrite`?

cvat/schema.yml
```diff
@@ -1531,6 +1531,64 @@ paths:
...
+          description: ''
+  /api/invitations/{key}/resend:
```
SpecLad, line 1561: Same comments here as for `/accept`.

cvat/schema.yml
```diff
@@ -7963,10 +8062,9 @@ components:
           - $ref: '#/components/schemas/RoleEnum'
           readOnly: true
         invitation:
-          type: string
-          readOnly: true
-          nullable: true
+          $ref: '#/components/schemas/BasicInvitation'
```
SpecLad, line 8065: This change looks backwards-incompatible. Is there a way to make it in a backwards-compatible way?

bsekachev:
Is it possible to redirect to login or default tasks page if invitation already accepted? 
Because now I can open this form many times: 
image

After a regular registration user gets authenticated immediately after registration. Here it suggest me to login after. Why is it different? 

Probably it is true, but I am asking just in case. If a user was invited by email, after registration the user's email gets verified without additional actions? 


Maybe we should discuss with somebody, but I believe light blue on orange button is not really contrast:
image
It is unpleasant to look at the button.

cvat-ui/src/components/accept-invitation-page/accept-invitation-page.tsx
```diff
@@ -0,0 +1,50 @@
...
+        email: queryParams.get('email') || '',
+        key: queryParams.get('key') || '',
+    };
+    const onRegister = useCallback((registerData: RegisterData) => {
+        dispatch(acceptInvitationAsync(
+            registerData,
+            invitationParams.key,
+            (orgSlug: string) => {
+                history.replace(`/auth/login?next=/tasks&activateOrganization=${orgSlug}`);
+            },
```
bsekachev, line 35: I do not think approach with activate command in query is good enough. Why do we need this? Why can't we activate organization in this callback? Why can't we add org slug to local storage and reload the page? Even not sure that we need to reload, we can look at current org in org watcher where now query parameter is tracked in onmount hook

cvat-ui/src/components/register-page/register-form.tsx
```diff
@@ -34,6 +34,8 @@ export interface RegisterData {
 interface Props {
     fetching: boolean;
     userAgreements: UserAgreement[];
+    predifinedEmail?: string;
```
bsekachev, line 37:
```suggestion
    predefinedEmail?: string;
```

cvat-ui/src/components/register-page/register-form.tsx
```diff
@@ -34,6 +34,8 @@ export interface RegisterData {
...
+    disableNavigation?: boolean;
```
bsekachev, line 38: Sounds too abstract
I would propose 

```suggestion
    hideLoginLink?: boolean;
```

or something more specific up to you

cvat/apps/organizations/models.py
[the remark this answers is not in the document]
klakhov: Now we will have explicit error:
image

cvat-core/src/user.ts
```diff
@@ -3,7 +3,7 @@
 //
 // SPDX-License-Identifier: MIT
 
-interface RawUserData {
+export interface RawUserData {
```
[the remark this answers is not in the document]
klakhov: Done

cvat-ui/src/components/organization-page/members-list.tsx
```diff
@@ -56,6 +60,21 @@ function MembersList(props: Props): JSX.Element {
...
+                            }}
+                            onDeleteInvitation={(key: string) => {
+                                dispatch(
+                                    removeOrganizationMemberAsync(organizationInstance, member, () => {
+                                        fetchMembers();
+                                    }),
+                                );
+                                dispatch(
+                                    deleteOrganizationInvitationAsync(organizationInstance, key),
+                                );
```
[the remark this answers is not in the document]
klakhov: Yes, its guaranteed. We can delete only org member. Removed second request

cvat-ui/src/components/accept-invitation-page/accept-invitation-page.tsx
klakhov: I've checked it, we can set `currentOrganization` and it will be automatically activated without reload, just on login

klakhov:
@bsekachev 
> Is it possible to redirect to login or default tasks page if invitation already accepted? Because now I can open this form many times: 
As we discussed on HLD meeting, We shouldn't do initial request to check invitation status, because after form is filled invitation can become invalid(exired, deleted etc...). But it would be nice feature for the future. 

> After a regular registration user gets authenticated immediately after registration. Here it suggest me to login after. Why is it different?
Maybe we should add it. But currently we consider this form as changing user account info (Like change password, as user account is already created. In password changing logic we redirect user to login form after successful change.) 

> Probably it is true, but I am asking just in case. If a user was invited by email, after registration the user's email gets verified without additional actions?
Yes, thats true

> Maybe we should discuss with somebody, but I believe light blue on orange button is not really contrast: image It is unpleasant to look at the button.
Changed it to white:
image

bsekachev:
@klakhov 

> As we discussed on HLD meeting, We shouldn't do initial request to check invitation status, because after form is filled invitation can become invalid(exired, deleted etc...). But it would be nice feature for the future.

If discussed on HLD, ok. 
But IMHO user should now if invitation was expired before filling the form, otherwise UX is worse.

bsekachev (approved):

cvat/schema.yml
klakhov: Fixed

cvat/schema.yml
klakhov: Fixed

cvat/schema.yml
klakhov: As we discussed on HLD meeting, we should accept invitation and update user data in one request. Doing this operation in two or three requests can lead to additional problems wich we will have to handle. (invitation expired, deleted, user nickname is already user etc.)

cvat/schema.yml
klakhov: There is a way - to request each invitation individually on memberships page, but this leads to doing 10-15 requests, wich is not really good. 
BTW I dont think this field was used in any way, because all the previous invitations were auto accepted.

azhavoro (approved):

cvat/schema.yml
SpecLad: I was thinking more along the lines of leaving this field as it is, and adding a new field with the `BasicInvitation` type.

> BTW I dont think this field was used in any way, because all the previous invitations were auto accepted.

Even if it wasn't used by the UI, it was still exposed. In particular, the SDK will attempt to parse all fields in a response, even if you don't use some of them, so this change will break it.

SpecLad (changes requested):

cvat/schema.yml
```diff
@@ -1531,6 +1531,60 @@ paths:
...
+      tags:
+      - invitations
+      security:
+      - sessionAuth: []
+        csrfAuth: []
+        tokenAuth: []
+      - signatureAuth: []
+      - basicAuth: []
+      responses:
+        '200':
```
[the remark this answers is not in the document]
klakhov: Done

cvat/schema.yml
```diff
@@ -6100,6 +6154,35 @@ components:
       - description
       - name
       - version
+    AcceptInvitationRequest:
```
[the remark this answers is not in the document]
klakhov: There was a mistake, this serializer doesnt accept `email` field, fixed that. Also, it has different save logic (it gets user email from the invitation, doesn't create user as its already created), so I dont think we can use `RegisterSerializerEx` here

cvat/schema.yml
```diff
@@ -1531,6 +1531,60 @@ paths:
...
+              $ref: '#/components/schemas/AcceptInvitationRequest'
+        required: true
+      security:
+      - {}
+      responses:
+        '200':
+          content:
+            application/vnd.cvat+json:
+              schema:
+                type: string
```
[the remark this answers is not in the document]
klakhov: Done

cvat/schema.yml
klakhov: Ok. I've reverted that change.
Personally, I dont like adding new field like `invitation_data` with invitation details, having a couple of invitation fields looks inconvenient. So my suggestion will be keep it as it was for now.

SpecLad (approved):

nmanovic:
Please let me play a littile bit with the PR. Probably, I have some questions.

PR #6918, opened 2023-09-28
Remove unnecessary COUNT for labels

nmanovic:
Optimized /api/projects, /api/tasks, and /api/jobs requests. Previously for unification, we decided to add a count of labels into every response for each project, task, and job. It leads to performance issues. UI and probably other clients don't use the information because it is useless. Removing it, helps to optimize the SQL request several times.


nmanovic:
Before:
image


After:
image

cvat/apps/engine/views.py
```diff
@@ -242,9 +242,6 @@ class ProjectViewSet(viewsets.GenericViewSet, mixins.ListModelMixin,
     ).prefetch_related(
         'tasks', 'label_set__sublabels__attributespec_set',
         'label_set__attributespec_set'
```
SpecLad, line 244: What about these label-related prefetches, are they needed?

tests/python/shared/fixtures/data.py
```diff
@@ -58,18 +60,63 @@ def tasks():
...
+    for label in labels:
+        project_id = label.get("project_id")
+        if not label["parent_id"] and project_id:
+            projects[project_id]["labels"]["count"] += 1
+
+    return projects
+
+
+@pytest.fixture(scope="session")
+def jobs():  # jobs with labels count
```
SpecLad, line 105: Misplaced comment?

tests/python/rest_api/test_tasks.py
```diff
@@ -1568,7 +1573,8 @@ def test_can_rename_label(self, tasks, labels, admin_user):
         resulting_labels = self._get_task_labels(task["id"], admin_user)
         assert DeepDiff(resulting_labels, task_labels, ignore_order=True) == {}
 
-    def test_cannot_rename_label_to_duplicate_name(self, tasks, labels, admin_user):
+    def test_cannot_rename_label_to_duplicate_name(self, tasks_wlc, labels, admin_user):
+        tasks = tasks_wlc
```
SpecLad, line 1577: IMO, renamings like these serve to obscure the underlying data structure. If a variable is supposed to have label counts, then let it keep the `_wlc` suffix.

SpecLad (changes requested):

changelog.d/20231011_162748_nikita_fix_perf_issues.md
```diff
@@ -0,0 +1,7 @@
+### Changed
+
+- /api/projects, /api/tasks, /api/jobs don't return information about
+  the count of labels anymore. The information is not useful and
+  significantly complicates SQL queries.
```
SpecLad, line 5: It would be useful to mention how the users can get this information now.

SpecLad (approved):

PR #6929, opened 2023-09-29
Avoid relying on PyAV-provided video frame count

SpecLad:
### Motivation and context

They are not reliable.

In particular, MP4 has a feature called "edit lists" that allows you to set a custom playback order for the media data. With edit lists, you could only specify that a particular range of frames should be played, or that a range should be played multiple times, etc. See the following for technical details:

https://developer.apple.com/documentation/quicktime-file-format/edit_list_atom

FFmpeg follows edit lists when decoding videos. However, the frame count returned by PyAV's `Stream.frames` property is the number of frames in the raw media data and does not reflect the modifications applied by an edit list.

When we build a video manifest, we use `Stream.frames` if it's non-zero. Therefore, in the presence of an edit list we will obtain a frame count that does not match the actual number of frames that we can get out of the video.

FWIW, edit lists are probably not the only way that `Stream.frames` could be inaccurate, it's just the reason behind a specific problem I encountered.

Since we already have to handle the situation where `Stream.frames` is not available, just pretend it doesn't exist and always count frames by traversing the entire video. I don't think it even matters much, since we have to do it anyway to build the rest of the manifest.

We also have to stop validating the frame count in a user-provided manifest, which is unfortunate, but it doesn't seem worthwhile to decode the entire video just for that.

### How has this been tested?

I checked that `dataset_manifest/create.py` now calculates the correct number of frames for a file with an edit list. I also tested the same file by uploading it to CVAT.

### Checklist

### License

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- TDB
+- Incorrectly determined video frame count when the video contains an MP4 edit list
+  (<https://github.com/opencv/cvat/pull/6929>)
 
 ### Security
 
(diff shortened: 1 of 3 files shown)


nmanovic (approved):
LGTM

PR #6952, opened 2023-10-06
Fix possible double writing of the uploaded client files in Upload-Multiple requests

### LINKED ISSUE #6878: Images uploaded via the SDK are twice as large as those uploaded via the web interface

1. Select some images to be uploaded via web interface and also SDK
2. Using the web interface, create a task with those images
3. Using the SDK, create a task and upload those images
4. Compare the size of the images inside the docker volume

### Current Behaviour
I am creating tasks and uploading images programmatically, but I realized that a project I expected to be 20 GB in size was finally 40 GB. Initially, I thought it might be because of parameters like zip_chunks or cache, but it's not.

I selected three images of 77.6MB, 56.9MB, and 42.8MB, created a CVAT task from the web interface, and uploaded these images. I have done the same process but with the SDK, and as a result, I can see that the images uploaded with the SDK occupy twice as much space in the docker volume.

This is the configuration used in the web interface:
image

This is the configuration used in SDK: 
```
data_params = {"image_quality": 70,
                "use_zip_chunks": True,
                "use_cache": True}
```

This is the comparison of size between task created in web interface (253) and task created using the SDK (255): 
image

Is it a bug or am I missing a parameter?

### Your Environment
- CVAT VERSION: 2.7.0
- CVAT SDK: 2.7.0
- Docker version `docker version` (e.g. Docker 17.0.05): 20.10.25
- Operating System and version (e.g. Linux, Windows, MacOS): Amazon Linux

zhiltsov-max:
Fixes #6878 

 In the case of big files (>2.5 MB by default), the uploaded files could be write-appended twice,
leading to bigger raw file sizes than expected. This PR fixes the behavior by excluding repetitive
writes where it was not supposed.

- Fixed double append-writing of the uploaded files when Upload-Multiple requests are used
- Fixed potential DB - disk inconsistencies in the case of upload errors
- Added tests

### Checklist

### License


tests/python/rest_api/test_tasks.py
```diff
@@ -762,6 +763,53 @@ def test_can_create_task_with_exif_rotated_images(self):
...
+
+        task_id, _ = create_task(self._USERNAME, task_spec, task_data)
+
+        # check that the uploaded unprocessed files have the original size
+        script = textwrap.dedent(
+            """\
+            import os
+            from cvat.apps.engine.models import Data
+            data = Data.objects.get(tasks__id={task_id})
+            print(os.stat(f"data/data/{{data.id}}/raw/{image_file_name}").st_size)
```
SpecLad, line 802: Wouldn't it be possible to check this by requesting the original chunk via the API and examining it? Seems like it would be more robust than peeking inside the container.

zhiltsov-max: It can work, I need to check.

zhiltsov-max: Updated

jcorrochanoj:
Would it be possible to have this fix in the next release? Thank you very much!

changelog.d/20231006_152644_fix_6878.md
```diff
@@ -0,0 +1,4 @@
+Fixed
```
SpecLad, line 1:
```suggestion
### Fixed
```

tests/python/rest_api/test_tasks.py
```diff
@@ -820,6 +820,53 @@ def test_can_create_task_with_exif_rotated_images(self):
...
+            "client_files": [image_file],
+            "image_quality": 70,
+            "use_cache": False,
+            "use_zip_chunks": True,
+        }
+
+        task_id, _ = create_task(self._USERNAME, task_spec, task_data)
+
+        # check that the original chunk image have the original size
+        # this is less accurate than checking the uploaded file directly, but faster
```
SpecLad, line 855: Could you clarify why this is less accurate?

zhiltsov-max: Because we're not checking the original problem exactly, as it was with raw uploaded files. The chunks may get processed data, which can be correct for some foreign reasons, which can shadow the initial problem.

SpecLad (approved):

PR #6958, opened 2023-10-06
Update nested objects on parent labels update

### LINKED ISSUE #6871: exported categories different from project categories

1. create a project with 1 category - eg. car
2. add a task with a single image to the project
3. annotate a car in the image
4. export the task (I'm using coco format)
5. confirm the categories in the export annotation json match the project(car)
6. add another category to the project - eg. person
8. export the task again
9. unexpectedly there is only the car in the export task annotation json file.

### Expected Behaviour
When exporting a task in coco format the categories in the annotation json should match those of the project, not just those present in the project when the task was last saved.

### Current Behaviour
When exporting a task in coco format, the categories present in the annotation json are not in line with those in the project.

zhiltsov-max:
### Motivation and context

Fixes #6871 

- Added batch update (note - no signals issued on this) for owning and nested objects (tasks, jobs) on parent (task, project) labels updates

### Checklist

### License

### DIFF
--- a/cvat-sdk/cvat_sdk/core/client.py
+++ b/cvat-sdk/cvat_sdk/core/client.py
@@ -55,6 +55,9 @@ class Config:
     """Directory in which to store cached server data"""
 
 
+_VERSION_OBJ = pv.Version(VERSION)
+
+
 class Client:
     """
     Provides session management, implements authentication operations
@@ -62,8 +65,8 @@ class Client:
     """
 
     SUPPORTED_SERVER_VERSIONS = (
-        pv.Version("2.7"),
-        pv.Version("2.8"),
+        pv.Version(f"{_VERSION_OBJ.epoch}!{_VERSION_OBJ.major}.{_VERSION_OBJ.minor}"),
+        pv.Version(f"{_VERSION_OBJ.epoch}!{_VERSION_OBJ.major}.{_VERSION_OBJ.minor+1}"),
     )
 
     def __init__(


cvat/apps/engine/serializers.py
```diff
@@ -1196,12 +1196,29 @@ def update(self, instance, validated_data):
...
+            self.update_children_objects_on_labels_update(instance)
+
         instance.task_labels_count = instance.label_set.filter(
             parent__isnull=True).count()
         instance.proj_labels_count = instance.project.label_set.filter(
             parent__isnull=True).count() if instance.project else 0
         return instance
 
+    def update_children_objects_on_labels_update(self, instance: models.Task):
+        # Use light wrappers and load only necessary values to avoid heavy requests
```
SpecLad, line 1210: I'm not sure it's necessary to load anything here. Wouldn't it be possible to do the update entirely within the database, using `QuerySet.update`?

cvat/apps/engine/serializers.py
```diff
@@ -1196,12 +1196,29 @@ def update(self, instance, validated_data):
...
+        updated_jobs = [
+            models.Job(**v)
+            for v in models.Job.objects.filter(
+                segment__task=instance
+            ).values("id", "updated_date")
```
SpecLad, line 1215: Wouldn't `only` be more appropriate here (since you load the values into the model class anyway)?

cvat/apps/engine/serializers.py
```diff
@@ -1196,12 +1196,29 @@ def update(self, instance, validated_data):
...
+        ]
+        for job in updated_jobs:
+            job.updated_date = instance.updated_date
```
SpecLad, line 1218: I think you should filter out jobs whose `updated_date` is already greater or equal than the new value, to avoid having the date go backwards.

cvat/apps/engine/serializers.py
```diff
@@ -1196,12 +1196,29 @@ def update(self, instance, validated_data):
...
+        if 'label_set' in validated_data and not instance.project_id:
+            self.update_children_objects_on_labels_update(instance)
+
         instance.task_labels_count = instance.label_set.filter(
             parent__isnull=True).count()
         instance.proj_labels_count = instance.project.label_set.filter(
             parent__isnull=True).count() if instance.project else 0
         return instance
 
+    def update_children_objects_on_labels_update(self, instance: models.Task):
```
SpecLad, line 1209:
```suggestion
    def update_children_on_labels_update(self, instance: models.Task):
```

`child_objects` would also be okay.

tests/python/rest_api/test_labels.py
```diff
@@ -819,3 +820,99 @@ def test_regular_user_delete_org_label(self, user_org_case):
...
+    @pytest.mark.parametrize("update_kind", ["addition", "removal", "modification"])
+    def test_project_label_update_triggers_nested_task_and_job_update(
+        self, update_kind, admin_user, labels, projects, tasks, jobs
+    ):
+        # Checks for regressions against the issue https://github.com/opencv/cvat/issues/6871
+
+        project = next(p for p in projects if p["tasks"]["count"] and p["labels"]["count"])
+        project_labels = [l for l in labels if l.get("project_id") == project["id"]]
+        nested_tasks = [t for t in tasks if t["project_id"] == project["id"]]
+        nested_task_ids = set(t["id"] for t in nested_tasks)
```
SpecLad, line 836:
```suggestion
        nested_task_ids = {t["id"] for t in nested_tasks}
```

tests/python/rest_api/test_labels.py
```diff
@@ -13,10 +13,11 @@
 from cvat_sdk import exceptions, models
 from cvat_sdk.api_client.api_client import ApiClient, Endpoint
 from cvat_sdk.core.helpers import get_paginated_collection
+from dateutil.parser import parse as parse_datetime
```
SpecLad, line 16: Should probably use `isoparse`, to make sure the API outputs are in the expected format.

SpecLad (changes requested):

SpecLad:
Please add a changelog entry.

cvat/apps/engine/serializers.py
zhiltsov-max: Yeah, seems to be so.

cvat/apps/engine/serializers.py
zhiltsov-max: I added this, but such situation is strange. I can probably imagine it happening in 2 parallel updates in 2 threads.

tests/python/rest_api/test_labels.py
zhiltsov-max: I don't think it adds anything to the test. These tests are not focused on a specific output format.

cvat/apps/engine/serializers.py
```diff
@@ -1204,8 +1204,17 @@ def update(self, instance, validated_data):
...
         instance.save()
+
+        if 'label_set' in validated_data and not instance.project_id:
+            self.update_child_objects_on_labels_update(instance)
+
         return instance
 
+    def update_child_objects_on_labels_update(self, instance: models.Task):
+        models.Job.objects.filter(
+            updated_date__lt=instance.updated_date
```
SpecLad, line 1215: I think you forgot the filter by `task_id`.

tests/python/rest_api/test_labels.py
SpecLad: Sure, but if they happen to catch an issue with the date output format, that would be useful, no? And it should be a trivial change.

cvat/apps/engine/serializers.py
zhiltsov-max: Yes, thanks. Fixed

tests/python/rest_api/test_labels.py
zhiltsov-max: Ok

SpecLad (approved):

PR #6961, opened 2023-10-07
Support rotation in detector results

WilliamKyle:
### Motivation and context

Fixes opencv/cvat#6891

### Checklist

### License

### DIFF
--- a/cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/tools-control.tsx
+++ b/cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/tools-control.tsx
@@ -1239,7 +1239,7 @@ export class ToolsControlComponent extends React.PureComponent<Props, State> {
                                     objectType: ObjectType.SHAPE,
                                     frame,
                                     occluded: false,
-                                    rotation: data.rotation == null ? 0 : parseFloat(data.rotation),
+                                    rotation: data.rotation || 0,
                                     source: core.enums.Source.AUTO,
                                     attributes: (data.attributes as { name: string, value: string }[])
                                         .reduce((acc, attr) => {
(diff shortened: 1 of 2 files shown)


cvat/apps/lambda_manager/views.py
```diff
@@ -621,6 +621,9 @@ def reset(self):
                         "source": "auto"
                     }
 
+                    if anno.get("rotation"):
+                        shape["rotation"] = anno.get("rotation")
```
bsekachev, line 625: ``rotation`` is only supported actually for ellipses and bounding boxes. 
So, we need additional check for shape type here I believe, otherwise it can lead to unexpected results later. 

```python
if shape["type"] in ("rectangle", "ellipse"):
    shape["rotation"] = anno.get("rotation", 0)
```

cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/tools-control.tsx
```diff
@@ -1239,6 +1239,7 @@ export class ToolsControlComponent extends React.PureComponent<Props, State> {
                                     objectType: ObjectType.SHAPE,
                                     frame,
                                     occluded: false,
+                                    rotation: data.rotation == null ? 0 : parseFloat(data.rotation),
```
bsekachev, line 1242: The same comment as for the server, plus why do you use parseFloat here? Does server return string? 
Also ``parseFloat`` for ``undefined`` returns NaN, it will cause issues with models that does not have ``rotation`` in output.  

Here I would suggest: ```rotation: data.rotation || 0``` in case of corresponding shapes.

cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/tools-control.tsx
```diff
@@ -1239,6 +1239,7 @@ export class ToolsControlComponent extends React.PureComponent<Props, State> {
                                     objectType: ObjectType.SHAPE,
                                     frame,
                                     occluded: false,
+                                    rotation: data.rotation || 0,
```
bsekachev, line 1242:
```suggestion
                                    rotation: [
                                        ShapeType.RECTANGLE, ShapeType.ELLIPSE,
                                    ].includes(data.type) ? (data.rotation || 0) : 0,
```

bsekachev (approved):
Works well for me, thank you for the contribution

PR #6968, opened 2023-10-09
Fix tracks splitting for tracks that have sub-tracks

kirill-sizov:
### Motivation and context

When uploading annotations for the whole task that has few jobs our server splits tracks. Each track we divide into a few tracks in that way that each track corresponds to only one job and all shapes of the track do not extend beyond the frame range corresponding to that job.

It mainly works, but we have a problem with skeleton tracks. Due to this problem we if upload skeleton tracks for the whole task all our tracks will be duplicated for each job but not split, it means that we will have tracks that extend beyond the frame range of jobs and it leads to bad side effects, in this case, our exported annotations from such task could be incorrect or we won't be able to export annotations at all.

And the reason for this problem is this small bug, which this PR fixes. Our code really slices sub-tracks for the skeleton track, but after this slicing server doesn't save the result.

### Checklist

### License


nmanovic (approved):
LGTM

PR #6969, opened 2023-10-09
Allow to upload tracks from COCO formats

kirill-sizov:
### Motivation and context

This PR resolves the following problems:

1. During import we ignore track_id attribute for annotations if these annotations do not come from one of these formats        `('cvat', 'datumaro', 'sly_pointcloud')` this prevents to import tracks in coco format -> this PR extends the list of "track formats"
2. When we convert a dataset from CVAT representation to Datumaro Dataset in all possible situations only the first shape of the skeleton track is a keyframe, all other shapes will be as marked as not keyframes it leads to problems when we try to import such tracks that were exported from CVAT -> this PR adds the following line of code: set `keyframe = True` for skeletons if at least one point of this skeleton is keyframe 
3. During import skeleton tracks our code always expects to see a "keyframe" attribute for the skeleton's element, but not all formats support attributes for skeleton points (for example COCO Keypoints doesn't allow to have attributes for each skeleton's point). And if the server doesn't see this attribute for some element we ignore this element -> this PR adds the following logic for skeleton track: if skeleton's elements don't have a keyframe attribute at all then consider such skeleton elements as keyframes = True

### Checklist

### License


tests/python/rest_api/test_tasks.py
```diff
@@ -2526,3 +2526,45 @@ def test_import_annotations(self, task_kind, annotation_kind, expect_success):
...
 
             assert b"Could not match item id" in capture.value.body
+
+    def test_can_export_and_import_skeleton_tracks_in_coco_format(self):
+        task = self.client.tasks.retrieve(14)
+        dataset_file = self.tmp_dir / "some_file.zip"
+        format_name = "COCO Keypoints 1.0"
+
+        original_annotations = task.get_annotations()
+        print(original_annotations)
```
nmanovic, line 2536: Please remove the print

tests/python/rest_api/test_tasks.py
```diff
@@ -2526,3 +2526,45 @@ def test_import_annotations(self, task_kind, annotation_kind, expect_success):
...
+
+        task.export_dataset(format_name, dataset_file, include_images=False)
+        task.remove_annotations()
+        task.import_annotations(format_name, dataset_file)
+
+        imported_annotations = task.get_annotations()
+        print(imported_annotations)
```
nmanovic, line 2543: Need to remove

kirill-sizov: Removed

nmanovic (approved):

PR #6972, opened 2023-10-10
Add a GitHub workflow to prepare a release pull request

SpecLad:
### Motivation and context

This automates the first half of the current release process. The second half will be implemented by another workflow.

The reason why it can't all be done in a single workflow is that it's useful to let developers inspect what'll go into the release before actually publishing it, and to apply any last-minute fixes, if necessary. It also allows CI to complete for the release PR.

To support the new workflow, add a `--set` option to `update_version.py` that sets the version to a custom value.

### How has this been tested?

Manual testing on my fork of CVAT. See a test run here: <https://github.com/SpecLad/cvat/actions/runs/6459901481>.

### Checklist

### License

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -1,0 +1,59 @@
+name: Prepare release
+on:
+  workflow_dispatch:
+    inputs:
+      newVersion:
+        description: "Version number for the new release"
+        required: true
+        default: X.Y.Z
+jobs:
+  main:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Validate version number
+        env:
+          NEW_VERSION: "${{ inputs.newVersion }}"
+        run: |
+          if ! [[ "$NEW_VERSION" =~ [0-9]+\.[0-9]+\.[0-9]+ ]]; then
+            echo "Invalid version number"
+            exit 1
+          fi
+
+      - uses: actions/checkout@v4
+
+      - name: Verify that the release is new
+        run: |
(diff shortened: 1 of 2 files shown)


SpecLad:
CC @azhavoro

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,52 @@
...
+        run:
+          git push -u origin "release-${{ inputs.newVersion }}"
+
+      - name: Create release pull request
+        env:
+          GH_TOKEN: "${{ github.token }}"
+        run: |
+          gh pr create \
+            --base=master \
+            --title="Release v2.7.4" \
```
kirill-sizov, line 51: I think it supposed to be something like this:
```suggestion
            --title="Release v${{ inputs.newVersion }}" \
```

kirill-sizov:
I also see that it would be nice to have some validation on the version number that prevents increasing one of X, Y or Z on more than 1. Or was it already implemented and I missed it?

.github/workflows/prepare-release.yml
SpecLad: Right, thanks for catching it.

SpecLad:
> I also see that it would be nice to have some validation on the version number that prevents increasing one of X, Y or Z on more than 1.

IMO, this isn't needed. There's no requirement that versions _have_ to be sequential, and we might end up wanting to skip a version for whatever reason.

Although now that you mention it, maybe I should at least add a check that the release does not already exist...

PR #6975, opened 2023-10-10
Add ability to limit one user to one task at a time

Marishka17:
### Motivation and context

Added ability to limit a user to one active rq job at a time.
Migrated to rq 1.15.1.
### How has this been tested?

Manually + REST API test that will be in another PR
### Checklist

### License


cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -3385,6 +3399,15 @@ def _test_api_v2_tasks_id_data_spec(self, user, spec, data,
         response = send_data_callback(task_id, user, data)
         self.assertEqual(response.status_code, status.HTTP_202_ACCEPTED, response.reason_phrase)
 
+        if get_status_callback:
+            response = get_status_callback(task_id, user)
+            while response.data['state'] not in ('Failed', 'Finished'):
+                response = get_status_callback(task_id, user)
```
zhiltsov-max, line 3405: It's a potential infinite loop here, and it can waste  a lot of CPU here. I suggest adding a default timeout and some small delay between the requests (e.g. `sleep(0.01)`).

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -40,7 +40,7 @@
 from cvat.apps.engine.tests.utils import get_paginated_collection
 from utils.dataset_manifest import ImageManifestManager, VideoManifestManager
```
zhiltsov-max, line 41: Could you add at least 1 test for this functionality? It can be done in this file, using the [`self.settings()`](https://docs.djangoproject.com/en/4.2/topics/testing/tools/#django.test.SimpleTestCase.settings) context manager.

cvat/apps/lambda_manager/views.py
```diff
@@ -445,21 +451,36 @@ def enqueue(self,
...
-                "task": task,
-                "job": job,
-                "quality": quality,
-                "cleanup": cleanup,
-                "conv_mask_to_poly": conv_mask_to_poly,
-                "mapping": mapping,
-                "max_distance": max_distance
-            })
+        user_id = request.user.id
+        cm = queue.connection.lock(f'{queue.name}-lock-{user_id}') if settings.LIMIT_ONE_USER_TO_ONE_AUTO_ANNOTATION_TASK_AT_A_TIME else nullcontext()
```
nmanovic, line 455: Why don't lock always? Why do you need the condition here?

nmanovic, line 455: We should acquire the lock with a timeout (e.g. 1h). If we kill the process and the lock is acquired, we should release it after some time. Otherwise, we can get into a situation where the lock is locked forever. What do you think?

Marishka17: I thought it would be better not to use locking in cases when it is not required.

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -3385,6 +3400,20 @@ def _test_api_v2_tasks_id_data_spec(self, user, spec, data,
...
+        if get_status_callback:
+            max_number_of_attempt = 10
+            state = None
+            while state not in ('Failed', 'Finished'):
+                assert max_number_of_attempt, "Too much time to create a task"
+                response = get_status_callback(task_id, user)
+                state = response.data['state']
+                sleep(0.01)
+                max_number_of_attempt -= 1
+            self.assertEqual(response.data['state'], expected_task_creation_status_state)
```
Marishka17, line 3412:
```suggestion
            self.assertEqual(state, expected_task_creation_status_state)
```

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -3400,9 +3401,14 @@ def _test_api_v2_tasks_id_data_spec(self, user, spec, data,
         self.assertEqual(response.status_code, status.HTTP_202_ACCEPTED, response.reason_phrase)
 
         if get_status_callback:
-            response = get_status_callback(task_id, user)
-            while response.data['state'] not in ('Failed', 'Finished'):
+            max_number_of_attempt = 10
```
zhiltsov-max, line 3404: 0.01 * 10 = 0.1 seconds, it's quite a small time period even in some normal situations, e.g. with video processing and chunks on weak machines. I suggest increasing it to 10 seconds instead for stability.

cvat/settings/base.py
```diff
@@ -706,6 +706,5 @@ class CVAT_QUEUES(Enum):
 # to check configuration and throw ImproperlyConfigured if thats a case
 EMAIL_BACKEND = None
 
-LIMIT_ONE_USER_TO_ONE_IMPORT_TASK_AT_A_TIME = strtobool(os.getenv('LIMIT_ONE_USER_TO_ONE_IMPORT_TASK_AT_A_TIME', 'false'))
-LIMIT_ONE_USER_TO_ONE_EXPORT_TASK_AT_A_TIME = strtobool(os.getenv('LIMIT_ONE_USER_TO_ONE_EXPORT_TASK_AT_A_TIME', 'false'))
-LIMIT_ONE_USER_TO_ONE_AUTO_ANNOTATION_TASK_AT_A_TIME = strtobool(os.getenv('LIMIT_ONE_USER_TO_ONE_AUTO_ANNOTATION_TASK_AT_A_TIME', 'false'))
+ONE_RUNNING_JOB_IN_QUEUE_PER_USER = strtobool(os.getenv('ONE_RUNNING_JOB_IN_QUEUE_PER_USER', 'false'))
+KEY_TO_EXCLUDE_FROM_DEPENDENCY = 'exclude_from_dependency'
```
zhiltsov-max, line 710: I don't think it makes sense to make this name configurable. It's just an implementation detail, but it is referenced from several places in the code.

tests/python/rest_api/test_tasks.py
```diff
@@ -2684,3 +2687,46 @@ def test_can_export_and_import_skeleton_tracks_in_coco_format(self):
...
+
+        task_data = {
+            "image_quality": 75,
+            "client_files": generate_image_files(3),
+        }
+
+        task_id, _ = create_task(self._USERNAME, task_spec, task_data)
+        return task_id
+
+    def test_can_create_tasks_in_well_formed_order(self):
```
zhiltsov-max, line 2716: This test doesn't look reliable because of unpredictable execution order, and now we also don't test the case with disabled ordering. Why we can't add tests for both configurations in the server unit tests?

Marishka17: Because server unit tests are not executed on the real instance? Generally, we don't need to test the order, we need to check that 1. some rq jobs depend on others, 2. There are no rq jobs that depend on several rq jobs. And looks like, these checks are predictable. Do you think it would be better if we keep server configuration with the disabled order in the public repo and move that test to private?

zhiltsov-max: > Generally, we don't need to test the order

If so, the test name should probably be changed.

> 1. some rq jobs depend on others, 2. ...

Probably, it's better to check that only 1 job has no deps, and all the others have 1 dep each. Otherwise we can get some unnoticed errors. Actually, the point 2 feels like an implementation detail.

> Do you think it would be better if we keep server configuration with the disabled order in the public repo and move that test to private?

No, I don't think so. If this feature is available in the public version, it should probably be tested in the public version. I'm worried that we currently test only 1 case.

.github/workflows/main.yml
```diff
@@ -174,6 +174,7 @@ jobs:
           COVERAGE_PROCESS_START: ".coveragerc"
         run: |
           pytest tests/python/ --cov --cov-report=json
+          pytest tests/python/rest_api/test_rq_job_dependencies.py --rebuild --adjusted-server-conf --cov --cov-report=json
```
nmanovic, line 177: I expect something like

```
ONE_RUNNING_JOB_IN_QUEUE_PER_USER=true pytest tests/python/rest_api/test_rq_job_dependencies.py --rebuild --cov --cov-report=json
```

Need to check. I hope it will work.

tests/docker-compose.adjusted.yml
```diff
@@ -0,0 +1,9 @@
+services:
+
+  cvat_redis:
```
nmanovic, line 3: I believe the file isn't necessary. You can define the environment variable externally. Redis should be available as redis:9379. Both cvat_redis and cvat_server are in the same network.

tests/python/rest_api/test_rq_job_dependencies.py
```diff
@@ -0,0 +1,58 @@
...
+            "image_quality": 75,
+            "client_files": generate_image_files(3),
+        }
+
+        task_id, _ = create_task(self._USERNAME, task_spec, task_data)
+        return task_id
+
+    def test_check_rq_job_dependencies_after_creating_tasks_in_parallel(self, request):
+        if not request.session.config.getoption("--adjusted-server-conf"):
+            pytest.skip("Server is not configured with ONE_RUNNING_JOB_IN_QUEUE_PER_USER setting")
```
nmanovic, line 42: Let's use https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-mark-skipif

tests/python/shared/fixtures/init.py
```diff
@@ -370,23 +376,27 @@ def session_start(
     rebuild = session.config.getoption("--rebuild")
     cleanup = session.config.getoption("--cleanup")
     dumpdb = session.config.getoption("--dumpdb")
```
nmanovic, line 378: I hope you don't need changes in the file.

cvat/apps/engine/backup.py
```diff
@@ -1081,21 +1087,26 @@ def _import(importer, request, queue, rq_id, Serializer, file_field_name, locati
...
-            meta={
-                'tmp_file': filename,
-                **get_rq_job_meta(request=request, db_obj=None)
-            },
-            depends_on=dependent_job,
-            result_ttl=settings.IMPORT_CACHE_SUCCESS_TTL.total_seconds(),
-            failure_ttl=settings.IMPORT_CACHE_FAILED_TTL.total_seconds()
-        )
+        user_id = request.user.id
+        cm = queue.connection.lock(f'{queue.name}-lock-{user_id}', timeout=30) if settings.ONE_RUNNING_JOB_IN_QUEUE_PER_USER else nullcontext()
```
nmanovic, line 1091: Let's move the code into a function. Something like
def get_rq_lock(request):
    ...

tests/python/rest_api/test_rq_job_dependencies.py
```diff
@@ -0,0 +1,58 @@
...
+        task_data = {
+            "image_quality": 75,
+            "client_files": generate_image_files(3),
+        }
+
+        task_id, _ = create_task(self._USERNAME, task_spec, task_data)
+        return task_id
+
+    def test_check_rq_job_dependencies_after_creating_tasks_in_parallel(self, request):
+        if not request.session.config.getoption("--adjusted-server-conf"):
```
zhiltsov-max, line 41: The parameter name is too generic now to make such decisions as it is done here. Could you try to make it more clear and focused?

tests/docker-compose.adjusted.yml
Marishka17: We run tests on the host, not in cvat server. We should open ports for the redis container to be able to interact with it.

.github/workflows/main.yml
Marishka17: If we still need to override some settings in docker compose file, I'd rather to leave ONE_RUNNING_JOB_IN_QUEUE_PER_USER there

Marishka17:
@nmanovic, @zhiltsov-max, Do you have any other comments?

changelog.d/20231023_132746_maria_limit_one_user_to_one_task_at_a_time.md
```diff
@@ -0,0 +1,15 @@
+### Added
+
+- Ability to limit one user to one task at a time
+  (<https://github.com/opencv/cvat/pull/6975>)
+
+### Fixed
+
+- Bug with viewing dependent rq jobs for downloading resources from
+cloud storage when file path contains sub-directories
```
zhiltsov-max, line 9: Could you make this description more clear for the platform users? It's quite hard to understand what is happening in this message.

cvat/apps/engine/utils.py
```diff
@@ -160,20 +193,33 @@ def configure_dependent_job(
     key: str,
     request: HttpRequest,
     result_ttl: float,
-    failure_ttl: float
+    failure_ttl: float,
+    should_be_dependent: bool = settings.ONE_RUNNING_JOB_IN_QUEUE_PER_USER,
 ) -> Job:
-    rq_job_id_download_file = rq_id + f'?action=download_{filename}'
+    rq_job_id_download_file = rq_id + f'?action=download_{key.replace("/", ".")}'
```
zhiltsov-max, line 199: Could you comment on this `replace` change? Why it needed? It can lead to task name collisions with some filenames with dots (in all invocations key is taken from filename).

Marishka17: How it can lead to task name collisions if rq_id is unique per import operation?
Replace is required because a key can contain a forward slash that is not an acceptable symbol for rq id generally because it leads to the inability to open page with rq job details (django-rq/queues/{id}/{rq_id}/)

zhiltsov-max: So what happens if we have a single task creation request with 2 files on a cloud storage:
```
a/b.ext
a.b.ext
```

Can the jobs collide?

zhiltsov-max (approved):

PR #7014, opened 2023-10-16
Add new community links to the `Links` README section

capjamesg:
This PR adds two links to the CVAT README `Links` section. The first link is a tutorial written by the Roboflow team on how to use CVAT and the second link shows the Roboflow guide to using models for automatic labeling in CVAT. I would love for the Roboflow + CVAT integration to be more accessible to new users; the faster someone can label, the quicker they can get a model into production.


nmanovic:
Thank you for your contribution.

PR #7026, opened 2023-10-18
SDK: determine supported server versions automatically

SpecLad:
### Motivation and context

I think the spirit of the `SUPPORTED_SERVER_VERSIONS` constant is to allow the same and the next minor versions as the SDK, so just derive those versions automatically instead of bumping them every minor release.

### How has this been tested?

I examined `cvat_sdk.Client.SUPPORTED_SERVER_VERSIONS` in the REPL.

### Checklist

- ~~[ ] I have linked related issues~~ (see [GitHub docs](

### License

### DIFF
--- a/cvat-sdk/cvat_sdk/core/client.py
+++ b/cvat-sdk/cvat_sdk/core/client.py
@@ -55,6 +55,9 @@ class Config:
     """Directory in which to store cached server data"""
 
 
+_VERSION_OBJ = pv.Version(VERSION)
+
+
 class Client:
     """
     Provides session management, implements authentication operations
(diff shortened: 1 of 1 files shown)


zhiltsov-max (approved):
> I think the spirit of the SUPPORTED_SERVER_VERSIONS constant is to allow the same and the next minor versions as the SDK

It was never supposed to be like this - the variable is literally the list of supported versions for each SDK release. But we can surely add this change to simplify maintenance of new releases, as it reflects our current strategy.

SpecLad:
> It was never supposed to be like this - the variable is literally the list of supported versions for each SDK release.

Okay, I'm going to remove this part from the commit message.

PR #7048, opened 2023-10-24
Compress `changeFrame` events

klakhov:
### Motivation and context

We have too many event records of `changeFrame` and `zoomImage` events. They are not really informative. This pr compresses `changeFrame` events and increases ignore events timeouts
 
### Checklist

- ~~[] I have updated the documentation accordingly~~

### License


cvat-core/src/logger-storage.ts
```diff
@@ -36,19 +46,37 @@ class LoggerStorage {
...
                         currentPayload.object_id === previousLog.payload.object_id &&
                         currentPayload.id === previousLog.payload.id
                     );
                 },
+                update: defaultUpdate,
+            },
+            [LogType.changeFrame]: {
+                lastLog: null,
+                timeThreshold: 4000,
+                ignore(previousLog: EventLogger, currentPayload: any): boolean {
```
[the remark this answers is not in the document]
klakhov: Curretly implementation compresses many change frame events into one with compressed `from` and `to`
Before:
`changeFrame: { from: 1, to: 2 }`
`changeFrame: { from: 2, to: 3 }`
`changeFrame: { from: 3, to: 4 }`

After:
`changeFrame: { from: 1, to: 4 }`

We can add parameters to that:
`changeFrame: { from: 1, to: 4, step: 1, count: 3 }`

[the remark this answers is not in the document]
nmanovic: Do we have "count" field as a standard field for an event?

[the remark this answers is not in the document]
klakhov: Yes, default "count" field is used, like for other events `draw:object` or `delete:object`

cvat-ui/src/actions/annotation-actions.ts
```diff
@@ -665,6 +665,9 @@ export function changeFrameAsync(
             await job.logger.log(LogType.changeFrame, {
                 from: frame,
                 to: toFrame,
+                // step is passed with sign so we can know the direction
+                step: toFrame - frame,
```
bsekachev, line 669: We always know the direction because we know ``from``, ``to`` isn't it?

klakhov: Thats true. We can add Math.abs here if we dont need sign of the step. From my perspective having sign helps to undersand direction faster while looking on analytics data.

 What do you think about it?

bsekachev: To be honest I just do not see why I would use this information. Usually, when debug, only final frame is important

bsekachev: What would be more useful, IMHO, is event duration

cvat-core/src/logger-storage.ts
```diff
@@ -109,10 +139,11 @@ Object.defineProperties(LoggerStorage.prototype.log, {
...
                 if (lastLog && ignoreRule.ignore(lastLog, payload)) {
-                    lastLog.payload = {
-                        ...lastLog.payload,
-                        ...payload,
-                    };
+                    lastLog.payload = ignoreRule.update(lastLog, payload);
+
+                    if (this.closeOnLog.includes(logType)) {
+                        await lastLog.close();
+                    }
```
bsekachev, line 146: This probably shall be handled out of ``(logType in this.ignoreRules) {`` condition body

bsekachev: Do we have any strong reasons to close log immediately after it was added?
IMHO it does not look obvious, when the log is closed implictitly. Why not to close the log from outside code after it was created?

cvat-ui/src/actions/annotation-actions.ts
```diff
@@ -662,10 +662,13 @@ export function changeFrameAsync(
...
             localStorage.setItem(`Job_${job.id}_frame`, `${toFrame}`);
-            await job.logger.log(LogType.changeFrame, {
+            const log = await job.logger.log(LogType.changeFrame, {
                 from: frame,
                 to: toFrame,
-            });
+                step: toFrame - frame,
+                count: 1,
+            }, true);
+            log.close();
```
bsekachev, line 671: It would be more correct to close this log after canvas got updated with a new image to compute correct duration. 
In case of multiple change frame events it will work almost as expected, but in case with one, duration will be about zero, what is not true

bsekachev (approved):

PR #7125, opened 2023-11-10
Update Datumaro format

zhiltsov-max:
### Motivation and context

Fixes https://github.com/opencv/cvat/issues/5424
Fixes https://github.com/opencv/cvat/issues/7375
Fixes https://github.com/cvat-ai/cvat/issues/8700

Depends on https://github.com/cvat-ai/datumaro/pull/34

This PR improves quality of life when using Datumaro format.

- Added support for direct .json uploading of annotations, similarly to the COCO and CVAT formats
- Added image metadata when exporting in the Datumaro format without images
- For related images in 3d tasks, datumaro export without images will include only the basenames (before: absolute server paths were exported)
- <s>Refactored `conv_mask_to_poly` uses to avoid code and logic duplication</s> (will be in another PR)

### How has this been tested?

Unit tests

### Checklist

### License

## Summary by CodeRabbit

- **New Features**
	- Introduced support for direct .json file import in Datumaro format.
- **Bug Fixes**
	- Fixed an issue where exporting without images in Datumaro format now correctly includes image information.
- **Refactor**
	- Renamed classes and methods in dataset management to support a broader range of media types, enhancing the system's flexibility.
- **Tests**
	- Added new tests to verify the behavior of importing and exporting in Datumaro format, ensuring robustness in dataset handling.


cvat/apps/dataset_manager/formats/datumaro.py
```diff
@@ -15,29 +14,23 @@
...
+@importer(name="Datumaro", ext="JSON, ZIP", version="1.0")
 def _import(src_file, temp_dir, instance_data, load_data_callback=None, **kwargs):
-    Archive(src_file.name).extractall(temp_dir)
-
-    dataset = Dataset.import_from(temp_dir, 'datumaro', env=dm_env)
+    if zipfile.is_zipfile(src_file):
+        zipfile.ZipFile(src_file).extractall(temp_dir)
+        dataset = Dataset.import_from(temp_dir, 'datumaro', env=dm_env)
+    else:
+        load_data_callback = None
```
Marishka17, line 32: As I can see `load_data_callback` is used only for importing project datasets so `zipfile.is_zipfile(src_file)` will always be true and it is not clear to me why we need to set the `load_data_callback` explicitly to `None`. Am I missing something?

tests/python/rest_api/test_tasks.py
```diff
@@ -2733,3 +2733,49 @@ def test_can_export_and_import_skeleton_tracks_in_coco_format(self):
...
+            )
+            assert response.status == HTTPStatus.OK
+
+        with zipfile.ZipFile(io.BytesIO(response.data)) as zip_file:
+            annotations = json.loads(zip_file.read("annotations/default.json"))
+
+        assert annotations["items"]
+        for item in annotations["items"]:
+            assert item["media"]["path"]
+            assert item["image"]["path"]
```
Marishka17, line 2758: What is the difference between a media path and an image path in Datumaro format?

tests/python/rest_api/test_tasks.py
```diff
@@ -2733,3 +2733,49 @@ def test_can_export_and_import_skeleton_tracks_in_coco_format(self):
...
+            assert item["image"]["size"]
+
+    def test_can_import_datumaro_json(self, admin_user, tasks):
```
Marishka17, line 2761: Wouldn't it be better to place the test in `TestImportTaskAnnotations`?

tests/python/rest_api/test_tasks.py
```diff
@@ -2733,3 +2733,49 @@ def test_can_export_and_import_skeleton_tracks_in_coco_format(self):
...
+        with zipfile.ZipFile(io.BytesIO(response.data)) as zip_file:
+            annotations = zip_file.read("annotations/default.json")
+
+        with Client(BASE_URL) as client, TemporaryDirectory() as tempdir:
+            client.config.status_check_period = 0.01
+            client.login((admin_user, USER_PASS))
+
+            annotations_path = Path(tempdir) / "annotations.json"
+            annotations_path.write_bytes(annotations)
+            client.tasks.retrieve(task["id"]).import_annotations("Datumaro 1.0", annotations_path)
```
Marishka17, line 2781: Probably you need to check annotations after import and compare them with the original

tests/python/rest_api/test_tasks.py
```diff
@@ -2733,3 +2733,49 @@ def test_can_export_and_import_skeleton_tracks_in_coco_format(self):
                 for tes in te.shapes
             ]
         )
+
+
+@pytest.mark.usefixtures("restore_db_per_function")
+class TestGetTaskAnnotations:
+    @pytest.mark.parametrize("mode", ["annotation", "interpolation"])
+    def test_datumaro_export_without_annotations_includes_image_info(self, admin_user, tasks, mode):
```
Marishka17, line 2741: I guess it will be better to move the test in `TestGetTaskDataset`

cvat/apps/dataset_manager/formats/datumaro.py
```diff
@@ -15,29 +14,23 @@
...
     dataset = Dataset.from_extractors(GetCVATDataExtractor(
         instance_data=instance_data, include_images=save_images), env=dm_env)
-    if not save_images:
-        dataset.transform(DeleteImagePath)
     dataset.export(temp_dir, 'datumaro', save_images=save_images)
 
     make_zip_archive(temp_dir, dst_file)
 
-@importer(name="Datumaro", ext="ZIP", version="1.0")
+@importer(name="Datumaro", ext="JSON, ZIP", version="1.0")
```
Marishka17, line 26: It looks like `Datumaro 3D` can also be extended to support importing annotations in JSON format

cvat/apps/engine/view_utils.py
```diff
@@ -111,3 +112,14 @@ def decorator(f):
         return f
 
     return decorator
+
+def parse_mask_to_poly_request_param(request) -> Optional[bool]:
+    """
+    Retrieves and parses the 'conv_mask_to_poly' parameter from the request.
+    """
+    from cvat.apps.dataset_manager.formats.transformations import MaskToPolygonTransformation
+    param_value = request.query_params.get(MaskToPolygonTransformation.PARAMETER_NAME, None)
```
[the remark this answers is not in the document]
zhiltsov-max: The parsing and default value logic was copied in 5 different places, and it doesn't belong to any of them. The updated version allows to control the input value (including none) in the format handler. If we supply a default value here, the logic will be duplicated once again.

cvat/apps/dataset_manager/formats/datumaro.py
zhiltsov-max: It's true, but with this implementation there is no need to think about how the function is called.

zhiltsov-max: It's true, but with this implementation there is no need to think about how the function is called.

tests/python/rest_api/test_tasks.py
zhiltsov-max: The meaning is the same, but each possible media type is saved in a different field. If the type info is provided, the image field is used. If no media type info is available, the generic media type (and field) is used.

cvat/apps/engine/view_utils.py
```diff
@@ -4,6 +4,7 @@
 
 # NOTE: importing in the utils.py header leads to circular importing
 
+from distutils.util import strtobool
```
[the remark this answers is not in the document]
zhiltsov-max: I agree in general, but this function is used in several places in the project, so I'd better leave it to another PR.

tests/python/rest_api/test_tasks.py
zhiltsov-max: Probably, the reasons for both fields to exist simultaneously are historical. Media path is just newer, and the image path is duplicated for backward compatibility.

cvat/apps/engine/view_utils.py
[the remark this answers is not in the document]
Marishka17: > The parsing and default value logic was copied in 5 different places, and it doesn't belong to any of them. The updated version allows to control the input value (including none) in the format handler.

Of course it is clear, but I still don't understand why we cannot use True as the default value in 1 place here because in all subsequent functions to which this parameter is passed (I mean import_task_annotations/import_job_annotations/import_dataset_as_project), it is not expected that it will be None at least because it has the default value `True` in [_import_annotations](https://github.com/opencv/cvat/blob/5cad9356682a74af90c0248f34d266f1457934f3/cvat/apps/engine/views.py#L2821),  [_import_project_dataset](https://github.com/opencv/cvat/blob/5cad9356682a74af90c0248f34d266f1457934f3/cvat/apps/engine/views.py#L3048).
I prefer to use `True` as default value here and in this case, you don't need to handle `None` in `MaskToPolygonTransformation.convert_dataset` or migrate to optional `conv_mask_to_poly` with `None` in all places. But mixing the 2 approaches looks odd to me.

cvat/apps/engine/view_utils.py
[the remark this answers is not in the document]
Marishka17: My point was not to migrate to the custom `strtobool` in all places in this PR, but just don’t add any more dependencies from it.

cvat/apps/dataset_manager/formats/datumaro.py
Marishka17: Maybe in this case we need to raise an exception if `load_data_callback` was passed with json file. (since it is unexpected behavior)

cvat/apps/engine/view_utils.py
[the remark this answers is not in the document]
zhiltsov-max: Both `_import_annotations` and `_import_project_dataset` also duplicate the logic they do not own, so they probably need a fix then.

Actually, I noticed that some formats did not use the parameter and could benefit from it (e.g. COCO, Datumaro, ...). But the default value would change the default behavior for these formats, and there was no way to actually get the original input value in the format handler. So I decided only to refactor what we have now.

cvat/apps/engine/views.py
```diff
@@ -420,7 +421,7 @@ def upload_finished(self, request):
         if self.action == 'dataset':
             format_name = request.query_params.get("format", "")
             filename = request.query_params.get("filename", "")
-            conv_mask_to_poly = strtobool(request.query_params.get('conv_mask_to_poly', 'True'))
+            conv_mask_to_poly = parse_mask_to_poly_request_param(request)
```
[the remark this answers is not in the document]
zhiltsov-max: Will be in another PR to keep this one focused.

nmanovic:
@coderabbitai review

Eldies (approved):

PR #7127, opened 2023-11-10
helm-chart: make backend templates more uniform

SpecLad:
### Motivation and context

The templates for the various backend service deployments differ in many ways, but a lot of those differences can be eliminated. That way, it becomes easier to compare different templates, to create new services, and to make changes across all templates.

Specific differences which are eliminated are as follows:

* Different sets of values are used for customization. This can be mitigated by putting the specific set of values in a local variable once. That way, only the variable assignment needs to differ.

* Containers and volumes are named differently. There's no need for this, since these names are local to a pod. Just name all analogous containers/volumes the same thing.

* Some deployments use different values for the `app` label. This just seems incorrect, since they all belong to the same app. Standardize on `cvat-app` as the value.

* Some deployments forget to check disableDistinctCachePerService. That seems like a bug.

* There are minor formatting differences.

Ideally I'd like to reduce these templates to just calling one big shared template with service-specific parameters, but there are still enough differences between the templates that I don't feel like doing that just yet.

### How has this been tested?

`helm template`

### Checklist

### License

### DIFF
--- a/changelog.d/20231113_174833_roman_refactor_values.md
+++ b/changelog.d/20231113_174833_roman_refactor_values.md
@@ -1,0 +1,5 @@
+### Changed
+
+- \[Helm\] All backend-related deployments now use `cvat-app` as the value
+  for the `app` label
+  (<https://github.com/opencv/cvat/pull/7127>)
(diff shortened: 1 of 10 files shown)


SpecLad:
I think I'm going to expand this, actually, hang on.

azhavoro (approved):

PR #7132, opened 2023-11-13
[Helm] update ingress templates

azhavoro:
Added:
- Generate ingress based on parameter `ingress: true/false`
- Ability to configure cvat hostname using an override value file instead of the ugly --set option in helm
- Removed some deprecated ingress specifications

### Checklist

### License


helm-chart/templates/analytics/ingress.yaml
```diff
@@ -1,48 +1,38 @@
 {{- $version := .Capabilities.KubeVersion.GitVersion -}}
```
SpecLad, line 1: I think this assignment can be removed.

helm-chart/templates/analytics/ingress.yaml
```diff
@@ -1,48 +1,38 @@
...
 {{- if and .Values.ingress.enabled .Values.analytics.enabled -}}
 {{- $fullName := include "cvat.fullname" . -}}
-{{- if semverCompare ">=1.19-0" $version -}}
```
SpecLad, line 4: You should add a [`kubeVersion`](https://helm.sh/docs/topics/charts/#the-chartyaml-file) field to `Chart.yml` to declare supported Kubernetes versions. This should also be reflected in the changelog.

helm-chart/templates/analytics/ingress.yaml
```diff
@@ -1,48 +1,38 @@
...
-    {{- with .Values.analytics.ingress.annotations }}
+    {{- end }}
+    {{- with .Values.ingress.annotations }}
       {{- toYaml . | nindent 4 }}
     {{- end }}
 spec:
-{{- if semverCompare ">=1.19-0" $version }}
-  ingressClassName: {{ .Values.ingress.className }}
+{{- if .Values.ingress.ingressClassName }}
+  ingressClassName: {{ .Values.ingress.ingressClassName | quote }}
```
SpecLad, line 21:
```suggestion
{{- with .Values.ingress.className }}
  className: {{ toYaml . }}
```

Note: the change from `ingress.className` to `ingress.ingressClassName` seems gratuitous, so I'm suggesting to revert it.

helm-chart/templates/analytics/ingress.yaml
```diff
@@ -1,48 +1,38 @@
...
 {{- end }}
   rules:
-  {{- range .Values.ingress.hosts }}
-    - host: {{ .host | quote }}
+    {{- if .Values.ingress.hostname }}
```
SpecLad, line 24: This seems odd. If `ingress.hostname` is not defined, then the created ingress will have no rules - but then what's the point of creating the ingress in the first place? Seems like `ingress.hostname` should be required if `ingress.enabled` is true.

helm-chart/templates/ingress.yaml
```diff
@@ -1,26 +1,20 @@
 {{- $version := .Capabilities.KubeVersion.GitVersion -}}
```
SpecLad, line 1: This can also be removed.

helm-chart/templates/ingress.yaml
```diff
@@ -1,26 +1,20 @@
...
   namespace: {{ .Release.Namespace }}
   labels:
     {{- include "cvat.labels" . | nindent 4 }}
-  {{- with .Values.ingress.annotations }}
   annotations:
-    {{- toYaml . | nindent 4 }}
-  {{- end }}
+    {{- with .Values.ingress.annotations }}
+      {{- toYaml . | nindent 4 }}
+    {{- end }}
```
SpecLad, line 14: This seems like a change for the worse. Won't this break if `ingress.annotations` is empty?

helm-chart/values.yaml
```diff
@@ -364,56 +355,30 @@ grafana:
...
-#          service:
-#            name: frontend-service
-#            port: 80
-#  tls:
-#    - hosts:
-#      - <your_domain>
-#    secretName: ingress-tls-cvat
+  ## @param ingress.pathType Ingress path type
+  ##
+  pathType: Prefix
```
SpecLad, line 363: Why do we need this to be configurable? We know that our paths are prefixes, no other option would make sense.

helm-chart/templates/ingress.yaml
```diff
@@ -33,26 +27,29 @@ spec:
...
-            backend:
-              serviceName: {{ .service.name }}
-              servicePort: {{ .service.port }}
-            {{- end }}
+  {{- if .Values.ingress.hostname }}
+  - host: {{ .Values.ingress.hostname }}
+    http:
+      paths:
+      {{- range tuple "api" "admin" "static" "django-rq" "profiler" }}
+      - path: {{ print "/" . }}
```
SpecLad, line 35:
```suggestion
      - path: "/{{ . }}"
```

helm-chart/values.yaml
```diff
@@ -364,56 +355,30 @@ grafana:
       enabled: false
 
 ingress:
+  ## @param ingress.enabled Enable ingress record generation for CVAT
```
SpecLad, line 358:
```suggestion
  ## @param ingress.enabled Enable ingress resource generation for CVAT
```

helm-chart/values.yaml
```diff
@@ -364,56 +355,30 @@ grafana:
...
-#            port: 8080
-#        - path : /
-#          pathType: "Prefix"
-#          service:
-#            name: frontend-service
-#            port: 80
-#  tls:
-#    - hosts:
-#      - <your_domain>
-#    secretName: ingress-tls-cvat
```
SpecLad, line 410: What about these parameters? They should either be documented, or their usages should be removed.

helm-chart/Chart.yaml
```diff
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.10.3
+version: 0.11.0
```
SpecLad, line 18: Please add a changelog entry documenting everything that has changed.

SpecLad (changes requested):

helm-chart/values.yaml
azhavoro: removed from ingress

helm-chart/templates/analytics/ingress.yaml
```diff
@@ -1,48 +1,33 @@
...
-    {{- with .Values.analytics.ingress.annotations }}
+    {{- end }}
+    {{- with .Values.ingress.annotations }}
       {{- toYaml . | nindent 4 }}
     {{- end }}
 spec:
-{{- if semverCompare ">=1.19-0" $version }}
-  ingressClassName: {{ .Values.ingress.className }}
+{{- if .Values.ingress.className }}
+  ingressClassName: {{ .Values.ingress.className | quote }}
```
SpecLad, line 20:
```suggestion
  ingressClassName: {{ .Values.ingress.className | toYaml }}
```

helm-chart/templates/ingress.yaml
```diff
@@ -19,40 +12,27 @@ metadata:
...
-  tls:
-  {{- range .Values.ingress.tls }}
-    - hosts:
-      {{- range .hosts }}
-        - {{ . | quote }}
-      {{- end }}
-      secretName: {{ .secretName }}
-  {{- end }}
+{{- if .Values.ingress.className }}
+  ingressClassName: {{ .Values.ingress.className | quote }}
```
SpecLad, line 16:
```suggestion
  ingressClassName: {{ .Values.ingress.className | toYaml }}
```

helm-chart/values.yaml
```diff
@@ -364,56 +355,27 @@ grafana:
...
+  ## @param ingress.hostname Default host for the ingress record
```
SpecLad, line 361:
```suggestion
  ## @param ingress.hostname Host for the ingress resource
```

(It's not really the default host, it's the only host.)

helm-chart/cvat.values.yaml
```diff
@@ -3,51 +3,6 @@ analytics:
 
 ingress:
   enabled: true
-  annotations:
-    traefik.ingress.kubernetes.io/router.entrypoints: web
-    kubernetes.io/ingress.class: traefik
```
SpecLad, line 8: You removed these annotations from here, but you didn't re-add them anywhere else. Is the user supposed to set them manually when enabling Traefik?

changelog.d/20231115_110934_andrey_update_helm.md
```diff
@@ -0,0 +1,7 @@
+### Changed
+
+- \[Helm\] Minimum compatible Kubernetes version is 1.19.0.
+  (<https://github.com/opencv/cvat/pull/7132>)
+
+- \[Helm\] The CVAT hostname can be configured with `ingress.hostname` option.
```
SpecLad, line 6: This isn't the only change. You've also removed `ingress.hosts` and `ingress.tls`.

helm-chart/values.yaml
SpecLad: Then how would the user configure TLS in a Traefik-less configuration? I think you should provide some replacement for these parameres.

helm-chart/cvat.values.yaml
azhavoro: There is no need to add the first line because this is the default behavior
The second line is deprecated, ingressclass is now used instead

changelog.d/20231115_110934_andrey_update_helm.md
azhavoro: I'm sure that the tls options never worked, but ok I'll add the entry

helm-chart/values.yaml
azhavoro: TLS has never worked before, I guess it's a separate task to add tls support out of the box

azhavoro: Ok, you are right, let's keep the part described in the spec, but anyway the tls configuration is too controller specific

helm-chart/cvat.values.yaml
SpecLad: > There is no need to add the first line because this is the default behavior

Okay.

> The second line is deprecated, ingressclass is now used instead

Indeed, but `ingressClass` defaults to blank. Shouldn't it default to `traefik` when the `traefik` setting is true?

helm-chart/values.yaml
```diff
@@ -364,56 +355,34 @@ grafana:
...
+  ## @param ingress.className IngressClass that will be be used to implement the Ingress (Kubernetes 1.18+)
+  ## This is supported in Kubernetes 1.18+ and required if you have more than one IngressClass marked as the default for your cluster
+  ## ref: https://kubernetes.io/blog/2020/04/02/improvements-to-the-ingress-api-in-kubernetes-1.18/
+  ##
+  className: ""
+  ## @param ingress.tls Enable TLS configuration for the host defined at `ingress.hostname` parameter
+  ## TLS certificates will be retrieved from a TLS secret defined in tlsSecretName parameter
+  ##
+  tls: false
+  ## @param ingress.tlsSecretName Ыpecifies the name of the secret containing TLS certificates. Ignored if ingress.tls is false
```
SpecLad, line 380:
```suggestion
  ## @param ingress.tlsSecretName Specifies the name of the secret containing TLS certificates. Ignored if ingress.tls is false
```

helm-chart/cvat.values.yaml
azhavoro: done

PR #7139, opened 2023-11-14
Fix issues related to rq job dependencies

Marishka17:
### Motivation and context

This PR resolves 2 issues related to rq jobs:
1. In some scenarios it is possible to reach a situation when `X2` rq job depends on `X1` -> running `X1` job is moved to FailedJobRegistry and not deleted -> user creates one more `X1` job that is enqueued after `X2` execution.  It was possible due to the second issue when a user tried to export annotations for task 1, then for task 2, and after restarting the worker container user tried again to export annotations for task 1. (cyclic dependence)
2. Looks like in rq implementation rq jobs that depend on `X` job will never be enqueued when `X` is moved to FailedJobRegistry due to AbandonedJobError. I've submitted the [issue](https://github.com/rq/rq/issues/2006) to the rq repository.
### How has this been tested?

Manually
### Checklist

### License


cvat/settings/base.py
```diff
@@ -723,3 +723,66 @@ class CVAT_QUEUES(Enum):
...
+                    job.cleanup(ttl=-1, pipeline=pipeline)
+                    failed_job_registry.add(job, job.failure_ttl)
+                    queue.enqueue_dependents(job)
+
+            pipeline.zremrangebyscore(self.key, 0, score)
+            pipeline.execute()
+
+    return job_ids
+
+rq.registry.StartedJobRegistry.cleanup = custom_started_job_registry_cleanup
```
azhavoro, line 788: Please add a comment why we need override this method, and also add a link to the relevant issue

cvat/apps/engine/utils.py
```diff
@@ -171,9 +176,24 @@ def define_dependent_job(queue: DjangoRQ, user_id: int, should_be_dependent: boo
         for job in queue.job_class.fetch_many(
             queue.deferred_job_registry.get_job_ids(), queue.connection
         )
-        if job and job.meta.get("user", {}).get("id") == user_id
+        if job and job.meta.get("user", {}).get("id") == user_id and job.is_deferred
```
Marishka17, line 179: We should check each job from `deferred_job_registry` that it is still deferred because `deferred_job_registry.get_job_ids()` calls `deferred_job_registry.cleanup()` which is dummy and has an "empty" implementation

cvat/rq_patching.py
```diff
@@ -0,0 +1,71 @@
...
+
+import rq.registry
+from rq.exceptions import AbandonedJobError, NoSuchJobError
+from rq.job import JobStatus
+from rq.utils import current_timestamp
+
+
+# NOTE: we should patch implementation of original method because
+# there is no enqueuing dependent jobs in original function
+# https://github.com/rq/rq/issues/2006
```
SpecLad, line 17: Since you already have a fix, maybe you could submit it upstream? That way we could potentially remove this hack later (or maybe not apply it in the first place, depending on how quickly the maintainers respond).

cvat/rq_patching.py
```diff
@@ -0,0 +1,72 @@
...
+                    failed_job_registry.add(job, job.failure_ttl)
+                    queue.enqueue_dependents(job)
+
+            pipeline.zremrangebyscore(self.key, 0, score)
+            pipeline.execute()
+
+    return job_ids
+
+def update_started_job_registry_cleanup() -> None:
+    rq.registry.StartedJobRegistry.cleanup = custom_started_job_registry_cleanup
```
SpecLad, line 72: I think you should assert on the version of rq here. That way, we'll at least remember to revise the patch if we update to a newer rq version.

cvat/apps/engine/utils.py
Marishka17: I have no exact explanation for that. Need to debug this more, but looks like it can happen if there are active or deferred jobs when restarting the worker container. I have seen several times that one job with `finished` status exists in 2 registries `DeferredJobRegistry` and `FinishedJobRegistry`. FinishedJobRegistry and  StartedJobRegistry have cleanup implementation and expired jobs will be deleted eventually. However, deferred jobs can be deleted from deferred job registry only directly like [here](https://github.com/rq/rq/blob/d11094af95188c197a5e11b1c4fa05b000485370/rq/queue.py#L1224C43-L1224C43). I'll send you screenshots with evidence.

SpecLad: Okay. Can you summarize this issue in a code comment?

SpecLad (approved):

PR #7167, opened 2023-11-22
Reduce the number of requests for task details

Marishka17:
### Motivation and context

Previously, we tried to get task details every time a user opened the upload/export job annotations modals to get information about task source/target storage. It was not effective and could lead to a large number of *PermissionDenied* exceptions in our analytics. We also requested task details and **all** task jobs every time a ground truth job was requested when opening the annotation view. 
### How has this been tested?

Manually
### Checklist

### License


cvat-core/src/session.ts
```diff
@@ -567,6 +571,22 @@ export class Job extends Session {
...
+                    ),
+                },
+                targetStorage: {
+                    get: () => (
+                        new Storage({
+                            location: data.target_storage?.location || StorageLocation.LOCAL,
+                            cloudStorageId: data.target_storage?.cloud_storage_id,
+                        })
+                    ),
+                },
```
bsekachev, line 589: Do you think we need to create new instance each time on get? 
It leads to strange behavior, like ``Job.targetStorage !== Job.targetStorage``

cvat-core/src/api-implementation.ts
```diff
@@ -183,6 +183,8 @@ export default function implementAPI(cvat) {
             sort: isString,
             search: isString,
             jobID: isInteger,
+            task_id: isInteger,
```
bsekachev, line 186: probably that should be taskID field. 
I could not find it is used anywhere as ``task_id`` before, so, we may try to rename it

cvat-ui/src/components/import-dataset/import-dataset-modal.tsx
```diff
@@ -103,35 +103,23 @@ function ImportDatasetModal(props: StateToProps): JSX.Element {
...
-                        if ((error as any).code !== 403) {
-                            Notification.error({
-                                message: `Could not get task instance ${instance.taskId}`,
-                                description: error.toString(),
-                            });
-                        }
-                    });
-                setInstanceType(`job #${instance.id}`);
+            setDefaultStorageLocation(instance.sourceStorage?.location || StorageLocation.LOCAL);
+            setDefaultStorageCloudId(instance.sourceStorage?.cloudStorageId || null);
```
bsekachev, line 112: As I see ``instance.sourceStorage`` is always defined for all the tasks/projects/jobs. 
So, we do not need this construction with ``?``

cvat-ui/src/components/import-dataset/import-dataset-modal.tsx
```diff
@@ -103,35 +103,23 @@ function ImportDatasetModal(props: StateToProps): JSX.Element {
...
+            let type: 'project' | 'task' | 'job';
+
+            if (isProject()) {
+                type = 'project';
+            } else if (isTask()) {
+                type = 'task';
+            } else {
+                type = 'job';
```
bsekachev, line 120:
```suggestion
            let type: 'project' | 'task' | 'job' = 'job';

            if (isProject()) {
                type = 'project';
            } else if (isTask()) {
                type = 'task';
```

tests/python/shared/assets/jobs.json
```diff
@@ -23,11 +23,13 @@
       "mode": "annotation",
       "organization": 2,
       "project_id": null,
+      "source_storage": null,
```
bsekachev, line 26: Just a reminder to update assets in our private repository

cvat-ui/src/components/import-dataset/import-dataset-modal.tsx
bsekachev: The same for ``instance.targetStorage`` in export-dataset-modal.tsx

Issue #6757, opened 2023-08-28
Fresh cvat installation result in huge django size


Djkmarco:
### My actions before raising this issue
- [x] Read/searched [the docs](https://github.com/cvat-ai/cvat/tree/master#documentation)
- [x] Searched [past issues](/issues)

### Steps to Reproduce (for bugs)

Fresh installation on cvat after removing previous installation, uninstalled docker and all related files. The disk is 4Tb and 75% is free.

### Expected Behaviour
Working CVAT on local machine

### Current Behaviour
Cannot connect to the server
Make sure the CVAT backend and all necessary services (Database, Redis and Open Policy Agent) are running and available. If you upgraded from version 2.2.0 or earlier, manual actions may be needed, see the [Upgrade Guide](https://opencv.github.io/cvat/docs/administration/advanced/upgrade_guide).

### Context
Relevant log:

```
2023-08-28 10:02:51,691 DEBG 'uvicorn-1' stderr output:
ERROR:health-check:warning: b6837d7489b2 96.0% disk usage exceeds 90%
Traceback (most recent call last):
  File "/opt/venv/lib/python3.10/site-packages/health_check/backends.py", line 30, in run_check
    self.check_status()
  File "/opt/venv/lib/python3.10/site-packages/health_check/contrib/psutil/backends.py", line 21, in check_status
    raise ServiceWarning(
health_check.exceptions.ServiceWarning: warning: b6837d7489b2 96.0% disk usage exceeds 90%

2023-08-28 10:02:51,775 DEBG 'uvicorn-1' stderr output:
[2023-08-28 10:02:51,774] ERROR django.request: Internal Server Error: /api/server/health/

2023-08-28 10:02:51,775 DEBG 'uvicorn-1' stderr output:
ERROR:django.request:Internal Server Error: /api/server/health/

INFO:     172.21.0.1:0 - "GET /api/server/health/?format=json&org= HTTP/1.0" 500 Internal Server Error
2023-08-28 10:02:51,775 DEBG 'uvicorn-1' stdout output:
INFO:     172.21.0.1:0 - "GET /api/server/health/?format=json&org= HTTP/1.0" 500 Internal Server Error
```
```

docker exec -it -u django cvat_server python -c 'import os; print(os.statvfs("/"))'
[...]

Issue #6564, opened 2023-07-26
Deploying CVAT on AWS EC2 fails 


il-nietos:
### My actions before raising this issue
- [x ] Read/searched [the docs](https://github.com/cvat-ai/cvat/tree/master#documentation)
- [x] Searched [past issues](/issues)

### Steps to Reproduce (for bugs)

1. Follow installation steps for Ubuntu in https://opencv.github.io/cvat/docs/administration/basics/installation/#ubuntu-1804-x86_64amd64

2. EC2 security group: Allow inbound traffic with the following specifications: 
type: Custom TCP
Protocol: TCP
Port range: 8080
Source: Anywhere IP

3. export CVAT_HOST=[public IPv4 address of EC2]

4. docker compose up -d

5. In chrome go to [public IPv4 address of EC2]:8080

### Expected Behaviour
For CVAT to run in the browser with [public IPv4 address of EC2]:8080

### Current Behaviour
When I go to chrome browser with [public IPv4 address of EC2]:8080, I get the message :
'```
Cannot connect to the server
Make sure the CVAT backend and all necessary services (Database, Redis and Open Policy Agent) are running and available. If you upgraded from version 2.2.0 or earlier, manual actions may be needed, see the [Upgrade Guide](https://opencv.github.io/cvat/docs/administration/advanced/upgrade_guide)'

```

### Context
I'm trying to deploy CVAT on an AWS EC2 with Ubuntu. 

### Your Environment
<!--- Include relevant details about the environment you experienced
the bug in -->
- Docker version: Docker version 24.0.5, build ced0996
- latest version of chrome
- Operating System and version: Ubuntu 20.04.6 LTS
- Other diagnostic information / logs:
    <details>
    <summary>Logs from `cvat_dv` container</summary>
PostgreSQL Database directory appears to contain a database; Skipping initialization

2023-07-26 15:18:17.216 UTC [1] LOG:  starting PostgreSQL 15.3 on x86_64-pc-linux-musl, compiled by gcc (Alpine 12.2.1_git20220924-r10) 12.2.1 20220924, 64-bit
2023-07-26 15:18:17.216 UTC [1] LOG:  listening on IPv4 address "0.0.0.0", port 5432
2023-07-26 15:18:17.216 UTC [1] LOG:  listening on IPv6 address "::", port 5432
2023-07-26 15:18:17.221 UTC [1] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
[...]

Issue #5724, opened 2023-02-17
CVAT silently stops working when the disk usage is above 90 %


jf99:
### My actions before raising this issue
- [x] Read/searched [the docs](https://github.com/cvat-ai/cvat/tree/master#documentation). However, the link from this issue template is broken.
- [x] Searched [past issues](/issues)
  By the way, https://github.com/opencv/cvat/issues/5449#issuecomment-1348394380 did not work for me.

CVAT silently stops working when the disk usage is above 90 %.

### Steps to Reproduce (for bugs)
1. Upload new data to CVAT until it occupies more than 90 % of the disk space.

### Expected Behaviour
It is good that CVAT detects if the disk is almost full, so that one can react, before 0 Bytes are left. However, CVAT should tell the user what is wrong through the web-interface. Also, it should not be possible to upload any new data when the disk usage is above the threshold – again, tell the user why his uploads are being rejected.

### Current Behaviour
A user can upload data even if less than 10 % of the disk are free. He does not recognize until he logs out from CVAT and tries to re-login. Not only is he then confronted with an unhelpful error message:
![cvat-error](https://user-images.githubusercontent.com/7962377/219625608-06a3cdc5-2664-4bda-bf49-ed9c9948fef5.png)
But also, the user has no chance to fix the problem, as he can no longer log in.

### Possible Solution
Prevent the user from uploading too much and tell him what's the problem.

### Context
I just wasted two hours on that and finally solved it by manually deleting files from CVAT's data directory.

### Your Environment
- Git hash commit (`git log -1`): v2.3.0 plus some minor (unrelated) modifications
- Docker version `docker version` (e.g. Docker 17.0.05): 23.0.1
- Are you using Docker Swarm or Kubernetes? No.
- Operating System and version (e.g. Linux, Windows, MacOS): Ubuntu 18.04.6 LTS

rogvidarge:
Had the same issue. Was trying all sorts of fixes, but the issue was simply that I had 94% disk utilization.

Meriipu:
wow I had a fresh install without any data and 120GB of free space (even if it is at 90-92%) and I had no idea why it was frozen at the spinning loading-page

the health check is called a warning so it did not indicate to me it is a fatal error
```
cvat_server             | 2023-02-21 23:59:12,622 DEBG 'runserver' stderr output:
[...]

Issue #5685, opened 2023-02-09
Installation problem:  Cannot connect to the server


SadSadSheep:
Hello,
I installed cvat using the Installation Guide.
When I go to  http://localhost:8080/ I get the following massage:

"Cannot connect to the server
Make sure the CVAT backend and all necessary services (Database, Redis and Open Policy Agent) are running and available. If you upgraded from version 2.2.0 or earlier, manual actions may be needed, see the [Upgrade Guide](https://opencv.github.io/cvat/docs/administration/advanced/upgrade_guide)."

Here is the output for  docker ps:

<pre>CONTAINER ID   IMAGE                                 COMMAND                  CREATED        STATUS          PORTS                                                                                          NAMES
cd9931a6942a   openpolicyagent/opa:0.45.0-rootless   &quot;/opa run --server -…&quot;   6 hours ago    Up 44 minutes                                                                                                  cvat_opa
ac94d4c39927   cvat/ui:dev                           &quot;/docker-entrypoint.…&quot;   21 hours ago   Up 44 minutes   80/tcp                                                                                         cvat_ui
84af64c63593   cvat/server:dev                       &quot;/usr/bin/supervisor…&quot;   21 hours ago   Up 44 minutes   8080/tcp                                                                                       cvat_server
48457694f40d   cvat/server:dev                       &quot;/usr/bin/supervisor…&quot;   21 hours ago   Up 44 minutes   8080/tcp                                                                                       cvat_worker_annotation
bed5976033cb   cvat/server:dev                       &quot;/usr/bin/supervisor…&quot;   21 hours ago   Up 44 minutes   8080/tcp                                                                                       cvat_worker_webhooks
57741601a48e   cvat/server:dev                       &quot;/usr/bin/supervisor…&quot;   21 hours ago   Up 44 minutes   8080/tcp                                                                                       cvat_utils
6b1f3d1944cd   cvat/server:dev                       &quot;/usr/bin/supervisor…&quot;   21 hours ago   Up 44 minutes   8080/tcp                                                                                       cvat_worker_import
8e140a998a0a   cvat/server:dev                       &quot;/usr/bin/supervisor…&quot;   21 hours ago   Up 44 minutes   8080/tcp                                                                                       cvat_worker_export
8f25809c1b19   postgres:15-alpine                    &quot;docker-entrypoint.s…&quot;   21 hours ago   Up 44 minutes   5432/tcp                                                                                       cvat_db
[...]

PR #7180, opened 2023-11-25
feat: Disable django disk usage health check

### LINKED ISSUE #5449: CVAT fails health check using >90% disk
Clone latest develop.
`docker-compose -f docker-compose.yml -f docker-compose.dev.yml build`
`docker-compose up -d`

### Expected Behaviour
Startup and run application.

### Current Behaviour
Computer runs very slowly as CVAT uses significant resources. When trying to login, the following occurs:

`get http://10.0.253.33:9001/api/server/health/?format=json&org=`
`{"Cache backend: default": "working", "DatabaseBackend": "working", "DiskUsage": "warning: 86c1cd71866c 92.4% disk usage exceeds 90%", "MemoryUsage": "working", "MigrationsHealthCheck": "working"}`

Server logs:

```
2022-12-12 04:54:32,063 DEBG 'runserver' stderr output:
[Mon Dec 12 04:54:32.063061 2022] [wsgi:error] [pid 178:tid 139765320791808] ERROR:health-check:warning: 35a75041c750 92.4% disk usage exceeds 90%
[Mon Dec 12 04:54:32.063076 2022] [wsgi:error] [pid 178:tid 139765320791808] Traceback (most recent call last):
[Mon Dec 12 04:54:32.063085 2022] [wsgi:error] [pid 178:tid 139765320791808]   File "/opt/venv/lib/python3.8/site-packages/health_check/backends.py", line 30, in run_check

2022-12-12 04:54:32,063 DEBG 'runserver' stderr output:
[Mon Dec 12 04:54:32.063125 2022] [wsgi:error] [pid 178:tid 139765320791808]     self.check_status()

2022-12-12 04:54:32,063 DEBG 'runserver' stderr output:
[Mon Dec 12 04:54:32.063420 2022] [wsgi:error] [pid 178:tid 139765320791808]   File "/opt/venv/lib/python3.8/site-packages/health_check/contrib/psutil/backends.py", line 21, in check_status
[Mon Dec 12 04:54:32.063422 2022] [wsgi:error] [pid 178:tid 139765320791808]     raise ServiceWarning(
[Mon Dec 12 04:54:32.063422 2022] [wsgi:error] [pid 178:tid 139765320791808] health_check.exceptions.ServiceWarning: warning: 35a75041c750 92.4% disk usage exceeds 90%

2022-12-12 04:54:32,082 DEBG 'runserver' stderr output:
[Mon Dec 12 04:54:32.082022 2022] [wsgi:error] [pid 178:tid 139767587452672] [remote 172.27.0.4:39110] [2022-12-12 04:54:32,081] ERROR django.request: Internal Server Error: /api/server/health/
[Mon Dec 12 04:54:32.082094 2022] [wsgi:error] [pid 178:tid 139767587452672] [remote 172.27.0.4:39110] ERROR:django.request:Internal Server Error: /api/server/health/
```

Podesta:
### Motivation and context
Currently, CVAT's server completely stops working once disk usage on the drive reaches over 90% usage. This is simply a default setting, that has never been changed: [heatlh-check docs](https://django-health-check.readthedocs.io/en/stable/settings.html). As it is a django thing, it is not documented anywhere on CVAT docs.

Not only it should not be CVAT's responsibility to check disk usage, the check is not very useful, as it checks for percentage use. So even though I may have terabytes of free disk on my server, it still breaks it if I have over 90% usage.

There are several issues related to this, to name a few:
Fixes #5449 
Related #5724, #5685, #6664, #6757, #6564 

Maybe it makes sense to reintroduce a similar feature in the future, but as is, I share @azhavoro view [here](https://github.com/opencv/cvat/issues/5724#issuecomment-1659731758), that it creates more problems than it solves. But to be sensible I believe it should at least:
 1. Do not fail silently to the end user.
 2. Trigger a warning instead of effectively blocking the server.
 3. Check for both percent disk free and total disk free.

### How has this been tested?
Images have been successfully built and used with the changes.

### Checklist

### License

### DIFF
--- a/changelog.d/20231125_005843_Podesta_remove_disk_health_check.md
+++ b/changelog.d/20231125_005843_Podesta_remove_disk_health_check.md
@@ -1,0 +1,4 @@
+### Changed
+
+- Django's health-check no longer checks for disk usage.
+  (<https://github.com/opencv/cvat/pull/7180>)
--- a/cvat/settings/production.py
+++ b/cvat/settings/production.py
@@ -12,3 +12,9 @@ NUCLIO['HOST'] = os.getenv('CVAT_NUCLIO_HOST', 'nuclio')
 # https://github.com/moggers87/django-sendfile2
 SENDFILE_BACKEND = 'django_sendfile.backends.nginx'
 SENDFILE_URL = '/'
+
+# Checking for disk usage should be the user responsibility
+# https://django-health-check.readthedocs.io/en/stable/settings.html
+HEALTH_CHECK = {
+    'DISK_USAGE_MAX': None,
+}
--- a/cvat/settings/testing_rest.py
+++ b/cvat/settings/testing_rest.py
@@ -20,7 +20,6 @@ ANALYTICS_CHECK_JOB_DELAY = 10000
 IMPORT_CACHE_CLEAN_DELAY = timedelta(seconds=30)
 
 # The tests should not fail due to high disk utilization of CI infrastructure that we have no control over
-# But let's keep this check enabled
 HEALTH_CHECK = {
-    'DISK_USAGE_MAX': 100,  # percent
+    'DISK_USAGE_MAX': None,
 }


ValV:
But is it necessary to completely disable max disk usage check? I agree that 90% is kinda inadequate threshold, but isn't it possible to set to some reasonable value, say, 99% (or in MB)?

My point is that such a health check is a good indicator that the host requires manual intervention. In scenario where we have a simple VM with only CVAT deployed, we can detect lack of disk space just in time. Assuming disk size check is disabled, we can detect the problem too late (when we have 0% of free space), so we may fail to log in to the VM to clean the disk.

Podesta:
I believe it's mostly about whether or not that should be cvat task to monitor disk usage. As far as I understand, that's just a default config that have been brought over from django by default. I don't think there was a conscious decision by the cvat team to say, we should block the server if someone uses over 90% of the disk.

But yeah, I agree that a better logic, taking into account both percentage usage and absolute free disk, would be a far better solution than the current implementation. I don't think django allows for that though.

As it stands, the current implementation, I see it as an issue. Nearly weekly there is someone that gets bitten by this.

gapsong (approved):
Looks good.

azhavoro:
Discussed internally, since this check can be configured with a custom settings file, let's not change the default behavior

PR #7183, opened 2023-11-25
[Helm] Allow pre-release k8s versions to fix AWS EKS deployments

### LINKED ISSUE #7182: Helm chart's kubeVersion constraint prevents deployments to AWS EKS

### Steps to Reproduce

Using the latest version, when you try `helm install` to an EKS cluster, you get an error that looks like this:

```
chart requires kubeVersion: >= 1.19.0 which is incompatible with Kubernetes v1.25.13-eks-43840fb
```

### Expected Behavior

EKS with k8s version 1.25 should be satisfactory.

### Possible Solution

The reason of the error is explained in this issue: https://github.com/helm/helm/issues/10375

In summary, Helm interprets AWS distro version's `-eks` postfix as a pre-release version, and fails the version constraint. 

Since AWS won't change it's naming - at least for the existing versions - I think we should implement the workaround in the Chart.yaml file.

### Context

_No response_

### Environment

_No response_

ozen:
Fixes #7182 by including pre-release versions as explained in the [helm docs](https://helm.sh/docs/chart_template_guide/function_list/#working-with-prerelease-versions).

### DIFF
--- a/changelog.d/20231129_112732_canyigitozen_helm_chart_kube_version.md
+++ b/changelog.d/20231129_112732_canyigitozen_helm_chart_kube_version.md
@@ -1,0 +1,3 @@
+### Changed
+
+- \[Helm\] Allow pre-release versions in kubernetes requirement to include AWS EKS versions (https://github.com/opencv/cvat/pull/7183)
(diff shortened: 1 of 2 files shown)


SpecLad:
Thanks, looks good to me. Would you mind adding a changelog entry?

ozen:
> Thanks, looks good to me. Would you mind adding a changelog entry?

Thanks. Added changelog entry.

SpecLad:
There is a remark issue, could you fix it?

SpecLad (approved):

PR #7184, opened 2023-11-26
[Helm] Don't try to install Traefik Middleware when Traefik is disabled

ozen:
This PR is complementary to #7132. 

#7132 updated ingress templates to work with `traefik.enabled=false` and `ingress.enabled=true` values.

However the templates in `templates/analytics/middlewares` path lead to creation of resources of kind `Middleware`, which is a Traefik CRD. Helm install fails if Traefik is not installed on the cluster.

Changing the top-level conditional from `ingress.enabled` to `traefik.enabled` solves the issue.


SpecLad:
Looks good, thanks. Could you add a changelog entry?

ozen:
> Looks good, thanks. Could you add a changelog entry?

Sure, added.

SpecLad (approved):

PR #7190, opened 2023-11-27
Fix list GT settings permissions in an org

zhiltsov-max:
### Motivation and context

- Fixed a 500 error for supervisors and workers when they tried to list settings in an org context
- Added possible 403 and 404 errors for the task filter

### How has this been tested?

Unit tests

### Checklist

### License


cvat/apps/quality_control/views.py
```diff
@@ -390,6 +390,14 @@ def get_queryset(self):
         queryset = super().get_queryset()
 
         if self.action == "list":
+            if task_id := self.request.query_params.get("task_id", None):
+                try:
+                    task = Task.objects.get(id=task_id)
+                except Task.DoesNotExist as ex:
+                    raise NotFound(f"Report {task_id} does not exist") from ex
+
+                self.check_object_permissions(self.request, task)
```
Marishka17, line 399: I'm not sure that we need to return 403 and 404 exceptions at all on list request when filters are used. I thought the expected result in such cases should be an empty list. Why should we tell a user that there is no task with such an id or that there is a task with such id but he doesn't have access to it on `GET  /api/quality/settings?task_id=task_id` ?

zhiltsov-max: The idea here is that the object is specified explicitly and not filtered out.

Marishka17: Probably in such case, you need to move this into permissions
1) to keep logic in one place
2) you don't need to call  check_object_permissions
3) obj that is passed to QualitySettingPermission.create will be only QualitySettings or None
```python
elif scope == cls.Scopes.LIST:
    if task_id := request.query_params.get("task_id"):
        try:
            task = Task.objects.get(id=task_id)
            permissions.append(TaskPermission.create_scope_view(
                request, task, iam_context=iam_context,
            ))
        except Task.DoesNotExist as ex:
            raise NotFound(f"Task {task_id} does not exist") from ex
```

zhiltsov-max: I can do this, but I don't like this approach, because it breaks the view class encapsulation and, in fact, these parameter checks break quite often. It's a natural consequence, because there is no way currently to tell the permission classes about the endpoint parameters and how they should be parsed.

Marishka17 (approved):
LGTM

PR #7208, opened 2023-11-30
Client code refactoring

bsekachev:
### Motivation and context
* I converted (cvat-data, cvat-core, cvat-canvas, cvat-canvas3d) to ESM. In fact they already were ESM, but still considered as CommonJS modules without explicit ``{ "type": "module" }`` in package.json. 
* CommonJS files renamed from .js to .cjs within ESM modules (it is a part of specification)
* I added ``cvat-core`` global types, now ``getCore()`` returns typed object 
* Updated browserlist config, Firefox < 102 is not used on the market according to browserlist info. I would like also update spec for Chrome for more modern version, but these old version, like 63, 64, 65 are still used by users around the world :(
* Updated many eslint packages, minor code refactoring related to these changes, updated linting rules
* And actually many other packages also were updated
* A little bit refactored jest tests, I tried to minimize using ``window`` there. I hope soon we will be able to get rid of ``jsdom`` dependency. Right now 3rdparty packages from ``cvat-data`` don't allow to do that
* Removed extra code from ``cvat-data`` module.
* Supported typescript workers with Webpack 5, removed worker loader
* Removed events.subscribe, events.unsubscribe (we did not even have implementation for these cvat-core methods) 

### Checklist

### License


cvat-core/src/api-implementation.ts
```diff
@@ -33,39 +33,40 @@ import QualityConflict from './quality-conflict';
...
-    cvat.server.share.implementation = async (directory: string, searchPrefix?: string) => {
+    });
+    implementationMixin(cvat.server.share, async (directory: string, searchPrefix?: string) => {
         const result = await serverProxy.server.share(directory, searchPrefix);
         return result.map((item) => ({ ...omit(item, 'mime_type'), mimeType: item.mime_type }));
-    };
-
-    cvat.server.formats.implementation = async () => {
+    });
+    implementationMixin(cvat.server.formats, async () => {
```
klakhov, line 61: Please, add newlines inbetween the implementations. Or maybe just between the groups cvat.`something`

cvat-core/src/download.worker.ts
klakhov, line 1: Header update?

cvat-core/src/index.ts
```diff
@@ -0,0 +1,150 @@
...
+    server: {
+        about: typeof serverProxy.server.about;
+        share: (dir: string) => Promise<{
+            mimeType: string;
+            name: string;
+            type: enums.ShareFileType;
+        }[]>;
+        formats: () => Promise<AnnotationFormats>;
+        userAgreements: typeof serverProxy.server.userAgreements,
+        register: any; // TODO: add types later
```
klakhov, line 54: Will it be added in this patch or wait for next one?

cvat-ui/src/components/import-dataset/import-dataset-modal.tsx
```diff
@@ -234,12 +234,15 @@ function ImportDatasetModal(props: StateToProps): JSX.Element {
...
-            ) as any);
+            dispatch(
+                importDatasetAsync(
+                    instance,
+                    uploadParams.selectedFormat as string,
+                    uploadParams.useDefaultSettings,
+                    uploadParams.sourceStorage,
+                    uploadParams.file || uploadParams.fileName as string,
+                    uploadParams.convMaskToPoly,
+                ) as any);
```
klakhov, line 245: I dont really think we need this `any`

cvat-core/src/index.ts
bsekachev: I was planning to add these types gradually. It is too big task for one PR

cvat-core/src/api-implementation.ts
bsekachev: Right now there are newlines exactly between groups, like between cvat.server and cvat.assets, etc. 
If to add these spaces between each row, we will lose some kind of visual difference


We can't add 2 new blank lines, because eslint complaints about it

cvat-data/src/ts/3rdparty/README.md
```diff
@@ -1,7 +1,7 @@
 ## 3rdparty components
 
 These files are from the [Broadway.js](https://github.com/mbebenita/Broadway) repository:
-- Decoder.js
+- Decoder.cjs
```
SpecLad, line 4: There is no `Decoder.cjs`.

cvat-data/src/ts/3rdparty_patch.diff
SpecLad, line 1: Why did you delete this?

cvat-canvas/webpack.config.cjs
SpecLad, line 1: Does webpack not support ESM configs?

cvat-data/src/ts/3rdparty_patch.diff
bsekachev: Because we did not use it

cvat-canvas/webpack.config.cjs
bsekachev: I didn't really understand your question

cvat-core/src/api-implementation.ts
klakhov: Ok, I see

cvat-data/src/ts/3rdparty_patch.diff
bsekachev: - Initially we used this patch file to patch the library during build
- As I discovered, patched version was already commited long time ago and we don't use the file anymore
- Finally I do not really see a reason to patch this file, their repository is dead, we don't receive any updates. License allows us to modify it, so..

cvat-canvas/webpack.config.cjs
bsekachev: If you mean webpack.config file is a common js module, so, I decided to leave configurations as they are. 
They write than webpack should support ESM configurations in our version. 

But it lead to issues with eslint/jest configurations when I tried to convert them to ESM.

SpecLad: Okay.

PR #7216, opened 2023-12-04
cvat-core: retry HTTP requests if a 429 status is returned

SpecLad:
### Motivation and context

This should help reduce breakage if a user hits an API rate limit.

### How has this been tested?

Manual testing.

### Checklist

### License

### DIFF
--- a/cvat-core/src/server-proxy.ts
+++ b/cvat-core/src/server-proxy.ts
@@ -6,6 +6,7 @@
 import FormData from 'form-data';
 import store from 'store';
 import Axios, { AxiosError, AxiosResponse } from 'axios';
+import axiosRetry from 'axios-retry';
 import * as tus from 'tus-js-client';
 import { ChunkQuality } from 'cvat-data';
 
@@ -282,6 +283,19 @@ Axios.interceptors.response.use((response) => {
     }
 
     return response;
+});
+
+axiosRetry(Axios, {
+    retryCondition: (error) => error.response && error.response.status === Axios.HttpStatusCode.TooManyRequests,
+    retryDelay: (retryCount, error) => {
+        const retryAfterValue = error.response.headers['retry-after'];
+
+        // Retry-After is allowed to be a number as well as a date.
+        // We're probably not going to use the latter option, though, so don't bother parsing it.
+        if (!retryAfterValue || !/^\d+$/.test(retryAfterValue)) return axiosRetry.exponentialDelay(retryCount, error);
+
+        return parseInt(retryAfterValue, 10) * 1000;
+    },
 });
 
 let token = store.get('token');
--- a/cvat-core/src/server-proxy.ts
+++ b/cvat-core/src/server-proxy.ts
@@ -6,10 +6,10 @@
 import FormData from 'form-data';
 import store from 'store';
 import Axios, { AxiosError, AxiosResponse } from 'axios';
-import axiosRetry from 'axios-retry';
 import * as tus from 'tus-js-client';
 import { ChunkQuality } from 'cvat-data';
 
+import './axios-config';
 import {
     SerializedLabel, SerializedAnnotationFormats, ProjectsFilter,
     SerializedProject, SerializedTask, TasksFilter, SerializedUser, SerializedOrganization,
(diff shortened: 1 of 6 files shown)


cvat-core/src/server-proxy.ts
```diff
@@ -284,6 +285,19 @@ Axios.interceptors.response.use((response) => {
...
+    retryCondition: (error) => error.response && error.response.status === Axios.HttpStatusCode.TooManyRequests,
+    retryDelay: (retryCount, error) => {
+        const retryAfterValue = error.response.headers['retry-after'];
+
+        // Retry-After is allowed to be a number as well as a date.
+        // We're probably not going to use the latter option, though, so don't bother parsing it.
+        if (!retryAfterValue || !/^\d+$/.test(retryAfterValue)) return axiosRetry.exponentialDelay(retryCount, error);
+
+        return parseInt(retryAfterValue, 10) * 1000;
+    },
```
bsekachev, line 298: If ``error.response.headers['retry-after']`` is supposed to be a number or not a number string:

```suggestion
    retryDelay: (retryCount: number, error: Error): number => {
        // Retry-After is allowed to be a number as well as a date.
        // We're probably not going to use the latter option, though, so don't bother parsing it.
        const retryTimeout = (+error.response.headers['retry-after']) * 1000;
        if (Number.isNaN(retryTimeout)) {
            return axiosRetry.exponentialDelay(retryCount, error);
        }
        return retryTimeout;
    },
```

bsekachev (commented):
We also have a worker ``cvat-core/src/download.worker.ts``  to download data chunks in another thread
Will this plugin have effect inside the worker?

bsekachev:
But to be honest, looking at the worker now, I do not remember why it was added. 
And I have some doubts we need it, need to check additionally

cvat-core/src/server-proxy.ts
SpecLad: By "number" I really mean "non-negative decimal integer". See the [definition of `Retry-After`](https://www.rfc-editor.org/rfc/rfc9110#field.retry-after).

Blindly converting to a JavaScript number could yield undesirable values like negative numbers or infinity, so I'd rather not do that.

bsekachev: Negative numbers are not covered by the code above.
For avoid infinity/floats we may use:

``Number.isInteger(retryTimeout) && retryTimeout > 0``

Instead of ``isNaN``

SpecLad:
> But to be honest, looking at the worker now, I do not remember why it was added. And I have some doubts we need it, need to check additionally

Well, you tell me if you want me to add retry code to the worker. 🤷‍♂️

cvat-core/src/server-proxy.ts
SpecLad: > It is a little bit more pretty, than regex using

I don't know about pretty, but I find my version more straightforward. I verify that:

1. The header is present.
2. The header value follows the expected grammar.

Then I convert the value to the expected type.

Your version relies on JS's numeric conversion doing the right thing for values not matching the grammar, which isn't even true:

```console
> +'0xA'
10
```

bsekachev: Why not true? You just converted HEX 0xA to decimal 10. 
And this is correct.

bsekachev: image

But we do not expect `+55` from the server, right? As well as ``0xA``..

bsekachev:
> Well, you tell me if you want me to add retry code to the worker. 🤷‍♂️

First of all need to check that this retry logic has effect inside the worker, if it does not, so, please add the code there.
In the future somebody will check if we really need this worker.

cvat-core/src/server-proxy.ts
SpecLad: > Why not true? You just converted HEX 0xA to decimal 10.
> And this is correct.

It's not correct, because hexadecimal is not allowed by the grammar.

> Let's check with regex any data come from the server then...?

Sure? If it's amenable to such validation, then why not.

> image

What's your point? Plus signs are not allowed by the grammar, so this string should be rejected - and that's what my implementation does.
 
> But we do not expect `+55` from the server, right? As well as `0xA`..

We don't - and that's why I reject such values.

SpecLad:
> First of all need to check that this retry logic has effect inside the worker, if it does not, so, please add the code there.

Okay, I'll do that.

(FWIW, I'm pretty sure the current code will not affect the worker, since the worker doesn't import `server-proxy.ts`.)

SpecLad:
I updated the code to share basic configuration (including retries) between the worker and `server-proxy.ts`.

bsekachev (approved):

PR #7259, opened 2023-12-13
Removed `PermissionDenied` error thrown before OPA call

klakhov:
### Motivation and context

At first this part of code was an optimization. If we know that user is not a member but tries to access org resourses we throw an error. But during development of other features(like invitations to organization) we need to give user an access to such resources. That leads to rewriting this part of code with lots of `if` statements.

Generally all permission decision should be made by OPA.

### Checklist

### License


Marishka17:
Probably you need to add an appropriate OPA test to list /users in an org and therefore update [users.csv](https://github.com/opencv/cvat/blob/develop/cvat/apps/iam/rules/tests/configs/users.csv) with something like `list,N/A,Organization,N/A,,GET,/users,None,Worker`.

klakhov:
@Marishka17 
I've added test generation for list scope, could you check please if thats ok?
All of them are basically `allow` because /users just filters everyone except the user who sent the request.

cvat/apps/iam/rules/tests/configs/users.csv
```diff
@@ -1,5 +1,6 @@
 Scope,Resource,Context,Ownership,Limit,Method,URL,Privilege,Membership
 list,User,N/A,N/A,,GET,/users,None,N/A
```
Marishka17, line 2:
```suggestion
list,User,Sandbox,N/A,,GET,/users,None,N/A
```

cvat/apps/iam/rules/tests/generators/users_test.gen.rego.py
```diff
@@ -127,8 +123,6 @@ def get_name(scope, context, ownership, privilege, membership, resource):
 def is_valid(scope, context, ownership, privilege, membership, resource):
     if context == "sandbox" and membership:
         return False
-    if scope == "list" and ownership != "None":
-        return False
```
Marishka17, line 131: Probably you need to change it to
```python
if scope == "list":
    return not ownership == "self"
```

cvat/apps/iam/rules/tests/generators/users_test.gen.rego.py
```diff
@@ -46,13 +46,9 @@ def read_rules(name):
...
-    if scope == "list":
-        return [None]
-    else:
-        return [
-            {"id": random.randrange(300, 400), "membership": {"role": role}} for role in ORG_ROLES
-        ]
-
+    return [
+        {"id": random.randrange(300, 400), "membership": {"role": role}} for role in ORG_ROLES
+    ]
```
Marishka17, line 51: You don't need these changes because resource should always be `None` when scope == "list"

Marishka17:
@klakhov, We should check 3 cases in such tests:
1) scope == list & is_sandbox => allow
2) scope == list & is_org & is_member => allow
3) scope == list & is_org & not is_member => reject

With the changes I've suggested, the mentioned cases will be covered and there will be several tests like
```
test_scope_LIST_context_ORGANIZATION_ownership_NONE_privilege_BUSINESS_membership_NONE_resource_NONE {
    not allow with input as {"scope": "list", "auth": {"user": {"id": 9, "privilege": "business"}, "organization": {"id": 120, "owner": {"id": 243}, "user": {"role": null}}}, "resource": null}
}
```

cvat/apps/iam/rules/tests/configs/users.csv
klakhov: Done

klakhov: Done

Marishka17 (approved):
LGTM

PR #7295, opened 2023-12-26
Remove empty masks 

klakhov:
### Checklist

### License

### DIFF
--- a/cvat-core/src/annotations-collection.ts
+++ b/cvat-core/src/annotations-collection.ts
@@ -13,7 +13,7 @@ import AnnotationsFilter from './annotations-filter';
 import { checkObjectType } from './common';
 import Statistics from './statistics';
 import { Attribute, Label } from './labels';
-import { ArgumentError, ScriptingError } from './exceptions';
+import { ArgumentError, IncorrectUpdateError, ScriptingError } from './exceptions';
 import ObjectState from './object-state';
 import { cropMask } from './object-utils';
 import config from './config';
@@ -997,6 +997,12 @@ export default class Collection {
                     throw new ArgumentError(
                         `Object shape must be one of: ${JSON.stringify(Object.values(ShapeType))}`,
                     );
+                }
+
+                if (state.shapeType === 'mask') {
+                    if (state.points.length === 5) {
+                        throw new IncorrectUpdateError('Cant create empty mask', [state.clientID]);
+                    }
                 }
 
                 if (state.objectType === 'shape') {
--- a/cvat-core/src/annotations-objects.ts
+++ b/cvat-core/src/annotations-objects.ts
@@ -6,7 +6,9 @@
 import config from './config';
 import ObjectState, { SerializedData } from './object-state';
 import { checkObjectType, clamp } from './common';
-import { DataError, ArgumentError, ScriptingError } from './exceptions';
+import {
+    DataError, ArgumentError, ScriptingError, IncorrectUpdateError,
+} from './exceptions';
 import { Label } from './labels';
 import {
     colors, Source, ShapeType, ObjectType, HistoryActions, DimensionType,
@@ -2143,6 +2145,9 @@ export class MaskShape extends Shape {
 
     protected validateStateBeforeSave(data: ObjectState, updated: ObjectState['updateFlags'], frame?: number): number[] {
         Annotation.prototype.validateStateBeforeSave.call(this, data, updated);
+        if (data.points.length === 5) {
+            throw new IncorrectUpdateError('cant save empty mask', [data.clientID]);
+        }
         if (updated.points) {
             const { width, height } = this.frameMeta[frame];
             return cropMask(data.points, width, height);
@@ -2192,9 +2197,16 @@ export class MaskShape extends Shape {
             }
         }
 
+        const incorrectMasks = [];
         for (const object of Object.values(updatedObjects)) {
             object.points = mask2Rle(masks[object.clientID]);
             object.updated = Date.now();
+            if (object.points.length === 1) {
+                incorrectMasks.push(object.clientID);
+            }
+        }
+        if (incorrectMasks.length !== 0) {
+            throw new IncorrectUpdateError('cant save empty mask', incorrectMasks);
         }
     }
 
--- a/cvat-ui/src/actions/annotation-actions.ts
+++ b/cvat-ui/src/actions/annotation-actions.ts
@@ -13,7 +13,7 @@ import {
     RectDrawingMethod, CuboidDrawingMethod, Canvas, CanvasMode as Canvas2DMode,
 } from 'cvat-canvas-wrapper';
 import {
-    getCore, MLModel, JobType, Job, QualityConflict,
+    getCore, MLModel, JobType, Job, QualityConflict, IncorrectUpdateError,
 } from 'cvat-core-wrapper';
 import logger, { LogType } from 'cvat-logger';
 import { getCVATStore } from 'cvat-store';
@@ -199,6 +199,7 @@ export enum AnnotationActionTypes {
     RESTORE_FRAME_FAILED = 'RESTORE_FRAME_FAILED',
     UPDATE_BRUSH_TOOLS_CONFIG = 'UPDATE_BRUSH_TOOLS_CONFIG',
     HIGHLIGHT_CONFLICT = 'HIGHLIGHT_CONFCLICT',
+    INCORRECT_ANNOTATIONS_UPDATE = 'INCORRECT_ANNOTATIONS_UPDATE',
 }
 
 export function saveLogsAsync(): ThunkAction {
@@ -1075,7 +1076,23 @@ export function updateAnnotationsAsync(statesToUpdate: any[]): ThunkAction {
             }
 
             const promises = statesToUpdate.map((objectState: any): Promise<any> => objectState.save());
-            const states = await Promise.all(promises);
+            let results = await Promise.allSettled(promises);
+            results = results.map((s, index) => {
+                if (s.status === 'fulfilled') {
+                    return Promise.resolve(s.value);
+                }
+                if (s.reason instanceof IncorrectUpdateError) {
+                    dispatch({
+                        type: AnnotationActionTypes.INCORRECT_ANNOTATIONS_UPDATE,
+                        payload: {
+                            error: s.reason,
+                        },
+                    });
+                    return statesToUpdate[index].delete(frame);
+                }
+                return Promise.reject(s.reason);
+            });
+            const states = await Promise.all(results);
 
             const needToUpdateAll = states
                 .some((state: any) => state.shapeType === ShapeType.MASK || state.parentID !== null);
@@ -1113,7 +1130,24 @@ export function createAnnotationsAsync(sessionInstance: NonNullable<CombinedStat
     return async (dispatch: ActionCreator<Dispatch>): Promise<void> => {
         try {
             const { filters, showAllInterpolationTracks } = receiveAnnotationsParameters();
-            await sessionInstance.annotations.put(statesToCreate);
+            try {
+                await sessionInstance.annotations.put(statesToCreate);
+            } catch (error) {
+                const states = await sessionInstance.annotations.get(frame, showAllInterpolationTracks, filters);
+                if (error instanceof IncorrectUpdateError) {
+                    await Promise.all(error.clientIDs.map((clientID: number) => (
+                        states.find(((state) => state.clientID === clientID))?.delete(frame))),
+                    );
+                    dispatch({
+                        type: AnnotationActionTypes.INCORRECT_ANNOTATIONS_UPDATE,
+                        payload: {
+                            error,
+                        },
+                    });
+                } else {
+                    throw error;
+                }
+            }
             const states = await sessionInstance.annotations.get(frame, showAllInterpolationTracks, filters);
             const history = await sessionInstance.actions.get();
 
--- a/cvat-core/src/exceptions.ts
+++ b/cvat-core/src/exceptions.ts
@@ -120,6 +120,21 @@ export class Exception extends Error {
 
 export class ArgumentError extends Exception {}
 
+export class IncorrectUpdateError extends Exception {
+    constructor(message, clientIDs) {
+        super(message);
+
+        Object.defineProperties(
+            this,
+            Object.freeze({
+                clientIDs: {
+                    get: () => clientIDs,
+                },
+            }),
+        );
+    }
+}
+
 export class DataError extends Exception {}
 
 export class ScriptingError extends Exception {}
--- a/cvat-ui/src/cvat-core-wrapper.ts
+++ b/cvat-ui/src/cvat-core-wrapper.ts
@@ -21,7 +21,7 @@ import QualityReport, { QualitySummary } from 'cvat-core/src/quality-report';
 import QualityConflict, { AnnotationConflict, ConflictSeverity } from 'cvat-core/src/quality-conflict';
 import QualitySettings from 'cvat-core/src/quality-settings';
 import { FramesMetaData, FrameData } from 'cvat-core/src/frames';
-import { ServerError } from 'cvat-core/src/exceptions';
+import { ServerError, IncorrectUpdateError } from 'cvat-core/src/exceptions';
 import {
     ShapeType, LabelType, ModelKind, ModelProviders,
     ModelReturnType, DimensionType, JobType,
@@ -93,6 +93,7 @@ export {
     AnalyticsEntry,
     AnalyticsEntryViewType,
     ServerError,
+    IncorrectUpdateError,
     EventLogger,
     FrameData,
     ActionParameterType,
--- a/cvat-ui/src/reducers/notifications-reducer.ts
+++ b/cvat-ui/src/reducers/notifications-reducer.ts
@@ -1901,6 +1901,23 @@ export default function (state = defaultState, action: AnyAction): Notifications
                 },
             };
         }
+        case AnnotationActionTypes.INCORRECT_ANNOTATIONS_UPDATE: {
+            return {
+                ...state,
+                errors: {
+                    ...state.errors,
+                    annotation: {
+                        ...state.errors.annotation,
+                        updating: {
+                            message: 'Could not update annotations',
+                            reason: action.payload.error,
+                            shouldLog: !(action.payload.error instanceof ServerError),
+                            className: 'cvat-notification-notice-update-annotations-failed',
+                        },
+                    },
+                },
+            };
+        }
         case BoundariesActionTypes.RESET_AFTER_ERROR:
         case AuthActionTypes.LOGOUT_SUCCESS: {
             return { ...defaultState };


cvat-core/src/annotations-collection.ts
```diff
@@ -999,6 +999,12 @@ export default class Collection {
                     );
                 }
 
+                if (state.shapeType === 'mask') {
+                    if (state.points.length === 5) {
```
bsekachev, line 1003: maybe correct comparison is length < 6?

xtl, ytl, xbr, ybr and at least one filled pixel would have length 6

bsekachev: Also two nested condition may be joined with &&

bsekachev: This check is useful anyway
But I believe we should fix also cvat-canvas

cvat-core/src/annotations-collection.ts
```diff
@@ -999,6 +999,12 @@ export default class Collection {
...
+                        throw new IncorrectUpdateError('Cant create empty mask', [state.clientID]);
```
bsekachev, line 1004: That is not IncorrectUpdateError, this is ArgumentError

bsekachev: btw clientID is empty for not saved mask in collection

cvat-ui/src/actions/annotation-actions.ts
```diff
@@ -1075,7 +1076,23 @@ export function updateAnnotationsAsync(statesToUpdate: any[]): ThunkAction {
...
+                    return Promise.resolve(s.value);
+                }
+                if (s.reason instanceof IncorrectUpdateError) {
+                    dispatch({
+                        type: AnnotationActionTypes.INCORRECT_ANNOTATIONS_UPDATE,
+                        payload: {
+                            error: s.reason,
+                        },
+                    });
+                    return statesToUpdate[index].delete(frame);
```
bsekachev, line 1091: Why we delete objects we tried to update? (as I understand it is a case when we totally erase a mask during editing) 
They should not be updated if it leads to their incorrect state

Try to create a bounding box and move it out of frame, it will be returned to the previous position automatically. I expect something like this for masks

cvat-core/src/annotations-objects.ts
```diff
@@ -2192,9 +2197,16 @@ export class MaskShape extends Shape {
             }
         }
 
+        const incorrectMasks = [];
         for (const object of Object.values(updatedObjects)) {
             object.points = mask2Rle(masks[object.clientID]);
             object.updated = Date.now();
+            if (object.points.length === 1) {
```
bsekachev, line 2204: < 2 would be more reliable

cvat-core/src/annotations-objects.ts
```diff
@@ -2192,9 +2197,16 @@ export class MaskShape extends Shape {
...
+                incorrectMasks.push(object.clientID);
+            }
+        }
+        if (incorrectMasks.length !== 0) {
+            throw new IncorrectUpdateError('cant save empty mask', incorrectMasks);
```
bsekachev, line 2209: I do not think raising exceptions is a good idea when we speak about warnings.
It will prevent any further code to execution until the exception is handled (and even if all the work is already done in this case, we must remember this fact always during future development, and avoid adding any code in other files/modules. In practice, I believe it will lead to problems)

Generally speaking, JS exceptions mechanism will not work for warnings

bsekachev: As far as I remember we discussed that these "empty" masks appeared as a result of removeUnderlyingPixels logic should be removed (in this case nothing happens with them). 

I see it is implemented in cvat-ui, but that is another module. Total spagetti code, we should find a way to handle everything in cvat-core

cvat-core/src/annotations-collection.ts
klakhov: Fixed if case.
Could you clarify please what should we fix in cvat-canvas?

cvat-ui/src/actions/annotation-actions.ts
klakhov: Fixed, now if mask becomes empty it is not updated

cvat-core/src/annotations-collection.ts
bsekachev: I suppose a case when an empty mask created (using polygon minus tool)

cvat-core/src/annotations-objects.ts
klakhov: Fixed

cvat-core/src/annotations-objects.ts
klakhov: Removed Error. Added logic with removing in cvat-core. (We still need to show some warning in UI so ive added a callback to config)

bsekachev:
Found inconsistencies:

Still can create empty mask in cvat-canvas module. It can be passed to cvat-core where exception is thrown and message is shown.
In case with other shapes, it is not created without a message

Create a normal mask, modify it to empty => exception is thrown, message is shown
In case of other shapes, it cancels change without a message

Found problems: 
Message: "As a result of removing the underlying pixels, some masks became empty and were subsequently deleted." disappears to fast. I was not even able to read it until the end for the first time.
Message: "As a result of removing the underlying pixels, some masks became empty and were subsequently deleted." appears everytime when I create a mask with enabled "removeUnderlyingPixels"


Let's try to modify the message, splitting it to title and description. Something like this?
image

cvat-core/src/annotations-collection.ts
```diff
@@ -1095,27 +1101,46 @@ export default class Collection {
...
-                (object as MaskShape).removeUnderlyingPixels(object.frame);
+            if (object.shapeType === ShapeType.MASK && config.removeUnderlyingMaskPixels.enabled) {
+                const {
+                    clientIDs,
+                    undo: undoWithRemoveIncorrectMasks,
+                    redo: redoWithRemoveIncorrectMasks,
+                } = (object as MaskShape).removeUnderlyingPixels(object.frame);
+                additionalUndo.push(undoWithRemoveIncorrectMasks);
+                additionalRedo.push(redoWithRemoveIncorrectMasks);
+                additionalClientIDs.push(clientIDs);
```
bsekachev, line 1116: Shall we be able to undo/redo any changes when removeUndrelyingPixels enabled?
I mean not only to restore deleted masks, but also cancel changes in all affected masks?

Right now that is what this feature lack of

klakhov: Added

klakhov:
I belive everything is consistent.
> Found inconsistencies:
> 
> Still can create empty mask in cvat-canvas module. It can be passed to cvat-core where exception is thrown and message is shown. In case with other shapes, it is not created, message does not appear
Why there is no exception with other shapes?

We have the same logic for all shapes in [checkNumOfPoints](https://github.com/opencv/cvat/pull/7295/files#diff-9fad337ebdaab50e76e5b1e57643ce314bec8d0ec776a800c83c6ec0dcf1a9baR9) function. If I try to create rect with <4 points then an errror occurs
```
            const state = new cvat.classes.ObjectState({
                shapeType: ShapeType.RECTANGLE,
                objectType: ObjectType.SHAPE,
                points: [0],
                occluded: false,
                attributes: {},
                frame: 1,
                label: jobInstance.labels[0],
            });
            const [clientID] = await jobInstance.annotations.put([state]);
```
image


> Create a normal mask, modify it to empty => exception is thrown, message is shown In case of other shapes, it cancels changes, message does not appear

If I try to set rectangle to <4 points and save it then the same exeption is thrown. Is there even a way to create such case for other shapes than masks?



> Found problems: Message: "As a result of removing the underlying pixels, some masks became empty and were subsequently deleted." disappears to fast. I was not even able to read it until the end for the first time. Message: "As a result of removing the underlying pixels, some masks became empty and were subsequently deleted." appears everytime when I create a mask with enabled "removeUnderlyingPixels"
> 
> Let's try to modify the message, splitting it to title and description. Something like this? image

Fixed the message.

bsekachev:
How can you create/edit a rectangle setting it less then 4 points?

bsekachev:
> We have the same logic for all shapes in [checkNumOfPoints](https://github.com/opencv/cvat/pull/7295/files#diff-9fad337ebdaab50e76e5b1e57643ce314bec8d0ec776a800c83c6ec0dcf1a9baR9) function. If I try to create rect with <4 points then an errror occurs

Yes, we have in cvat-core
But `cvat-canvas` will not even try to save a rectangle with less number of points for a rectangle. 
The same for polyshapes. For example when polygon has only three points, the interface don't allow to remove any more points:
image
image

bsekachev:
I believe locking polygon-minus and eraser tool if mask is empty in cvat-canvas would be a good solution in first.
This way let users avoid misclicking a button. Probably it is one of ways why such masks are created, not intentionally 
One more way, is when we finished drawing a mask using eraser/polygon minus, creating a new shape will have this tool as default one. What is also not really user friendly. 

Finally if a user created an empty mask in cvat-canvas, it should not raise shape.drawn event

klakhov:
> How can you create/edit a rectangle setting it less then 4 points?

Using cvat-core api via code.

> I believe locking polygon-minus and eraser tool if mask is empty in cvat-canvas would be a good solution in first.
> Finally if a user created an empty mask in cvat-canvas, it should not raise shape.drawn event

Okay, ill think about that

cvat-canvas/src/typescript/canvasModel.ts
```diff
@@ -103,6 +103,7 @@ export interface BrushTool {
     color: string;
     form: 'circle' | 'square';
     size: number;
+    configureBlockedTools: (blockToolsConfiguration: Record<'eraser' | 'polygon-minus', boolean>) => void,
```
bsekachev, line 106: Since it is a callback, I would suggest callback style naming. Maybe:

```suggestion
    onBlockUpdated: (blockedTools: Record<'eraser' | 'polygon-minus', boolean>) => void,
```

cvat-canvas/src/typescript/canvasView.ts
```diff
@@ -283,7 +283,9 @@ export class CanvasViewImpl implements CanvasView, Listener {
             }
         }
 
-        if (data) {
+        const emptyMask = data && data.shapeType === 'mask' && data.points.length < 6;
```
bsekachev, line 286:
```suggestion
        const isEmptyMask = data && data.shapeType === 'mask' && data.points.length < 6;
```

cvat-canvas/src/typescript/canvasView.ts
```diff
@@ -361,7 +363,9 @@ export class CanvasViewImpl implements CanvasView, Listener {
 
     private onEditDone = (state: any, points: number[], rotation?: number): void => {
         this.canvas.style.cursor = '';
-        if (state && points) {
+        const emptyMask = state && points && state.shapeType === 'mask' && points.length < 6;
```
bsekachev, line 366:
```suggestion
        const isEmptyMask = state && points && state.shapeType === 'mask' && points.length < 6;
```

cvat-canvas/src/typescript/canvasView.ts
bsekachev: I believe would be better to check a mask is not empty for both editing and drawing in `masksHandler.ts`.

cvat-canvas/src/typescript/masksHandler.ts
```diff
@@ -294,6 +295,42 @@ export class MasksHandlerImpl implements MasksHandler {
...
+        }
+        const rle = zipChannels(imageData);
+        const emptyMask = rle.length < 2;
+        this.tool.configureBlockedTools({
+            eraser: emptyMask,
+            'polygon-minus': emptyMask,
+        });
+    }
+
+    private createDrawnObjectsArray(): (fabric.Polygon | fabric.Circle | fabric.Rect | fabric.Line | fabric.Image)[] {
```
bsekachev, line 322:
```suggestion
    private createDrawnObjectsArray(): MasksHandlerImpl['drawnObjects'] {
```

cvat-canvas/src/typescript/masksHandler.ts
```diff
@@ -294,6 +295,42 @@ export class MasksHandlerImpl implements MasksHandler {
...
+        const imageData = this.imageDataFromCanvas(wrappingBbox);
+        if (this.brushMarker) {
+            this.canvas.add(this.brushMarker);
+        }
+        const rle = zipChannels(imageData);
+        const emptyMask = rle.length < 2;
+        this.tool.configureBlockedTools({
+            eraser: emptyMask,
+            'polygon-minus': emptyMask,
+        });
```
bsekachev, line 319: 1. Let's wrap adding brushMarker back to try ... finally 
2. Probably we can rewrite as `private funcName = lodash.debounce( ... )`
3. I would rename as `updateBlockedTools` or something like that

cvat-canvas/src/typescript/masksHandler.ts
```diff
@@ -326,6 +362,7 @@ export class MasksHandlerImpl implements MasksHandler {
             defaultCursor: 'inherit',
         });
         this.canvas.imageSmoothingEnabled = false;
+        this.drawnObjects = this.createDrawnObjectsArray();
```
bsekachev, line 365: Initialized as a regular array in `releasePaste` 
So, when drawing after paste, it will not work

bsekachev:
Small issue with colors during editing: 
Peek 2024-01-15 12-58

cvat-canvas/src/typescript/masksHandler.ts
```diff
@@ -551,6 +588,10 @@ export class MasksHandlerImpl implements MasksHandler {
                 }
             }
 
+            if (drawData.brushTool) {
+                this.checkEmpty();
```
bsekachev, line 592: checkEmpty uses ``this.tool``.
``this.tool`` initialized only in `this.updateBrushTools`
`this.updateBrushTools` not called in case of insertion. 

So, the condition is incorrect and probably this code should be inserted to `else` branch above

bsekachev:
There is an issue with inserting. Class has changed
Peek 2024-01-15 13-07

cvat-core/src/annotations-collection.ts
```diff
@@ -1006,6 +1006,12 @@ export default class Collection {
                     );
                 }
 
+                if (state.shapeType === 'mask' && state.points.length < 6) {
+                    throw new ArgumentError(
+                        'Could not create empty mask',
+                    );
```
bsekachev, line 1012:
```suggestion
                    throw new ArgumentError('Could not create an empty mask');
```

cvat-core/src/annotations-collection.ts
```diff
@@ -1095,27 +1101,49 @@ export default class Collection {
...
+                const {
+                    clientIDs,
+                    emptyMaskOccured,
+                    undo: undoWithUnderlyingPixels,
+                    redo: redoWithUnderlyingPixels,
+                } = (object as MaskShape).removeUnderlyingPixels(object.frame);
+                additionalUndo.push(undoWithUnderlyingPixels);
+                additionalRedo.push(redoWithUnderlyingPixels);
+                additionalClientIDs.push(clientIDs);
+                globalEmptyMaskOccured = emptyMaskOccured || globalEmptyMaskOccured;
```
bsekachev, line 1119: grammar issue occured => occurred

cvat-core/src/annotations-objects.ts
```diff
@@ -2228,10 +2235,39 @@ export class MaskShape extends Shape {
...
+            stashedPoints = [];
+            for (const object of Object.values(updatedObjects)) {
+                const points = mask2Rle(masks[object.clientID]);
+                stashedPoints.push([...object.points]);
+                object.points = points;
+                if (points.length < 2) {
+                    object.removed = true;
+                    emptyMaskOccured = true;
+                }
+                object.updated = Date.now();
```
bsekachev, line 2260: if (points.length < 2) {
                    object.removed = true;
                    emptyMaskOccured = true;
                } else {
                    object.points = points;
                    object.updated = Date.now();
                }

cvat-core/src/annotations-objects.ts
```diff
@@ -2228,10 +2235,39 @@ export class MaskShape extends Shape {
...
-        }
+        let emptyMaskOccured = false;
+        let stashedPoints = [];
+        const undo = (): void => {
+            for (const [index, object] of Object.values(updatedObjects).entries()) {
+                if (object.points.length < 2) {
+                    object.removed = false;
+                }
+                object.points = stashedPoints[index];
+                object.updated = Date.now();
```
bsekachev, line 2246: Why not? 

```suggestion
                if (object.points.length < 2) {
                    object.removed = false;
                } else {
                    object.points = stashedPoints[index];
                    object.updated = Date.now();
                }
```

cvat-core/src/annotations-objects.ts
```diff
@@ -2228,10 +2235,39 @@ export class MaskShape extends Shape {
...
+            }
+        };
+
+        const redo = (): void => {
+            stashedPoints = [];
+            for (const object of Object.values(updatedObjects)) {
+                const points = mask2Rle(masks[object.clientID]);
```
bsekachev, line 2253: There will be a problem with huge memory consumption

In closure it refers to variable `masks` that contains all unpacked masks stored on the frame (imagine that there are 100+ masks per frame).
And undo/redo stores up to 128 actions as far as I remember 

It gigabytes of data

cvat-canvas/src/typescript/canvasView.ts
klakhov: Do you mean calling `onEditDone(null)`/`onDrawDone(null)` in `draw`/`edit`?

bsekachev: Yep

cvat-canvas/src/typescript/canvasModel.ts
klakhov: Done

cvat-canvas/src/typescript/masksHandler.ts
klakhov: Fixed

cvat-canvas/src/typescript/masksHandler.ts
klakhov: Renamed.
Why do we need to wrap it in try catch? I doesnt look we use try anywhere with canvas.add?
I dont think we need to rewrite with `private funcName = lodash.debounce( ... )` because this funtion is used in proxy wrapper where we cant really access `this` without alias

cvat-canvas/src/typescript/masksHandler.ts
klakhov: Moved to `this.updateBrushTools`

cvat-core/src/annotations-objects.ts
klakhov: `masks` store not all masks on the frame but only affected ones, but still there can be a problem.
Ive added limitations to annotation history, so there only can be `20` undos for removing underlying pixels. (maybe we should change this number to even lower value)

klakhov:
@bsekachev 

The issue with copying mask that was changing class seems to be reproducible on develop, its separate bug but Ive fixed it.

I cant really reproduce the issue with color changing on last verion of PR, so maybe its fixed already.

cvat-core/src/object-utils.ts
```diff
@@ -32,14 +32,8 @@ export function checkNumberOfPoints(shapeType: ShapeType, points: number[]): voi
...
         }
     } else if (shapeType === ShapeType.MASK) {
-        const [left, top, right, bottom] = points.slice(-4);
-        const [width, height] = [right - left, bottom - top];
-        if (width < 0 || !Number.isInteger(width) || height < 0 || !Number.isInteger(height)) {
-            throw new DataError(`Mask width, height must be positive integers, but got ${width}x${height}`);
-        }
-
-        if (points.length !== width * height + 4) {
-            throw new DataError(`Points array must have length ${width}x${height} + 4, got ${points.length}`);
```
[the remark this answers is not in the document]
klakhov: Reverted first check

bsekachev (approved):

PR #7339, opened 2024-01-10
Replace keydb with kvrocks

### LINKED ISSUE #7345: OPA Helm Chart
I just want to mention the typo in the `helm-chart/values.yaml` under `opa`. I guess it was supposed to be `imagePullPolicy` instead of `imagepullploicy`, wasn't it?

azhavoro:
### Motivation and context
Resolved #7345

### Checklist

### License


helm-chart/templates/cvat_kvrocks/deployment.yml
```diff
@@ -0,0 +1,90 @@
+{{- if .Values.cvat.kvrocks.enabled }}
+apiVersion: apps/v1
+kind: Deployment
```
SpecLad, line 3: Wouldn't a StatefulSet make more sense here? I assume every replica of Kvrocks would need its own storage volume; that matches StatefulSet semantics better.

docker-compose.dev.yml
```diff
@@ -111,4 +111,4 @@ services:
 
   cvat_redis_ondisk:
     ports:
-      - '6479:6379'
+      - '6479:6666'
```
SpecLad, line 114: The "6479" number can probably also be replaced with 6666 everywhere; there's not much reason to keep it if the internal port number is different.

helm-chart/templates/cvat_kvrocks/deployment.yml
```diff
@@ -0,0 +1,90 @@
...
+    spec:
+      containers:
+        - name: cvat-kvrocks-app-container
+          image: {{ .Values.cvat.kvrocks.image }}:{{ .Values.cvat.kvrocks.tag }}
+          imagePullPolicy: {{ .Values.cvat.kvrocks.imagePullPolicy }}
+          args:
+            - --dir
+            - /var/lib/kvrocks/data
+            - --requirepass
+            - "$(CVAT_REDIS_ONDISK_PASSWORD)"
```
SpecLad, line 42: When does this variable get expanded? It is Kvrocks that does it?

helm-chart/values.yaml
```diff
@@ -222,6 +222,57 @@ cvat:
           protocol: TCP
           name: http
 
+  kvrocks:
+    enabled: true
+    external:
+      host: 127.0.0.1
```
SpecLad, line 228: This doesn't seem like a useful default... how could an external Kvrocks be located at localhost?

helm-chart/values.yaml
```diff
@@ -222,6 +222,57 @@ cvat:
...
+    existingSecret: "cvat-kvrocks-secret"
+    secret:
+      create: true
+      name: cvat-kvrocks-secret
+      password: cvat_kvrocks
+    replicas: 1
+    image: apache/kvrocks
+    tag: 2.7.0
+    imagepullploicy: IfNotPresent
```
SpecLad, line 237:
```suggestion
    imagePullPolicy: IfNotPresent
```

helm-chart/values.yaml
```diff
@@ -222,6 +222,57 @@ cvat:
...
+    labels: {}
+    #  test: test
+    annotations: {}
+    # test.io/test: test
+    resources: {}
+    affinity: {}
+    tolerations: []
+    # nodeAffinity:
```
SpecLad, line 245:
```suggestion
    nodeAffinity: {}
```

Seems like it should be consistent with the other values.

helm-chart/values.yaml
```diff
@@ -222,6 +222,57 @@ cvat:
...
+    #       - key: kubernetes.io/e2e-az-name
+    #         operator: In
+    #         values:
+    #         - e2e-az1
+    #         - e2e-az2
+    additionalEnv: []
+    # Example:
+    #  - name: volume-from-secret
+    # - name: TEST
+    #  value: "test"
```
SpecLad, line 258: The examples are misaligned. Also, the first example seems strange? There's a name, but no value, and the name doesn't feel like an environment variable name.

helm-chart/values.yaml
```diff
@@ -222,6 +222,57 @@ cvat:
...
+  kvrocks:
+    enabled: true
+    external:
+      host: 127.0.0.1
+    existingSecret: "cvat-kvrocks-secret"
+    secret:
+      create: true
+      name: cvat-kvrocks-secret
+      password: cvat_kvrocks
+    replicas: 1
```
SpecLad, line 234: Does it actually work if you set this parameter to more than one? I would assume running Kvrocks with replication is more involved than that. And if it doesn't work, then we shouldn't provide the option.

SpecLad (changes requested):

SpecLad:
Please add a changelog entry.

helm-chart/templates/cvat_kvrocks/deployment.yml
azhavoro: this var is expanded by k8s: https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#use-environment-variables-to-define-arguments

SpecLad: I see. Good to know.

helm-chart/templates/cvat_kvrocks/statefulset.yml
```diff
@@ -0,0 +1,107 @@
...
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+  {{- if .Values.cvat.kvrocks.defaultStorage.enabled }}
+  volumeClaimTemplates:
+  - metadata:
+      name: {{ .Release.Name }}-kvrocks-data
+      namespace: {{ .Release.Namespace }}
+      annotations:
+        helm.sh/resource-policy: keep
```
SpecLad, line 88: This annotation isn't needed because Helm doesn't manage volumes created by a StatefulSet. Kubernetes itself does, and by default it will retain the volume when the pod is deleted. So you don't need to do anything special to retain the data between deployments.

azhavoro: removed

SpecLad (approved):

tisonkun:
Glad to see CVAT choose Apache Kvrocks as the on-disk Redis solution.

Would you mind to add you in Kvrocks Users page? I prepare a PR for you to preview https://github.com/apache/kvrocks-website/pull/200 but your opinion is significant and perhaps you'd propose a different brand kit.

cc @SpecLad @bsekachev

tisonkun:
Hi @nmanovic! @bsekachev suggested a different logo at https://github.com/apache/kvrocks-website/pull/200#issuecomment-197258769. Please take a look if there is unaligned.

PR #7348, opened 2024-01-11
Add the ability to disable sign-in/sign-up in the UI

Marishka17:
### Checklist

### License


cvat-ui/src/reducers/server-api-reducer.ts
```diff
@@ -0,0 +1,64 @@
...
+import { ServerAPIState } from '.';
+
+const defaultState: ServerAPIState = {
+    schema: null,
+    fetching: false,
+    initialized: false,
+    isRegistrationEnabled: true,
+    isBasicLoginEnabled: true,
+    isPasswordResetEnabled: true,
+    isPasswordChangeEnabled: true,
```
klakhov, line 16: Do we really want the configuration to be true for everything by default? As far as I remember, before this PR reset/change passwords were disabled by default.

cvat-core/src/server-proxy.ts
```diff
@@ -1841,6 +1841,17 @@ async function installedApps() {
     }
 }
 
+async function getApiSchema() {
+    const { backendAPI } = config;
+
+    try {
+        const response = await Axios.get(`${backendAPI}/schema/?scheme=json`);
```
klakhov, line 1848: Now this request returns full schema. I dont really remember what we decided about this in our team discussion, but do we really need that? Looks like a lot of information is useless and can be filtered

Marishka17: We decided to use the whole schema without filtration. I believe it is better to request schema one time and then use data from the state. For example, [here](https://github.dev/opencv/cvat/blob/2b33548411b39a8e8dd60a2a72fce4ca93145adc/cvat-core/src/api-implementation.ts#L510) we need information about components. Probably we can replace /server/about endpoint with schema also.

cvat-ui/src/reducers/server-api-reducer.ts
```diff
@@ -0,0 +1,68 @@
...
+import { ServerAPIActions, ServerAPIActionTypes } from 'actions/server-actions';
+import { ServerAPIState } from '.';
+
+const defaultState: ServerAPIState = {
+    schema: null,
+    fetching: false,
+    initialized: false,
+    configuration: {
+        isRegistrationEnabled: false,
+        isBasicLoginEnabled: false,
```
klakhov, line 15: I belive those two shouldnt be disabled by default. It breaks login form if something goes wrong on the server and schema request fails:
image
It was only about change/reset

Marishka17: @klakhov, I disagree with you because I guess we should use the same default values for the endpoint's configuration. It does not matter if we cannot identify enabled login/register endpoints or reset/change password endpoints. We have enabled by default these endpoints on the server (on the public version). If we enable all endpoints by default on the client it will improve user experience if the schema request fails but it can render UI not based on the server configuration. Looks like the first option is more preferable

klakhov: Ok, I agree. Disabling form worsens UX. Anyway, the request will fail if the setting is disabled.

klakhov (approved):

cvat-core/src/api-implementation.ts
```diff
@@ -136,6 +136,11 @@ export default function implementAPI(cvat: CVATCore): CVATCore {
         return result;
     });
 
+    implementationMixin(cvat.server.apiSchema, async () => {
+        const result = await serverProxy.server.apiSchema();
+        return result;
+    });
```
bsekachev, line 142:
```suggestion
    implementationMixin(cvat.server.apiSchema, serverProxy.server.apiSchema);
```

cvat-core/src/index.ts
```diff
@@ -65,6 +65,7 @@ export default interface CVATCore {
         setAuthData: any;
         removeAuthData: any;
         installedApps: any;
+        apiSchema: any;
```
bsekachev, line 68: Will it be possible to add something more specific than `any`?

cvat-ui/src/components/cvat-app.tsx
```diff
@@ -558,6 +564,16 @@ class CVATApplication extends React.PureComponent<CVATAppProps & RouteComponentP
...
+                            {isPasswordResetEnabled && (
+                                <Route exact path='/auth/password/reset' component={ResetPasswordPageComponent} />
+                            )}
+                            {isPasswordResetEnabled && (
+                                <Route
+                                    exact
+                                    path='/auth/password/reset/confirm'
+                                    component={ResetPasswordPageConfirmComponent}
+                                />
+                            )}
```
bsekachev, line 576: These routes still rendered below (lines 577-582)

bsekachev (commented):
LGTM, just several minor comments

cvat-core/src/index.ts
bsekachev: We do not need to define precise type for the whole scheme, but at least what we should do it for things we use on client

bsekachev (approved):

egeekial:
Now that this has been merged, how can I disable sign-ups in the UI?

UUVtQD:
I have the same question - could either of you provide guidance on disabling self sign-up, or direct us to the relevant section in the documentation? @Marishka17 @bsekachev 
Thanks!

davidblom603:
Would also like to know how to disable the self sign up

Marishka17:
Hello, @egeekial, @UUVtQD, @davidblom603, you can just comment out the corresponding endpoint [on the server](https://github.com/cvat-ai/cvat/blob/efbafe9a8574a8a5bbf0477f22cb9605f2a27346/cvat/apps/iam/urls.py#L31) and rebuild CVAT from the source (`docker compose -f docker-compose.yml -f docker-compose.dev.yml build`).

davidblom603:
Thanks! Works great.

0leynik:
@Marishka17 @bsekachev Thanks for the work!

@egeekial, @UUVtQD, @davidblom603
You can disable registration and password reset.
Replace the code with the code below:
1. https://github.com/cvat-ai/cvat/blob/952f7af8365c0348fba274ecaa5342d2683941db/cvat-ui/src/reducers/server-api-reducer.ts#L14-L17
```
isRegistrationEnabled: false,
isBasicLoginEnabled: true,
isPasswordResetEnabled: false,
isPasswordChangeEnabled: true,
```
2. https://github.com/cvat-ai/cvat/blob/952f7af8365c0348fba274ecaa5342d2683941db/cvat-ui/src/reducers/server-api-reducer.ts#L35-L38
```
const isRegistrationEnabled = false;
const isBasicLoginEnabled = Object.keys(schema.paths).includes('/api/auth/login');
const isPasswordResetEnabled = false;
const isPasswordChangeEnabled = Object.keys(schema.paths).includes('/api/auth/password/change');
```

PR #7416, opened 2024-01-31
Drop support of the TFRecord format

SpecLad:
### Motivation and context

Usage statistics on app.cvat.ai show that this format is rarely used, with single-digit numbers of projects/tasks/jobs being exported or imported using this format. Moreover, TensorFlow's popularity appears to be shrinking, so I'm not expecting this format to make a comeback.

Meanwhile, supporting this format has a cost that has to be borne by everyone deploying CVAT, because it requires TensorFlow to be installed and loaded. This has various ill effects:

1. Loading time is increased. In my testing, even a command as trivial as `manage.py --help` is slowed down by 3.6 seconds. This may not seem like much, but the effect is compounded, because we have multiple processes (server + workers) all loading the same codebase. Plus, the container entrypoint may execute several Django commands.

2. Memory usage is increased. TensorFlow adds ~100MB of RAM usage per process with data alone; and the libraries add more (although it's hard to estimate the impact of library code, since it can be shared between processes in RAM).

3. Docker image size is increased by ~1.5GB (when unpacked). This is more than half of the current total size! Building time is increased as well.

Overall, it seems that the drawbacks of keeping support for this format outweigh the benefits, so it's time to drop it.

### Checklist

### License


azhavoro:
LGTM, could you resolve the conflict?

SpecLad:
> LGTM, could you resolve the conflict?

Resolved.

azhavoro (approved):

PR #7450, opened 2024-02-07
Don't save all fields of a model when bumping the update date

SpecLad:
### Motivation and context

Currently, there are a lot of places where changing one object requires updating the `updated_date` field on other objects. Usually, the `.save` method is used for this purpose, which results in a generic SQL `UPDATE` query that replaces the values in all columns.

Introduce and use a specialized method that only changes the date. This provides the following advantages:

* Prevents the date updates from overwriting updates to other fields that may happen at the same time.

* Possibly makes the updates more efficient due to fewer columns being updated (I have no data on this, though).

* Makes it easier to recognize the purpose of the query when analyzing logs/live statistics.

* Makes the purpose of the code more obvious.

Additionally, update `__save_job_handler`, where we don't want to save _just_ the date, but we still only need to save two fields.

### How has this been tested?

I verified that when you save the annotations for a job, only the `updated_date` field of the task is modified.

### Checklist

### License

### DIFF
--- a/cvat/apps/dataset_manager/task.py
+++ b/cvat/apps/dataset_manager/task.py
@@ -12,7 +12,6 @@ from datumaro.components.errors import DatasetError, DatasetImportError, Dataset
 
 from django.db import transaction
 from django.db.models.query import Prefetch
-from django.utils import timezone
 from rest_framework.exceptions import ValidationError
 
 from cvat.apps.engine import models, serializers
@@ -389,9 +388,7 @@ class JobAnnotation:
         self.ir_data.tags = tags
 
     def _set_updated_date(self):
-        db_task = self.db_job.segment.task
-        db_task.updated_date = timezone.now()
-        db_task.save()
+        self.db_job.segment.task.touch()
 
     def _save_to_db(self, data):
         self.reset()
@@ -404,7 +401,7 @@ class JobAnnotation:
     def _create(self, data):
         if self._save_to_db(data):
             self._set_updated_date()
-            self.db_job.save()
+            self.db_job.touch()
 
     def create(self, data):
         self._create(data)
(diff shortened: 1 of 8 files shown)


cvat/apps/engine/models.py
```diff
@@ -316,15 +316,23 @@ class Image(models.Model):
     class Meta:
         default_permissions = ()
 
-class Project(models.Model):
+class TimestampedModel(models.Model):
+    created_date = models.DateTimeField(auto_now_add=True)
+    updated_date = models.DateTimeField(auto_now=True)
```
Marishka17, line 321: 1. Why aren't we going to update base classes for several models in other applications (that have date fields)?
1. Will it be better to separate this class into 2 with `created_date` and `updated_date` + `touch`? Otherwise, now we have some models that are not inherited from `TimestampedModel` but have the same custom `created_date` field.

cvat/apps/dataset_manager/task.py
```diff
@@ -389,9 +388,7 @@ def _save_tags_to_db(self, tags):
         self.ir_data.tags = tags
 
     def _set_updated_date(self):
-        db_task = self.db_job.segment.task
-        db_task.updated_date = timezone.now()
-        db_task.save()
+        self.db_job.segment.task.touch()
```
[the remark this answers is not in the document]
SpecLad: I think the fact that it only touched the task was a bug. I updated the code to make it touch the job as well (or rather I moved that call from `_create` to here).

cvat/apps/engine/models.py
SpecLad: > Why aren't we going to update base classes for several models in other applications (that have date fields)?

I think I just forgot to look in other apps. Thanks for reminding me; updated.

> Will it be better to separate this class into 2 with `created_date` and `updated_date` + `touch`?

I guess in principle it would be better, but I think this refactoring would go beyond the scope of what I'm trying to do in this PR. If in the future I have to add some shared functionality involving only `created_date`, I'll consider something like this.

Marishka17 (approved):

PR #7460, opened 2024-02-08
Defer most signal handling code until after the end of the current transaction

SpecLad:
### Motivation and context

Non-database actions like deleting directories, sending webhooks, scheduling reports should only be done after the current transaction is committed. If we do it immediately, and the transaction is later aborted, then we will have (for example) sent a webhook about an event that didn't actually happen.

There's also a secondary benefit to moving this action outside of the transaction; the less time we spend inside a transaction, the better, since a transaction may lock out other clients from working on the affected DB rows.

In addition, prevent the affected actions from crashing the view handler with an exception (using the `robust=True` option). I don't think it's reasonable to (for example) return a 500 response to a `PATCH` request just because we failed to send the corresponding webhook.

There is one more type of action that should be modified in this way (sending events), but it would be easier to do that after a refactoring that I did in another patch, so I'll do it later.

### Checklist

### License

### DIFF
--- a/cvat/apps/quality_control/signals.py
+++ b/cvat/apps/quality_control/signals.py
@@ -2,6 +2,7 @@
 #
 # SPDX-License-Identifier: MIT
 
+from django.db import transaction
 from django.db.models.signals import post_save
 from django.dispatch import receiver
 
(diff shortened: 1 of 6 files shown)


cvat/apps/quality_control/signals.py
```diff
@@ -37,8 +38,15 @@ def __save_job__update_quality_metrics(instance, created, **kwargs):
     else:
         assert False
 
-    for task in tasks:
-        qc.QualityReportUpdateManager().schedule_quality_autoupdate_job(task)
+    def schedule_autoupdate_jobs():
+        for task in tasks:
+            if task.id is None:
+                # The task may have been deleted after the on_commit call.
+                continue
```
Marishka17, line 45: Why are you sure that instance from the list will be actual after deleting the row in the database? I've checked that e.g. after updating a project with several tasks in one transaction and deleting one project task after closing the first transaction -> related task instance from the list has "outdated" properties. Probably you need to check directly that an object exists in db.

SpecLad: My goal here isn't really to _ensure_ that the task exists in the database. It's not possible to do that anyhow, since we're not in a transaction, and another request could delete the task right after we check it.

Rather, I just want to make sure that `task.id` is not `None` (even if it's outdated), because `schedule_quality_autoupdate_job` references it. I'm not sure what'll happen if `schedule_quality_autoupdate_job` encounters an ID of `None`, and I can't assume that it will crash safely, so I'd rather prevent this from happening.

Marishka17 (approved):
LGTM

PR #7466, opened 2024-02-09
Upgrade React and Antd till the latest version

bsekachev:
### Motivation and context

- [x] Annotation menu
- [x] OpenCV is not loaded automatically
- [x] Eslint
- [x] Stylelint
- [x] Tests

### Checklist

### License


cvat-ui/src/components/actions-menu/styles.scss
```diff
@@ -1,22 +1,9 @@
...
-    > li:hover {
-        background-color: $hover-menu-color;
-    }
-
-    .ant-menu-submenu-title {
-        margin: 0;
-        width: 13em;
-    }
-}
+@import '../../base';
```
klakhov, line 6:
```suggestion
@import 'base';
```

cvat-ui/src/components/actions-menu/styles.scss
```diff
@@ -1,22 +1,9 @@
...
 
 .cvat-menu-dump-submenu-item,
```
klakhov, line 8: I cant find usage of this classes in our code now

cvat-ui/src/components/analytics-page/styles.scss
```diff
@@ -129,7 +130,7 @@
     padding: 6px 8px;
     border: 1px solid lightgray;
     margin-left: 8px;
```
klakhov, line 132:
```suggestion
    margin-left: $grid-unit-size;
```

klakhov (commented):
Sidepanel became white:
image

klakhov (commented):
There is something wrong with initial position of opencv tool popup 
opencvplacement

bsekachev:
@klakhov  Applied

klakhov (commented):
Exclamation mark and excessive left padding now appear on organization switch modal. Also I cant close it without selecting something new
image

Do we really want to use blue hover effect? I think it doesn't stand out enough on gray background. Maybe we can use scaling effect instead.
image

Here I suppose using bold black font would be better instead of blue too. 
image

If we dont have an assignee, maybe we dont want to show the line with it at all? Or at least say something like not assigned so we dont have emty space.
I like the idea of showing stage+state in one status row like `annotation in progress` or `annotation rejected`, it looks cool. But it looks strange with `new`, maybe we should use an alias for that something like `annotation not started`. 
image

Small padding problem on models page(it seems the same problem is on the projects page):
image

Major version update for package.json?

bsekachev:
> Exclamation mark and excessive left padding now appear on organization switch modal

I believe this is general antd change: 
image

> Major version update for package.json?

I do not see reasons to update major version. Generally all the changes are cosmetic.

bsekachev:
>  Also I cant close it without selecting something new

Applied

> Do we really want to use blue hover effect?

Default antd button, I reversed it as it was with custom css.

> Here I suppose using bold black font would be better instead of blue too.

The same.

 > If we dont have an assignee, maybe we dont want to show the line with it at all? Or at least say something like not assigned so we dont have emty space.

The main idea is to keep height the same. Sliding bottom part is annoying a little bit. 

> I like the idea of showing stage+state in one status row like annotation in progress or annotation rejected, it looks cool. But it looks strange with new, maybe we should use an alias for that something like annotation not started.

The main idea behind it was to keep vertical space, using horizontal free space. I would avoid customizing logic in described way.

> Small padding problem on models page(it seems the same problem is on the projects page):

Fixed on projects/jobs/models. Also fixed top bars on different resolutions.

klakhov (approved):
Looks good!

PR #7486, opened 2024-02-19
Single object annotation mode

bsekachev:
### Motivation and context
image

Added query parameter to open specific mode:
`?defaultWorkspace=standard`
`?defaultWorkspace=review`
`?defaultWorkspace=tags`
`?defaultWorkspace=single_shape`

Query parameters to setup label and number of points in single shape mode:
`?defaultLabel=label_name`
`?defaultPointsCount=1`

Additional change: canvas will not trigger unnecessary `canvas.canceled` events when something is done (drawing, merging, etc).
Only two things should trigger such events: explicit canvas API call and when something is wrong happened during drawing (e.g. drawn object is not correct)

### Checklist

### License


klakhov:
image

cvat-canvas/src/typescript/drawHandler.ts
```diff
@@ -467,11 +473,13 @@ export class DrawHandlerImpl implements DrawHandler {
         };
 
         this.canvas.on('mousedown.draw', (e: MouseEvent): void => {
-            if (initialPoint.x === null || initialPoint.y === null) {
-                const translated = translateToSVG(this.canvas.node as any as SVGSVGElement, [e.clientX, e.clientY]);
-                [initialPoint.x, initialPoint.y] = translated;
-            } else {
-                this.drawInstance.fire('drawstop');
+            if (e.button === 0 && !e.altKey) {
```
klakhov, line 476: Why alt key here?

cvat-canvas/src/typescript/masksHandler.ts
```diff
@@ -627,6 +627,8 @@ export class MasksHandlerImpl implements MasksHandler {
                     enabled: true,
                     shapeType: 'mask',
                 };
+
+                this.onDrawRepeat({ enabled: true, shapeType: 'mask' });
```
klakhov, line 631: Why do we call it twice? isnt second call already contains data with enabled and mask shape type?

cvat-ui/src/components/annotation-page/single-shape-workspace/single-shape-sidebar/single-shape-sidebar.tsx
```diff
@@ -0,0 +1,481 @@
...
+        pointsCountIsPredefined: true,
+        pointsCount: defaultPointsCount || 1,
+        labels: jobInstance.labels.filter((label) => label.type !== LabelType.TAG && label.type !== LabelType.SKELETON),
+        label: null,
+        labelType: jobInstance.labels[0]?.type || LabelType.ANY,
+    });
+
+    const savingRef = useRef(false);
+    const nextFrame = useCallback((): void => {
+        (frame < jobInstance.stopFrame ? jobInstance.annotations.search(frame + 1, jobInstance.stopFrame, {
```
klakhov, line 163: May we rewrite this with ordinary `if`? Ternary operator with length of 20 lines is quite hard to read

cvat-ui/src/components/annotation-page/top-bar/player-navigation.tsx
```diff
@@ -85,18 +91,52 @@ function PlayerNavigation(props: Props): JSX.Element {
...
+    const disabledStyle: CSSProperties = {
+        pointerEvents: 'none',
+        opacity: 0.5,
+    };
+
     const deleteFrameIcon = !frameDeleted ? (
         <CVATTooltip title={`Delete the frame ${deleteFrameShortcut}`}>
-            <DeleteOutlined className='cvat-player-delete-frame' onClick={showDeleteFrameDialog} />
+            <DeleteOutlined
+                style={workspace === Workspace.SINGLE_SHAPE ? disabledStyle : {}}
```
klakhov, line 121: Lets create a separate variable instead of checking `workspace === Workspace.SINGLE_SHAPE` every time

cvat-ui/src/components/annotation-page/top-bar/player-buttons.tsx
```diff
@@ -92,23 +141,39 @@ function PlayerButtons(props: Props): JSX.Element {
...
+        opacity: 0.5,
+    };
+
     return (
         <Col className='cvat-player-buttons'>
+            <GlobalHotKeys keyMap={subKeyMap} handlers={handlers} />
             <CVATTooltip title='Go to the first frame'>
-                <Icon className='cvat-player-first-button' component={FirstIcon} onClick={onFirstFrame} />
+                <Icon
+                    style={workspace === Workspace.SINGLE_SHAPE ? disabledStyle : {}}
```
klakhov, line 164: The same thing with `workspace === Workspace.SINGLE_SHAPE` in this file

klakhov:
- The background color of the sidebar is not consistent with our other sidebars, I will suggest to make it gray (maybe with border)
- If its Single shape mode, maybe we shouldnt allow users to create more that one shape if someone decided to stay on the page?
- It looks there is no Next button anymore, so it may be confusing what to press. We need to clarify the message about navigation panel and add a next(and maybe prev) frames shortcut hints.
image

cvat-canvas/src/typescript/drawHandler.ts
bsekachev: As for other shapes, when we hold alt, points are not set when clicking canvas (so, you may drag the canvas in drawing mode, for example)

cvat-canvas/src/typescript/masksHandler.ts
bsekachev: In case with masks no 
For masks first call initiates drawing and second is necessary to setup brush tools

bsekachev:
> The background color of the sidebar is not consistent with our other sidebars, I will suggest to make it gray (maybe with border)

We do not have tips in other modes. So, what consistency are you speaking about? It is a default color for Alert and it will be inconsistent to make it gray. 

> If its Single shape mode, maybe we shouldnt allow users to create more that one shape if someone decided to stay on the page?

We decided do not set strong restrictions. Finally, users may do what they want. Including annotating many bounding boxes on one frame if this mode is convenient for them because it automatically restarts drawing process. 

> It looks there is no Next button anymore, so it may be confusing what to press. We need to clarify the message about navigation panel and add a next(and maybe prev) frames shortcut hints.

I removed these buttons to not create custom components for navigation. Navigation is possible with default shortcuts that described in shortcuts list. Moreover, by default users do not need even navigate because auto next frame is on.

cvat-ui/src/components/annotation-page/single-shape-workspace/single-shape-sidebar/single-shape-sidebar.tsx
bsekachev: Done

cvat-ui/src/components/annotation-page/top-bar/player-buttons.tsx
bsekachev: Done

cvat-ui/src/components/annotation-page/top-bar/player-navigation.tsx
bsekachev: Done

zhiltsov-max (approved):

klakhov (approved):

PR #7493, opened 2024-02-19
Disable authentication for public endpoints

Marishka17:
### Motivation and context

We have several endpoints that should be available for everyone. Despite this with the current implementation (empty permissions + enabled authentication for these endpoints) it is possible to receive 401 responses if some invalid token is specified in the header. This happens because authentication is attempted [before](https://github.dev/encode/django-rest-framework/blob/2ef77b18335322f10c8c6160b6f7794ff6bc92b7/rest_framework/views.py#L414-L416) checking permissions initially. So, e.g. when a header with an expired token is specified, TokenAuthentication throws an exception during authentication. This PR fixes the described issue.
### How has this been tested?

Manually
### Checklist

### License

### DIFF
--- a/cvat/apps/engine/urls.py
+++ b/cvat/apps/engine/urls.py
@@ -32,15 +32,18 @@ urlpatterns = [
 
     # documentation for API
     path('api/schema/', SpectacularAPIView.as_view(
-        permission_classes=[] # This endpoint is available for everyone
+        permission_classes=[], # This endpoint is available for everyone
+        authentication_classes=[],
     ), name='schema'),
     path('api/swagger/', SpectacularSwaggerView.as_view(
         url_name='schema',
-        permission_classes=[] # This endpoint is available for everyone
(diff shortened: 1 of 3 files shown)


SpecLad:
I suppose this will fix it, but why is it even an issue? Does this create problems for users?

Also, I think this will prevent requests for these endpoints from being correctly associated with users who made them, which would affect, for example, the rate limiting logic.

Marishka17:
> Does this create problems for users

Yes, to some extent, the `/api/schema` endpoint is now used to render the UI login page correctly. If for some reason invalid/expired tokens are kept in local storage (e.g. a user logs in 2 browsers -> logs out in the first one -> refreshes a page in the second one), then a user receives an exception with 401 code on `api/schema` request, and UI components are rendered not according to server configuration.

IMHO, such an approach is more appropriate than removing auth data and sending `api/schema` request one more time when the first request fails.

PR #7537, opened 2024-02-29
Requests page

klakhov:
### Checklist

### License

### Release Notes

- **New Features**
  - Enhanced progress tracking for dataset export and backup operations.
  - Improved import dataset progress visibility with dynamic messages.

- **Bug Fixes**
  - Fixed navigation flow in multiple Cypress test scripts by adding `cy.goBack()` after export operations.
  - Corrected notification handling in various test cases.

- **Refactor**
  - Removed unused state properties and hooks related to `importing` and `backupIsActive` across multiple components.
  - Restructured state management in reducers for export and import actions.

- **Tests**
  - Updated Cypress commands for downloading exports and verifying downloads.
  - Added new commands for better test flow management in Cypress scripts.

- **Documentation**
  - Updated copyright year in various components.

- **Chores**
  - Added new error handling and notification messages for request-related actions.


cvat/apps/engine/serializers.py
```diff
@@ -2170,3 +2177,120 @@ def _update_assets(guide):
...
+    operation = RQJobOperationSerializer(source="*")
+    percent = serializers.SerializerMethodField()
+    enqueue_date = serializers.DateTimeField(source="enqueued_at", read_only=True)
+    start_date = serializers.DateTimeField(
+        required=False, allow_null=True, source="started_at", read_only=True
+    )
+    finished_date = serializers.DateTimeField(
+        required=False, allow_null=True, source="ended_at", read_only=True
+    )
+    expire_date = serializers.SerializerMethodField()
```
zhiltsov-max, line 2232:
```suggestion
    expiry_date = serializers.SerializerMethodField()
```
Or `expiration`, both would be more correct from the language viewpoint.

cvat/apps/engine/views.py
```diff
@@ -3156,3 +3182,192 @@ def _import_project_dataset(request, rq_id_template, rq_func, db_obj, format_nam
...
 
     return Response(serializer.data, status=status.HTTP_202_ACCEPTED)
+
+@extend_schema(tags=['requests'])
+@extend_schema_view(
+    list=extend_schema(
+        summary='List RQ jobs',
+        parameters=[
+            OpenApiParameter(name='task_id', location=OpenApiParameter.QUERY, type=OpenApiTypes.INT),
+            OpenApiParameter(name='action', location=OpenApiParameter.QUERY, type=OpenApiTypes.STR, enum=['create']),
```
zhiltsov-max, line 3192: This list of filters looks incomplete, comparing with the supported targets and actions. What do you think about adding filters like `project_id`, `job_id`, and other actions and subjects (e.g. backup, ...)? Also, what do you think about adding `status`, and time filters, sorting?

zhiltsov-max (commented):
@Marishka17, I think, it makes sense to add named constants or use other means of encapsulation for shared strings, such as `meta['user']`, actions, etc. Currently, the fact that they are the same in different jobs is coincidental.

cvat/apps/engine/urls.py
```diff
@@ -24,6 +24,7 @@
 router.register('cloudstorages', views.CloudStorageViewSet)
 router.register('assets', views.AssetsViewSet)
 router.register('guides', views.AnnotationGuidesViewSet)
+router.register('requests', views.RequestViewSet,  basename="request")
```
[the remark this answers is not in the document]
Marishka17: I'm not sure that I understand which pattern you mean. Are you talking about using plural instead of singular? In that case, looks like the standard pattern is `ObjectViewSet`..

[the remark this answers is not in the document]
zhiltsov-max: Yes, I meant changing it to `RequestsViewSet`, but it seems it was my bad.

cvat/apps/engine/views.py
```diff
@@ -3156,3 +3182,192 @@ def _import_project_dataset(request, rq_id_template, rq_func, db_obj, format_nam
...
+        if not rq_job.is_started:
+            return HttpResponseBadRequest(f"Only started RQ jobs can be cancelled")
+
+        rq_job.cancel(enqueue_dependents=settings.ONE_RUNNING_JOB_IN_QUEUE_PER_USER)
+
+        return Response(status=status.HTTP_200_OK)
+
+    @method_decorator(never_cache)
+    @_handle_redis_exceptions
+    def destroy(self, request, pk: str):
```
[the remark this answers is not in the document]
Marishka17: I've added this endpoint just because without it user cannot "directly" delete finished/failed jobs. That means that after some time the functionality for managing requests can become useless because it will require iterating over a huge amount of jobs and finding the required one.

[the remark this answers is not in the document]
zhiltsov-max: Well, typically, we use filters and sorting for this. Which are suggested in another comment.

cvat/schema.yml
```diff
@@ -9222,6 +9354,107 @@ components:
...
+        percent:
+          type: number
+          format: double
+          readOnly: true
+          nullable: true
+        enqueue_date:
+          type: string
+          format: date-time
+          readOnly: true
+        start_date:
```
SpecLad, line 9381: `started_date` would be more consistent.

cvat/schema.yml
```diff
@@ -9222,6 +9354,107 @@ components:
...
+          type: string
+          format: date-time
+          readOnly: true
+          nullable: true
+        finished_date:
+          type: string
+          format: date-time
+          readOnly: true
+          nullable: true
+        expire_date:
```
SpecLad, line 9391:
```suggestion
        expires_date:
```

cvat/schema.yml
```diff
@@ -9222,6 +9354,107 @@ components:
...
+          allOf:
+          - $ref: '#/components/schemas/UserIdentifiers'
+          readOnly: true
+        result_url:
+          type: string
+          format: uri
+          readOnly: true
+          nullable: true
+      required:
+      - operation
```
SpecLad, line 9406: This list seems a bit barren. What about `id`, `status`, `enqueue_date`, `owner` - shouldn't these be required too?

cvat/schema.yml
```diff
@@ -9222,6 +9354,107 @@ components:
...
+    RQJobDetailsStatusEnum:
+      enum:
+      - queued
+      - started
+      - failed
+      - finished
```
SpecLad, line 9412: If the intent here is for this to be the opposite of `failed`, then `succeeded` would be a more natural choice.

cvat/schema.yml
```diff
@@ -9222,6 +9354,107 @@ components:
...
+          nullable: true
+        job_id:
+          type: integer
+          readOnly: true
+          nullable: true
+        format:
+          type: string
+          readOnly: true
+          nullable: true
+        name:
```
SpecLad, line 9445: What's this supposed to be?

cvat/schema.yml
```diff
@@ -9222,6 +9354,107 @@ components:
...
+        * `started` - started
+        * `failed` - failed
+        * `finished` - finished
+    RQJobOperation:
+      type: object
+      properties:
+        type:
+          type: string
+          readOnly: true
+        target:
```
SpecLad, line 9425: Is this field really necessary? Seems like it should be possible to get the same information from the `id`s that are present.

cvat/schema.yml
```diff
@@ -5403,6 +5511,9 @@ paths:
   /api/tasks/{id}/status:
     get:
       operationId: tasks_retrieve_status
+      description: 'This method is deprecated and will be removed in version 2.14.0.
```
SpecLad, line 5514: If this is deprecated, how about updating the SDK so that it doesn't use it anymore?

cvat/schema.yml
```diff
@@ -9222,6 +9354,107 @@ components:
...
+          type: string
+          readOnly: true
+        operation:
+          $ref: '#/components/schemas/RQJobOperation'
+        percent:
+          type: number
+          format: double
+          readOnly: true
+          nullable: true
+        enqueue_date:
```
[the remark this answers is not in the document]
Marishka17: I have just taken the API interface which was discussed and approved on HLD

cvat/schema.yml
Marishka17: No, the intent here is to provide which rq statuses are supported by the server

cvat/schema.yml
Marishka17: This param will be removed, it was presented in the original API design, but I was waiting for the client update to stop using it.

cvat/schema.yml
Marishka17: I agree with you that it is kind of sending duplicated data. The same applies to the `type` field.

cvat/schema.yml
```diff
@@ -4281,6 +4281,114 @@ paths:
...
+    get:
+      operationId: requests_list
+      summary: List RQ jobs
+      parameters:
+      - in: query
+        name: action
+        schema:
+          type: string
+          enum:
+          - create
```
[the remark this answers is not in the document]
zhiltsov-max: I think it's a filter

klakhov:
@Marishka17  
Could you check permissions for the lambda models please? It seems UI tests are failiing because of consistent error `could not get models from the server`.

cvat/schema.yml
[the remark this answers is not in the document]
bsekachev: I will vote for `enqueued_date`.
In django admin we have `started/queued`. I am not sure this is the same time in case with RQ jobs. 
image

bsekachev:
@klakhov 

- Please, merge the develop branch.
- I can't create any task. The server was restarted. `/requests/` endpoint works. Did I miss something additional? 
image

bsekachev:
The same thing on `/tasks` page. Requests are with code 200. 

image

bsekachev:
There is seven elements on the first page and one element on the second
Why was it implemented this way? Why not 10 per page? Empty page at the bottom allows to have at least one element more even without scrolling 

image
image

bsekachev:
- I can't see automatic annotation request on requests page
image

bsekachev:
When backup a task, first it says me that downloading will start automatically.
After that it suggest me to open another page to start downloading. Why?  

The style of these two notifications is different (font size, i mean) 

image

bsekachev:
Additionally, I will recommend to start downloading automatically whenever it is possible. 
Without necessity to open another page and click something else.

bsekachev:
Do we really store backup only 12 minutes?

image


All started/enqueued are equal to page refresh time:
image

bsekachev:
Why does it suggest me to DELETE if the method is not allowed? 

image

Marishka17:
> Why does it suggest me to DELETE if the method is not allowed?

@bsekachev, I removed DELETE endpoint on the server, but forgot to remove corresponding action on UI. @klakhov, API has been changed on the server a little bit, as was suggested in the comments above, could you please check that and adapt UI?

Marishka17:
> I can't see automatic annotation request on requests page

This was not mentioned in the HLD as will be implemented. Right now server handles only import/export queues. I can add automatic annotation queue after vacation.

bsekachev:
@Marishka17 

Alright. Current HLD only describes the first iteration. 
But in this case, I expect @klakhov to track the following iterations of the feature and prepare new updated HLD.

bsekachev:
@klakhov 

Too many requests. We can't afford this patch on production server. 
It sends requests to often (each 2 seconds...?? )

image

Period should be more smart. For example there is not any reason to ask often the status if rq job is deferred or queued. 
If progress increases slowly, we also may increase pooling period. 
Please, think how to implement it in better way.

bsekachev:
It did not update status on `Requests` page.
All these tasks are already done (And I can see them to be done in a new tab, please see preview of another tab):
image

Request also has shown that the process has finished (see task id and corresponding line on `requests` page):
image

bsekachev:
What does expire for Import annotations job? 
image

bsekachev:
Why do we duplicate any exception messages?
image

bsekachev:
Long exception has to be scrollable. 
image

I believe the height of the card should be fixed.

bsekachev:
Changelog entry?

cvat-core/src/requests-manager.ts
```diff
@@ -0,0 +1,278 @@
...
+        if (storedID in this.listening) {
+            if (callback) {
+                this.listening[storedID].onUpdate.push(callback);
+            }
+            return this.listening[storedID].promise;
+        }
+        const promise = new Promise<Request>((resolve, reject) => {
+            const timeoutCallback = (): void => {
+                serverProxy.requests.status(id, params).then((response) => {
+                    const request = new Request({ ...response });
```
bsekachev, line 202: Why we create `Request` instance from a variable with name `response`?

bsekachev:
After refreshing a page, I may request exporting at the same format again.

Before:
image

After:
image

We need to come to consistent solution (do not disable these items at all OR disable them even if the page has been refreshed).

cvat-core/src/project-implementation.ts
```diff
@@ -96,9 +98,14 @@
...
+        options?: { requestStatusCallback?: (request: Request) => void },
     ) {
-        const result = exportDataset(this, format, saveImages, useDefaultSettings, targetStorage, customName);
-        return result;
+        const { requestStatusCallback } = options || {};
+        const rqID = await exportDataset(this, format, saveImages, useDefaultSettings, targetStorage, customName);
+        if (rqID) {
+            return requestsManager.listen(rqID, { callback: requestStatusCallback });
+        }
+        return new Request({ status: RQStatus.FINISHED, message: '' });
```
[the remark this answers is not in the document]
klakhov: We dont want to use ternary operator here as the line is hard to read

cvat-core/src/server-proxy.ts
```diff
@@ -2429,6 +2246,54 @@
...
+    const { backendAPI } = config;
+
+    try {
+        const response = await fetchAll(`${backendAPI}/requests`);
+
+        return response.results;
+    } catch (errorData) {
+        throw generateError(errorData);
+    }
+}
```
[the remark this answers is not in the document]
klakhov: We dont need additional information here

klakhov:
For UI part:
- Task status requests were fixed, now task creation displays normally
- Imroved amount of requests on the page, now it looks like tasks list(10 per page with scroll)
- Fixed font size on notifications
- Fixed timestamps on the requests cards + hidden the expiration time for tasks create and import
- Removed delete option
- Improved status requests delay system (now only 5 consecutive requests are allowed, others will be re-scheduled)
- Improved error display for the card, added scrollable element if its too large
- Disabled requesting export if its already started(as it was before)
- As for downloading automatically, as I remember we discussed on HLD that we should download only by button click

bsekachev:
I ran several processes as on the screen, than pressed F5 to refresh the page. After refresh: 
image

bsekachev:
image

1. Download once
2. Wait for a 30 seconds / one minute
3. Try to download one more time

It runs new export: 
image

cvat-core/src/annotations.ts
```diff
@@ -161,20 +162,23 @@ export function importDataset(
     file: File | string,
     options: {
         convMaskToPoly?: boolean,
-        updateStatusCallback?: (s: string, n: number) => void,
+        requestStatusCallback?: (request: Request) => void,
+        uploadStatusCallback?:(s: string, n: number) => void,
```
bsekachev, line 166:
```suggestion
        uploadStatusCallback?: (s: string, n: number) => void,
```

cvat-core/src/annotations.ts
```diff
@@ -161,20 +162,23 @@ export function importDataset(
...
     } = {},
-): Promise<void> {
-    const updateStatusCallback = options.updateStatusCallback || (() => {});
+): Promise<string> {
+    const uploadStatusCallback = options.uploadStatusCallback || (() => {});
+    const requestStatusCallback = options.requestStatusCallback || (() => {});
     const convMaskToPoly = 'convMaskToPoly' in options ? options.convMaskToPoly : true;
     const adjustedOptions = {
-        updateStatusCallback,
+        requestStatusCallback,
```
bsekachev, line 173: We need a check for this new argument like for `uploadStatusCallback`
Message should be more specific if we have many callbacks now:

``throw new ArgumentError('Callback must be a function');`` ->  ``throw new ArgumentError('Request status callback must be a function');``

cvat-core/src/api.ts
```diff
@@ -374,6 +375,27 @@ function build(): CVATCore {
                 },
             },
         },
+        requests: {
+            async list(): Promise<{
+                requests: Request[];
+                count: number;
+            }> {
```
bsekachev, line 382: For the most of resources we have returning lists like: PaginatedResource<Resource> where `count` is an additional property of returning array.

But for lambda manager and request manager we use another approach.

I believe it should be unified.

cvat-core/src/enums.ts
```diff
@@ -50,6 +50,7 @@ export enum RQStatus {
     FINISHED = 'finished',
     FAILED = 'failed',
     UNKNOWN = 'unknown',
+    DEFERRED = 'deferred',
```
bsekachev, line 53: Did you test it?

I have several deferred requests, but they are returned as queued
image

cvat-core/src/index.ts
```diff
@@ -150,6 +152,19 @@ export default interface CVATCore {
     frames: {
         getMeta: any;
     };
+    requests: {
+        list: () => Promise<{ requests: Request[], count: number }>;
+        listen: (
+            rqID: string | null,
```
bsekachev, line 158: What does listening for `null` request mean?

cvat-core/src/index.ts
```diff
@@ -150,6 +152,19 @@ export default interface CVATCore {
...
+            options?: {
+                callback?: (request: Request) => void,
```
bsekachev, line 160: Shall this callback be optional? What is a reason to listen for something if we do not want to get updates?

cvat-core/src/index.ts
```diff
@@ -150,6 +152,19 @@ export default interface CVATCore {
...
+                filter?: RequestsFilter,
```
bsekachev, line 161: What are you going to filter if you listen for a specific request?

cvat-core/src/project-implementation.ts
```diff
@@ -96,9 +98,14 @@ export default function implementProject(projectClass) {
...
```
bsekachev, line 108: I believe the following approach is more correct:

On `cvat-core` level we do not want to run implicit listening inside `exportDataset`, `importDataset`, `exportBackup`, `importBackup`. These methods should return `Request` instance. If `cvat-ui` wants to listen for them, it calls `listen` explicitly. This way we will get rid of new options, like `requestStatusCallback` and it will make code easier. 

Need to think additionally about task creating. For simplicity, my basic idea is to split between `create` and `save`  (as the first one is worker process, the second one is quicker):
- `create` returns `Request` instance.
- `save` returns `Task` instance as before.

It is not unified with for example Project.save or with Issue.save. But maybe we do not need this unification as these processes are not performed in Workers.

bsekachev: For now, never mind, let's discuss after full code review, when I have mode understanding

cvat-core/src/enums.ts
Marishka17: > I have several deferred requests, but they are returned as queued

This is consciously changed on the server, a user shouldn't know that his job is deferred. For him job is just in a queue.

cvat-core/src/project-implementation.ts
```diff
@@ -96,9 +98,14 @@ export default function implementProject(projectClass) {
         useDefaultSettings: boolean,
         targetStorage: Storage,
         customName?: string,
+        options?: { requestStatusCallback?: (request: Request) => void },
     ) {
-        const result = exportDataset(this, format, saveImages, useDefaultSettings, targetStorage, customName);
-        return result;
+        const { requestStatusCallback } = options || {};
```
bsekachev, line 103: `requestStatusCallback` type is not checked

cvat-core/src/project-implementation.ts
```diff
@@ -107,24 +114,37 @@ export default function implementProject(projectClass) {
         file: File | string,
         options?: {
             convMaskToPoly?: boolean,
-            updateStatusCallback?: (s: string, n: number) => void,
+            uploadStatusCallback?: (s: string, n: number) => void,
+            requestStatusCallback?: (request: Request) => void,
         },
     ) {
-        return importDataset(this, format, useDefaultSettings, sourceStorage, file, options);
+        const { requestStatusCallback } = options || {};
```
bsekachev, line 121: `requestStatusCallback` type is not checked

cvat-core/src/project-implementation.ts
```diff
@@ -107,24 +114,37 @@ export default function implementProject(projectClass) {
...
 
     projectClass.prototype.backup.implementation = async function (
         targetStorage: Storage,
         useDefaultSettings: boolean,
         fileName?: string,
+        options?: { requestStatusCallback?: (request: Request) => void },
     ) {
-        const result = await serverProxy.projects.backup(this.id, targetStorage, useDefaultSettings, fileName);
-        return result;
+        const { requestStatusCallback } = options || {};
```
bsekachev, line 132: `requestStatusCallback` type is not checked

cvat-core/src/project-implementation.ts
```diff
@@ -107,24 +114,37 @@ export default function implementProject(projectClass) {
...
+        if (rqID) {
+            return requestsManager.listen(rqID, { callback: requestStatusCallback });
+        }
+        return new Request({ status: RQStatus.FINISHED, message: '' });
     };
 
-    projectClass.restore.implementation = async function (storage: Storage, file: File | string) {
-        const result = await serverProxy.projects.restore(storage, file);
-        return result;
+    projectClass.restore.implementation = async function (
```
bsekachev, line 140: `requestStatusCallback` type is not checked

cvat-core/src/project-implementation.ts
```diff
@@ -96,9 +98,14 @@ export default function implementProject(projectClass) {
...
         customName?: string,
+        options?: { requestStatusCallback?: (request: Request) => void },
     ) {
-        const result = exportDataset(this, format, saveImages, useDefaultSettings, targetStorage, customName);
-        return result;
+        const { requestStatusCallback } = options || {};
+        const rqID = await exportDataset(this, format, saveImages, useDefaultSettings, targetStorage, customName);
+        if (rqID) {
+            return requestsManager.listen(rqID, { callback: requestStatusCallback });
+        }
```
bsekachev, line 107: Basically `exportDataset`, from this implementaiton will always return rq id or throw an exception.
So, we do not need this `if` condition. Just update typings for `exportDataset`

cvat-core/src/project-implementation.ts
```diff
@@ -107,24 +114,37 @@ export default function implementProject(projectClass) {
...
         fileName?: string,
+        options?: { requestStatusCallback?: (request: Request) => void },
     ) {
-        const result = await serverProxy.projects.backup(this.id, targetStorage, useDefaultSettings, fileName);
-        return result;
+        const { requestStatusCallback } = options || {};
+        const rqID = await serverProxy.projects.backup(this.id, targetStorage, useDefaultSettings, fileName);
+        if (rqID) {
+            return requestsManager.listen(rqID, { callback: requestStatusCallback });
+        }
```
bsekachev, line 136: The same comment as for `exportDataset`

cvat-core/src/enums.ts
bsekachev: > a user shouldn't know that his job is deferred

1. Why do you think so? 
2. For a user on UI we will still may show "Queued", no problem.
3. But this information is required to implement correct long-time-pooling algorithm  (when user has 500+ jobs queued, will client ask the server about each status every N seconds??? Definitely No -> throttling issues. So, for which Requests should client fetch the information? I would say that we not even should now which jobs are deferred [at minimum], but also information about dependend jobs is important)

cvat-core/src/requests-manager.ts
```diff
@@ -0,0 +1,357 @@
...
+            this.#finishedDate = request.finished_date;
+        }
+        if (request.expiry_date !== undefined) {
+            this.#expiryDate = request.expiry_date;
+        }
+        if (request.owner !== undefined) {
+            this.#owner = new User(request.owner);
+        }
+    }
+}
```
bsekachev, line 144: Why do not implement setters? The method makes sense to reduce code duplication, but in this case another question:

- We have setters for `status` and `message`, but not for others. Why?

cvat-core/src/requests-manager.ts
```diff
@@ -0,0 +1,357 @@
...
+    async list(): Promise<{ requests: Request[], count: number }> {
+        const result = await serverProxy.requests.list();
+        const requests = [];
+        for (const request of result) {
+            requests.push(
+                new Request({
+                    ...request,
+                }),
+            );
+        }
```
bsekachev, line 178: `map`?

cvat-core/src/requests-manager.ts
```diff
@@ -0,0 +1,357 @@
...
+
+    async listen(
+        id: string | null,
+        options?: {
+            callback?: (request: Request) => void,
+            filter?: RequestsFilter,
+            initialRequest?: Request,
+        },
+    ): Promise<Request> {
+        let storedID = null;
```
bsekachev, line 190: Why `storedID`? I will suggest to rename to `requestID`

cvat-core/src/requests-manager.ts
```diff
@@ -0,0 +1,357 @@
...
+        if (storedID in this.listening) {
+            if (callback) {
+                this.listening[storedID].onUpdate.push(callback);
+            }
+            return this.listening[storedID].promise;
+        }
+        const promise = new Promise<Request>((resolve, reject) => {
+            const timeoutCallback = (): void => {
+                if (!(storedID in this.listening)) {
+                    return;
```
bsekachev, line 213: How is that possible if you clear timeout before deleting the request from this.listening?

cvat-core/src/requests-manager.ts
```diff
@@ -0,0 +1,357 @@
...
+                        this.listening[storedID].timeout = window.setTimeout(timeoutCallback, delay);
+                        return;
+                    }
+                }
+                if (this.requestStack.length >= REQUESTS_COUNT) {
+                    this.requestStack.pop();
+                }
+                this.requestStack.unshift(timestamp);
+
+                serverProxy.requests.status(id, params).then((serializedRequest) => {
```
bsekachev, line 232: I will suggest to rewrite with `async/await`.

cvat-core/src/requests-manager.ts
```diff
@@ -0,0 +1,357 @@
...
+                            .forEach((update) => update(new Request({
+                                status: RQStatus.FAILED,
+                                message: `Could not get a status of the request ${storedID}. ${error.toString()}`,
+                            })));
+                        reject(error);
+                    }
+                }).finally(() => {
+                    if (storedID in this.listening) {
+                        this.listening[storedID].timeout = null;
+                    }
```
bsekachev, line 276: In then you have:

```
this.listening[storedID].timeout = window.setTimeout(timeoutCallback, this.delayFor(storedID));
```

In finally you set it to `null`. Why?

cvat-core/src/requests-manager.ts
```diff
@@ -0,0 +1,357 @@
...
+                this.requestStack.unshift(timestamp);
+
+                serverProxy.requests.status(id, params).then((serializedRequest) => {
+                    const request = new Request({ ...serializedRequest });
+                    const { status } = request;
+                    // check it was not cancelled
+                    if (storedID in this.listening) {
+                        const { onUpdate } = this.listening[storedID];
+                        if ([RQStatus.QUEUED, RQStatus.STARTED, RQStatus.DEFERRED].includes(status)) {
+                            onUpdate.forEach((update) => update(request));
```
bsekachev, line 239: We can't guarantee one of update callback do not change the request object.

In this case, I will suggest different ways to ensure that it does not happen:
1. Do Request object fully read only
2. Create new Request instance each time for each callback
3. Using typescript define that this argument passes as read only

cvat-ui/src/actions/requests-async-actions.ts
```diff
@@ -0,0 +1,165 @@
...
+    return core.requests
+        .listen(filter ? null : id, {
+            callback: (updatedRequest) => {
+                updateRequestProgress(updatedRequest, dispatch);
+            },
+            filter,
+            initialRequest: request,
+        })
+        .catch((error: Error) => {
+            request.updateFields({ status: RQStatus.UNKNOWN, progress: 0, message: '' });
```
bsekachev, line 45: What is the reason to  update fields outside of cvat-core?

cvat-core/src/server-proxy.ts
```diff
@@ -855,24 +856,16 @@ function exportDataset(instanceType: 'projects' | 'jobs' | 'tasks') {
...
-                            setTimeout(request, 3000);
-                        } else if (status === 201) {
-                            params.action = 'download';
-                            resolve(`${baseURL}?${new URLSearchParams(params).toString()}`);
-                        } else if (isCloudStorage && status === 200) {
-                            resolve();
+                        if (response.status === 202) {
+                            resolve(response.data.rq_id);
                         }
+                        resolve();
```
bsekachev, line 868: Are there any cases, when success response does not include `rq_id`?

cvat-core/src/server-proxy.ts
```diff
@@ -972,13 +941,18 @@ async function importDataset(
         }
     }
     try {
-        return await wait();
+        return rqId;
```
bsekachev, line 944: I do not think we need try catch here :)

cvat-core/src/server-proxy.ts
```diff
@@ -907,30 +900,6 @@ async function importDataset(
...
-                } catch (error) {
-                    reject(generateError(error));
-                }
-            }
-            setTimeout(requestStatus, 2000);
-        });
-    }
     const isCloudStorage = sourceStorage.location === StorageLocation.CLOUD_STORAGE;
 
     if (isCloudStorage) {
```
bsekachev, line 905: We do not need `if .. else`` here anymore

cvat-core/src/server-proxy.ts
```diff
@@ -987,23 +961,16 @@ async function backupTask(id: number, targetStorage: Storage, useDefaultSettings
...
-                    setTimeout(request, 3000);
-                } else if (status === 201) {
-                    params.action = 'download';
-                    resolve(`${url}?${new URLSearchParams(params).toString()}`);
-                } else if (isCloudStorage && status === 200) {
-                    resolve();
+                if (response.status === 202) {
+                    resolve(response.data.rq_id);
                 }
+                resolve();
```
bsekachev, line 973: The same question about rq_id here

cvat-core/src/server-proxy.ts
```diff
@@ -1074,15 +1018,16 @@ async function restoreTask(storage: Storage, file: File | string) {
                 headers: { 'Upload-Finish': true },
             });
     }
-    return wait();
+    const rqId = response.data.rq_id;
```
bsekachev, line 1021: Here we also do not really need ``if ... else`` anymore

Just: `if (isCloudStorage) { response = ..; return response.data.rq_id }` 

By the way I noticed that requests are not try..catched here.

cvat-core/src/server-proxy.ts
```diff
@@ -1093,23 +1038,16 @@ async function backupProject(
...
-
-                if (status === 202) {
-                    setTimeout(request, 3000);
-                } else if (status === 201) {
-                    params.action = 'download';
-                    resolve(`${url}?${new URLSearchParams(params).toString()}`);
-                } else if (isCloudStorage && status === 200) {
-                    resolve();
+                if (response.status === 202) {
+                    resolve(response.data.rq_id);
```
bsekachev, line 1048: The same question about rq_id

cvat-core/src/server-proxy.ts
```diff
@@ -1182,92 +1094,21 @@ async function restoreProject(storage: Storage, file: File | string) {
                 headers: { 'Upload-Finish': true },
             });
     }
-    return wait();
+
+    const rqId = response.data.rq_id;
```
bsekachev, line 1098: The same comment about if ... else as before

bsekachev: And about exception handling

cvat-core/src/server-proxy.ts
```diff
@@ -1380,8 +1221,7 @@ async function createTask(
     }
 
     try {
-        const createdTask = await listenToCreateTask(response.data.id, onUpdate);
-        return createdTask;
+        return response.data.id;
```
bsekachev, line 1224: Please, remove try .. catch.

cvat-core/src/server-proxy.ts
```diff
@@ -1817,7 +1634,7 @@ async function uploadAnnotations(
         }
     }
     try {
-        return await wait();
+        return rqId;
```
bsekachev, line 1637: 1. Please, remove try catch from return value
2. Please, refactor if/else for cloud storage / not a cloud storage

cvat-core/src/session-implementation.ts
```diff
@@ -506,10 +525,11 @@ export function implementTask(Task) {
     };
 
     Task.prototype.listenToCreate.implementation = async function (
-        onUpdate: (state: RQStatus, progress: number, message: string) => void = () => {},
+        options,
     ): Promise<TaskClass> {
         if (Number.isInteger(this.id) && this.size === 0) {
-            const serializedTask = await serverProxy.tasks.listenToCreate(this.id, onUpdate);
+            const request = await requestsManager.listen(null, options);
```
bsekachev, line 531: Listen for null ????????

cvat-ui/src/components/tasks-page/task-item.tsx
```diff
@@ -58,17 +59,21 @@ class TaskItemComponent extends React.PureComponent<TaskItemProps & RouteCompone
...
+                                message: request.message,
+                                progress: Math.floor(request.progress * 100),
+                                state: request.status,
+                            },
+                        });
+                    }
+                },
+                filter: { taskID: taskInstance.id, action: 'create' },
+            },
+            ).then((createdTask: Task) => {
```
bsekachev, line 76: Why do not we have normal `rq_id` for a task creation process? 
This implementation with additional option filter and null rq_id does not look pretty

cvat-core/src/session-implementation.ts
```diff
@@ -331,9 +333,14 @@ export function implementJob(Job) {
...
+        options?: { requestStatusCallback?: (request: Request) => void },
     ) {
-        const result = await exportDataset(this, format, saveImages, useDefaultSettings, targetStorage, customName);
-        return result;
+        const { requestStatusCallback } = options || {};
+        const rqID = await exportDataset(this, format, saveImages, useDefaultSettings, targetStorage, customName);
+        if (rqID) {
+            return requestsManager.listen(rqID, { callback: requestStatusCallback });
+        }
+        return new Request({ status: RQStatus.FINISHED, message: '' });
```
bsekachev, line 343: Export dataset should always return normal rqID or throw Error

cvat-core/src/session-implementation.ts
```diff
@@ -530,15 +550,24 @@ export function implementTask(Task) {
...
+        options?: { requestStatusCallback?: (request: Request) => void },
     ) {
-        const result = await serverProxy.tasks.backup(this.id, targetStorage, useDefaultSettings, fileName);
-        return result;
+        const { requestStatusCallback } = options || {};
+        const rqID = await serverProxy.tasks.backup(this.id, targetStorage, useDefaultSettings, fileName);
+        if (rqID) {
+            return requestsManager.listen(rqID, { callback: requestStatusCallback });
+        }
+        return new Request({ status: RQStatus.FINISHED, message: '' });
```
bsekachev, line 560: Backup should always return normal rqID or throw Error

cvat-core/src/session-implementation.ts
```diff
@@ -783,9 +813,14 @@ export function implementTask(Task) {
...
+        options?: { requestStatusCallback?: (request: Request) => void },
     ) {
-        const result = await exportDataset(this, format, saveImages, useDefaultSettings, targetStorage, customName);
-        return result;
+        const { requestStatusCallback } = options || {};
+        const rqID = await exportDataset(this, format, saveImages, useDefaultSettings, targetStorage, customName);
+        if (rqID) {
+            return requestsManager.listen(rqID, { callback: requestStatusCallback });
+        }
+        return new Request({ status: RQStatus.FINISHED, message: '' });
```
bsekachev, line 823: Export dataset should always return normal rqID or throw Error

cvat-core/src/requests-manager.ts
```diff
@@ -0,0 +1,313 @@
...
+    get progress(): number {
+        return this.#progress;
+    }
+
+    get message(): string {
+        return this.#message;
+    }
+    set message(message) {
+        this.#message = message;
+    }
```
bsekachev, line 74: How do you use it?

cvat-core/src/requests-manager.ts
```diff
@@ -0,0 +1,313 @@
...
+    get id(): string {
+        return this.#id;
+    }
+
+    get status(): RQStatus {
+        return this.#status.toLowerCase() as RQStatus;
+    }
+    set status(status) {
+        this.#status = status;
+    }
```
bsekachev, line 63: How do you use it?

cvat-core/src/server-proxy.ts
```diff
@@ -1128,42 +1068,22 @@ async function restoreProject(storage: Storage, file: File | string) {
...
+        try {
+            params.filename = file;
+            response = await Axios.post(url,
+                new FormData(), {
+                    params,
+                });
+            return response.data.rq_id;
+        } catch (errorData) {
+            throw generateError(errorData);
+        }
```
bsekachev, line 1084: Since `catch` is the same, I will recommend to use common try in the function body. Please, check other changed functions in this file likewise

cvat-core/src/session-implementation.ts
```diff
@@ -491,7 +492,15 @@ export function implementTask(Task) {
...
 
-        const task = await serverProxy.tasks.create(taskSpec, taskDataSpec, onUpdate);
+        const { taskID, rqID } = await serverProxy.tasks.create(
+            taskSpec,
+            taskDataSpec,
+            options?.uploadStatusCallback || (() => {}),
+        );
+        await requestsManager.listen(rqID, {
+            callback: options?.requestStatusCallback,
+        });
```
bsekachev, line 502: Did you decide to generate server id on client and use common endpoint?

cvat-core/src/index.ts
klakhov: Why not, if some code wants to listen but dont need to update any UI

cvat-core/src/project-implementation.ts
klakhov: Implemented as we discussed

cvat-core/src/requests-manager.ts
klakhov: Made the Request read-only

cvat-core/src/server-proxy.ts
klakhov: Yes, in case of export and backups, if the export is finished but not downloaded, new export is not started and we get 200 response without `rq_id` (the same thing for other same comments)

cvat-core/src/index.ts
bsekachev: Could you please provide example?

cvat-core/src/requests-manager.ts
klakhov: Removed

cvat-core/src/server-proxy.ts
bsekachev: So, then something is not working here.
For example:

- serverProxy.backupTask or backupProject returns string or undefined.
- it means Task.backup, and Project.backup also returns string or undefined
- string or undefined is used to start listening (in case with undefined, it seems there is typing error)

image

bsekachev: Perhaps the same for export, but I did not check

cvat-core/src/session-implementation.ts
klakhov: Yes, I think it has advantages: We have guessing logic in one place - serverProxy, so we dont need to invent client ids and then parse it to send filter request.  
If something changes on server, we can change one line on the client(but I suppose next change on the server should be returning correct id)

bsekachev: Okay

cvat-core/src/requests-manager.ts
klakhov: Removed

cvat-core/src/enums.ts
bsekachev: Okay, let's postpone this algorithm, but in this case please remove all 'deferred' from cvat-core

cvat-ui/src/actions/requests-actions.ts
```diff
@@ -0,0 +1,106 @@
...
+} from 'cvat-core-wrapper';
+
+const core = getCore();
+
+export enum RequestsActionsTypes {
+    GET_REQUESTS = 'GET_REQUESTS',
+    GET_REQUESTS_SUCCESS = 'GET_REQUESTS_SUCCESS',
+    GET_REQUESTS_FAILED = 'GET_REQUESTS_FAILED',
+    GET_REQUESTS_STATUS_SUCCESS = 'GET_REQUESTS_STATUS_SUCCESS',
+    GET_REQUESTS_STATUS_FAILED = 'GET_REQUESTS_STATUS_FAILED',
```
bsekachev, line 18:
```suggestion
    GET_REQUEST_STATUS_SUCCESS = 'GET_REQUEST_STATUS_SUCCESS',
    GET_REQUEST_STATUS_FAILED = 'GET_REQUEST_STATUS_FAILED',
```

cvat-ui/src/actions/requests-actions.ts
```diff
@@ -0,0 +1,106 @@
+// Copyright (C) 2024 CVAT.ai Corporation
+//
+// SPDX-License-Identifier: MIT
+
+import { ActionUnion, createAction } from 'utils/redux';
+import { RequestsQuery } from 'reducers';
+import {
+    Request, InstanceType, getCore,
+} from 'cvat-core-wrapper';
```
bsekachev, line 9:
```suggestion
import { Request, InstanceType, getCore } from 'cvat-core-wrapper';
```

bsekachev:
Minor bug:
1. Try to backup project, wait until done
image

2. Try to backup it one more time. You immediately will get two messages. One of them says it was exported to the cloud storage
image

But export was local

bsekachev:
It looks like the feature to delete requests absent in the implementation

image

the method `core.requests.delete` exists in the cvat-core, but I can't see it somewhere to be used.

cvat-ui/src/components/requests-page/styles.scss
```diff
@@ -0,0 +1,175 @@
...
+    display: flex;
+    align-items: center;
+}
+
+.cvat-requests-percent {
+    padding-left: $grid-unit-size * 2;
+}
+
+.cvat-requests-card {
+    height: $grid-unit-size * 11;
```
bsekachev, line 58: 96px looks better

```suggestion
    height: $grid-unit-size * 12;
```

bsekachev: image
image

cvat-ui/src/reducers/import-reducer.ts
```diff
@@ -1,22 +1,22 @@
...
 import { ImportState } from '.';
 
-const core = getCore();
-
 const defaultProgress = 0.0;
 
-export function defineActititiesField(instance: any): 'projects' | 'tasks' | 'jobs' {
-    if (instance instanceof core.classes.Project) {
+export function defineActititiesField(instance: InstanceType | RequestInstanceType): 'projects' | 'tasks' | 'jobs' {
+    const instanceType = getInstanceType(instance);
```
bsekachev, line 15:
```
return `${getInstanceType(instance)}s`; 
```

cvat-ui/src/actions/tasks-actions.ts
```diff
@@ -258,16 +259,38 @@ ThunkAction<Promise<void>, {}, {}, AnyAction> {
...
+                            message = 'CVAT queued the task to import';
+                            helperMessage = 'You may close the window.';
+                        } else if (status === RQStatus.FAILED) {
+                            message = 'Images processing failed';
+                        } else if (status === RQStatus.FINISHED) {
+                            message = 'Task creation finished';
+                        } else {
+                            message = 'Unknown status received';
+                        }
+                    }
```
bsekachev, line 288: What is the difference between these two callbacks? 
Do we really need both? They look very similar

cvat-ui/src/actions/requests-async-actions.ts
```diff
@@ -0,0 +1,138 @@
+// Copyright (C) 2024 CVAT.ai Corporation
+//
+// SPDX-License-Identifier: MIT
+
+import { ThunkAction } from 'utils/redux';
+import {
+    CombinedState,
+    RequestsQuery, StorageLocation,
+} from 'reducers';
```
bsekachev, line 9:
```suggestion
import { CombinedState, RequestsQuery, StorageLocation } from 'reducers';
```

cvat-ui/src/actions/requests-async-actions.ts
```diff
@@ -0,0 +1,138 @@
...
+        dispatch(requestsActions.getRequests(query));
+
+        const state: CombinedState = getState();
+
+        try {
+            const requests = await core.requests.list();
+            dispatch(requestsActions.getRequestsSuccess(requests));
+
+            if (notify) {
+                const shownNotifications = JSON.parse(localStorage.getItem('requestsNotifications') || '[]');
```
bsekachev, line 37: 0. Parsing item from local storage should be try catched, otherwise now it will get into global try catch and show "get requests failed message". That is not true. 
1. Local storage is limited by 5 MB, we will quickly make it full with big number of request (as I see you only append something inside. But never remove. So, gradually it will be full)
2. Finally, I believe we may totally omit this logic. Thus, a user will only see notification if the process finished when the page is opened. I believe it is enough.

cvat-ui/src/actions/requests-async-actions.ts
```diff
@@ -0,0 +1,138 @@
...
+                        ) {
+                            dispatch(importDatasetAsync(
+                                instance as RequestInstanceType,
+                                format,
+                                undefined,
+                                undefined,
+                                undefined,
+                                undefined,
+                                listeningPromise,
+                            ));
```
bsekachev, line 117: May you please explain logic behind this? 

It makes signature and implemenetion of `exportBackupAsync`, `exportDatasetAsync`, `importDatasetAsync` more difficult, but I can't see the reason.

cvat-ui/src/actions/tasks-actions.ts
klakhov: Removed

cvat-ui/src/actions/requests-async-actions.ts
klakhov: Reworked as discussed, using separate listening functions

cvat-ui/src/actions/requests-async-actions.ts
klakhov: Removed this logic

cvat-ui/src/actions/requests-actions.ts
```diff
@@ -0,0 +1,99 @@
...
+    ),
+    getRequestsFailed: (error: any) => createAction(RequestsActionsTypes.GET_REQUESTS_FAILED, {
+        error,
+    }),
+    getRequestStatusSuccess: (request: Request) => (
+        createAction(RequestsActionsTypes.GET_REQUEST_STATUS_SUCCESS, {
+            request,
+        })
+    ),
+    getRequestStatusFailed: (request: Request, error: any) => (
```
bsekachev, line 40: This action is never called

cvat-ui/src/actions/export-actions.ts
```diff
@@ -61,68 +71,105 @@ export const exportActions = {
...
+    },
+): Promise<void> {
+    const { instance, format, saveImages } = params;
+    const resource = saveImages ? 'dataset' : 'annotations';
+
+    const instanceType = getInstanceType(instance);
+    try {
+        let result;
+        let target: 'cloudstorage' | 'local' | undefined;
+        if (rqID) {
```
bsekachev, line 101: Declared type is string, but we have this condition here

cvat-ui/src/actions/requests-async-actions.ts
```diff
@@ -0,0 +1,106 @@
...
+                            ) {
+                                dispatch(exportActions.exportDataset(instance as RequestInstanceType, format, field));
+                                listenExportDatasetAsync(
+                                    rqID,
+                                    dispatch,
+                                    { instance: instance as RequestInstanceType, format, saveImages: type.includes('dataset') },
+                                );
+                            }
+                        }
+                    } else if (operationType === 'import') {
```
bsekachev, line 75: import backup?

cvat-ui/src/actions/requests-actions.ts
```diff
@@ -0,0 +1,99 @@
...
+    CANCEL_REQUEST_FAILED = 'CANCEL_REQUEST_FAILED',
+    DELETE_REQUEST = 'DELETE_REQUEST',
+    DELETE_REQUEST_FAILED = 'DELETE_REQUEST_FAILED',
+}
+
+export const requestsActions = {
+    getRequests: (query?: RequestsQuery) => createAction(RequestsActionsTypes.GET_REQUESTS, { query }),
+    requestFinished: (request: Request) => createAction(RequestsActionsTypes.REQUEST_FINISHED, { request }),
+    requestFailed: (request: Request) => createAction(RequestsActionsTypes.REQUEST_FAILED, { request }),
+    getRequestsSuccess: (requests: Request[]) => createAction(
```
bsekachev, line 29: Request[] is a wrong type, as you read property `count`, not existing on type Array

Correct one is `Awaited<ReturnType<typeof core['requests']['list']>>`

cvat-ui/src/actions/requests-async-actions.ts
```diff
@@ -0,0 +1,106 @@
...
+                    } else if (target === 'job') {
+                        instance = { id: jobID as number, type: target };
+                    } else if (target === 'project') {
+                        instance = { id: projectID as number, type: target };
+                    }
+
+                    if (operationType === 'export') {
+                        if (operationTarget === 'backup') {
+                            if (!(state.export.tasks.backup.current?.[(instance as RequestInstanceType).id] ||
+                                state.export.projects.backup.current?.[(instance as RequestInstanceType).id])
```
bsekachev, line 56: Please, check the condition.
It seems like there will be a bug if we backup both project and task with the same id.

Probably better to rewrite something like:
```
                            if (
                                (instance.type === 'task' && !(instance.id in state.export.tasks.backup.current)) ||
                                (instance.type === 'project' && !(instance.id in state.export.projects.backup.current))
                            ) {
                                // dispatch
                                // listen
                            }
```

bsekachev: That is relevant for  import/export annotations/dataset
And import/export backups

cvat-ui/src/actions/requests-async-actions.ts
```diff
@@ -0,0 +1,106 @@
...
+                            ) {
+                                dispatch(exportActions.exportBackup(instance as RequestInstanceType));
```
bsekachev, line 58: We already discussed it, so, what is the reason to call here:

- exportActions.exportDataset
- importActions.importDataset
- exportActions.exportBackup

And generally store import/export as activities. 
Only to lock some controls on UI?

cvat-ui/src/actions/export-actions.ts
klakhov: Added `undefined`

cvat-ui/src/actions/requests-async-actions.ts
klakhov: Added

cvat-ui/src/actions/import-actions.ts
```diff
@@ -1,16 +1,19 @@
...
 import {
-    getCore, Storage, Job, Task, Project,
+    getCore, Storage, Job, Task, Project, InstanceType,
 } from 'cvat-core-wrapper';
 import { EventScope } from 'cvat-logger';
 import { getProjectsAsync } from './projects-actions';
 import { AnnotationActionTypes, fetchAnnotationsAsync } from './annotation-actions';
+import {
+    getInstanceType, listen, RequestInstanceType, RequestsActions,
+} from './requests-actions';
```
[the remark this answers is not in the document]
klakhov: We dont want to use type imports, please dont suggest that anymore

cvat-ui/src/actions/requests-async-actions.ts
```diff
@@ -0,0 +1,115 @@
...
+                                dispatch(importActions.importDataset(instance as RequestInstanceType, format));
+                                listenImportDatasetAsync(
+                                    rqID,
+                                    dispatch,
+                                    { instance: instance as RequestInstanceType },
+                                );
+                            }
+                        }
+                    }
+                });
```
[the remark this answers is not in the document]
klakhov: We dont want to use for of here

cvat-ui/src/actions/requests-actions.ts
```diff
@@ -0,0 +1,92 @@
...
+        RequestsActionsTypes.GET_REQUESTS_SUCCESS, { requests },
+    ),
+    getRequestsFailed: (error: any) => createAction(RequestsActionsTypes.GET_REQUESTS_FAILED, {
```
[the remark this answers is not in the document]
klakhov: We dont want to specify Error type for now, as we need to change it all across the application.

klakhov:
@coderabbitai pause

bsekachev:
Something does not work when I am within organization. I see two problems:

1. In organization I can't see request for creating a task within the same organization
2. But I see request for creating a task that was initiated outside the organization 

@Marishka17  they are probably server-side issues.

cvat-ui/src/reducers/import-reducer.ts
```diff
@@ -36,7 +33,12 @@ const defaultState: ImportState = {
     tasks: {
         dataset: {
             modalInstance: null,
-            current: {},
+            uploadState: {
+                id: null,
+                format: '',
+                progress: 0,
+                status: '',
+            },
```
bsekachev, line 41: If it was supposed to add upload progress bar (like we have when importing a dataset to a project) to tasks and jobs
It does not work. If not, please, explain

cvat-core/src/api.ts
```diff
@@ -374,6 +375,26 @@ function build(): CVATCore {
...
+                return result;
+            },
+            async cancel(rqID: string) {
+                const result = await PluginRegistry.apiWrapper(cvat.requests.cancel, rqID);
+                return result;
+            },
+            async listen(
+                rqID: string,
+                options?: {
+                    callback?: (request: Request) => void,
```
bsekachev, line 390: Still not clear why callback is optional

cvat-ui/src/components/requests-page/requests-page.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { updateHistoryFromQuery } from 'components/resource-sorting-filtering';
+import EmptyListComponent from './empty-list';
+import RequestsList, { PAGE_SIZE } from './requests-list';
+
+export default function RequestsPageComponent(): JSX.Element {
+    const history = useHistory();
+    const dispatch = useDispatch();
+    const fetching = useSelector((state: CombinedState) => state.requests.fetching);
+    const count = useSelector((state: CombinedState) => state.requests.count);
+    const query = useSelector((state: CombinedState) => state.requests.query);
```
bsekachev, line 23: Please, minimize useSelectors

cvat-ui/src/components/requests-page/requests-page.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+    const updatedQuery = { ...query };
+    const queryParams = new URLSearchParams(history.location.search);
+    for (const key of Object.keys(updatedQuery)) {
+        (updatedQuery as Indexable)[key] = queryParams.get(key) || null;
+        if (key === 'page') {
+            updatedQuery.page = updatedQuery.page ? +updatedQuery.page : 1;
+        }
+    }
```
bsekachev, line 32: We only have one query parameter, the code may be significantly simpler

cvat-ui/src/components/requests-page/styles.scss
```diff
@@ -0,0 +1,175 @@
+// Copyright (C) 2024 CVAT.ai Corporation
+//
+// SPDX-License-Identifier: MIT
+
+@import '../../base';
```
bsekachev, line 5:
```suggestion
@import 'base';
```

cvat-ui/src/components/requests-page/styles.scss
```diff
@@ -0,0 +1,175 @@
...
+        background-color: #fff;
+    }
+
+    &::-webkit-scrollbar-thumb {
+        background-color: #babac0;
+        border-radius: 16px;
+        border: 6px solid #fff;
+    }
+}
+
```
bsekachev, line 23: Why did you decide to use vendor-specific css selectors?

cvat-ui/src/components/requests-page/styles.scss
```diff
@@ -0,0 +1,175 @@
...
+    max-height: $grid-unit-size * 8;
+    overflow: hidden auto;
+    display: block;
+}
+
+.cvat-requests-page-actions-button {
+    border: none;
+    padding: 0;
+    padding-top: 22px;
```
bsekachev, line 134: Please, consider using $grid-unit-size in css rules

cvat-ui/src/components/requests-page/styles.scss
```diff
@@ -0,0 +1,175 @@
...
+        }
+    }
+
+    .ant-card-body {
+        width: 100%;
+        height: 100%;
+    }
+
+    .ant-collapse-header {
+        padding: 0 !important;
```
bsekachev, line 85: May we get rid of `important`?

cvat-ui/src/components/requests-page/requests-list.tsx
```diff
@@ -0,0 +1,74 @@
...
+    const query = useSelector((state: CombinedState) => state.requests.query);
+
+    const queryParams = new URLSearchParams(history.location.search);
+    const updatedQuery = { ...query };
+    for (const key of Object.keys(updatedQuery)) {
+        (updatedQuery as Indexable)[key] = queryParams.get(key) || null;
+        if (key === 'page') {
+            updatedQuery.page = updatedQuery.page ? +updatedQuery.page : 1;
+        }
+    }
```
bsekachev, line 42: We already have this code in `requests-page` component

cvat-ui/src/components/requests-page/requests-list.tsx
```diff
@@ -0,0 +1,74 @@
...
+    displayRequests.sort((a, b) => moment(b.enqueuedDate).valueOf() - moment(a.enqueuedDate).valueOf());
+    return displayRequests.slice((newPage - 1) * PAGE_SIZE, newPage * PAGE_SIZE);
+}
+
+function RequestsList(): JSX.Element {
+    const dispatch = useDispatch();
+    const history = useHistory();
+    const requests = useSelector((state: CombinedState) => state.requests.requests);
+    const count = useSelector((state: CombinedState) => state.requests.count);
+    const query = useSelector((state: CombinedState) => state.requests.query);
```
bsekachev, line 33: Please, minimize useSelectors

cvat-ui/src/components/requests-page/requests-list.tsx
```diff
@@ -0,0 +1,74 @@
...
+                </Col>
+            </Row>
+            <Row justify='center' align='middle'>
+                <Pagination
+                    className='cvat-tasks-pagination'
+                    onChange={(newPage: number) => {
+                        dispatch(getRequestsAsync({
+                            ...query,
+                            page: newPage,
+                        }));
```
bsekachev, line 61: serverProxy:getRequestsAsync uses `fetchAll`. By the way, why???

If it already fetched all list, why do you call getRequestsAsync?

cvat-ui/src/components/requests-page/requests-list.tsx
```diff
@@ -0,0 +1,74 @@
...
+                    current={page}
+                    pageSize={PAGE_SIZE}
+                    showQuickJumper
+                />
+            </Row>
+        </>
+    );
+}
+
+export default RequestsList;
```
bsekachev, line 74: Please, add React.memo where necessary. I see it is missed in other files also. 
Otherwise these components will perform extra re-rendering.

cvat-core/src/config.ts
```diff
@@ -17,6 +19,16 @@ const config = {
     },
     onOrganizationChange: null,
     globalObjectsCounter: 0,
+
+    requestsStatusDelays: (window as any).Cypress ? {
```
bsekachev, line 23: Please, remove Cypress mention from **cvat-core** code.

cvat-core/src/requests-manager.ts
```diff
@@ -0,0 +1,199 @@
...
+                delete this.listening[rqID];
+            }
+        });
+    }
+
+    private delayFor(rqID: string): number {
+        const state = this.listening[rqID];
+        const { request, requestDelayIdx } = state;
+
+        // request is not checked yet, call it immideately
```
bsekachev, line 153:
```suggestion
        // request was not checked yet, call it immediately
```

cvat-core/src/server-response-types.ts
```diff
@@ -495,3 +495,26 @@ export interface SerializedAPISchema {
         url: string;
     };
 }
+
+export interface SerializedRequest {
+    id?: string;
```
bsekachev, line 500: why having `id` is not guaranteed? 

tha same about `created_date` and `owner`

bsekachev:
- The card shows a link to a task that has been deleted 
image

- One more found problem is that in `develop` if you try to create a task and creating thread will be failed (for example try to upload a text file as data, or broken image file), UI will send a request to remove the task after failing. However with this patch - it will not, so, we may see failed tasks in the list.

bsekachev:
I already mentioned this, but did not get answer.
image

Why is TTL 20 minutes for exports??? 
Now downloading does not start automatically anymore, so, it will not work. 
If a user does not catch manually a finished job in 20 minutes, will it be necessary to export from scratch?

cvat-ui/src/components/requests-page/styles.scss
```diff
@@ -0,0 +1,175 @@
...
+            }
+        }
+    }
+
+    .ant-card-body {
+        width: 100%;
+        height: 100%;
+    }
+
+    .ant-collapse-header {
```
bsekachev, line 84: As far as I see you are not  using Collapse element within request card

cvat-ui/src/reducers/import-reducer.ts
klakhov: Removed. That field is used only for projects

cvat-core/src/api.ts
klakhov: Made it required

cvat-ui/src/components/requests-page/styles.scss
klakhov: I believe we cant make pretty scrollbar with standard selectors. It works fine in chrome as its our main supported browser

cvat-ui/src/components/requests-page/requests-list.tsx
klakhov: Removed that call.
Dont we need to check for all requests status? There can be more than 10 in progress. We want notifications for those when they are completed

cvat-core/src/server-response-types.ts
klakhov: Yes. When we create task, we use the same request callback, so we create fake Request instances. We dont have an ID while uploading data, also it wouldnt be pretty to construct a created_date and owner here.

cvat-ui/src/components/requests-page/styles.scss
klakhov: Removed.

cvat-ui/src/components/requests-page/styles.scss
klakhov: Removed

cvat-ui/src/components/requests-page/styles.scss
klakhov: Added

klakhov:
- Fixed links to deleted tasks
- Added back deleting of tasks failed to create
Concerning TTL for exports, its not actually 20mins, its just a visual bug. It seems we get timestamp for expire from server in wrong format so its displayed incorrecty. I beleive Maria is already working on this.

cvat/schema.yml
```diff
@@ -5376,6 +5825,70 @@ paths:
...
+      - tasks
+      security:
+      - sessionAuth: []
+        csrfAuth: []
+        tokenAuth: []
+      - signatureAuth: []
+      - basicAuth: []
+      responses:
+        '202':
+          description: Exporting has been started
```
SpecLad, line 5887: This returns a request ID, right? Can you add a response schema?

cvat/schema.yml
```diff
@@ -9274,6 +9811,115 @@ components:
...
+          nullable: true
+        created_date:
+          type: string
+          format: date-time
+          readOnly: true
+        enqueued_date:
+          type: string
+          format: date-time
+          readOnly: true
+          nullable: true
```
SpecLad, line 9846: Is there a meaningful difference between `created_date` and `enqueued_date`, as far as the user is concerned? If a request was created, it must have also been enqueued.

Marishka17: Probably not for now, but I think it depends on whether we're going to inform the user about deferred jobs or continue to mask them as queued.

klakhov:
I will close this PR and re-open it. It became messy with lots of comments from AI tools and reviews.

cvat/schema.yml
SpecLad: Then let's only expose one of these fields until such a time when there's a reason to expose the other one.

(FWIW, my vote is for `created_date`, as that would be consistent with other API resources.)

PR #7553, opened 2024-03-05
Speed up resource updates when there are no matching webhooks

SpecLad:
### Motivation and context

The signal handling code in `cvat.apps.webhooks.signals` always fetches the old version of the resource from the database and serializes it. But if no webhook is defined for the current resource, this work is pointless.

Select the webhooks first, and only then (if necessary) serialize the old resource. Also, move the diffing of the old and new versions to the `post_save` signal handler, to avoid serializing the new version a second time. This leads to a considerable speedup for resources with complicated serializers, such as tasks and jobs, when no webhooks are attached. In my experiments, saving of annotations was sped up by 20-25% with this patch.

This patch also makes the signal handlers more hygienic with the use of temporary attributes. Both the attributes it uses are now prefixed with `_webhooks` to avoid collisions, and removed after they are used.

### How has this been tested?

I think the existing tests are sufficient to cover this.

### Checklist

### License

### DIFF
--- a/cvat/apps/webhooks/signals.py
+++ b/cvat/apps/webhooks/signals.py
@@ -136,25 +136,32 @@ def get_sender(instance):
 @receiver(pre_save, sender=Invitation, dispatch_uid=__name__ + ":invitation:pre_save")
 @receiver(pre_save, sender=Membership, dispatch_uid=__name__ + ":membership:pre_save")
 def pre_save_resource_event(sender, instance, **kwargs):
-    try:
-        old_instance = sender.objects.get(pk=instance.pk)
-    except ObjectDoesNotExist:
+    instance._webhooks_selected_webhooks = []
+
+    if instance.pk is None:
+        created = True
+    else:
+        try:
+            old_instance = sender.objects.get(pk=instance.pk)
+            created = False
+        except ObjectDoesNotExist:
+            created = True
+
+    resource_name = instance.__class__.__name__.lower()
+
+    event_type = event_name("create" if created else "update", resource_name)
+    if event_type not in map(lambda a: a[0], EventTypeChoice.choices()):
         return
 
-    old_serializer = get_serializer(instance=old_instance)
-    serializer = get_serializer(instance=instance)
-    diff = get_instance_diff(old_data=old_serializer.data, data=serializer.data)
-
-    if not diff:
+    instance._webhooks_selected_webhooks = select_webhooks(instance, event_type)
+    if not instance._webhooks_selected_webhooks:
         return
 
-    before_update = {
-        attr: value["old_value"]
-        for attr, value in diff.items()
-    }
-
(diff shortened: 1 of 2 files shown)


cvat/apps/webhooks/signals.py
```diff
@@ -136,25 +136,32 @@ def get_sender(instance):
...
-    serializer = get_serializer(instance=instance)
-    diff = get_instance_diff(old_data=old_serializer.data, data=serializer.data)
+    if instance.pk is None:
+        created = True
+    else:
+        try:
+            old_instance = sender.objects.get(pk=instance.pk)
+            created = False
+        except ObjectDoesNotExist:
+            created = True
```
Marishka17, line 148: Why do we need to handle it like this? If `instance.pk is not None` but the instance does not exist anymore, it means that the object was deleted before this code was executed. After initializing `created = True` the code below will check the wrong event "created:object". IMHO, we should stop handling here (return from function).

cvat/apps/webhooks/signals.py
```diff
@@ -165,30 +172,37 @@ def pre_save_resource_event(sender, instance, **kwargs):
...
+    del instance._webhooks_old_data
+
+    if not created and old_data is not None:
+        diff = get_instance_diff(old_data=old_data, data=serializer.data)
+
+        if diff:
+            data["before_update"] = {
+                attr: value["old_value"]
+                for attr, value in diff.items()
+            }
```
Marishka17, line 202: Probably it will be better to set `_webhooks_old_data` attribute only when object was updated and `_webhooks_selected_webhooks` is not empty
```suggestion
    if not created and (old_data := getattr(instance, '_webhooks_old_data', None)):
        del instance._webhooks_old_data

        if diff := get_instance_diff(old_data=old_data, data=serializer.data):
            data["before_update"] = {
                attr: value["old_value"]
                for attr, value in diff.items()
            }
```

cvat/apps/webhooks/signals.py
SpecLad: > If `instance.pk is not None` but the instance does not exist anymore,

I wasn't thinking of this case, but rather of the case where an object is created with a preset public key value. I don't think we _do_ that anywhere right now, but I figured the code should be robust and handle that situation.

But also, in the case you describe, I don't see how my code is incorrect? If the instance was deleted prior the call to `save`, then `.save` will re-create it, so `created:*` seems like the correct event.

cvat/apps/webhooks/signals.py
SpecLad: I thought through this code again, and I decided to update it in a different way. I want to make sure that the value of `created` is consistent between the `pre_save` and `post_save` handlers, so I now initialize it based on the value of `_webhooks_old_data`.

Marishka17 (approved):

PR #7556, opened 2024-03-05
Remove double `updated_date` bumping when saving annotations

SpecLad:
### Motivation and context

Currently, `JobAnnotation.update` needlessly bumps the `updated_date` of the job and the task twice (both via `_delete` and `_create`). These updates are costly in terms of time due to the various side effects associated with them (such as webhook sending and event logging), so it's better to avoid doing them gratuitously.

Change the code so that the bumps only happen once.

I simplified the code a bit; previously, the code in `_delete` would only bump the date if something was actually deleted (i.e., the annotations passed in actually existed in the database). Now we just check if there were any annotations we _wanted_ to delete. This means that if a request only deletes non-existent annotations, the date will still get bumped. I don't think it's a significant issue, since this is a pathological case.

### How has this been tested?

Manual testing

### Checklist

### License

### DIFF
--- a/changelog.d/20240305_193358_roman_less_bumpy.md
+++ b/changelog.d/20240305_193358_roman_less_bumpy.md
@@ -1,0 +1,5 @@
+### Fixed
+
+- Job and task `updated_date` are no longer bumped twice when updating
+  annotations
+  (<https://github.com/opencv/cvat/pull/7556>)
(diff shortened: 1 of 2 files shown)


nmanovic (approved):
LGTM

But our API isn't optimal. It looks like UI should make separate requests if annotations are created, deleted and updated at the same time.

PR #7560, opened 2024-03-06
Fix job meta requests

klakhov:
### Motivation and context

Pr fixes two problems: 
- PATCH /data/meta is sent every time the job is saved independently of changes in meta data
- GET /data/meta is sent twice on job load

### Checklist

### License

### DIFF
--- a/cvat-core/src/frames.ts
+++ b/cvat-core/src/frames.ts
@@ -11,10 +11,11 @@ import PluginRegistry from './plugins';
 import serverProxy from './server-proxy';
 import { SerializedFramesMetaData } from './server-response-types';
 import { Exception, ArgumentError, DataError } from './exceptions';
+import { FieldUpdateTrigger } from './common';
 
 // frame storage by job id
 const frameDataCache: Record<string, {
-    meta: Omit<SerializedFramesMetaData, 'deleted_frames'> & { deleted_frames: Record<number, boolean> };
+    meta: FramesMetaData;
     chunkSize: number;
     mode: 'annotation' | 'interpolation';
     startFrame: number;
@@ -38,7 +39,7 @@ const frameDataCache: Record<string, {
 
 export class FramesMetaData {
     public chunkSize: number;
-    public deletedFrames: number[];
+    public deletedFrames: Record<number, boolean>;
     public includedFrames: number[];
     public frameFilter: string;
     public frames: {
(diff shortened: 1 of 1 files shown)


bsekachev:
@klakhov Could you please check tests?

cvat-core/src/frames.ts
```diff
@@ -65,9 +71,30 @@ export class FramesMetaData {
...
+                if (property === 'deleted_frames') {
+                    const handler = {
+                        set(target, prop, value) {
+                            target[prop] = value;
+                            updateTrigger.update('deletedFrames');
+                            return true;
+                        },
+                        deleteProperty(target, prop) {
+                            const result = delete target[prop];
+                            if (result) {
```
bsekachev, line 88: Shall we additionally check if the property exists?  It would be safer, because in general JS allows deleting not-existing properties. 

Also, this check below:

```
if (frame in meta.deletedFrames) {
        delete meta.deletedFrames[frame];
    }
```

may be removed, leaving just `delete` operator

tests/cypress/e2e/features/masks_basics.js
```diff
@@ -163,6 +163,7 @@ context('Manipulations with masks', { scrollBehavior: false }, () => {
     describe('Tests to make sure that empty masks cannot be created', () => {
         beforeEach(() => {
             cy.removeAnnotations();
+            cy.saveJob('PUT');
```
bsekachev, line 166: Could you please clarify why it worked before?

cvat-core/src/frames.ts
```diff
@@ -379,14 +414,31 @@ Object.defineProperty(FrameData.prototype.data, 'implementation', {
     writable: false,
 });
 
+async function getJobMeta(jobID: number): Promise<FramesMetaData> {
+    let meta: FramesMetaData;
```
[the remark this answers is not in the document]
klakhov: Done.

tests/cypress/e2e/features/masks_basics.js
klakhov: `cy.saveJob('PATCH', 200, 'removeUnderlyingPixelsUndoRedo')` lines inside `Empty masks are deleted using remove underlying pixels feature` case are depending on `PATCH` annotations requests. But after `removeAnnotations`, first update of the annotations is sent via `PUT` request. 

So before the fix first call of the `cy.saveJob('PATCH'...` catched not the `PATCH /annotations` , but `PATCH data/meta` and worked well. After the fix  `PATCH data/meta` is not sent, so the test failed.

We can eighter: 
1. Look for `PUT` request in first call of the `cy.saveJob(...` inside the test 
2. Save job after each remove annotations, so all saves inside test are looking for `PATCH` requests. 
Ive decided to go the second way.

bsekachev (approved):

PR #7576, opened 2024-03-08
Make analytics report update job scheduling more efficient

SpecLad:
### Motivation and context

Currently, `schedule_analytics_report_autoupdate_job` attempts to debounce job scheduling by examining existing jobs before scheduling a new one. Unfortunately, the `scheduler.get_jobs` function, which it uses for this purpose, scales poorly. Not only does it fetch IDs of all scheduled jobs (and not just ones related to the current object), but it then fetches information about every job, one by one. The current logic doesn't even need this information, but RQ Scheduler provides no method to get just the IDs.

Replace the current logic with a new lightweight approach that uses a custom Redis key to block scheduling of additional jobs.

### How has this been tested?

Manual testing.

### Checklist

### License

### DIFF
--- a/cvat/apps/analytics_report/report/create.py
+++ b/cvat/apps/analytics_report/report/create.py
@@ -66,20 +66,20 @@ class AnalyticsReportUpdateManager:
     def _get_queue(self):
         return django_rq.get_queue(settings.CVAT_QUEUES.ANALYTICS_REPORTS.value)
 
-    def _make_queue_job_prefix(self, obj) -> str:
+    def _make_queue_job_id_base(self, obj) -> str:
         if isinstance(obj, Task):
-            return f"{self._QUEUE_JOB_PREFIX_TASK}{obj.id}-"
+            return f"{self._QUEUE_JOB_PREFIX_TASK}{obj.id}"
         else:
-            return f"{self._QUEUE_JOB_PREFIX_PROJECT}{obj.id}-"
+            return f"{self._QUEUE_JOB_PREFIX_PROJECT}{obj.id}"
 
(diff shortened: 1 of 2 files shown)


cvat/apps/analytics_report/report/create.py
```diff
@@ -68,18 +68,18 @@ def _get_queue(self):
...
 
-    def _make_initial_queue_job_id(self, obj) -> str:
-        return f"{self._make_queue_job_prefix(obj)}initial"
+    def _make_queue_job_id(self, obj, start_time: timezone.datetime) -> str:
+        return f"{self._make_queue_job_prefix(obj)}-{start_time.timestamp()}"
 
-    def _make_regular_queue_job_id(self, obj, start_time: timezone.datetime) -> str:
-        return f"{self._make_queue_job_prefix(obj)}{start_time.timestamp()}"
+    def _make_autoupdate_blocker_key(self, obj) -> str:
+        return f"cvat:analytics:autoupdate-blocker:{self._make_queue_job_prefix(obj)}"
```
klakhov, line 82: I dont think we should put something named `prefix` to the end of the key, its better to be in the beggining

klakhov (commented):
Generally, patch works for me.

cvat/apps/analytics_report/report/create.py
SpecLad: Yeah, that's fair enough. I think the format of the key itself is fine, but I can rename the function to something more semantically fitting.

SpecLad: Renamed.

klakhov (approved):

PR #7582, opened 2024-03-09
[GSoC2024] Added quality reporting for Tag annotations

### LINKED ISSUE #7424: Incorrect confusion matrix when GT job is created with a percentage of the entire task dataset (less than 100%)

### Steps to Reproduce

1. Create a task
2. Create a GT job with some percentage of annotations (less than 100)
3. Annotate the GT job, change stage/state
4. Create a quality report for the task
5. Check comparision_summary -> annotations -> confusion_matrix

There is an example with the resulting incorrect confusion matrix.
image

### Expected Behavior

_No response_

### Possible Solution

_No response_

### Context

_No response_

### Environment

_No response_

Viditagarwal7479:
Fixes #7424 

This PR adds quality computations for Tag annotations.

### Motivation and context
- Earlier in the code when we use annotation type as Tag then incorrect quality report was generated for it. 
- On analyzing the code I found that in this [line](https://github.com/opencv/cvat/blob/develop/cvat/apps/quality_control/quality_reports.py#L151) the annotation of type `dm.AnnotationType.label` is missing. Thus while generating the quality report annotations of type tag were not considered in the report. 
- On adding `dm.AnnotationType.label` there, [match_annotations](https://github.com/opencv/cvat/blob/bfb902fca4c44c1aa44be490193ba87a282aafc0/cvat/apps/quality_control/quality_reports.py#L1597) function started to break as the output of `self.match_labels` wasn't coherent with the output of other annotation types. 
- On trying to fix this by changing the output format of [match_annotations](https://github.com/opencv/cvat/blob/bfb902fca4c44c1aa44be490193ba87a282aafc0/cvat/apps/quality_control/quality_reports.py#L1597) function things started to break down in further calculation part where particularly objects like `itertools.chain(matches, mismatches)` where used [[1](https://github.com/opencv/cvat/blob/bfb902fca4c44c1aa44be490193ba87a282aafc0/cvat/apps/quality_control/quality_reports.py#L1690), [2](https://github.com/opencv/cvat/blob/bfb902fca4c44c1aa44be490193ba87a282aafc0/cvat/apps/quality_control/quality_reports.py#L1712), [3](https://github.com/opencv/cvat/blob/bfb902fca4c44c1aa44be490193ba87a282aafc0/cvat/apps/quality_control/quality_reports.py#L1784)]
- So, the most appropriate approach which I was able to think was to change the output of `match_labels` function but this wasn't feasible as it depended on the implementation of `datumaro`
- Since, the label of type Bbox is already working correctly I emulated a label of type Tag as a Bbox, a square with coordinates at (0,0) and side of 1 unit. After this made changes accordingly so that this emulation is coherent across all the functions which will be seeing the annotations. So, it doesn't happens that one function has memorized the label as type of `AnnotationType.Label` but afterwards it becomes `AnnotationType.Bbox` for it.

### How has this been tested?
Performed annotations of type Mask and Tag in the updated code and saw that the quality report now generated is correct. 
1. Create a task
2. Create a GT job for it (having less 100% data)
3. Complete both the jobs
4. Generate the quality report
5. Repeat the same for another annotation type.

### Checklist

### License

## Summary by CodeRabbit

- **New Features**
	- Enhanced Quality Report calculations to include annotations of type Tag.
- **Bug Fixes**
	- Resolved issues with mismatching groups and attributes in quality conflict reports.
- **Documentation**
	- Updated JSON files to reflect new entries and detailed information on tasks, jobs, and user activities.
- **Tests**
	- Increased test coverage for quality reports and conflicts with new scenarios and data entries.


Viditagarwal7479:
It's failing on `rest_api/test_quality_control.py::TestQualityReportMetrics::test_settings_affect_metrics[line_thickness-22]` in test file
`tests/python/rest_api/test_quality_control.py:1204: AssertionError` but when I am running this test locally it's passing without any issue.
image of terminal showing all test cases have passed

nmanovic:
@Viditagarwal7479 , could you please add a line into CHANGELOG.md (see comments inside the file)
Also, please look at Helm report.

Viditagarwal7479:
> @Viditagarwal7479 , could you please add a line into CHANGELOG.md (see comments inside the file) Also, please look at Helm report.

Sure, I shall make appropriate changes in CHANGELOG.md
I am not able to understand why the Helm Report is failing looking at the logs, when I try to replicate the same it's passing as I have attached the screenshot also in the previous comment. 
I shall again give it a look, it would be really helpful if someone could provide the slightest of hint regarding this.

cvat/apps/quality_control/quality_reports.py
```diff
@@ -156,6 +157,7 @@ class ComparisonParameters(_Serializable):
         dm.AnnotationType.polygon,
         dm.AnnotationType.polyline,
         dm.AnnotationType.skeleton,
+        dm.AnnotationType.label,
```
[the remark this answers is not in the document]
Viditagarwal7479: Ah thank you so much for the clarification related to `AnnotationType.label`. Another possible implementation of tag comparison can be to use a custom label matching function instead of using the one in `datumaro`. 

As you correctly mentioned to reproduce the original problem first, first looking at it then going to matching labels.

On experimenting out with few examples I didn't come across any issue in the formation of confusion matrix or any other variable of quality report. 
In that case, @Marishka17 can you please provide some more details related to the job. If possible share the backup of that task?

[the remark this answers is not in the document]
zhiltsov-max: Please check the answer https://github.com/opencv/cvat/issues/7424#issuecomment-1988760249

Viditagarwal7479:
This PR is working on an issue which isn't the one mentioned in description (fixes). Will create a issue corresponding to this PR after fixing the issue of confusion matrix. Thanks!

Viditagarwal7479:
Hello @zhiltsov-max ,
Yesterday, I created a [new PR](https://github.com/opencv/cvat/pull/7604) corresponding to the issue tagged in this PR.
In the meantime, it gets merged, should I create a new issue and begin the discussion corresponding to the issue with the quality report of annotation type `AnnotationType.Label`, which this PR is trying to solve?

zhiltsov-max:
Let's finish that PR first.

zhiltsov-max:
As the [PR](https://github.com/opencv/cvat/pull/7604) was merged and the former issue resolved, this PR can be continued now as a new feature implementation, since image tag comparison was not planned for the quality feature initially. Thus, please share your ideas on how you'd like tag comparison to be implemented or close the PR if you don't plan to continue it.

Viditagarwal7479:
Hello @zhiltsov-max ,
Thank you for helping me to merge that PR.
Yes, I am looking at the code base on how we can modify the existing match function for the annotation type of `dm.AnnotationType.label`. Once I am done drafting the GSOC proposal, I shall begin implementing this feature.

Viditagarwal7479:
Hello @zhiltsov-max ,
I think we implement this by writing a custom function to perform [matching of labels](https://github.com/Viditagarwal7479/cvat/blob/3e0679f2dd9a430fd4581870178c379e10d8d3ce/cvat/apps/quality_control/quality_reports.py#L1005) such a function is already present for other types of annotations (overriding the implementation of `datumaro`) then further we skip the calculation of metrics which aren't needed for tag annotation, i.e. jaccard_index (IoU).

Viditagarwal7479:
How can we compute the value of `label_jaccard_indices`? From [this line](https://github.com/Viditagarwal7479/cvat/blob/19c5ad514a99aacf1101f0b19088e255563d6065/cvat/apps/quality_control/quality_reports.py#L1970] we can see that `label_jaccard_indices` value is calculated from the confusion matrix so in the case where we have a combination of various types of annotations might result in incorrect calculation.

Viditagarwal7479:
A question beyond the context of this PR,
What's the difference between [Helm Test](https://github.com/opencv/cvat/actions/runs/8489328219/job/23259297527?pr=7582) and [rest_api_testing](https://github.com/opencv/cvat/actions/runs/8489328235/job/23259362262?pr=7582#logs)? Both of them are running the `REST API and SDK tests`.

Viditagarwal7479:
> How can we compute the value of `label_jaccard_indices`? From [this line](https://github.com/Viditagarwal7479/cvat/blob/19c5ad514a99aacf1101f0b19088e255563d6065/cvat/apps/quality_control/quality_reports.py#L1970) we can see that `label_jaccard_indices` value is calculated from the confusion matrix so in the case where we have a combination of various types of annotations might result in incorrect calculation.


Wow! Just came to know that the Jaccard index is **not** just mIoU, which is only valid for segmentation and bounding box tasks. 
[Wikipedia page of Jaccard Index](https://en.wikipedia.org/wiki/Jaccard_index): The Jaccard index, also known as the Jaccard similarity coefficient, is a statistic used for gauging the similarity and diversity of sample sets. However, they are identical in generally taking the ratio of Intersection over Union.

Viditagarwal7479:
The Helm test is failing unexpectedly due to this reason,
```
>           raise ApiException(http_resp=r)
E           cvat_sdk.api_client.exceptions.ApiException: Status Code: 400
E           Reason: Bad Request
```
This doesn't seem to be correlated with the changes in this PR.

Viditagarwal7479:
I don't think we would need additional tests for this change, as the [TestQualityReportMetrics::test_report_summary](https://github.com/Viditagarwal7479/cvat/blob/d3689a6223fb76c0570679dab5803f606872b10f/tests/python/rest_api/test_quality_control.py#L1085) checks for the quality report for the `task_id, 22` which has combination for various types of annotation including the tag. Thus its quality report also got updated after the above changes.

cvat/apps/quality_control/quality_reports.py
```diff
@@ -1941,6 +1967,10 @@ def _generate_annotations_summary(
         gt_ann_counts = np.sum(confusion_matrix, axis=0)
         total_annotations_count = np.sum(confusion_matrix)
 
+        # [https://en.wikipedia.org/wiki/Jaccard_index] The Jaccard index is a statistic used for gauging the \
+        # similarity and diversity of sample sets. However, they are identical in generally taking the ratio of \
+        # Intersection over Union.
+        # Equation: https://scikit-learn.org/stable/modules/model_evaluation.html#jaccard-similarity-coefficient-score
```
zhiltsov-max, line 1973: This doesn't seem to be needed here.

cvat/apps/quality_control/quality_reports.py
```diff
@@ -1000,6 +1004,21 @@ def _match_ann_type(self, t, *args):
...
+        b_labels = self._get_ann_type(dm.AnnotationType.label, item_b)
+
+        returned_values = _match_segments(
+            a_labels,
+            b_labels,
+            distance=lambda a, b: 1 - float(a is None or b is None),
+            label_matcher=lambda a, b: a.label == b.label,
+        )
+        if self.return_distances:
+            returned_values = returned_values + ({},)
```
zhiltsov-max, line 1018: If you want to use `_match_segments`, it's probably better to use `self._match_segments` instead (and remove some code here). Otherwise the datumaro's one (which is inherited here) looks fine, it just needs some return value adjustment.

zhiltsov-max:
> A question beyond the context of this PR,
What's the difference between [Helm Test](https://github.com/opencv/cvat/actions/runs/8489328219/job/23259297527?pr=7582) and [rest_api_testing](https://github.com/opencv/cvat/actions/runs/8489328235/job/23259362262?pr=7582#logs)? Both of them are running the REST API and SDK tests.

The Helm tests mostly focus on Helm deployment capability, even though basically they are the same. The Helm tests are unable to start side services, some of the "normal" tests are disabled there. Keep in mind that the Helm tests can often fail for obscure reasons, it can often be solved by restarting them.

> Wow! Just came to know that the Jaccard index is not just mIoU, which is only valid for segmentation and bounding box tasks.

Yes, this metric in the quality reports is not about segmentation. For segmentation there is `shape.mean_iou`

zhiltsov-max (commented):
It doesn't seem totally correct that these annotations are included in the `annotation_components.shape` section. Probably, this should be changed.

Viditagarwal7479:
> It doesn't seem totally correct that these annotations are included in the `annotation_components.shape` section. Probably, this should be changed.

How about we divide `frame_results` (matches, mismatches, gt_unmatched, ds_unmatched, pairwise_distances) into `shape_frame_results` and `tag_frame_results`?

cvat/apps/quality_control/quality_reports.py
Viditagarwal7479: Thanks for pointing out, I missed `self._match_segments` is the better version of the function, I will fix it. The values in the datumaro's one corresponding to frame results were incorrect.

zhiltsov-max:
> How about we divide frame_results (matches, mismatches, gt_unmatched, ds_unmatched, pairwise_distances) into shape_frame_results and tag_frame_results?

It's possible. In the report, it could be included in a field like `frame_results.[ ].by_type.tag` or simply `frame_results.[ ].tags`, though I don't think it's really needed now.

Viditagarwal7479:
> though I don't think it's really needed now

In the sense that fixing this thing related to "shape" might not be a major concern as of now?

zhiltsov-max:
> In the sense that fixing this thing related to "shape" might not be a major concern as of now?

I meant adding some field for tags in reports. The tag values just should not be included in the `annotation_components` section, but they still should be included in the frame results.

Viditagarwal7479:
> > In the sense that fixing this thing related to "shape" might not be a major concern as of now?
> 
> I meant adding some field for tags in reports. The tag values just should not be included in the `annotation_components` section, but they still should be included in the frame results.

Oh true, values calculated for tag shouldn't be used in the `annotation_comopents`. I consider that it's better we add `frame_results.kind["shape"]` so directly these values could be used for the `annotation_comopents`.

cvat/apps/quality_control/quality_reports.py
Viditagarwal7479:
```py
a_unmatched = a_labels - b_labels
b_unmatched = b_labels - a_labels
```
This is from the `datumaro`'s `match_labels` implementation due to this line `a_unmatched` and `b_unmatched` are computed incorrectly and obviously due to this `mispred` isn't computed.

zhiltsov-max: Could you share an example when this is incorrect? "Mispred" is not defined for simple label comparison.

Viditagarwal7479:
In this [tests/python/rest_api/test_quality_control.py::TestQualityReportMetrics::test_unmodified_task_produces_the_same_metrics](https://github.com/Viditagarwal7479/cvat/blob/b5f58cd3116a94d6d0d5a57e4a44f45026f90281/tests/python/rest_api/test_quality_control.py#L1101) is failing.

But on my local, it's passing,
locally test passing screenshot

zhiltsov-max:
I could reproduce the problem, please try to restart the tests and debug.

cvat/apps/quality_control/quality_reports.py
Viditagarwal7479: Suppose there are two labels, "Dog" and "Cat". A frame is marked as "Dog" in GT Job and "Cat" in DS Job.
In this case,
[Expectation, if we use custom function]`matches`, `gt_unmatched` and `ds_unmatched` should be empty, whereas `mispred` contains information about that frame.
[Reality, if we use `datumaro`'s] `matches` and `mispred` are empty.
To make the results for the tag to make sense with existing parameters then, we have to make the implementation which can produce the outputs as in `Expectation`

Viditagarwal7479:
Do you know why is this [e2e_testing (features)](https://github.com/opencv/cvat/actions/runs/8536815462/job/23386768489?pr=7582#logs) test failing? In the previous PR, also I observed that this fails at times and automatically gets passed at times after the `codecov` bot posts the report.

zhiltsov-max:
It's a known problem with cypress tests, they can randomly fail. In such cases, we typically just restart the CI, until it's green.

cvat/apps/quality_control/quality_reports.py
zhiltsov-max: I can't totally agree about the expected result. It can be so if there is only 1 label per image, but there is no such restriction in CVAT or Datumaro.

Viditagarwal7479: After you mentioned this, I tried it on several combinations of `gt_ann` and `ds_ann` for multi_label scenarios. It fails for the scenario, 
`gt_ann`: [tag_1, tag_2]
`ds_ann`: [tag_2, tag_3]
In such a case, it will try to compare tag_1 to tag_2 and tag_2 to tag_3, thus producing incorrect results.
Amazed by how you sensed all this just by looking at the code!

Viditagarwal7479: Ahh, now, while trying to fix this. I got the feel of why mispred isn't there in datumaro's match_label.

`gt_ann`: [tag_1, tag_2, tag_3]
`ds_ann`: [tag_2, tag_4]
Now here, 
`matches`: [tag_2]
Comparing [tag_1, tag_3] and [tag_4] is ambiguous.

https://scikit-learn.org/stable/modules/generated/sklearn.metrics.multilabel_confusion_matrix.html Based on this also in the case of multi-label tag annotation, we have a class-wise confusion matrix, not a single confusion matrix. 

So, one thing we can do is to restrict this PR only to the annotations where each frame as single label. Or incorporate a label-wise (i.e., class-wise) confusion matrix in quality_report.

Viditagarwal7479: > About the matrix - it's the same situation as with shapes, since there are no guarantees GT and DS have the same number of shapes or they overlap. It's already handled by the code. Per-label matrix is also there.

Thanks to this, it clicked that the distance function in the case of shapes is able to handle this. The distance function was not separate two tags of different labels, tags of the same labels are closer than of different labels.

> It's pretty simple if you look at this from this [perspective](https://en.wikipedia.org/wiki/Jaccard_index). Actually, it's used all over in this code - e.g. IoU, GT vs DS. Label matching as well.

Didn't got this...

zhiltsov-max: > Didn't got this...

This is to the comment:

> After you mentioned this, I tried it on several combinations of gt_ann and ds_ann for multi_label scenarios. It fails for the scenario,
gt_ann: [tag_1, tag_2]
ds_ann: [tag_2, tag_3]
In such a case, it will try to compare tag_1 to tag_2 and tag_2 to tag_3, thus producing incorrect results.
Amazed by how you sensed all this just by looking at the code!

For tags and other annotation types, we're just trying to find intersection of 2 sets of objects. Matches = intersection, gt_extra = gt - intersection, ds_extra = ds - intersection, just by definition. It's not related to the confusion matrix directly. Mispred is a part of intersection, that has different labels, it's separated to help with the following analysis.

Viditagarwal7479: True, now it makes sense. The current code takes this into account.

Viditagarwal7479: https://github.com/cvat-ai/cvat/pull/7582/commits/6ec3d2b2bf5bd5826eb57b25cc51a04215be37c6

cvat/apps/quality_control/quality_reports.py
```diff
@@ -1000,6 +1004,20 @@ def _match_ann_type(self, t, *args):
         else:
             return None
 
+    def match_labels(self, item_a, item_b):
+        def label_distance(a, b):
+            if a is None or b is None:
+                return 0
+            return 0.5 + (a.label == b.label) / 2
```
zhiltsov-max, line 1011: Is this formula better than simple `a.label == b.label`?

Viditagarwal7479: - Case 1: **One of the labels isn't present**, then in such case, the distance should be `0`
- Case 2: When there's an **intersection between a set of labels**, they should have a distance of `1`
- Case 3: There's a **mismatch of labels**, i.e., a case between the above cases. The distance can be anything be 0 to 1 but fixed across all labels here `0.5`

Wherever there's `distance` in mind try to make an analogy with IoU in case of shape.

https://github.com/cvat-ai/cvat/pull/7582#discussion_r1550197708 based on our conversation in this, I got this idea.

zhiltsov-max: Ok, the question is only about the last line, so let's discuss the point 3. I don't think we can really use any constant, because if it's less than `self.iou_thresold`, it can be filtered out during matching. Probably, it must be bigger.

Viditagarwal7479: If that constant value is equal to or more than `self.iou_threshold`, then we are good to go. One thing we can do is to pass another named parameter in the `self._match_segments` function related to `dist_thresh` by this, we can fix the dist_thresh constant `0.5` for the label case. 

How come you are able to think so so far-fetched relations? I wouldn't have realised this until somebody raised an issue that after raising the iou_threshold the results are coming incorrect.

Viditagarwal7479: Hello @zhiltsov-max a gentle reminder to this.

zhiltsov-max: Yes, probably we can fix the threshold to some value for this check, as it's not really applicable to tag comparison. Maybe it will be better to return back to the non-class `_match_segments` then, or add a function parameter.

cvat/apps/quality_control/quality_reports.py
```diff
@@ -1028,7 +1048,11 @@ def _match_segments(
                 extra_args["label_matcher"] = label_matcher
 
             returned_values = _match_segments(
-                a_objs, b_objs, distance=distance, dist_thresh=self.iou_threshold, **extra_args
+                a_objs,
+                b_objs,
+                distance=distance,
+                dist_thresh=dist_thresh or self.iou_threshold,
```
zhiltsov-max, line 1054:
```suggestion
                dist_thresh=dist_thresh if dist_thresh is not None else self.iou_threshold,
```

The value of `0` passed would be ignored.

Viditagarwal7479: Ah, true.

zhiltsov-max (commented):
Basically, the changes look good to me. However, during testing in UI I've noticed 2 problems:

1. 
image
It would be nice to show the source of the tag somehow. Currently, both GT and current job tags are shown in just 1 list. As a solution, I can suggest showing 2 sorted lists one over another, or add an extra suffix like `(Ground truth)` to each tag.

2. In the list of conflicts on the frame, there can be a conflict for tags. If I hover the mouse on the conflict card in the list, then move the cursor out, then hover again, the UI crashes:

<details>

image


```
TypeError
Cannot read properties of undefined (reading 'shapeType')

TypeError: Cannot read properties of undefined (reading 'shapeType')
    at PL.deactivateShape (http://localhost:8080/assets/cvat-ui.40e9762cea80b6f30e54.min.js:2:7197338)
    at PL.deactivate (http://localhost:8080/assets/cvat-ui.40e9762cea80b6f30e54.min.js:2:7198107)
    at PL.highlight (http://localhost:8080/assets/cvat-ui.40e9762cea80b6f30e54.min.js:2:7202929)
    at PL.notify (http://localhost:8080/assets/cvat-ui.40e9762cea80b6f30e54.min.js:2:7185128)
    at oj.notify (http://localhost:8080/assets/cvat-ui.40e9762cea80b6f30e54.min.js:2:7017095)
    at oj.highlight (http://localhost:8080/assets/cvat-ui.40e9762cea80b6f30e54.min.js:2:7024207)
    at IL.highlight (http://localhost:8080/assets/cvat-ui.40e9762cea80b6f30e54.min.js:2:7218944)
    at YPe.componentDidUpdate (http://localhost:8080/assets/cvat-ui.40e9762cea80b6f30e54.min.js:2:9217490)
    at us (http://localhost:8080/assets/cvat-ui.40e9762cea80b6f30e54.min.js:2:5429428)
    at Sl (http://localhost:8080/assets/cvat-ui.40e9762cea80b6f30e54.min.js:2:5447053)
```

</details>

You can fix this in another PR if you want. After this is fixed, we can merge the PR.

cvat/apps/quality_control/quality_reports.py
```diff
@@ -1728,21 +1772,21 @@ def _find_closest_unmatched_shape(shape: dm.Annotation):
                     )
                 )
 
-        for unmatched_ann in gt_unmatched:
+        for shape_unmatched_ann in gt_unmatched:
```
zhiltsov-max, line 1775:
```suggestion
        for unmatched_ann in gt_unmatched:
```

cvat/apps/quality_control/quality_reports.py
```diff
@@ -1728,21 +1772,21 @@ def _find_closest_unmatched_shape(shape: dm.Annotation):
...
             conflicts.append(
                 AnnotationConflict(
                     frame_id=frame_id,
                     type=AnnotationConflictType.MISSING_ANNOTATION,
-                    annotation_ids=[self._dm_ann_to_ann_id(unmatched_ann, self._gt_dataset)],
+                    annotation_ids=[self._dm_ann_to_ann_id(shape_unmatched_ann, self._gt_dataset)],
```
zhiltsov-max, line 1780:
```suggestion
                    annotation_ids=[self._dm_ann_to_ann_id(unmatched_ann, self._gt_dataset)],
```

cvat/apps/quality_control/quality_reports.py
```diff
@@ -1728,21 +1772,21 @@ def _find_closest_unmatched_shape(shape: dm.Annotation):
...
                 )
             )
 
-        for unmatched_ann in ds_unmatched:
+        for shape_unmatched_ann in ds_unmatched:
```
zhiltsov-max, line 1784:
```suggestion
        for unmatched_ann in ds_unmatched:
```

cvat/apps/quality_control/quality_reports.py
```diff
@@ -1728,21 +1772,21 @@ def _find_closest_unmatched_shape(shape: dm.Annotation):
...
             conflicts.append(
                 AnnotationConflict(
                     frame_id=frame_id,
                     type=AnnotationConflictType.EXTRA_ANNOTATION,
-                    annotation_ids=[self._dm_ann_to_ann_id(unmatched_ann, self._ds_dataset)],
+                    annotation_ids=[self._dm_ann_to_ann_id(shape_unmatched_ann, self._ds_dataset)],
```
zhiltsov-max, line 1789:
```suggestion
                    annotation_ids=[self._dm_ann_to_ann_id(unmatched_ann, self._ds_dataset)],
```

zhiltsov-max:
> I was looking at the code this [lines of code](https://github.com/Viditagarwal7479/cvat/blob/0c70c6b7e67bd585447b49770e5e404c88f2e8ab/cvat-ui/src/components/annotation-page/standard-workspace/objects-side-bar/issues-list.tsx#L129-L132) make the sidebar button on clicking which the gt annotations show up in the main canvas, I am not able to figure out how to render these gt annotations is being controlled. Could you suggest this a bit, so I can get a starting point and thus figure the rest out?

@klakhov, @bsekachev - please, comment on UI details.

> So, should I consider trying it by creating a new issue and another PR related to this?

Yes, I think it would be the right way of doing this.

Viditagarwal7479:
> > I was looking at the code this [lines of code](https://github.com/Viditagarwal7479/cvat/blob/0c70c6b7e67bd585447b49770e5e404c88f2e8ab/cvat-ui/src/components/annotation-page/standard-workspace/objects-side-bar/issues-list.tsx#L129-L132) make the sidebar button on clicking which the gt annotations show up in the main canvas, I am not able to figure out how to render these gt annotations is being controlled. Could you suggest this a bit, so I can get a starting point and thus figure the rest out?
> 
> @klakhov, @bsekachev - please, comment on UI details.
> 
> > So, should I consider trying it by creating a new issue and another PR related to this?
> 
> Yes, I think it would be the right way of doing this.

Thanks, then I am opening a new issue regarding the same and a PR, too, once @klakhov, and @bsekachev comment on the UI details.

klakhov:
Hi @Viditagarwal7479,

Regarding UI part of the PR, consider looking on this component tree(sidebar object list):  [ObjectList](https://github.com/cvat-ai/cvat/blob/develop/cvat-ui/src/components/annotation-page/standard-workspace/objects-side-bar/objects-list.tsx#L103) -> [ObjectItemContainer](https://github.com/cvat-ai/cvat/blob/develop/cvat-ui/src/containers/annotation-page/standard-workspace/objects-side-bar/object-item.tsx#L406) -> [ObjectItem](https://github.com/cvat-ai/cvat/blob/develop/cvat-ui/src/components/annotation-page/standard-workspace/objects-side-bar/object-item.tsx#L116) -> [ItemBasics](https://github.com/cvat-ai/cvat/blob/develop/cvat-ui/src/components/annotation-page/standard-workspace/objects-side-bar/object-item-basics.tsx#L97). As you can see, `objectState` knows if its ground truth or not, thats how the component determines if it should render `GT` prefix on the sidebar. You can use the same approach somewhere inside [FrameTags](https://github.com/cvat-ai/cvat/blob/develop/cvat-ui/src/components/annotation-page/tag-annotation-workspace/frame-tags.tsx#L46) component to make GT tags look different from usual ones(make new row, assign new color etc.).

About the crash after hovering the confilict on sidebar, please look on the highlight pipeline. The idea here is to highlight the annotations [on hover](https://github.com/cvat-ai/cvat/blob/develop/cvat-ui/src/components/annotation-page/standard-workspace/objects-side-bar/issues-list.tsx#L223) the client ids are [sent to canvasInstance](https://github.com/cvat-ai/cvat/blob/develop/cvat-ui/src/components/annotation-page/canvas/views/canvas2d/canvas-wrapper.tsx#L492). But tags shouldnt be sent to `cvat-canvas` as they're rendered only in `cvat-ui`. Thats why it maybe crushing.

Viditagarwal7479:
Thank you, @klakhov for helping me out on the UI part,
I have created the PR corresponding to the [first issue](https://github.com/cvat-ai/cvat/issues/7773) and an [issue related](https://github.com/cvat-ai/cvat/issues/7775) to the second issue.

Viditagarwal7479:
Hello @zhiltsov-max can we now merge this one too?

zhiltsov-max:
@Viditagarwal7479, there was the https://github.com/cvat-ai/cvat/pull/7776 PR, do you want to do anything there? If not, please close it. Then we can merge this PR.

zhiltsov-max (approved):
Ok, it seems like this PR is not blocked anymore.

cvat/apps/quality_control/quality_reports.py
```diff
@@ -1000,6 +1004,21 @@ def _match_ann_type(self, t, *args):
         else:
             return None
 
+    def match_labels(self, item_a, item_b):
+        def label_distance(a, b):
+            if a is None or b is None:
+                return 0
+            return 0.5 + (a.label == b.label) / 2
```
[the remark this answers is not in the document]
zhiltsov-max: @coderabbitai, returning 0 for mismatching labels would be conflicting with the `match_segments()` logic. We need to get label mismatches for existing mismatching labels, while returning 0 would classify them into missing labels. Not that it was entirely incorrect, but this is the implementation decision for now in this PR.

zhiltsov-max:

Viditagarwal7479:
> @Viditagarwal7479, there was the #7776 PR, do you want to do anything there? If not, please close it. Then we can merge this PR.

I will give it one more try and try to solve it.

Issue #7394, opened 2024-01-24
Incorrect display of images extracted from rotated video (no zip chunks)


phausamann:
### Actions before raising this issue

- [X] I searched the existing issues and did not find anything similar.
- [X] I read/searched [the docs](https://github.com/cvat-ai/cvat/tree/master#documentation)

### Steps to Reproduce

1. Create a video with a rotation record in the metadata, e.g. with `ffmpeg -i <input_video> -c copy -metadata:s:v:0 rotate=90 <output_video>`
2. Create a new task from the rotated video. Uncheck "prefer zip chunks"
3. Wait for the task to be created and open any job

### Expected Behavior

The video images should be displayed correctly, taking into account the rotation. Instead, the image dimensions are correct, but the displayed image is warped:

![image](https://github.com/opencv/cvat/assets/4687996/c686fca5-730d-4562-b459-65d15c3a244f)

Note that this doesn't occur when "prefer zip chunks" is checked.

### Possible Solution

This seems to be a frontend issue, introduced in v2.6.0. My suspicion is this PR: https://github.com/opencv/cvat/pull/6585

The extracted images themselves are not affected. Downgrading the `cvat_ui` image version to v2.5.2 fixes the issue when opening the same task.

### Environment

```Markdown
- Git hash a33f7f57, but can be reproduced with all versions >=2.6.0
- Ubuntu 20.04
- Docker version 24.0.1
```

bsekachev:
Hello,
Just keep in mind, that you may perform rotation in CVAT

![image](https://github.com/opencv/cvat/assets/40690378/2fbd1d80-0978-4a5b-9f93-d0dd22b59b1f)

tahamukhtar20:
@nmanovic @bsekachev please assign me this issue.

tahamukhtar20:
I didn't post a solution before for review, it skipped my mind that I had to do that, please let me know if there are any issues with this, and I'll reimplement that. 
[...]

PR #7583, opened 2024-03-10
[GSoC2024] fix warped image loading

tahamukhtar20:
### Motivation and context

Fixes: #7394 

This PR fixes the issue stated in the above-mentioned issue, the issue was caused when a 90 deg-rotated video was added with "Prefer Zip Chunks" disabled, and the image loaded was warped, which is now fixed using the static cropImage function found in this [commit](https://github.com/opencv/cvat/commit/844a72b82c241c2f612debc113de54b2f3612e7e#diff-2ca4b22442f076f5f5c3086ee757649fbaa903a75baca70991e419e2bd1d822fR281) as this was the commit where this issue initially started

### How has this been tested?

Added various images to test the bug fix, this did not require any new test cases.

### Checklist

### License

### DIFF
--- a/changelog.d/20240326_195957_tahamukhtar20.md
+++ b/changelog.d/20240326_195957_tahamukhtar20.md
@@ -1,0 +1,4 @@
+### Fixed
+
+- 90 deg-rotated video was added with "Prefer Zip Chunks" disabled was warped, fixed using the static cropImage function.
+  (<https://github.com/opencv/cvat/pull/7583>)
--- a/changelog.d/20240326_195957_tahamukhtar20.md
+++ b/changelog.d/20240326_195957_tahamukhtar20.md
@@ -1,4 +1,5 @@
 ### Fixed
 
-- 90 deg-rotated video was added with "Prefer Zip Chunks" disabled was warped, fixed using the static cropImage function.
+- 90 deg-rotated video was added with "Prefer Zip Chunks" disabled 
+was warped, fixed using the static cropImage function.
   (<https://github.com/opencv/cvat/pull/7583>)
--- a/cvat-data/src/ts/cvat-data.ts
+++ b/cvat-data/src/ts/cvat-data.ts
@@ -213,14 +213,13 @@ export class FrameDecoder {
         const source = new Uint32Array(imageBuffer);
 
         const bufferSize = width * height * 4;
-        const buffer = new ArrayBuffer(bufferSize);
-        const rgbaInt32 = new Uint32Array(buffer);
-        const rgbaInt8Clamped = new Uint8ClampedArray(buffer);
-
         if (imageWidth === width) {
             return new ImageData(new Uint8ClampedArray(imageBuffer, 0, bufferSize), width, height);
         }
 
+        const buffer = new ArrayBuffer(bufferSize);
+        const rgbaInt32 = new Uint32Array(buffer);
+        const rgbaInt8Clamped = new Uint8ClampedArray(buffer);
         let writeIdx = 0;
         for (let row = 0; row < height; row++) {
             const start = row * imageWidth;
--- a/cvat-data/src/ts/cvat-data.ts
+++ b/cvat-data/src/ts/cvat-data.ts
@@ -200,6 +200,39 @@ export class FrameDecoder {
         return null;
     }
 
+    static cropImage(
+        imageBuffer: ArrayBuffer,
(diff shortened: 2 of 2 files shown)


tahamukhtar20:
I didn't post a solution before for review, it skipped my mind that I had to do that, please let me know if there are any issues with this, and I'll reimplement that.
I've already implemented this solution in the PR attached above, here is how I solved it currently, I noticed that this was working fine in v-2.5.1 and not in v-2.6.2 so I checked out the commits in between both versions and identified that the culprit was this piece of cvat-data code change in this [commit](https://github.com/opencv/cvat/commit/844a72b82c241c2f612debc113de54b2f3612e7e#diff-2ca4b22442f076f5f5c3086ee757649fbaa903a75baca70991e419e2bd1d822fR281). Now what I did was identify what was missing here, I noticed that imageBitmap was now being used across the entire application but something was missing that was in the previous PR that wasn't in this. which was that the crop image function's code was completely missing and that was why the issue was happening in the first place. So I added that in the code base in a way that nothing else was affected and tested it out in various environments making sure it worked for all types of images

tahamukhtar20:
@bsekachev Sorry for pinging you like this, I think you should also have a look at this Pull Request as the piece of code that I changed here was directly changed by you a few months back.

cvat-data/src/ts/cvat-data.ts
```diff
@@ -239,8 +272,15 @@ export class FrameDecoder {
...
                     const width = Math.round(this.renderWidth / scaleFactor);
 
-                    const array = new Uint8ClampedArray(e.data.buf.slice(0, width * height * 4));
-                    createImageBitmap(new ImageData(array, width)).then((bitmap) => {
+                    createImageBitmap(FrameDecoder.cropImage(
+                        e.data.buf,
+                        e.data.width,
+                        e.data.height,
+                        0,
+                        0,
```
bsekachev, line 280: - We do not need constant parameters.
- Function needs to be rewritten based on this constant values (code may be reduced). Is not it?

tahamukhtar20: Thanks for the response, I'll look into it.

tahamukhtar20: @bsekachev please have a look now.

nmanovic:
@SpecLad , could you please help with the review?

cvat-data/src/ts/cvat-data.ts
```diff
@@ -200,6 +200,37 @@ export class FrameDecoder {
...
+        const bufferSize = width * height * 4;
+        const buffer = new ArrayBuffer(bufferSize);
+        const rgbaInt32 = new Uint32Array(buffer);
+        const rgbaInt8Clamped = new Uint8ClampedArray(buffer);
+
+        if (imageWidth === width) {
+            return new ImageData(new Uint8ClampedArray(imageBuffer, 0, bufferSize), width, height);
+        }
+
+        let writeIdx = 0;
```
bsekachev, line 224:
```suggestion
        if (imageWidth === width) {
            return new ImageData(new Uint8ClampedArray(imageBuffer, 0, bufferSize), width, height);
        }

        const buffer = new ArrayBuffer(bufferSize);
        const rgbaInt32 = new Uint32Array(buffer);
        const rgbaInt8Clamped = new Uint8ClampedArray(buffer);
        let writeIdx = 0;
```

bsekachev: Perhaps we only need to create buffer when we are going to use it

bsekachev:
@tahamukhtar20 

Could you please check changelog entry?

``scriv create --edit``

tahamukhtar20:
@bsekachev added the changelog entry.

changelog.d/20240326_195957_tahamukhtar20.md
```diff
@@ -0,0 +1,4 @@
+### Fixed
+
+- 90 deg-rotated video was added with "Prefer Zip Chunks" disabled was warped, fixed using the static cropImage function.
```
bsekachev, line 3:
```suggestion
- 90 deg-rotated video was added with "Prefer Zip Chunks" disabled 
was warped, fixed using the static cropImage function.
```

bsekachev (approved):

bsekachev:
Great, thank you for the contribution!

PR #7619, opened 2024-03-15
quality_control: refactor confusion matrix initialization

SpecLad:
### Motivation and context

The existing code creates a row and column for each category in the dataset, plus an extra row/column at the end for unmatched annotations. However, it's not immediately clear what order these rows/columns are in, because all access is done indirectly through the `confusion_matrix_labels_rmap` dictionary.

IMO, this is an unnecessary level of indirection, which obscures the structure of the matrix. It's easier to just assign rows/columns 0..(N-1) to the categories and N to "unmatched".

Do just that. Additionally, factor out the code creating the matrix into a function, because it's used in two places.

### How has this been tested?

Manual testing.

### Checklist

### License

### DIFF
--- a/cvat/apps/quality_control/quality_reports.py
+++ b/cvat/apps/quality_control/quality_reports.py
@@ -1874,17 +1874,7 @@ class DatasetComparator:
         invalid_labels_count = len(mismatches)
         total_labels_count = valid_labels_count + invalid_labels_count
 
-        confusion_matrix_labels = {
-            i: label.name
-            for i, label in enumerate(self._gt_dataset.categories()[dm.AnnotationType.label])
-            if not label.parent
-        }
-        confusion_matrix_labels[None] = "unmatched"
-        confusion_matrix_labels_rmap = {k: i for i, k in enumerate(confusion_matrix_labels.keys())}
-        confusion_matrix_label_count = len(confusion_matrix_labels)
-        confusion_matrix = np.zeros(
-            (confusion_matrix_label_count, confusion_matrix_label_count), dtype=int
-        )
+        confusion_matrix_labels, confusion_matrix = self._make_zero_confusion_matrix()
         for gt_ann, ds_ann in itertools.chain(
             # fully matched annotations - shape, label, attributes
             matches,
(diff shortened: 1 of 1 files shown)


cvat/apps/quality_control/quality_reports.py
```diff
@@ -2025,15 +2012,11 @@ def generate_report(self) -> ComparisonReport:
         label_recalls = _arr_div(matched_ann_counts, gt_ann_counts)
 
         valid_annotations_count = np.sum(matched_ann_counts)
-        missing_annotations_count = np.sum(confusion_matrix[confusion_matrix_labels_rmap[None], :])
-        extra_annotations_count = np.sum(confusion_matrix[:, confusion_matrix_labels_rmap[None]])
+        missing_annotations_count = np.sum(confusion_matrix[-1, :])
```
zhiltsov-max, line 2015: Why the matrix accumulation code was removed?

cvat/apps/quality_control/quality_reports.py
```diff
@@ -1906,15 +1896,11 @@ def _find_closest_unmatched_shape(shape: dm.Annotation):
...
```
zhiltsov-max, line 1899: It seems like now this -1 is a good candidate for a named constant, as it's used in several places.

cvat/apps/quality_control/quality_reports.py
```diff
@@ -1994,27 +1994,14 @@ def generate_report(self) -> ComparisonReport:
...
 
-        confusion_matrix_labels = {
-            i: label.name
-            for i, label in enumerate(self._gt_dataset.categories()[dm.AnnotationType.label])
-            if not label.parent
-        }
-        confusion_matrix_labels[None] = "unmatched"
-        confusion_matrix_labels_rmap = {k: i for i, k in enumerate(confusion_matrix_labels.keys())}
-        if confusion_matrices:
-            confusion_matrix = np.sum(confusion_matrices, axis=0)
```
zhiltsov-max, line 2013: Why the matrix accumulation code was removed?

SpecLad: It's not removed, it's just done in the loop above now.

zhiltsov-max: It's less efficient in the new variant, I suppose. There can be quite a few frames in the task, and the old variant summed them with numpy.

cvat/apps/quality_control/quality_reports.py
SpecLad: Yeah, fair point. Replaced with a constant.

cvat/apps/quality_control/quality_reports.py
SpecLad: Okay, I did a little benchmark, and at matrix sizes where the time taken non-negligible the new variant is faster. 🤷

zhiltsov-max: Looks strange, but if so, okay. Maybe it's better to just include this in the `accumulate()` then. There is also a parallel [PR](https://github.com/opencv/cvat/pull/7604) that touches this topic.

zhiltsov-max (approved):

PR #7623, opened 2024-03-17
[GSoC2024] added filter to not include gt_job when the test intends to create one.

### LINKED ISSUE #7622: Some tests create gt_job for tasks which already have a gt_job

### Steps to Reproduce

1. Add a new task in the testing database, which has gt_job
2. Run tests

### Expected Behavior

1. The test should pass

### Possible Solution

As mentioned in this [comment](https://github.com/opencv/cvat/pull/7604#issuecomment-2002166466), tests should check whether the task they are picking up has gt_job or not when they intend to create one or require a specific kind of gt_job which needs to be created.

### Context

I created this [PR](https://github.com/opencv/cvat/pull/7604) in which I added a new task in the test database. After this, some tests which fall in the above category started to fail.

### Environment

_No response_

Viditagarwal7479:
Fixes #7622 

Added a check to ensure that tests which need to create a gt_job aren't picking up tasks which already have a gt_job

### How has this been tested?
I added a new task to the testing database and ran these updated tests.

### Checklist

~- [ ]  I have increased versions of npm packages if it is necessary

### License


zhiltsov-max (approved):

tests/python/rest_api/test_jobs.py
```diff
@@ -728,6 +732,7 @@ def _get_or_create_gt_job(self, user, task_id, frames):
         with make_api_client(user) as api_client:
             (task_jobs, _) = api_client.jobs_api.list(task_id=task_id, type="ground_truth")
             if task_jobs.results:
+                # TODO: to modify this to only get gt_job only with the mentioned frames
```
zhiltsov-max, line 735: I think it's better to add an assert here that there is no existing GT job, if specific frames are requested.

zhiltsov-max (changes requested):
Please resolve the comment, and it can be merged.

Viditagarwal7479:
The `jobs.json` and `tasks.json` files include information about `admin1` only. In the above test, it's making API requests as `admin2` because of which there is a mismatch in validating whether `gt_job` exists for that task_id. 
[here](https://github.com/Viditagarwal7479/cvat/blob/2d85a435982057cce2a981a4743272b2880eaee8/tests/python/rest_api/test_jobs.py#L591) it's checking presence of `gt_job` through the json files
[here](https://github.com/Viditagarwal7479/cvat/blob/2d85a435982057cce2a981a4743272b2880eaee8/tests/python/rest_api/test_jobs.py#L733) it's checking presence of `gt_job` through API request

Viditagarwal7479:
Probably, there's some other issue,
Only entry in `jobs.json` for `user: admin1` and `task_id:5` :
```json
    {
      "assignee": {
        "first_name": "Worker",
        "id": 9,
        "last_name": "Fourth",
        "url": "http://localhost:8080/api/users/9",
        "username": "worker4"
      },
      "bug_tracker": null,
      "created_date": "2022-02-16T06:25:48.168000Z",
      "data_chunk_size": 72,
      "data_compressed_chunk_type": "imageset",
      "dimension": "2d",
      "frame_count": 25,
      "guide_id": null,
      "id": 7,
      "issues": {
        "count": 1,
        "url": "http://localhost:8080/api/issues?job_id=7"
      },
      "labels": {
        "url": "http://localhost:8080/api/labels?job_id=7"
      },
      "mode": "interpolation",
      "organization": null,
      "project_id": null,
      "source_storage": null,
      "stage": "annotation",
      "start_frame": 0,
      "state": "in progress",
      "status": "annotation",
      "stop_frame": 24,
      "target_storage": null,
      "task_id": 5,
      "type": "annotation",
      "updated_date": "2022-06-22T09:18:45.296000Z",
      "url": "http://localhost:8080/api/jobs/7"
    },
```
But in the [API request](https://github.com/Viditagarwal7479/cvat/blob/2d85a435982057cce2a981a4743272b2880eaee8/tests/python/rest_api/test_jobs.py#L733) 
```py
{'assignee': None,
 'bug_tracker': None,
 'created_date': datetime.datetime(2024, 3, 18, 14, 2, 41, 301446, tzinfo=tzutc()),
 'data_chunk_size': 72,
 'data_compressed_chunk_type': 'imageset',
 'dimension': '2d',
 'frame_count': 4,
 'guide_id': None,
 'id': 32,
 'issues': {'count': 0, 'url': 'http://localhost:8080/api/issues?job_id=32'},
 'labels': {'url': 'http://localhost:8080/api/labels?job_id=32'},
 'mode': 'interpolation',
 'organization': None,
 'project_id': None,
 'source_storage': None,
 'stage': 'annotation',
 'start_frame': 0,
 'state': 'new',
 'status': 'annotation',
 'stop_frame': 24,
 'target_storage': None,
 'task_id': 5,
 'type': 'ground_truth',
 'updated_date': datetime.datetime(2024, 3, 18, 14, 2, 41, 301465, tzinfo=tzutc()),
 'url': 'http://localhost:8080/api/jobs/32'}
```
I tried doing this multiple times by restarting the server, but the result remained the same.

zhiltsov-max:
Both admins should have equal access rights for any server objects, they should see the same (all) jobs and tasks.

Viditagarwal7479:
YES!!! I figured it out!
The tests that are previous to the failing tests are creating `gt_job` corresponding to a `task_id`. Due to this the current session has `gt_job` for that `task_id` but the `jobs.json` doesn't has any information regarding this.

zhiltsov-max:
Great news! Please add `@pytest.mark.usefixtures("restore_db_per_function")` to these tests to fix the problem.

Viditagarwal7479:
Wow, that's it! I am trying to fix this for quite a while and came up with this,
```py
(_, gt_job_del_response) = api_client.jobs_api.destroy(gt_job.id)
assert gt_job_del_response.status == HTTPStatus.NO_CONTENT, "newly created gt_job couldn't be deleted"
```
was just about to push the changes and saw your message. Thank You ;)

Viditagarwal7479:
[`test_can_get_gt_job_chunk`](https://github.com/Viditagarwal7479/cvat/blob/2d85a435982057cce2a981a4743272b2880eaee8/tests/python/rest_api/test_jobs.py#L680) is still failing after adding 
```py 
@pytest.mark.usefixtures("restore_db_per_function")
```
but passes on using 
```py
(_, gt_job_del_response) = api_client.jobs_api.destroy(gt_job.id)
assert gt_job_del_response.status == HTTPStatus.NO_CONTENT, "newly created gt_job couldn't be deleted"
```

zhiltsov-max:
Hi, yes, it seems like these specific tests have an issue with server data cache, as the [comment](https://github.com/opencv/cvat/blob/53bf3500edcf34ce39a22c444b1e81ec24a9b7d5/tests/python/rest_api/test_jobs.py#L571-L573) says. I think we need to think if we can improve these tests somehow to avoid doing hacks.

Viditagarwal7479:
Ahh!
Could you explain this line?

> if we don't clean the db, the gt jobs created will be reused, and their ids won't conflict

How about instead of restoring the db this way, we just delete the newly created gt_job through the API call?

Viditagarwal7479:
Ah, `request.addfinalizer()` this approach seems great to me.
Should I make these changes?

zhiltsov-max:
Yes, please update the tests relying on optional gt job creation.

Viditagarwal7479:
I also observed that [`test_can_get_gt_job_chunk`](https://github.com/opencv/cvat/blob/f513aa162ba2abfeb6625408ae4b35ae6223c180/tests/python/rest_api/test_jobs.py#L677) is non-determinant. It sometimes passes and, at times, fails at `original-interpolation` (mostly) and rarely at `original-annotation`.

Here's some output that I recorded when `original-interpolation` failed,
```py
print(set(chunk.namelist()))
>>> {'000000.jpeg', '000008.jpeg', '000006.jpeg', '000003.jpeg', '000007.jpeg', '000004.jpeg', '000002.jpeg', '000001.jpeg', '000005.jpeg', '000010.jpeg', '000009.jpeg'}
print(set("{:06d}.jpeg".format(i) for i in frame_range))
>>> {'000000.jpeg', '000017.jpeg', '000011.jpeg', '000024.jpeg', '000008.jpeg', '000016.jpeg', '000001.jpeg', '000020.jpeg', '000004.jpeg', '000002.jpeg', '000012.jpeg', '000015.jpeg', '000006.jpeg', '000009.jpeg', '000021.jpeg', '000022.jpeg', '000023.jpeg', '000010.jpeg', '000005.jpeg', '000014.jpeg', '000018.jpeg', '000003.jpeg', '000007.jpeg', '000013.jpeg', '000019.jpeg'}
print(task_meta)
>>> {'chunk_size': 72,
 'deleted_frames': [],
 'frame_filter': '',
 'frames': [{'has_related_context': False,
             'height': 720,
             'name': 'test_video_1.mp4',
             'related_files': 0,
             'width': 1280}],
 'image_quality': 70,
 'included_frames': None,
 'size': 25,
 'start_frame': 0,
 'stop_frame': 24}
```
Here I suspect that, as per the code [here](https://github.com/opencv/cvat/blob/f513aa162ba2abfeb6625408ae4b35ae6223c180/tests/python/rest_api/test_jobs.py#L704) the `chunk_size` is `72` thus it will select all the frames the task has.

tests/python/rest_api/test_jobs.py
```diff
@@ -567,16 +569,18 @@ def test_get_gt_job_in_org_task(
             self._test_get_job_403(user["username"], job["id"])
 
 
-@pytest.mark.usefixtures(
-    # if the db is restored per test, there are conflicts with the server data cache
-    # if we don't clean the db, the gt jobs created will be reused, and their
-    # ids won't conflict
-    "restore_db_per_class"
-)
```
zhiltsov-max, line 575: Why these lines were removed?

zhiltsov-max, line 575: Why these lines were removed?

Viditagarwal7479: This was restoring the db state. But in the tests here we are ensuring that the db state is restored by deleting the new objects created (which is only gt job here). So thought to remove, but if you say that there might be risk of db state getting changed I shall again add it.

tests/python/rest_api/test_jobs.py
```diff
@@ -574,9 +576,17 @@ def test_get_gt_job_in_org_task(
...
-    @pytest.mark.usefixtures("restore_db_per_function")
+
+    def delete_gt_job(self, user, gt_job_id):
+        with make_api_client(user) as api_client:
+            (_, gt_job_del_response) = api_client.jobs_api.destroy(gt_job_id)
+            assert (
+                gt_job_del_response.status == HTTPStatus.NO_CONTENT
+            ), "newly created gt_job couldn't be deleted"
+
+    # @pytest.fixture
```
zhiltsov-max, line 587:
```suggestion
```

tests/python/rest_api/test_jobs.py
```diff
@@ -574,9 +576,17 @@ def test_get_gt_job_in_org_task(
     "restore_db_per_class"
 )
 class TestGetGtJobData:
-    @pytest.mark.usefixtures("restore_db_per_function")
+
+    def delete_gt_job(self, user, gt_job_id):
```
zhiltsov-max, line 580:
```suggestion
    def _delete_gt_job(self, user, gt_job_id):
```

tests/python/rest_api/test_jobs.py
```diff
@@ -600,6 +611,8 @@ def test_can_get_gt_job_meta(self, admin_user, tasks, task_mode):
         with make_api_client(user) as api_client:
             (gt_job_meta, _) = api_client.jobs_api.retrieve_data_meta(gt_job.id)
 
+        request.addfinalizer(lambda: self.delete_gt_job(user, gt_job.id))
```
zhiltsov-max, line 614: I think it's better to put right after the `_get_or_create_gt_job()` call in each case.

tests/python/rest_api/test_jobs.py
```diff
@@ -728,6 +745,7 @@ def _get_or_create_gt_job(self, user, task_id, frames):
         with make_api_client(user) as api_client:
             (task_jobs, _) = api_client.jobs_api.list(task_id=task_id, type="ground_truth")
             if task_jobs.results:
+                assert not frames, "specific frames requested but gt_job exists"
```
zhiltsov-max, line 748: It makes sense to split the `self._get_or_create_gt_job` function in 2 now to exclude the possibility of incorrect invocation. Please split, update the code, and this PR seems to be finished.

tests/python/rest_api/test_jobs.py
Viditagarwal7479: Some functions later on fetch the metadata of this newly created gt_job. So, this change might cause problem.

zhiltsov-max: Finalizers are called on the test tear down. Which potential problems do you see?

tests/python/rest_api/test_jobs.py
```diff
@@ -724,11 +742,11 @@ def test_can_get_gt_job_chunk(self, admin_user, tasks, task_mode, quality):
                     assert image.size > (1, 1)
                     assert np.any(image_data != 0)
 
-    def _get_or_create_gt_job(self, user, task_id, frames):
+    def _create_gt_job(self, user, task_id, frames):
         with make_api_client(user) as api_client:
             (task_jobs, _) = api_client.jobs_api.list(task_id=task_id, type="ground_truth")
             if task_jobs.results:
-                gt_job = task_jobs.results[0]
+                raise RuntimeError(f"gt_job already exist for task {task_id}")
```
zhiltsov-max, line 749: There is no need to check this additionally, the job creation request will fail with an error if there is a gt job already.

tests/python/rest_api/test_jobs.py
Viditagarwal7479: Ohh is that so, sorry I thought they are called in the order they are called.

Viditagarwal7479:
It's showing "1 review requesting changes by reviewers with write access" but I have incorporated all the requested changes so far.

tests/python/rest_api/test_jobs.py
```diff
@@ -574,9 +576,16 @@ def test_get_gt_job_in_org_task(
...
 )
 class TestGetGtJobData:
-    @pytest.mark.usefixtures("restore_db_per_function")
+
+    def _delete_gt_job(self, user, gt_job_id):
+        with make_api_client(user) as api_client:
+            (_, gt_job_del_response) = api_client.jobs_api.destroy(gt_job_id)
+            assert (
+                gt_job_del_response.status == HTTPStatus.NO_CONTENT
+            ), "newly created gt_job couldn't be deleted"
```
zhiltsov-max, line 585:
```suggestion
            api_client.jobs_api.destroy(gt_job_id)
```

The call will raise an exception if the request fails.

tests/python/rest_api/test_jobs.py
```diff
@@ -710,6 +726,8 @@ def test_can_get_gt_job_chunk(self, admin_user, tasks, task_mode, quality):
         # with placeholders in the frames outside the job.
         # This is required by the UI implementation
         with zipfile.ZipFile(chunk_file) as chunk:
+            print(set(chunk.namelist()))
+            print(set("{:06d}.jpeg".format(i) for i in frame_range))
```
zhiltsov-max, line 730:
```suggestion
```

zhiltsov-max (approved):

zhiltsov-max:

tests/python/rest_api/test_jobs.py
Viditagarwal7479: Oh thanks was unaware about it.

tests/python/rest_api/test_jobs.py
Viditagarwal7479: Ah, missed them while debugging 😅

Viditagarwal7479:
Thank you! For helping me through out.

PR #7624, opened 2024-03-17
[GSoC2024] Implemented drag and drop feature

### LINKED ISSUE #7370: Can't set order attributes

1. go task or project constructor raw
2. change attributes order

test1, test2, test3, test4, test5 -> test1, test2, test3, test5, test4

image

3.  click done -> i think this makes sorting attributes by attributes id

test1, test2, test3, test5, test4  -> test1, test2, test3, test4, test5

image

I want to change attributes order

When I change constructor list then, state is continuously maintained, so that it is also applied when labeling page

image

test1, test2, test3, test4, test5 -> test1, test2, test3, test5, test4

image

I think it can make enhance labeling speed

This might not be an alternative, but rather a feature development, but I'll mention it anyway.
How about adding a function to change the order in the Constructor settings?
There is a UI example in the following [link](https://www.darins.page/articles/designing-a-reorderable-list-component) and image.

image

### Additional context

Thank you

KrishavRajSingh:
Added the feature of dragging attributes 

### Motivation and context

resolves #7370

### Checklist

- [*] I submit my changes into the `develop` branch
- [*] I have created a changelog fragment 
- [*] I have linked related issues (see [GitHub docs](

### License

## Summary by CodeRabbit

- **New Features**
  - Introduced drag-and-drop functionality for label attributes in the label editor.

- **Style**
  - Updated styles to support drag-and-drop handles with a pointer cursor.


klakhov (commented):
Generally im not sure if this solution is good in UX terms. Design with arrows is a bit outdated and not so comfortable to use (if I want to place 10th attribute to the first place I need to make 10 clicks :worried:). 
Lets return to original proposal, as the author of [7370](https://github.com/opencv/cvat/issues/7370) mentiones, it sould be implemented as drag-n-drop. I belive we need to use dots icon that is shown only on hover. 
It should be something similar to:
list
In CVAT, we use react-grid-layout to help us implement draggable and responsible layouts on [analytics page](https://github.com/opencv/cvat/blob/develop/cvat-ui/src/components/analytics-page/analytics-performance.tsx) and [canvas-layout](https://github.com/opencv/cvat/blob/develop/cvat-ui/src/components/annotation-page/canvas/grid-layout/canvas-layout.tsx). I would suggest to implement this feature using specific rows-only draggable layout.


Also, please, add a link to the issue you're trying to resolve in PR description. And dont forget to add a changelog enty.

KrishavRajSingh:
Ok, I will make the changes

KrishavRajSingh:
> Generally im not sure if this solution is good in UX terms. Design with arrows is a bit outdated and not so comfortable to use (if I want to place 10th attribute to the first place I need to make 10 clicks 😟). Lets return to original proposal, as the author of [7370](https://github.com/opencv/cvat/issues/7370) mentiones, it sould be implemented as drag-n-drop. I belive we need to use dots icon that is shown only on hover. It should be something similar to: list [ list ](https://private-user-images.githubusercontent.com/50956430/313683472-2619dace-9433-4878-a686-f07b09e7d01c.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTA5MjIxMzQsIm5iZiI6MTcxMDkyMTgzNCwicGF0aCI6Ii81MDk1NjQzMC8zMTM2ODM0NzItMjYxOWRhY2UtOTQzMy00ODc4LWE2ODYtZjA3YjA5ZTdkMDFjLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDAzMjAlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwMzIwVDA4MDM1NFomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWI1NWNmNWI2NGUwYWQxNWMzMDgwODg2Yzg4ODVjOGUxNTgxZjQ1Njc2OWViZWU3MjQxNWIzMDcwYzliZjFkMzgmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.yp5plSNv1OWt0wC2d3XUc6TOl7Tb52IOdw0PmurhGJI) [ ](https://private-user-images.githubusercontent.com/50956430/313683472-2619dace-9433-4878-a686-f07b09e7d01c.gif?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MTA5MjIxMzQsIm5iZiI6MTcxMDkyMTgzNCwicGF0aCI6Ii81MDk1NjQzMC8zMTM2ODM0NzItMjYxOWRhY2UtOTQzMy00ODc4LWE2ODYtZjA3YjA5ZTdkMDFjLmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDAzMjAlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwMzIwVDA4MDM1NFomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWI1NWNmNWI2NGUwYWQxNWMzMDgwODg2Yzg4ODVjOGUxNTgxZjQ1Njc2OWViZWU3MjQxNWIzMDcwYzliZjFkMzgmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.yp5plSNv1OWt0wC2d3XUc6TOl7Tb52IOdw0PmurhGJI) In CVAT, we use react-grid-layout to help us implement draggable and responsible layouts on [analytics page](https://github.com/opencv/cvat/blob/develop/cvat-ui/src/components/analytics-page/analytics-performance.tsx) and [canvas-layout](https://github.com/opencv/cvat/blob/develop/cvat-ui/src/components/annotation-page/canvas/grid-layout/canvas-layout.tsx). I would suggest to implement this feature using specific rows-only draggable layout.
> 
> Also, please, add a link to the issue you're trying to resolve in PR description. And dont forget to add a changelog entry.

Hey I think react-grid-layout doesn't works on Forms.

klakhov:
What do you mean by doesnt work on Forms? 
I dont think it can work directly on the the Form.Item components. As usage examples of react-grid-layouts shows, we need to pass just a div inside of it. My first thoughts of the solution are: we can wrap Form.Item in div and then manage layout somewhere in the state of the label form component. At least [antd form docs](https://ant.design/components/form#formlist) show that it should be possible. Probably we need to refactor the label form component as its heavily dependent on antd form so we can manage attributes layout correctly.

Could you describe the soultion you are trying to implement so we can discuss it?

KrishavRajSingh:
> What do you mean by doesnt work on Forms? I dont think it can work directly on the the Form.Item components. As usage examples of react-grid-layouts shows, we need to pass just a div inside of it. My first thoughts of the solution are: we can wrap Form.Item in div and then manage layout somewhere in the state of the label form component. At least antd form docs [here](https://ant.design/components/form#formlist) and [here](https://ant.design/components/form#why-formitem-not-update-value-when-children-is-nest) show that it should be possible. Probably we need to refactor the label form component as its heavily dependent on antd form so we can manage attributes layout correctly.
> 
> Could you describe the soultion you are trying to implement so we can discuss it?

I was trying to implement the same but after wrapping with <ReactGridLayout>, this.renderAttributes was not returning anything. Most probably because fieldInstances was empty when calling from <ReactGridLayout>.
This is a gist of the changes I made <script src="https://gist.github.com/KrishavRajSingh/0b7c9ca11b75bd677413e0be0b055702.js"></script>

klakhov:
I think we dont need to pass `Form.List` into the grid, instead we should create a grid inside of a list. As you can see, `Form.List` expects a function that will render a list of `Form.Items`. We already use this function to render a plain list, but instead of that we can firstly create a layout and then return a react grid layout component. [See the example](https://pastebin.com/KiwPnvQW). The thing is that we need to manage layout correcly in the component to keep the order of attributes in list. Also, need to work with the styles.

KrishavRajSingh:
> I think we dont need to pass `Form.List` into the grid, instead we should create a grid inside of a list. As you can see, `Form.List` expects a function that will render a list of `Form.Items`. We already use this function to render a plain list, but instead of that we can firstly create a layout and then return a react grid layout component. [See the example](https://pastebin.com/KiwPnvQW). The thing is that we need to manage layout correcly in the component to keep the order of attributes in list. Also, need to work with the styles.

Hey, I have almost completed it but I am facing a problem. The thing is when we change the layout, I update the attributes list and this layout updates based on this new attributes list. These were the changes that I have made: [https://pastebin.com/V9qYW46u](url). 
[Screencast from 23-03-24 04:34:48 PM IST.webm](https://github.com/opencv/cvat/assets/99531396/aa189389-6199-4f6a-b9a1-972b9c3a297e)

klakhov:
Hi @KrishavRajSingh , As I can see you are trying to update form after layout changes and then it re-renders. I think we can fix this problem by managing layout somewhere outside of the form. There should be some state for the layout. By default it should be empty, then on first render of the form we set it up. If user makes some changes, we set the state of the layout object. When the form is submitted (`handleSubmit` func) we can use the layout state to provide correct order of the attributes to the `onSubmit` func.

bsekachev:
Also, I would recommend to look at sorting component. It may be found on projects/tasks/cloud storages pages
Drag & Drop are implemented there

image

And this solution will not work when attributes were already saved on the server.

klakhov:
Hi @KrishavRajSingh,
I was trying to invistigate why **ui-only fix** works for this issue in the first place. A brief look over the CVAT server, shows that there are no ordering fields, updated date, or explicit ordering in queryset for the Attribute model. There is no `ORDER BY` in sql queries to DB and no sorting in django. But still the experiments show that attributes are returned from the server with implicit sorting by `updated_date`(even we dont really have such field on our side). That means the fix works only because of some kind of optimization on DB side(which maybe caches rows in order of updation, I dont really know).  The fact that it uses some ordering is undocumented and cant be relied on.
So it seems we need to extend the solution with more reliable approach to sorting.. We need something on the server side to support the ui. Can you think about it? Maybe we need to introduce ordering field, maybe we should add updated_date to attribute and sort by it.

KrishavRajSingh:
I am looking into this.

KrishavRajSingh:
> Contributor

Thanks for this, I was finally able to implement this drag&drop.

https://github.com/opencv/cvat/assets/99531396/c2f8a07f-0caf-4d8a-adfb-8fccbae96948

KrishavRajSingh:
> Hi @KrishavRajSingh,
> I was trying to invistigate why **ui-only fix** works for this issue in the first place. A brief look over the CVAT server, shows that there are no ordering fields, updated date, or explicit ordering in queryset for the Attribute model. There is no `ORDER BY` in sql queries to DB and no sorting in django. But still the experiments show that attributes are returned from the server with implicit sorting by `updated_date`(even we dont really have such field on our side). That means the fix works only because of some kind of optimization on DB side(which maybe caches rows in order of updation, I dont really know).  The fact that it uses some ordering is undocumented and cant be relied on.
> So it seems we need to extend the solution with more reliable approach to sorting.. We need something on the server side to support the ui. Can you think about it? Maybe we need to introduce ordering field, maybe we should add updated_date to attribute and sort by it.

I think a separate issue should be opened for this. What do you think??

KrishavRajSingh:
> Hi @KrishavRajSingh, I was trying to invistigate why **ui-only fix** works for this issue in the first place. A brief look over the CVAT server, shows that there are no ordering fields, updated date, or explicit ordering in queryset for the Attribute model. There is no `ORDER BY` in sql queries to DB and no sorting in django. But still the experiments show that attributes are returned from the server with implicit sorting by `updated_date`(even we dont really have such field on our side). That means the fix works only because of some kind of optimization on DB side(which maybe caches rows in order of updation, I dont really know). The fact that it uses some ordering is undocumented and cant be relied on. So it seems we need to extend the solution with more reliable approach to sorting.. We need something on the server side to support the ui. Can you think about it? Maybe we need to introduce ordering field, maybe we should add updated_date to attribute and sort by it.

Hey can you tell me where to perform sorting?? 
Actually I am new to django and the server feels little overwhelming.

KrishavRajSingh:
> Hi @KrishavRajSingh, I was trying to invistigate why **ui-only fix** works for this issue in the first place. A brief look over the CVAT server, shows that there are no ordering fields, updated date, or explicit ordering in queryset for the Attribute model. There is no `ORDER BY` in sql queries to DB and no sorting in django. But still the experiments show that attributes are returned from the server with implicit sorting by `updated_date`(even we dont really have such field on our side). That means the fix works only because of some kind of optimization on DB side(which maybe caches rows in order of updation, I dont really know). The fact that it uses some ordering is undocumented and cant be relied on. So it seems we need to extend the solution with more reliable approach to sorting.. We need something on the server side to support the ui. Can you think about it? Maybe we need to introduce ordering field, maybe we should add updated_date to attribute and sort by it.

One more thing, if we are changing the order of attributes and saving it, why do we need to sort it based on updated_date??

klakhov:
@KrishavRajSingh 
You need to look on the [LabelView](), [LabelModel](), [AttributeModel](). For sorting refer to the [django querysets]().

> One more thing, if we are changing the order of attributes and saving it, why do we need to sort it based on updated_date??

The problem here is that the order should be based on some variable, for example it may be `updated_date`. With current implemetation there is no variable by which server sorts the attributes. It returns them from database as-is which means the order is defined by database internal implementation which can change from time to time.

bsekachev:
@klakhov 
@KrishavRajSingh 

I do not think `updated_time` is a good choice to sort attributes. 
There may be other ways to re-save attributes in the database and bump updated_date (e.g. adding a new attribute value)

Even if now we do not have such ways, in the future we may. Semantic of `updated_date` does not correspond to the way you are going to use it.

klakhov:
@bsekachev 
What would you suggest to use? Separate variable for ordering(`dislplay_order`/`order`)?

bsekachev:
If we want this feature, than yes, something like this.

KrishavRajSingh:
@klakhov @bsekachev 
Do we need this feature??
If yes, do I need to introduce a field called "display_order" instead of "updated_date" and change its value when it is updated or something else??
If no,  should I remove this updated_date as sorting field??

klakhov:
@KrishavRajSingh,
Yes, let's introduce a separate sorting field. We need to change its value based on order that is sent from UI.
Also, I think we should keep the updated_date field. Consider inheritance from the [TimestampedModel](https://github.com/cvat-ai/cvat/blob/develop/cvat/apps/engine/models.py#L318) instead of adding the field explicitly.

KrishavRajSingh:
> @KrishavRajSingh, Yes, let's introduce a separate sorting field. We need to change its value based on order that is sent from UI. Also, I think we should keep the updated_date field. Consider inheritance from the [TimestampedModel](https://github.com/cvat-ai/cvat/blob/develop/cvat/apps/engine/models.py#L318) instead of adding the field explicitly.

One more thing, the order will contain integers like 0, 1, 2, 3 which we will use to display the attributes in ui and when we add an attribute we will increase it by one. Correct??

klakhov:
@KrishavRajSingh, Yes, I suppose it sould be like this. Could you discribe what will happen when we will save new attribute order?

KrishavRajSingh:
> @KrishavRajSingh, Yes, I suppose it sould be like this. Could you discribe what will happen when we will save new attribute order?


Order need to be introduced in the frontend.
The order should be kept same when we are changing the order of attributes, only the value of attribute changes. 
image

And when a new attribute is added, the order value is increased by one.
And then the attribute can be saved with their order.

Open to suggestions if you have any other good idea.

KrishavRajSingh:
> @KrishavRajSingh, Yes, I suppose it sould be like this. Could you discribe what will happen when we will save new attribute order?

Hey I am trying to send display_order as property of the attributes but the request made does not contain it.
image

Request
image

Is there something I am missing??

KrishavRajSingh:
Do I still need updated_date as I am using display_order for the sorting field??

klakhov:
Hi @KrishavRajSingh ,
I belive we need updated_date as its good practice. Changing the order should update the model and its updated date. As Ive said earlier, consider looking on the [TimestampedModel](https://github.com/cvat-ai/cvat/blob/develop/cvat/apps/engine/models.py#L318) instead of adding the field explicitly as all our models with such field use this base class.
Please, let us know when the PR is ready, we will review the UI and server part.

bsekachev:
@klakhov 

> I belive we need updated_date as its good practice

IMHO it is worthy of a dedicated pull request if we want these fields in Attribute and Label model. 
That is not related with this one.

klakhov:
Alright, lets keep it for dedicated PR.

@KrishavRajSingh, lets remove updated_date for now.

KrishavRajSingh:
> Alright, lets keep it for dedicated PR.
> 
> @KrishavRajSingh, lets remove updated_date for now.

Removed it, you may review it.

klakhov:
@KrishavRajSingh , You need to update the shema `python manage.py spectacular > cvat/schema.yml` and push the changes in schema.yml so we can run the tests

KrishavRajSingh:
> @KrishavRajSingh , You need to update the shema `python manage.py spectacular > cvat/schema.yml` and push the changes in schema.yml so we can run the tests

Done

klakhov:
Small problem, after I change the order the drag icon becomes inactive and I cant drag once again:
Peek 2024-05-23 16-35

cvat-ui/src/components/labels-editor/label-form.tsx
```diff
@@ -5,7 +5,9 @@
 
 import React, { RefObject } from 'react';
 import { Row, Col } from 'antd/lib/grid';
-import Icon, { DeleteOutlined, PlusCircleOutlined } from '@ant-design/icons';
+import Icon, {
+    DeleteOutlined, HolderOutlined, PlusCircleOutlined,
+} from '@ant-design/icons';
```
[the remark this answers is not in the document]
klakhov: Please, remove unused imports

cvat/apps/engine/views.py
```diff
@@ -49,7 +49,7 @@
 from cvat.apps.engine.frame_provider import FrameProvider
 from cvat.apps.engine.media_extractors import get_mime
 from cvat.apps.engine.models import (
-    ClientFile, Job, JobType, Label, SegmentType, Task, Project, Issue, Data,
+    ClientFile, Job, JobType, Label, AttributeSpec, SegmentType, Task, Project, Issue, Data,
```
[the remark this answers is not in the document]
klakhov: Remove unused import.

klakhov:
Also padding after attribute list became too small. Previously we had:
image

klakhov:
Can we disable vertical resizing too? Horizontal is disabled as I can see 
Peek 2024-05-23 16-54

klakhov:
Also please check failing linters and tests

cvat/apps/engine/views.py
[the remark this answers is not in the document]
KrishavRajSingh: > Remove unused import.

I have not imported this. This is here since 3 years ago.
image
Should i remove this??

KrishavRajSingh:
> Also please check failing linters and tests

I have fixed almost everything, but I am not able to run the tests locally. Please help me with that.
Thanks.

KrishavRajSingh:
> @KrishavRajSingh To run tests locally, please refer to [our docs](https://docs.cvat.ai/docs/contributing/running-tests/). I suppose you need the part with E2E tests and Rest API tests. Also, please look on linter errors.

Hey, I am getting thia issue while generating cvat-sdk for tests.
image
image

klakhov:
Hi @KrishavRajSingh, sorry for the long response, we will try to look on the issue.

KrishavRajSingh:
> Hi @KrishavRajSingh, sorry for the long response, we will try to look on the issue.

Hey @klakhov , you may review this now. Sorry for taking this long.

KrishavRajSingh:
Also I was informed that I need to make a similar drag and drop for Annotation labels, so should I make this here or should I make a separate PR?

bsekachev:
@klakhov 

Please, look at the updates

cvat-ui/src/components/labels-editor/label-form.tsx
```diff
@@ -150,7 +171,7 @@ export default class LabelForm extends React.Component<Props> {
         const { key } = fieldInstance;
         const attrNames = this.formRef.current?.getFieldValue('attributes')
             .filter((_attr: any) => _attr.id !== attr.id).map((_attr: any) => _attr.name);
-
+        const height = window.innerWidth < 1404 ? 2 : 1;
```
klakhov, line 174: Why do we need this? It causes differences in heights in between elements.

cvat-ui/src/components/labels-editor/label-form.tsx
```diff
@@ -63,6 +74,16 @@ export default class LabelForm extends React.Component<Props> {
         const {
             label, onSubmit, onSkeletonSubmit, onCancel, resetSkeleton,
         } = this.props;
+        const orderedAttr: SerializedAttribute[] = []; // to store the updated and newly ordered attributes
+        const { layout } = this.state;
+        if (layout) {
+            // Updating attributes based on layout before sending it to server
+            const attr = values.attributes;
+            layout.forEach((obj: any, index: number) => {
```
klakhov, line 82:
```suggestion
            layout.forEach((obj: Layout, index: number) => {
```

cvat-ui/src/components/labels-editor/styles.scss
```diff
@@ -249,4 +249,21 @@ textarea.ant-input.cvat-raw-labels-viewer {
...
         }
     }
-}
\ No newline at end of file
+}
+
+.cvat-drag-drop {
+    cursor: pointer;
+    height: 2rem;
+    width: 0.5rem;
```
klakhov, line 257: Please, use `$grid-unit-size` instead of `rem`s

cvat-ui/src/components/labels-editor/label-form.tsx
```diff
@@ -228,20 +264,34 @@ export default class LabelForm extends React.Component<Props> {
         const { key } = fieldInstance;
         const locked = attr.id as number >= 0;
         const existingValues = attr.values;
-
+        const { layout } = this.state;
+        const updatedLayout = layout.map((item, index) => {
+            if (index === key) {
+                return { ...item, h: 2 };
```
klakhov, line 270: Do we use two types of `height` to correctly display the validation error?

Lets declare two const variables for the height 1 and 2 with descriptive names

cvat-ui/src/components/labels-editor/label-form.tsx
```diff
@@ -228,20 +264,34 @@ export default class LabelForm extends React.Component<Props> {
...
             }
 
             for (const value of values) {
                 if (!patterns.validateAttributeValue.pattern.test(value)) {
+                    this.setState({ layout: updatedLayout });
                     return Promise.reject(new Error(`Invalid attribute value: "${value}"`));
                 }
             }
-
+            const updatedLayout1 = layout.map((item, index) => {
```
klakhov, line 288: Lets declare a function, something like `createLayout(height)` so we dont repeat this code.
Also please, use something more meaningful than `updatedLayout` and `updatedLayout1`. defaultLayout/errorLayout

klakhov:
- There is a bug with textarea:
textarea-bug

- Lets make dots icon smaller, and with less padding. Now it takes too much space. Something like
image

- If I move attribute away from the page I see x-scrollbar, I dont think we need it:
image

- When I click 'Add attribute' the from element is added, but when I start typing there is some strange padding appears(It seems it happens when my page width is small):
strange padding

cvat-ui/src/components/labels-editor/label-form.tsx
KrishavRajSingh: THe thing is when we change the width below 1404px, the 'name' moves to next line which can't be displayed. So in order to see that we need to increase the height.
[Screencast from 16-07-24 02:19:30 PM IST.webm](https://github.com/user-attachments/assets/a489a90d-d860-4560-95a4-c13efe8e83d4)

KrishavRajSingh:
@klakhov you may review this.

cvat-ui/src/components/labels-editor/styles.scss
```diff
@@ -249,4 +249,30 @@ textarea.ant-input.cvat-raw-labels-viewer {
...
+    justify-content: center;
+    align-items: center;
+}
+
+.cvat-attribute-inputs-wrapper {
+    height: 4 * $grid-unit-size;
+}
+
+.cvat-save-cancel-btn {
+    padding-top: 0.1rem;
```
klakhov, line 269: The same comment as before, please dont use rem, use $grid-unit-size

cvat-ui/src/components/labels-editor/styles.scss
```diff
@@ -249,4 +249,30 @@ textarea.ant-input.cvat-raw-labels-viewer {
...
+}
+
+.react-grid-layout {
+    overflow-x: clip;
+}
+
+.anticon-holder > svg {
+    height: 3 / 2 * $grid-unit-size;
```
klakhov, line 277:
```suggestion
    height: $grid-unit-size * 1.5;
```

klakhov (commented):
Hi @KrishavRajSingh, 

As I see now you are trying to adjust height of elements based on having/not-having error using some unspecified constants as `8`,`6`, `25` etc. I dont think this approach is correct.  Consider this: I've incorrectly filled up a attibute name with two errors and now adjustments are not working:
image
In other case, what if we will try to add the `LabelForm` component on other page where the width of the container would be different. It will break the adjustments too.

We need to calculate height automatically based on current form component height(with errors) and  base row height. We can use some constants but maybe for ratios, not for the height itself. There is a [thread about this in RGL repo](https://github.com/react-grid-layout/react-grid-layout/issues/971). Can we try to get the actual width of the components and resize its grid height based on it?


Also, the same comment as before, we need to make the drag element padding smaller, its too large now.
image

klakhov:
This PR has no activity for more than a month, I will close it for now. If you wish to finish the contribution feel free to re-open :smile:

PR #7632, opened 2024-03-18
quality_control: factor out duplicate annotation summary creation logic

SpecLad:
In addition, remove code creating an unused annotation summary object from `generate_report`.

### How has this been tested?

Manual testing.

### Checklist

### License

### DIFF
--- a/cvat/apps/quality_control/quality_reports.py
+++ b/cvat/apps/quality_control/quality_reports.py
@@ -1886,37 +1886,9 @@ class DatasetComparator:
             gt_label_idx = gt_ann.label if gt_ann else self._UNMATCHED_IDX
             confusion_matrix[ds_label_idx, gt_label_idx] += 1
 
-        matched_ann_counts = np.diag(confusion_matrix)
-        ds_ann_counts = np.sum(confusion_matrix, axis=1)
-        gt_ann_counts = np.sum(confusion_matrix, axis=0)
-        label_accuracies = _arr_div(
-            matched_ann_counts, ds_ann_counts + gt_ann_counts - matched_ann_counts
-        )
-        label_precisions = _arr_div(matched_ann_counts, ds_ann_counts)
-        label_recalls = _arr_div(matched_ann_counts, gt_ann_counts)
-
-        valid_annotations_count = np.sum(matched_ann_counts)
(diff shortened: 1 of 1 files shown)


cvat/apps/quality_control/quality_reports.py
```diff
@@ -1955,21 +1927,49 @@ def _make_zero_confusion_matrix(self) -> Tuple[List[str], np.ndarray]:
...
+        )
+        label_precisions = _arr_div(matched_ann_counts, ds_ann_counts)
+        label_recalls = _arr_div(matched_ann_counts, gt_ann_counts)
+
+        valid_annotations_count = np.sum(matched_ann_counts)
+        missing_annotations_count = np.sum(confusion_matrix[cls._UNMATCHED_IDX, :])
+        extra_annotations_count = np.sum(confusion_matrix[:, cls._UNMATCHED_IDX])
+        total_annotations_count = np.sum(confusion_matrix)
+        ds_annotations_count = np.sum(ds_ann_counts[: cls._UNMATCHED_IDX])
+        gt_annotations_count = np.sum(gt_ann_counts[: cls._UNMATCHED_IDX])
```
zhiltsov-max, line 1949: Since we're refactoring this, I think, it would be nice to exclude just the specific column/row from the computations. Currently, it implicitly assumes the value to be -1, even though it has a named constant. This can lead to unnoticed bugs, when changed. On option is to sum everything and then subtract the extra, or to try to use some indexing feature from numpy, described in [this section](https://numpy.org/doc/stable/user/basics.indexing.html#basics-indexing).

SpecLad: > This can lead to unnoticed bugs, when changed.

I get the concern, but I have trouble imagining changes that are a) non-gratuitous, and b) would not require you to revisit the rest of the code anyway. Like, I guess you could move the unmatched counts to index 0, but a) why would you do that? and b) this would mean that the confusion matric building code would have to be altered too. And sure, you could make that code more future-proof too (e.g., bring back the `rmap` dict), but that would make the code harder to understand in the present.

IMO, it's fine to pick a layout for the confusion matrix and commit to it. This makes the code simpler and easier to understand.

FWIW, for these sums specifically, it wouldn't be that much harder to sum all and subtract unmatched, but I've worked on adding macro averages, and that trick doesn't work there. So I'd prefer to be able to rely on the layout of the matrix.

PR #7641, opened 2024-03-19
Add macro average precision/recall to quality reports

SpecLad:
### Motivation and context

This provides another way to evaluate annotation quality.

### How has this been tested?

Manual testing.

### Checklist

### License

### DIFF
--- a/cvat/apps/quality_control/quality_reports.py
+++ b/cvat/apps/quality_control/quality_reports.py
@@ -265,6 +265,14 @@ class ComparisonReportAnnotationsSummary(_Serializable):
     def recall(self) -> float:
         return self.valid_count / (self.gt_count or 1)
 
+    @property
+    def macro_average_precision(self) -> float:
+        return np.average(self.confusion_matrix.precision[: DatasetComparator._UNMATCHED_IDX])
+
+    @property
+    def macro_average_recall(self) -> float:
+        return np.average(self.confusion_matrix.recall[: DatasetComparator._UNMATCHED_IDX])
+
     def accumulate(self, other: ComparisonReportAnnotationsSummary):
         for field in [
             "valid_count",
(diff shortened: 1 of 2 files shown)


cvat/apps/quality_control/quality_reports.py
```diff
@@ -265,6 +265,14 @@ def precision(self) -> float:
     def recall(self) -> float:
         return self.valid_count / (self.gt_count or 1)
 
+    @property
+    def macro_average_precision(self) -> float:
+        return np.average(self.confusion_matrix.precision[: DatasetComparator._UNMATCHED_IDX])
```
zhiltsov-max, line 270: Please check that 0/0 is not converted to 0, and it doesn't reduce the average quality. It can happen if there are no annotations of some class.

cvat/apps/quality_control/quality_reports.py
```diff
@@ -265,6 +265,14 @@ def precision(self) -> float:
     def recall(self) -> float:
         return self.valid_count / (self.gt_count or 1)
```
zhiltsov-max, line 267: What about accuracy? Why only precision and recall?

SpecLad: The overall accuracy value is not the average of per-class accuracies. It's going to be smaller than each of them, actually, because it has the same denominator, but the numerator is smaller. (Conceptually, it's harder to be a multiclass classifier than a binary classifier for each class.)

Theoretically, you could average the per-class accuracies and call that a "macro average accuracy", but I'm not sure that's a useful metric. 🤷‍♂️

zhiltsov-max: What do you mean by overall accuracy? Micro accuracy?

> Theoretically, you could average the per-class accuracies and call that a "macro average accuracy", but I'm not sure that's a useful metric. 🤷‍♂️

Yes, that's what I'm suggesting to add.

Issue #6528, opened 2023-07-20
Cannot create a task with  👀 character in the uploaded filename


azhavoro:
### My actions before raising this issue
- [ ] Read/searched [the docs](https://github.com/cvat-ai/cvat/tree/master#documentation)
- [ ] Searched [past issues](/issues)

### Steps to Reproduce (for bugs)

1.
1.
1.
1.

### Your Environment

- Git hash commit (`git log -1`):
- Docker version `docker version` (e.g. Docker 17.0.05):
- Are you using Docker Swarm or Kubernetes?
- Operating System and version (e.g. Linux, Windows, MacOS):
- Code example or link to GitHub repo or gist to reproduce problem:
- Other diagnostic information / logs:
    <details>
    <summary>Logs from `cvat` container</summary>
    </details>

AndrewLuGit:
I would like to be assigned this issue

AndrewLuGit:
The cause of this issue seems to be because of this line in `views.py`: https://github.com/opencv/cvat/blob/f22322f4349a91bbb645dfcacf1e799d870a6f16/cvat/apps/engine/views.py#L984

When there are characters that are not encodable in latin-1 such as  👀 in the file name, in the response header the file name gets encoded using MIME encoding, and so retrieving the file name from the response header can lead to unexpected results. My proposed solution is to simply decode the file name in the `_maybe_append_upload_info_entry` method, like so:
```python
def _maybe_append_upload_info_entry(self, filename: str):
    task_data = cast(Data, self._object.data)

    raw, encoding = decode_header(filename)[0]

    if encoding is not None:
        filename = raw.decode(encoding)

    filename = self._prepare_upload_info_entry(filename)
    task_data.client_files.get_or_create(file=filename)
```

AndrewLuGit:
@nmanovic am I good to create a PR for this?

nmanovic:
@AndrewLuGit , please do.

azhavoro:
Fixed in https://github.com/cvat-ai/cvat/pull/7646

PR #7646, opened 2024-03-20
[GSoC2024] Decode uploaded filename from request header

AndrewLuGit:
### Motivation and context

This fixes issue #6528, as filenames with special characters must be decoded from the MIME header back into UTF-8.

### Checklist

### License


cvat/apps/engine/views.py
```diff
@@ -934,6 +935,11 @@ def _prepare_upload_info_entry(self, filename: str) -> str:
     def _maybe_append_upload_info_entry(self, filename: str):
         task_data = cast(Data, self._object.data)
 
+        raw, encoding = decode_header(filename)[0]
```
azhavoro, line 938: Please add a note here about when it can be a MIME-compatible header string and which code generates this string. Is this part of TUS specification or something else?
Also, please add exception handling here

changelog.d/20240320_174524_andrew.yunxiang.md
```diff
@@ -0,0 +1,4 @@
+### Fixed
+
+- Creating tasks with special characters in uploaded filename
```
azhavoro, line 3: Please make sure that other upload-related functions work with such characters in file names.

azhavoro (changes requested):
Thanks for the contribution!

Please consider to add a API test to check this case.

cvat/apps/engine/views.py
AndrewLuGit: As I was working on exception handling, I realized that there may be the case where the original uploaded file already has a MIME-encoded filename, and in that case we wouldn't want to decode the filename. I think it would be better if we could get the filename directly from the request header instead of from the response header, since the MIME encoding doesn't come from the TUS specification, but rather from when the code creates the response. What do you think?

changelog.d/20240320_174524_andrew.yunxiang.md
AndrewLuGit: Bulk (non-TUS) uploads work fine. I'm not sure how to do remote and cloud uploads. Can I test those through the testing framework?

AndrewLuGit:
@azhavoro I'm not sure how to add the API test case, since this problem only happens through TUS uploads. How do I make sure a file is uploaded through TUS using the testing framework? Does the test case need to be in cypress instead of pytest?

cvat/apps/engine/views.py
azhavoro: I agree that taking the filename from the request rather than the response header is the right direction.

changelog.d/20240320_174524_andrew.yunxiang.md
azhavoro: Yes, it looks like the cypres test is the easiest way to initiate uploading via TUS.

azhavoro:
> @azhavoro I'm not sure how to add the API test case, since this problem only happens through TUS uploads. How do I make sure a file is uploaded through TUS using the testing framework? Does the test case need to be in cypress instead of pytest?

Yes, it looks like the cypres test is the easiest way to initiate uploading via TUS.

azhavoro:
Thanks for the contribution!

azhavoro (approved):

Issue #7571, opened 2024-03-08
"Switch outside" property does not work with the Datumaro format while tracking instances


Krupal09:
### Actions before raising this issue

- [X] I searched the existing issues and did not find anything similar.
- [X] I read/searched [the docs](https://opencv.github.io/cvat/docs/)

### Steps to Reproduce

1. Upload a series of images to the CVAT tool
2. Utilize polygon tracking feature to annotate objects over a set of consecutive images
3. At certain point (for example, at image number 10), opt for "switch outside" property to stop tracking one of the objects
4. Save the annotations in Datumaro format (Menu -> Export job dataset -> Datumaro format)
5. Create a fresh job and upload the same set of images as mentioned in the step 1
6. Upload the annotations generated in step 4
7. Go to the image where the "switch outside" property was opted for (, which was carried out in step 3)
8. The object's annotation polygon still shows up and keeps on penetrating to the further frames

### Expected Behavior

1. Upload a series of images to the CVAT tool
2. Utilize polygon tracking feature to annotate objects over a set of consecutive images
3. At certain point (for example, at image number 10), opt for "switch outside" property to stop tracking one of the objects
4. Save the annotations in Datumaro format (Menu -> Export job dataset -> Datumaro format)
5. Create a fresh job and upload the same set of images as mentioned in the step 1
6. Upload the annotations generated in step 4
7. Go to the image where the "switch outside" property was opted for (, which was carried out in step 3)
---->> 8. The object's annotation polygon should not show up after uploading the saved Datumaro annotation  <<---------

### Context

I am tracking objects over multiple frames and want to remove their annotations from further images when objects are not present on the current frame. The problem with the current state is that all labels from the previous frames are being accumulated to the current frame !

### Environment

Operating system : Linux 22.04
OS type : 64-bit
Using google chrome's latest version
Using CVAT docker on local system

bsekachev:
Please, fill the environment.

Krupal09:
As it is clearly visible in attached image, annotations from previous images are being accumulated in the current images which is not expected. Any help in this regard would be really appreciated.

![Screenshot from 2024-03-08 10-52-57](https://github.com/opencv/cvat/assets/49275269/8d1e7224-0cbb-4a10-a910-f59e3eb1bc62)

Krupal09:
> Please, fill the environment.

@bsekachev Thank you for your quick response ! I have updated the environment field. Kindly let me know if further details are required.

[...]

PR #7669, opened 2024-03-23
[GSoC2024] Fix Proposal for Issue #7571

Yeek020407:
### Motivation and context

Main Issue: #7571 
Related Issue: #2339 

I've reproduced the issue mentioned in #7571 when exporting and importing annotations using both the Datumaro and Coco 1.0 formats. Specifically, the "Switch outside" attribute isn't being applied as expected. After some investigation, I pinpointed the root cause to be the absence of the "outside" attribute in the exported annotations.

To address this, I've made adjustments to the binding.py file to bypass the track_id during annotation import. This modification appears to solve the issue regarding the "Switch outside" attribute. However, it introduces a new concern: the potential loss of information, including keyframes and track_id.

While this workaround offers a temporary fix, I'm contemplating a more holistic approach that entails properly handling the "outside" attribute during both the export and import processes of annotations. This method could preserve the integrity of the annotations while ensuring the functionality of the "Switch outside" attribute.

I'm reaching out for feedback or suggestions on my proposed solution. Is there a preference for one of these approaches, or might there be another avenue I haven't considered? 

Looking forward to your insights.

### Checklist

### License

### DIFF
--- a/cvat/apps/dataset_manager/bindings.py
+++ b/cvat/apps/dataset_manager/bindings.py
@@ -1757,9 +1757,9 @@ class CvatToDmAnnotationConverter:
         if shape.type == ShapeType.RECTANGLE:
             dm_attr['rotation'] = shape.rotation
 
-        if hasattr(shape, 'track_id'):
-            dm_attr['track_id'] = shape.track_id
-            dm_attr['keyframe'] = shape.keyframe
+        # if hasattr(shape, 'track_id'):
+        #     dm_attr['track_id'] = shape.track_id
+        #     dm_attr['keyframe'] = shape.keyframe
 
         dm_points = shape.points
 


zhiltsov-max:
Hi, thank you for sending the PR! Just disabling tracks in the export doesn't seem as a proper solution. Probably, something along [these lines](https://github.com/opencv/cvat/blob/d0e1143c809e0f37dc12e70784570cae1ecb9985/cvat/apps/dataset_manager/bindings.py#L2083-L2140) should be adjusted, as this is the common logic for several formats. The handling logic (and so, the solution) should be similar to [this fragment](https://github.com/opencv/cvat/blob/d0e1143c809e0f37dc12e70784570cae1ecb9985/cvat/apps/dataset_manager/formats/mot.py#L73-L83).

Yeek020407:
> Hi, thank you for sending the PR! Just disabling tracks in the export doesn't seem to be a proper solution, as the whole point of using the outside property is to use it with tracks. Probably, something along [these lines](https://github.com/opencv/cvat/blob/d0e1143c809e0f37dc12e70784570cae1ecb9985/cvat/apps/dataset_manager/bindings.py#L2083-L2140) should be adjusted, as this is the common logic for several formats. The handling logic (and so, the solution) should be similar to [this fragment](https://github.com/opencv/cvat/blob/d0e1143c809e0f37dc12e70784570cae1ecb9985/cvat/apps/dataset_manager/formats/mot.py#L73-L83).

Thanks for your advice. Will make necessary changes soon!

Yeek020407:
Hi @zhiltsov-max , I have made necessary changes and add a test to it. Please have a look

changelog.d/20240403_135916_yeek020407_switch_outside.md
```diff
@@ -0,0 +1,4 @@
+### Fixed
+
+- Check the end of outside attribute for tracked format
```
zhiltsov-max, line 3:
```suggestion
- Formats with the custom `track_id` attribute should import `outside`track shapes properly (e.g. `COCO`, `Datumaro`, `PASCAL VOC`)
```

cvat/apps/dataset_manager/bindings.py
```diff
@@ -2151,6 +2151,26 @@ def reduce_fn(acc, v):
...
+                prev_shape_idx += 1
+                track['shapes'].insert(prev_shape_idx, prev_shape)
+            prev_shape = shape
+            prev_shape_idx += 1
+
+        # if the last shape 'outside' is False, we need to add to stop the tracking
+        if not prev_shape.outside and prev_shape.frame+instance_data.frame_step <= frame_number:
+            prev_shape = prev_shape._replace(outside=True,
+                    frame=prev_shape.frame + instance_data.frame_step)
+            track['shapes'].append(prev_shape)
```
zhiltsov-max, line 2172: Please fix the code formatting in the added fragment.

cvat/apps/dataset_manager/tests/test_rest_api_formats.py
```diff
@@ -1045,6 +1045,47 @@ def test_api_v2_tasks_annotations_dump_and_upload_many_jobs_with_datumaro(self):
                     data_from_task_after_upload = self._get_data_from_task(task_id, include_images)
                     compare_datasets(self, data_from_task_before_upload, data_from_task_after_upload)
 
+    def test_api_v2_tasks_annotations_dump_and_upload_with_datumaro_outside_true(self):
```
zhiltsov-max, line 1048: Please move the test into `/tests/python`. We're trying to move away from the django-based tests, so all the new tests should be added in the "new" tests, with some exceptions for server internals checks.

Yeek020407:
Screenshot 2024-04-07 at 10 11 39 PM

Hii, @zhiltsov-max, I have transferred the code to tests/python. I have checked with all the REST API, SDK and CLI tests. Please take a look.

Yeek020407:
> Hi, please fix the issues reported by the CI.

Hope the changes will fix the issues flagged by the CI.

changelog.d/20240403_135916_yeek020407_switch_outside.md
```diff
@@ -0,0 +1,4 @@
+### Fixed
+
+- Formats with the custom `track_id` attribute should import `outside`track shapes properly (e.g. `COCO`, `Datumaro`, `PASCAL VOC`)
```
zhiltsov-max, line 3:
```suggestion
- Formats with the custom `track_id` attribute should import `outside` track shapes properly (e.g. `COCO`, `COCO Keypoints`, `Datumaro`, `PASCAL VOC`)
```

cvat/apps/dataset_manager/tests/test_rest_api_formats.py
zhiltsov-max, line 1048: I still can see the test in the server tests. Please move it into `tests/python`.

tests/python/rest_api/test_tasks.py
```diff
@@ -2858,3 +2858,97 @@ def test_can_export_and_import_skeleton_tracks_in_coco_format(self):
                 for tes in te.shapes
             ]
         )
+
+    @pytest.mark.parametrize("format_name", ["Datumaro 1.0", "COCO 1.0", "PASCAL VOC 1.1"])
```
zhiltsov-max, line 2862: Probably, it makes sense to add or extend an existing test for the `COCO keypoints` format, as it's also affected.

zhiltsov-max (changes requested):
Please update the tests, and we can merge this.

Yeek020407:
Hi @zhiltsov-max , could you help me with this? I have tried declaring the tracked coco keypoints annotations, but it keeps giving me error on line 2918, which I'm not really sure why.
Screenshot 2024-04-09 at 10 55 30 PM
Here is the error code


And after checking the code for coco keypoints, I realised that there is a need to also append outside true to track['element']['shapes']  (the changes are on line 2182 bindings.py).  Is this something aligned with the requirement?

zhiltsov-max:
Hi, please check server logs with
`docker logs test_cvat_server_1`, `docker logs test_cvat_worker_import_1`, `docker logs test_cvat_worker_export_1`.

> And after checking the code for coco keypoints, I realised that there is a need to also append outside true to track['element']['shapes'] (the changes are on line 2182 bindings.py). Is this something aligned with the requirement?

I'm not sure I got the question. Each skeleton point (stored in the 'element' list) is a separate track, but it can also be affected by the parent skeleton `outside`.

Yeek020407:
I realised if I only change the last parent skeleton `outside` to true, the skeleton points will still appear for all the frames. So, I have also added `outside` to skeleton points.

I am unsure about whether I should follow the parent skeleton `outside` , or append `outside` individually to each skeleton point.

zhiltsov-max:
> I am unsure about whether I should follow the parent skeleton outside , or append outside individually to each skeleton point.

The logic should be as following. If the parent skeleton is outside, its points should also be outside. If the skeleton is visible, each point can be controlled separately. Probably, it's not necessary to copy skeleton's `outside` to the points, if the skeleton is `outside`, as it should be handled by UI and by the server during the task export, though it's better double-checking this.

Yeek020407:
Hi @zhiltsov-max  , I have added the test for coco keypoints with `outside` true. 

However, while running pytest, `test_can_export_and_import_skeleton_tracks_in_coco_format` in test_tasks.py had failed. 

This is track[‘shapes’] under original_annotation:
{"type": "rectangle", "frame": 0, "points": [3.0, 2.0, 2.0, 3.0], “outside”=false},
{"type": "rectangle", "frame": 3, "points": [3.0, 2.0, 2.0, 3.0], “outside”=false},


and this is track['shapes'] under imported_annotation
{"type": "rectangle", "frame": 0, "points": [3.0, 2.0, 2.0, 3.0], “outside”=false},
{"type": "rectangle", "frame": 1, "points": [3.0, 2.0, 2.0, 3.0], “outside”=false},
{"type": "rectangle", "frame": 2, "points": [3.0, 2.0, 2.0, 3.0], “outside”=false},
{"type": "rectangle", "frame": 3, "points": [3.0, 2.0, 2.0, 3.0], “outside”=false},
{"type": "rectangle", "frame": 4, "points": [3.0, 2.0, 2.0, 3.0], “outside”=false},
{"type": "rectangle", "frame": 5, "points": [3.0, 2.0, 2.0, 3.0], “outside”=false},
{"type": "rectangle", "frame": 6, "points": [3.0, 2.0, 2.0, 3.0], “outside”=false},


This is because when we export the annotation, we will add annotations to all the frames if `outside` is not true. Hence the original_annotation that we get by using `task.get_annotations()` does not reflect what we actually exported out.

I am not sure how I should proceed with this. Is changing the annotation used by this test case an appropriate way of dealing with this, or my current way of dealing with the `outside` attribute is wrong?

zhiltsov-max:
I think there should be a check for the `keyframe` annotation attribute, and all annotations in between 2 keyframes should not be imported.

Yeek020407:
Okay, got it. Will be working on this now. 

However, for the case of `test_can_export_and_import_skeleton_tracks_in_coco_format`, I think it does not include `keyframe` attribute in annotation.json. Does that mean I need to modify the annotation?

zhiltsov-max:
> Does that mean I need to modify the annotation?

Yes, I suppose so. Probably, it was supposed to be there. I think, earlier we had logic that all the imported frames from non-track native formats should be keyframes, but we have probably moved away from this idea since then.

Yeek020407:
Hi, @zhiltsov-max  , here is the summary of the changes:

1. Added to_boolean() to all the attributes.get(). 
This is because I realised that in PASCAL VOC format, the keyframe value is ‘False’ (string) instead of False (boolean). Hence, I made this function and added it to all attributes.get() to make sure it will not convert to the wrong boolean.

2. Added ending_tracks to binding.py import_dm_annotations()
It is a temporary holding area for the tracks and elements with `keyframe` and `outside` equal to false after the last `keyframe` = true. In the last part of the function, ending_tracks is used to determine whether or not to append `outside` = true to tracks[‘shapes’] and tracks[‘elements’][‘shapes’]

Please have a look.

Yeek020407:
Screenshot 2024-04-11 at 12 47 31 PM

while testing with coco keypoints format on https://app.cvat.ai/, I found this bug. If I annotate the first frame with skeleton track, put third frame as keyframe and fifth frame with switch outside, It will give  
`ValueError: operands could not be broadcast together with shapes (0,) (4,)`

I am currently solving this issue, but I think this is a different issue from the current one. Should I create a new issue on this?

cvat/apps/dataset_manager/bindings.py
```diff
@@ -2103,15 +2114,17 @@ def reduce_fn(acc, v):
...
+                        if keyframe or outside:
+                            tracks[track_id]['shapes'].append(track)
+                            ending_tracks[track_id]['shapes'] = []
+                        else:
+                            ending_tracks[track_id]['shapes'].append(track)
 
                         if ann.type == dm.AnnotationType.skeleton:
                             for element in ann.elements:
-                                element_keyframe = dm.util.cast(element.attributes.get('keyframe', None), bool, True)
+                                element_keyframe = dm.util.cast(to_boolean(element.attributes.get('keyframe', None)), bool, True)
```
zhiltsov-max, line 2125:
```suggestion
                                element_keyframe = dm.util.cast(element.attributes.get('keyframe', None), to_boolean, True)
```

cvat/apps/dataset_manager/util.py
```diff
@@ -80,3 +80,19 @@ def deepcopy_simple(v):
...
+
+    :param value: The input value to convert.
+    :return: True, False, or None based on the input.
+    """
+    if value is None:
+        return None
+    elif isinstance(value, str):
+        return value.lower() == 'true'
+    else:
+        return bool(value)
```
zhiltsov-max, line 97: Please consider using 
```suggestion
    from attrs.converters import to_bool
    return to_bool(value)
```

Yeek020407: Is replacing all `to_boolean` with `to_bool` a better idea?

cvat/apps/dataset_manager/bindings.py
```diff
@@ -2031,9 +2034,10 @@ def reduce_fn(acc, v):
                     # because in some formats return type can be different
                     # from bool / None
                     # https://github.com/openvinotoolkit/datumaro/issues/719
-                    occluded = dm.util.cast(ann.attributes.pop('occluded', None), bool) is True
-                    keyframe = dm.util.cast(ann.attributes.get('keyframe', None), bool) is True
-                    outside = dm.util.cast(ann.attributes.pop('outside', None), bool) is True
+                    occluded = dm.util.cast(to_boolean(ann.attributes.pop('occluded', None)), bool) is True
+
+                    keyframe = dm.util.cast(to_boolean(ann.attributes.get('keyframe', None)), bool) is True
+                    outside = dm.util.cast(to_boolean(ann.attributes.pop('outside', None)), bool) is True
```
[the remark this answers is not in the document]
Yeek020407: Yea This makes more sense.

cvat/apps/dataset_manager/bindings.py
```diff
@@ -2103,15 +2114,17 @@ def reduce_fn(acc, v):
                             attributes=attributes,
                         )
 
-                        tracks[track_id]['shapes'].append(track)
+                        if keyframe or outside:
+                            tracks[track_id]['shapes'].append(track)
+                            ending_tracks[track_id]['shapes'] = []
+                        else:
+                            ending_tracks[track_id]['shapes'].append(track)
```
zhiltsov-max, line 2121: Probably, it should not be here, as we can't expect the parsed `item`s to be in the CVAT frame order. The order of track shapes is only established in the list in the line [2184](https://github.com/cvat-ai/cvat/blob/bb71350278816b563035b5b9f702ded7ac323177/cvat/apps/dataset_manager/bindings.py#L2184), so this code can possibly be added after that line.

Yeek020407: Instead of separating the shapes into `tracks` and `ending_tracks`, maybe I just dump everything into ending_tracks first, then sort and process it at line [2179](https://github.com/cvat-ai/cvat/blob/bb71350278816b563035b5b9f702ded7ac323177/cvat/apps/dataset_manager/bindings.py#L2179). Is this something you mean?

cvat/apps/dataset_manager/util.py
zhiltsov-max: Sure, this will work too.

cvat/apps/dataset_manager/bindings.py
zhiltsov-max: Basically, yes, but I'm not sure why `ending_tracks` are needed in the first place. You can collect all the track shapes in `tracks`, then iterate over each track and build a new finalized list of shapes. In the loop we, basically, just need to infer the keyframes and add outside frames in the beginning of each track gap.

Please note that all `outside` frames must be keyframes as well. Each beginning of a non-outside shape group should be lead with a keyframe.

```

add outside = true here (where sequences of ":" start)
  v    v     v       v
[----------------------------]        current task or job frames
 |::|||::::||:::|||||::::::::         "|" is a shape, ":" - no shape

 ^  ^      ^    ^ 
add keyframe = true here (where sequences of "|" start)
```

Yeek020407: The reason I used `ending_tracks` is because I think it can help to skip some redundant reiteration of tracks. But I didn't consider the sequence of tracks at that time.

Yeek020407:
Hi @zhiltsov-max , I have made the necessary changes based on your advice. The `finalize_tracks` basically rearranges and remove `tracks['shapes']` with outside=false from `tracks`.

cvat/apps/dataset_manager/bindings.py
```diff
@@ -2137,7 +2138,9 @@ def reduce_fn(acc, v):
                                     z_order=element.z_order,
                                     source=element_source,
                                     attributes=element_attributes,
-                                ))
+                                )
+
+                                tracks[track_id]['elements'][element.label].shapes.append(new_element_shape)
```
zhiltsov-max, line 2143: What was the point of changing these lines?

cvat/apps/dataset_manager/bindings.py
```diff
@@ -2151,11 +2154,67 @@ def reduce_fn(acc, v):
...
+                                    frame=prev_shape.frame + instance_data.frame_step)
+                            current_shapes.append(prev_shape)
+                    prev_shape = shape
+
+                if not prev_shape.outside and prev_shape.frame+instance_data.frame_step <= stop_frame:
+                    prev_shape = prev_shape._replace(outside=True, keyframe=True,
+                            frame=prev_shape.frame + instance_data.frame_step)
+                    current_shapes.append(prev_shape)
+
+    for track in finalize_tracks.values():
```
zhiltsov-max, line 2214:
```suggestion
```

You're already iterating over the same tracks.

cvat/apps/dataset_manager/bindings.py
```diff
@@ -2107,11 +2109,9 @@ def reduce_fn(acc, v):
 
                         if ann.type == dm.AnnotationType.skeleton:
                             for element in ann.elements:
-                                element_keyframe = dm.util.cast(element.attributes.get('keyframe', None), bool, True)
+                                element_keyframe = dm.util.cast(element.attributes.get('keyframe', None), to_bool, True)
```
zhiltsov-max, line 2112: The variable is not used.

cvat/apps/dataset_manager/bindings.py
```diff
@@ -2151,11 +2154,67 @@ def reduce_fn(acc, v):
...
+                            prev_shape = prev_shape._replace(outside=True, keyframe=True,
+                                    frame=prev_shape.frame + instance_data.frame_step)
+                            current_shapes.append(prev_shape)
+                    prev_shape = shape
+
+                if not prev_shape.outside and prev_shape.frame+instance_data.frame_step <= stop_frame:
+                    prev_shape = prev_shape._replace(outside=True, keyframe=True,
+                            frame=prev_shape.frame + instance_data.frame_step)
+                    current_shapes.append(prev_shape)
+
```
zhiltsov-max, line 2213: Probably, we need to skip empty tracks with no shapes and no elements inside.

cvat/apps/dataset_manager/bindings.py
```diff
@@ -2151,11 +2154,67 @@ def reduce_fn(acc, v):
...
+            'label': track['label'],
+            'group': track['group'],
+            'source': track['source'],
+            'shapes': [],
+            'elements':{},
+        }
+        track['shapes'].sort(key=lambda t: t.frame)
+        prev_shape = track['shapes'][0]
+
+        for shape in track['shapes'][1:]:
```
zhiltsov-max, line 2170: Probably, the first shape can be missed, if there is only 1 shape in the track.

cvat/apps/dataset_manager/bindings.py
```diff
@@ -2151,11 +2154,67 @@ def reduce_fn(acc, v):
...
+                finalize_tracks[track_id]['shapes'].append(prev_shape)
+            else:
+                has_skip = instance_data.frame_step < shape.frame - prev_shape.frame
+                if has_skip and not prev_shape.outside:
+                    prev_shape = prev_shape._replace(outside=True, keyframe=True,
+                            frame=prev_shape.frame + instance_data.frame_step)
+                    finalize_tracks[track_id]['shapes'].append(prev_shape)
+            prev_shape = shape
+
+        if not prev_shape.outside and prev_shape.frame+instance_data.frame_step <= stop_frame:
```
zhiltsov-max, line 2181: Probably, `instance_data.frame_step` is not needed here.

cvat/apps/dataset_manager/bindings.py
```diff
@@ -2151,11 +2154,67 @@ def reduce_fn(acc, v):
...
+    finalize_tracks = {}
+    stop_frame = int(instance_data.meta[instance_data.META_FIELD]['stop_frame'])
+    for track_id, track in tracks.items():
+        finalize_tracks[track_id] = {
+            'label': track['label'],
+            'group': track['group'],
+            'source': track['source'],
+            'shapes': [],
+            'elements':{},
+        }
```
zhiltsov-max, line 2166: I'm still don't see why this variable is needed, as you could just do everything with the current `track` in the loop, and maybe using a new `shapes` and `elements` variables.

cvat/apps/dataset_manager/bindings.py
```diff
@@ -2151,11 +2154,67 @@ def reduce_fn(acc, v):
...
+            if prev_shape.keyframe or prev_shape.outside:
+                finalize_tracks[track_id]['shapes'].append(prev_shape)
+            else:
+                has_skip = instance_data.frame_step < shape.frame - prev_shape.frame
+                if has_skip and not prev_shape.outside:
+                    prev_shape = prev_shape._replace(outside=True, keyframe=True,
+                            frame=prev_shape.frame + instance_data.frame_step)
+                    finalize_tracks[track_id]['shapes'].append(prev_shape)
+            prev_shape = shape
+
```
zhiltsov-max, line 2180: I don't see this code adds keyframe = true to the 1st visible shape in each visible section of the track. Probably, it needs to be done somewhere (`if not prev_shape or prev_shape.outside:`).

tests/python/rest_api/test_tasks.py
```diff
@@ -2819,42 +2819,207 @@ def test_import_annotations(self, task_kind, annotation_kind, expect_success):
...
+            if "label_id" not in track:
+                track["label_id"] = labels[0].id
+            for element_idx, element in enumerate(track["elements"]):
+                if "label_id" not in element:
+                    element["label_id"] = sublabels[element_idx].id
+
+        response = patch_method(
+            "admin1", f"tasks/{task_id}/annotations", annotations, action="create"
+        )
+        assert response.status_code == 200, f"Cannot update task's annotations: {response.content}"
```
zhiltsov-max, line 2848: Probably, it can be replaced by `put`.

zhiltsov-max (changes requested):

Yeek020407:
Hi @zhiltsov-max , I have modified the code based on your comments. But there are still left with two that need your clarification.

cvat/apps/dataset_manager/bindings.py
```diff
@@ -2151,11 +2151,82 @@ def reduce_fn(acc, v):
...
+    stop_frame = int(instance_data.meta[instance_data.META_FIELD]['stop_frame'])
+    for track_id, track in tracks.items():
+        # skip tracks with no shapes and elements
+        if len(track['shapes']) == 0 and len(track['elements']) == 0:
+            continue
+
+        track['shapes'].sort(key=lambda t: t.frame)
+        prev_shape = track['shapes'][0]
+        # add keyframe to the first shape
+        prev_shape = prev_shape._replace(keyframe=True)
```
zhiltsov-max, line 2163: Probably, it should be the first non-outside shape, not any first shape.

cvat/apps/dataset_manager/bindings.py
```diff
@@ -2151,11 +2151,89 @@ def reduce_fn(acc, v):
...
+
+            if prev_shape.outside or prev_shape.keyframe:
+                new_shapes.append(prev_shape)
+            elif not prev_shape.outside and prev_shape.frame + instance_data.frame_step <= stop_frame:
+                prev_shape = prev_shape._replace(outside=True, keyframe=True,
+                        frame=prev_shape.frame + instance_data.frame_step)
+                new_shapes.append(prev_shape)
+
+        else:
+            if prev_shape.keyframe or prev_shape.outside:
```
zhiltsov-max, line 2189: Probably, there is no point in adding a single `outside` shape in the track.

Yeek020407:
@zhiltsov-max Could you also take a look at the two reviews that you made previously? I am not very sure about some parts.

cvat/apps/dataset_manager/bindings.py
```diff
@@ -2151,11 +2151,89 @@ def reduce_fn(acc, v):
...
+        prev_shape = track['shapes'][0]
+        # add keyframe to the first non-outside shape
+        if not prev_shape.outside:
+            prev_shape = prev_shape._replace(keyframe=True)
+        new_shapes = []
+
+        # avoid skipping the first frame
+        if len(track['shapes']) > 1:
+            for shape in track['shapes'][1:]:
+                if prev_shape.keyframe or prev_shape.outside:
```
zhiltsov-max, line 2170: Probably, the shape has only be added if it's visible. Actually, I recommend to exclude all the outside shapes to simplify the code. This way, you'll only need to check if the previous was `None` or there was a skip.

zhiltsov-max:
> @zhiltsov-max Could you also take a look at the two reviews that you made previously? I am not very sure about some parts.

> The variable is not used.

The variable was added in this PR and never used. It's clearly visible in the code editor. Probably, you supposed to do something different.


> Probably, `instance_data.frame_step` is not needed here.

This suggestion has probably been my mistake, as it's related to frame skip after the last visible frame.

cvat/apps/dataset_manager/bindings.py
zhiltsov-max: Sorry, incorrect comment.

Yeek020407:
> The variable was added in this PR and never used. It's clearly visible in the code editor. Probably, you supposed to do something different.

I have checked with the export and import of coco keypoints format. However, I realised that when we import the annotations, the annotations are always missing this element_keyframe information, causing it to be set as true

`element_keyframe = dm.util.cast(element.attributes.get('keyframe', None), bool, True);`

cvat/apps/dataset_manager/bindings.py
```diff
@@ -2151,11 +2151,87 @@ def reduce_fn(acc, v):
...
+            if prev_shape.outside or prev_shape.keyframe:
+                new_shapes.append(prev_shape)
+            elif not prev_shape.outside and prev_shape.frame + instance_data.frame_step <= stop_frame:
+                prev_shape = prev_shape._replace(outside=True, keyframe=True,
+                        frame=prev_shape.frame + instance_data.frame_step)
+                new_shapes.append(prev_shape)
+
+        else:
+            if prev_shape.keyframe:
+                new_shapes.append(prev_shape)
```
zhiltsov-max, line 2189:
```suggestion
new_shapes = []
        prev_shape = None
        # infer and keep only the keyframe shapes
        for shape in track['shapes']:
            cur_is_visible = shape and not shape.outside
            if not cur_is_visible:
                continue

            prev_is_visible = prev_shape and not prev_shape.outside
            if not prev_is_visible or (
                has_gap := prev_shape.frame + instance_data.frame_step < shape.frame
            ):
                if has_gap:
                    prev_shape = prev_shape._replace(outside=True, keyframe=True,
                        frame=prev_shape.frame + instance_data.frame_step)
                    new_shapes.append(prev_shape)

                shape = shape._replace(keyframe=True)
                new_shapes.append(shape)
                prev_shape = shape

        if prev_shape and not prev_shape.outside and (
            prev_shape.frame + instance_data.frame_step <= stop_frame
            # has a gap before the current instance segment end
        ):
            prev_shape = prev_shape._replace(outside=True, keyframe=True,
                frame=prev_shape.frame + instance_data.frame_step)
            new_shapes.append(prev_shape)
```

I suggest that we use something like this. Note that this needs to be replicated for track elements as well. Maybe it's worth be extracted as a function to avoid duplicating code.

cvat/apps/dataset_manager/bindings.py
```diff
@@ -2151,11 +2151,87 @@ def reduce_fn(acc, v):
                 raise CvatImportError("Image {}: can't import annotation "
                     "#{} ({}): {}".format(item.id, idx, ann.type.name, e)) from e
 
-    for track in tracks.values():
+    stop_frame = int(instance_data.meta[instance_data.META_FIELD]['stop_frame'])
+    for track_id, track in tracks.items():
+        # skip tracks with no shapes and elements
+        if len(track['shapes']) == 0 and len(track['elements']) == 0:
```
zhiltsov-max, line 2157: It's better to check this before the `instance_data.add_track()` call below.

zhiltsov-max:
> I realised that when we import the annotations, the annotations are always missing this element_keyframe information

It can be so for the correct tracks exported by CVAT in these formats, but this is not guaranteed to be so in arbitrary input data, which can be human-crafted or created by other tools.

> causing it to be always set as true
> element_keyframe = dm.util.cast(element.attributes.get('keyframe', None), bool, True);
> Do you have any idea on this?

Probably, this default of true was correct at some point before, but not now. I suggest that we just use false instead, and then use the keyframe algorithm added in this PR.

Yeek020407:
Hi @zhiltsov-max , I have refactored the code and made some changes to your suggested change. I added the check for  `prev_shape is None` so that the issue of `has_gap` is not defined and will not occur


> Probably, this default of true was correct at some point before, but not now. I suggest that we just use false instead, and then use the keyframe algorithm added in this PR. Actually, the idea of adding this algorithm is exactly because the input data is not expected to be reliable in this regard, and with the algorithm, it's not really important what's the default value.

I have tried the method you suggested, however, while I ran the test, there will be an issue on this test `test_can_export_and_import_skeleton_tracks_in_coco_format`, checking on the number of tracks['elements']['shapes']

This is because when we export and import the annotation again, we lose the keyframe=true information. Hence not all the number of element shapes is imported, causing assertion error.

All the code committed had been tested. Please take a look.

cvat/apps/dataset_manager/bindings.py
```diff
@@ -2187,49 +2186,29 @@ def reduce_fn(acc, v):
...
-
-                else:
-                    if prev_shape.keyframe:
-                        new_element_shapes.append(prev_shape)
-
             track['elements'] = new_elements
 
-        track['elements'] = list(track['elements'].values())
-        instance_data.add_track(instance_data.Track(**track))
+        if not(len(track['shapes']) == 0 and len(track['elements']) == 0):
```
zhiltsov-max, line 2209: `track['shapes'] or track['elements']`

zhiltsov-max:
> I have tried the method you suggested, however, while I ran the test, there will be an issue on this test test_can_export_and_import_skeleton_tracks_in_coco_format, while checking on the number of tracks['elements']['shapes']

> This is because when we export and import the annotation again, we lose the keyframe=true information. Hence not all the number of element shapes is imported, causing assertion error.

Ok, makes sense. Yes, the intermediate keyframes were overlooked, thanks for noticing.

cvat/apps/dataset_manager/bindings.py
```diff
@@ -2151,10 +2151,64 @@ def reduce_fn(acc, v):
...
+                new_element_shapes = append_necessary_outside_attribute(element.shapes)
+                new_elements[element_id] = instance_data.Track(
+                                label=element.label,
+                                group=element.group,
+                                source=element.source,
+                                shapes=new_element_shapes,
+                            )
+            track['elements'] = new_elements
+
+        if not(len(track['shapes']) == 0 or len(track['elements']) == 0):
```
zhiltsov-max, line 2209:
```suggestion
        if track['shapes']) or track['elements']:
```
We need to add only if there's anything to be added.

zhiltsov-max:
Looks good overall, but it would be nice to add a couple of tests for the corner-cases handled by this code, such as only 1 outside frame, outside without keyframe, intermediate keyframe, no keyframes in a track etc.

Yeek020407:
> Looks good overall, but it would be nice to add a couple of tests for the corner-cases handled by this code, such as only 1 outside frame, outside without keyframe, intermediate keyframe, no keyframes in a track etc.

sure will work on this now.

cvat/apps/dataset_manager/bindings.py
```diff
@@ -2151,10 +2151,64 @@ def reduce_fn(acc, v):
                 raise CvatImportError("Image {}: can't import annotation "
                     "#{} ({}): {}".format(item.id, idx, ann.type.name, e)) from e
 
-    for track in tracks.values():
-        track['elements'] = list(track['elements'].values())
-        instance_data.add_track(instance_data.Track(**track))
+    def append_necessary_outside_attribute(shapes):
```
zhiltsov-max, line 2154:
```suggestion
    def _validate_track_shapes(shapes):
```
I suggest that it is renamed to something more general or more close to the actual actions done.

Yeek020407:
Hi, @zhiltsov-max , I have added tests for corner cases for all tracked formats.

And a gentle reminder on one of the previous reviews, I hope to get your advice on that.

cvat/apps/dataset_manager/bindings.py
```diff
@@ -2151,23 +2151,21 @@ def reduce_fn(acc, v):
...
+                prev_shape = shape
                 continue
 
-            prev_is_visible = prev_shape and not prev_shape.outside
-            if not prev_is_visible or (
+            prev_is_visible = prev_shape is not None and not prev_shape.outside
+            if prev_shape is None or shape.keyframe:
+                shape = shape._replace(keyframe=True)
+                new_shapes.append(shape)
+            elif not prev_is_visible or (
```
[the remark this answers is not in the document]
zhiltsov-max: > And a gentle reminder on one of the previous https://github.com/cvat-ai/cvat/pull/7669#discussion_r1565921246, I hope to get your advice on that.

```suggestion
            has_gap = False
            if not prev_is_visible or shape.keyframe (
```

tests/python/rest_api/test_tasks.py
```diff
@@ -2819,42 +2820,523 @@ def test_import_annotations(self, task_kind, annotation_kind, expect_success):
...
-        task.remove_annotations()
+
+        # get the original annotations
+        response = get_method("admin1", f"tasks/{task.id}/annotations")
+        assert response.status_code == 200, f"Cannot get task's annotations: {response.content}"
+        original_annotations = response.json()
+
+        # delete all annotations
+        response = delete_method("admin1", f"tasks/{task_id}/annotations")
+        assert response.status_code == 204, f"Cannot delete task's annotations: {response.content}"
```
zhiltsov-max, line 2856: This is done automatically during the import.

tests/python/rest_api/test_tasks.py
```diff
@@ -2819,42 +2820,523 @@ def test_import_annotations(self, task_kind, annotation_kind, expect_success):
...
-                tes.frame
-                for t in imported_annotations.tracks
-                for te in t.elements
-                for tes in te.shapes
+                tes["frame"]
+                for t in imported_annotations["tracks"]
+                for te in t["elements"]
+                for tes in te["shapes"]
             ]
         )
```
zhiltsov-max, line 2892: Consider using a single DeepDiff call instead, like [here](https://github.com/cvat-ai/cvat/blob/fbc2610d230035a685d2bafb22ee3492884e76b4/tests/python/rest_api/test_tasks.py#L2231-L2241).

tests/python/rest_api/test_tasks.py
```diff
@@ -2819,42 +2820,519 @@ def test_import_annotations(self, task_kind, annotation_kind, expect_success):
...
+                    "elements": [],
+                }
+            ],
+        }
+
+        original_annotations, imported_annotations = self.delete_annotation_and_import_annotations(
+            task_id, annotations, format_name, dataset_file
+        )
+
+        self.compare_original_and_import_annotations(original_annotations, imported_annotations)
```
zhiltsov-max, line 3041: As above, it makes sense to add an extra check that there is the keyframe field set in the imported shape.

tests/python/rest_api/test_tasks.py
```diff
@@ -2819,42 +2820,519 @@ def test_import_annotations(self, task_kind, annotation_kind, expect_success):
...
```
zhiltsov-max, line 3069: As above.

Yeek020407:
Hi @zhiltsov-max , it does make sense to check for `keyframe` for this corner case. However, after checking with the database and GET Method for task annotations [(_init_tracks_from_db)](https://github.com/cvat-ai/cvat/blob/fbc2610d230035a685d2bafb22ee3492884e76b4/cvat/apps/dataset_manager/task.py#L574)  [(get_task_data)](https://github.com/cvat-ai/cvat/blob/fbc2610d230035a685d2bafb22ee3492884e76b4/cvat/apps/dataset_manager/task.py#L865), I found out that there is no keyframe value in the database.

Screenshot 2024-04-16 at 8 00 36 AM

Currently the alternative solution in testing this corner case I can think of is by checking the number of `tracks['shapes']`. From the algorithm used in `bindings.py`, we know that only `keyframe` or `outside` = `True` will be imported, hence, if there exist two tracks['shapes'], it means annotation is imported correctly. If you have any better ideas for testing this corner case, feel free to share them with me and I will implement them.

I have also deleted `test_can_export_and_import_skeleton_tracks_in_coco_format` test as this test is the same as `test_export_and_import_coco_keypoints_with_intermediate_keyframe`

zhiltsov-max:
Ok, yes, only the keyframes are stored in the DB and returned with this request to annotations. I agree that we can check the shapes without looking for the keyframe attribute, as you suggested.

Yeek020407:
> Ok, yes, only the keyframes are stored in the DB and returned with this request to annotations. I agree that we can check the shapes without looking for the keyframe attribute, as you suggested. We still need to check the outside attribute though.

Yeah, thanks for reminding me of this. I have added it into `test_export_and_import_tracked_format_with_outside_true` and `test_export_and_import_coco_keypoints_with_outside_true`.

However, for the one outside case, I think the intended outcome for the annotations would be empty, as there is no visible frame, so `assert len(imported_annotations["tracks"]) == 0` is added instead of checking for `outside` = `True`

zhiltsov-max:
Ok, could you also add a more complex test, maybe similar to what was discussed in https://github.com/cvat-ai/cvat/pull/7669#discussion_r1561234398 ? The currently added tests are working, but they are pretty simple and can miss some problems.

Yeek020407:
@zhiltsov-max , I have added tests to check annotations with two gaps and check if the number of `outside`=`True` ==2. 

While creating this test, I found this bug in `bindings.py`
`if not prev_is_visible or shape.keyframe or (
   has_gap := prev_shape.frame + instance_data.frame_step < shape.frame ) :`
Frame with `outside`=`True` is always `keyframe`=`True`, so whenever we run this line of code, it will exit the checking in the second condition, causing has_gap to always set to False. Hence, I have changed the position between them to solve this bug.

I can't really think of any other more complex test cases at this moment. Do you have any suggestions?

Yeek020407:
@zhiltsov-max  , for the error that occurred in cypress (features/ground_truth_jobs.js), I have run it on my laptop, and it seems like there is no problem over here. Could you give some advice on this issue?
Screenshot 2024-04-16 at 11 12 19 PM

zhiltsov-max:
> Could you give some advice on this issue?

Don't worry, cypress tests often fail for no reason. We typically restart tests in such cases.

Drise13:
Hi all, I think I'm running into this issue. I got here from the description in https://github.com/cvat-ai/cvat/issues/7757. Running in a locally hosted environment. I have several keyframes it looks like with `points: []`:

```
 "tracks": [
            {
                "frame": 0,
                "group": 0,
                "source": "manual",
                "shapes": [
                    {
                        "type": "skeleton",
                        "occluded": false,
                        "outside": false,
                        "z_order": 0,
                        "rotation": 0.0,
                        "points": [
                            11.899662168245413,
                            733.991168169599,
                            263.1463499080511,
                            0.0
                        ],
                        "frame": 0,
                        "attributes": []
                    },
                    {
                        "type": "skeleton",
                        "occluded": false,
                        "outside": false,
                        "z_order": 1,
                        "rotation": 0.0,
                        "points": [],
                        "frame": 67,
                        "attributes": []
                    },
                    {
                        "type": "skeleton",
                        "occluded": false,
                        "outside": false,
                        "z_order": 2,
                        "rotation": 0.0,
                        "points": [],
                        "frame": 239,
                        "attributes": []
                    },
```

etc. I'm showing 40 occurrences total for my task json.

Is there a workaround for this? I have several places where I have a skeleton with only one point visible and I'd rather not delete the entire ID. I was hoping to have this dataset exported and training tonight, but I fear that's not likely at this stage.

Yeek020407:
> Is there a workaround for this? I have several places where I have a skeleton with only one point visible and I'd rather not delete the entire ID. I was hoping to have this dataset exported and training tonight, but I fear that's not likely at this stage.

Hi, I am currently still figuring out where the bug comes from. And for this pull request, it does not solve the error. I don't know if other contributors have solved it; maybe you can search for them.

Drise13:
I'm not super familar with the backend of CVAT. Where can I view the data that CVAT is using in the background to drive all the functionality? It knows these objects exist _somewhere_ since it correctly loads them and displays them. If I have to delve into the data backend to get my dataset information, I'm fine with that.

Yeek020407:
> I'm not super familiar with the backend of CVAT. Where can I view the data that CVAT is using in the background to drive all the functionality? It knows these objects exist _somewhere_ since it correctly loads them and displays them. If I have to delve into the data backend to get my dataset information, I'm fine with that.

Maybe you can take a look at this file [coco.py](https://github.com/cvat-ai/cvat/blob/592c77f43a7a6c6ecba164b80a29294f153c2229/cvat/apps/dataset_manager/formats/coco.py#L40), this is the entry point when you import and export the annotations as coco-keypoints. You can set breakpoints here and check if the annotations are handled correctly. [guide for running debugger](https://docs.cvat.ai/docs/contributing/development-environment/)

Most of the logic in handling annotations is in this file [bindings.py](https://github.com/cvat-ai/cvat/blob/592c77f43a7a6c6ecba164b80a29294f153c2229/cvat/apps/dataset_manager/bindings.py) and this directory [dataset_manager](https://github.com/cvat-ai/cvat/tree/592c77f43a7a6c6ecba164b80a29294f153c2229/cvat/apps/dataset_manager). Can have a look on these.

Drise13:
I found this snippet which I find interesting.

```
        {
            "id": 96,
            "label_id": 26,
            "frame": 7007,
            "group": 0,
            "source": "manual",
            "shapes": [
                {
                    "id": 32148,
                    "type": "skeleton",
                    "frame": 7007,
                    "occluded": false,
                    "outside": false,
                    "z_order": 0,
                    "rotation": 0.0,
                    "points": [
                        562.3782810744306,
                        368.7790275248524,
                        583.5190472403112,
                        336.1720830995109
                    ],
                    "attributes": []
                },
                {
                    "id": 32229,
                    "type": "skeleton",
                    "frame": 7132,
                    "occluded": false,
                    "outside": false,
                    "z_order": 0,
                    "rotation": 0.0,
                    "points": [],
                    "attributes": []
                },
                {
                    "id": 32230,
                    "type": "skeleton",
                    "frame": 7249,
                    "occluded": false,
                    "outside": false,
                    "z_order": 0,
                    "rotation": 0.0,
                    "points": [],
                    "attributes": []
                },
                {
                    "id": 32364,
                    "type": "skeleton",
                    "frame": 7316,
                    "occluded": false,
                    "outside": false,
                    "z_order": 0,
                    "rotation": 0.0,
                    "points": [],
                    "attributes": []
                },
                {
                    "id": 32365,
                    "type": "skeleton",
                    "frame": 7588,
                    "occluded": false,
                    "outside": false,
                    "z_order": 0,
                    "rotation": 0.0,
                    "points": [],
                    "attributes": []
                },
                {
                    "id": 32539,
                    "type": "skeleton",
                    "frame": 8153,
                    "occluded": false,
                    "outside": false,
                    "z_order": 0,
                    "rotation": 0.0,
                    "points": [],
                    "attributes": []
                }
            ],
            "attributes": [],
            "elements": [
                {
                    "id": 97,
                    "label_id": 27,
                    "frame": 7007,
                    "group": 0,
                    "source": "manual",
                    "shapes": [
                        {
                            "id": 32149,
                            "type": "points",
                            "frame": 7007,
                            "occluded": false,
                            "outside": true,
                            "z_order": 0,
                            "rotation": 0.0,
                            "points": [
                                562.3782810744306,
                                368.7790275248524
                            ],
                            "attributes": []
                        },
                        {
                            "id": 32150,
                            "type": "points",
                            "frame": 7011,
                            "occluded": false,
                            "outside": true,
                            "z_order": 0,
                            "rotation": 0.0,
                            "points": [
                                510.58669067443407,
                                330.97962462485157
                            ],
                            "attributes": []
                        },
                        {
                            "id": 32151,
                            "type": "points",
                            "frame": 7014,
                            "occluded": false,
                            "outside": true,
                            "z_order": 0,
                            "rotation": 0.0,
                            "points": [
                                476.12965942443407,
                                305.13978087485157
                            ],
                            "attributes": []
                        },
                        {
                            "id": 32152,
                            "type": "points",
                            "frame": 7017,
                            "occluded": false,
                            "outside": true,
                            "z_order": 0,
                            "rotation": 0.0,
                            "points": [
                                443.77809692443407,
                                283.49915587485884
                            ],
                            "attributes": []
                        },
                        {
                            "id": 32153,
                            "type": "points",
                            "frame": 7021,
                            "occluded": false,
                            "outside": true,
                            "z_order": 0,
                            "rotation": 0.0,
                            "points": [
                                405.67262817443407,
                                254.79993712485884
                            ],
                            "attributes": []
                        },
                        {
                            "id": 32154,
                            "type": "points",
                            "frame": 7024,
                            "occluded": false,
                            "outside": true,
                            "z_order": 0,
                            "rotation": 0.0,
                            "points": [
                                382.11403442443407,
                                235.16321837485884
                            ],
                            "attributes": []
                        },
                        {
                            "id": 32155,
                            "type": "points",
                            "frame": 7027,
                            "occluded": false,
                            "outside": true,
                            "z_order": 0,
                            "rotation": 0.0,
                            "points": [
                                363.87184692443407,
                                220.07728087485884
                            ],
                            "attributes": []
                        },
                        {
                            "id": 32156,
                            "type": "points",
                            "frame": 7032,
                            "occluded": false,
                            "outside": true,
                            "z_order": 0,
                            "rotation": 0.0,
                            "points": [
                                341.83669067443407,
                                198.84290587485884
                            ],
                            "attributes": []
                        },
                        {
                            "id": 32157,
                            "type": "points",
                            "frame": 7035,
                            "occluded": true,
                            "outside": false,
                            "z_order": 0,
                            "rotation": 0.0,
                            "points": [
                                662.3042077872451,
                                365.5625657567143
                            ],
                            "attributes": []
                        },
                        {
                            "id": 32158,
                            "type": "points",
                            "frame": 7039,
                            "occluded": true,
                            "outside": false,
                            "z_order": 0,
                            "rotation": 0.0,
                            "points": [
                                645.8619028999965,
                                349.7876841499965
                            ],
                            "attributes": []
                        },
                        
```

It seems to differentiate between skeleton and points, even though I only have skeleton and rectangle tracks in my annotations.  I'm thinking it tracks the points individually, since it seems to be tracking occlusion and outside correctly there. So what is the purpose of the 'skeleton' section?

Drise13:
The more I think about it, why do skeleton shapes have point data at all? It looks like the points are tracked as elements of the skelton shape.

Drise13:
Ok, so I have a workaround. Borrowing ideas from https://github.com/cvat-ai/cvat/issues/7589 I was able to find a way to get my annotation json file from `localhost:8080/api/jobs/<Job_ID>/annotations?scheme=json`. I opened it in vscode, found all instances of skeletons with empty points and removed them. Then at the bottom of the annotation api page, I submitted the cleaned json file via PUT in the form box. I went back into my job and ran an export.

At first it gave me 

```
  <image id="7448" name="BaseballTrainingData/auto-annotated-bat/frame-08138.png" subset="Train" task_id="11" width="1920" height="1080">
    <skeleton label="bat-kp" source="manual" z_order="0">
    </skeleton>
    <box label="bat" source="manual" occluded="0" xtl="881.56" ytl="761.19" xbr="1055.99" ybr="1001.40" z_order="0">
    </box>
    <skeleton label="bat-kp" source="manual" z_order="0">
    </skeleton>
    <skeleton label="bat-kp" source="manual" z_order="0">
    </skeleton>
    <skeleton label="bat-kp" source="manual" z_order="0">
    </skeleton>
  </image>
```

But, I clicked save and now it is giving me the correct output:

```
  <image id="10602" name="BaseballTrainingData/auto-annotated-bat/frame-11292.png" width="1920" height="1080">
    <box label="bat" source="manual" occluded="0" xtl="1218.20" ytl="1047.70" xbr="1381.48" ybr="1080.00" z_order="0">
    </box>
    <skeleton label="bat-kp" source="manual" z_order="0">
      <points label="bat-knob" source="manual" outside="1" occluded="0" points="1023.90,1064.01">
      </points>
      <points label="bat-barrel-end" source="manual" outside="0" occluded="0" points="1379.36,1064.33">
      </points>
    </skeleton>
  </image>
```

I suspect there is something fishy going on where multiple skeleton shapes are being created. Essentially, any skeleton shape with no points is invalid and should not exist in the DB. Not sure how they got there, but I don't think the formatter is the issue. I have a feeling it is somewhere on the shape creation side, but I don't have the tools set up to find out. Hope that helps.

P.S. I exported as CVAT for Images, but it has the same broadcast error.

zhiltsov-max:
@Drise13, hi. In CVAT, a `skeleton` is a composite annotation, which stores its parts in the `elements` field, so its points indeed are stored as separate shapes, 1 point per shape. The `points` field is not used with this annotation type or at least is ignored at annotation import and export.

> I suspect there is something fishy going on where multiple skeleton shapes are being created. Essentially, any skeleton shape with no points is invalid and should not exist in the DB.

This PR, in turn, should exclude any tracks with no visible shapes on import, so it can partially solve this problem, but it's mainly focused on a different issue.

zhiltsov-max:
@Drise13, you wrote about some code snippet in https://github.com/cvat-ai/cvat/pull/7669#issuecomment-2060291528. Could you provide information about how you could obtain it?

Drise13:
> @Drise13, you wrote about some code snippet in [#7669 (comment)](https://github.com/cvat-ai/cvat/pull/7669#issuecomment-2060291528). Could you provide information about how you could obtain it? Specifically, the `points` field is expected to be empty in `skeleton` annotations.

By snippet, I meant "subset of the json data file". If the skeleton points are supposed to be empty, why are there both ones with and without points? Additionally, when running an export to CVAT, it tries to interpolate the tracks of the skeleton which has no points and throws a broadcast error. My intuition tells me that skeleton tracks should not try to be interpolated and only use the points. That seems to work correctly when I cleaned out the skeletons with no points. It may be that it fixed it because there were no other keyframes to try to interpolate from, but I'm not certain on that.

zhiltsov-max:
@Drise13,

> By snippet, I meant "subset of the json data file".

What is this file?

> Additionally, when running an export to CVAT, it tries to interpolate the tracks of the skeleton which has no points and throws a broadcast error. My intuition tells me that skeleton tracks should not try to be interpolated and only use the points.

Agree, it seems like the source of the problem, according to your findings.

> As for the purpose of this PR, if this is not the right PR for this investigation, should I open a new issue instead?

There is the issue which you mentioned before. If it describes the problem you faced, then it should be discussed there.

cvat/apps/dataset_manager/bindings.py
```diff
@@ -2151,10 +2151,62 @@ def reduce_fn(acc, v):
...
+            ) or shape.keyframe:
+                if has_gap:
+                    prev_shape = prev_shape._replace(outside=True, keyframe=True,
+                        frame=prev_shape.frame + instance_data.frame_step)
+                    new_shapes.append(prev_shape)
 
+                shape = shape._replace(keyframe=True)
+                new_shapes.append(shape)
+
+            prev_shape = shape
```
zhiltsov-max, line 2177: This algorithm misses some required outside shapes, please check the test task

```suggestion
        for shape in shapes:
            prev_is_visible = prev_shape and not prev_shape.outside
            cur_is_visible = shape and not shape.outside
            if not cur_is_visible:
                if prev_is_visible:
                    shape = shape._replace(keyframe=True)
                    new_shapes.append(shape)
                prev_shape = shape
                continue

            has_gap = False
            if prev_is_visible:
                has_gap = prev_shape.frame + instance_data.frame_step < shape.frame

            if not prev_is_visible or has_gap or shape.keyframe:
                if has_gap:
                    prev_shape = prev_shape._replace(outside=True, keyframe=True,
                        frame=prev_shape.frame + instance_data.frame_step)
                    new_shapes.append(prev_shape)

                shape = shape._replace(keyframe=True)
                new_shapes.append(shape)

            prev_shape = shape
```

zhiltsov-max (changes requested):
Please add a more complex test like this 
[task_test_7669_outside_in_dm_formats_backup_2024_04_17_18_55_18.zip](https://github.com/cvat-ai/cvat/files/15015799/task_test_7669_outside_in_dm_formats_backup_2024_04_17_18_55_18.zip)
[job_11805_annotations_2024_04_17_16_43_29_coco keypoints 1.0.zip](https://github.com/cvat-ai/cvat/files/15015800/job_11805_annotations_2024_04_17_16_43_29_coco.keypoints.1.0.zip)

Yeek020407:
@zhiltsov-max Thanks for the suggestion. I have added the new test for this.

cvat/apps/dataset_manager/bindings.py
```diff
@@ -2151,10 +2151,66 @@ def reduce_fn(acc, v):
...
+    def _validate_track_shapes(shapes):
+        new_shapes = []
+        prev_shape = None
+        # infer the keyframe shapes and keep only them
+        for shape in shapes:
+            prev_is_visible = prev_shape and not prev_shape.outside
+            cur_is_visible = shape and not shape.outside
+            if not cur_is_visible:
+                if prev_is_visible:
+                    shape = shape._replace(keyframe=True)
```
zhiltsov-max, line 2163: Actually, it seems like we need to check the distance after the prev shape here as well, like it's done below. Because it's possible that the current invisible shape has a gap after the last visible shape, and we need to insert an outside keyframe after the last visible shape in this case.

cvat/apps/dataset_manager/bindings.py
```diff
@@ -2151,10 +2151,66 @@ def reduce_fn(acc, v):
...
+            if not prev_is_visible or has_gap or shape.keyframe:
+                if has_gap:
+                    prev_shape = prev_shape._replace(outside=True, keyframe=True,
+                        frame=prev_shape.frame + instance_data.frame_step)
+                    new_shapes.append(prev_shape)
 
+                shape = shape._replace(keyframe=True)
+                new_shapes.append(shape)
+
+            prev_shape = shape
```
zhiltsov-max, line 2181:
```suggestion
            prev_is_visible = prev_shape and not prev_shape.outside
            cur_is_visible = shape and not shape.outside

            has_gap = False
            if prev_is_visible:
                has_gap = prev_shape.frame + instance_data.frame_step < shape.frame

            if has_gap:
                prev_shape = prev_shape._replace(outside=True, keyframe=True,
                    frame=prev_shape.frame + instance_data.frame_step)
                new_shapes.append(prev_shape)

            if prev_is_visible != cur_is_visible or cur_is_visible and (has_gap or shape.keyframe):
                shape = shape._replace(keyframe=True)
                new_shapes.append(shape)

            prev_shape = shape
```

Seems like this should fix the problem

Yeek020407: Yeah, this makes the logic clearer and allows for handling all the cases.

I am currently thinking of how to test this [case](https://github.com/cvat-ai/cvat/pull/7669#discussion_r1570266503). Using our current test to test the case of two consecutive outside attributes might be hard. This is because when we export the annotation, we will ignore the second `outside` = `True`.


"|" - shape
"*" - outside true
":" - no shape

before export
[ | | | * : : * : : | | | | ]

After export it out:
[ | | | : : : : : : | | | | ] (it will skip outside=True [here](https://github.com/cvat-ai/cvat/blob/a5269b056341db726da01c2d779d6e3dbb20c8bc/cvat/apps/dataset_manager/bindings.py#L436)

Import the annotation:
[ | | | * : : : : : | | | | ]

So, the imported annotation will only contain one outside instead of two.
This case is more likely to handle human-crafted or annotations created by other tools.

Do you have any idea about testing this case?

zhiltsov-max: Actually, the example above wasn't manually edited, it's just a CVAT export in COCO keypoints for the skeleton track described (3 points, 6 frames, 1st frame visible, 2nd and 3rd fully outside, then points appear by one consequently). So it goes like this: `[ | * : | | |  ]`

Yeek020407: Sorry for the confusion made. I did not notice the example when writing the message. Will take a look now.

Yeek020407:
Hi @zhiltsov-max , I have added the example you provided into our test and added some checking on the number of `outside`  = `True` in each elements.

cvat/apps/dataset_manager/bindings.py
```diff
@@ -2151,10 +2151,60 @@ def reduce_fn(acc, v):
                 raise CvatImportError("Image {}: can't import annotation "
                     "#{} ({}): {}".format(item.id, idx, ann.type.name, e)) from e
 
-    for track in tracks.values():
-        track['elements'] = list(track['elements'].values())
-        instance_data.add_track(instance_data.Track(**track))
+    def _validate_track_shapes(shapes):
```
zhiltsov-max, line 2154:
```suggestion
    def _validate_track_shapes(shapes):
        shapes = sorted(shapes, key=lambda t: t.frame)
        
```

cvat/apps/dataset_manager/bindings.py
```diff
@@ -2151,10 +2151,60 @@ def reduce_fn(acc, v):
...
+        ):
+            prev_shape = prev_shape._replace(outside=True, keyframe=True,
+                frame=prev_shape.frame + instance_data.frame_step)
+            new_shapes.append(prev_shape)
+
+        return new_shapes
+
+    stop_frame = int(instance_data.meta[instance_data.META_FIELD]['stop_frame'])
+    for track_id, track in tracks.items():
+        track['shapes'].sort(key=lambda t: t.frame)
```
zhiltsov-max, line 2189:
```suggestion
```

cvat/apps/dataset_manager/bindings.py
```diff
@@ -2151,10 +2151,60 @@ def reduce_fn(acc, v):
...
+        track['shapes'] = _validate_track_shapes(track['shapes'])
+
+        if ann.type == dm.AnnotationType.skeleton:
+            new_elements = {}
+            for element_id, element in track['elements'].items():
+                element.shapes.sort(key=lambda t: t.frame)
```
zhiltsov-max, line 2195:
```suggestion
```

cvat/apps/dataset_manager/bindings.py
```diff
@@ -2151,10 +2151,60 @@ def reduce_fn(acc, v):
...
+                new_element_shapes = _validate_track_shapes(element.shapes)
+                new_elements[element_id] = instance_data.Track(
+                                label=element.label,
+                                group=element.group,
+                                source=element.source,
+                                shapes=new_element_shapes,
+                            )
```
zhiltsov-max, line 2202:
```suggestion
                new_elements[element_id] = element._replace(shapes=new_element_shapes)
```

cvat/apps/dataset_manager/bindings.py
```diff
@@ -2151,10 +2151,55 @@ def reduce_fn(acc, v):
                 raise CvatImportError("Image {}: can't import annotation "
                     "#{} ({}): {}".format(item.id, idx, ann.type.name, e)) from e
 
-    for track in tracks.values():
-        track['elements'] = list(track['elements'].values())
-        instance_data.add_track(instance_data.Track(**track))
-
+    def _validate_track_shapes(shapes):
+        shapes = sorted(shapes, key=lambda t: t.frame)
+
```
zhiltsov-max, line 2156:
```suggestion

```

zhiltsov-max (approved):

NicolasPetermann134:
Hi @zhiltsov-max !
I'm facing the same issue: ValueError: operands could not be broadcast together with shapes (0,) (6,)

I found two annotations in backup where "points":[] for a skeleton track. 

I removed them and tried to create from backup but I am facing this error - KeyError: "There is no item named 'task.json' in the archive"

I am 100% that in my zip file there is a task.json tough. 
my zip has following structure: 
dataset 
 - annotations.json
 - task.json
 - data
    - all images

Would highly appreciate some help. 
Regards

zhiltsov-max:
@NicolasPetermann134,

I don't see how it's related to this PR or issues focused.

> my zip has following structure:

What is `dataset`, a directory? If it's a directory, it should not be present in the backup archive.

NicolasPetermann134:
@zhiltsov-max 
I commented on this case as I saw here https://github.com/cvat-ai/cvat/pull/7669#issuecomment-2060319103 the idea of deleting the instances with empty points...

dataset is the zipped file. 
I have here shared my backups https://1drv.ms/f/s!AiG5XaoBkxRVm9YYwD1G7yHEUrE01g?e=tPhDNN

One is the original backup which is not modified - this one I can use to create a new task without any issues. 
For the other one I only modified annotation.json - I deleted there the instances with empty points, I didn't touch the task.json but I get the error KeyError: "There is no item named 'task.json' in the archive"

What could be the issue?

zhiltsov-max:
@Eldies , please look at the issue.

NicolasPetermann134:
hi @Eldies, any news on this?

NicolasPetermann134:
ftr: I could solve the problem. I could export the dataset easily without errors when I tried it on the self hosted version.

PR #7681, opened 2024-03-26
Fix two problems with the task segmentation algorithm

### LINKED ISSUE #7675: Redundant jobs created in task when segment_size is explicitly set

### Steps to Reproduce

Create a task with the following parameters:

* Remote source: https://archive.org/download/SampleVideo1280x7205mb/SampleVideo_1280x720_5mb.mp4
* Segment size: 739
* Overlap: 5

This creates two jobs: one with frames 0-738, another with frames 734-738. The second job is completely redundant, since all frames in it are also in the first job.

### Expected Behavior

Only the first job should be created.

### Possible Solution

_No response_

### Context

This happens if you already have a video task that was created without explicitly specifying a segment size, and you try to create a new task programmatically by copying parameters from the old task.

### Environment

```Markdown
Reproduced with commit eb5b6126.
```

SpecLad:
### Motivation and context

1. When a task has non-zero overlap and exactly as many frames as needed to create 1 or more complete segments, the current algorithm generates a redundant segment at the end. For example, if size is 5, segment size is 3, and overlap is 1, it generates segments (0, 2), (2, 4), and (4, 4).

   The algorithm attempts to compensate for this, but it only works in the case where the segment size is unspecified (and defaults to the total size).

   Update the algorithm to handle this correctly in the general case.

2. The algorithm selects a default overlap size of 5 if the media file is a video. However, this might not be a valid value if the task has a very small segment size. In this case, a range of undesirable behaviors may occur, depending on the segment size:

   * segments getting generated such that more than 2 segments cover a single frame;

   * task creation crashing with an exception;

   * a task being created with no segments at all.

   Fix this by clamping the default overlap size the same way as a user-specified one.

Fixes #7675.

### How has this been tested?

Unit tests.

### Checklist

### License

### DIFF
--- a/changelog.d/20240326_133323_roman_segmentation_fix.md
+++ b/changelog.d/20240326_133323_roman_segmentation_fix.md
@@ -1,0 +1,9 @@
+### Fixed
+
+- Fixed an invalid default overlap size being selected for video tasks
+  with small segments
+  (<https://github.com/opencv/cvat/pull/7681>)
+
+- Fixed redundant jobs being created for tasks with non-zero overlap
+  in certain cases
+  (<https://github.com/opencv/cvat/pull/7681>)
--- a/cvat/apps/engine/task.py
+++ b/cvat/apps/engine/task.py
@@ -6,7 +6,6 @@
 import itertools
 import fnmatch
 import os
-import sys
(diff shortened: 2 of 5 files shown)


nmanovic (approved):
LGTM

Issue #7598, opened 2024-03-12
Linting warnings in some typescript files


tahamukhtar20:
### Actions before raising this issue

- [X] I searched the existing issues and did not find anything similar.
- [X] I read/searched [the docs](https://opencv.github.io/cvat/docs/)

### Steps to Reproduce

run lint and you will face some warnings
similarly when you push some changes you will observe that there are some warnings visible=>
![image](https://github.com/opencv/cvat/assets/91777330/4b3ed5c4-6879-45bc-8fdd-1cd8b150a37c)

### Expected Behavior

No linting warnings

### Possible Solution

adding the required return types to fix this issue

### Environment

```Markdown
- Ubuntu 23.04
- Chrome Browser
```

tahamukhtar20:
@bsekachev What do you think about this one, can I work on this?

bsekachev:
Hello,
Are you going to fix all typescript warnings in the product? 
Or just enable linting? The second option will not really work well for people who deliver new features because they will also need to fix existing errors and warnings.

tahamukhtar20:
No @bsekachev, I think linting is already enabled, there are some typescript warnings in some files, I noticed these initially when I was working on my issue which was previously assigned to me, so the first option, fixing all typescript errors in the product.

bsekachev:
Under linting I actually meant adding one more stage on git pre-commit (now it runs eslint on changed files, for example) and adding a new action on GitHub.

tahamukhtar20:
Yes, so there are some typescript warnings right now, for example, there are some types which are not added in the code, the code runs fine without those types but, it throws some warnings, =>
![image](https://github.com/opencv/cvat/assets/91777330/3ec677ec-7d3b-46ae-b5ec-d816afa316ae)
in this image we just need to add a return type which would be void in this case to remove the warning. 
I'm looking to fix such errors and not add any new action on GitHub as that would cause problems for other contributors and their workflows.
So, @bsekachev should I work on this? Please let me know if this needs some more explanation.

bsekachev:
Hello, 
You are welcome to contribute to fix these warnings.

PR #7689, opened 2024-03-27
[GSoC2024] add return types to fix typescript warnings

tahamukhtar20:
Fixes: #7598
### Motivation and context
There are some minor typescript errors in the files, this pull request fixes those.

### How has this been tested?
Ran the code after changing, and it was functional like before, this didn't need any additional test.

### Checklist

### License


tahamukhtar20:
@bsekachev I've made the necessary changes, please change the last file Lambda Manager as I have added a few Promise<any> there because it is a very reused file and to me, Promise<any> makes the most sense there. Please check and let me know if there is any issue and I'll fix it ASAP. Thanks.

tahamukhtar20:
I think this PR is not ready yet, there are a few changes which I just realized need to be added, I'll add those and ping you accordingly, thanks.

cvat-core/src/lambda-manager.ts
```diff
@@ -43,7 +43,7 @@ class LambdaManager {
         return { models, count: lambdaFunctions.length };
     }
 
-    async run(taskID: number, model: MLModel, args: any) {
```
bsekachev, line 46: Promise<any> should be specified more concrete in this file

changelog.d/20240327_224004_tahamukhtar20.md
```diff
@@ -0,0 +1,4 @@
+### Fixed
+
+- Fixed typescript linting errors in ts files
```
SpecLad, line 3: This fix has no impact on the user, so I don't think it needs a changelog entry.

cvat-core/src/lambda-manager.ts
bsekachev: @tahamukhtar20

tahamukhtar20: I'll get to these tomorrow after submitting my proposal. Thanks for the reminder.

bsekachev:
Hello, do you have any plans regarding the pull request?

tahamukhtar20:
Yes @bsekachev , I left a comment above under your review, please let me know. Thanks.

bsekachev:
Yes, one week ago you mentioned "tomorrow". That is why I am asking :)

tahamukhtar20:
> Yes, one week ago you mentioned "tomorrow". That is why I am asking :)

Sorry about that, @bsekachev  I wanted to ask something about your comment under the Promise<any> review, you said we should use something more concrete here instead of any, but that is Lambda manager, and we are very uncertain what will be returned here, hence doesn't `any` make sense here?

bsekachev:
I believe we are certain enough. Otherwise how would it be possible to work with this API? 
As far as I remember `run()` returns a string, representing rq identificator to fetch statuses later. 
`call()` is more complex, but, more or less, you may find it out from this file `cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/tools-control.tsx`. 

Perhaps the output is different for trackers, detectors and interactors. 
So, it should be like `Promise<TrackerResults | DetectorResults | InteractorResults>`, and define these three types additionally.

tahamukhtar20:
Hi @bsekachev, I've added the required types and currently added the interfaces on top of the Lambda Manager file, please let me know if that is fine or if there is some other place more appropriate for the interfaces.
Additionally, I have updated the concerned code related to the lambda.call function, but I had to make some type assertions(e.g. number[] and number[][] in the tools-control.tsx file) to make it work with the existing code without making too many changes. Please let me know if this is the correct implementation for this.
Thank you.

cvat-core/src/lambda-manager.ts
```diff
@@ -6,14 +6,36 @@
...
 export interface ModelProvider {
     name: string;
     icon: string;
     attributes: Record<string, string>;
 }
 
+export interface InteractorResults {
+    mask: number[][];
+    points: number[][] | number[];
+    bounds: [number, number, number, number]
```
bsekachev, line 20:
```suggestion
    points?: [number, number][];
    bounds?: [number, number, number, number]
```

cvat-core/src/lambda-manager.ts
```diff
@@ -6,14 +6,36 @@
...
+    label: string;
+    outside?: boolean;
+    points?: number[];
+    mask?: number[];
+    elements: SerializedShape[];
+}
+
+export interface TrackerResults {
+    states: any[]
+    shapes: number[][]
```
bsekachev, line 36:
```suggestion
    states: any[];
    shapes: number[][];
```

cvat-core/src/lambda-manager.ts
```diff
@@ -6,14 +6,36 @@
...
+export interface SerializedShape {
+    type: ShapeType;
+    rotation?: number;
+    attributes: { name: string; value: string }[];
+    label: string;
+    outside?: boolean;
+    points?: number[];
+    mask?: number[];
+    elements: SerializedShape[];
+}
```
bsekachev, line 32:
```suggestion
export interface DetectedShape {
    type: ShapeType;
    rotation?: number;
    attributes: { name: string; value: string }[];
    label: string;
    outside?: boolean;
    points?: number[];
    mask?: number[];
    elements: DetectedShape[];
}
```

cvat-core/src/lambda-manager.ts
```diff
@@ -68,7 +90,7 @@ class LambdaManager {
         return result.id;
     }
 
-    async call(taskID, model, args) {
+    async call(taskID, model, args): Promise<TrackerResults | InteractorResults | SerializedShape[]> {
```
bsekachev, line 93:
```suggestion
    async call(taskID, model, args): Promise<TrackerResults | InteractorResults | DetectedShape[]> {
```

bsekachev:
Please, avoid changes in `tools-control.tsx`, I will work with types inconsistency on my own, I see there are some issues.

tahamukhtar20:
@bsekachev Updated

bsekachev (approved):

bsekachev:
Just FYI: https://github.com/cvat-ai/cvat/pull/7785

tahamukhtar20:
I found some more typescript warnings because of missing return types, maybe I can create a follow-up pull request for this if you allow it.

PR #7692, opened 2024-03-28
Split off documentation building into its own pipeline

SpecLad:
### Motivation and context

It has occurred to me that the documentation build happens inside the main workflow, which is configured to ignore changes to the docs when determining whether to run. So if a PR only changes the docs... we don't build the docs. That makes no sense.

Fix this by extracting the documentation job into its own workflow, that always runs. The downside of this is that we have to duplicate the work of generating the SDK, but IMO, it's an acceptable sacrifice.

A side effect of this is that a failure to build the docs no longer prevents the publishing of Docker images, but I don't think that was needed in the first place.

### Checklist

### License

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,66 @@
+name: Docs
+on:
+  push:
+    branches:
+      - 'master'
+      - 'develop'
+  pull_request:
+    types: [ready_for_review, opened, synchronize, reopened]
+
+concurrency:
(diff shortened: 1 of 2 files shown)


zhiltsov-max (approved):
Should we add this into the nightly builds as well?

zhiltsov-max:
I think it's the same, as other nightly checks - just a way to ensure everything still builds without problems.

SpecLad:
(I forgot to add concurrency settings in the initial version; added now.)

Issue #7363, opened 2024-01-17
Cannot read properties of null (reading 'annotations')


bsekachev:
### Actions before raising this issue

- [X] I searched the existing issues and did not find anything similar.
- [X] I read/searched [the docs](https://github.com/cvat-ai/cvat/tree/master#documentation)

### Steps to Reproduce

1. Open a job
2. Upload annotations to the job
3. Go to task page before upload is done
![image](https://github.com/opencv/cvat/assets/40690378/1499849e-e555-492b-a924-398df2779deb)

### Environment

```Markdown
v2.10.0
```

MKBenjie:
I'm interested in this issue and am ready to tackle it. Could you please assign it to me?

MKBenjie:
I was able to reproduce this bug.

## Issue

When I navigate away from the page while uploading annotations the message "Cannot read properties of null (reading 'annotations')" along with "Annotations have been loaded to the task" are displayed to the user. This error occurs because the FETCH_ANNOTATIONS_FAILED action is triggered since annotations have to be loaded and fetched yet I  navigated away from the page, even though the annotations loading process is still a success. This leads to unnecessary error message being displayed.

## Root Cause

The root cause of the issue is that the FETCH_ANNOTATIONS_FAILED action is dispatched without considering the possibility of navigation interruption during the annotations fetching process.

## Proposed Solution

To resolve this issue, I will implement a mechanism to detect navigation interruptions during the annotations fetching process. This involves modifying the fetchAnnotationsAsync function to handle navigation interruptions by checking if the interruption occurred before dispatching the FETCH_ANNOTATIONS_FAILED action. Additionally, I will update the error handling logic to consider navigation interruptions so that the warming isn't displayed if it was due to navigation interruptions.

nmanovic:
@MKBenjie , please send us a PR which solves the issue. Could you please add a test as well?

MKBenjie:
Created the PR and added a test as well

nmanovic:
@MKBenjie , could you please share a link on the PR? I cannot find it.

bsekachev:
@nmanovic 

https://github.com/opencv/cvat/pull/7704

PR #7704, opened 2024-03-30
[GSoC2024] Cannot read properties of null (reading 'annotations') #7363

MKBenjie:
### Checklist

### License


cvat-core/tests/api/annotations.cjs
```diff
@@ -87,6 +87,22 @@ describe('Feature: get annotations', () => {
...
+        const mockAnnotationsGet = jest.spyOn(job.annotations, 'get');
+
+        // Reject the promise with an error to simulate a page interrupt
+        mockAnnotationsGet.mockRejectedValueOnce(new Error("Cannot read properties of null (reading 'annotations')"));
+
+        // Expectation: Retrieving annotations should throw an error
+        await expect(job.annotations.get(0)).rejects.toThrow("Cannot read properties of null (reading 'annotations')");
+
+        // Ensuring that job.annotations.get() was called with the correct arguments
+        expect(mockAnnotationsGet).toHaveBeenCalledWith(0);
```
bsekachev, line 105: 1. The code does not make sense, because the test tests itself
2. Perhaps we need Cypress E2E tests to cover the bugfix, not jest. Because issue on `cvat-ui` level. On `cvat-core` level we can't even reproduce the problem.

cvat-ui/src/actions/annotation-actions.ts
```diff
@@ -295,28 +295,59 @@ async function fetchAnnotations(frameNumber?: number): Promise<{
     };
 }
 
+function navigationInterruptionDetected(callback: (interrupted: boolean) => void): void {
+    let currentPath = window.location.pathname;
+
+    // Listen for popstate events (back/forward navigation)
+    window.addEventListener('popstate', () => {
```
bsekachev, line 302: ``popstate`` event listener is never removed. Memory leak, because `fetchAnnotationsAsync` may be called plenty of times.

cvat-ui/src/actions/annotation-actions.ts
```diff
@@ -295,28 +295,59 @@ async function fetchAnnotations(frameNumber?: number): Promise<{
...
+        const newPath = window.location.pathname;
+        if (newPath !== currentPath) {
+            console.log('Navigation interruption detected.');
```
bsekachev, line 305: we do not need logging in production code

cvat-ui/src/actions/annotation-actions.ts
```diff
@@ -295,28 +295,59 @@ async function fetchAnnotations(frameNumber?: number): Promise<{
...
-            dispatch({
-                type: AnnotationActionTypes.FETCH_ANNOTATIONS_SUCCESS,
-                payload: {
-                    states,
-                    history,
-                    minZ,
-                    maxZ,
-                },
+            // Check if navigation interruption occurred
+            navigationInterruptionDetected((interrupted: boolean) => {
```
bsekachev, line 321: From the code I see:

1. We fetch annotations in line 318
2. We call `navigationInterruptionDetected` with callback passed
2.1. Callback will be called only if `postate` event was dispatched with `true` argument.
2.2. Callback never called with `false` argument. 

Condition in line 322 is always false, so, I am not sure how this code is supposed to work. Did you test it?

cvat-core/tests/api/annotations.cjs
MKBenjie: The test was aiming to simulate a scenario where fetching annotations encounters an error. It was meant for verifying that the method correctly throws an error with the expected message when encountering a page interrupt.

If the issue is more related to UI behavior, then let me focus on E2E tests with Cypress.

cvat-ui/src/actions/annotation-actions.ts
MKBenjie: Ok, I should remove the event listener after it's triggered

cvat-ui/src/actions/annotation-actions.ts
MKBenjie: Let me remove all

cvat-ui/src/actions/annotation-actions.ts
MKBenjie: Ok, I also see errors in my logic of handling this now. What do you think of handling this issue in such away.

1. I separate the handling of navigation interruptions from the handling of fetching errors.
2. I use a flag to track whether a navigation interruption occurred during the fetching process.
3. Then check if navigation interruption occurred , dispatch success action if no interruption AND dispatch FETCH_ANNOTATIONS_FAILED only if not due to navigation interruption
4. Then navigationInterruptionDetected function now returns a function that removes the event listener when it's no longer needed. So that that the event listener is cleaned up to prevent memory leaks.

bsekachev:
Thank you for trying to contribute.
Unfortunately we are not ready to accept the patch as is, sorry.

PR #7714, opened 2024-04-02
Fixed annotation menu for GT objects in GT job

bsekachev:
### Checklist

### License

### DIFF
--- a/cvat-ui/src/actions/annotation-actions.ts
+++ b/cvat-ui/src/actions/annotation-actions.ts
@@ -13,7 +13,7 @@ import {
     RectDrawingMethod, CuboidDrawingMethod, Canvas, CanvasMode as Canvas2DMode,
 } from 'cvat-canvas-wrapper';
 import {
-    getCore, MLModel, JobType, Job, QualityConflict,
+    getCore, MLModel, JobType, Job, QualityConflict, ObjectState,
 } from 'cvat-core-wrapper';
 import logger, { EventScope } from 'cvat-logger';
 import { getCVATStore } from 'cvat-store';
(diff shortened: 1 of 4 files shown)


cvat-ui/src/actions/annotation-actions.ts
```diff
@@ -277,10 +277,22 @@ async function fetchAnnotations(frameNumber?: number): Promise<{
...
+            get(_state, prop) {
+                if (prop === 'isGroundTruth') {
+                    // ground truth objects are not considered as gt objects, relatively to a gt jobs
+                    // to avoid extra css styles, or restrictions applied
+                    return false;
+                }
+
+                return Reflect.get(_state, prop);
+            },
+        }));
```
klakhov, line 294: Generally Im ok with fix.

What do you think about adding a module to `cvat-ui` containing proxy handlers, maybe somewhere in utils? Usually such handlers are independent objects. It can improve code readability by reducing functions size as we have several such handlers.

bsekachev: Is that a common practice or ...? 
Do you have any references?

klakhov: Not sure if thats a common practive or not, but Ive seen a couple of examples where handler object is created somehere outside of the `new Proxy...` in [simple mdn example](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy#description) and in [vuejs source code](https://github.com/vuejs/core/blob/0dea7f9a260d93eb6c39aabac8c94c2c9b2042dd/packages/reactivity/src/ref.ts#L287). So I thought maybe we can use that approach to reduce function size. What do you think?

bsekachev: I can't see real benefits from this approach.
One more thing is that we use proxy in cvat-core, cvat-canvas, and cvat-ui. So, it turns that we need 3 modules just to store Proxy handlers

klakhov: Ok, lets leave it as it is.

klakhov (approved):

Issue #3430, opened 2021-07-18
Automatically create superuser


tadejsv:
I would suggest that the CVAT container automatically creates a superuser on startup, with username/password/email set by the `DJANGO_SUPERUSER_*` [env variables](https://docs.djangoproject.com/en/3.2/ref/django-admin/#createsuperuser).

Right now, users have to manually create the superuser account using `docker exec` after the container starts. But since they will always need a superuser, wouldn't it make more sense to just create one from the entrypoint script? This would also make automatically setting up CVAT much easier, as it eliminates all manual actions.

If this is something that you think would be a good addition to CVAT, I can take a shot at creating a PR.

bsekachev:
@tadejsv 

IMHO If we have default settings for these variables, many users will not change them, it is less secure. 

If we obligate a user to specify these variables before the build, UX is not really well, because the build will be failed without them. What is more, what is the difference if in both cases a user must do some additional steps?

tadejsv:
Regarding defaults: we can leave the default as it is now: so these new env variables unset, and no superuser created. So the user has an option to create a superuser automatically, but it is not an obligation.

As for the difference compared to now: let's take the use care where you want to set up all the infrastructure automatically, for example using Terraform (which I am using now), or some other IaaC product. Currently, after the setup is complete, you will need to somehow log into the instance that is running the container, and manually execute the shell command - it's an "annoying" step, and also forces you to open up your instance for SSH connections. You also need to manually note somewhere what the username/password of superuser will be (or where are they stored), because it is not evident from configuration.

If this is automated, I can just specify these environmental variables in my configuration (which also allows me to do it in a more organized way - for example to store/retrieve them using AWS Secrets Manager), and when the infrastructure is created, everything works out of the box.

tadejsv:
@bsekachev any update on this? If you still don't agree with the proposal feel free to close the issue

nmanovic:
@tadejsv , it is fine to have an alternative way to create super users. Could you please contribute the functionality?

AKxy4321:
@nmanovic  Can I take this issue

AKxy4321:
@nmanovic Are there any limitations on the Open Source Library FiftyOne for the importing and exporting datasets problem, because if there are no problems then I can automate the dataset conversion using that and add some other features to make it more convenient for the user to use.

nmanovic:
@AKxy4321 , no limitations, but we are using Datumaro library for the purpose currently.

PR #7720, opened 2024-04-02
[GSoC2024] Automatically Creating Superuser - Issue #3430

AKxy4321:
#3430 
Added method to create default superuser if not specified or created before.
Also added codes to delete and check all users present for simiplicity and debugging purposes

### Checklist

### License

### DIFF
--- a/check_users.py
+++ b/check_users.py
@@ -1,0 +1,11 @@
+import os
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "cvat.settings.development")
+
+import django
+django.setup()
+
+from django.contrib.auth.models import User
+
+users = User.objects.all()
+for user in users:
+    print(user.username)
--- a/delete_users.py
+++ b/delete_users.py
@@ -1,0 +1,15 @@
+import os
+os.environ.setdefault("DJANGO_SETTINGS_MODULE", "cvat.settings.development")
+
+import django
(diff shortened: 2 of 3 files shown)


AKxy4321:
Added method to create default superuser if no superuser is created
I have added the changes to manage.py, the relevant code will execute only if runserver argument is present.
If present, it then checks if any superuser exists, if no one exists then it will create a default one.
 
Added check and delete users code to increase convenience while developing

nmanovic (changes requested):
Hi @AKxy4321 , thank you for the PR. Let's try to follow recommendation in the original issue. Several comments:
* We don't need check_users.py and delete_users.py scripts. They can be deleted.
* We don't need to modify manage.py script. It isn't necessary.
* A good approach to implement the feature is to extend cvat/backend_entrypoint.sh
* Please add changelog

Several extra notes:
* We will need to create a superuser only in case DJANGO_SUPERUSER_* env variables are set. Please see https://docs.djangoproject.com/en/5.0/ref/django-admin/#createsuperuser. Also, look at the stackoverflow recommendation: https://stackoverflow.com/questions/6244382/how-to-automate-createsuperuser-on-django
* We will need to create a superuser only in case we don't have a superuser at all. Need to check that before creation.

Good luck!

PR #7734, opened 2024-04-05
Split cvat/apps/iam/permissions.py into per-app permissions modules

SpecLad:
### Motivation and context

This improves the dependency graph of apps with regards to IAM and other apps. IAM is a generic app, so it should not depend on specific apps, like `webhooks`.

It's also easier to work with a bunch of smaller files rather than one giant one.

Having permissions classes in separate modules introduces the risk that they will never be loaded directly. To mitigate that, add a function to ensure permissions are loaded and call it from every app.

In a future PR, I plan to move the `.rego` files into their respective apps, as well.

### How has this been tested?

I made sure that `OpenPolicyAgentPermission.__subclasses__()` returns all permission classes. I also manually checked one endpoint in each app to ensure that permission was still being denied when appropriate.

### Checklist

### License


cvat/apps/analytics_report/permissions.py
```diff
@@ -0,0 +1,80 @@
+from django.conf import settings
```
nmanovic, line 1: copyright is missing. Need to add.

cvat/apps/engine/apps.py
```diff
@@ -13,3 +13,6 @@ def ready(self):
         import cvat.apps.engine.signals
         # Required in order to silent "unused-import" in pyflake
         assert cvat.apps.engine.signals
+
```
nmanovic, line 16: Need to update the copyright. Please use https://marketplace.visualstudio.com/items?itemName=ymotongpoo.licenser to add/update headers automatically.

cvat/apps/lambda_manager/apps.py
```diff
@@ -3,3 +3,7 @@
 
 class LambdaManagerConfig(AppConfig):
     name = 'cvat.apps.lambda_manager'
+
+    def ready(self) -> None:
```
nmanovic, line 7: What do you think about cvat.apps.iam.permissions.AppConfigEx class with implemented ready() method by default? We can try to handle several additional modules such as signals.py and permissions.py. All apps have to use AppConfigEx.

nmanovic (changes requested):
I really like the PR. It isolates apps. But need to add headers in some files and I would recommend to implement the default AppConfig class with overloaded ready method by default. Use it to handle modules such as signals.py and permissions.py

cvat/apps/analytics_report/permissions.py
SpecLad: Thanks, updated. I also updated the copyright banners in `*/apps.py`.

Note that since I just moved code from one place to another, I copied the original copyright years from `iam/permissions.py`.

cvat/apps/lambda_manager/apps.py
SpecLad: IMO, a class that loads both signals and permissions isn't a very compelling abstraction. Signals have nothing to do with permissions, or with IAM as a whole. An app may only use one of them (in fact, the app on which you posted this comment does not use signals).

I suppose I could add two separate mixin classes, so you could have something like:

```python
class EngineConfig(AppConfig, IamAppMixin, SignalAppMixin):
    name = 'cvat.apps.engine'
```

but the amount of code saved in this approach is pretty marginal.

nmanovic (approved):

Issue #6502, opened 2023-07-18
REST API handles multiline attributes incorrect way


bsekachev:
### My actions before raising this issue
- [x] Read/searched [the docs](https://github.com/cvat-ai/cvat/tree/master#documentation)
- [x] Searched [past issues](/issues)

### Steps to Reproduce (for bugs)
Server request:

![image](https://github.com/opencv/cvat/assets/40690378/608e757b-6f2a-4dba-831b-a01073a20443)

Request payload:

![image](https://github.com/opencv/cvat/assets/40690378/389eac30-08a2-4145-9861-765eb8060a84)

Response payload:
![image](https://github.com/opencv/cvat/assets/40690378/39ae7080-c373-4152-9a7c-a7ac972b72ea)

The same behavior for PATCH projects/<id> endpoint with "labels" field.

### Expected Behaviour
"values" property keeps after request

### Current Behaviour
\n -> ","

### Your Environment

- Git hash commit (`git log -1`): v2.5.0
- Docker version `docker version` (e.g. Docker 17.0.05):
- Are you using Docker Swarm or Kubernetes?
- Operating System and version (e.g. Linux, Windows, MacOS):
- Code example or link to GitHub repo or gist to reproduce problem:
- Other diagnostic information / logs:
    <details>
    <summary>Logs from `cvat` container</summary>
    </details>

mach-12:
Hello @bsekachev, @nmanovic. If you could check #7736 and assign this one to me. It took quite some time to get this one right.

ritikraj26:
Hi @bsekachev @nmanovic ,
I would like to take up this issue. Could you assign this to me?

ritikraj26:
@nmanovic 
Raised the PR [here](https://github.com/cvat-ai/cvat/pull/7740)

PR #7736, opened 2024-04-06
[GSoC2024] Modified Attribute Serializer to retain values after request

mach-12:
### Motivation and context

Fixing #6502: While creating or editing a multiline text attribute on a label, it was replacing the default newlines with commas.

image

This behavior was not appropriate. The cause of this issue was identified as the `DelimitedStringListField` serializer:

https://github.com/cvat-ai/cvat/blob/70a7cc0944ea1684defce29687d56529f7a24455/cvat/apps/engine/serializers.py#L226-L240

The fix was implemented by overriding the `to_representation()` method in the parent `AttributeSerializer`. This method converts the `ListField` into a list with only a single object with the text concatenated if the `input_type` in the context is 'text'.

Nested Label Serializers:
- `LabelSerializer`
  - `SublabelSerializer`
    - `AttributeSerializer`
      - `DelimitedStringListField`

### How has this been tested?

Changes persist as intended in the Tasks, Projects and Job views. 

Screenshot 2024-04-07 at 12 22 01 AM

Screenshot 2024-04-07 at 12 22 19 AM

image

### Checklist

### License

### DIFF
--- a/changelog.d/20240407_000430_mann.compi_multilines_attribute_issue.md
+++ b/changelog.d/20240407_000430_mann.compi_multilines_attribute_issue.md
@@ -1,0 +1,4 @@
+### Fixed
+
+- Corrected handling multiline text attributes so they don’t get converted into comma-separated values by REST API
+  (<https://github.com/opencv/cvat/pull/7736>)
--- a/cvat/apps/engine/serializers.py
+++ b/cvat/apps/engine/serializers.py
@@ -225,10 +225,7 @@ class UserSerializer(serializers.ModelSerializer):
 
 class DelimitedStringListField(serializers.ListField):
     def to_representation(self, value):
-        if isinstance(value, str):
-            value = value.split('\n')
-
-        return super().to_representation(value)
+        return super().to_representation(value.split('\n'))
 
     def to_internal_value(self, data):
         return '\n'.join(super().to_internal_value(data))
--- a/cvat/apps/engine/serializers.py
+++ b/cvat/apps/engine/serializers.py
@@ -225,15 +225,25 @@ class UserSerializer(serializers.ModelSerializer):
 
 class DelimitedStringListField(serializers.ListField):
     def to_representation(self, value):
-        return super().to_representation(value.split('\n'))
+        if isinstance(value, str):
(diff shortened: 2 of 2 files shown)


Marishka17 (changes requested):
Hello, thank you for the contribution.

I see one problem here:
In the current REST API implementation, it is possible to specify several values for text attributes too (even though it looks useless). But in that case, after these changes, all values will be merged into one and this is wrong. 

We use one [CharField](https://github.com/cvat-ai/cvat/blob/f1d308e8a01c72f51263aad9cb125d3031a95dcc/cvat/apps/engine/models.py#L860) to store values and [\n](https://github.com/cvat-ai/cvat/blob/f1d308e8a01c72f51263aad9cb125d3031a95dcc/cvat/apps/engine/serializers.py#L231) to separate them. Probably you can change the separator (e.g. `\\n`), but in that case, Django migration must be added.

Marishka17:
> But how does a user trigger this intended use case in the first place

For example, users can use REST API directly

> How will the User know how to use it?

A user can check the server endpoint specification.

> So I need to:
> Define new separator,
> Add appropriate db Migration

It is possible to fix this issue using one more way - we can restrict specifying values for text attributes since this makes no sense at all. In that case, we don't need to change the separator.

IMHO, the second approach is more preferable.

bsekachev:
Hi @mach-12 

Do you have any plans regarding the pull request?

mach-12:
Yes, I need some help with understanding how to restrict specification of text values. Going to finish this PR

bsekachev:
@Marishka17, please, work together with @mach-12 to finish the PR.

Marishka17:
Hello @mach-12,  sorry for the delay, I'm going to close this PR now, but do not hesitate to reopen it when you implement more appropriate solution.

You can use the following approach to fix the issue:
- on the server: add `AttributeSerialize.validate` method and raise `ValidationError` if `values` is not empty and `input_type == models.AttributeType.TEXT` 
- on the server: add migration to clean the `values` field for all text attributes
- on the client: exclude sending default value in attribute `values` field for text attributes

PR #7744, opened 2024-04-09
Improve code for working with the OPA bundle

SpecLad:
There are three changes here:

* Don't save the bundle to a file. There seems to be no reason to do that; the bundle is very small (<8000 bytes), and we already load it fully into memory to generate the ETag, so just generate it in memory and keep it there.

  This also fixes a possible race condition where a server process starts to read the bundle file right as another server process starts up and overwrites it.

* Generate the bundle on-demand rather than at startup. This obviates the need for the `IAM_OPA_BUNDLE` environment variable.

* Generate the ETag once, instead of on every request. There's no reason to hash the same bundle over and over.

### Motivation and context

A patch I'm working on (the sequel to #7734) requires the bundle to be generated on-demand (and it seems to me like a simpler approach in general). Removing the bundle file was a consequence of that. The etag change I implemented because I was updating the related code anyway, and it seemed like a good idea.

### How has this been tested?

I checked that the `/api/auth/rules` endpoint still works and that OPA is still getting 304 responses on repeat requests.

### Checklist

### License

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -167,7 +167,6 @@
             "env": {
                 "CVAT_SERVERLESS": "1",
                 "ALLOWED_HOSTS": "*",
-                "IAM_OPA_BUNDLE": "1",
                 "DJANGO_LOG_SERVER_HOST": "localhost",
                 "DJANGO_LOG_SERVER_PORT": "8282"
             },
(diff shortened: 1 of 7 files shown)


nmanovic (approved):
Looks really great!

Issue #7762, opened 2024-04-13
Top Bar is disabled in model page


thekavikumar:
### Actions before raising this issue

- [X] I searched the existing issues and did not find anything similar.
- [X] I read/searched [the docs](https://docs.cvat.ai/docs/)

### Steps to Reproduce

1. Go to https://app.cvat.ai/models
![image](https://github.com/cvat-ai/cvat/assets/50093149/dc28b8ff-db5b-43e2-aac6-3abc6c676713)
![image](https://github.com/cvat-ai/cvat/assets/50093149/61aab72c-117c-4224-9c58-08e5507a9297)

I have noticed that in the model page, the top bar is disabled when there are no items, which makes sense. However, I cannot find any code that enables it when there are items. By default, when there are no items on any page, the top bar is not disabled. Therefore, I believe we should follow the same approach for the model page. 

### Expected Behavior

The top bar should be enabled when there are no items on the page. This allows us to match the styles with other pages. Any other ideas are also welcome.

thekavikumar:
@nmanovic please review this issue and assign me.

thekavikumar:
As you can see in app.cvat.ai also when models are available on the model page the top bar is still disabled.
Which does not allow the user to search for the models, if we have a huge volume of models then it will become complicated to go through them one by one instead of searching them using a search bar.

syedbarimanjan:
There is no use of the top at the moment the functionality is not implemented which is why it is disabled by default. @thekavikumar

thekavikumar:
you mean the search functionality? @syedbarimanjan

syedbarimanjan:
> you mean the search functionality? @syedbarimanjan

Yes all of those are not implemented the search, filtering and all.

thekavikumar:
@syedbarimanjan I see, alright!

PR #7765, opened 2024-04-14
[GSoC2024] Top Bar Disabled in Model Page (Resolved Issue#7762)

thekavikumar:
I've modified the top bar to get enabled when there are models available.

Fix: #7762


bsekachev:
Unfortunately, it will not work, these features are not implemented on server for models. 
Also changes in cvat-core are necessary and in cvat-ui.

thekavikumar:
@bsekachev Oh alright, then I will close the pr.

PR #7796, opened 2024-04-23
Fixed showing annotations at 0 frame

klakhov:
### Motivation and context

After the PR #7714 annotations are not loaded to the store with `GET_JOB_SUCCESS` action, instead we load them in separate async action `fetchAnnotationsAsync`. 
That leads to the problem: `CanvasWrapper` component is initially mounted with empty `annotations` array. That leads to such function calls: `updateCanvas` -> ` canvasInstance.setup` -> [the promise is created](https://github.com/cvat-ai/cvat/blob/develop/cvat-canvas/src/typescript/canvasModel.ts#L600) that will set this.data.objects to empty array.
Then the annotations are put to the store, `CanvasWrapper` runs update hook, where the chain is called again: `updateCanvas` -> ` canvasInstance.setup` -> but instead of promise, [the objects are assigned right away](https://github.com/cvat-ai/cvat/blob/develop/cvat-canvas/src/typescript/canvasModel.ts#L551). After that when the first promise will be executed the objects will be empty which leads to empty canvas wihout annotations.

This PR fixes the problem by introducing `initialized` state status for annotations so `CanvasWrapper` is not mounted before annotations are loaded. (+it removes one unnecessary re-render) 

### Checklist

### License

### DIFF
--- a/changelog.d/20240423_142255_klakhov_fix_missing_annotations.md
+++ b/changelog.d/20240423_142255_klakhov_fix_missing_annotations.md
@@ -1,0 +1,4 @@
+### Fixed 
+
+- Annotations are not shown on the `0` frame
+  (<https://github.com/cvat-ai/cvat/pull/7796>)
--- a/changelog.d/20240423_142255_klakhov_fix_missing_annotations.md
+++ b/changelog.d/20240423_142255_klakhov_fix_missing_annotations.md
@@ -1,4 +1,4 @@
-### Fixed 
+### Fixed
 
-- Annotations are not shown on the `0` frame
+- Annotations are not shown on the `0` frame sometimes
   (<https://github.com/cvat-ai/cvat/pull/7796>)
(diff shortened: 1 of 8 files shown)


changelog.d/20240423_142255_klakhov_fix_missing_annotations.md
```diff
@@ -0,0 +1,4 @@
+### Fixed 
+
+- Annotations are not shown on the `0` frame
```
bsekachev, line 3:
```suggestion
- Annotations are not shown on the `0` frame sometimes
```

cvat-canvas/src/typescript/canvasModel.ts
```diff
@@ -608,6 +611,8 @@ export class CanvasModelImpl extends MasterImpl implements CanvasModel {
                     } else {
                         this.data.exception = new Error('Unknown error occured when fetching image data');
                     }
+                    this.data.objects = prevObjects;
+                    this.data.zLayer = prevZLayer;
```
bsekachev, line 615: Perhaps it should be:

```js
if (this.data.objects === objectStates && this.data.zLayer === zLayer) {
     this.data.objects = prevObjects;
     this.data.zLayer = prevZLayer;
}
```

Otherwise these values may be not relevant to the moment when the promise is already rejected (one more `setup` has been called in the time range).

bsekachev: Also let's add a comment with explanation to the code

PR #7803, opened 2024-04-25
Disable further checking permissions if at least one is False

bsekachev:
### Motivation and context
To avoid extra requests to the database, opa container and code execution on the server

### Checklist

### License

### DIFF
--- a/cvat/apps/iam/permissions.py
+++ b/cvat/apps/iam/permissions.py
@@ -2015,20 +2015,14 @@ class PolicyEnforcer(BasePermission):
         if self.is_metadata_request(request, view) or obj and is_public_obj(obj):
             return True
 
-        permissions: List[OpenPolicyAgentPermission] = []
         iam_context = get_iam_context(request, obj)
+        for perm_class in OpenPolicyAgentPermission.__subclasses__():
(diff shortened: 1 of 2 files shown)


cvat/apps/iam/permissions.py
```diff
@@ -2025,6 +2025,8 @@ def check_permission(self, request, view, obj) -> bool:
         for perm in permissions:
             result = perm.check_access()
             allow &= result.allow
+            if not allow:
+                break
```
SpecLad, line 2029: I don't think the `allow` variable is needed at all, given this change. You can just replace this with:

```python
if not result.allow:
    return False
```

and the last line with

```python
return True
```

bsekachev: agree

SpecLad: In fact... I don't think the `permissions` list is needed either. You can merge this loop with the previous loop, and just check each permission as it is created.

nmanovic:
@bsekachev , let's add a line into changelog. The optimization looks good. Need double check. Do you have any numbers to estimate the enhancement?

SpecLad (approved):

PR #7815, opened 2024-04-28
fix: Apply custom patches

jackylamhk:
- Update Django settings
- Disable CVAT self-signup
- Update nginx and uvicorn config


jackylamhk:
Accidentally opened this PR for the upstream branch - please ignore.

PR #7903, opened 2024-05-17
[WIP][Dependent] Fix task creation with cloud storage data

Marishka17:
### Motivation and context

This PR contains several fixes:
- Optimized memory usage by not keeping all downloaded images/part of images in memory while creating a manifest file
- Optimized the number of requests to CS providers by downloading only images from a specified range (`use_cache` in True/False)
### How has this been tested?

This table contains metrics (required time and memory usage) obtained before this fix and after.
 
| with/without fix| Threads count | Time for partially files downloading & preparing manifest | Memory usage diff after `_create_task_manifest_from_cloud_data` (MB) |
| ------ | ------ | ------ | ------ |
| without | 1 | 0:01:02.797817 | 222 |
| without | 2 | 0:00:11.307688 | 225 |
| with | 1 | 0:00:12.622483 | 16.5 |
| with | 2 | 0:00:13.004692 | 17.2 |

Testing dataset contains 4000 images, its size is about 400MB, 64 KB was downloaded on average for each image.

<details>
   <summary>Details</summary> 
      
      after fix:
      1 thread
      Required time:  0:00:12.622483
      280.7 MiB      0.0 MiB           1                   if data['stop_frame'] is not None:
         727                                                             stop_frame = min(stop_frame, data['stop_frame'])
         728    297.2 MiB     16.5 MiB           2                   _create_task_manifest_from_cloud_data(
         729    280.7 MiB      0.0 MiB           1                       db_data.cloud_storage, sorted_media, manifest,
         730    280.7 MiB      0.0 MiB           1                       start_frame=db_data.start_frame, stop_frame=stop_frame, step=db_data.get_frame_step()
         731                                                         )
         732                                         
         733    297.2 MiB      0.0 MiB           1       av_scan_paths(upload_dir)
      
      2 threads
      Required time:  0:00:13.004692
      726    281.1 MiB      0.0 MiB           1                   if data['stop_frame'] is not None:
         727                                                             stop_frame = min(stop_frame, data['stop_frame'])
         728    298.3 MiB     17.2 MiB           2                   _create_task_manifest_from_cloud_data(
         729    281.1 MiB      0.0 MiB           1                       db_data.cloud_storage, sorted_media, manifest,
         730    281.1 MiB      0.0 MiB           1                       start_frame=db_data.start_frame, stop_frame=stop_frame, step=db_data.get_frame_step()
         731                                                         )
         732                                         
         733    298.3 MiB      0.0 MiB           1       av_scan_paths(upload_dir)
      
      without fix:
      1 thread
      Required time:  0:01:02.797817
      688    280.5 MiB      0.0 MiB           1               if manifest_file:
         689                                                         # Define task manifest content based on cloud storage manifest content and uploaded files
         690                                                         _create_task_manifest_based_on_cloud_storage_manifest(
         691                                                             sorted_media, cloud_storage_manifest_prefix,
         692                                                             cloud_storage_manifest, manifest)
         693                                                     else: # without manifest file but with use_cache option
         694                                                         # Define task manifest content based on list with uploaded files
         695    502.6 MiB    222.1 MiB           1                   _create_task_manifest_from_cloud_data(db_data.cloud_storage, sorted_media, manifest)
         696                                         
         697    502.6 MiB      0.0 MiB           1       av_scan_paths(upload_dir)
      
      2 threads
      Required time:  0:00:11.307688
      280.8 MiB      0.0 MiB           1               if manifest_file:
         689                                                         # Define task manifest content based on cloud storage manifest content and uploaded files
         690                                                         _create_task_manifest_based_on_cloud_storage_manifest(
         691                                                             sorted_media, cloud_storage_manifest_prefix,
         692                                                             cloud_storage_manifest, manifest)
         693                                                     else: # without manifest file but with use_cache option
         694                                                         # Define task manifest content based on list with uploaded files
         695    506.0 MiB    225.2 MiB           1                   _create_task_manifest_from_cloud_data(db_data.cloud_storage, sorted_media, manifest)
         696                                         
         697    506.0 MiB      0.0 MiB           1       av_scan_paths(upload_dir)
</details>

### Checklist

### License

## Summary by CodeRabbit

- **New Features**
  - Introduced threading logic for bulk downloads to enhance performance based on CPU capabilities and file count.
  - Added new classes `Named` and `NamedBytesIO` for better handling of file attributes in image processing.

- **Improvements**
  - Optimized memory usage and request handling for creating manifest files.
  - Enhanced handling of image sources and iteration logic in `DatasetImagesReader`.

- **Configuration**
  - Added global settings for cloud data downloading thread management.

- **Bug Fixes**
  - Improved retrieval of preview information from manifests to ensure robustness.

- **Documentation**
  - Updated internal documentation to reflect new threading logic and image processing enhancements.


utils/dataset_manifest/core.py
```diff
@@ -182,37 +193,69 @@ def step(self):
...
             else:
-                yield dict()
+                for image in self._sources:
+                    yield self._get_img_properties(image)
+        else:
+            sources = iter(self._sources) if not self._is_generator_used else self._sources
+
+            assert self.start is not None
+            assert self.stop is not None
+            assert self.step is not None
```
Marishka17, line 250: remove debug code

utils/dataset_manifest/core.py
```diff
@@ -383,8 +428,9 @@ def _parse_line(self, line):
                 offset = self._index[line]
                 manifest_file.seek(offset)
                 properties = manifest_file.readline()
+                # fixme: remove from common class
```
Marishka17, line 431: remove or fix comments not related to this PR

cvat/apps/engine/task.py
```diff
@@ -686,7 +713,13 @@ def _update_status(msg: str) -> None:
                     cloud_storage_manifest, manifest)
             else: # without manifest file but with use_cache option
                 # Define task manifest content based on list with uploaded files
-                _create_task_manifest_from_cloud_data(db_data.cloud_storage, sorted_media, manifest)
+                stop_frame = len(sorted_media)
+                if data['stop_frame'] is not None:
```
Marishka17, line 717: refactor code to remove duplicates

Marishka17:
@zhiltsov-max, Could you please take a look at the PR?  I've added small comments that should be fixed but I can fix them only in the evening probably.

utils/dataset_manifest/core.py
```diff
@@ -182,37 +193,69 @@ def step(self):
...
+        if self._use_image_hash:
+            image_properties['checksum'] = md5_hash(img)
 
-                yield image_properties
+        return image_properties
+
+    def __iter__(self):
+        if not self.are_sources_filtered:
+            if self.stop is not None:
+                index_range = range(self.start, self.stop, self.step)
```
zhiltsov-max, line 227: Probably, +1 is required for stop.

utils/dataset_manifest/core.py
```diff
@@ -182,37 +193,69 @@ def step(self):
...
+                for image in self._sources:
+                    yield self._get_img_properties(image)
+        else:
+            sources = iter(self._sources) if not self._is_generator_used else self._sources
+
+            assert self.start is not None
+            assert self.stop is not None
+            assert self.step is not None
+
+            index_range = range(self.start, self.stop, self.step)
```
zhiltsov-max, line 252: Probably, + 1 is required for stop.

utils/dataset_manifest/core.py
```diff
@@ -182,37 +193,69 @@ def step(self):
...
+            for idx in range(self.stop):
```
zhiltsov-max, line 253: Probably, + 1 is required for stop.

utils/dataset_manifest/core.py
```diff
@@ -611,6 +665,13 @@ def create(self, content=None, _tqdm=None):
 
         self.set_index()
 
+    def get_first_not_empty_item(self) -> ImageProperties:
+        for _, image in self:
+            if image:
+                return image
+
+        raise InvalidManifestError('Manifest is empty')
```
zhiltsov-max, line 673: It feels like this API change should be at least reflected in the PR description.

cvat/apps/engine/task.py
```diff
@@ -481,11 +481,25 @@ def _create_task_manifest_from_cloud_data(
...
+    filtered_sorted_media = [sorted_media[idx] for idx in index_range]
+    content_generator = cloud_storage_instance.bulk_download_to_memory(filtered_sorted_media)
+
+    manifest.link(
+        sources=content_generator,
+        DIM_3D=dimension == models.DimensionType.DIM_3D,
+        stop=stop_frame,
+        start=start_frame,
+        step=step,
+        are_sources_filtered=True,
```
zhiltsov-max, line 501: Can this parameter be removed? E.g. if the `sources` contain something like a list of paths or bytes or the resulting structure (which is now just a dict).

utils/dataset_manifest/core.py
```diff
@@ -182,37 +193,69 @@ def step(self):
...
-                if self._use_image_hash:
-                    image_properties['checksum'] = md5_hash(img)
+        if self._use_image_hash:
+            image_properties['checksum'] = md5_hash(img)
 
-                yield image_properties
+        return image_properties
+
+    def __iter__(self):
+        if not self.are_sources_filtered:
```
zhiltsov-max, line 225: Consider replacing all these branches with just 1 algorithm instead. It will help to avoid problems such as updating only one branch when all of them need to be updated.

utils/dataset_manifest/core.py
```diff
@@ -182,37 +193,69 @@ def step(self):
...
+            assert self.stop is not None
+            assert self.step is not None
+
+            index_range = range(self.start, self.stop, self.step)
+            for idx in range(self.stop):
+                if idx in index_range:
+                    image = next(sources)
+                    yield self._get_img_properties(image)
+                else:
+                    yield dict()
```
zhiltsov-max, line 258: Consider this as a function, which can be applied to an iterator and return an iterator.

zhiltsov-max (commented):
There are quite a few places where stop is changed from (the last valid index) to (the first index after the last valid index). In most places in CVAT it's used in the first meaning, I suggest avoiding changing this to avoid confusion.

utils/dataset_manifest/core.py
```diff
@@ -139,23 +140,33 @@ def __iter__(self) -> Iterator[Union[int, Tuple[int, int, str]]]:
...
+    ):
+        self._is_generator_used = isgenerator(sources)
+
+        if not self._is_generator_used:
+            raw_data_used = not isinstance(sources[0], str)
+            func = (lambda x: x.filename) if raw_data_used else None
+            self._sources = sort(sources, sorting_method, func=func)
+        else:
+            if sorting_method != SortingMethod.PREDEFINED:
+                raise ValueError('Only SortingMethod.PREDEFINED can be used with generator')
```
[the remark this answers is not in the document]
Marishka17: No, we cannot apply sorting in that case because we don't want to keep all files in memory to perform sorting. Besides that, `ThreadPoolExecutor.map` keeps sequence order and the list of files is sorted before downloading step.

cvat/apps/engine/cache.py
```diff
@@ -312,7 +312,7 @@ def _prepare_cloud_preview(self, db_storage):
             manifest.set_index()
```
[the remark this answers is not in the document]
Marishka17: Not really because without this optimization it was impossible to create a task from the cloud storage with `use_cahce==True` with a huge dataset because the background job would crash due to reaching the memory limit.

cvat/apps/engine/cloud_provider.py
```diff
@@ -196,29 +202,39 @@ def optimally_image_download(self, key: str, chunk_size: int = 65536) -> BytesIO
...
-                return pool.map(func, files)
+        if threads_number is None:
+            threads_number = min(CPU_NUMBER, MAX_THREADS_NUMBER, max(math.ceil(len(files) / NUMBER_OF_FILES_PER_THREAD), 1))
         else:
-            slogger.glob.warning('Download files to memory in series in one thread.')
-            return [func(f) for f in files]
+            threads_number = min(threads_number, CPU_NUMBER, MAX_THREADS_NUMBER)
+
+        if threads_number == 1:
+            slogger.glob.info('Download files to memory in series in one thread.')
```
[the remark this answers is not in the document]
Marishka17: removed

cvat/apps/engine/task.py
```diff
@@ -481,11 +481,25 @@ def _create_task_manifest_from_cloud_data(
...
+    stop_frame: int,
+    start_frame: int = 0,
+    step: int = 1,
     dimension: models.DimensionType = models.DimensionType.DIM_2D,
 ) -> None:
     cloud_storage_instance = db_storage_to_storage_instance(db_storage)
-    content = cloud_storage_instance.bulk_download_to_memory(sorted_media)
-    manifest.link(sources=content, DIM_3D=dimension == models.DimensionType.DIM_3D)
+    index_range = range(start_frame, stop_frame, step)
+    filtered_sorted_media = [sorted_media[idx] for idx in index_range]
```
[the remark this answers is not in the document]
Marishka17: Are you talking about real storage content? Or about storage content defined in server code? In any case, I believe this should be fixed in a separate PR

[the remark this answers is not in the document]
Marishka17: yeah, thanks for pointing that out

utils/dataset_manifest/core.py
```diff
@@ -139,23 +140,33 @@ def __iter__(self) -> Iterator[Union[int, Tuple[int, int, str]]]:
...
+            self._sources = sources
         self._meta = meta
         self._data_dir = kwargs.get('data_dir', None)
         self._use_image_hash = use_image_hash
         self._start = start
-        self._stop = stop if stop else len(sources)
+        self._stop = stop if stop or self._is_generator_used else len(sources)
```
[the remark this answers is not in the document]
Marishka17: added using `islice` in `__iter__` method

utils/dataset_manifest/core.py
Marishka17: I believe I've updated all places where `stop_frame` is defined and then used in ranges

cvat/apps/engine/task.py
Marishka17: Initially, I would like to add handling not only cases when the `sources` variable is generator returned after partial downloading of images but also add the ability to create manifest from images generator without passing stop frame. But I've removed handling this case since we do not use it anywhere.

utils/dataset_manifest/core.py
Marishka17: I do not see here a necessity to move this code into a separate function since now __iter__ does not look like a huge function

cvat/apps/engine/cloud_provider.py
```diff
@@ -35,6 +37,16 @@
...
 
+CPU_NUMBER = get_cpu_number()
+MAX_THREADS_NUMBER = 4
+NUMBER_OF_FILES_PER_THREAD = 1000
+
+def normalize_threads_number(threads_number: Optional[int], number_of_files: int) -> int:
+    if threads_number is None:
+        return min(CPU_NUMBER, MAX_THREADS_NUMBER, max(math.ceil(number_of_files / NUMBER_OF_FILES_PER_THREAD), 1))
+
+    return min(threads_number, CPU_NUMBER, MAX_THREADS_NUMBER)
```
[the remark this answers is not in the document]
Marishka17: I prefer my version

utils/dataset_manifest/core.py
```diff
@@ -182,35 +194,41 @@
...
-                yield image_properties
+        return image_properties
+
+    def __iter__(self):
+        sources = self._sources if self._is_generator_used else islice(self._sources, self.start, self.stop + 1, self.step)
+
+        for idx in range(self.stop + 1):
+            if idx in range(self.start, self.stop + 1, self.step):
+                image = next(sources)
+                yield self._get_img_properties(image)
```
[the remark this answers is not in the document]
Marishka17: your code is incorrect

cvat/apps/engine/task.py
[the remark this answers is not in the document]
zhiltsov-max: The real storage contents.
1. The user selects some files in task creation UI
2. The task creation request is put into the queue
3. The storage contents are modified
4. The task creation starts

utils/dataset_manifest/core.py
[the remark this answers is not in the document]
zhiltsov-max: What do you mean by "keep all the files in the memory"? Getting the list of files from the storage should be enough for sorting and/or filtering the range. After the requested segment of the list is obtained, it can be sorted as requested. Then, the actual images can be downloaded for further processing.

utils/dataset_manifest/core.py
```diff
@@ -182,35 +194,41 @@ def step(self):
...
```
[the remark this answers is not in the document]
zhiltsov-max: @coderabbitai, this code must produce empty dicts for skipped indices in the range.

cvat/apps/engine/cloud_provider.py
```diff
@@ -196,29 +206,30 @@ def optimally_image_download(self, key: str, chunk_size: int = 65536) -> BytesIO
...
+        *,
+        threads_number: Optional[int] = None,
+    ) -> None:
+        threads_number = normalize_threads_number(threads_number, len(files))
+
         args = zip(files, [os.path.join(upload_dir, f) for f in files])
         if threads_number > 1:
             with ThreadPool(threads_number) as pool:
-                return pool.map(lambda x: self.download_file(*x), args)
+                pool.map(lambda x: self.download_file(*x), args)
```
zhiltsov-max, line 231: Probably, you need to wait for the results of the call. The function returns futures to be awaited without waiting for their completion. A simple way is to add a `list()` call for the map result.

https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.map

utils/dataset_manifest/core.py
```diff
@@ -182,35 +194,41 @@ def step(self):
...
-        for idx in range(self._stop):
-            if idx in self.range_:
-                image = next(sources)
-                img = Image.open(image, mode='r')
+    def _get_img_properties(self, image: Union[str, BytesIO]) -> Dict[str, Any]:
+        img = Image.open(image, mode='r')
+        if self._data_dir:
+            img_name = os.path.relpath(image, self._data_dir)
+        else:
+            img_name = os.path.basename(image) if isinstance(image, str) else image.filename
```
zhiltsov-max, line 202: The `BytesIO` class doesn't have the `filename` member.

utils/dataset_manifest/core.py
zhiltsov-max: Agree, now the function seems to be small enough.

cvat/apps/engine/task.py
```diff
@@ -657,7 +671,20 @@ def _update_status(msg: str) -> None:
...
+
+            if media['image']:
+                start_frame = db_data.start_frame
+
+                stop_frame = len(filtered_data)
+                if data['stop_frame'] is not None:
+                    stop_frame = min(stop_frame, data['stop_frame'] + 1)
+
+                step = db_data.get_frame_step()
+                if start_frame or step != 1 or stop_frame != len(filtered_data):
```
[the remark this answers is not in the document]
zhiltsov-max: Seems to be fixed now.

cvat/apps/engine/cloud_provider.py
zhiltsov-max, line 231: The `map()` function returns a generator. The results can be missing when the code here exits. Consider adding a `list()` call for the map result.

https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.map

utils/dataset_manifest/core.py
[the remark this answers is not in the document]
zhiltsov-max: If the list of files is first obtained from the storage, gets filtered, sorted, and then the requested files are downloaded and passed here in a generator, it's ok.

cvat/apps/engine/cloud_provider.py
Marishka17: Honestly, it is not true since `ThreadPool` was used here, but okay, I switched to `ThreadPoolExecutor`

utils/dataset_manifest/core.py
Marishka17: Fixed types

cvat/apps/engine/cloud_provider.py
zhiltsov-max: Ok, you're right, I checked the wrong type. The updated version looks ok to me, but I'd probably just use a `map` call, as it was earlier, for simplicity.

Marishka17: > but I'd probably just use a map call, as it was earlier, for simplicity

I've used `map` in the `bulk_download_to_memory` method because we need to keep the order of files, but here we don't need that and I guess using `as_complited` will be faster, but I've not checked yet.

cvat/apps/engine/cloud_provider.py
```diff
@@ -31,10 +32,27 @@
...
 ImageFile.LOAD_TRUNCATED_IMAGES = True
 
+CPU_NUMBER = get_cpu_number()
+
+def normalize_threads_number(threads_number: Optional[int], number_of_files: int) -> int:
+    if threads_number is None:
+        return min(CPU_NUMBER, settings.CVAT_MAX_THREADS_NUMBER_FOR_DATA_DOWNLOADING, max(math.ceil(number_of_files / settings.CVAT_NUMBER_OF_FILES_PER_THREAD), 1))
+
+    return min(threads_number, CPU_NUMBER, settings.CVAT_MAX_THREADS_NUMBER_FOR_DATA_DOWNLOADING)
+
```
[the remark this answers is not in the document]
Marishka17: @coderabbitai, they are configurable now since these settings were moved into settings/base.py. I don't think that it is necessary to add such env variables. Please check the settings one more time and don't repeat other people's comments.

cvat/apps/engine/cloud_provider.py
zhiltsov-max: Seems like `wait()` would be a more focused choice for this use case. But speaking of downloading performance, I doubt any of these choices would affect it. This is handled by the pool task queue, which executes the submitted tasks after the previous ones are completed (not collected).

https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.wait

cvat/apps/engine/cloud_provider.py
```diff
@@ -31,10 +32,43 @@
...
 
+CPU_NUMBER = get_cpu_number()
+
+def normalize_threads_number(
+    threads_number: Optional[int], number_of_files: int
+) -> int:
+    threads_number = (
+        min(
+            CPU_NUMBER,
+            settings.CVAT_MAX_THREADS_NUMBER_FOR_DATA_DOWNLOADING,
```
zhiltsov-max, line 56:
```suggestion
            settings.CVAT_MAX_THREADS_NUMBER_FOR_DATA_DOWNLOADING,
```
Consider renaming to `CVAT_MAX_DATA_DOWNLOADING_THREADS` for brevity.

Marishka17:
@coderabbitai, can you please remove all your comments from this PR?

cvat/apps/engine/task.py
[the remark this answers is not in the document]
Marishka17: Honestly, right now I'm not sure how it should be fixed and should be fixed at all, but I suggest submitting a new issue.

utils/dataset_manifest/core.py
Marishka17: why? I guess it is internal API and it's not worth mentioning that it was added

Marishka17:
@zhiltsov-max, I've applied all your comments, could you please take a look at the PR one more time?

utils/dataset_manifest/core.py
zhiltsov-max: Ok, I see it's handled internally, so it can be ignored.

cvat/apps/engine/cloud_provider.py
```diff
@@ -176,51 +210,52 @@ def optimally_image_download(self, key: str, chunk_size: int = 65536) -> BytesIO
...
+        threads_number: Optional[int] = None,
+    ) -> None:
+        threads_number = normalize_threads_number(threads_number, len(files))
+
+        with ThreadPoolExecutor(max_workers=threads_number) as executor:
+            futures = [executor.submit(self.download_file, f, os.path.join(upload_dir, f)) for f in files]
+            done, _ = wait(futures, return_when=FIRST_EXCEPTION)
+            for future in done:
+                if ex := future.exception():
+                    raise ex
```
zhiltsov-max, line 258: I feel that this implementation can store all the thread results in memory

Given a simple script like this:
```python
from concurrent.futures import ThreadPoolExecutor
import sys
from time import sleep


class ObjectWithAllocationReporting:
    def __init__(self, i) -> None:
        self.id = i
        print("created", self.id)

    def __del__(self) -> None:
        print("deleted", self.id)

def main() -> int:
    def _foo_with_result(i: int) -> ObjectWithAllocationReporting:
        return ObjectWithAllocationReporting(i)

    def _foo_without_result(i: int) -> None:
        ObjectWithAllocationReporting(i)
        # can also be a wrapper around a function with a result
        # can also add del here to be sure

    for callback in [
        _foo_with_result,
        _foo_without_result,
    ]:
        print(callback)
        with ThreadPoolExecutor(2) as pool:
            results = pool.map(callback, range(6))
            sleep(5)
            print("slept")
            list(results)

    return 0

if __name__ == "__main__":
    sys.exit(main())
```

Output:
```
<function main.<locals>._foo_with_result at 0x7fc98bd54e50>
created 0
created 1
created 2
created 3
created 4
created 5
slept
deleted 5
deleted 4
deleted 3
deleted 2
deleted 1
deleted 0
<function main.<locals>._foo_without_result at 0x7fc98bd54ee0>
created 0
deleted 0
created 1
deleted 1
created 2
deleted 2
created 3
created 4
deleted 3
deleted 4
created 5
deleted 5
slept
```

Probably, it's not what we want for big datasets, as there is going to be some time between the tasks are started and the results are collected.

zhiltsov-max:
@Marishka17, do you want to finalize the PR?

Marishka17:
> do you want to finalize the PR?

@zhiltsov-max, probably yes

PR #7907, opened 2024-05-17
Fixed: Queued jobs are not considered in deferring logic

bsekachev:
### Checklist

### License

## Summary by CodeRabbit

- **Bug Fixes**
  - Enhanced job handling to ensure jobs in "started" status are stopped before cancellation.
  - Added a conditional check in the `handle_rq_exception` function to handle `SystemExit` exceptions.

- **New Features**
  - Improved exception handling for specific scenarios, including `SystemExit`.

- **Performance Improvements**
  - Optimized job execution by managing database connections more efficiently.

- **Chores**
  - Removed outdated formatter settings from the project configuration.


cvat/apps/engine/backup.py
```diff
@@ -967,6 +969,11 @@ def export(db_instance, request, queue_name):
             # in case the server is configured with ONE_RUNNING_JOB_IN_QUEUE_PER_USER
             # we have to enqueue dependent jobs after canceling one
             rq_job.cancel(enqueue_dependents=settings.ONE_RUNNING_JOB_IN_QUEUE_PER_USER)
+            try:
+                send_stop_job_command(rq_job.connection, rq_job.id)
+            except InvalidJobOperation:
+                pass
```
Marishka17, line 975: May be replaced with [suppress](https://docs.python.org/3/library/contextlib.html#contextlib.suppress) from contextlib
```suggestion
            with suppress(InvalidJobOperation):
                send_stop_job_command(rq_job.connection, rq_job.id)
```

cvat/apps/engine/utils.py
```diff
@@ -165,25 +165,22 @@ def define_dependent_job(
...
         # Since there is no cleanup implementation in DeferredJobRegistry,
         # this registry can contain "outdated" jobs that weren't deleted from it
         # but were added to another registry. Probably such situations can occur
         # if there are active or deferred jobs when restarting the worker container.
-        if job and job.meta.get("user", {}).get("id") == user_id and job.is_deferred
+        for registry in [queue.started_job_registry, queue.deferred_job_registry]:
+            pipeline.zrange(registry.key, 0, -1)
+        pipeline.lrange(queue.key, 0, -1)
+        results = pipeline.execute()
+        keys = [bytes_key.decode('utf-8') for result in results for bytes_key in result]
```
Marishka17, line 178: Why did you change this to use redis API instead of rq? Looks like we can receive expired jobs here since the`StartedJobRegistry.cleanup` method is not being called.

cvat/apps/engine/utils.py
```diff
@@ -165,25 +165,22 @@ def define_dependent_job(
...
+
+    all_user_jobs = [
+        job for job in queue.job_class.fetch_many(keys, queue.connection)
+        if job and job.meta.get("user", {}).get("id") == user_id
```
Marishka17, line 182: How about checking that a job is not deferred?

cvat/apps/engine/views.py
```diff
@@ -2975,6 +2978,11 @@ def _export_annotations(
             # in case the server is configured with ONE_RUNNING_JOB_IN_QUEUE_PER_USER
             # we have to enqueue dependent jobs after canceling one
             rq_job.cancel(enqueue_dependents=settings.ONE_RUNNING_JOB_IN_QUEUE_PER_USER)
+            try:
+                send_stop_job_command(rq_job.connection, rq_job.id)
+            except InvalidJobOperation:
+                pass
```
Marishka17, line 2984:
```suggestion
            with suppress(InvalidJobOperation):
                send_stop_job_command(rq_job.connection, rq_job.id)
```

cvat/apps/engine/utils.py
bsekachev: - The first problem I found here was that queued jobs are not considered in this logic (what looks like error, as it allows a user to push as many jobs to **queued** (ignoring deferred) as he wants), if all workers are busy with other tasks of other users.

The initial fix was like:

```
started_user_jobs = [
        job
        for job in queue.job_class.fetch_many(
            queue.started_job_registry.get_job_ids(), queue.connection
        )
        if job and job.meta.get("user", {}).get("id") == user_id
    ]
deferred_user_jobs = [
    job
    for job in queue.job_class.fetch_many(
        queue.deferred_job_registry.get_job_ids(), queue.connection
    )
    # Since there is no cleanup implementation in DeferredJobRegistry,
    # this registry can contain "outdated" jobs that weren't deleted from it
    # but were added to another registry. Probably such situations can occur
    # if there are active or deferred jobs when restarting the worker container.
    if job and job.meta.get("user", {}).get("id") == user_id and job.is_deferred
]
queued_user_jobs = [
    job
    for job in queue.job_class.fetch_many(
        queue.get_job_ids(), queue.connection
    )
    if job and job.meta.get("user", {}).get("id") == user_id and job.is_deferred
]
all_user_jobs = started_user_jobs + deferred_user_jobs + queued_user_jobs
```

bsekachev: Yes, because I was trying to reproduce it specially for you :)

bsekachev: And checked one more time, with the current implementation `export:project.id128-by-admin` after updating the resource goes immediately to `queued` as result of logic with avoiding circular dependencies. 

That is also not good, but that is about using the same IDs.

cvat/apps/engine/utils.py
bsekachev: > we know

Hmm, I do not know such situations to be honest 
Do you have any specific scenarios to reproduce?

bsekachev: Okay, now I saw that accidentally removed code you are talking about

bsekachev: @Marishka17 

I returned this check

cvat/apps/engine/utils.py
```diff
@@ -165,25 +165,18 @@ def define_dependent_job(
...
+    queues = [queue.deferred_job_registry, queue, queue.started_job_registry]
+    # Since there is no cleanup implementation in DeferredJobRegistry,
+    # this registry can contain "outdated" jobs that weren't deleted from it
+    # but were added to another registry. Probably such situations can occur
+    # if there are active or deferred jobs when restarting the worker container.
+    filters = [lambda job: job.is_deferred, lambda _: True, lambda _: True]
+    all_user_jobs = []
+    for q, f in zip(queues, filters):
+        job_ids = q.get_job_ids()
+        jobs = queue.job_class.fetch_many(job_ids, queue.connection)
```
Marishka17, line 177:
```suggestion
        jobs = q.job_class.fetch_many(job_ids, q.connection)
```

cvat/rq_patching.py
```diff
@@ -69,7 +71,22 @@ def custom_started_job_registry_cleanup(self, timestamp: Optional[float] = None)
 
     return job_ids
 
+def custom_hearbeat(self, timestamp: datetime, ttl: int, pipeline: Optional['Pipeline'] = None, xx: bool = False):
```
Marishka17, line 74:
```suggestion
def custom_heartbeat(self, timestamp: datetime, ttl: int, pipeline: Optional['Pipeline'] = None, xx: bool = False):
```

cvat/rq_patching.py
```diff
@@ -69,7 +71,22 @@ def custom_started_job_registry_cleanup(self, timestamp: Optional[float] = None)
...
+    self.last_heartbeat = timestamp
+    connection = pipeline if pipeline is not None else self.connection
+    connection.hset(self.key, 'last_heartbeat', utcformat(self.last_heartbeat))
+
+    # the only difference with the default implementation
+    # is to avoid adding job to started job registry
+    # that is not clear why they adds a job to started_job_registry when the job has failed
+    # also, as far as I see they also commented this line in unrealeased (2.0) version
+    exc_info = sys.exc_info()
+    is_stopped_export_job = isinstance(exc_info[1], (StopRequested, SystemExit))
```
Marishka17, line 84: It is not clear why this condition means that an `export` job has been stopped (and generally it should not mean that). Consider using a more abstract name.

cvat/rqworker.py
```diff
@@ -1,16 +1,34 @@
...
+    # This modification ensures that jobs stopped intentionally
+    # do not get their status updated or placed in the failed registry
+    # as the main server code must delete them at all
+    def handle_job_failure(self, job, queue, **kwargs):
+        # pylint: disable=access-member-before-definition
+        if self._stopped_job_id == job.id:
+            self._stopped_job_id = None
+            self.set_current_job_id(None)
+        else:
+            super().handle_job_failure(job, queue, **kwargs)
```
Marishka17, line 27: I see one problem here - a started rq job can also be stopped from django admin panel and with these changes such a job won't be handled correctly.

bsekachev: > won't be handled correctly

Do you have ideas why? If it sends kill signal to the work horse process and does not set `self._stopped_job_id`, it is probably problem with rq.

Finally, this fix does not suppose to solve all known problems with rq jobs, so:

- the problem was not added with this patch, and
- you do not see any problems in regular user scenarios, 
 
I do not think it will block us from merging.

Marishka17 (approved):

PR #7951, opened 2024-05-28
Update packages with vulnerability

novda:
### Checklist

### License

## Summary by CodeRabbit

- **Chores**
  - Updated `PIP` version to `24.0` for improved package management.
  - Upgraded `setuptools` to `>=70.0.0` for enhanced compatibility and features.
  - Updated `redis` package to `5.0.4` for better performance and stability.
  - Upgraded `fakeredis` to `2.23.2` for improved testing capabilities.


cvat-cli/requirements/base.txt
```diff
@@ -1,3 +1,3 @@
 cvat-sdk~=2.15.0
 Pillow>=10.3.0
-setuptools>=65.5.1 # not directly required, pinned by Snyk to avoid a vulnerability
+setuptools>=70.0.0
```
SpecLad, line 3: As far as I can tell, setuptools is still not directly required, so why remove the comment?

Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG PIP_VERSION=22.0.2
+ARG PIP_VERSION=24.0
```
SpecLad, line 1: IIRC, pip changed the HTTP cache location between these two versions, so you'll need to update the `--mount` options for the `RUN` commands that involve pip.

cvat-cli/requirements/base.txt
novda: Misunderstood the comment, returned it

Dockerfile
novda: That's right, it turns out the path had [moved](https://pip.pypa.io/en/stable/topics/caching/) to http-v2, changed the path, thanks!

Dockerfile
```diff
@@ -24,7 +24,7 @@ RUN apt-get update && \
 
 ARG PIP_VERSION
 ENV PIP_DISABLE_PIP_VERSION_CHECK=1
-RUN --mount=type=cache,target=/root/.cache/pip/http \
+RUN --mount=type=cache,target=/root/.cache/pip/http-v2 \
```
SpecLad, line 27: This command runs the old pip, so it should still have the old location.

Dockerfile
```diff
@@ -78,14 +78,14 @@ RUN sed -i '/^av==/d' /tmp/utils/dataset_manifest/requirements.txt
...
     DATUMARO_HEADLESS=1 python3 -m pip wheel --no-deps \
     -r /tmp/cvat/requirements/${CVAT_CONFIGURATION}.txt \
     -w /tmp/wheelhouse
 
-FROM golang:1.20.5 AS build-smokescreen
+FROM golang:1.22.3 AS build-smokescreen
 
-RUN git clone --depth=1 -b v0.0.4 https://github.com/stripe/smokescreen.git
+RUN git clone --depth=1 https://github.com/stripe/smokescreen.git
```
SpecLad, line 88: You have to check out a specific commit; otherwise, the builds will not be reproducible.

novda: Done

Dockerfile
```diff
@@ -161,6 +161,7 @@ RUN if [ "$CLAM_AV" = "yes" ]; then \
 # Install wheels from the build image
 RUN python3 -m venv /opt/venv
 ENV PATH="/opt/venv/bin:${PATH}"
+RUN python -m pip install --upgrade setuptools
```
SpecLad, line 164: What's this for?

novda: Here setuptools is already installed from apt base and there setuptools version 59.6.0 which has vulnerabilities, to install the latest version(70.0.0 now) of setuptools I had to add this line

SpecLad: What are the vulnerabilities?

SpecLad: Actually, regardless of the answer, I would suggest _removing_ setuptools instead of upgrading. I don't think we need it in the production configuration, and if we did need it, it would be listed in `production.txt` and installed alongside all other packages.

novda: As docker scout shows, the vulnerability is https://github.com/advisories/GHSA-r9hx-vwmv-q579
Deletion also solves the problem, done ✅

SpecLad (approved):

novda:
@SpecLad 
It turned out that setuptools is needed when building an application, without setuptools our main CI [failed](https://github.com/cvat-ai/cvat/actions/runs/9299460392/job/25593442172) on try run `docker load --input /tmp/cvat_server/image.tar`. So I returned the setuptools update by pip

SpecLad:
> @SpecLad It turned out that setuptools is needed when building an application, without setuptools our main CI [failed](https://github.com/cvat-ai/cvat/actions/runs/9299460392/job/25593442172) on try run `docker load --input /tmp/cvat_server/image.tar`. So I returned the setuptools update with pip

Looks like this is caused by googleapis/python-storage#40, which has already been resolved in a newer version of `google-cloud-storage`. Could you add a reference to this bug in a comment, so that we can remember to remove setuptools after we update `google-cloud-storage`?

novda:
> > @SpecLad It turned out that setuptools is needed when building an application, without setuptools our main CI [failed](https://github.com/cvat-ai/cvat/actions/runs/9299460392/job/25593442172) on try run `docker load --input /tmp/cvat_server/image.tar`. So I returned the setuptools update with pip
> 
> Looks like this is caused by [googleapis/python-storage#740](https://github.com/googleapis/python-storage/issues/740), which has already been resolved in a newer version of `google-cloud-storage`. Could you add a reference to this bug in a comment, so that we can remember to remove setuptools after we update `google-cloud-storage`?

Done ✅

PR #7958, opened 2024-05-28
Using dedicated event to store working time

### LINKED ISSUE #7884: Lack of documenation on working_time metrics of Analytics

It is impossible to understand what working_time means without going to the code or without analyzing explicitly the log events.
It is necessary to document how it is calculated. What is the "duration" of some event types etc.

Here is the small summary of how I understand the working_time:

UI collects the "events" when user interacts with buttons etc.
The structure of one single record in the event series described [here](https://docs.cvat.ai/docs/administration/advanced/analytics/#events-log-structure)
2.1 Each record has it's own timestamp
2.2 Some records has a duration
UI sends the message with the series of events by 3 triggers:
3.1. When user click "Save"
3.2 When user loads the job
3.3 Periodically by timer
When the "events" message (you can find it in Network page in your browser if needed) arrives, server calculates the working time and updates the payload with the calculated working_time.

Here is the example of how this working_time is calculated. 

If you go from the bottom to the top you'll see the first event with working_time 0. Second one is the timestamp of the first - the timestamp of the current event. Third is the second timestamp - current timestamp (third one) etc.

working_time, timestamp, event
0.355	2024-04-28T18:06:46.099Z	drag:object
0.392	2024-04-28T18:06:45.744Z	drag:object
0.451	2024-04-28T18:06:45.352Z	drag:object
0.497	2024-04-28T18:06:44.901Z	drag:object
0.593	2024-04-28T18:06:44.404Z	drag:object
0.565	2024-04-28T18:06:43.811Z	drag:object
0	2024-04-28T18:06:43.246Z	drag:object

There are some events that has a duration field. 2 of them are change:frame and save:job.

Working_time for these types is: duration of the event + time difference between the previous timestamp and a current timestamp.

Example:

working_time, timestamp, event
4.089	2024-04-28T18:07:01.346Z	change:frame
0.662	2024-04-28T18:07:00.225Z	drag:object
0.466	2024-04-28T18:06:59.563Z	resize:object
0.582	2024-04-28T18:06:59.097Z	drag:object
1.048	2024-04-28T18:06:58.515Z	drag:object

Let's say we've pushed the "play" button and changed 30 frames from frame 10 to frame 40. This means that UI would generate 30 change:frame events. Instead, these 30 events are collapsed in 1 event with a duration of 4 seconds and a timestamp of the first change:event.

To summarize the above.

Each event has it's own timestamp
Some event types has a duration. One of these kind is change:frame. Others has a duration as 0.
If there is a duration > 0 for the event, the duration is added to the working_time

A note about the time limits. If the time difference between the current timestamp and previous timestamp is <= 100 s. then the working time is calculated just like it is described above. If the time difference is > 100s, the working_time is 0 for the current event and the working_time is calculated from the second timestamp as usual.

_No response_

### Additional context

_No response_

bsekachev:
### Motivation and context
- Parsing JSON payloads to get `working_time` in general leads to low performance in Clickhouse requests. This patch will not fix it right now, but with this patch, after a period of time we may switch to new quick approach to calculate working time.
- There will not be a lot of `send:working_time` events, we may store this scope of events for a longer time (e.g. 5 years instead of one by default). 
- Finally storing working time in such events like `click:element` or `send:exception`, or `debug:info` seems not logical. 
- Also, the history showed, that as result in different bugs, these events may sometime lose information about `job_id`, `task_id`, etc. 

Resolved #7884

### Checklist

### License

## Summary by CodeRabbit

- **New Features**
  - Introduced automatic working time tracking per job, enhancing time management and reporting.

- **Bug Fixes**
  - Improved event processing to ensure accurate timestamp updates and payload handling.

- **Refactor**
  - Streamlined internal logic for handling client events and working time calculations.

- **Security**
  - Added proper permission handling to prevent unauthorized access.

### DIFF
--- a/cvat/apps/events/handlers.py
+++ b/cvat/apps/events/handlers.py
@@ -3,9 +3,9 @@
 # SPDX-License-Identifier: MIT
 
 from copy import deepcopy
-from datetime import datetime
 from contextlib import suppress
 from typing import Optional, Union
+import datetime
 import traceback
 import json
 import rq
@@ -600,6 +600,11 @@ def handle_viewset_exception(exc, context):
     return response
 
 def handle_client_events_push(request, data):
+    TIME_THRESHOLD = datetime.timedelta(seconds=100)
+    WORKING_TIME_SCOPE = 'send:working_time'
+    WORKING_TIME_RESOLUTION = datetime.timedelta(milliseconds=1)
+    org = request.iam_context["organization"]
+
     def get_end_timestamp(event: dict) -> datetime.datetime:
         COLLAPSED_EVENT_SCOPES = frozenset(("change:frame",))
         if event["scope"] in COLLAPSED_EVENT_SCOPES:
@@ -608,9 +613,6 @@ def handle_client_events_push(request, data):
         return event["timestamp"]
 
     def generate_wt_event(job_id: int | None, wt: datetime.timedelta, common: dict) -> EventSerializer | None:
-        WORKING_TIME_SCOPE = 'send:working_time'
-        WORKING_TIME_RESOLUTION = datetime.timedelta(milliseconds=1)
-
         if wt.total_seconds():
             task_id = None
             project_id = None
@@ -638,12 +640,7 @@ def handle_client_events_push(request, data):
                 "payload": json.dumps({ "working_time": value })
             })
 
-            event.is_valid(raise_exception=True)
             return event
-
-    TIME_THRESHOLD = datetime.timedelta(seconds=100)
-
-    org = request.iam_context["organization"]
 
     if previous_event := data["previous_event"]:
         previous_end_timestamp = get_end_timestamp(previous_event)
@@ -676,7 +673,7 @@ def handle_client_events_push(request, data):
         previous_job_id = job_id
 
     common = {
-        "timestamp": str(datetime.datetime.now(datetime.timezone.utc)),
+        "timestamp": datetime.datetime.now(datetime.timezone.utc),
         "user_id": request.user.id,
         "user_name": request.user.username,
         "user_email": request.user.email,
@@ -687,6 +684,7 @@ def handle_client_events_push(request, data):
     for job_id in working_time_per_job:
         event = generate_wt_event(job_id, working_time_per_job[job_id], common)
         if event:
+            event.is_valid(raise_exception=True)
             record_server_event(
                 scope=event.validated_data['scope'],
                 request_id=request_id(),
--- a/cvat/apps/events/handlers.py
+++ b/cvat/apps/events/handlers.py
@@ -135,7 +135,7 @@ def user_name(instance=None):
 
 def user_email(instance=None):
     current_user = get_user(instance)
-    return _get_value(current_user, "email")
+    return _get_value(current_user, "email") or None
 
 def organization_slug(instance):
     if isinstance(instance, Organization):
@@ -666,7 +666,7 @@ def handle_client_events_push(request, data: dict):
             "timestamp": data["events"][0]["timestamp"],
             "user_id": request.user.id,
             "user_name": request.user.username,
-            "user_email": request.user.email,
+            "user_email": request.user.email or None,
             "org_id": getattr(org, "id", None),
             "org_slug": getattr(org, "slug", None),
         }
--- a/cvat/apps/events/handlers.py
+++ b/cvat/apps/events/handlers.py
@@ -604,6 +604,16 @@ def handle_client_events_push(request, data):
     WORKING_TIME_RESOLUTION = datetime.timedelta(milliseconds=1)
     org = request.iam_context["organization"]
 
+    def build_complex_id(event):
+        job_id = event.get("job_id")
+        task_id = event.get("task_id")
+        project_id = event.get("project_id")
+        return ':'.join((str(job_id), str(task_id), str(project_id)))
+
+    def parse_complex_id(id):
+        job_id, task_id, project_id = map(lambda x: None if x == 'None' else int(x), id.split(':'))
+        return job_id, task_id, project_id
+
     def get_end_timestamp(event: dict) -> datetime.datetime:
         COLLAPSED_EVENT_SCOPES = frozenset(("change:frame",))
         if event["scope"] in COLLAPSED_EVENT_SCOPES:
@@ -612,7 +622,7 @@ def handle_client_events_push(request, data):
         return event["timestamp"]
 
     def generate_wt_event(id: str, wt: datetime.timedelta, common: dict) -> EventSerializer | None:
-        job_id, task_id, project_id = map(lambda x: None if x == 'None' else int(x), id.split(':'))
+        job_id, task_id, project_id = parse_complex_id(id)
         if wt.total_seconds():
             value = wt // WORKING_TIME_RESOLUTION
             event = EventSerializer(data={
@@ -633,22 +643,16 @@ def handle_client_events_push(request, data):
 
             return event
 
-    def get_combined_id(event):
-        job_id = event.get("job_id")
-        task_id = event.get("task_id")
-        project_id = event.get("project_id")
-        return ':'.join((str(job_id), str(task_id), str(project_id)))
-
     if previous_event := data["previous_event"]:
         previous_end_timestamp = get_end_timestamp(previous_event)
-        previous_id = get_combined_id(previous_event)
+        previous_id = build_complex_id(previous_event)
     elif data["events"]:
         previous_end_timestamp = data["events"][0]["timestamp"]
-        previous_id = get_combined_id(data["events"][0])
+        previous_id = build_complex_id(data["events"][0])
 
     working_time_per_id = {}
     for event in data["events"]:
-        id = get_combined_id(event)
+        id = build_complex_id(event)
         working_time = datetime.timedelta()
 
         timestamp = event["timestamp"]
--- a/cvat/apps/events/handlers.py
+++ b/cvat/apps/events/handlers.py
@@ -655,14 +655,16 @@ def handle_client_events_push(request, data: dict):
             previous_end_timestamp = end_timestamp
 
         if previous_ids not in working_time_per_ids:
-            working_time_per_ids[previous_ids] = datetime.timedelta()
+            working_time_per_ids[previous_ids] = {
+                "value": datetime.timedelta(),
+                "timestamp": timestamp,
+            }
 
-        working_time_per_ids[previous_ids] += working_time
+        working_time_per_ids[previous_ids]["value"] += working_time
         previous_ids = read_ids(event)
 
     if data["events"]:
         common = {
-            "timestamp": data["events"][0]["timestamp"],
             "user_id": request.user.id,
             "user_name": request.user.username,
             "user_email": request.user.email or None,
@@ -670,8 +672,9 @@ def handle_client_events_push(request, data: dict):
             "org_slug": getattr(org, "slug", None),
         }
 
-        for ids in working_time_per_ids:
-            event = generate_wt_event(ids, working_time_per_ids[ids], common)
+        for ids, working_time in working_time_per_ids.items():
+            common["timestamp"] = working_time["timestamp"]
+            event = generate_wt_event(ids, working_time["value"], common)
             if event:
                 event.is_valid(raise_exception=True)
                 record_server_event(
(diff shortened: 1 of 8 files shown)


cvat/apps/events/views.py
```diff
@@ -1,23 +1,41 @@
...
-from cvat.apps.engine.log import vlogger
+from cvat.apps.iam.filters import ORGANIZATION_OPEN_API_PARAMETERS
+
 from .export import export
+from .handlers import handle_client_events_push
+
 
 class EventsViewSet(viewsets.ViewSet):
     serializer_class = None
+    SERVER_ONLY_SCOPES = frozenset((
```
SpecLad, line 25: You should use a whitelist rather than a blacklist. A blacklist can quickly get out of date, because nothing will break if a new server-side event is added without updating the blacklist.

cvat/apps/events/views.py
```diff
@@ -31,11 +49,20 @@ def create(self, request):
         serializer = ClientEventsSerializer(data=request.data, context={"request": request})
         serializer.is_valid(raise_exception=True)
 
-        for event in serializer.data["events"]:
-            message = JSONRenderer().render(event).decode('UTF-8')
+        for event in serializer.validated_data['events']:
+            scope = event['scope']
+            if scope in EventsViewSet.SERVER_ONLY_SCOPES:
+                raise ValidationError(f'Event scope **{scope}** is not allowed from client')
```
SpecLad, line 55: This seems like the kind of thing that should be in `ClientEventsSerializer`.

cvat/apps/events/serializers.py
```diff
@@ -52,42 +43,26 @@ def to_internal_value(self, data):
...
         time_correction = receive_time - send_time
 
-        if previous_event := data["previous_event"]:
-            previous_end_timestamp = self._end_timestamp(previous_event)
-        elif data["events"]:
-            previous_end_timestamp = data["events"][0]["timestamp"]
+        if data["previous_event"]:
+            data["previous_event"].update({
+                "timestamp": data["previous_event"]["timestamp"] + time_correction,
+            })
```
SpecLad, line 49:
```suggestion
            data["previous_event"]["timestamp"] += time_correction
```

site/content/en/docs/administration/advanced/analytics.md
```diff
@@ -154,6 +155,38 @@ Client events:
...
 
+### Working time calculation
+
+Here is a short overview of how CVAT deals with the user's working time:
+
+- The user interface collects events when a user interacts with the interface
+(resizing canvas, drawing objects, clicking buttons, etc)
+The structure of one single event is described [here](#events-log-structure).
+
+- The user interface sends these events in bulks to the server.
```
SpecLad, line 166:
```suggestion
- The user interface sends these events in bulk to the server.
```

site/content/en/docs/administration/advanced/analytics.md
```diff
@@ -154,6 +155,38 @@ Client events:
...
+    - Regardless different events may have `duration` field, we consider only `change:frame`
+    event as a continuous event when computing working time
+    - So, the only `duration` of the `change:frame` event is added to the working time.
+    Thresholding is not involved here.
+
+- After calculation, the server generates `send:working_time` events with time value in payload.
+These events may be, or not be bounded to a certain job/task/project.
+CVAT defines belonging of the working time based on events, that were used to calculate.
+
+- If Clickhouse was configured, CVAT saves the event in
```
SpecLad, line 187: Is it possible for Clickhouse to not be configured? I don't think it's optional.

site/content/en/docs/administration/advanced/analytics.md
```diff
@@ -154,6 +155,38 @@ Client events:
 
 <!--lint enable maximum-line-length-->
 
+### Working time calculation
+
+Here is a short overview of how CVAT deals with the user's working time:
+
+- The user interface collects events when a user interacts with the interface
+(resizing canvas, drawing objects, clicking buttons, etc)
```
SpecLad, line 163: Please indent subsequent lines in each list item.

site/content/en/docs/administration/advanced/analytics.md
```diff
@@ -154,6 +155,38 @@ Client events:
...
+
+- When events reach the server, it calculates working time based on timestamps of the events.
+
+- Working time is computed as a sum of time differences between
+the end of the previous event and the end of the next event.
+    - But if the difference is more then threshold (~100 seconds), the difference is not added to the working time
+    - Regardless different events may have `duration` field, we consider only `change:frame`
+    event as a continuous event when computing working time
+    - So, the only `duration` of the `change:frame` event is added to the working time.
+    Thresholding is not involved here.
```
SpecLad, line 181: I find this layout confusing, because there are two components to the working time, but one of them is explained in the parent item, and one in the sub-item. I would suggest refactoring this to something like:

- The working time for an event is computed as the sum of the following:
  - The difference between the start time of the event and the end time of the previous event, if it is not more than 100 seconds.
  - The duration of the event, for events of type `change:frame`.

site/content/en/docs/administration/advanced/analytics.md
```diff
@@ -154,6 +155,38 @@ Client events:
...
+
+- After calculation, the server generates `send:working_time` events with time value in payload.
+These events may be, or not be bounded to a certain job/task/project.
+CVAT defines belonging of the working time based on events, that were used to calculate.
```
SpecLad, line 185:
```suggestion
These events may or may not be bound to a certain job/task/project,
depending on the client-side events that were used to generate them.
```

SpecLad (changes requested):

cvat/apps/events/handlers.py
```diff
@@ -141,7 +135,7 @@ def user_name(instance=None):
 
 def user_email(instance=None):
     current_user = get_user(instance)
-    return _get_value(current_user, "email")
+    return _get_value(current_user, "email") or None
```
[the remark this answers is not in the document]
bsekachev: Email may be empty string in our database.
EventSerializer does not accept blank field (empty string). Only null or nothing. 
It will throw validation error. 

Usually we do not run event.is_valid() for events generated on the server, but I did it for working time event and this error occurred

But yes, this change does not affect working time logic

cvat/apps/events/handlers.py
```diff
@@ -594,3 +588,96 @@ def handle_viewset_exception(exc, context):
...
+
+    def generate_wt_event(
+        ids: tuple[int | None, int | None, int | None],
+        wt: datetime.timedelta,
+        common: dict
+    ) -> EventSerializer | None:
+        job_id, task_id, project_id = ids
+        if wt.total_seconds():
+            value = wt // WORKING_TIME_RESOLUTION
+            event = EventSerializer(data={
```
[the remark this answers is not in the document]
bsekachev: Only for extra validation
During development this validation helped me, and I believe it is a good idea to validate any data we are going to save in the database

cvat/apps/events/handlers.py
```diff
@@ -594,3 +588,96 @@ def handle_viewset_exception(exc, context):
...
+
+        if previous_ids not in working_time_dict:
+            working_time_dict[previous_ids] = datetime.timedelta()
+
+        working_time_dict[previous_ids] += working_time
+        previous_ids = read_ids(event)
+
+    if data["events"]:
+        common = {
+            "timestamp": data["events"][0]["timestamp"],
```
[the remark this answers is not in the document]
bsekachev: I did it first, but there is one problem with it. In this case event will get timestamp later, then `Job.updated_date` field and when analytics report requested, working time events with timestamp > `Job.updated_date` will be ignored.

cvat/apps/events/views.py
bsekachev: Agree

site/content/en/docs/administration/advanced/analytics.md
bsekachev: CVAT will work without grafana, clickhouse and vector. So, I think it is optional

cvat/apps/events/handlers.py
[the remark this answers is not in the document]
bsekachev: I think we can remember time each time when see new triple (job_id, task_id, project_id) and use it.

site/content/en/docs/administration/advanced/analytics.md
bsekachev: Let's omit this part, I do not think it is really important here

site/content/en/docs/administration/advanced/analytics.md
bsekachev: I am not sure what you meant here, tried couple of ideas, but remark lint was not happy :/

SpecLad: I meant:

```suggestion
  (resizing canvas, drawing objects, clicking buttons, etc)
```

cvat/apps/events/handlers.py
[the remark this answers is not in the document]
SpecLad: Okay (although IMO, email being an empty string in the DB is a bug).

cvat/apps/events/handlers.py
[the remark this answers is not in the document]
SpecLad: I'm not sure I agree. Validation is for transforming data from untrusted (i.e. coming from the client) to trusted. If the data is coming from the server, like here, then it is already trusted. Furthermore, using `EventSerializer` specifically is problematic for a few reasons:

* `Serializer.is_valid` raises a `ValidationError`, which is transformed by DRF to `400 Bad Request`, i.e. a client-side problem. However, if the data here is invalid, then the problem is on the server.

* `EventSerializer` is meant to validate events coming from the client, and we may not necessarily want the same requirements for events originating from the server.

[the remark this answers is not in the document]
SpecLad: I'd also add that if you want to do any sanity checking, then the right place to do it would be in `record_server_event`, so that the checks are shared between all events.

cvat/apps/events/handlers.py
```diff
@@ -594,3 +588,99 @@ def handle_viewset_exception(exc, context):
...
+
+        end_timestamp = get_end_timestamp(event)
+        if end_timestamp > previous_end_timestamp:
+            working_time += end_timestamp - previous_end_timestamp
+            previous_end_timestamp = end_timestamp
+
+        if previous_ids not in working_time_per_ids:
+            working_time_per_ids[previous_ids] = {
+                "value": datetime.timedelta(),
+                "timestamp": timestamp,
```
SpecLad, line 660: I think it would be more accurate to initialize `timestamp` to `datetime()` here and for each event do:

```python
if previous_end_timestamp > working_time_per_ids[previous_ids]["timestamp"]:
    working_time_per_ids[previous_ids]["timestamp"] = previous_end_timestamp
```

That way, the timestamp of the working time event is the end timestamp of the last event that was used to determine it - which is the earliest time at which the working time would be known.

cvat/apps/events/views.py
```diff
@@ -31,11 +34,15 @@ def create(self, request):
         serializer = ClientEventsSerializer(data=request.data, context={"request": request})
         serializer.is_valid(raise_exception=True)
 
-        for event in serializer.data["events"]:
-            message = JSONRenderer().render(event).decode('UTF-8')
+        handle_client_events_push(request, serializer.validated_data)
```
SpecLad, line 37: I think this should be after the event loop, so that the working time events are logged after the events that were used to generate them.

bsekachev: I do not think that is so important
Timestamp of generated events == timestamp of the first event in sequence used to generate

cvat/apps/events/handlers.py
bsekachev: I do not think we should use the last event.
There is a problem with timestamp correction (request time added to the event timestamp).

As discussed with @azhavoro , we need this correction, because it allows to sync time between client and server.
On the other way if request took long (e.g. 30 seconds), we will get the same problem with event timestamp >= Job.updated_date. It looks like a problem.

But using the first event in the sequence allows to minimize effect of this bug

cvat/apps/events/handlers.py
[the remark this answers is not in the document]
bsekachev: Alright, makes sence

cvat/apps/events/handlers.py
SpecLad: Okay. Also, I see that you changed the contents of the event to record the working time in the duration field. I like that, now the timestamp of the first original event makes sense as the timestamp of the working time event, since we can now view the WT event as spanning from the start of the first original event to the end of the last one.

SpecLad (approved):

PR #7974, opened 2024-05-31
Creating Consensus Job

Viditagarwal7479:
Fixes Creating `consensus jobs` for `normal jobs` in https://github.com/cvat-ai/cvat/issues/7973

[List of features present in this PR](https://github.com/cvat-ai/cvat/pull/7974#issuecomment-2156201401)

### Checklist

### License

## Summary by CodeRabbit

- **New Features**
  - Introduced a new `consensus job per segment` configuration option for tasks.
  - Added the ability to specify a `source job ID` for jobs.

- **Enhancements**
  - Advanced configuration form now includes an option to set `consensus job per segment`.
  - Improved task and job creation workflows to support consensus jobs.

- **Bug Fixes**
  - Fixed issues related to job and task serialization and deserialization to include new fields.

- **Migration**
  - Multiple database migrations to add and alter fields for `consensus job per segment` and `source job ID`.


Viditagarwal7479:
In the above commit itself, I have added the constraint on the value of consensus jobs, [0, 10] - {1}.

Viditagarwal7479:
## List of features added in this PR

1. Separate collapsable menu for setting `consensus_job_per_segement` and `agreement_score_threshold` while creating the task
image
image

2. Task having consensus jobs have a additional tag in their task name as "Consensus Based Annotation"
3. `agreement_score_threshold` parameter can be changed after task is created
4. Consensus jobs corresponding are in a collapsed view under their parent job, with the number of consensus jobs mentioned
image
image

5. `api/jobs?parent_job_id={id}` added additional filter option to filter out consensus jobs in corresponding to the `parent_job_id`
6. Added frontend side and server side validation for these newly added parameters
7. Annotations made in consensus jobs aren't included in the annotation export of the task and the task backup. Though they can be exported individually.
8. These newly added paramterers are shown appropriately in `GET` request when for task and job details, `PATCH` request when creating and updating the task, while updating task only `agreement_score_threshold` can be adjusted among the newly added parameters

Viditagarwal7479:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=cvat-ai_cvat&pullRequest=7974) **Quality Gate failed**
> 
> Failed conditions  [3.4% Duplication on New Code](https://sonarcloud.io/component_measures?id=cvat-ai_cvat&pullRequest=7974&metric=new_duplicated_lines_density&view=list) (required ≤ 3%)
> 
> [See analysis details on SonarCloud](https://sonarcloud.io/dashboard?id=cvat-ai_cvat&pullRequest=7974)



On going through its detailed analysis, I also don't seem to find any redundant code block that can be removed.

cvat/apps/engine/migrations/0079_job_source_job_id_task_consensus_job_per_segment.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 4.2.11 on 2024-05-30 12:55
```
zhiltsov-max, line 1: Could you squash the migrations into just one in the PR?

cvat-sdk/cvat_sdk/core/proxies/tasks.py
```diff
@@ -78,7 +78,7 @@ def upload_data(
         """
         params = params or {}
 
-        data = {"image_quality": 70}
+        data = {"image_quality": 70, "consensus_job_per_segment": 0, "agreement_score_threshold": 0}
```
zhiltsov-max, line 81: These new parameters should have server defaults, they should not be required.

cvat/apps/dataset_manager/bindings.py
```diff
@@ -779,6 +780,8 @@ def meta_for_task(db_task, host, label_mapping=None):
             ("start_frame", str(db_task.data.start_frame)),
             ("stop_frame", str(db_task.data.stop_frame)),
             ("frame_filter", db_task.data.frame_filter),
+            ("consensus_job_per_segment", str(db_task.consensus_job_per_segment)),
+            ("agreement_score_threshold", str(db_task.agreement_score_threshold)),
```
zhiltsov-max, line 784: What is the point in adding this information into export?

cvat/apps/dataset_manager/task.py
```diff
@@ -742,6 +743,8 @@ def _patch_data(self, data, action):
         splitted_data = {}
         jobs = {}
         for db_job in self.db_jobs:
+            if db_job.parent_job_id is not None:
```
zhiltsov-max, line 746: It will be better to put this into the query filter.

cvat/apps/engine/serializers.py
```diff
@@ -1096,6 +1097,8 @@ class TaskReadSerializer(serializers.ModelSerializer):
     source_storage = StorageSerializer(required=False, allow_null=True)
     jobs = JobsSummarySerializer(url_filter_key='task_id', source='segment_set')
     labels = LabelsSummarySerializer(source='*')
+    consensus_job_per_segment = serializers.ReadOnlyField(required=False)
```
zhiltsov-max, line 1100:
```suggestion
    consensus_jobs_per_segment = serializers.ReadOnlyField(required=False)
```

cvat/apps/engine/models.py
```diff
@@ -677,6 +679,7 @@ class Job(TimestampedModel):
 
     type = models.CharField(max_length=32, choices=JobType.choices(),
```
zhiltsov-max, line 680: Please add a new job type in the enum.

cvat/apps/engine/models.py
```diff
@@ -418,6 +418,8 @@ class Task(TimestampedModel):
         blank=True, on_delete=models.SET_NULL, related_name='+')
     target_storage = models.ForeignKey('Storage', null=True, default=None,
         blank=True, on_delete=models.SET_NULL, related_name='+')
+    consensus_job_per_segment = models.IntegerField(default=0, blank=True)
+    agreement_score_threshold = models.FloatField(default=0, blank=True)
```
zhiltsov-max, line 422: Consider extracting this field into a separate model. If we add more parameters in future, it will be easier if it's stored in a dedicated structure.

zhiltsov-max (changes requested):

cvat/apps/dataset_manager/bindings.py
Viditagarwal7479: My bad; I missed making modifications here while moving from task data to task.

cvat/apps/engine/models.py
Viditagarwal7479: But then we would also have to check that that there isn't any logic based on job type due to which the consensus job doesn't work properly.

Like some functionalities which it exhibited since it was inheriting the job type, would now not work, and break?

Viditagarwal7479: Despite that, shouldn't the job type structure look like this?

- Job Types
  - Ground Truth
  - Annotation
     - Normal Job
     - Consensus Job

And inevitably the information about whether it's a normal or consensus job is present in parameter `parent_job_id`...?

zhiltsov-max: > Despite that, shouldn't the job type structure look like this? [...]

I don't see a point in making such a nested structure at this point. The consensus jobs are not the same as annotation jobs, they have different handling logic in annotation import and export. Probably, they should not be included in the task quality report. It's better to just keep the type list flat, unless there is a clear use case for such a structure.

> But then we would also have to check that that there isn't any logic based on job type due to which the consensus job doesn't work properly.
> Like some functionalities which it exhibited since it was inheriting the job type, would now not work, and break?

You would need to check this in any case, as these jobs aren't the same as "normal" jobs.

cvat-ui/src/components/create-task-page/advanced-configuration-form.tsx
```diff
@@ -92,7 +92,7 @@ function validateURL(_: RuleObject, value: string): Promise<void> {
     return Promise.resolve();
 }
 
-const isInteger = ({ min, max }: { min?: number; max?: number }) => (
+const isInteger = ({ min, max, toBeSkipped }: { min?: number; max?: number; toBeSkipped?: number }) => (
```
zhiltsov-max, line 95: Consider renaming the `isInteger` function or adding another validator that checks for specific values.

Viditagarwal7479: With this, the functionality is added with minimal modifications, and the function gets reused in consensus-configuration. The existing function also checks for the range and whether it's an integer, as the function's name suggests...

cvat/schema.yml
```diff
@@ -1891,6 +1891,11 @@ paths:
           enum:
```
zhiltsov-max, line 1891: Probably, this file needs to be updated to reflect the latest changes.

cvat-ui/src/components/task-page/details.tsx
```diff
@@ -92,8 +94,9 @@ class DetailsComponent extends React.PureComponent<Props, State> {
     }
 
     private renderTaskName(): JSX.Element {
-        const { name } = this.state;
+        const { name, consensusJobsPerSegment } = this.state;
         const { task: taskInstance, onUpdateTask } = this.props;
+        const taskName = name + (consensusJobsPerSegment > 0 ? ' (Consensus Based Annotation)' : '');
```
zhiltsov-max, line 99: I suggest that the task name should not be modified, and an icon or a tag should be shown instead. Take for example the "Ground Truth" label on GT jobs.

cvat/apps/engine/serializers.py
```diff
@@ -1119,12 +1122,13 @@ class TaskWriteSerializer(WriteOnceMixin, serializers.ModelSerializer):
     project_id = serializers.IntegerField(required=False, allow_null=True)
     target_storage = StorageSerializer(required=False, allow_null=True)
     source_storage = StorageSerializer(required=False, allow_null=True)
+    consensus_jobs_per_segment = serializers.IntegerField(required=False, allow_null=True)
```
zhiltsov-max, line 1125: Could you rename this parameter to something without "segment" in the name? Segments exist only in the server implementation, they are not exposed anywhere in the API. I can suggest `consensus_jobs_per_normal_job`.

Viditagarwal7479: Sure, I will rename it to `consensus_jobs_per_normal_job`

cvat-ui/src/components/job-item/job-item.tsx
```diff
@@ -113,14 +114,29 @@ function JobItem(props: Props): JSX.Element {
     }
     const frameCountPercent = ((job.frameCount / (task.size || 1)) * 100).toFixed(0);
     const frameCountPercentRepresentation = frameCountPercent === '0' ? '<1' : frameCountPercent;
+    let jobName = `Job #${job.id}`;
+    if (task.consensusJobsPerSegment && job.type !== JobType.GROUND_TRUTH) {
+        jobName = job.type === JobType.CONSENSUS ? `Consensus Job #${job.id}` : `Normal Job #${job.id}`;
```
klakhov, line 119: Do we really need `normal job` instead of just `job`?

cvat-ui/src/components/job-item/job-item.tsx
```diff
@@ -113,14 +114,29 @@ function JobItem(props: Props): JSX.Element {
...
+    }
+
+    let consensusJob: Job[] = [];
```
klakhov, line 122:
```suggestion
    let consensusJobs: Job[] = [];
```

cvat-ui/src/components/job-item/job-item.tsx
```diff
@@ -113,14 +114,29 @@ function JobItem(props: Props): JSX.Element {
...
+    if (task.consensusJobsPerSegment) {
+        consensusJob = task.jobs.filter((eachJob: Job) => eachJob.parentJobId === id).reverse();
+    }
+    const consensusJobView: React.JSX.Element[] = consensusJob.map((eachJob: Job) => (
```
klakhov, line 126:
```suggestion
    const consensusJobViews: React.JSX.Element[] = consensusJob.map((eachJob: Job) => (
```

cvat-ui/src/components/task-page/job-list.tsx
```diff
@@ -53,6 +53,9 @@ function setUpJobsList(jobs: Job[], query: JobsQuery): Job[] {
         result = result.filter((job, index) => jsonLogic.apply(filter, converted[index]));
     }
 
+    // primarily only normal jobs should be shown
```
klakhov, line 56: If we use `parentJobId` parameter, that means not only normal jobs are shown, but GT jobs are also shown

klakhov (commented):
- When im trying to create a task with 1 segment consenus jobs are created as expected. But when Im trying to create a task with any non-default segment size, I dont see any consensus jobs
- There should be more padding in between consensus tab and general job information. Also actions menu should be higher. (marked with arrows)
image
- Actions menu runs away when i open the tab :smile: :
image
- The consensus tag moves to strange place in case of large task name. I would suggest not to mix it with name. Maybe we can adjust the highlighted block to use consensus tag? :
image

cvat-ui/src/actions/tasks-actions.ts
```diff
@@ -213,6 +213,7 @@ ThunkAction<Promise<void>, {}, {}, AnyAction> {
             sorting_method: data.advanced.sortingMethod,
             source_storage: new Storage(data.advanced.sourceStorage || { location: StorageLocation.LOCAL }).toJSON(),
             target_storage: new Storage(data.advanced.targetStorage || { location: StorageLocation.LOCAL }).toJSON(),
+            consensus_jobs_per_segment: 0,
```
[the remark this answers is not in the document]
Viditagarwal7479: Ah, correct changed it to `data.consensus_jobs_per_normal_job` to also make it coherent with the rest.

cvat-ui/src/components/task-page/job-list.tsx
Viditagarwal7479: Sorry for not writing the comment properly. Primarily only normal jobs and GT jobs would be shown as the consensus job will be collapsed inside the corresponding normal job.

cvat-ui/src/components/job-item/job-item.tsx
Viditagarwal7479: In the regular sync meet, when talking about jobs, to not get confused whether talking about consensus ones or the normal, we used to the "normal job", which swept in here as well. 

If you say I will change "Normal Job" to just "Job".

cvat-ui/src/components/create-task-page/create-task-content.tsx
```diff
@@ -965,8 +997,8 @@ class CreateTaskContent extends React.PureComponent<Props & RouteComponentProps,
                 {this.renderSubsetBlock()}
                 {this.renderLabelsBlock()}
                 {this.renderFilesBlock()}
+                {this.renderConsensusBlock()}
```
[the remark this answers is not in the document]
Viditagarwal7479: To not mix the parameters related to consensus jobs with the advanced parameters.


Earlier, in this PR itself, I had created a new block for these types of parameters (🥲) when we had to configure two such parameters. When this changed to a single parameter `consensus jobs per normal job`, we thought a whole collapse block would be a waste for just a single parameter, so I removed the collapse block and just placed an input field.

Viditagarwal7479:
> When im trying to create a task with 1 segment consenus jobs are created as expected. But when Im trying to create a task with any non-default segment size, I dont see any consensus jobs

Thanks for pointing out it was happening due to conflicting `onSubmit` property on both `advancedBlock` and `consensusBlock`. Changed it to `onChange` for `consensusBlock` just like `basicBlock`

cvat-ui/src/components/create-task-page/consensus-configuration-form.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+const initialValues: ConsensusConfiguration = {
+    consensusJobsPerSegment: 0,
+};
+
+interface Props {
+    onSubmit(values: ConsensusConfiguration): Promise<void>;
+}
+
+const isNumber = ({
```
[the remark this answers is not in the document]
Viditagarwal7479: Thanks. Now I have updated it accordingly and moved it under `utils` as `validate-integer.ts`

cvat-ui/src/components/header/header.tsx
```diff
@@ -412,11 +412,11 @@ function HeaderComponent(props: Props): JSX.Element {
         .map(({ component, weight }): typeof menuItems[0] => [component({ targetProps: props }), weight]),
     );
 
-    const getButtonClassName = (value: string): string => {
+    const getButtonClassName = (value: string, highlightable = true): string => {
```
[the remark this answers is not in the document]
Viditagarwal7479: This PR compares my branch to another branch, which isn't updated with `develop`; that's why you see that change in this PR. It's actually a change made by Boris around 17 days ago, which came in this file while I updated this branch with `develop`

Viditagarwal7479:
> Actions menu runs away when i open the tab 😄 

Does this look better?

image

image

cvat-ui/src/components/job-item/job-item.tsx
klakhov: Ok, lets leave it as is for now, its discussable I believe

Viditagarwal7479:
> The consensus tag moves to strange place in case of large task name. I would suggest not to mix it with name. Maybe we can adjust the highlighted block to use consensus tag?

image

On zooming out by 50%:
image

cvat-ui/src/components/create-task-page/create-task-content.tsx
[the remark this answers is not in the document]
Viditagarwal7479: As you have mentioned here and also Boris pointed out in the mid-term meet. I have moved the `consensus jobs per normal job` parameter into the `advance configuration` collapse block.

cvat-ui/src/components/create-task-page/create-task-content.tsx
```diff
@@ -152,13 +153,15 @@ function filterFiles(remoteFiles: RemoteFile[], many: boolean): RemoteFile[] {
 class CreateTaskContent extends React.PureComponent<Props & RouteComponentProps, State> {
     private basicConfigurationComponent: RefObject<BasicConfigurationForm>;
     private advancedConfigurationComponent: RefObject<AdvancedConfigurationForm>;
+    // private consensusConfigurationComponent: RefObject<ConsensusConfigurationForm>;
```
zhiltsov-max, line 156: Please clean up comments

zhiltsov-max (approved):

PR #8007, opened 2024-06-09
[GSoC'24] Registeration of Shortcuts via modules and dummy conflict detector

tahamukhtar20:
Part of: #747 

### Motivation and context

This PR is part of the GSOC project: Keyboard Shortcut Customization

### How has this been tested?

It has been tested by checking if the existing shortcuts work the same way as before.

### Checklist

### License


bsekachev:
Hi @tahamukhtar20 

There are plenty failed end to end tests with this pull request
Could you please take a look?

bsekachev:
Example: 
New organization pipeline -- The first user creates an organization and activates it -- after all hook (failed)

cvat-ui/src/utils/mousetrap-react.tsx
```diff
@@ -34,17 +34,21 @@ export default function GlobalHotKeys(props: Props): JSX.Element {
     const { children, keyMap, handlers } = props;
     useEffect(() => {
         for (const key of Object.keys(keyMap)) {
-            const { sequences } = keyMap[key];
-            const handler = handlers[key];
-            Mousetrap.bind(sequences, handler, 'keydown');
-            applicationKeyMap[key] = keyMap[key];
+            if (keyMap[key] && keyMap[key].sequences) {
```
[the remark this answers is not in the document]
bsekachev: For me it is like (e.g. in MergeControl component) you have:

```
<GlobalHotKeys
                keyMap={{ SWITCH_MERGE_MODE: shortcuts.SWITCH_MERGE_MODE.details }}
                ....
            />
``` 

Where in keyMap you declare key SWITCH_MERGE_MODE, but value maybe undefined if shortcut registration is not completed yet. 

Then in Moustrap handler you are trying to perform destructurization:
```
    const { sequences, action } = keyMap[key];
```

What do you think?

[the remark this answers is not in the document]
tahamukhtar20: it shouldn't be, but this is what the error showed. It's not happening locally. I'll try to investigate it.

cvat-ui/src/actions/shortcuts-actions.ts
```diff
@@ -3,16 +3,21 @@
 //
 // SPDX-License-Identifier: MIT
 
+import { KeyMapItem } from 'utils/mousetrap-react';
 import { ActionUnion, createAction } from 'utils/redux';
 
 export enum ShortcutsActionsTypes {
     SWITCH_SHORTCUT_DIALOG = 'SWITCH_SHORTCUT_DIALOG',
+    REGISTER_SHORTCUT = 'REGISTER_SHORTCUT',
```
bsekachev, line 11: As this action allows to register a list of shortcuts: 

```suggestion
    REGISTER_SHORTCUTS = 'REGISTER_SHORTCUTS',
```

cvat-ui/src/utils/enums.tsx
```diff
@@ -21,3 +21,7 @@ export enum StorageStatuses {
     FORBIDDEN = 'FORBIDDEN',
     NOT_FOUND = 'NOT_FOUND',
 }
+
+export enum ViewType {
```
bsekachev, line 25: Let's rename:

```suggestion
export enum ShortcutScope {
```

cvat-ui/src/utils/mousetrap-react.tsx
```diff
@@ -11,7 +11,7 @@ export interface KeyMapItem {
     description: string;
     sequences: string[];
     displayedSequences?: string[];
-    action: 'keydown' | 'keyup' | 'keypress';
+    view: string;
```
bsekachev, line 14:
```suggestion
    scope: ShortcutScope;
```

cvat-ui/src/utils/hooks.ts
```diff
@@ -130,3 +133,9 @@ export function useAuthQuery(): Record<string, string> | null {
     const queryParams = new URLSearchParams(history.location.search);
     return authQuery(queryParams);
 }
+
+export function useRegisterShortcuts(shortcuts: Record<string, KeyMapItem>): void {
+    const store = getCVATStore();
+    const { registerShortcuts } = shortcutsActions;
+    store.dispatch(registerShortcuts(shortcuts));
+}
```
bsekachev, line 141: That is not a hook, so:

- let's remove `use` in the beginning: `useRegisterShortcuts` -> `registerComponentShortcuts`
- let's move this function to `shortcuts-actions.ts`

cvat-ui/src/components/annotation-page/attribute-annotation-workspace/attribute-annotation-sidebar/attribute-annotation-sidebar.tsx
```diff
@@ -50,6 +52,65 @@ interface LabelAttrMap {
     [index: number]: any;
 }
 
+const componentShortcuts = {
```
bsekachev, line 55: Now in each component we have:

1. const componentShortcuts = { SHORTCUT_1: { ... }, SHORTCUT_2: { ... } }
2. const subKeyMap = { SHORTCUT_1: { ... }, SHORTCUT_2: { ... } } (actually the same as componentShortcuts)
3. const handlers = { SHORTCUT_1: () => {}, SHORTCUT_2: () => {} }

I will propose to generate subKeyMap dinamically, e.g.:
```
                <GlobalHotKeys
                    keyMap={Object.fromEntries(Object.keys(componentShortcuts).map((key) => [key, keyMap[key]]))}
                    handlers={handlers}
                />
```

And add types to `handlers`, e.g.:
```
const handlers: Record<keyof typeof componentShortcuts, (event?: KeyboardEvent) => void> = {
```

It will make as sure that we do not forget to declare any handlers. 

What do you think?

bsekachev: Also, in this case we may prevent undefined values passed to <GlobalHotKeys> as discussed before:

```
Object.fromEntries(Object.keys(componentShortcuts).filter((key) => !!keyMap[key]).map((key) => [key, keyMap[key]]))
```

This code may be separated as an utility function.

cvat-ui/src/utils/mousetrap-react.tsx
```diff
@@ -34,17 +35,21 @@ export default function GlobalHotKeys(props: Props): JSX.Element {
     const { children, keyMap, handlers } = props;
     useEffect(() => {
         for (const key of Object.keys(keyMap)) {
-            const { sequences, action } = keyMap[key];
-            const handler = handlers[key];
-            Mousetrap.bind(sequences, handler, action);
-            applicationKeyMap[key] = keyMap[key];
+            if (keyMap[key] && keyMap[key].sequences) {
```
bsekachev, line 38: It seems we do not need this condition anymore, as `subKeyMap` will ignore shortcuts with `undefined` values, right? 

The same in return.

bsekachev:
Hello, 

Not all the shortcuts are transformed with this patch (because not all were initially defined in shortcuts reducer).
Try to search in project by `GlobalHotKeys` to understand what files include declared keymaps.

For example in file `cvat-ui/src/components/annotation-page/attribute-annotation-workspace/attribute-annotation-sidebar/attribute-editor.tsx` you may find the following:

```
        sortedValues.forEach((value: string, index: number): void => {
            const key = `SET_${index}_VALUE`;
            keyMap[key] = {
                name: `Set value "${value}"`,
                description: `Change current value for the attribute to "${value}"`,
                sequences: [`${index}`],
                action: 'keydown',
            };

            handlers[key] = (event: KeyboardEvent | undefined) => {
                if (event) {
                    event.preventDefault();
                }

                onChange(value);
            };
        });
```

**On one hand**, if we want to avoid shortcuts conflicts, using common design, now we have to declare these shortcuts globally as well (register in storage) (probably with specific scope (e.g. `attribute annotation mode`)); 

**On the other hand**, I do not think we want to be able to modify such kind of shortcuts in the future (and even list them in total shortcuts list). So, maybe we should introduce `isMutable`, `isEnumerable` flags to keyMapItem or something like this, and handle them properly. The reason of such restriction is that these shortcuts are very dynamic (for different labels and attributes they are different, with own `name`, and `description`). 

Alternatively, we may unify these shortcuts, like: 
```
name: "Set 1st value to the current attribute"
description: "Change current value for the attribute to the 1st value in the list"
```

and only provide correct handlers in attribute annotation sidebar component. 

**From first look, I like the second option more.** 


As I remember, there is similar problem in tag annotation mode (this file: cvat-ui/src/components/annotation-page/tag-annotation-workspace/tag-annotation-sidebar/shortcuts-select.tsx).

In this component, I believe it make sense to be able to see and configure these shortcuts (again, with specific scope, e.g. `tag annotation mode`)

cvat-ui/src/components/annotation-page/attribute-annotation-workspace/attribute-annotation-sidebar/attribute-editor.tsx
```diff
@@ -24,6 +28,71 @@ interface InputElementParameters {
     onChange(value: string): void;
 }
 
+const componentShortcuts = {
+    SET_0_VALUE: {
+        name: 'Set 0th value to the current attribute',
```
bsekachev, line 33: What is 0th value?

Order is incorrect in this list.

0 -> first value
1 -> second value, etc.

cvat-ui/src/components/annotation-page/tag-annotation-workspace/tag-annotation-sidebar/shortcuts-select.tsx
```diff
@@ -21,6 +26,81 @@ type Props = {
     labels: Label[];
 };
 
+const componentShortcuts = {
+    SETUP_0_TAG: {
+        name: 'Setup 0 tag',
+        description: 'Setup tag with "label" label',
```
bsekachev, line 32: "label" label?

bsekachev: I suppose description for any value is the same. Something like:

`Create a new tag with corresponding class. The class may be setup in tag annotation sidebar`

Name, also the same for each shortcut:
e.g.: `Create a new tag`

bsekachev:
LGTM, let's wait for CI.

We are going to make a new release soon, after that this PR will be merged.
But we still need to apply the same approach for attribute annotation mode as you did in tag annotation mode

tahamukhtar20:
Yes sure, I will make those changes in the next pull request, the UI one.

GeorgePearse:
I think this maybe removed the ability to change the "active class" with hotkeys on a per task basis? 

Without being able to change the class names that these are applied to, these hotkeys are really not useful

image

PR #8018, opened 2024-06-12
feat: Add a new `published` job stage

likelikeslike:
We added a new stage for `Job` in CVAT called `published`, which allows users to make one `Job` accessible to all members of their organization.

### Motivation and context

First, thank you for the amazing project that has significantly helped us with our annotation program.

In our work, we encountered a problem: we want to provide annotation examples for new annotators to help them understand the annotation standards easily. However, with the current CVAT design, the only way to do this is by duplicating several `Tasks` and assigning each one to an annotator. While this method achieves our goal, it is not that convenient, and there is a risk that the annotators might accidentally change the pre-annotated examples.

Our ideal solution is that each annotator can share a single `Job` where they can only add comments by opening an `Issue`.

To achieve this, we added a new stage for `Job`, called `published`. If we change a `Job` to the `published` stage, any annotator with the link to this `Job` can view it and add comments.

### How has this been tested?

Yes, we followed the [Development environment](https://docs.cvat.ai/docs/contributing/development-environment/) guidelines and tested it on Ubuntu 22.04. The results were as we expected.

However, due to our limited knowledge of this project, we made the changes as minor as possible. We may also need your help to make this PR more useful in the future.

### Checklist

- [ x ] I submit my changes into the `develop` branch
~~- [ ] I have created a changelog fragment ~~
- [ x ] I have updated the documentation accordingly

### License


likelikeslike:
Show the published jobs count in the `Tasks` page.

image

likelikeslike:
Hi @bsekachev, could you please take a moment to review this PR and let me know if it can be merged? Thank you!

MChanlong:
how can i use your improved version, i also want to create a task multi-users can work in the same job and the database can store different annotations from different users

likelikeslike:
> how can i use your improved version, i also want to create a task multi-users can work in the same job and the database can store different annotations from different users

I think the easiest way is to create a PR and merge this branch to your forked one, it seems that the CVAT team don't have a plan to merge this PR for now.

And for your case, I think you may need to make some changes in `cvat/apps/engine/rules/*.rego` to allow editing of annotations.

bsekachev:
Hi guys

Thank you for the contribution, however we are not ready to accept it as we do not have real requests for something like this.

- For access by many org members there should be implemented something like multiple assignment feature. 
- For automatic work distribution there should be some kind of automation setup, configuring `assignee` field.

> provide annotation examples for new annotators

Basically this is responsibility of existing annotations guides. If from your opinion, examples from annotation guides are not enough, please share your thoughts in the issue.

Issue #7973, opened 2024-05-31
[GSoC'24] Quality Control: Consensus


Viditagarwal7479:
### Actions before raising this issue

- [X] I searched the existing issues and did not find anything similar.
- [X] I read/searched [the docs](https://docs.cvat.ai/docs/)

### Is your feature request related to a problem? Please describe.

Allowing annotating same data multiple times and merge them. https://github.com/cvat-ai/cvat/wiki/GSoC-2024#idea-quality-control-consensus.

### Describe the solution you'd like

1. Creating multiple copies of same `job`
2. Aggregating their annotations
3. Merging them into the `source job`
4. Generating report for the same
5. Reviewing the merging process?

PR #8042, opened 2024-06-18
[GSoC'24] Merge consensus jobs

Viditagarwal7479:
Fixes `Aggregating their annotations` and  `Merging them into the parent job` in #7973 

### Checklist

### License


cvat/apps/engine/views.py
```diff
@@ -902,6 +903,22 @@ def export_backup(self, request, pk=None):
...
 
+    @extend_schema(summary="Aggregate data of a task",
+        responses={
+            '201': OpenApiResponse(description='Consensus Jobs Aggregated'),
+            '202': OpenApiResponse(description='Agreegation of Consensus Jobs started'),
+            '400': OpenApiResponse(description='Agreegating a task without data is not allowed'),
+        },
+    )
+    @action(methods=['PUT'], detail=True, url_path=r'aggregate/?$')
+    def aggregate(self, request, pk=None):
```
zhiltsov-max, line 914: Let's move this endpoint into the `/api/consensus/` API section. It will be easier to maintain and extend it there in future. Also, please change the method to POST and document the `rq_id` parameter.

zhiltsov-max (commented):
Please update the PR with the latest changes, it seems outdated.

cvat/apps/consensus/merge_consensus_jobs.py
```diff
@@ -0,0 +1,102 @@
+from typing import List, Dict
```
zhiltsov-max, line 1:
```suggestion
# Copyright (C) 2024 CVAT.ai Corporation
#
# SPDX-License-Identifier: MIT

from typing import List, Dict
```

cvat/apps/consensus/__init__.py
zhiltsov-max, line 1: Please add this component into `dev/format_python_code.py`, and then run the script.

cvat/apps/consensus/merge_consensus_jobs.py
```diff
@@ -0,0 +1,102 @@
...
+from cvat.apps.dataset_manager.bindings import import_dm_annotations
+from rest_framework import status
+from rest_framework.response import Response
+from cvat.apps.dataset_manager.task import patch_job_data, PatchAction
+from cvat.apps.engine.utils import get_rq_lock_by_user, get_rq_job_meta, define_dependent_job, process_failed_job
+from cvat.apps.engine.serializers import RqIdSerializer
+from django.utils import timezone
+from django.db import transaction
+
+def get_consensus_jobs(task_id: int):
```
zhiltsov-max, line 17:
```suggestion
def get_consensus_jobs(task_id: int) -> dict[int, list[int]]:
```

Please add other missing type annotations in functions in this file (except for some django stuff, like `request`).

cvat/apps/consensus/merge_consensus_jobs.py
```diff
@@ -0,0 +1,102 @@
...
+    jobs = {} # parent_job_id -> [consensus_job_id]
+    for job in Job.objects.select_related("segment").filter(segment__task_id=task_id, type=JobType.ANNOTATION.value).order_by('id'):
+        if job.parent_job_id is not None:
+            if job.parent_job_id not in jobs:
+                jobs[job.parent_job_id] = []
+            jobs[job.parent_job_id].append(job.id)
```
zhiltsov-max, line 23:
```suggestion
    for job in Job.objects.select_related("segment").filter(segment__task_id=task_id, type=JobType.CONSENSUS.value):
        assert job.parent_job_id
        jobs.setdefault(job.parent_job_id, []).append(job.id)
```

cvat/apps/consensus/merge_consensus_jobs.py
```diff
@@ -0,0 +1,102 @@
...
+
+    for parent_job_id, job_ids in jobs.items():
+        consensus_dataset = list(map(get_annotations, job_ids))
+
+        merged_dataset: dm.Dataset = merger(consensus_dataset)
+
+        # check if the merged dataset has annotations
+        for item in merged_dataset:
+            if not item.annotations:
+                return 400
```
zhiltsov-max, line 43: This check doesn't seem to be correct, as some images can have nothing to be annotated. I'd recommend just removing it, as it seems to be some debugging code for development. It should be covered by a unit test later.

cvat/apps/consensus/merge_consensus_jobs.py
```diff
@@ -0,0 +1,102 @@
...
+        imported annotations
+        """
+
+        parent_job = JobDataProvider(parent_job_id)
+
+        # imports the annotations in the `parent_job.job_data` instance
+        import_dm_annotations(merged_dataset, parent_job.job_data)
+
+        # updates the annotations in the job
+        patch_job_data(parent_job_id, parent_job.job_data.data.serialize(), PatchAction.UPDATE)
```
zhiltsov-max, line 60: Probably, you can replace these 2 calls (delete then update) by a single call to `put_job_data`.

cvat/apps/consensus/merge_consensus_jobs.py
```diff
@@ -0,0 +1,102 @@
...
+        return 201
```
zhiltsov-max, line 61: If you want to use HTTP status codes here, please use the named constants instead. However, I'd suggest to return None or throw exceptions on errors. The exceptions will be thrown anyway, as you're not catching them here. Unlike exceptions, return codes will make the job `finished`, instead of `failed`, so the result parsing logic is more complex.

cvat/apps/consensus/merge_consensus_jobs.py
Viditagarwal7479: Doing was adding the new merged annotations without removing the old ones. The comment block of comment after the delete operations explains this.

zhiltsov-max: The semantic of `put` supposes overwriting, so the existing annotations must be removed.

dev/format_python_code.sh
```diff
@@ -23,6 +23,7 @@ for paths in \
     "tests/python/" \
     "cvat/apps/quality_control" \
     "cvat/apps/analytics_report" \
+    "cvat/apps/consensus" \
```
zhiltsov-max, line 26: Please also copy the pyproject.toml file from any other app ([e.g. this one](https://github.com/cvat-ai/cvat/blob/develop/cvat/apps/quality_control/pyproject.toml)) to format the code correctly.

cvat/apps/consensus/merge_consensus_jobs.py
Viditagarwal7479: Should I mark this as resolved?

cvat-core/src/api-implementation.ts
```diff
@@ -31,7 +31,7 @@ import Webhook from './webhook';
 import { ArgumentError } from './exceptions';
 import {
     AnalyticsReportFilter, QualityConflictsFilter, QualityReportsFilter,
-    QualitySettingsFilter, SerializedAsset,
+    SettingsFilter, SerializedAsset,
```
klakhov, line 34: Why do we rename this? I think `SettingsFilter` is too broad term. 
For me, it refers to some settings inside cvat annotation view.

cvat-ui/src/components/actions-menu/actions-menu.tsx
```diff
@@ -116,6 +119,16 @@ function ActionsMenuComponent(props: Props): JSX.Element {
         </Menu.Item>
     ), 50]);
 
+    if (consensusJobsPerSegment) {
+        menuItems.push([(
+            <Menu.Item
+                key={Actions.SHOW_TASK_CONSENSUS_CONFIGURATION}
+            >
+                Consensus Configuration
```
klakhov, line 127:
```suggestion
                Consensus configuration
```

cvat-ui/src/actions/consensus-actions.ts
```diff
@@ -0,0 +1,56 @@
...
+    mergeTaskConsensusJobsSuccess: (taskID: number) => (
+        createAction(ConsensusActionTypes.MERGE_CONSENSUS_JOBS_SUCCESS, { taskID })
+    ),
+    mergeTaskConsensusJobsFailed: (taskID: number, error: any) => (
+        createAction(ConsensusActionTypes.MERGE_CONSENSUS_JOBS_FAILED, { taskID, error })
+    ),
+};
+
+export const mergeTaskConsensusJobsAsync = (
+    taskInstance: any,
```
klakhov, line 43:
```suggestion
    taskInstance: Task
```

cvat-ui/src/components/consensus/consensus-modal.tsx
```diff
@@ -0,0 +1,115 @@
...
+        Notification.info({
+            message: 'Merging consensus jobs...',
+            description: 'Merge Report will be available as the merging process is completed successfully.',
+            className: 'cvat-notification-notice-export-backup-start',
+        });
+    }, [instance]);
+
+    return (
+        <Modal
+            title={<Text> Consensus Configurations </Text>}
```
klakhov, line 77:
```suggestion
            title={<Text> Consensus Configuration </Text>}
```

klakhov (commented):
Here are my ideas for the consensus settings modal:
- Notifications. Lets show that the settings are being saved not by additional notification "updating ..." but by adding a loading icon into the save button.  Also, I dont think we need Success notification. Lets not show it eighter. Just close the modal when the save is finished. 
image
- Save/Reset buttons. Need to make 'Save' as 'primary' not 'default. Also, do we really need reset functionality? In quality settings we have lots of settings and dont have reset button. From the other side it can be useful. So, if we want to keep it, I would suggest two things here: 
1. Change the large button for icon 
image from ant icons, and place it in the top of the modal
image
2. Instead of resetting right away, we need to confirm this action using simple Modal.confirm
Also its good to add cancel button just to close the modal. So, in the end we will have 'Save' as primary and 'Cancel' as default buttons in this row.
- Merge button. Currently it has some problems with styles, text is not white and backrogund becomes white on hover. From my perspective the action button is too large. 
image
We have a merge icon from the antd. So the bottom menu can look like this:
image
We can display merge status above the save/cancel as text. Lock the merge button when the process is active. Also we can show a tooltip on hovering the merge icon. What do you think?

klakhov:
The consensus tab is not yet ready, am I right?

Viditagarwal7479:
> The consensus tab is not yet ready, am I right?

Yes, how it will look is yet to be finalised.

cvat-core/src/api-implementation.ts
Viditagarwal7479: The `QualitySettingsFilter` now not only applies to `QualitySettings` but also `ConsensusSettings`. 
> I think SettingsFilter is too broad term

True, I will think about a better name and then discuss.

Viditagarwal7479: How about `AnalyticsSettingsFilter`?

Viditagarwal7479:
> Merge button. Currently it has some problems with styles, text is not white and backrogund becomes white on hover. 

This, along with the colour and hover action of the rest of the buttons, has been updated.
> In quality settings we have lots of settings and dont have reset button

Should I add a reset button to it, in a separate PR, probably?

If we add the reset button on the top right and the merge button on the bottom left as an icon with a tooltip. In that case, isn't there a possibility that it could be missed by users, and they end up searching for the merge button, or might they never use the reset functionality? Or maybe I am overthinking too much.

cvat/apps/consensus/merge_consensus_jobs.py
```diff
@@ -0,0 +1,181 @@
...
+)
+from cvat.apps.quality_control.quality_reports import JobDataProvider
+
+
+def get_consensus_jobs(task_id: int) -> Tuple[Dict[int, List[Tuple[int, User]]], List[Job]]:
+    jobs = {}  # parent_job_id -> [(consensus_job_id, assignee)]
+
+    for job in Job.objects.select_related("segment").filter(
+        segment__task_id=task_id, type=JobType.CONSENSUS.value
+    ):
```
zhiltsov-max, line 42:
```suggestion
    ).exclude(state=StateChoice.NEW, parent_job__isnull=True):
```

cvat/apps/consensus/merge_consensus_jobs.py
```diff
@@ -0,0 +1,181 @@
...
+        assert job.parent_job_id
+
+        # if the job is in NEW state, it means that the job isn't annotated
+        if job.state == StateChoice.NEW.value:
+            continue
```
zhiltsov-max, line 47:
```suggestion
```

cvat/apps/consensus/merge_consensus_jobs.py
```diff
@@ -0,0 +1,181 @@
...
+    # remove parent jobs that are not in annotation stage
+    for parent_job_id in parent_job_ids:
+        parent_job = Job.objects.filter(id=parent_job_id, type=JobType.ANNOTATION.value).first()
+
+        if parent_job.stage == StageChoice.ANNOTATION.value:
+            parent_jobs.append(parent_job)
+        else:
+            jobs.pop(parent_job_id)
+
+    return jobs, parent_jobs
```
zhiltsov-max, line 62:
```suggestion
    job_map = {}  # parent_job_id -> [(consensus_job_id, assignee)]
    parent_jobs: dict[int, Job] = {}
    for job in Job.objects.prefetch_related("segment", "parent_job").filter(
        segment__task_id=task_id, 
        type=JobType.CONSENSUS.value, 
        parent_job__stage=StageChoice.ANNOTATION.value,
        parent_job__isnull=False,
    ).exclude(state=StateChoice.NEW.value):
        job_map.setdefault(job.parent_job_id, []).append((job.id, job.assignee))
        parent_jobs.setdefault(job.parent_job_id, dob.parent_job)

    return job_map, list(parent_jobs.values())
```

cvat/apps/consensus/merge_consensus_jobs.py
```diff
@@ -0,0 +1,181 @@
...
+        ]
+
+        merged_dataset = merger(consensus_datasets)
+
+        assignee_report_data: Dict[User, Dict[str, Union[List[float], float]]] = {}
+
+        for idx, _ in enumerate(consensus_job_ids):
+            assignee_report_data.setdefault(assignees[idx], {})
+            assignee_report_data[assignees[idx]].setdefault("consensus_score", []).append(
+                merger._dataset_mean_consensus_score[id(consensus_datasets[idx])]
```
zhiltsov-max, line 106:
```suggestion
                merger.dataset_mean_consensus_score[id(consensus_datasets[idx])]
```
Please refactor so that it wasn't a private field access.

cvat/apps/consensus/merge_consensus_jobs.py
```diff
@@ -0,0 +1,181 @@
...
+
+from cvat.apps.consensus.consensus_reports import (
+    ComparisonReport,
+    generate_job_consensus_report,
+    generate_task_consensus_report,
+    save_report,
+)
+from cvat.apps.consensus.models import ConsensusSettings
+
+# from datumaro.components.operations import IntersectMerge
```
zhiltsov-max, line 23: Please remove extra comments

cvat/apps/consensus/consensus_reports.py
```diff
@@ -0,0 +1,493 @@
...
+    ConsensusSettings,
+)
+from cvat.apps.consensus.new_intersect_merge import IntersectMerge
+from cvat.apps.dataset_manager.util import bulk_create
+from cvat.apps.engine.models import Job, Task, User
+from cvat.apps.quality_control.quality_reports import AnnotationId, JobDataProvider, _Serializable
+
+
+@define(kw_only=True)
+class AnnotationConflict(_Serializable):
```
zhiltsov-max, line 37:
```suggestion
class AnnotationConflict(Serializable):
```
Please rename so that it wasn't a private symbol access.

cvat/apps/consensus/consensus_reports.py
```diff
@@ -0,0 +1,493 @@
...
+        annotation_ids = []
+        error_annotations = []
+
+        for arg in error.args:
+            if isinstance(arg, Annotation):
+                error_annotations.append(arg)
+
+        for annotation in error_annotations:
+            # the annotation belongs to which consensus dataset
+            idx = merger._dataset_map[merger._item_map[merger._ann_map[id(annotation)][1]][1]][1]
```
zhiltsov-max, line 265: There should be a function for this in the merger class.

cvat/apps/consensus/consensus_reports.py
```diff
@@ -0,0 +1,493 @@
...
+
+
+def prepare_report_for_downloading(db_report: ConsensusReport, *, host: str) -> str:
+    # copied from quality_reports.py
+    # Decorate the report for better usability and readability:
+    # - add conflicting annotation links like:
+    # <host>/tasks/62/jobs/82?frame=250&type=shape&serverID=33741
+    # - convert some fractions to percents
+    # - add common report info
+
```
zhiltsov-max, line 467: In quality reports, there was assignee info added. It might make sense to include it here as well.

cvat/apps/consensus/consensus_reports.py
```diff
@@ -0,0 +1,493 @@
...
+            target_last_updated=job.updated_date,
+            data=job_comparison_report.to_json(),
+            conflicts=[c.to_dict() for c in job_comparison_report.conflicts],
+            consensus_score=job_consensus_score,
+            assignee=job.assignee,
+        )
+        task_mean_consensus_score += job_consensus_score
+        job_reports[job.id] = job_report
+
+    task_mean_consensus_score /= len(jobs)
```
zhiltsov-max, line 378: Please try to avoid anything beyond just saving in this function.

cvat-ui/src/actions/consensus-actions.ts
```diff
@@ -0,0 +1,50 @@
+// Copyright (C) 2024 CVAT.ai Corporation
+//
+// SPDX-License-Identifier: MIT
+
+import { ActionUnion, createAction, ThunkAction } from 'utils/redux';
+import { ConsensusSettings, Task } from 'cvat-core-wrapper';
+
+export enum ConsensusActionTypes {
+    OPEN_CONSENSUS_MODAL = 'OPEN_CONSENSUS_MODAL',
```
klakhov, line 9: Consensus modal code should be removed I suppose

cvat/apps/consensus/models.py
```diff
@@ -0,0 +1,211 @@
...
+    agreement_score_threshold = models.FloatField(default=0)
+    quorum = models.IntegerField(default=-1)
+    iou_threshold = models.FloatField(default=0.5)
+    sigma = models.FloatField(default=0.1)
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+
+        if self.quorum == -1:
+            self.quorum = self.task.consensus_jobs_per_regular_job // 2
```
zhiltsov-max, line 75: I think it's better to do either in serializer or in the using code. Otherwise, this field access can invoke extra db requests or this init call can be skipped if the object is created by other means.

cvat-ui/src/components/actions-menu/actions-menu.tsx
```diff
@@ -111,6 +134,25 @@ function ActionsMenuComponent(props: Props): JSX.Element {
...
+        ), 55]);
+        menuItems.push([(
+            <Menu.Item
+                key={Actions.MERGE_CONSENSUS_JOBS}
+                disabled={isTaskInMergingConsensus}
+                icon={isTaskInMergingConsensus && <LoadingOutlined />}
+            >
+                Merge Consensus Jobs
+            </Menu.Item>
+        ), 55]);
```
klakhov, line 153: The weight here should be higher (60?)

cvat-ui/src/components/analytics-page/analytics-page.tsx
```diff
@@ -23,12 +23,14 @@ import CVATLoadingSpinner from 'components/common/loading-spinner';
...
 import AnalyticsOverview, { DateIntervals } from './analytics-performance';
 import TaskQualityComponent from './task-quality/task-quality-component';
+import TaskConsensusComponent from './task-consensus/task-consensus-component';
 
 const core = getCore();
 
 enum AnalyticsTabs {
     OVERVIEW = 'overview',
     QUALITY = 'quality',
+    CONSENSUS = 'consensus',
```
klakhov, line 33: Why do we have consensus tab on the analytics page? As we discussed, it should be separate page

cvat-ui/src/components/analytics-page/consensus-analytics-page.tsx
```diff
@@ -0,0 +1,255 @@
...
+    if (type === 'project') {
+        return +useParams<{ pid: string }>().pid;
+    }
+    if (type === 'job') {
+        return +useParams<{ jid: string }>().jid;
+    }
+    return +useParams<{ tid: string }>().tid;
+}
+
+type InstanceType = 'project' | 'task' | 'job';
```
klakhov, line 56: Isnt consensus page only for tasks?

cvat/apps/consensus/new_intersect_merge.py
```diff
@@ -0,0 +1,1983 @@
...
+)
+from datumaro.components.extractor import CategoriesInfo, DatasetItem
+from datumaro.components.media import Image, MediaElement, MultiframeImage, PointCloud, Video
+from datumaro.util import find
+from datumaro.util.annotation_util import find_instances, max_bbox, mean_bbox
+from datumaro.util.attrs_util import ensure_cls
+from scipy.optimize import linear_sum_assignment
+
+
+def OKS(a, b, sigma=0.1, bbox=None, scale=None, visibility_a=None, visibility_b=None):
```
zhiltsov-max, line 50: Probably, it's better to just import the function from `quality_reports.py` instead of copying.

cvat-ui/src/components/analytics-page/task-consensus/task-consensus-component.tsx
```diff
@@ -0,0 +1,221 @@
...
+
+    return (
+        <div className='cvat-task-quality-page'>
+            {fetching ? (
+                <CVATLoadingSpinner size='large' />
+            ) : (
+                <>
+                    <Row>
+                        <Text type='secondary'>
+                            {`Created ${taskReport?.id ? moment(taskReport.createdDate).fromNow() : ''}`}
```
klakhov, line 204: When we dont have a report it just says `Created` and empty space. Lets show `Created` only when we have the report

cvat/apps/consensus/new_intersect_merge.py
```diff
@@ -0,0 +1,1983 @@
...
+            for s in sources:
+                if not issubclass(s.media_type(), media_type) or not issubclass(
+                    media_type, s.media_type()
+                ):
+                    # Symmetric comparision is needed in the case of subclasses:
+                    # eg. Image and ByteImage
+                    raise MediaTypeError("Datasets have different media types")
+            return media_type
+
+        return None
```
zhiltsov-max, line 415: Why this class is also copied here?

cvat/apps/consensus/new_intersect_merge.py
```diff
@@ -0,0 +1,1983 @@
...
+    else:
+        a = [list(a.get_bbox())]
+        b = [list(b.get_bbox())]
+
+    # Note that mask_utils.iou expects (dt, gt). Check this if the 3rd param is True
+    return float(mask_utils.iou(b, a, [0]))
+
+
+@attrs
+class IntersectMerge(MergingStrategy):
```
zhiltsov-max, line 481: Could you inherit the original class and extend it here, instead of copying completely?

cvat/apps/consensus/new_intersect_merge.py
```diff
@@ -0,0 +1,1983 @@
...
+    if not is_bbox:
+        assert img_h is not None and img_w is not None
+        a = _to_rle(a, img_h=img_h, img_w=img_w)
+        b = _to_rle(b, img_h=img_h, img_w=img_w)
+    else:
+        a = [list(a.get_bbox())]
+        b = [list(b.get_bbox())]
+
+    # Note that mask_utils.iou expects (dt, gt). Check this if the 3rd param is True
+    return float(mask_utils.iou(b, a, [0]))
```
zhiltsov-max, line 477: Probably, it should be imported from `quality_reports.py`

cvat/apps/consensus/new_intersect_merge.py
```diff
@@ -0,0 +1,1983 @@
...
+        shape = self._merge_cluster_shape_nearest(cluster)
+        shape_score = sum(max(0, self.distance(shape, s)) for s in cluster) / len(cluster)
+        return shape, shape_score
+
+    # def __init__(self, categories=None):
+    #     _ShapeMerger.__init__(self)
+    #     SkeletonMatcher.__init__(self, categories)
+
+
+def match_segments(
```
zhiltsov-max, line 1927: It can be imported from `quality_reports.py`

cvat-ui/src/components/task-page/top-bar.tsx
```diff
@@ -26,6 +26,10 @@ export default function DetailsComponent(props: DetailsComponentProps): JSX.Elem
         history.push(`/tasks/${taskInstance.id}/analytics`);
     }, [history]);
 
+    const onViewConsensusAnalytics = (): void => {
+        history.push(`/tasks/${taskInstance.id}/analytics/consensus`);
```
klakhov, line 30: Lets place the consensus page just on /tasks/id/consensus

cvat-ui/src/reducers/notifications-reducer.ts
```diff
@@ -729,6 +732,43 @@ export default function (state = defaultState, action: AnyAction): Notifications
                 },
             };
         }
+        case ConsensusActionTypes.MERGE_CONSENSUS_JOBS_FAILED: {
+            const { taskID } = action.payload;
+            if (action.payload.error.code === 400) {
+                action.payload.error.message = "Consensus Jobs aren't annotated.";
```
klakhov, line 738: Why we specify this message on UI? Shouldnt it be coming from the server already?

cvat-ui/src/components/consensus/styles.scss
```diff
@@ -0,0 +1,39 @@
+@use 'sass:color';
+@import '../export-backup/styles';
+
+.consensus-modal-form {
```
klakhov, line 4: Its not used as I see, we should remove it

cvat-ui/src/components/consensus/styles.scss
```diff
@@ -0,0 +1,39 @@
...
+    color: white !important;
+
+    &:hover {
+      background-color: color.adjust(#1890ff, $lightness: 7%) !important;
+      color: white !important;
+    }
+  }
+}
+
+.custom-modal-center-title {
```
klakhov, line 20: The same here

cvat/apps/consensus/views.py
```diff
@@ -0,0 +1,475 @@
...
+    def create(self, request, *args, **kwargs):
+        self.check_permissions(request)
+        input_serializer = ConsensusReportCreateSerializer(data=request.data)
+        input_serializer.is_valid(raise_exception=True)
+        task_id = input_serializer.validated_data["task_id"]
+
+        queue_name = settings.CVAT_QUEUES.CONSENSUS.value
+        queue = django_rq.get_queue(queue_name)
+        # rq_id = request.query_params.get(self.CREATE_REPORT_RQ_ID_PARAMETER, None)
+        rq_id = request.data.get("rq_id", f"merge_consensus:task.id{task_id}-by-{request.user}")
```
zhiltsov-max, line 266: Please move the parameter into query params.

cvat-core/src/consensus-conflict.ts
```diff
@@ -0,0 +1,105 @@
...
+    get conflictType(): ConsensusConflictType {
+        return this.#conflictType;
+    }
+
+    get description(): string {
+        return this.#description;
+    }
+}
+
+export default class ConsensusConflict {
```
klakhov, line 61: I belive we need to re-use QualityConfilct class. Now we have too much of code duplication here. The difference in `QualityConflict` and `ConsensusConflict` is not big. We can try to make `generic Conflict` class to adress the difference in `type` and `severity` fields. Something like [this](https://www.tutorialsteacher.com/typescript/typescript-generic-class)

On the second thought, do we really need two types of conflicts(and two classes)? For what reason? Can we just adapt QualityConflict to maybe AnnotationConflict and use is as in quality and in consensus?

klakhov (commented):
There are some visual problems:
- Height of the components should be the same:
image
- Need to set width to 100%:
image
- It shouldnt be `analytics`, its `consensus`:
image

cvat/apps/consensus/new_intersect_merge.py
Viditagarwal7479: we earlier thought not to use components from one Django app in another Django app, but rather create a file containing such common functions. So should I still use the implementation from `quality_reports.py`?

zhiltsov-max: Yes, please don't duplicate this logic.

cvat-ui/src/components/analytics-page/analytics-page.tsx
Viditagarwal7479: Ah, I added this just for testing purpose. Anyway there's a seprate page for this.

cvat-core/src/api.ts
```diff
@@ -401,6 +401,26 @@ function build(): CVATCore {
                 return result;
             },
         },
+        consensus: {
+            async assignee_reports(filter = {}) {
```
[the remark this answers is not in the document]
Viditagarwal7479: Yes, sorry while shifting from python part to javascript part, naturally snake case flows.

supervisord/worker.consensus.conf
```diff
@@ -0,0 +1,27 @@
...
+# childlogdir=%(ENV_HOME)s/logs/            ; where child log files will live
+
+# [program:rqworker-consensus]
+# command=%(ENV_HOME)s/wait_for_deps.sh
+#     python3 %(ENV_HOME)s/manage.py rqworker -v 3 consensus
+#         --worker-class cvat.rqworker.DefaultWorker
+# environment=VECTOR_EVENT_HANDLER="SynchronousLogstashHandler",CVAT_POSTGRES_APPLICATION_NAME="cvat:worker:consensus"
+# numprocs=%(ENV_NUMPROCS)s
+# process_name=%(program_name)s-%(process_num)d
+# autorestart=true
```
zhiltsov-max, line 27: Hi, please uncomment the worker deployment.

cvat/apps/consensus/models.py
```diff
@@ -0,0 +1,211 @@
...
+from django.core.exceptions import ValidationError
+from django.db import models
+from django.forms.models import model_to_dict
+
+from cvat.apps.engine.models import Job, ShapeType, Task, User
+
+
+class ConsensusConflictType(str, Enum):
+    NoMatchingItemError = "no_matching_item"
+    FailedAttrVotingError = "failed_attribute_voting"
```
zhiltsov-max, line 19: It's possible to get failed attribute voting even if there are no attributes defined in the task.

cvat-ui/src/components/tasks-page/task-item.tsx
```diff
@@ -239,6 +239,10 @@ class TaskItemComponent extends React.PureComponent<TaskItemProps & RouteCompone
             history.push(`/tasks/${taskInstance.id}/analytics`);
```
zhiltsov-max, line 239: It would be nice to remember unfolded normal jobs having consensus jobs when opening these jobs from the task page. It gets a little bit annoying to open these lists each time you're returning back from a consensus job review and want to open another one.

cvat-core/src/api-implementation.ts
```diff
@@ -520,6 +526,60 @@ export default function implementAPI(cvat: CVATCore): CVATCore {
         const settings = await serverProxy.analytics.quality.settings.get(params);
         return new QualitySettings({ ...settings });
     });
+    implementationMixin(cvat.consensus.reports, async (filter: ConsensusReportsFilter) => {
+        checkFilter(filter, {
+            page: isInteger,
+            pageSize: isPageSize,
+            projectID: isInteger,
+            taskID: isInteger,
+            jobID: (value: any) => isInteger(value) || value === null,
```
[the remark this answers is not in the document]
Viditagarwal7479: In case of filtering task report, the jobID is `null`. So, to handle that.

cvat-core/src/api-implementation.ts
Viditagarwal7479: Ah, nice. Makes sense.

cvat/apps/consensus/merge_consensus_jobs.py
Viditagarwal7479: It's `parent_job_id` not `parent_job`. So instead of `job.parent_job` should I use Job.prefetch multiple times for each `parent_job_id`?

cvat/apps/consensus/new_intersect_merge.py
Viditagarwal7479: Should I replace the one in `quality_reports.py` with this? Since this particular implementation can perform IoU on both masks and bbox?

cvat/apps/consensus/consensus_reports.py
Viditagarwal7479:
```py
    def add_item_error(self, error, *args, **kwargs):
        self.errors.append(error(self._item_id, *args, **kwargs))
```
There's only this function which handles error, but it doesn't do all that we want...

cvat-core/src/assignee-consensus-report.ts
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    get assignee(): User | null {
+        return this.#assignee;
+    }
+
+    get assignee_id(): number {
+        return this.#assignee ? this.#assignee.id : 0;
+    }
+
```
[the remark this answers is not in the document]
Viditagarwal7479: Sorry. Yes, don't remember exactly but I was getting some error earlier while doing so, NVM `assignee.id` is working fine now, so removing `assigneeID`

cvat/apps/consensus/consensus_reports.py
Viditagarwal7479: assignee info is in `QualityReport` similar to what we have in `ConsensusReport` but not in `AnnotationConflict`. Am I mistaking something for something else?

cvat/apps/consensus/views.py
Viditagarwal7479: The `rq_id` will be the same for a particular task and user, so if the current query parameters are the same, the same `rq_id` will also be formed. So, should I still add it to query params?

zhiltsov-max: I'm not sure what the question is about. I'm just asking to move the parameter from the request body to the query, because in all other similar places it's passed in the query.

cvat/apps/consensus/new_intersect_merge.py
zhiltsov-max: No, the one from quality reports must be used here. Quality reports use a different function for bbox comparison, it also must be used in consensus.

cvat/apps/consensus/consensus_reports.py
zhiltsov-max: I'm speaking about [this function](https://github.com/cvat-ai/cvat/blob/a69e1228ac8ab120ca9f1274348e83ddcb89d4be/cvat/apps/quality_control/quality_reports.py#L2490-L2498) for assignee formatting in the report output. Please use it here for consistent representation of assignee in these two kinds of reports.

cvat/apps/consensus/merge_consensus_jobs.py
zhiltsov-max: There is no such relation as `parent_job_id`. It's an automatic field from Django in the `job` objects. It's stored in the job table, so fetched together with other job fields. Parent job is not a column, it's a relation, so it's fetched in `prefetch_related()`.

Viditagarwal7479: In the model, it's defined as a `parent_job_id` itself. That's why when I replaced the mentioned code with this snippet it failed with the same message.

zhiltsov-max: Well, it's better define the field as a foreign key to the same model, an example is here: https://github.com/cvat-ai/cvat/blob/a69e1228ac8ab120ca9f1274348e83ddcb89d4be/cvat/apps/engine/models.py#L818

cvat/apps/consensus/consensus_reports.py
zhiltsov-max: I can see there is `get_ann_source()`, which does half of the work. But you can always define your own function in the class, if there is no fitting one. It will be better than accessing private fields.

Viditagarwal7479: Ah, I see. I will create a similar function which does rest of the work, instead of accessing the private fields.

Viditagarwal7479:
```py
idx = merger._dataset_map[merger._item_map[merger._ann_map[id(annotation)][1]][1]][1]
annotation_ids.append(consensus_job_data_providers[idx].dm_ann_to_ann_id(annotation))
```
By any chance do you have any clue why `consensus_job_data_providers[idx].dm_ann_to_ann_id(annotation)` would be failing because of keyerror, i.e., the annotation isn't a valid annotation for the job data provider I am using...?

zhiltsov-max: No, it's hard to say it without the full context. Here are 2 possible reasons:
- it's a nested point in a skeleton, which wasn't added to the index
- `idx` is not the index of the source provider

Viditagarwal7479: It was the 2nd one. It was happening because the `self.error` list in `IntersectMerge` was storing information about all the errors encountered since it's initialisation.

cvat/apps/consensus/consensus_reports.py
```diff
@@ -252,9 +252,12 @@ def generate_assignee_consensus_report(
         )
 
     for assignee_id, assignee_info in assignee_report_data.items():
-        assignee_report_data[assignee_id]["consensus_score"] = sum(
+        assignee_consensus_score = sum(assignee_info["consensus_score"]) / len(
             assignee_info["consensus_score"]
-        ) / len(assignee_info["consensus_score"])
+        )
```
zhiltsov-max, line 256: I think simply adding `(len(...) or 0)` would also fix the problem.

cvat/apps/consensus/new_intersect_merge.py
```diff
@@ -0,0 +1,1107 @@
...
+    def _instance_bbox(
+        self, instance_anns: Sequence[dm.Annotation]
+    ) -> Tuple[float, float, float, float]:
+        return dm.ops.max_bbox(
+            a.get_bbox() if isinstance(a, dm.Skeleton) else a
+            for a in instance_anns
+            if hasattr(a, "get_bbox") and not a.attributes.get("outside", False)
+        )
+
+    def distance(a, b):
```
zhiltsov-max, line 244:
```suggestion
    def distance(self, a, b):
```

cvat/apps/consensus/new_intersect_merge.py
```diff
@@ -0,0 +1,1107 @@
...
+                        src_a,
+                        src_b,
+                    )
+                else:
+                    matches, _, _, _ = self.match_annotations_two_sources(
+                        src_a,
+                        src_b,
+                    )
+                for a, b in matches:
+                    adjacent[id(a)].append(id(b))
```
zhiltsov-max, line 370: What does this code do?

cvat/apps/consensus/new_intersect_merge.py
```diff
@@ -0,0 +1,1107 @@
...
+                dist_thresh=dist_thresh if dist_thresh is not None else self.iou_threshold,
+                **extra_args,
+            )
+
+        if self.return_distances:
+            returned_values = returned_values + (distances,)
+
+        return returned_values
+
+    def match_annotations_two_sources(self, a, b):
```
zhiltsov-max, line 322: Please add output type info, it's not clear how this method should be defined in subclasses.

cvat/apps/consensus/new_intersect_merge.py
Viditagarwal7479: This finds all matching annotations exist in the remaining sources for an annotation `a`. Based on this only the cluster formation and subsequently merging happens.

cvat/apps/consensus/consensus_reports.py
Viditagarwal7479: Oh, yes. That's nice. I didn't think about this. Thanks!

cvat/apps/consensus/new_intersect_merge.py
```diff
@@ -319,7 +320,13 @@ def _match_segments(
 
         return returned_values
 
-    def match_annotations_two_sources(self, a, b):
+    def match_annotations_two_sources(self, item_a: dm.Annotation, item_b: dm.Annotation) -> Tuple[
+        List[dm.Annotation],
+        List[dm.Annotation],
+        List[dm.Annotation],
+        List[dm.Annotation],
+        Optional[Dict[int, float]],
```
zhiltsov-max, line 328: This annotation means there are always 5 elements in the output tuple, which is not the same 4 or 5 elements.

cvat/apps/consensus/merge_consensus_jobs.py
Viditagarwal7479: Ah, on doing so it prompts that the object of type Job isn't JSON serializable...

cvat/apps/consensus/new_intersect_merge.py
zhiltsov-max: If it's a class method, why it returns these 2 kinds of output? I seems like only one of them is used, and so I suggest that the function is refactored to return only 1 kind of resulting value.

Viditagarwal7479: Two kinds of output?

Viditagarwal7479: Are you talking about match annotations two sources?

zhiltsov-max: Yes, about this `if` with 2 possible output types with 4 or 5 values:
```
if self.return_distances:
                    matches, _, _, _, _ = self.match_annotations_two_sources(
                        src_a,
                        src_b,
                    )
else:
                    matches, _, _, _ = self.match_annotations_two_sources(
                        src_a,
                        src_b,
                    )
```

Viditagarwal7479: In skeleton matcher the self.return distance is true and there we use the 5th return value

Viditagarwal7479: Instead we can make it return only two values matches and distance. The distance will be empty dict when self.return distance is false

zhiltsov-max: So, does any of the classes in the matcher or merger hierarchies uses the returned distances? If no, I don't see a point why this class method returns different value sets.

Viditagarwal7479: The one in skeleton matcher

zhiltsov-max: I can't find the specific code that uses the 5th value from a `self.match_annotations_two_sources()` call. In skeleton merger the distances are just remembered in a class field.

Viditagarwal7479: https://github.com/cvat-ai/cvat/blob/11ef2707eb3681bc0b821816ae74bb6af7c2d7dc/cvat/apps/consensus/new_intersect_merge.py#L805

zhiltsov-max: Do these distances come from a `match_annotations_two_sources()` call?

Viditagarwal7479: Yes, in the above line, it gets from self._match_segments which calls match annotations two sources

zhiltsov-max: Please refactor the code, it should not work this way.

cvat/apps/engine/migrations/0083_job_parent_job_task_consensus_jobs_per_regular_job_and_more.py
```diff
@@ -13,8 +14,14 @@ class Migration(migrations.Migration):
...
         migrations.AddField(
             model_name="job",
-            name="parent_job_id",
-            field=models.PositiveIntegerField(blank=True, default=None, null=True),
+            name="parent_job",
+            field=models.ForeignKey(
+                blank=True,
+                null=True,
+                on_delete=django.db.models.deletion.CASCADE,
+                related_name="childern_job",
```
zhiltsov-max, line 22:
```suggestion
                related_name="children_jobs",
```

cvat-core/src/consensus-conflict.ts
Viditagarwal7479: 1. The conflict types are different for both
2. There are certain params which are different like QualityConflicts have warnings and error type conflict whereas this isn't present in ConsensusConflicts

Viditagarwal7479: 1. The conflict types are different for both
2. There are certain params which are different like QualityConflicts have warnings and error type conflict whereas this isn't present in ConsensusConflicts

cvat-core/src/api-implementation.ts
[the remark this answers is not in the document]
zhiltsov-max: From this perspective, both `taskID` and `projectID` can also be null, and it must be the same as `undefined`. In the current design, the `target` field is used for distinction. E.g `taskid=42&target=task` to search just for the task report.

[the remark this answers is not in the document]
Viditagarwal7479: Makes sense, I got confused between `undefined`, `null` and use of `target` param

cvat/apps/consensus/merge_consensus_jobs.py
```diff
@@ -35,51 +35,70 @@
...
-        job_map.setdefault(job.parent_job_id, []).append((job.id, job.assignee))
-        parent_jobs.setdefault(job.parent_job_id, job.parent_job)
-
-    return job_map, list(parent_jobs.values())
+class MergeConsensusJobs:
+    jobs: Dict[int, List[Tuple[int, User]]]
+    parent_jobs: List[Job]
+    merger: IntersectMerge
+    consensus_settings: ConsensusSettings
+    assignee_jobs_count: Dict[User, int]
```
zhiltsov-max: This declaration makes these fields class fields, not instance fields. I think it's not what was supposed.

cvat/apps/consensus/merge_consensus_jobs.py
```diff
@@ -0,0 +1,282 @@
...
+            self.task_id,
+            self.parent_jobs,
+            task_report_data,
+            job_comparison_reports,
+            assignee_reports,
+            task_mean_consensus_score,
+        )
+
+    @transaction.atomic
+    def merge_single_consensus_jobs(self, parent_job_id: int) -> None:
```
zhiltsov-max, line 189:
```suggestion
    def merge_single_consensus_job(self, parent_job_id: int) -> None:
```

cvat/apps/consensus/merge_consensus_jobs.py
```diff
@@ -0,0 +1,282 @@
...
+                sigma=self.consensus_settings.sigma,
+                torso_r=self.consensus_settings.line_thickness,
+            )
+        )
+        if not self.jobs:
+            raise ValidationError(
+                "No annotated consensus jobs found or no regular jobs in annotation stage"
+            )
+
+    def get_consensus_jobs(self, task_id: int) -> None:
```
zhiltsov-max, line 63:
```suggestion
    def _get_consensus_jobs(self, task_id: int) -> None:
```

If this method changes the current object, it's better to make the function private.

cvat/apps/consensus/merge_consensus_jobs.py
```diff
@@ -0,0 +1,282 @@
...
+            )
+            .exclude(state=StateChoice.NEW.value)
+        ):
+            job_map.setdefault(job.parent_job_id, []).append((job.id, job.assignee))
+            parent_jobs.setdefault(job.parent_job_id, job.parent_job)
+
+        self.jobs = job_map
+        self.parent_jobs = list(parent_jobs.values())
+
+    def get_assignee_jobs_count(self) -> None:
```
zhiltsov-max, line 82:
```suggestion
    def _get_assignee_jobs_count(self) -> None:
```

cvat/apps/consensus/merge_consensus_jobs.py
```diff
@@ -0,0 +1,282 @@
...
+        for assignees in self.jobs.values():
+            for _, assignee in assignees:
+                if assignee not in assignee_jobs_count:
+                    assignee_jobs_count[assignee] = 1
+                else:
+                    assignee_jobs_count[assignee] += 1
+        self.assignee_jobs_count = assignee_jobs_count
+
+    @staticmethod
+    def get_annotations(job_id: int) -> dm.Dataset:
```
zhiltsov-max, line 93:
```suggestion
    def _get_annotations(job_id: int) -> dm.Dataset:
```
I don't think this function is supposed for use outside of this class.

cvat/apps/consensus/apps.py
```diff
@@ -0,0 +1,18 @@
+# Copyright (C) 2024 CVAT.ai Corporation
```
zhiltsov-max, line 1: Consider renaming the file into `merging_manager.py` or something similar to reflect it's contents better.

cvat/apps/consensus/merge_consensus_jobs.py
```diff
@@ -0,0 +1,282 @@
+# Copyright (C) 2024 CVAT.ai Corporation
```
zhiltsov-max, line 1: Consider renaming the file into `merging_manager.py` or something similar to reflect it's contents better.

cvat/apps/consensus/new_intersect_merge.py
```diff
@@ -0,0 +1,878 @@
+# Copyright (C) 2024 CVAT.ai Corporation
```
zhiltsov-max, line 1: Consider renaming this file into `merging.py` or `intersect_merge.py`.

cvat/apps/consensus/views.py
```diff
@@ -0,0 +1,474 @@
...
+                    raise NotFound(f"Report {report_id} does not exist") from ex
+
+                self.check_object_permissions(self.request, report)
+
+                if report.target == ConsensusReportTarget.TASK:
+                    queryset = queryset.filter(Q(report=report)).distinct()
+                elif report.target == ConsensusReportTarget.JOB:
+                    queryset = queryset.filter(report=report)
+                else:
+                    assert False
```
zhiltsov-max, line 120: I guess the filters should not be the same in these two cases.

cvat/apps/consensus/views.py
```diff
@@ -0,0 +1,474 @@
...
+            if task_id := self.request.query_params.get("task_id", None):
+                # NOTE: This filter is too complex to be implemented by other means
+                try:
+                    task = Task.objects.get(id=task_id)
+                except Task.DoesNotExist as ex:
+                    raise NotFound(f"Task {task_id} does not exist") from ex
+
+                self.check_object_permissions(self.request, task)
+
+                queryset = queryset.filter(Q(task__id=task_id)).distinct()
```
zhiltsov-max, line 455: Why a custom implementation is needed here? From what I can see, it just does what a default simple filter could do.

zhiltsov-max, line 455: I think this query is incorrect. Probably, you need to return all task reports, not a random one.

cvat/apps/consensus/views.py
```diff
@@ -0,0 +1,474 @@
...
+            else:
+                perm = AssigneeConsensusReportPermission.create_scope_list(self.request)
+                queryset = perm.filter(queryset)
+
+        if consensus_report_id := self.request.query_params.get("consensus_report_id", None):
+            queryset = queryset.filter(consensus_report_id=consensus_report_id)
```
zhiltsov-max, line 461:
```suggestion
            if consensus_report_id := self.request.query_params.get("consensus_report_id", None):
                queryset = queryset.filter(consensus_report_id=consensus_report_id)
```
I think you might need to either remove this branch to use just the default simple filter or add some permission check. Currently it's just the same as the default implementation of a simple filter.

cvat/apps/engine/backup.py
```diff
@@ -726,7 +730,7 @@ def _import_gt_jobs(self, jobs):
                 })
                 job_serializer.is_valid(raise_exception=True)
                 job_serializer.save()
-            elif job_type == models.JobType.ANNOTATION:
+            elif job_type == models.JobType.ANNOTATION or job_type == models.JobType.CONSENSUS:
                 continue
             else:
                 assert False
```
zhiltsov-max, line 736:
```suggestion
            else:
                continue
```

cvat/apps/engine/models.py
```diff
@@ -726,7 +727,7 @@ class Job(TimestampedModel):
 
     type = models.CharField(max_length=32, choices=JobType.choices(),
         default=JobType.ANNOTATION)
-    parent_job_id = models.PositiveIntegerField(null=True, blank=True, default=None)
+    parent_job = models.ForeignKey('self', on_delete=models.CASCADE, null=True, blank=True, related_name='childern_jobs')
```
zhiltsov-max, line 730:
```suggestion
    parent_job = models.ForeignKey('self', on_delete=models.CASCADE, null=True, blank=True, related_name='children_jobs')
```

cvat/apps/engine/backup.py
```diff
@@ -398,27 +398,31 @@ def serialize_task():
...
 
-            if self._db_task.segment_size == 0 and segment_type == models.SegmentType.RANGE:
-                segment.update(serialize_custom_file_mapping(db_segment))
+                if self._db_task.segment_size == 0 and segment_type == models.SegmentType.RANGE:
+                    segment.update(serialize_custom_file_mapping(db_segment))
 
-            return segment
+                segments.append(segment)
+
+            return segments
```
zhiltsov-max, line 420: Have you tried to run merging on tasks restored from backups? I'm not sure it's working.

cvat/apps/consensus/serializers.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def validate(self, attrs):
+        for k, v in attrs.items():
+            if (k.endswith("_threshold") or k == "line_thickness") and not 0 <= v <= 1:
+                raise serializers.ValidationError(f"{k} must be in the range [0; 1]")
+            elif k == "quorum" and not 0 <= v <= 10:
+                # since we have constrained max. consensus jobs per regular job to 10
+                raise serializers.ValidationError(f"{k} must be in the range [0; 10]")
+            elif k == "sigma" and not 0.05 <= v <= 0.2:
+                raise serializers.ValidationError(f"{k} must be in the range [0.05; 0.2]")
```
zhiltsov-max, line 122: What are these restrictions?

cvat/apps/consensus/merge_consensus_jobs.py
```diff
@@ -0,0 +1,282 @@
...
+        patch_job_data(parent_job_id, None, PatchAction.DELETE)
+        # if we don't delete exising annotations, the imported annotations
+        # will be appended to the existing annotations, and thus updated annotation
+        # would have both existing + imported annotations, but we only want the
+        # imported annotations
+
+        parent_job_data_provider = JobDataProvider(parent_job_id)
+
+        # imports the annotations in the `parent_job.job_data` instance
+        import_dm_annotations(merged_dataset, parent_job_data_provider.job_data)
```
zhiltsov-max, line 129: After merging I get annotations with the `MANUAL` source. Could you try to update it to be `CONSENSUS` instead?

image

cvat/apps/consensus/signals.py
```diff
@@ -0,0 +1,37 @@
...
+    if created:
+        if isinstance(instance, Task):
+            task = instance
+        elif isinstance(instance, Job):
+            task = instance.segment.task
+        else:
+            assert False
+
+        ConsensusSettings.objects.get_or_create(
+            task=task, quorum=task.consensus_jobs_per_regular_job // 2
```
zhiltsov-max, line 36:
```suggestion
            task=task, quorum=math.ceil(task.consensus_jobs_per_regular_job / 2)
```

cvat/apps/consensus/signals.py
```diff
@@ -0,0 +1,37 @@
+# Copyright (C) 2024 CVAT.ai Corporation
+#
+# SPDX-License-Identifier: MIT
+
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+
+# from cvat.apps.quality_control import quality_reports as qc
```
zhiltsov-max, line 8:
```suggestion
```

cvat/apps/consensus/signals.py
```diff
@@ -0,0 +1,37 @@
+# Copyright (C) 2024 CVAT.ai Corporation
+#
+# SPDX-License-Identifier: MIT
+
```
zhiltsov-max, line 4:
```suggestion

import math
```

cvat/apps/consensus/models.py
```diff
@@ -16,10 +16,8 @@
 
 class ConsensusConflictType(str, Enum):
     NoMatchingItemError = "no_matching_item"
-    FailedAttrVotingError = "failed_attribute_voting"
     NoMatchingAnnError = "no_matching_annotation"
     AnnotationsTooCloseError = "annotation_too_close"
-    WrongGroupError = "wrong_group"
```
zhiltsov-max: Why these 2 error types were removed?

cvat/apps/consensus/serializers.py
Viditagarwal7479: We discussed that the `oks_sigma` should be between `0.5%` and `20%`. That's why

cvat/apps/consensus/models.py
Viditagarwal7479: The attribute voting error here corresponds to annotation attrs, like rotation. But as we discussed that the user only knows task attributes as attributes.
Also, these two errors were not useful also in the sense of the user, as by default, we should have the attrs from the merged annotation which still happens but was raising these errors due to `quorum`

cvat/apps/consensus/serializers.py
zhiltsov-max: This might be true for some tasks, but I see no point in introducing such restrictions. You can check if this value is positive, and maybe < 1, if you want to validate it, but the current values are too restrictive.

cvat/apps/consensus/models.py
zhiltsov-max: > The attribute voting error here corresponds to annotation attrs, like rotation. But as we discussed that the user only knows task attributes as attributes.

Rotation is included in attributes in datumaro representation of CVAT annotations along with user-defined attributes. I don't see why conflicts in user-defined attributes should be ignored.

How it's related to groups? Probably, you need to check if the resulting groups are valid, if they were present before merging.

cvat/apps/consensus/serializers.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def validate(self, attrs):
+        for k, v in attrs.items():
+            if (k.endswith("_threshold") or k == "line_thickness") and not 0 <= v <= 1:
+                raise serializers.ValidationError(f"{k} must be in the range [0; 1]")
+            elif k == "quorum" and not 0 <= v <= 10:
+                # since we have constrained max. consensus jobs per regular job to 10
+                raise serializers.ValidationError(f"{k} must be in the range [0; 10]")
+            elif k == "sigma" and not 0 < v < 1:
+                raise serializers.ValidationError(f"{k} must be in the range [0; 100]")
```
zhiltsov-max, line 122:
```suggestion
                raise serializers.ValidationError(f"{k} must be in the range [0; 1]")
```
API clients will send 1, not 100.

cvat/apps/consensus/views.py
Viditagarwal7479:
```
                if report.target == QualityReportTarget.TASK:
                    queryset = queryset.filter(
                        Q(report=report) | Q(report__parent=report)
                    ).distinct()
                elif report.target == QualityReportTarget.JOB:
                    queryset = queryset.filter(report=report)
                else:
                    assert False
```
I copied it from `QualityConflictViewSet` just remove `parent` param as we don't have it here...

zhiltsov-max: I don't see how it resolves the problem with 2 different filters doing exactly the same in different lines of code. Also, why there is no parent? Aren't job reports nested in task reports?

cvat/apps/consensus/views.py
zhiltsov-max: Ok, it seems to be my mistake about the randomness, but there is still no point is using `Q` and `distinct` here.

cvat/apps/consensus/views.py
Viditagarwal7479: Job reports have task_id as null, and the same for Task reports. No there's no nesting.

zhiltsov-max:
Let's merge this PR and finalize the remaining bits in the PR with tests. I believe we got understanding about most of the implementation aspects here and only some polishing and bugfixes are left to be done.

cvat/apps/consensus/consensus_reports.py
```diff
@@ -0,0 +1,554 @@
...
+        # String keys are needed for json dumping
+        d["frame_results"] = {str(k): v for k, v in d["frame_results"].items()}
+        return dump_json(d).decode()
+
+    @classmethod
+    def from_json(cls, data: str) -> ComparisonReport:
+        return cls.from_dict(parse_json(data))
+
+
+def _get_error_type(error: dm.errors) -> str:
```
zhiltsov-max, line 237:
```
-> Optional[str]
```

PR #8044, opened 2024-06-18
Add propagate shapes action

klakhov:
### Motivation and context

The PR adds new action `Propagate shapes` that allows to propagate all filtered shapes on the current frame simultaneously

### How has this been tested?

Manual testing

### Checklist

### License

## Summary by CodeRabbit

- **New Features**
  - Introduced a new "Propagate Shapes" action to enable simultaneous propagation of filtered shapes across frames.

- **Updates**
  - Updated version numbers for `cvat-core` to `15.0.7` and `cvat-ui` to `1.63.12` indicating a new release.

- **Enhancements**
  - Improved annotation workflows with the addition of frame selection types and enhanced frame selection handling in the UI.
  - Conditional rendering and disabling of frame input fields based on the active frame selection type.


changelog.d/20240619_102529_klakhov_propagate_shapes_simultaneously.md
```diff
@@ -0,0 +1,4 @@
+### Added
+
+- `Propagate shapes` action that allows to propagate all filtered shapes simultaneously
```
bsekachev, line 3:
```suggestion
- `Propagate shapes` action to create copies of visible shapes on multiple frames forward or backward
```

cvat-core/src/object-utils.ts
```diff
@@ -356,3 +357,35 @@ export function rle2Mask(rle: number[], width: number, height: number): number[]
 
     return decoded;
 }
+
+export function propagateShapes(shapes: SerializedShape[], from: number, to: number): SerializedShape[] {
+    const getCopyFromShape = (shape: SerializedShape): SerializedShape => ({
```
bsekachev, line 362: Probably we want somehow to re-use the same code in action and in `propagateObjectAsync`

bsekachev: Some utility functions like `mask2Rle`, `rle2Mask` are already is a part of cvat-core external API

cvat-core/src/object-utils.ts
```diff
@@ -356,3 +357,35 @@ export function rle2Mask(rle: number[], width: number, height: number): number[]
...
+        attributes: shape.attributes,
```
bsekachev, line 363: It is a good idea to create a deep copy here

cvat-core/src/annotations-actions.ts
```diff
@@ -82,6 +95,45 @@ class RemoveFilteredShapes extends BaseSingleFrameAction {
...
+
+    public get name(): string {
+        return 'Propagate shapes';
+    }
+
+    public get parameters(): ActionParameters | null {
+        return {
+            'Target frame': {
+                type: ActionParameterType.NUMBER,
+                values: ['0', 'frameCount', '1'],
```
bsekachev, line 126: I like the idea of using "special values", but probably we should be more specific here.
So, I will propose to use callback, accepting an object with job and returning values, e.g.:

`({ job }) => job.startFrame` instead of `'0'` and `({ job }) => job.stopFrame` instead of `'frameCount'`

Additionally converted to strings. 

Also, frame count is 12 in a job with 11 frames, looks like a small issue. 
This approach will be quite general. 

Later we may add more input data to the callback, supporting backward compatibility

bsekachev: Or instead of `job`, maybe to use `instance: Job | Task` as actions are already implemented for both. 
Need to think. .

cvat-core/src/annotations-actions.ts
```diff
@@ -82,6 +95,45 @@ class RemoveFilteredShapes extends BaseSingleFrameAction {
...
+    public async destroy(): Promise<void> {
+        // nothing to destroy
+    }
+
+    public async run(
+        instance,
+        { collection: frameCollection, frameData: { number } },
+    ): Promise<SingleFrameActionOutput> {
+        const { shapes } = frameCollection;
+        const propagatedShapes = propagateShapes(shapes, number, this.#targetFrame);
```
bsekachev, line 114:
```suggestion
        { collection: { shapes }, frameData: { number } },
    ): Promise<SingleFrameActionOutput> {
        const propagatedShapes = propagateShapes(shapes, number, this.#targetFrame);
```

cvat-core/src/object-utils.ts
klakhov: Added this function to cvat-core api.

cvat-core/src/annotations-actions.ts
klakhov: As we dont really run actions on tasks right now, I suppose `Job` would be enough

cvat-core/src/annotations-actions.ts
```diff
@@ -28,12 +29,20 @@ export enum ActionParameterType {
     NUMBER = 'number',
 }
 
+// For SELECT values should be a list of possible options
+// For NUMBER values should be a list with [min, max, step],
+// or a callback ({ job }) => [min, max, step]
 type ActionParameters = Record<string, {
     type: ActionParameterType;
-    values: string[];
+    values: string[] | (({ job }: { job: Job }) => string[]);
```
bsekachev, line 37: What if we run action on a task? 
It is not possible on UI, but actions API assumes such ability. What job will be passed? 

It is a reason, why better to pass `instance: Job | Task`.

cvat-core/src/object-utils.ts
```diff
@@ -356,3 +358,60 @@ export function rle2Mask(rle: number[], width: number, height: number): number[]
 
     return decoded;
 }
+
+export function propagateShapes<T extends SerializedShape | ObjectState>(
+    shapes: T[], from: number, to: number,
+): T[] {
+    const getCopy = (shape: T): any => {
```
bsekachev, line 365: Probably return type is `SerializedData` from `cvat-core/src/object-state.ts`

cvat-ui/src/components/annotation-page/annotations-actions/annotations-actions-modal.tsx
```diff
@@ -204,23 +218,31 @@ function ActionParameterComponent(props: Props & { onChange: (value: string) =>
...
 
+    const job = store.getState().annotation.job.instance as Job;
     useEffect(() => {
         onChange(value);
     }, [value]);
 
+    let computedValues = values as string[];
+    if (typeof values === 'function') {
+        computedValues = values({ job });
+    }
```
bsekachev, line 232:
```suggestion
    const computedValues = typeof values === 'function' ? values({ job }) : values; 
```

cvat-ui/src/components/annotation-page/annotations-actions/annotations-actions-modal.tsx
```diff
@@ -414,6 +438,7 @@ function AnnotationsActionsModalContent(props: { onClose: () => void; }): JSX.El
                                         value={frameFrom}
                                         min={(jobInstance as Job).startFrame}
                                         max={frameTo}
+                                        disabled={currentFrameAction}
```
bsekachev, line 441: There is no reason to render this section at all if only current frame allowed

bsekachev: Maybe to render simple text like: "Running the action is only allowed on current frame"

klakhov: Added simplier text

cvat-core/src/object-utils.ts
klakhov: Well, its actually `SerializedShape | SerializedData` depending on the input datatype

cvat-core/src/annotations-actions.ts
bsekachev: But if we will replace it by `instance` in the future, it will break backward compatibility

bsekachev:
- Default value for target frame "0" is useless as I am already on the same frame. 
I will propose to use the same default values as we use in default propagation functionality (+50 or -50 frames as I remember depending on possible propagation directions).

- Also when I am on frame 0 and target frame is 0, running propagate action freezes the browser (infinite loop?)

- Found a difference in behavior how masks are propagated using default propagation and actions. Example:
  - First frame is 1000x1000, mask is on [900:1000, 900:1000].
  - Second frame is 500x500. 
  - Propagate, using classic way: empty masks are created at top left corner. But export to Segmentation mask format works. 
  - Propagate, using action: masks are copied, but outside of the frame. As result export to Segmentation mask format does not work. 

I am not sure, that copies in both cases are correct, but the second case creates more issues, not allowing to export task. 

image

bsekachev:
Also, please, add tests to cover found issues.

klakhov:
> * Default value for target frame "0" is useless as I am already on the same frame.

In default propagate we use `0`and `lastFrame`, so Ive added last frame as default

> * Also when I am on frame 0 and target frame is 0, running propagate action freezes the browser (infinite loop?)

Fixed


> * Found a difference in behavior how masks are propagated using default propagation and actions.

Fixed as discussed (cropMask if needed in `MaskShape` constructor)

Also added tests for default propagate aciton, action on current frame and mask case

cvat-core/src/annotations-objects.ts
```diff
@@ -2187,6 +2187,11 @@ export class MaskShape extends Shape {
 
     constructor(data: SerializedShape, clientID: number, color: string, injection: AnnotationInjection) {
         super(data, clientID, color, injection);
+        const [left, top, right, bottom] = this.points.slice(-4);
+        const { width, height } = this.frameMeta[this.frame];
+        if (left > width || top > height || right > width || bottom > height) {
```
bsekachev, line 2192: left >= width? 

I suppose if we have image with width 1024px, the latest pixel has coordinate 1023 on X axis and left position 1024 is already out of the image

bsekachev: The same for other coordinates

cvat-core/src/annotations-objects.ts
```diff
@@ -2187,6 +2187,11 @@ export class MaskShape extends Shape {
 
     constructor(data: SerializedShape, clientID: number, color: string, injection: AnnotationInjection) {
         super(data, clientID, color, injection);
+        const [left, top, right, bottom] = this.points.slice(-4);
+        const { width, height } = this.frameMeta[this.frame];
```
bsekachev, line 2191: This line maybe within the condition below as width, height are not used anywhere else

cvat-core/src/annotations-objects.ts
klakhov: applied

cvat-core/src/annotations-objects.ts
klakhov: How this line can be put into the condition? The values width and height are used to determine if we need to execute the condition

bsekachev: Oh, exactly

bsekachev (approved):
LGTM

PR #8049, opened 2024-06-19
Improve mask import and export performance

zhiltsov-max:
### Motivation and context

CVAT uses correct, but suboptimal algorithms for RLE enconding and decoding. This results in bad import and export performance for masks. Quality reports are also affected, as they use the same code to represent datasets.

- improved RLE encoding and decoding algorithms for mask conversions in import and export

In tests, the average speedup is:
- import: ~7.5x
- export: ~6x
- CVAT to COCO (Datumaro) mask conversions: ~17x
- COCO (Datumaro) to CVAT mask conversions: ~6x

### Checklist

### License

## Summary by CodeRabbit

- **Refactor**
  - Improved shape conversion and annotation processes by updating internal class and method names for consistency and clarity.
- **New Features**
  - Added new `rle` method to the `MaskConverter` class, enhancing functionality for mask-related operations.


cvat/apps/dataset_manager/formats/transformations.py
```diff
@@ -36,28 +37,68 @@ def transform_item(self, item):
...
+    def rle(cls, arr: np.ndarray) -> np.ndarray:
+        "Computes RLE for a flat array"
+        # adapted from https://stackoverflow.com/a/32681075
+
+        n = len(arr)
+        if n == 0:
+            return np.array([])
+
+        pairwise_unequal = arr[1:] != arr[:-1]
+        return np.diff(np.where(pairwise_unequal)[0], prepend=-1, append=n - 1)
```
SpecLad, line 100:
```suggestion
        return np.diff(np.nonzero(pairwise_unequal)[0], prepend=-1, append=n - 1)
```

This is easier to understand, and [recommended by NumPy docs](https://numpy.org/doc/stable/reference/generated/numpy.where.html).

cvat/apps/dataset_manager/formats/transformations.py
```diff
@@ -36,28 +37,68 @@ def transform_item(self, item):
...
+        left = int(x)
+        bottom = int(max(y, y + h - 1))
+        right = int(max(x, x + w - 1))
+        tight_binary_mask = dm_mask.image[top : bottom + 1, left : right + 1]
+
+        # obtain RLE
+        cvat_rle = cls.rle(tight_binary_mask.reshape(-1))
+        cvat_rle = cvat_rle.tolist()
+
+        # CVAT RLE starts from 0
```
SpecLad, line 83: IMO, this is not even a property of CVAT RLE - you just need a known starting symbol in order to decode this RLE format. So perhaps this part should be moved to `rle`?

zhiltsov-max: RLE just returns the run lengths in the sequence. Here we set the starting value to be 0, if it's not 0 - so, we can modify the input sequence.

SpecLad: > Here we set the starting value to be 0, if it's not 0 - so, we can modify the input sequence.

I don't understand. "Modify the input sequence"? Why would you need to do that?

> I think if there is a desire to include it in a function, the reason should be to avoid keeping this detail in mind.

I can't tell if you're agreeing with me or not. 🤔 My reason is to make `rle` output a complete encoding - i.e. one that can be decoded without additional information.

zhiltsov-max: Updated

SpecLad (approved):

PR #8062, opened 2024-06-20
[GSoC'24] Dependent on #8007, Basic UI interface to customize shortcuts

tahamukhtar20:
### Motivation and context
Depends on #8007

### Checklist

### License


bsekachev:
Hello,

1. As mentioned before, let's make width of input elements fixed (200px?)
2. When I start input a shortcut (e.g. A+B+C), I expect to see `A+` while holding A button, then `A+B+` while holding A+B button, and then `A+B+C+` while holding `A+B+C` buttons. Now the input field is empty until I do **keyup**.
3. Please, make sure it works with all keys on keyboard (quick look showed that it does not work with Esc, Space and Windows buttonS). 
4. Pressing `Backspace` removes previous shortcut and adds `backspace` shortcut. 
5. Let's rename `control` -> `ctrl`. 
6. Look at **Lock/unlock all objects**. There is combination `t l`. And this is not the same as `t+l` in mousetrap. Please read how it works in corresponding Mousetrap documentation. We need to support both ways. 
7. I would vote for removing shortcuts paddings on sides. It will give us more space for input elements and descriptions. Especially on small screens. 

image

bsekachev:
Let's show 'No data' when shortcuts are not found: https://ant.design/components/empty 

image

tahamukhtar20:
- [ ] Update the restore defaults function
- [ ] Update the conflict detector according to the new sequences with the delay
- [ ] Somehow prevent the default behaviour of the keys, such as when f5 is pressed the page refreshes that shouldn't be the case, similarly when the f4 is pressed it only works once, investigate that as well.
- [ ] Make sure that when shift is pressed there is a mapping that will change the keys i.e. ! => 1, @ => 2 and so on or somehow prevent the default because this is different for different keyboards.
- [ ] Disable caps lock, num lock and scroll lock keys.
- [ ] Update the tag and attribute annotation modes in the first pull request.

PR #8095, opened 2024-06-27
Requests page

klakhov:
### Motivation and context

This PR introduces new page with information about data processing(status and progress). For now it will support: task creation, import/export.

For previous discussions refer to: #7537

### Checklist

### License


klakhov:
@coderabbitai pause

zhiltsov-max:
1. The sorting header seems to be missing on the Requests tab. It's there on all the other tabs, but not in this one
image
2. I'd just show the downloading button on the card. I don't see the point in the actions section with just 1 button. Actions can include some additional actions, but downloading is the key purpose of these cards.
image
3. When we click export, the annotations are not downloaded automatically. To export, we need to make several extra clicks. If we export multiple tasks, this will be multiplied by the number of tasks exported. To me, it doesn't look a good decision. Probably, we can download automatically, if we know there were some downloading requests in this UI session (i.e. before the window was refreshed). If the window was reopened, the exports can be found in the list, as implemented.

novda:
I see that the latest test runs were successful, looks like an unstable rest-api test. Marked a failed test as unstable, if any test fails which is not related to the PR, please tag me

cvat-sdk/cvat_sdk/core/proxies/tasks.py
```diff
@@ -132,27 +132,25 @@ def upload_data(
             self._client.logger.info("Awaiting for task %s creation...", self.id)
             while True:
                 sleep(status_check_period)
-                (status, response) = self.api.retrieve_status(self.id)
+                requests, response = self._client.api_client.requests_api.list(
+                    action="create", task_id=self.id
+                )
+                assert response.status == 200
+                assert 1 == len(requests.results)
```
zhiltsov-max, line 139: It feels like the task data uploading endpoint could also return the rq_id for future status checks, as there is no response body at the moment. The pattern used here looks a little bit unreliable, as the request made above can fail and some other uploading request can be returned here instead.

cvat-sdk/cvat_sdk/core/proxies/tasks.py
```diff
@@ -132,27 +132,25 @@ def upload_data(
...
+                if "finished" == status.lower():
                     break
-                elif (
-                    status.state.value
-                    == models.RqStatusStateEnum.allowed_values[("value",)]["FAILED"]
-                ):
-                    raise exceptions.ApiException(
-                        status=status.state.value, reason=status.message, http_resp=response
-                    )
+                elif "failed" == status.lower():
```
zhiltsov-max, line 152: Probably, we need to check the statuses somehow. If their names are changed in the API, we can never know the SDK needs to be updated. I suggest that there is at least an assert or a test for this. The previous way of enum name access was far from convenient, but it also checked names automatically and would even be updated automatically after the schema changes.

cvat/apps/engine/mixins.py
```diff
@@ -462,16 +503,21 @@ def import_annotations(self, request, db_obj, import_func, rq_func, rq_id_templa
...
-    def deserialize(self, request, import_func):
+        export_backup_manager = BackupExportManager(db_object, request, version=1)
+        response = export_backup_manager.export()
+
+        if request.query_params.get('action') != 'download':
+            response.headers['Deprecated'] = True
+
+        return response
+
+    # FUTURE-TODO: migrate to new API
```
zhiltsov-max, line 519: Probably, the same comment should be added to several other v1 functions.

cvat/apps/engine/background_operations.py
```diff
@@ -0,0 +1,666 @@
...
+    ) -> None:
+        self.version = version
+        self.db_instance = db_instance
+        self.obj_type = db_instance.__class__.__name__.lower()
+        if self.obj_type not in self.SUPPORTED_RESOURCES:
+            raise ValueError(
+                "Unexpected type of db_instance: {}".format(type(db_instance))
+            )
+
+        self.callback = callback
```
zhiltsov-max, line 66: I suggest that this field is either renamed or any documentation is added to clarify it's purpose. It's not used in the base class and the name itself is quite generic.

bsekachev:
@klakhov 

Please, resolve conflicts

klakhov:
> 1. The sorting header seems to be missing on the Requests tab. It's there on all the other tabs, but not in this one

The sorting/filtering functionality was not planned in this iteration

> 2. I'd just show the downloading button on the card. I don't see the point in the actions section with just 1 button. Actions can include some additional actions, but downloading is the key purpose of these cards.

I will think about this. We may need to re-design the card a little bit. Probably this enhancement can be moved to second iteration.


> 3. When we click export, the annotations are not downloaded automatically. To export, we need to make several extra clicks. If we export multiple tasks, this will be multiplied by the number of tasks exported. To me, it doesn't look a good decision. Probably, we can download automatically, if we know there were some downloading requests in this UI session (i.e. before the window was refreshed). If the window was reopened, the exports can be found in the list, as implemented.

I belive we discussed automatic downloading several times on different meetings, in the end we decided to stick with the simplier version - download by clicks(sometimes we may want to have control what we download and in what order). I agree that sometimes it can be not-so-useful. Probably there should be a setting controlling the behavior. Anyway, the feature looks like an enhancement that needs discussion.

bsekachev:
I think `Expires on` on resources exported to a cloud storage may confuse users a little bit

image

bsekachev:
As discussed, let's increase default value for `DATASET_CACHE_TTL`. 
I believe we may use 24 hours instead of 10 hours and also, avoid dividing this value by 3 for projects

`cvat/apps/dataset_manager/views.py`, line 42

CC: @zhiltsov-max

klakhov:
> I think `Expires on` on resources exported to a cloud storage may confuse users a little bit
> 
> image
> 
> Upd: Better to use `Finished on` in this case.

Applied

zhiltsov-max (approved):

PR #8103, opened 2024-07-01
Security update

novda:
### Motivation and context

Docker Scout found several vulnerabilities, updated or removed them
### Checklist

### License

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -83,7 +83,7 @@ RUN --mount=type=cache,target=/root/.cache/pip/http-v2 \
     -r /tmp/cvat/requirements/${CVAT_CONFIGURATION}.txt \
     -w /tmp/wheelhouse
 
-FROM golang:1.22.3 AS build-smokescreen
+FROM golang:1.22.4 AS build-smokescreen
 
 RUN git clone --depth=1 https://github.com/stripe/smokescreen.git
 RUN cd smokescreen && git checkout eb1ac09 && go build -o /tmp/smokescreen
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -182,6 +182,10 @@ RUN if [ "${CVAT_DEBUG_ENABLED}" = 'yes' ]; then \
         python3 -m pip install --no-cache-dir debugpy; \
     fi
 
+# Removing pip due to security reasons. See: https://scout.docker.com/vulnerabilities/id/CVE-2018-20225
+RUN /bin/bash -c "source /opt/venv/bin/activate && pip uninstall pip -y"
```
SpecLad, line 186: IMO, this vulnerability is pretty dubious (indeed, you can see that it's marked as disputed on the CVE website). Even if we consider it real, removing pip at this point does nothing for security, seeing as we don't actually use pip at runtime. Can we just ignore it?

Dockerfile
```diff
@@ -182,6 +182,10 @@ RUN if [ "${CVAT_DEBUG_ENABLED}" = 'yes' ]; then \
...
+RUN apt-get remove -y python3-pip
```
SpecLad, line 187: I don't think this package is installed in this image in the first place.

Dockerfile
novda: Yes, I agree with you, but some tools(like docker scout) marks it as a high vulnerability(even if we don't use pip at runtime) and I thought it would be better to remove it than to explain to customers that the vulnerability is dubious.

Dockerfile
novda: removed the unnecessary deletion

Dockerfile
SpecLad: Okay, I suppose that's fine. But then you ought add a more detailed explanation in the comment than just "security reasons".

Also, that command looks overcomplicated. Wouldn't it be sufficient to do `python -mpip uninstall -y pip`?

Dockerfile.ci
```diff
@@ -25,6 +25,7 @@ RUN apt-get update && \
 COPY cvat/requirements/ /tmp/cvat/requirements/
 COPY utils/dataset_manifest/requirements.txt /tmp/utils/dataset_manifest/requirements.txt
 
+RUN /opt/venv/bin/python3 -m ensurepip
```
SpecLad, line 28:
```suggestion
RUN python3 -m ensurepip
```

Dockerfile
novda: Added more detailed comment. 

Yea, `python -m pip uninstall -y pip` is work! Fixed it.

SpecLad (approved):

PR #8129, opened 2024-07-06
only store Object rotated state in updated array

### LINKED ISSUE #8013: When default rotation is 0，can't exec update
https://github.com/cvat-ai/cvat/blob/d7cbf8c9ed31cb307b100d65ff7dd783fc29d3b4/cvat-canvas/src/typescript/canvasView.ts#L2067

when the state.rotation === 0，shape.rotate can not exec。
for example:
   1、draw a rectangle，rotate it;
   2、undo the action；
   3、the rectangle can't undo to default。

KrishavRajSingh:
fixes #8013 
While rotating the rectangle, two states were stored in array, one Object Rotated and the other Changed Points. So, I just added a check that if the object is rotated, there is no need to change points.

### Checklist

### License

## Summary by CodeRabbit

- **Enhancements**
  - Improved handling of shape updates to account for rotation.
  - Reformatted method call parameters in MaskShape for better readability.


bsekachev:
Hello,

Firstly, the same problem exists also for tracks as you only made a fix in Shape class, but not in Track. 

Second thing is that I believe the problem is not in cvat-core. In one action user may update only points or rotation, but not both. However the function `onCanvasEditDone` in `canvas-wrapper` modifies both fields. I will recommend you to look at this.

KrishavRajSingh:
> Hello,
> 
> Firstly, the same problem exists also for tracks as you only made a fix in Shape class, but not in Track.
> 
> Second thing is that I believe the problem is not in cvat-core. In one action user may update only points or rotation, but not both. However the function `onCanvasEditDone` in `canvas-wrapper` modifies both fields. I will recommend you to look at this.

You may review this

bsekachev (approved):

PR #8144, opened 2024-07-09
Make `LabeledDataSerializer.validated_data` usable

SpecLad:
### Motivation and context

Currently, the views that use `LabeledDataSerializer` for input do something unconventional: they create the serializer with the input data, call `is_valid()`, but then use the original data instead of the `validated_data` member. I believe this is because `validated_data` in this case is unusable because of the `source` attributes on some of the fields in the nested serializers.

For example, in `LabeledImageSerializer`, the `attributes` field has a source of `labeledimageattributeval_set`. This works well when serializing `LabeledImage` objects, but when you're deserializing, this creates a dict with a `labeledimageattributeval_set` key. Such objects are incompatible with functions like `patch_job_data`, which expect the `attributes` key instead.

In the current code, using `data` instead of `validated_data` seems to work okay-ish. It _is_ a bit confusing, though, because it's unconventional. For example, the `default` values set in serializer fields are effectively useless, because they're only filled in `validated_data`.

However, I'm currently working on a feature where a `LabeledDataSerializer` is incorporated into another serializer, and this problem means that I can't use `validated_data` for the parent serializer either, and that means I can't implement custom `to_internal_value` or `create` methods. So I'd much rather fix this.

While we could do it by making `patch_job_data` and others accept `labeledimageattributeval_set`, this seems counterproductive. The name `attributes` is much easier to read & understand. So instead, change the models so that the attributes of an annotation object can be accessed via `.attributes` and the shapes of a track via `.tracks`. That way, the `source` attributes become unnecessary. This fixes the problem _and_ makes the code clearer at the same time.

### Checklist

### License

## Summary by CodeRabbit

- **Bug Fixes**
  - Standardized attribute references from `trackedshape` to `shape` across the application.

- **Refactor**
  - Updated methods and functions to use more descriptive and consistent attribute names.
  - Modified foreign key relationships to include `related_name` and `related_query_name` parameters.
  - Transitioned to use `validated_data` instead of `serializer.data` for safer data handling.

- **Chores**
  - Introduced migrations to update foreign key relationships in various models.


zhiltsov-max:
As far as I remember, there was a significant performance problem related to unnecessary DB data validation, when the annotations were downloaded from the server.

SpecLad:
> As far as I remember, there was a significant performance problem related to unnecessary DB data validation in the serializer, when the annotations were downloaded from the server.

Sure, but how does that impact this PR?

bsekachev:
LGTM


The following serializers: 

- LabeledImageSerializerFromDB
- LabeledShapeSerializerFromDB
- LabeledTrackSerializerFromDB

became almost useless. 
As I see, the only thing they do now, is removing `id` field from items of `attributes` list.

PR #8183, opened 2024-07-17
Fixed importing skeletons in a specific scenario

bsekachev:
### Motivation and context
1. Create a minor task, with one segment, let's say 4 frames.
1. Create a skeleton shape on the first frame, propagate it until the latest frame, save, close, export task annotations as CVAT for Videos. 
1. Create the same task, but with two segments (segment size is 2), thus you will have two segments. Upload annotations. You will not be able to open the second task anymore. 

In the second task:
the second segment will have two skeletons from the first segment. But elements of this skeletons will not have shapes.

That happens because skeleton dumped as two shapes into format:

```
<track>
    <skeleton frame=1>
         <points outside=0>
    </skeleton>
    <skeleton frame=2>
         <points outside=1>
     </skeleton>
</track>
```

When imported, skeleton does not have "outside" flags and for skeleton two shapes are defined:
```
{"type": "skeleton", "frame": 1, "outside": False }
{"type": "skeleton", "frame": 2, "outside": False } # I believe this is also error, but this pull request does not fix that
```

Based on this information, method `_is_track_inside` consider such skeletons to exist in the second segment, however any child elements does not exist in the second segment.  

The second corner case related to skeletons:

`skeleton` will have one `element` with two `shape` in the same scenario 
```
{"type": "points", "frame": 1, "outside": False }
{"type": "points", "frame": 2, "outside": True }
```

And current implementation of `_is_track_inside` will return true for this point (and for skeleton consequently) [in the second segment where start frame is 2 and stop frame is 3]. 
However that is not exactly correct as this segment only has outside frame.  

### Checklist

### License

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of shapes in a track, addressing corner cases and ensuring correct overlap checks.
  - Enhanced logic for filtering and slicing track shapes, including better handling of leading outside shapes and removing unnecessary ones.
  - Added checks for skeleton elements without shapes to ensure correct processing.

### DIFF
--- a/cvat/apps/dataset_manager/annotation.py
+++ b/cvat/apps/dataset_manager/annotation.py
@@ -72,14 +72,22 @@ class AnnotationIR:
             # a <= b
             return 0 <= min(b, stop) - max(a, start)
 
+        elements = track.get('elements', [])
+        if elements:
+            # skeletons usually have only one shape defined on initial frame
+            # anyway, for them we decided if they are inside range based on child elements
+            return any(AnnotationIR._is_track_inside(t, start, stop) for t in elements)
+
         prev_shape = None
         for shape in track['shapes']:
             if prev_shape and not prev_shape['outside'] and \
                     has_overlap(prev_shape['frame'], shape['frame']):
                 return True
-            prev_shape = shape
 
-        if not prev_shape['outside'] and prev_shape['frame'] <= stop:
+            if shape['frame'] >= start:
+                prev_shape = shape
+
+        if prev_shape and not prev_shape['outside'] and prev_shape['frame'] <= stop:
             return True
 
         return False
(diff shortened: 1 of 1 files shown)


zhiltsov-max (commented):
Seems to be working. I think I have found another problem. In the exported file, tracks from the last job don't have a finishing shape. If there are extra frames in after the last one in annotations in the target task, the last tracks will be unexpectedly continued.

cvat/apps/dataset_manager/annotation.py
```diff
@@ -72,11 +72,23 @@ def has_overlap(a, b):
...
+            # skeletons usually have only one shape defined on initial frame
+            # anyway, for them we decided if they are inside range based on child elements
+            return any(AnnotationIR._is_track_inside(t, start, stop) for t in elements)
+
         prev_shape = None
         for shape in track['shapes']:
+            if shape['frame'] == start and shape['outside']:
+                # corner case when the only shape on segment is outside frame on start frame
+                prev_shape = shape
+                continue
```
zhiltsov-max, line 86: I'm not totally sure this function should be responsible for deciding anything about segments. Technically, there could have been the stop frame, if it wasn't dropped in the `slice_track` function. Thus, there is another possible place to fix the problem - in [this line](https://github.com/cvat-ai/cvat/blob/5bb83a150fbdfecb0c46b5b0a314ed6c98215e5e/cvat/apps/dataset_manager/annotation.py#L159) (I can guess that that check also was introduced because of this problem).

PR #8186, opened 2024-07-17
[GSoC'24] Dependent on #8007, UI, Working Conflict Detector and Shortcuts adjusted according to different Scopes, with the addition of Local Storage

tahamukhtar20:
### Checklist

### License

## Summary by CodeRabbit

- **New Features**
  - Introduced new action types for managing keyboard shortcuts, allowing for dynamic updates.
  - Added a new component for managing multiple keyboard shortcuts with conflict resolution.
  - Enhanced settings modal to include keyboard shortcuts management.

- **Improvements**
  - Refined scope for many keyboard shortcuts to ensure they are relevant to specific contexts in the application, improving user experience.
  - Enhanced shortcut conflict detection and management within the shortcuts reducer.

- **Style**
  - Updated styles for the shortcuts settings section in the UI for better organization.

- **Documentation**
  - Added comments and updated interfaces to improve clarity on shortcut management functionalities.


bsekachev:
@tahamukhtar20 

Please, resolve conflicts and write TODO list in the PR description what steps should we done before merging

KTXKIKI (approved):

KTXKIKI (changes requested):
1

bsekachev:
@tahamukhtar20 
May you please resolve conflicts?

tahamukhtar20:
> @tahamukhtar20 May you please resolve conflicts?

Resolved

bsekachev:
Tried to set shortcut `f2 f3`

Clicking **Okay** but nothing happens. Also there is an exception in the console.
image

Please, cover this issue by a test.

bsekachev:
**UX issue:** in order to save shortcuts, I have to scroll the whole list. List is quite big what makes saving a time consuming process. **Save** button should be always visible.

tahamukhtar20:
I'll look into the issue, but the exception in console is added intentionally.

bsekachev:
We do not need any exceptions in console intentionally. Usually such exceptions means that there are issues with code.
Sometimes we add some exceptions if we want to see them in analytics, but this information we definitely do not.

bsekachev:
image

1. Collapse all scopes.
2. In search put "sdfsdfsd" (see No data, as expected)
3. Clear search 
4. Now all scopes are opened

bsekachev:
Many shortcuts are not in correct scope

Switch automatic bordering => Standard Mode
Split mode => Standard Mode
Paste shape => Standard Mode
Switch algorithm blocker => Standard Mode
Toggle layout grid => Global

Lock/unlock an object => Available on both: Standard and Attribute Annotation Mode
The same for occluded and pinned

Please, check each shortcut is in correct scope.

bsekachev:
Let's remove **Shortcuts** from each title:
image

bsekachev:
image

Added **ctrl + space** shortcut. It may open a list of shortcuts, but can't close.

**Please add a test case.**

bsekachev:
Shortcut conflicts with itself?

image

Added `f f f`. `f` does not work anymore

bsekachev:
This button worthy of additional confirmation as user may accidentally click it. And in this case user's setup will be lost. 

image

bsekachev:
image

I suppose it will not work if we change sequences or add new sequences. Please, check. 
Probably we need to split to 10 different shortcuts.

bsekachev:
I see both `right` and `arrowright`. Are these keys different?

bsekachev:
Please, do different checks of your patch on your own first.

- For each shortcut in the list, try to add couple of sequences and make sure they work.
- Play with conflicts detector.
- Try different special keys and their combinations (ctrl, alt, tab, shift, page up, page down, home, end, insert, delete)

bsekachev:
At least on Linux Ctrl + Tab switches to another browser Tab. Did you try this shortcut? Why this default behavior is not prevented?

Please, left `Tab` button as it was. 

image

tahamukhtar20:
> I see both `right` and `arrowright`. Are these keys different?

these are not different, this might have been an error carry forward from the original code. I'll update it to right.

tahamukhtar20:
> At least on Linux Ctrl + Tab switches to another browser Tab. Did you try this shortcut? Why this default behavior is not prevented?
> 
> Please, left `Tab` button as it was.
> 
> image

All the default behaviours have been prevented, but apparently some behaviours cant be prevented for some reason, I have changed it ctrl+tab for now because it was causing a conflict with
```
NEXT_OBJECT: {
        name: 'Next object',
        description: 'Go to the next object',
        sequences: ['tab'],
        scope: ShortcutScope.ATTRIBUTE_ANNOTATION_WORKSPACE,
    },
```

tahamukhtar20:
one solution of this I can think of is to manually check which sequences can't be prevented and those are not allowed to register.

bsekachev:
> it was causing a conflict with

That is not conflict.
The first one is in STANDARD workspace
The second on is in ATTRIBUTE ANNOTATION MODE workspace

tahamukhtar20:
> > it was causing a conflict with
> 
> That is not conflict. The first one is in STANDARD workspace The second on is in ATTRIBUTE ANNOTATION MODE workspace

Oh okay, it had an incorrect scope. will update scope.

tahamukhtar20:
> Added **ctrl + space** shortcut. It may open a list of shortcuts, but can't close.
> 
> **Please add a test case.**

Just to let you know, the keys were hard coded in the code before for the modals =>
image
I've updated the code so that the sequences are dynamic, but the support is only added for the Show Shortcuts and Show Settings for now because only these two shortcuts are associated with modals =>
image

tahamukhtar20:
> At least on Linux Ctrl + Tab switches to another browser Tab. Did you try this shortcut? Why this default behavior is not prevented?
> 
> Please, left `Tab` button as it was.
> 
> image

https://stackoverflow.com/questions/59952382/using-preventdefault-to-override-opening-new-tab

As seen in this, some sequences can't be prevented because those are protected.

tahamukhtar20:
Another thing, I tried adding conditions which might have helped in notifying the user that this shortcut is unavailable or something, but I'm afraid that isn't possible, there is a list of shortcuts which chrome says are it's shortcuts =>

https://support.google.com/chrome/answer/157179?hl=en&co=GENIE.Platform%3DDesktop#zippy=%2Ctab-window-shortcuts

But the thing is that this list is weird, half of the shortcuts can't be prevented where whereas the other half can be prevented. as an example ctrl + 1 is a reserved shortcut according to this webpage saying that it switches to the first tab, but it can be prevented using javascript, but that isn't the case for the ctrl + tab, ctrl + shift + tab, ctrl + n etc. I think we might need to include this in the docs that some reserved shortcuts aren't available. Any suggestions are appreciated.

tahamukhtar20:
And about testing it with various keyboard layouts, it only takes the underlying key according to the keyboards of QWERTY layout.
Extra symbols and keys which don't exist in https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode, are not supported.

tahamukhtar20:
> Just to let you know, the keys were hard coded in the code before for the modals =>

Another thing, you should know, the modal doesn't close if you use shortcuts with delay, this is because the Mousetrap callback is activated on keyboard event, so it doesn't wait for the user to complete it's combination and instead only checks if that combination is present or not, so in this case, f1 would work, ctrl + f1 would also work but something like f1 f2 f2 won't.

tahamukhtar20:
Other than that, I've thoroughly tested this; there shouldn't be any issues like this anymore. But still, in case you find any I'll fix those as soon as possible.

bsekachev:
> I think we might need to include this in the docs that some reserved shortcuts aren't available.

Yep, let's go through this way.

bsekachev:
> f1 would work, ctrl + f1 would also work but something like f1 f2 f2 won't.

I am not sure I understand your idea here.
If we have setup `f1` and `f1 f2 f2` for example to switch a modal, the modal will be switched after user enters `f1` not waiting `f1 f2 f2`. But these two shortcuts are conflicting. 

On another hand, if we use only `f1 f2 f2` to switch a modal, it should work.

bsekachev:
Cancel => Standard Workspace, Single Shape Workspace

Rotate clockwise => Standard workspace, Review workspace
Rotate anticlockwise => Standard workspace, Review workspace  

There are issues with a list of shortcuts in 3D mode. For example we may paste shape, but can't copy shape. How would it be possible?
Also, group features are available there as far as I remember. Please, check for others shortcuts in this scope (occluded, next keyframe, etc)

Such shortcuts as occluded, next keyframe, lock, pinned, etc are available in standard, standard 3d, attribute annotation mode
Some of them are also available in review mode. 

**I would ask you to check again that each shortcut is in correct scope.**

bsekachev:
image

Let's update title: **Assign attribute value "value"**

bsekachev:
**Change color** also available in review mode and in standard 3d mode

bsekachev:
I will suggest to collapse all titles by default. 

image

bsekachev:
image

Please, put: 
**Next frame** in scope **Annotation page**
on a single line

"Do you want to" -> "Would you like to"

bsekachev:
I would propose a small enhancement, add opacity: 0.5 to inactive shortcuts in attribute annotation, tag annotation and objects sidebar scopes. **Like this:**  

Also on screenshot "Create a new tag 0" should go after "Create a new tag 9". 

image

bsekachev:
Similar thing in the list of active shortcuts:
image

Actually these shortcuts will do nothing, they are specified, but labels for them do not exist. 
I will just propose to consider a shortcut without registered handler as non active. What do you think?

bsekachev:
Also, acccording to your discovers, I will propose to add something like warning at the top of the list:

"Some key combinations may be reserved by the browser and cannot be overridden in CVAT"

image

Maybe using Alert component:  https://ant.design/components/alert

cvat-ui/src/components/annotation-page/attribute-annotation-workspace/attribute-annotation-sidebar/attribute-editor.tsx
```diff
@@ -256,14 +258,56 @@ interface ListParameters {
...
+    const sortedValues = ['true', 'false'];
+    const filteredValues = values.filter((value: string): boolean => value !== config.UNDEFINED_ATTRIBUTE_VALUE);
+    const prevValuesRef = useRef<string[] | null>(null);
+
+    useDynamicLabels(componentShortcuts);
+
+    useEffect(() => {
+        if (!isEqual(values, prevValuesRef.current)) {
+            const updatedComponentShortcuts = {
+                ...Object.keys(componentShortcuts).reduce((acc: any, key) => {
```
bsekachev, line 273: Extra destructurization

```suggestion
            const updatedComponentShortcuts = Object.keys(componentShortcuts).reduce((acc: any, key) => {
```

cvat-ui/src/components/annotation-page/attribute-annotation-workspace/attribute-annotation-sidebar/attribute-editor.tsx
```diff
@@ -256,14 +258,56 @@ interface ListParameters {
...
-    if (inputType === 'checkbox') {
-        const sortedValues = ['true', 'false'];
-        if (values[0].toLowerCase() !== 'true') {
-            sortedValues.reverse();
+    const { keyMap } = useSelector((state: CombinedState) => state.shortcuts);
+    const sortedValues = ['true', 'false'];
+    const filteredValues = values.filter((value: string): boolean => value !== config.UNDEFINED_ATTRIBUTE_VALUE);
+    const prevValuesRef = useRef<string[] | null>(null);
+
+    useDynamicLabels(componentShortcuts);
```
bsekachev, line 268: I am not sure I understand what the hook does and why we need it here. Could you please elaborate?

cvat-ui/src/components/annotation-page/standard-workspace/objects-side-bar/labels-list.tsx
```diff
@@ -3,28 +3,36 @@
 //
 // SPDX-License-Identifier: MIT
 
-import React, { useCallback, useState } from 'react';
+import React, {
+    useCallback,
+    useEffect,
+    useState,
+} from 'react';
```
bsekachev, line 10:
```suggestion
import React, { useCallback, useEffect, useState } from 'react';
```

cvat-ui/src/components/annotation-page/attribute-annotation-workspace/attribute-annotation-sidebar/attribute-editor.tsx
tahamukhtar20: This is to reset the shortcuts to their original state once the component unmounts, there is a ref inside because when added to the dependency array it was causing infinite rerenders.

cvat-ui/src/components/annotation-page/standard-workspace/objects-side-bar/labels-list.tsx
```diff
@@ -3,28 +3,36 @@
...
-        scope: ShortcutScope.ALL,
-    },
-};
+const componentShortcuts: Record<string, KeyMapItem> = {};
+
+for (const i of [1, 2, 3, 4, 5, 6, 7, 8, 9, 0]) {
+    componentShortcuts[`SWITCH_LABEL_${i}`] = {
+        name: `Switch label ${i}`,
+        description: `Changes the label to label ${i} for the activated
+        object or for the next drawn object if no objects are activated`,
```
bsekachev, line 31:
```suggestion
        name: `Switches label`,
        description: `Changes label of a selected object or default label of the next created object if no one object is activated`,
```

cvat-ui/src/components/annotation-page/standard-workspace/objects-side-bar/labels-list.tsx
```diff
@@ -37,11 +45,40 @@ function LabelsListComponent(): JSX.Element {
...
-
     const [keyToLabelMapping, setKeyToLabelMapping] = useState<Record<string, number>>(
         Object.fromEntries(labelIDs.slice(0, 10).map((labelID: number, idx: number) => [(idx + 1) % 10, labelID])),
     );
 
+    useDynamicLabels(componentShortcuts);
+
+    useEffect(() => {
+        const updatedComponentShortcuts = {
+            ...Object.keys(componentShortcuts).reduce((acc: any, key) => {
```
bsekachev, line 56:
```suggestion
        const updatedComponentShortcuts = {
            ...Object.keys(componentShortcuts).reduce((acc: any, key) => {
```
```suggestion
        const updatedComponentShortcuts = Object.keys(componentShortcuts).reduce((acc: any, key) => {
```

cvat-ui/src/components/annotation-page/standard-workspace/objects-side-bar/labels-list.tsx
```diff
@@ -37,11 +45,40 @@ function LabelsListComponent(): JSX.Element {
...
+                        ...componentShortcuts[key],
+                        sequences: keyMap[key].sequences,
+                    };
+                }
+                return acc;
+            }, {}),
+        };
+
+        for (const [id, labelID] of Object.entries(keyToLabelMapping)) {
+            if (labelID) {
```
bsekachev, line 68: How is that possible that `labelID` may be something else rather than number?

cvat-ui/src/components/annotation-page/standard-workspace/objects-side-bar/labels-list.tsx
```diff
@@ -37,11 +45,40 @@ function LabelsListComponent(): JSX.Element {
...
+                updatedComponentShortcuts[`SWITCH_LABEL_${id}`] = {
+                    ...updatedComponentShortcuts[`SWITCH_LABEL_${id}`],
+                    name: `Switch label to ${labels.filter((label: any) => label.id === labelID)[0].name}`,
```
bsekachev, line 71: Please, use Array.find

cvat-ui/src/components/annotation-page/standard-workspace/objects-side-bar/labels-list.tsx
```diff
@@ -37,11 +45,40 @@ function LabelsListComponent(): JSX.Element {
...
+                    description: `Changes the label to ${
+                        labels.filter((label: any) => label.id === labelID)[0].name
```
bsekachev, line 73: Please, use Array.find only once. You do not need to search twice

cvat-ui/src/components/annotation-page/standard-workspace/objects-side-bar/labels-list.tsx
```diff
@@ -78,34 +115,42 @@ function LabelsListComponent(): JSX.Element {
...
-                    const bothAreTags = activeObjectType === ObjectType.TAG && label.type === ObjectType.TAG;
-                    const labelIsApplicable = label.type === LabelType.ANY ||
-                        activeShapeType === label.type || bothAreTags;
-                    if (labelIsApplicable) {
-                        dispatch(rememberObject({ activeLabelID: labelID }));
-                        message.destroy();
-                        message.success(`Default label has been changed to "${label.name}"`);
-                    }
-                }
+                handleHelper(labelID);
```
bsekachev, line 150: If you moved code to the function ``handleHelper``, is there any reason to left this line of code?
Why not put everything to the function ``handleHelper``? Currently in the function ``handleHelper`` we are looking for the label again

```
const label = labels.filter((_label: any) => _label.id === labelID)[0];
```

cvat-ui/src/components/annotation-page/attribute-annotation-workspace/attribute-annotation-sidebar/attribute-editor.tsx
bsekachev: I see, let's then rename to `useResetShortcutsOnUnmount`

cvat-ui/src/utils/hooks.ts
```diff
@@ -130,3 +132,23 @@ export function useAuthQuery(): Record<string, string> | null {
...
+
+    useEffect(() => {
+        keyMapRef.current = keyMap;
+    }, [keyMap]);
+
+    useEffect(() => () => {
+        const revertedShortcuts = Object.entries(componentShortcuts).reduce((acc: any, [key, value]) => {
+            acc[key] = {
+                ...value,
+                sequences: keyMapRef.current[key] ? keyMapRef.current[key].sequences : value.sequences,
```
bsekachev, line 148:
```suggestion
                sequences: keyMapRef.current[key]?.sequences ?? value.sequences,
```

bsekachev: Also I think that taking `value.sequences` as default value may be incorrect choice as it can lead to conflicts with registered shortcuts. Thus, better to take empty array, don't you think?

tahamukhtar20: I'll check

cvat-ui/src/components/annotation-page/tag-annotation-workspace/tag-annotation-sidebar/shortcuts-select.tsx
```diff
@@ -131,29 +70,34 @@ function ShortcutsSelect(props: Props): JSX.Element {
         setShortcutLabelMap(newShortcutLabelMap);
     }, []);
 
+    useDynamicLabels(componentShortcuts);
+
     useEffect(() => {
         const updatedComponentShortcuts = {
-            ...componentShortcuts,
+            ...Object.keys(componentShortcuts).reduce((acc: any, key) => {
```
bsekachev, line 77: extra destructurization

cvat-ui/src/components/annotation-page/tag-annotation-workspace/tag-annotation-sidebar/shortcuts-select.tsx
```diff
@@ -131,29 +70,34 @@ function ShortcutsSelect(props: Props): JSX.Element {
...
+                if (keyMap[key]) {
```
bsekachev, line 78: May you please explain when this condition may be falsy?

bsekachev: As I understand it is only in order to TypeScript does not show any warnings
From the point of code, it would be clearer if we just have an action to update existing shortcuts.

This method would accept partial body for many shortcuts and apply it to the storage, e.g.

```
function updateShortcutAction({ shortcutID: Partial<ShortcutBody> }) ... 
```

Otherwise we need checks like this and a little weird thing about it is:
We refer to `keyMap` in `useEffect`, and according to idea, `keyMap` has to be in useEffect dependencies. However I understand it will lead to infinite rendering loop.

bsekachev: By the way, probably the same issue forces you to use `useRef` in hook to reset defaults

cvat-ui/src/components/header/settings-modal/multiple-shortcuts-display.tsx
```diff
@@ -0,0 +1,169 @@
+import React, { useRef, useState, useEffect } from 'react';
```
bsekachev, line 1: Please add a default license header to newly created files

cvat-ui/src/components/header/settings-modal/multiple-shortcuts-display.tsx
```diff
@@ -0,0 +1,169 @@
...
+import { Select, Modal } from 'antd/lib';
+import { conflict, conflictDetector } from 'utils/conflict-detector';
+import { ShortcutScope } from 'utils/enums';
+import { KeyMapItem } from 'utils/mousetrap-react';
+import { registerComponentShortcuts } from 'actions/shortcuts-actions';
+import { getKeyfromCode } from 'utils/key-code-mapper';
+import { isEqual } from 'lodash';
```
bsekachev, line 8: 3rdparty imports should go first

bsekachev:
Common comment: please, reduce using `any` type in the patch. Type shall be as specific as possible

cvat-ui/src/components/header/settings-modal/multiple-shortcuts-display.tsx
```diff
@@ -0,0 +1,169 @@
...
+    const [timer, setTimer] = useState<NodeJS.Timeout | null>(null);
+
+    useEffect(() => () => {
+        if (timer) clearTimeout(timer);
+    }, [timer]);
+
+    function unsetExistingShortcut(
+        conflictingShortcuts: Record<string, KeyMapItem>,
+        updatedSequence: string[],
+        shortcut: Record<string, KeyMapItem>): void {
```
bsekachev, line 38: This function always accept only one shortcut. Why do we need record here?

cvat-ui/src/components/header/settings-modal/multiple-shortcuts-display.tsx
```diff
@@ -0,0 +1,169 @@
...
+        const updatedShortcuts: Record<string, KeyMapItem> = {};
+        for (const [key, value] of Object.entries(conflictingShortcuts)) {
+            const currentItem: Record<string, KeyMapItem> = {
+                [key]: value,
+            };
```
bsekachev, line 43: Why not do the same thing simpler? 

```suggestion
        for (const key of Object.keys(conflictingShortcuts)) {
            const conflictingItem: KeyMapItem = conflictingShortcuts[key];
```

cvat-ui/src/components/header/settings-modal/multiple-shortcuts-display.tsx
```diff
@@ -0,0 +1,169 @@
...
+            if (isEqual(Object.keys(currentItem), Object.keys(shortcut))) {
```
bsekachev, line 44: I am not sure I understand this piece of code.
If you comparing keys arrays, the condition will be truthy only when we `currentItem` and `shortcut` contain the same item. Was this your idea?

cvat-ui/src/components/header/settings-modal/multiple-shortcuts-display.tsx
```diff
@@ -0,0 +1,169 @@
...
+                };
+            } else {
+                const commonSequence = updatedSequence.map((s) => {
+                    for (const seq of value.sequences) {
+                        if (conflict(seq, s)) {
+                            return seq;
+                        }
+                    }
+                    return '';
+                });
```
bsekachev, line 59: You need Array.filter here

cvat-ui/src/components/header/settings-modal/multiple-shortcuts-display.tsx
```diff
@@ -0,0 +1,169 @@
...
+                    ...value,
+                    sequences: [...currentSequence.filter((s) => !conflict(s, newSequence)), newSequence],
+                };
+            } else {
+                const commonSequence = updatedSequence.map((s) => {
+                    for (const seq of value.sequences) {
+                        if (conflict(seq, s)) {
+                            return seq;
+                        }
+                    }
```
bsekachev, line 57: You need Array.some here

cvat-ui/src/components/header/settings-modal/multiple-shortcuts-display.tsx
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    function conflictNotifier(keyMapId: string, updatedSequence: string[]): void {
+        const shortcut = {
+            [keyMapId]: { ...keyMap[keyMapId], sequences: updatedSequence },
+        };
+        const conflictingShortcuts: Record<string, KeyMapItem> | null = conflictDetector(shortcut, keyMap);
+        if (conflictingShortcuts) {
+            Modal.confirm({
+                title: 'Shortcut conflict',
```
bsekachev, line 74:
```suggestion
                title: 'Conflicting shortcuts detected',
```

cvat-ui/src/components/header/settings-modal/multiple-shortcuts-display.tsx
```diff
@@ -0,0 +1,169 @@
...
+                content: (
+                    <p>
+                        This sequence conflicts with the following shortcuts:
```
bsekachev, line 77:
```suggestion
                        Added sequence conflicts with the following shortcuts:
```

cvat-ui/src/components/header/settings-modal/multiple-shortcuts-display.tsx
```diff
@@ -0,0 +1,169 @@
...
+                        <strong>{Object.values(conflictingShortcuts)[0].name}</strong>
+                        {' '}
+                        in the scope
+                        {' '}
+                        <strong>
+                            {ShortcutScope[
+                                Object.values(conflictingShortcuts)[0].scope].split('_').join(' ')}
+                        </strong>
+                        <br />
+                        Would you like to unset the conflicting shortcuts?
```
bsekachev, line 88: Do you think will it be possible to have conflict more than with one shortcut? 
Just from my understanding, if shortcut conflicts with two existing. Then before user change these two existing are already conflicted between each other (what is not possible to set)

cvat-ui/src/components/header/settings-modal/multiple-shortcuts-display.tsx
```diff
@@ -0,0 +1,169 @@
...
+                    </p>
+                ),
+                onOk: () => {
+                    onKeySequenceUpdate(keyMapId, updatedSequence);
+                    unsetExistingShortcut(conflictingShortcuts, updatedSequence, shortcut);
```
bsekachev, line 93:
```suggestion
                    unsetExistingShortcut(conflictingShortcuts, updatedSequence, shortcut);
                    onKeySequenceUpdate(keyMapId, updatedSequence);
```

bsekachev: Just reversed the order. Logically it looks more correct

cvat-ui/src/components/header/settings-modal/multiple-shortcuts-display.tsx
```diff
@@ -0,0 +1,169 @@
...
+                },
+                onCancel: () => {},
```
bsekachev, line 95: Do we need this empty function?

cvat-ui/src/components/header/settings-modal/multiple-shortcuts-display.tsx
tahamukhtar20: actually it is possible for two to not conflict with each other but conflict with a 3rd shortcut, example,
alt + c and alt + d are registered, these don't conflict with each other now if I set a new shortcut that is has the sequence alt, it will conflict with both.

cvat-ui/src/components/header/settings-modal/multiple-shortcuts-display.tsx
tahamukhtar20: Yes, there is a different condition if the conflict is with itself and a different condition otherwise.

cvat-ui/src/components/header/settings-modal/multiple-shortcuts-display.tsx
tahamukhtar20: It can accept multiple as there can be multiple conflicts as explained the comment below.

cvat-ui/src/components/header/settings-modal/multiple-shortcuts-display.tsx
tahamukhtar20: it doesn't work that way, the shortcut doesn't unset because of the way the code is written.

tahamukhtar20: At least if the conflict is the with itself.

cvat-ui/src/components/header/settings-modal/multiple-shortcuts-display.tsx
```diff
@@ -0,0 +1,169 @@
...
+    const { sequences } = item;
+    const selectRef = useRef<any>(null);
+    const [focus, setFocus] = useState(false);
+    const [pressedKeys, setPressedKeys] = useState<string[][]>([[]]);
+    const [currentIdx, setCurrentIdx] = useState<number>(0);
+    const [timer, setTimer] = useState<NodeJS.Timeout | null>(null);
+
+    useEffect(() => () => {
+        if (timer) clearTimeout(timer);
+    }, [timer]);
```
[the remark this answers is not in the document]
tahamukhtar20: I tried your approach, maybe it was my implementation or whether the issue was with the approach, but the timeout isn't cleared, and because of that it tries to finalize the shortcut multiple times, hence giving me multiple pop-ups incase of a conflict

bsekachev:
You are trying to save a lot of extra data to local storage:
- Normalized keymap
- Name, scope, description
- Default state

These things are defined by CVAT, they should not be stored in local storage.
Additionally, local storage is very limited by its size, and we should econom it

image
image

image

bsekachev:
In case of restoring defaults if data in local storage are broken, the feature does not work:
image

Actually, I first was checking that with incorrect value in local storage CVAT will be able to load without errors.
Hovewer I discovered that you use local storage to restore default values, what is:
1. Does not work in this case
2. I really do not understand why we need to use local storage at all for the feature. Any shortcut says default sequences in runtime during registration. Moreover, these default shortcuts may be changed in the future, to align somehow with newly added shortcuts. And it will cause even more issues with your implementation.

tahamukhtar20:
No actually, the restore defaults doesn't use the localStorage, it just parse the localStorage data to delete the shortcuts stored so that the next time data is loaded it's not loading the previously stored shortcuts. 
```
                const shortcuts = useSelector((state: CombinedState) => state.shortcuts);
                const currentSettings = localStorage.getItem('clientSettings');
                dispatch(shortcutsActions.setShortcuts({
                    ...shortcuts,
                    keyMap: { ...shortcuts.defaultState },
                }));
                if (currentSettings) {
                    const parsedSettings = JSON.parse(currentSettings);
                    delete parsedSettings.shortcuts;
                    localStorage.setItem('clientSettings', JSON.stringify(parsedSettings));
                }
```

tahamukhtar20:
> I was able to install obviously conflicting shortcut by modifying local storage directly: image
> 
> Why is it a problem? Imagine a case:
> 
> 1. User customizes shortcut, let's say F3 to do something.
> 2. CVAT team introduces new feature, lying in the same scope, with the same default shortcut.
> 3. User gets conflicting shortcuts setup.
> 
> I believe, correct way here is to prioritize user-customized shortcuts (loaded from local storage), over CVAT default shortcuts in case of conflicts.
> 
> Please, cover this issue by test.

I'll check

bsekachev:
I tried to add dummy "new feature shortcut".
It only was added in private tab (right one). In usual mode (left) it does not display.

image

cvat-ui/src/utils/mousetrap-react.tsx
```diff
@@ -38,7 +40,14 @@ export default function GlobalHotKeys(props: Props): JSX.Element {
         for (const key of Object.keys(keyMap)) {
             const { sequences } = keyMap[key];
             const handler = handlers[key];
-            Mousetrap.bind(sequences, handler, 'keydown');
+            Mousetrap.bind(sequences, (event, combo) => {
+                const e = event || window.event;
```
bsekachev, line 44: image

cvat-ui/src/components/header/settings-modal/shortcut-settings.tsx
```diff
@@ -0,0 +1,157 @@
...
+import { KeyMap } from 'utils/mousetrap-react';
+import { Empty, Modal } from 'antd';
+import { shortcutsActions } from 'actions/shortcuts-actions';
+import { useDispatch, useSelector } from 'react-redux';
+import { CombinedState } from 'reducers';
+import MultipleShortcutsDisplay from './multiple-shortcuts-display';
+
+interface Props {
+    keyMap: KeyMap;
+    normalizedKeyMap: Record<string, string>;
```
bsekachev, line 29: image

bsekachev:
Try to press a button and then click outside of field before 1 second ends. Now you do not have shortcut and you can't remove this text from the field.

Expected behavioud in this case, if user clicks somewhere outside, we should immediately apply the sequence 

issue

bsekachev:
issue1

Double modal opens

bsekachev:
??? issue 2

bsekachev:
Click in the field, click outside the field. Empty sequence registered
Снимок экрана 2024-08-20 103347

bsekachev:
Still reproducable
 
image

tahamukhtar20:
Shouldn't be, I'll check

bsekachev:
Shortcuts like `f3 f3 f3` still not work as expected.
From my own research, you can achieve that by modifying `Mousetrap.prototype.stopCallback` for example

E.g.

```
Mousetrap.prototype.stopCallback = function (e: KeyboardEvent, element: Element, combo: string): boolean {
    const store = getCVATStore();
    const { keyMap } = store.getState().shortcuts;
    const shortcuts = [];
    const keys: string[] = ['SWITCH_SHORTCUTS', 'SWITCH_SETTINGS'];
    for (const key of keys) {
        shortcuts.push(...keyMap[key].sequences);
    }

    let comboStarted = false;
    if (shortcuts.some((shortcut) => shortcut.startsWith(combo))) {
        e?.preventDefault();
        comboStarted = true;
    }

    // stop when modals are opened
    const someModalsOpened = Array.from(
        window.document.getElementsByClassName('ant-modal'),
    ).some((el) => (el as HTMLElement).style.display !== 'none');
    if (someModalsOpened && !comboStarted) {
        return true;
    }

    // stop for input, select, and textarea
    return element.tagName === 'INPUT' ||
        element.tagName === 'SELECT' ||
        element.tagName === 'TEXTAREA';
};
```

Please, check your patch one more time.

bsekachev:
I see these conflicts when just open the tool:
image

These shortcuts do not really conflict to each other. Shall we add more scopes? 

E.g.
- Standard workspace controls
- Standard 3D workspace controls
- Review workspace controls

tahamukhtar20:
Perhaps, because we can only check conflicts between scopes atomically, it won't be a good practice to exclude just these conflicts when conflicts occur.

bsekachev:
There may be some issues with re-used components in this case, e.g. merge control. Used in both 2D and 3D but it is registered with:

```
const componentShortcuts = {
    SWITCH_MERGE_MODE: {
        name: 'Merge mode',
        description: 'Activate or deactivate mode to merging shapes',
        sequences: ['m'],
        scope: ShortcutScope.CONTROLS_SIDE_BAR,
    },
};
```

in `annotation-page/standard-workspace/controls-side-bar/merge-control.tsx`

Maybe it is a reason to refactor the code in this case and put to different places as two independend shortcuts, e.g.:

In `cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/controls-side-bar.tsx`


```
const componentShortcuts = {
    SWITCH_MERGE_MODE_2D: {
        name: 'Merge mode',
        description: 'Activate or deactivate mode to merging shapes',
        sequences: ['m'],
        scope: ShortcutScope.CONTROLS_SIDE_BAR_2D,
    },
};
```

And in `cvat-ui/src/components/annotation-page/standard3D-workspace/controls-side-bar/controls-side-bar.tsx`

```
const componentShortcuts = {
    SWITCH_MERGE_MODE_3D: {
        name: 'Merge mode',
        description: 'Activate or deactivate mode to merging shapes',
        sequences: ['m'],
        scope: ShortcutScope.CONTROLS_SIDE_BAR_3D,
    },
};
```

There are some others re-used in both 2D and 3D workspaces.

bsekachev:
Setup `f2` to SWITCH_SHORTCUTS in local storage. Now it works correct when UI opens, hovewer there is a weird error in console:

image

It literally says that the shortcut conflicts with all shortcuts?

tahamukhtar20:
No actually, the logging is a bit misleading, what it really means is that the shortcuts which I want to register, in this case, I pass the whole data from the localstore to the conflict detector, contain a conflict, I'll update the logging so that it gives only the shortcut which it conflicts with.

bsekachev:
> Shortcuts like f3 f3 f3 still not work (to open/close modal with settings for example) as expected.
From my own research, you can achieve that by modifying Mousetrap.prototype.stopCallback for example

Still waiting for the solution here, it is quite critical. Currently I understand that what I proposed caused some issues, but you were not able yet to sort out what exactly happens. So, there are three possible options:

1. To sort out what happened, maybe it was easy to fix
2. Propose another solution
3. To describe why supporting shortcuts like `f3 f3 f3` is not possible by technical reasons. Additionally, in this case it have to be well documented.

tahamukhtar20:
I'll let you know if I come across any solution, currently some seemingly random errors of some colour not being correct appear in the test cases, but will let you know if I find anything.

bsekachev:
By the way this column currently does not seem useful for a user. So, may be dropped I believe. 

image

bsekachev:
Tag annotation mode does not have any controls
image

For single shape annotation mode name and description are wrong
image

Drawing always starts automatically. Here N is only to finish drawing of a current shape. And corresponding description. 

Also, we do not need both N and Shift+N here. Please, left only N.

bsekachev:
Will we get any conflicts if move all from single shape workspace controls to single shape workspace? 

As in this case, again, this workspace does not have any control bar aside.  

image

bsekachev:
image

In this case correct option is Review Workspace Controls as we have a button to create an issue in sidebar.

tahamukhtar20:
> Will we get any conflicts if move all from single shape workspace controls to single shape workspace?
> 
> As in this case, again, this workspace does not have any control bar aside.
> 
> image

We will get a conflict of draw mode with create new issue in the review workspace I think

tahamukhtar20:
> image
> 
> In this case correct option is Review Workspace Controls as we have a button to create an issue in sidebar.

That will cause conflicts with the controls side bars that have Draw mode in them.

tahamukhtar20:
Since we have not names the scopes as control side bar, instead as controls, so I don't think anyone would think this is the control side bar, just the controls available in that workspace.

bsekachev:
Paste shape and copy shape should be probably in the same scope (objects sidebar).
I do not see any reason to put them in different. Could you please refactor?  Additionally,  in this case we do not need to duplicate `paste shape` in different scopes

bsekachev:
We will get a conflict of draw mode with create new issue in the review workspace I think

Single shape workspace and review controls should not conflict to each other at all. So, why do you expect conflict here?

tahamukhtar20:
Yes they shouldn't but you see that there are multiple switch draw modes, all of those are currently in their corresponding workspace controls, and the create issue is in the review workspace, currently, not in the controls of the review workspace. The problem is that if I move the switch draw modes to workspaces instead of controls, we can't have multiple draw modes. You're right about that single shape workspace and review controls will not conflict, but moving the draw mode to workspace will cause issues in between workspaces, as there are three draw modes if we remove the tag annotation one, single shape, standard and standard 3d.

bsekachev:
image

Please, normalize the name of the categories as proposed: names, lower and higher case characters, sorting.

bsekachev:
I am not really understand your idea. 

Any shortcuts from these spaces:
- Tag annotation workspace
- Single shape annotation workspace
- Review workspace controls
- Standard workspace controls 

Will not conflict to each other as they are never rendered simultaneously.

tahamukhtar20:
Wait, yes you are right, I think I explained incorrectly before. I'll update the scopes to workspace and from review to review controls.

bsekachev:
Try to setup F4+F6 shortcut. 
image

Now clicking only F6 opens the list of shortcuts

bsekachev:
There is an issue with customization of **Switch draw** shortcut.

Shift + N was initially used to remove selected shape and re-draw it from scratch. And it is being handled like this:

image

Hovewer in case of any customization this logic will not work. Probably we need to refactor by splitting into two different shortcuts.

bsekachev:
image
 By a reason, this shortcut does not display in settings

tahamukhtar20:
I'll check

tahamukhtar20:
Okay, just one minute I'll update.

bsekachev (approved):

bsekachev:
Great job!
Let's try to merge it now

PR #8192, opened 2024-07-18
Optimized database requests when removing resources

bsekachev:
### Checklist

### License

## Summary by CodeRabbit

- **Bug Fixes**
  - Optimized deletion process to ensure all related resources are deleted efficiently.
  - Improved model field relationships to prevent issues when deleting related objects.
  
- **Performance Improvements**
  - Enhanced methods for cascading deletions, reducing potential orphaned records and improving database integrity.

- **General Improvements**
  - Unified deletion methods for tasks and jobs to streamline resource management and reduce redundancy.
  - Ensured segments are appropriately deleted under specified conditions, maintaining data consistency.


cvat/apps/engine/utils.py
```diff
@@ -412,3 +412,7 @@ def directory_tree(path, max_depth=None) -> str:
 
 def is_dataset_export(request: HttpRequest) -> bool:
     return to_bool(request.query_params.get('save_images', False))
+
+def chunked_list(lst, chunk_size):
+    for i in range(0, len(lst), chunk_size):
+        yield lst[i:i + chunk_size]
```
Marishka17, line 418: Perhaps [take_by](https://github.com/cvat-ai/datumaro/blob/2a4d9dbbd86f2e5fc5f8db2cfd2defdf464e9645/datumaro/util/__init__.py#L60) could be used instead

cvat/apps/engine/views.py
```diff
@@ -1746,7 +1746,13 @@ def perform_destroy(self, instance):
         if instance.type != JobType.GROUND_TRUTH:
             raise ValidationError("Only ground truth jobs can be removed")
 
-        return super().perform_destroy(instance)
+        segment = instance.segment
+        if segment:
```
Marishka17, line 1750: In which case segment can be None?

cvat/apps/engine/models.py
```diff
@@ -325,6 +325,18 @@ class Meta:
     def touch(self) -> None:
         self.save(update_fields=["updated_date"])
 
+@transaction.atomic(savepoint=False)
+def clear_annotations_in_jobs(job_ids):
+    for job_ids_chunk in chunked_list(job_ids, chunk_size=1000):
```
Marishka17, line 330: Did you compare performance with different `chunk_size`? Why are you using 1000 here and 10000 in the `dataset_manager.task.py::JobAnnotation::_delete`?

cvat/apps/dataset_manager/task.py
```diff
@@ -437,24 +438,52 @@ def update(self, data):
         if not self._data_is_empty(self.data):
             self._set_updated_date()
 
+    def _delete_job_labeledimages(self, ids__UNSAFE):
```
Marishka17, line 441: Could you please add type annotations to the new functions?

Marishka17 (commented):
I have a few minor comments, but overall patch works for me.

cvat/apps/engine/views.py
bsekachev: I do not have any ideas here, but moved this logic as is from `Job.delete` method 
Maybe @zhiltsov-max may comment

zhiltsov-max: I don't think this should happen in practice, it looks more like a default safe removal pattern.

cvat/apps/engine/utils.py
bsekachev: I would prefer to use something internal here
Utility function in 3rdparty library may be renamed/removed and hardly ever it will affect major library version
Internal is just safer

cvat/apps/engine/models.py
bsekachev: I did not really compare performance between different chunks, just tried to find something more or less logical and avoid huge SQL requests. Actually I do not think performance between one request with 10000 INs significantly differs with two requests and 5000 INs each (probably the first one will be even quicker). 

In case with annotations, user can easily update > 10000 objects in one save request (e.g. when running an annotations action), splitting it by batches in 100 or 1000 objects, would not be efficient IMHO.  At the same time we can't keep the batch unlimited as the SQL request in general is limited in length for different database backends. 

I will test, btw how it works with many thousands objects .

A task with more than 1000 jobs is something extremely rare on practice and 1000 will be enough almost always to finish operation in one request. Why not set 10000 here? I do not have strong answer here, it just does not look necessary

cvat/apps/engine/views.py
Marishka17: In that case, If this should not happen in practice, does such handling look like silent error suppression? (for me - yes). But, @zhiltsov-max, why do we need to call segment.delete() directly here? It will be deleted anyway after job deletion.

cvat/apps/engine/utils.py
Marishka17: I agree, but since our team maintains datumaro that we use, it depends on us. + We already use this method in engine.

cvat/apps/engine/views.py
zhiltsov-max: In our implementation, segments own jobs, so they need to be deleted explicitly. I agree that it's currently meaningless to check that the segment is none, as it's prohibited by the field definition.

cvat/apps/engine/models.py
```diff
@@ -912,7 +936,7 @@ def __str__(self):
 
 class Annotation(models.Model):
     id = models.BigAutoField(primary_key=True)
-    job = models.ForeignKey(Job, on_delete=models.CASCADE)
+    job = models.ForeignKey(Job, on_delete=models.DO_NOTHING)
```
zhiltsov-max, line 939: From [django docs](https://docs.djangoproject.com/en/4.2/ref/models/fields/#django.db.models.DO_NOTHING), this may leave objects in an inconsistent state. Could you explain why automatic removal on the DB side didn't work for us? Maybe we should just configure `on_delete` manually instead?

bsekachev: > Could you explain why automatic removal on the DB side didn't work for us? 

I explained on the meeting. The first point is that will not work with different backends, especially when foreign keys already configured.

Postgres already have setup constraint for ForeignKey fields, and this constraint:

1. Does not allow to remove a row with other rows which refer to this one. So, data will stay consistent. 
2. Goes first, so, we may configure on_delete on level of the database, but it will not work as the first ForeignKey constraint will just abort the operation

bsekachev: Try to create a task with a label with an attribute, annotate object and remove LabeledShape without removing before LabeledShapeAttributeVal. You will get an error raised from Postgres.

cvat/apps/engine/utils.py
bsekachev: Until that is not a part of API, it is not safe. 

> We already use this method in engine

That does not mean that such using is correct.

bsekachev: Being a part of API means that these methods are at least documented.

cvat/apps/engine/views.py
bsekachev: Returned this code to model, I believe that would be a little bit safer if we remove job from admin panel for example

cvat/apps/engine/models.py
bsekachev: Decreased both to 1000.

bsekachev:
@Marishka17 @zhiltsov-max Any more comments?

cvat/apps/dataset_manager/task.py
```diff
@@ -437,24 +438,52 @@ def update(self, data):
...
             self._set_updated_date()
 
+    def _delete_job_labeledimages(self, ids__UNSAFE: list[int]) -> None:
+        # ids__UNSAFE is a list, received from the user
+        # we MUST filter it by job_id additionally before applying to any queries
+        ids = self.db_job.labeledimage_set.filter(pk__in=ids__UNSAFE).values_list('id', flat=True)
+        models.LabeledImageAttributeVal.objects.filter(image_id__in=ids).delete()
+        self.db_job.labeledimage_set.filter(pk__in=ids).delete()
+
+    def _delete_job_labeledshapes(self, ids__UNSAFE: list[int], is_subcall=False) -> None:
```
zhiltsov-max, line 448:
```suggestion
    def _delete_job_labeledshapes(self, ids__UNSAFE: list[int], *, is_subcall: bool = False) -> None:
```

cvat/apps/dataset_manager/task.py
```diff
@@ -437,24 +438,52 @@ def update(self, data):
...
+        else:
+            ids = self.db_job.labeledshape_set.filter(pk__in=ids__UNSAFE).values_list('id', flat=True)
+            child_ids = self.db_job.labeledshape_set.filter(parent_id__in=ids).values_list('id', flat=True)
+            if len(child_ids):
+                self._delete_job_labeledshapes(child_ids, True)
+
+        models.LabeledShapeAttributeVal.objects.filter(shape_id__in=ids).delete()
+        self.db_job.labeledshape_set.filter(pk__in=ids).delete()
+
+    def _delete_job_labeledtracks(self, ids__UNSAFE: list[int], is_subcall=False) -> None:
```
zhiltsov-max, line 462:
```suggestion
    def _delete_job_labeledtracks(self, ids__UNSAFE: list[int], *, is_subcall: bool = False) -> None:
```

cvat/apps/dataset_manager/task.py
bsekachev: Why do you want to add `*` here? Number of arguments is known in this method

Marishka17 (approved):

PR #8198, opened 2024-07-18
Fix export and export cleanup job hangs in scheduled state

zhiltsov-max:
### Motivation and context

There are 2 schedulers supported by django_rq and by python RQ: `rq_scheduler` and a newer, builtin queue scheduler in RQ. rq_scheduler seems to die slowly in favor of the builtin scheduler. The schedulers have compatible API, but not the implementation. The existing job retry implementation relies on `retry()` calls, which, in turn, rely on the builtin RQ scheduler. CVAT uses rq_scheduler a for some tasks, so it its executed. The builtin RQ scheduler needs the `--with-scheduler` startup parameter on the worker processes. Thus, the jobs were hanging in the scheduled state, as the builtin RQ scheduler was not running on the queues. As CVAT is currently using rq_scheduler, it's decided to continue using it to avoid disruption and use of 2 schedulers together. The implementation in this PR does best efforts to be correct, but it's has potential problems with multiple same jobs running in parallel.

In future we need to migrate to the builtin RQ scheduler, as it is the only one maintained as of February 2023.

### Checklist

### License

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved export and export cache clean operations by adding a retry mechanism to handle job retries, preventing hangs.

- **Chores**
  - Updated internal process for handling job retries using RQ scheduler for better reliability.

### DIFF
--- a/cvat/apps/dataset_manager/views.py
+++ b/cvat/apps/dataset_manager/views.py
@@ -13,11 +13,13 @@ import django_rq
 import rq
 from django.conf import settings
 from django.utils import timezone
+from rq_scheduler import Scheduler
 
 import cvat.apps.dataset_manager.project as project
 import cvat.apps.dataset_manager.task as task
 from cvat.apps.engine.log import ServerLogManager
 from cvat.apps.engine.models import Job, Project, Task
+from cvat.apps.engine.utils import get_rq_lock_by_user
 
 from .formats.registry import EXPORT_FORMATS, IMPORT_FORMATS
 from .util import (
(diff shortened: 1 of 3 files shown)


azhavoro (approved):

cvat/apps/dataset_manager/views.py
```diff
@@ -58,6 +60,42 @@ def get_export_cache_ttl(db_instance: str | Project | Task | Job) -> timedelta:
...
+
+        with get_rq_lock_by_user(settings.CVAT_QUEUES.EXPORT_DATA.value, user_id):
+            scheduler.enqueue_in(
+                time_delta,
+                current_rq_job.func,
+                *current_rq_job.args,
+                **current_rq_job.kwargs,
+                job_id=current_rq_job.id,
+                meta=current_rq_job.meta,
+                depends_on=current_rq_job.dependency_ids,
```
Marishka17, line 88: @zhiltsov-max, Probably, there are several bugs:
1) `dependency_ids` can be outdated at the moment when this job will be enqueued, so one user can clog a queue with his jobs (despite `ONE_RUNNING_JOB_IN_QUEUE_PER_USER`==True)
2) let's imagine a situation when:
there were 3 rq jobs: `J1` <- `J2(allow_dependency_failures=True)` <- `J3(allow_dependency_failures=True)`
`J1` was executed successfully -> `J2` was started -> `J1` was deleted by CVAT code after downloading a file -> `J2` was scheduled with allow_dependency_failures = None (because it is not copied in the current implementation), -> `J1` was recreated after one more request from a user and now `J1` depends on the `J3`. What will be at the end? Will J3 be enqueued after `J2` is placed to the scheduled job registry?

PR #8216, opened 2024-07-24
Add REST API tests for /requests API && test both versions of the export API

Marishka17:
### Motivation and context

This PR fixes the following issues:
- [export API v1] do not reinitialize dataset export process when downloading a result file if a resource (project|task|job) has been updated since the first initialized export request
- [export API v1] return `rq_id` for all requests with 202 status code (not only for initialization requests)
- [requests API] Fixed filtering by format && added resource to allowed filters

REST API tests updates:
- Added tests to check requests filtration using simple filters
- Added tests to check specific requests retrieving
- Updated all tests that export project|task|job datasets|annotations|backups:
  - to test both API versions (including API mixing)
  - to use only appropriate resources by checking the default export location
- Added fixtures to filter projects/tasks assets
- Updated default target|source buckets to `import/export` bucket to exclude the same bucket usage as a data source in several tests (when bucket content is used as task data) and as a bucket for results
### Checklist

### License

## Summary by CodeRabbit

- **New Features**
	- Enhanced job handling for exports, improving error management and job state tracking.
	- Introduced a new `resource` field in the request handling system to improve data categorization.
	- Added new filtering capabilities for API queries, allowing users to filter by the `resource` field.

- **Bug Fixes**
	- Improved status checks and handling for job requests.
	- Introduced exception handling for forbidden access during project backup attempts.

- **Tests**
	- Refactored test suites to improve coverage and ensure compatibility across versions with new methods and exception handling.
	- New tests added to validate request handling functionality.


tests/python/rest_api/test_jobs.py
```diff
@@ -1321,47 +1327,82 @@ def _check_cvat_for_video_job_annotations(content, values_to_be_checked):
...
             )
         )
-        response = self._export_annotations(username, job_id)
-        assert response.data
 
+        self._test_export_annotations(
+            username, job["id"], api_version=api_version, local_download=self.is_local_download(job)
+        )
+
+    @pytest.mark.parametrize("api_version", (1, 2))
```
zhiltsov-max, line 1405: Is it possible to mix these APIs?

tests/python/rest_api/test_projects.py
```diff
@@ -597,8 +703,9 @@ def test_can_import_dataset_in_org(self, admin_user):
             ("Datumaro 1.0", "Datumaro 1.0"),
         ),
     )
+    @pytest.mark.parametrize("api_version", (1, 2))
     def test_can_export_and_import_dataset_with_skeletons(
```
zhiltsov-max, line 707: It doesn't seem to be a test that should be related to this update.

tests/python/rest_api/test_projects.py
```diff
@@ -708,20 +800,31 @@ def test_exported_project_dataset_structure(
...
 
-        response = self._export_annotations(username, pid, format=anno_format)
-        assert response.data
-        with zipfile.ZipFile(BytesIO(response.data)) as zip_file:
+        dataset = self._test_export_annotations(
+            username,
+            pid,
+            api_version=api_version,
+            format=anno_format,
+            local_download=(not (project["target_storage"] or {}).get("cloud_storage_id")),
```
zhiltsov-max, line 808: It feels like local downloading should be a test parameter in a separate test and just set to True in all these tests.

tests/python/rest_api/test_requests.py
```diff
@@ -0,0 +1,314 @@
...
+        self.user = find_users(privilege="user")[0]["username"]
+
+    def create_resources(self):
+        self.project_ids = self.create_projects()
+        self.task_ids = self.create_tasks()
+        self.job_ids = []
+        for tid in self.task_ids:
+            self.job_ids.extend(self.get_job_ids(tid))
+
+    def make_requests(self):
```
zhiltsov-max, line 69: Probably, it should be a fixture.

tests/python/rest_api/test_requests.py
```diff
@@ -0,0 +1,314 @@
...
+def validate_id(filter_: str, filter_value: int, requests: List[models.Request]) -> bool:
+    return {filter_value} == {getattr(r.operation, filter_) for r in requests}
+
+
+def validate_resource(filter_: str, filter_value: str, requests: List[models.Request]) -> bool:
+    return {filter_value} == {r.operation.target.value for r in requests}
+
+
+def validate_format(filter_: str, filter_value: str, requests: List[models.Request]) -> bool:
+    return {filter_value} == {r.operation.format for r in requests}
```
zhiltsov-max, line 48: Probably, these functions are supposed to be class members in the list testing class.

tests/python/rest_api/test_requests.py
```diff
@@ -0,0 +1,314 @@
...
+
+
+@pytest.mark.usefixtures("restore_db_per_class")
+@pytest.mark.usefixtures("restore_redis_inmem_per_function")
+@pytest.mark.timeout(30)
+class TestListRequests:
```
zhiltsov-max, line 54: Please check if `CollectionSimpleFilterTestBase` can be of any help here.

tests/python/rest_api/test_jobs.py
Marishka17: [Done](https://github.com/cvat-ai/cvat/pull/8216/commits/bf726bbd33112137e7985d4f34cbd973d76b6243)

tests/python/rest_api/test_projects.py
Marishka17: You are right, fixed

tests/python/rest_api/test_requests.py
Marishka17: Replaced this logic by using `CollectionSimpleFilterTestBase`

cvat/apps/engine/background.py
```diff
@@ -270,11 +251,44 @@ def _handle_rq_job_v1(
...
+        if action not in {None, "", "download"}:
+            raise serializers.ValidationError(
+                f"Unexpected action {action!r} specified for the request"
+            )
+
+        if not rq_job:
+            return None if action != "download" else \
+                HttpResponseBadRequest(
+                    "Result file cannot be send since there is no related job on the server. "
+                    "Initialize background process first without sending action=download parameter."
```
zhiltsov-max, line 265:
```suggestion
                    "Unknown export request id. "
                    "Please request export first by sending a request without the action=download parameter."
```
I remember we had an agreement that users should not know anything about background processes.

cvat/apps/engine/background.py
```diff
@@ -320,15 +330,13 @@ def _handle_rq_job_v1(
             # Such dependencies are never removed or finished,
             # as there is no TTL for deferred jobs,
             # so the current job can be blocked indefinitely.
+            cancel_and_reenqueue(rq_job)
+
+        if is_result_outdated():
+            cancel_and_reenqueue(rq_job)
+
+        return Response(status=status.HTTP_202_ACCEPTED)
```
zhiltsov-max, line 338: Probably, all 202 should have the same output format. Currently, there are cases with and without data in the body.

Marishka17:
@zhiltsov-max, Any more comments?

zhiltsov-max:
My tests seem to be working now.

cvat/apps/engine/background.py
```diff
@@ -80,24 +82,26 @@ def setup_background_job(self, queue: DjangoRQ, rq_id: str) -> None:
...
-        )
+        rq_job_status = rq_job.get_status(refresh=False)
+
+        if rq_job_status in {RQJobStatus.STARTED, RQJobStatus.QUEUED}:
+            return Response(
+                data=f"Export process is already {rq_job_status}",
+                status=status.HTTP_409_CONFLICT,
+            )
+
+        if rq_job_status == RQJobStatus.SCHEDULED:
```
Marishka17, line 100:
```suggestion
        if rq_job_status in (RQJobStatus.SCHEDULED, RQJobStatus.DEFERRED):
```

tests/python/rest_api/utils.py
```diff
@@ -68,9 +412,14 @@ def _get_field(cls, d: Dict[str, Any], path: Union[str, FieldPath]) -> Optional[
         assert path
         for key in path:
             if isinstance(d, dict):
+                assert isinstance(key, str)
                 d = d.get(key)
             else:
-                d = None
+                if callable(key):
```
SpecLad, line 418: I think the definition of `FieldPath` needs to be amended to account for this.

tests/python/shared/fixtures/data.py
```diff
@@ -60,6 +62,50 @@ def tasks():
...
+def filter_assets(resources: Iterable, **kwargs):
+    filtered_resources = []
+    exclude_prefix = "exclude_"
+
+    for resource in resources:
+        for key, value in kwargs.items():
+            if exclude := key.startswith(exclude_prefix):
+                key = key[len(exclude_prefix) :]
+
+            op = operator.ne if exclude else operator.eq
```
SpecLad, line 74: This is a confusing use of the `:=` operator, because the variable is not used in the `if` body (but is used outside of it). I would suggest this instead:

```suggestion
            op = operator.eq
            
            if key.startswith(exclude_prefix):
                key = key[len(exclude_prefix) :]
                op = operator.ne
```

cvat/schema.yml
```diff
@@ -4576,6 +4576,15 @@ paths:
         description: A simple equality filter for the project_id field
         schema:
           type: integer
+      - name: resource
```
SpecLad, line 4579: Why is it called "resource", when it seems to correspond to the `target` field?

tests/python/rest_api/utils.py
Marishka17: It has already been updated https://github.com/cvat-ai/cvat/blob/1f37236e09844f97716ba147595fcd3c6d30055f/tests/python/rest_api/utils.py#L392

cvat/schema.yml
Marishka17: Hm, probably I thought that using `resource` and `subresource` is better than `target` and `subresource`

Marishka17:
@zhiltsov-max, @SpecLad  
I've added using locks when handling and setting up RQ jobs to eliminate race conditions and make behavior more predictable (since in the current implementation we can modify rq job (delete/cancel/create a new one with the same id)). There is a comparison of 3 cases (other cases were not compared):

**Before fix:**
1. _[API v1] finished|failed X1 job, action=download_
    request 1: deletes X1, returns results file or exception details
    request 2: deletes* X1 (idempotent action), returns results file or exception details

2. _[API v1] finished X1 job, no action==download, result is outdated (request timestamp from meta < last updated time)_
    request 1: 
      - X1.delete() -> deletes X1 from the `FinishedJobRegistry` and Redis,
      - X1.cancel() -> enqueues X1 dependents, addes X1 to `CancelledJobRegistry`
      - then creates a new X1
      
    request 2 with small delay: deletes X1 from `CancelledJobRegistry`, creates new X1* job, but that X1* will depend on X1 from 1st request, because X1 was placed to export queue or to `StartedJobRegistry`. That means that X1 will be present in 2 registries (started and deferred), will be "dependent on itself" and will be executed twice.

3. _[API v2] finished|failed X1 job_
    request 1: deletes X1, creates new one, returns 202
    request 2: similar to the situation described above: X1 job exists in started and deferred registries, is executed twice
                          
**After fix:**
1. _[API v1] finished|failed X1 job, action=download_
    request 1: deletes X1, returns results file or exception details
    request 2: return 400, Unknown export request id

2. _[API v1] finished X1 job, no action==download, result is outdated (request timestamp from meta < last updated time)_
    request 1: the same as listed above
    request 2 with small delay: returns a response based on updated rq job status.

3. _[API v2] finished|failed X1 job_
    request 1: deletes X1, creates new one, returns 202
    request 2: return 409

tests/python/rest_api/utils.py
SpecLad: My bad, must've missed it.

cvat/schema.yml
SpecLad: Well... it would've been better, had it been that way from the start. But the field is already called `target`, so it doesn't really make sense for the corresponding filter to be `resource`.

cvat/apps/engine/background.py
```diff
@@ -601,11 +630,12 @@ def export(self) -> Response:
         )
         rq_job = queue.fetch_job(rq_id)
 
-        with suppress(ReenqueueJobException):
+        # ensure that there is no race condition when processing parallel requests
+        with get_rq_lock_by_job(rq_job), suppress(ReenqueueJobException):
```
zhiltsov-max, line 634: Isn't it better to acquire the lock before the job is fetched? I think it's not guaranteed that job data in the fetched object is actual after the lock is acquired.

cvat/apps/engine/background.py
```diff
@@ -144,12 +146,14 @@ def get_timestamp(self, time_: datetime) -> str:
...
+def recreate(rq_job: RQJob, should_be_canceled: bool = True) -> None:
+    # Cancel and delete a job and then raise ReenqueueJobException
+    # to indicate that the job should be re-created
+    if should_be_canceled:
+        # In the case the server is configured with ONE_RUNNING_JOB_IN_QUEUE_PER_USER
+        # we have to enqueue dependent jobs after canceling one.
+        rq_job.cancel(enqueue_dependents=settings.ONE_RUNNING_JOB_IN_QUEUE_PER_USER)
+
     rq_job.delete()
     raise ReenqueueJobException()
```
zhiltsov-max, line 158: - Using exceptions to control execution flow is typically considered an antipattern 
- I can't find any places where this exception is actually used, in all cases it's being suppressed

cvat/apps/engine/background.py
```diff
@@ -13,6 +13,7 @@
 
 import django_rq
 from attrs.converters import to_bool
+from contextlib import suppress
```
zhiltsov-max, line 16: Consider adding these new files into `dev/format_python_code.sh`

cvat/apps/engine/background.py
```diff
@@ -361,11 +377,11 @@ def export(self) -> Response:
 
         rq_job = queue.fetch_job(rq_id)
 
-        with suppress(ReenqueueJobException):
+        # ensure that there is no race condition when processing parallel requests
+        with get_rq_lock_by_job(rq_job), suppress(ReenqueueJobException):
```
zhiltsov-max, line 381: The same comment about the locking order.

cvat/apps/engine/background.py
```diff
@@ -80,24 +84,26 @@ def setup_background_job(self, queue: DjangoRQ, rq_id: str) -> None:
...
 
-        return Response(
-            data=f"Export process is already {'started' if rq_job_status == RQJobStatus.STARTED else 'queued'}",
-            status=status.HTTP_409_CONFLICT,
-        )
+        rq_job_status = rq_job.get_status()
+
+        if rq_job_status in {RQJobStatus.STARTED, RQJobStatus.QUEUED}:
+            return Response(
+                data=f"Export process is already {rq_job_status}",
```
zhiltsov-max, line 98:
```suggestion
                data=f"Export process is already started",
```
I think this difference between started and queued is an implementation detail, that should be encapsulated.

cvat/apps/engine/background.py
Marishka17: I missed that, thanks

zhiltsov-max (approved):

SpecLad (approved):
Needs a rebase, but otherwise LGTM.

PR #8223, opened 2024-07-25
Reduce import memory consumption

Bobronium:
#### memray stats for develop
Total time: 00:10:23
Peak memory usage: 4.2 GiB

#### memray stats for this branch
Total time: 00:05:55
Peak memory usage: 1.3 GiB

Check out [flamegraphs.zip](https://github.com/user-attachments/files/16380505/flamegraphs.zip) for full report.

Not yet entirely sure changes in 89659961af1720c015550521989bad50feaf5559 are sufficient. Was operating on the assumption that other importers are converting points to floats on import. Can confirm that this is the case for datumaro format.

### Checklist

### License

## Summary by CodeRabbit

- **New Features**
  - Introduced a new `LazyList` class for efficient data handling and lazy evaluation of elements.
  
- **Improvements**
  - Enhanced numerical accuracy in shape point processing by converting string attributes to float types during loading.
  - Updated the deletion process in the dataset manager for more efficient handling of annotations and data updates.
  - Modified the annotations handling function to allow for optional deep copying, improving performance flexibility.

- **Bug Fixes**
  - Corrected data type handling for shape points to prevent potential type-related errors during processing.


cvat/apps/events/handlers.py
```diff
@@ -373,8 +373,8 @@ def handle_delete(scope, instance, store_in_deletion_cache=False, **kwargs):
         user_email=uemail,
     )
 
-def handle_annotations_change(instance, annotations, action, **kwargs):
-    _annotations = deepcopy(annotations)
+def handle_annotations_change(instance, annotations, action, should_deepcopy=True, **kwargs):
```
zhiltsov-max, line 376: I think this breaks encapsulation of the function, if a parameter with such name is required to be specified by the client code. I can see several ways to improve it:
- rename to "annotations_mutable" or add a kw-parameter "mutable_annotations"
- infer it automatically inside the function, if the operation is "delete". The problem here is this code is not responsible for this logic. The same problem is with the implemented passing of `should_deepcopy=False` in the `delete()` call - it looks like it actually must be forwarded all the way from the place the class instance is created to be fully correct.

cvat/apps/events/handlers.py
```diff
@@ -373,8 +373,8 @@ def handle_delete(scope, instance, store_in_deletion_cache=False, **kwargs):
...
+    _annotations = deepcopy(annotations) if should_deepcopy else annotations
```
bsekachev, line 377: In my opinion we are trying to put into event too many information in this function, maybe that is worthy of discussing

cvat/apps/events/handlers.py
bsekachev: I think the function may be just refactored to avoid modifying its arguments

cvat/apps/events/handlers.py
```diff
@@ -374,19 +373,22 @@ def handle_delete(scope, instance, store_in_deletion_cache=False, **kwargs):
...
         }
 
-        label_id = shape.get("label_id", None)
+        label_id = data.get("label_id", None)
         if label_id:
-            data["label_id"] = label_id
+            filtered_data["label_id"] = label_id
 
-        return data
+        shapes = data.get("shates", None)
```
bsekachev, line 387: shates?

cvat/apps/events/handlers.py
```diff
@@ -440,13 +442,9 @@ def filter_shape_data(shape):
...
     tracks_by_type = {shape_type[0]: [] for shape_type in ShapeType.choices()}
-    for track in _annotations.get("tracks", []):
-        track_shapes = track.pop("shapes")
-        track = filter_shape_data(track)
-        track["shapes"] = []
-        for track_shape in track_shapes:
-            track["shapes"].append(filter_shape_data(track_shape))
-        tracks_by_type[track_shapes[0]["type"]].append(track)
+    for track in annotations.get("tracks", []):
+        filtered_track = filter_data(track)
```
bsekachev, line 446: Maybe it would be safer by concept to split the logic into several functions:
- `filter_track`
- `filter_tag`
- `filter_shape`

As these 3 data structures are different by their idea.

And left one more common `filter_data` copying (id, attributes, frame as common properties). 
But that is all about refactoring

cvat/apps/events/handlers.py
zhiltsov-max: So now it still will be copying, but without calling `deepcopy`. Do we really need to send all these annotations, can it be replaced by some summary? Can we modify the recipient side to get only required fields?

Bobronium: > So now it still will be copying, but without calling deepcopy

The main difference here is that we're only copying outer layers of the data, while nested fields remain the same objects. I'd say this has negligible impact on memory/performance, especially compared to previous implementation. 

> Do we really need to send all these annotations, can it be replaced by some summary? Can we modify the recipient side to get only required fields?

While I think it certainly worth a discussion, but I don't have enough context to answer these questions yet.

cvat/apps/events/handlers.py
Bobronium: Agreed. I've added `filter_track` and `filter_shape`, adding `filter_tag` didn't make sense since it would be identical to `filter_data`:

```python
def filter_tag(tag):
    return filter_data(tag)
 ```

Bobronium:
Closing in favour of 3 separate PRs:
2. Array fields optimization (https://github.com/cvat-ai/cvat/pull/8229)
3. Logging function optimization (https://github.com/cvat-ai/cvat/pull/8228)

PR #8228, opened 2024-07-29
Remove deepcopy from handle_annotations_change

Bobronium:
### Motivation and context
Decided to split changes in this PR: https://github.com/cvat-ai/cvat/pull/8223

2. Array fields optimization (https://github.com/cvat-ai/cvat/pull/8229)
3. Logging function optimization (this PR)

### Checklist

### License

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved data integrity by ensuring annotations change handling occurs every time an item is deleted.

- **New Features**
	- Enhanced data filtering logic for annotations, allowing for more efficient processing without unnecessary deep copies.
	- Introduced a new function for streamlined filtering of track data, simplifying data management.

- **Refactor**
	- Consolidated data filtering into reusable functions for better maintainability and clarity.

### DIFF
--- a/cvat/apps/dataset_manager/task.py
+++ b/cvat/apps/dataset_manager/task.py
@@ -477,10 +477,10 @@ class JobAnnotation:
 
     def delete(self, data=None):
         deleted_data = self._delete(data)
-        handle_annotations_change(self.db_job, deleted_data, "delete")
-
         if not self._data_is_empty(deleted_data):
             self._set_updated_date()
+
+        handle_annotations_change(self.db_job, deleted_data, "delete")
 
     @staticmethod
     def _extend_attributes(attributeval_set, default_attribute_values):
--- a/cvat/apps/events/handlers.py
+++ b/cvat/apps/events/handlers.py
@@ -4,7 +4,6 @@
 
 import datetime
 import traceback
-from copy import deepcopy
 from typing import Optional, Union
 
 import rq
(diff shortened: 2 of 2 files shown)


cvat/apps/events/handlers.py
```diff
@@ -440,13 +441,9 @@ def filter_shape_data(shape):
...
-    for track in _annotations.get("tracks", []):
-        track_shapes = track.pop("shapes")
-        track = filter_shape_data(track)
-        track["shapes"] = []
-        for track_shape in track_shapes:
-            track["shapes"].append(filter_shape_data(track_shape))
-        tracks_by_type[track_shapes[0]["type"]].append(track)
+    for track in annotations.get("tracks", []):
+        filtered_track = filter_track(track)
+        tracks_by_type[track["shapes"][0]["type"]].append(filtered_track)
```
bsekachev, line 446: Not related to this PR, but I think there may be server error when track["shapes"] is an empty array 
In theory that is possible.

cvat/apps/dataset_manager/task.py
```diff
@@ -477,11 +477,11 @@ def _delete(self, data=None):
 
     def delete(self, data=None):
         deleted_data = self._delete(data)
-        handle_annotations_change(self.db_job, deleted_data, "delete")
-
         if not self._data_is_empty(deleted_data):
             self._set_updated_date()
 
+        handle_annotations_change(self.db_job, deleted_data, "delete")
```
bsekachev, line 483: Also not related to the PR, but we should fix it in the future as it looks critical:

We should not call `handle_annotations_change` within a postgres transaction block

E.g:
patch_job_data -> create -> handle_annotations_change
patch_job_data -> update -> handle_annotations_change
patch_job_data -> delete -> handle_annotations_change

As patch_job_data is performed in atomic transaction and all time it keeps connection opened to the database and locks some rows in the database. 

At the same time `handle_annotations_change` interacts with another database (Clickhouse)

bsekachev (approved):

cvat/apps/dataset_manager/task.py
zhiltsov-max: Basically, that's what transactions are created for - to roll back the whole operation, if it fails at any point. Need to think more abut it.

zhiltsov-max (approved):

cvat/apps/dataset_manager/task.py
bsekachev: With such logic we may just put all endpoints into transaction. 
Obviously, it will not work.

Transactions only for data consistency inside the database. 
That should not be responsible for external services.

zhiltsov-max: > With such logic we may just put all endpoints into transaction.
> Obviously, it will not work.

Actually, it's [one of the widespread approaches](https://docs.djangoproject.com/en/5.0/topics/db/transactions/#tying-transactions-to-http-requests). The only concern here is the performance, but it just means we have to avoid locks by using lock-free approaches and that long-running operations should be factored out into async calls.

PR #8229, opened 2024-07-29
Lazy list for array field

Bobronium:
### Motivation and context
Decided to split changes in this PR: https://github.com/cvat-ai/cvat/pull/8223

2. Array fields optimization (this PR)
3. Logging function optimization (https://github.com/cvat-ai/cvat/pull/8228)

### Checklist

### License

## Summary by CodeRabbit

- **New Features**
	- Introduced the `LazyList` class for efficient, on-demand parsing of list elements from strings.
	- Added support for custom transformations through a converter function.
	- Enhanced lazy evaluation with new decorators for improved performance on list operations.
  
- **Tests**
	- Implemented a comprehensive test suite for the `LazyList` class, validating core functionalities and ensuring robustness.

### DIFF
--- a/cvat/apps/engine/lazy_list.py
+++ b/cvat/apps/engine/lazy_list.py
@@ -1,3 +1,7 @@
+# Copyright (C) 2024 CVAT.ai Corporation
+#
+# SPDX-License-Identifier: MIT
+
 from functools import wraps
 from itertools import islice
 from typing import Any
--- a/cvat/apps/engine/lazy_list.py
+++ b/cvat/apps/engine/lazy_list.py
@@ -1,0 +1,254 @@
+from functools import wraps
+from itertools import islice
+from typing import Any
+from typing import Callable
+from typing import Iterator
+
+from typing import TypeVar
+from typing import overload
+
+
+T = TypeVar("T", bound=int | float | str)
+
+
+def _parse_self_and_other_before_accessing(list_method: Callable[..., Any]) -> Callable[..., Any]:
+    @wraps(list_method)
+    def wrapper(self: 'LazyList', other: Any) -> 'LazyList':
+        self._parse_up_to(-1)
+        if isinstance(other, LazyList):
+            other._parse_up_to(-1)
+        if not isinstance(other, list):
+            # explicitly calling list.__add__ with
+            # np.ndarray raises TypeError instead of it returning NotImplemented
+            # this prevents python from executing np.ndarray.__radd__
+            return NotImplemented
+
+        return list_method(self, other)
+
+    return wrapper
+
+
+def _parse_self_before_accessing(list_method: Callable[..., Any]) -> Callable[..., Any]:
+    """Wrapper for original list methods. Forces LazyList to parse itself before accessing them."""
+    @wraps(list_method)
+    def wrapper(self: 'LazyList', *args, **kwargs) -> 'LazyList':
+        self._parse_up_to(-1)
+
+        return list_method(self, *args, **kwargs)
+
+    return wrapper
+
+
+class LazyListMeta(type):
+    def __new__(
+        mcs,
+        name: str,
+        bases: tuple[type, ...],
+        namespace: dict[str, Any],
+    ):
+        # add pre-parse for list methods
+        for method_name in [
+            "append",
+            "copy",
+            "insert",
+            "pop",
+            "remove",
+            "reverse",
+            "sort",
+            "clear",
+            "index",
+            "count",
+            "__setitem__",
+            "__delitem__",
+            "__contains__",
+            "__len__",
+            "__reversed__",
+            "__mul__",
+            "__rmul__",
+            "__imul__",
+        ]:
+            namespace[method_name] = _parse_self_before_accessing(
+                getattr(list, method_name)
+            )
+
+        for method_name in [
+            "extend",
+            "__add__",
+            "__iadd__",
+            "__eq__",
+            "__gt__",
+            "__ge__",
+            "__lt__",
+            "__le__",
+        ]:
+            namespace[method_name] = _parse_self_and_other_before_accessing(
+                getattr(list, method_name)
+            )
+
+        return super().__new__(mcs, name, bases, namespace)
+
+
+class LazyList(list[T], metaclass=LazyListMeta):
+    """
+    Evaluates elements from the string representation as needed.
+    Lazy evaluation is supported for __getitem__ and __iter__ methods.
+    Using any other method will result in parsing the whole string.
+    Once instance of LazyList is fully parsed (either by accessing list methods
+    or by iterating over all elements), it will behave just as a regular python list.
+    """
+    __slots__ = ("_string", "_separator", "_converter", "_probable_length", "_parsed")
+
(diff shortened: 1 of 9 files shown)


cvat/apps/engine/models.py
```diff
@@ -181,6 +183,242 @@ def choices(cls):
...
+    _converter: Callable[[str], T]
+    _probable_length: int | None
+
+    def __init__(self, string: str = "", separator: str = ",", converter: Callable[[str], T] = lambda s: s) -> None:
+        super().__init__()
+        self.string = string
+        self._separator = separator
+        self._converter = converter
+        self._probable_length = None
+        self.parsed = False
```
zhiltsov-max, line 238: I think, you can reduce the amount of code here by using `attrs.define` or `dataclass`.

cvat/apps/engine/models.py
```diff
@@ -193,19 +431,20 @@ def __init__(self, *args, store_sorted:Optional[bool]=False, unique_values:Optio
...
 
     def to_python(self, value):
-        if isinstance(value, list):
+        if isinstance(value, list | LazyList):
             return value
 
         return self.from_db_value(value, None, None)
 
     def get_prep_value(self, value):
+        if isinstance(value, LazyList) and not value.parsed and not (self._unique_values or self._store_sorted):
```
zhiltsov-max, line 443: Probably, this access to private fields should be redesigned somehow.

cvat/apps/engine/models.py
```diff
@@ -181,6 +183,242 @@ def choices(cls):
...
+        "__add__",
+        "__eq__",
+        "__iadd__",
+        "__gt__",
+        "__ge__",
+        "__lt__",
+        "__le__",
+        "__eq__",
+    ]:
+        locals()[method] = _parse_both_before_accessing(getattr(list, method))
```
zhiltsov-max, line 417: Modifying locals [is UB](https://docs.python.org/3/library/functions.html#locals) in python. Typically, it can be replaced either by overriding specific methods, `__getattribute__`, or a metaclass. One more way is to create the class in a function, which can also write some fields later.

cvat/apps/engine/models.py
```diff
@@ -181,6 +183,242 @@ def choices(cls):
...
+        parsed_elements = list.__repr__(self).removesuffix("]")
+        unparsed_elements = self.string[current_position:]
+        return (
+            f"LazyList({parsed_elements}... + {unparsed_elements}', "
+            f"({list.__len__(self) / self._compute_max_length(self.string) * 100:.02f}% parsed))"
+        )
+
+    def __deepcopy__(self, memodict: Any = None) -> list[T]:
+        """
+        Since our elements are scalar, this should be sufficient
```
zhiltsov-max, line 259: From the current class definition, T is arbitrary.

cvat/apps/engine/models.py
```diff
@@ -181,6 +183,242 @@ def choices(cls):
...
+        "reverse",
+        "sort",
+        "clear",
+        "index",
+        "count",
+        "__setitem__",
+        "__delitem__",
+        "__contains__",
+        "__len__",
+        "__contains__",
```
zhiltsov-max, line 398: `__contains__` is mentioned twice.

cvat/apps/engine/models.py
```diff
@@ -181,6 +183,242 @@ def choices(cls):
     def __str__(self):
         return self.value
 
+
+T = TypeVar("T")
+
+
+def _parse_both_before_accessing(fn):
```
zhiltsov-max, line 190: Please give a more clear name to the function. It's not clear, what this "both" is.

cvat/apps/engine/tests/test_lazy_list.py
```diff
@@ -0,0 +1,166 @@
...
+    def test_parse_before_accessing_decorator(self):
+        lazy_list_copy = LazyList(string="1,2,3", converter=int)
+        lazy_list_copy.append(4)
+        self.assertEqual(lazy_list_copy, [1, 2, 3, 4])
+
+    def test_parse_both_before_accessing_decorator(self):
+        other_list = LazyList(string="4,5", converter=int)
+        result = self.lazy_list + other_list
+        self.assertEqual(result, [1, 2, 3, 4, 5])
+
```
zhiltsov-max, line 163: It makes sense to add a test for calling `len()` in the middle of the first iteration.
```
for x in lazy_list:
    len(lazy_list)
```

cvat/apps/engine/models.py
Bobronium: Though in this particular case I didn't see any complications, you're right, `locals()` might be too fragile to rely on.

Using `__getattribute__`, on the other hand, would introduce all sorts of complications, so I went with moving this code to metaclass.

cvat/apps/engine/models.py
```diff
@@ -181,6 +183,253 @@ def choices(cls):
...
+            "__le__",
+        ]:
+            namespace[method_name] = _parse_self_and_other_before_accessing(
+                getattr(list, method_name)
+            )
+
+        return super().__new__(mcs, name, bases, namespace)
+
+
+class LazyList(list[T], metaclass=LazyListMeta):
```
zhiltsov-max, line 267: It makes sense to move this class into a separate file, as it's quite big already and has no direct relations to models.

cvat/apps/engine/models.py
Bobronium: I went with `dataclass` initially, however I think it's best to keep these attributes underscored, yet keep `__init__` arguments public (i.e. `LazyList('1,2,3', converter=int)` instead of `LazyList('1,2,3', _converter=int)`. 

AFAIK, there's no simple way to allow that with `dataclass` or `attr.define`.

cvat/apps/engine/models.py
Bobronium: Done

cvat/apps/engine/models.py
Bobronium: Done

cvat/apps/engine/models.py
zhiltsov-max: For attrs, it's explained [here](https://www.attrs.org/en/stable/examples.html#basics) ("For private attributes, ..."). Basically, it's the default behavior for underscored fields.

cvat/apps/engine/lazy_list.py
```diff
@@ -0,0 +1,254 @@
+from functools import wraps
```
zhiltsov-max, line 1: Please add license header.

cvat/apps/engine/models.py
zhiltsov-max: It's possible to turn on code formatting for the new files, if you want. Just add a copy of `pyproject.toml` into `engine/` and update paths in `dev/format_python_code.py`.

Bobronium: I put `pyproject.toml` in a root dir and removed black config from other `pyproject.toml` files. Couldn't get rid of configuring `isort` per directory though since it starts to confuse first-party packages with third-party.

zhiltsov-max: > Couldn't get rid of configuring isort per directory though since it starts to confuse first-party packages with third-party.

Yes, it was the reason to copy the file into each component.

PR #8272, opened 2024-08-07
Per segment chunks

zhiltsov-max:
### Motivation and context

- Changed chunk generation from per-task chunks to per-segment chunks
- Fixed a memory leak in video reading on the server side (only in media_extractors, so there are several more left)
- Fixed a potential hang in `import` worker or the server process on process shutdown
- Disabled multithreading in video reading in endpoints (not in static chunk generation)
- Refactored static chunk generation code (moved after job creation)
- Refactored various server internal APIs for frame retrieval
- Updated UI logic to access chunks, added support for non-sequential frames in chunks
- Added a new server configuration option `CVAT_ALLOW_STATIC_CACHE` (boolean) to enable and disable static cache support. The option is disabled by default (it's changed from the previous behavior)
- Added tests for the changes made
- Added missing original chunk type field in job responses
- Fixed invalid kvrocks cleanup in tests for Helm deployment
- Added a new 0-based `index` parameter in `GET /api/jobs/{id}/data/?type=chunk` to simplify indexing
  - GT job chunks with non-sequential frames have no placeholders inside

When this update is applied to the server, there will be a data storage setting migration for the tasks. Existing tasks using static chunks (`task.data.storage_method == FILE_SYSTEM`) will be switched to the dynamic cache (i.e. to `== CACHE)`). The remaining files should be removed manually, there will be a list of such tasks in the migration log file. 

After this update, you'll have an option to enable or disable static cache use during task creation. This allows, in particular, prohibit new tasks using the static cache. With this option, any tasks using static cache will use the dynamic cache instead on data access.

User-observable changes:
- Job chunk ids now start from 0 for each job instead of using parent task ids
- The `use_cache = false` or `storage_method = filesystem` parameters in task creation can be ignored by the server
- Task chunk access may be slower for some chunks (particularly, for tasks with overlap configured, for chunks on segment boundaries, and for tasks previously using static chunks)
- The last chunk in a job will contain only the frames from the current job, even if there are more frames in the task

### Checklist

### License

## Summary by CodeRabbit

- **New Features**
	- Introduced a new server setting to disable media chunks on the local filesystem.
	- Enhanced frame prefetching with a `startFrame` parameter for improved chunk calculations.
	- Added a new property, `data_original_chunk_type`, for enhanced job differentiation in the metadata.

- **Bug Fixes**
	- Resolved memory management issues to prevent leaks during video processing.
	- Corrected naming inconsistencies related to the `prefetchAnalyzer`.

- **Documentation**
	- Included configuration for code formatting tools to ensure consistent code quality across the project.

- **Refactor**
	- Restructured classes and methods for improved clarity and maintainability, particularly in media handling and task processing.

- **Chores**
	- Updated formatting scripts to include additional directories for automated code formatting.


cvat/apps/engine/models.py
```diff
@@ -256,29 +256,29 @@ def get_original_cache_dirname(self):
...
+    def _get_compressed_chunk_name(self, segment_id: int, chunk_number: int) -> str:
+        return self._get_chunk_name(segment_id, chunk_number, self.compressed_chunk_type)
 
-    def _get_original_chunk_name(self, chunk_number):
-        return self._get_chunk_name(chunk_number, self.original_chunk_type)
+    def _get_original_chunk_name(self, segment_id: int, chunk_number: int) -> str:
+        return self._get_chunk_name(segment_id, chunk_number, self.original_chunk_type)
 
-    def get_original_chunk_path(self, chunk_number):
+    def get_original_segment_chunk_path(self, chunk_number: int, segment: int) -> str:
```
nmanovic, line 275: segment_id to be consistent with the code above?

zhiltsov-max: Fixed

zhiltsov-max:
@coderabbitai review

cvat/apps/engine/task.py
```diff
@@ -960,206 +939,317 @@ def update_progress(progress):
...
 
     slogger.glob.info("Found frames {} for Data #{}".format(db_data.size, db_data.id))
-    _save_task_to_db(db_task, job_file_mapping=job_file_mapping)
+    _create_segments_and_jobs(db_task, job_file_mapping=job_file_mapping)
+
+    if (
+        settings.MEDIA_CACHE_ALLOW_STATIC_CACHE and
+        db_data.storage_method == models.StorageMethodChoice.FILE_SYSTEM
+    ):
+        _create_static_chunks(db_task, media_extractor=extractor)
```
Marishka17, line 1121: Is it expected behavior?
- when there was a `manifest.jsonl` file uploaded together with images, `use_cache` == False, `MEDIA_CACHE_ALLOW_STATIC_CACHE` == True -> task created successfully with static chunks, but before that manifest index was initialized. That means that despite use_cache==False and MEDIA_CACHE_ALLOW_STATIC_CACHE==True values, we prepare manifest (if it was not passed), initialize a manifest index, and then prepare static chunks. It seems odd.

If I remember correctly, previously it was not allowed to pass a manifest file when use_cache == False.

cvat/apps/engine/task.py
```diff
@@ -551,14 +550,17 @@ def _update_status(msg: str) -> None:
     else:
         assert False, f"Unknown file storage {db_data.storage}"
 
+    if (
+        db_data.storage_method == models.StorageMethodChoice.FILE_SYSTEM and
+        not settings.MEDIA_CACHE_ALLOW_STATIC_CACHE
+    ):
+        db_data.storage_method = models.StorageMethodChoice.CACHE
```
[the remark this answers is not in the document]
zhiltsov-max: As we discussed, we doesn't make sense to provide such an option to users. As far as I remember, currently we don't provide any task creation logs on the requests page, where such a notification could be shown.

cvat/apps/engine/task.py
zhiltsov-max: Yes, I changed it to generate manifest in any case. Basically, the idea is to use it for optimized data access on the server side regardless of the task creation method. It's also possible to use it even if the task was created with static chunks, as the server config may change.

zhiltsov-max:
> @zhiltsov-max, I don't remember whether we decided to validate passed/calculated chunk_size based on segment_size or not. However, If we are not going to prohibit task creation (or adapt chunk_size) when chunk_size > segment_size, then GET /api/jobs/<id>/data/meta should return min(chunk_size, segment_size).

We decided to optionally implement chunk size restrictions based on the data size, but this PR was already quite big, so I didn't add it. The existing implementation should handle the situation you're talking about in [this function](https://github.com/cvat-ai/cvat/blob/1a9a8136e68338b1c5d03f5faeeb8ba674cefc7c/cvat/apps/engine/frame_provider.py#L296) for tasks and [here](https://github.com/cvat-ai/cvat/blame/1a9a8136e68338b1c5d03f5faeeb8ba674cefc7c/cvat/apps/engine/task.py#L1247C25-L1248) for static chunks for segments.

cvat/apps/engine/task.py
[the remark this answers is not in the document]
Marishka17: > currently we don't provide any task creation logs on the requests page, where such a notification could be shown
 
 but we provide the current process status where such info is displayed
image

cvat/apps/engine/cache.py
```diff
@@ -1,248 +1,359 @@
...
+                prev_frame <= cur_frame
+            ), f"Requested frame ids must be sorted, got a ({prev_frame}, {cur_frame}) pair"
+
+        db_data = db_task.data
+
+        raw_data_dir = {
+            models.StorageChoice.LOCAL: db_data.get_upload_dirname(),
+            models.StorageChoice.SHARE: settings.SHARE_ROOT,
+            models.StorageChoice.CLOUD_STORAGE: db_data.get_upload_dirname(),
         }[db_data.storage]
```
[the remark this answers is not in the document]
zhiltsov-max: Yes, it's a good idea.

cvat/apps/engine/migrations/0083_move_to_segment_chunks.py
```diff
@@ -0,0 +1,55 @@
...
+            v[0]
+            for v in (
+                data_with_static_cache_query
+                .order_by('id')
+                .values_list('id')
+                .iterator(chunk_size=100000)
+            )
+        )
+
+        data_with_static_cache_query.update(storage_method="cache")
```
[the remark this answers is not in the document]
zhiltsov-max: It makes sense to double check it, but I supposed it to generate a SQL like:
```sql
UPDATE data_table
SET storage_method = "cache"
WHERE storage_method = "file_system";
```

So no batched iteration is needed.

cvat/apps/engine/frame_provider.py
```diff
@@ -3,226 +3,570 @@
...
-    def get_frames(self, start_frame, stop_frame, quality=Quality.ORIGINAL, out_type=Type.BUFFER):
-        for idx in range(start_frame, stop_frame):
-            yield self.get_frame(idx, quality=quality, out_type=out_type)
+def make_frame_provider(data_source: Union[models.Job, models.Task, Any]) -> IFrameProvider:
+    if isinstance(data_source, models.Task):
+        frame_provider = TaskFrameProvider(data_source)
+    elif isinstance(data_source, models.Job):
+        frame_provider = JobFrameProvider(data_source)
+    else:
+        raise TypeError(f"Unexpected data source type {type(data_source)}")
```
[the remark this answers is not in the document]
zhiltsov-max: I agree, but in the same time I'd like to make `SegmentFrameProvider` private to `frame_provider.py` and `cache.py`. Need to think about it.

cvat/apps/engine/task.py
```diff
@@ -917,38 +923,11 @@ def _update_status(msg: str) -> None:
...
-
-        status_message = 'CVAT is preparing data chunks'
-        if not progress:
-            status_message = '{} {}'.format(status_message, progress_animation[update_progress.call_counter])
-        job.meta['status'] = status_message
-        job.meta['task_progress'] = progress or 0.
-        job.save_meta()
-        update_progress.call_counter = (update_progress.call_counter + 1) % len(progress_animation)
-
     compressed_chunk_writer_class = Mpeg4CompressedChunkWriter if db_data.compressed_chunk_type == models.DataChoice.VIDEO else ZipCompressedChunkWriter
```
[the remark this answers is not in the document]
zhiltsov-max: Thanks, I'll check.

cvat/apps/engine/task.py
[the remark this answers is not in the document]
zhiltsov-max: So if there is such a log message, it will be immediately overwritten by the "Completed" status, won't it? We would generate this message in the very end of the task creation.

cvat/apps/engine/migrations/0083_move_to_segment_chunks.py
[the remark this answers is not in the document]
Marishka17: You're right

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -4115,79 +4138,34 @@ def _test_api_v2_tasks_id_data_create_can_use_server_images_and_manifest(self, u
...
-                image_sizes, StorageMethodChoice.CACHE, StorageChoice.SHARE,
-                expected_task_creation_status_state='Failed',
-                expected_task_creation_status_reason='Incorrect file mapping to manifest content')
-
-        for copy_data in [True, False]:
-            with self.subTest(current_function_name(), copy=copy_data):
+        for use_cache in [True, False]:
+            task_data['use_cache'] = use_cache
+
+            for copy_data in [True, False]:
```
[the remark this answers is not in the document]
zhiltsov-max: There is another test below in this loop that doesn't depend in `copy_data`, but depends on `use_cache`.

cvat/apps/engine/task.py
```diff
@@ -917,38 +923,11 @@ def _update_status(msg: str) -> None:
...
-    kwargs = {}
-    if validate_dimension.dimension == models.DimensionType.DIM_3D:
-        kwargs["dimension"] = validate_dimension.dimension
-    compressed_chunk_writer = compressed_chunk_writer_class(db_data.image_quality, **kwargs)
-    original_chunk_writer = original_chunk_writer_class(original_quality, **kwargs)
 
     # calculate chunk size if it isn't specified
     if db_data.chunk_size is None:
-        if isinstance(compressed_chunk_writer, ZipCompressedChunkWriter):
+        if issubclass(compressed_chunk_writer_class, ZipCompressedChunkWriter):
```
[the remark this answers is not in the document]
zhiltsov-max: Done

cvat/apps/engine/frame_provider.py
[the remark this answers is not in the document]
zhiltsov-max: Added `SegmentFrameProvider` in this function.

cvat/apps/engine/cache.py
```diff
@@ -1,248 +1,359 @@
...
-        slogger.glob.info(f'Starting to get chunk from cache: key {key}')
+        slogger.glob.info(f"Starting to get chunk from cache: key {key}")
         try:
             item = self._cache.get(key)
         except pickle.UnpicklingError:
-            slogger.glob.error(f'Unable to get item from cache: key {key}', exc_info=True)
+            slogger.glob.error(f"Unable to get item from cache: key {key}", exc_info=True)
             item = None
-        slogger.glob.info(f'Ending to get chunk from cache: key {key}, is_cached {bool(item)}')
+        slogger.glob.info(f"Ending to get chunk from cache: key {key}, is_cached {bool(item)}")
```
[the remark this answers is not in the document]
zhiltsov-max: Done

cvat/apps/engine/cache.py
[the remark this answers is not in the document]
zhiltsov-max: Done

changelog.d/20240812_161617_mzhiltso_job_chunks.md
```diff
@@ -0,0 +1,4 @@
+### Added
+
+- A server setting to disable media chunks on the local filesystem
```
bsekachev, line 3:
```suggestion
- A server setting to disable permanent media chunks on local filesystem
```

changelog.d/20240812_161734_mzhiltso_job_chunks.md
```diff
@@ -0,0 +1,4 @@
+### Changed
+
+- Jobs now have separate chunk ids starting from 0, instead of using ones from the task
```
bsekachev, line 3: I would suggest to formulate somehow more clear for end user (not sure actually how exactly). 
But this record does not say me (as a CVAT user anything). This is probably only important for REST API users 

Maybe: **Numeration of data chunks in any job always starts with 0**

bsekachev: By the way this change is breaking. Thus it means that we have to up major version of REST API. 
Hovewer I am not sure we want do it now. Probably worthy of internal discussion

bsekachev (commented):
- From first look `FrameDecoder` and `PrefetchAnalyizer` themself may know nothing about `startFrame` if their public methods always accepts frame numbers relatively to the segment start frame. That may do architecture a little easier. 

- We need to drop all cache in any jobs which are not the first in a task (it will be a little difficult, so, maybe easier just to drop all cache, but this is about our enterprise repository. Nevertheless, must be done at the same release). 

- If REST API for each job enumerates chunks from zero (Job View Set), we probably should follow the same approach for context images and frames. Hovewer I do not see any changes related to it. E.g. when we request: `/api/jobs/<id>/data?type=context_image&number=<number>`, `<number>` probably should be relative to the job start. 

The same about `frame` parameter, hovewer it is not used on UI as far as I remember. 
image

The same about `frame` parameter.

changelog.d/20240812_161734_mzhiltso_job_chunks.md
bsekachev: Why did you decide so?

changelog.d/20240812_161617_mzhiltso_job_chunks.md
zhiltsov-max: There were only permanent chunks on the local filesystem.

changelog.d/20240812_161734_mzhiltso_job_chunks.md
zhiltsov-max: We never established SemVer for the server component. There were discussions a couple of years ago, we decided to postpone it. Currently, we can have any changes in the server API in any release, the version is the same as in the release itself.

zhiltsov-max:
> If REST API for each job enumerates chunks from zero (Job View Set), we probably should follow the same approach for context images and frames. Hovewer I do not see any changes related to it. E.g. when we request: /api/jobs/<id>/data?type=context_image&number=<number>, <number> probably should be relative to the job start.

It can be a good idea, but it wasn't planned for the feature. Only chunks were supposed to change. I can do a separate PR later for this, once it's needed.

changelog.d/20240812_161734_mzhiltso_job_chunks.md
bsekachev: We increase server version each release (minor or patch) depends on changes. CVAT version = its server version

changelog.d/20240812_161617_mzhiltso_job_chunks.md
bsekachev: Regular users (even those who may install self-hosted solution) do not know such implementation details.

bsekachev: I just want to clarify that this is exactly about permanent chunks, explicitly

SpecLad:
Have you checked the impact of this change on the dataset layer in the SDK? I would assume that some adaptation would be needed, since it relies on the chunk layout to find frames by index.

zhiltsov-max:
> Have you checked the impact of this change on the dataset layer in the SDK? I would assume that some adaptation would be needed, since it relies on the chunk layout to find frames by index.

Yes, we only use task chunks in SDK, they are unchanged by this PR. The tests pass.

bsekachev:
One more idea is that before releasing it, we may somehow notify users to force them to update client immediately after release. It may be especially important in this case.

changelog.d/20240812_161734_mzhiltso_job_chunks.md
zhiltsov-max: So your suggestion is to make it CVAT 3.0?

bsekachev: As I mentioned in the beginning: 

> Hovewer I am not sure we want do it now. Probably worthy of internal discussion

As I see strong reasons do not update version right now.

bsekachev: Maybe instead of modifying existing `chunk_number` parameter, we have to add one more and deprecate previous behaviour

changelog.d/20240812_161617_mzhiltso_job_chunks.md
zhiltsov-max: Ok, but it's just what is passed in the task data endpoint API: `storage_method = 'file_system' | 'cache'`.

changelog.d/20240812_161617_mzhiltso_job_chunks.md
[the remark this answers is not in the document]
SpecLad: Just FYI, when I wrote "pick one", I meant to pick one category for the heading. You _can_ put multiple headings in one fragment.

That said, it doesn't really matter if you put your entries in one fragment or many, since the end result is the same.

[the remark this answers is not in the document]
zhiltsov-max: Great, nice to know.

changelog.d/20240812_161734_mzhiltso_job_chunks.md
zhiltsov-max: Added the `index` parameter.

changelog.d/20240812_161617_mzhiltso_job_chunks.md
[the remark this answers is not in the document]
zhiltsov-max: Joined fragments into one.

cvat/apps/engine/cache.py
```diff
@@ -1,248 +1,359 @@
...
+            create_callback=lambda: self.prepare_masked_range_segment_chunk(
+                db_job.segment, chunk_number, quality=quality
+            ),
+        )
 
-        item = self._get_or_set_cache_item(
-            key, create_function=lambda: self._prepare_cloud_preview(db_storage)
+    def get_or_set_segment_preview(self, db_segment: models.Segment) -> DataWithMime:
+        return self._get_or_set_cache_item(
+            f"segment_preview_{db_segment.id}",
```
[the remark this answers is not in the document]
zhiltsov-max: I've added several functions to handle this. It can be refactored further, once we find a better implementation.

changelog.d/20240812_161617_mzhiltso_job_chunks.md
zhiltsov-max: Updated the message

cvat-data/src/ts/cvat-data.ts
```diff
@@ -155,17 +155,43 @@ export class FrameDecoder {
...
+        }
+    }
+
+    private arraysEqual(a: number[], b: number[]): boolean {
+        return (
+            a.length === b.length &&
+            a.every((element, index) => element === b[index])
+        );
+    }
+
```
bsekachev, line 182: Is there any reason to make it class method?

cvat-data/src/ts/cvat-data.ts
```diff
@@ -155,17 +155,43 @@ export class FrameDecoder {
...
+            const prev = frameNumbers[i - 1];
+            const current = frameNumbers[i];
+            if (current <= prev) {
+                throw new Error(
+                    'frameNumbers must be sorted in ascending order, ' +
+                    `got a (${prev}, ${current}) pair instead`,
+                );
+            }
+        }
+    }
```
bsekachev, line 174: Is there any reason to make it class method?

cvat-data/src/ts/cvat-data.ts
```diff
@@ -155,17 +155,43 @@ export class FrameDecoder {
         }
     }
 
+    private validateFrameNumbers(frameNumbers: number[]): void {
+        if (!frameNumbers || !frameNumbers.length) {
```
bsekachev, line 159:
```suggestion
        if (!Array.isArray(frameNumbers) || !frameNumbers.length) {
```

cvat-data/src/ts/cvat-data.ts
```diff
@@ -175,12 +201,14 @@ export class FrameDecoder {
                 // it was other chunk
                 this.requestedChunkToDecode.onReject(new RequestOutdatedError());
             }
-        } else if (this.chunkIsBeingDecoded === null || this.chunkIsBeingDecoded.start !== start) {
+        } else if (this.chunkIsBeingDecoded === null ||
+            !this.arraysEqual(frameNumbers, this.requestedChunkToDecode.frameNumbers)
```
bsekachev, line 205: From the previous condition `this.requestedChunkToDecode` is null, but here you use `this.requestedChunkToDecode.frameNumbers`

cvat-data/src/ts/cvat-data.ts
```diff
@@ -253,16 +281,17 @@ export class FrameDecoder {
             releaseMutex();
         };
         try {
-            const { start, end, block } = this.requestedChunkToDecode;
-            if (start !== blockToDecode.start) {
+            const { frameNumbers, chunkNumber, block } = this.requestedChunkToDecode;
+            if (!this.arraysEqual(frameNumbers, blockToDecode.frameNumbers)) {
```
bsekachev, line 285: Will not it be enough to compare `chunkNumber` and `blockToDecode.chunkNumber`?

cvat-data/src/ts/cvat-data.ts
zhiltsov-max: I thought it can be optimized somehow for this case, but maybe we can just check `chunkNumber` instead, which was added later.

cvat-data/src/ts/cvat-data.ts
zhiltsov-max: I think it should work.

cvat-data/src/ts/cvat-data.ts
zhiltsov-max: Removed, probably not needed anymore.

cvat-core/src/frames.ts
```diff
@@ -144,6 +151,48 @@ export class FramesMetaData {
...
+    getFrameStep(): number {
+        if (this.frameFilter) {
+            const frameStepParts = this.frameFilter.split('=', 2);
+            if (frameStepParts.length !== 2) {
+                throw new Error(`Invalid frame filter '${this.frameFilter}'`);
+            }
+            return parseInt(frameStepParts[1], 10);
+        }
+        return 1;
+    }
```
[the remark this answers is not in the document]
zhiltsov-max: I changed it to be a field. It seems like `frameFilter` in this class is immutable, so a cached field should perform better.

cvat-data/src/ts/cvat-data.ts
zhiltsov-max: Fixed

cvat-data/src/ts/cvat-data.ts
bsekachev: Just suggestion to move it out of the class as it does not use `this` context somehow. Generally it is a minor question.

bsekachev:
Perhaps need to prepare a pull request in private repos to update code according to new changes
image

bsekachev:
I will suggest to make a global replacement, like this to go to the same naming approach

image

bsekachev:
If to press `Play` in a GT job 

image

I think the problem is somewhere in this code:
image

As for a ground truth job this condition may not to work.

cvat/apps/engine/default_settings.py
```diff
@@ -0,0 +1,16 @@
+# Copyright (C) 2024 CVAT.ai Corporation
```
nmanovic, line 1: Why do we need the file? Why don't define the settings in the global settings file?

zhiltsov-max: You can redefine these settings in the global settings file if you need it, this file defines the default values.

bsekachev (approved):

tests/docker-compose.configurable_static_cache.yml
Marishka17, line 1: Why do we need a separate `docker-compose.configurable_static_cache.yml`? Why not pass the environment variable through the main docker-compose.yml? (+ IMHO, It should be possible to configure static cache enabling/disabling not only in tests)

zhiltsov-max: Makes sense, added.

cvat/apps/engine/cache.py
```diff
@@ -1,323 +1,569 @@
...
+        self, obj: Union[models.Task, models.Segment, models.Job, models.CloudStorage]
+    ) -> str:
+        if isinstance(obj, models.Task):
+            return f"task_{obj.id}"
+        elif isinstance(obj, models.Segment):
+            return f"segment_{obj.id}"
+        elif isinstance(obj, models.Job):
+            return f"job_{obj.id}"
+        elif isinstance(obj, models.CloudStorage):
+            return f"cloudstorage_{obj.id}"
```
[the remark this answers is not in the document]
zhiltsov-max: The fact that it can be done this way is more of a coincidence than a rule. Doing so can lead to unexpected changes in future, e.g. if a model is renamed.

cvat/apps/engine/cache.py
```diff
@@ -1,323 +1,569 @@
...
+        def create_item() -> _CacheItem:
+            slogger.glob.info(f"Starting to prepare chunk: key {key}")
+            item_data = create_callback()
+            slogger.glob.info(f"Ending to prepare chunk: key {key}")
+
+            if item_data[0]:
+                item = (item_data[0], item_data[1], self._get_checksum(item_data[0].getbuffer()))
                 self._cache.set(key, item)
+            else:
+                item = (item_data[0], item_data[1], None)
```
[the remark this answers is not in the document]
zhiltsov-max: Fixed

zhiltsov-max:
> If I create a task with 4 images and start_frame 2, I'll receive a 400 code on the task preview request.

Fixed

> I'm not sure whether it is related to the current PR or not, but I guess the progress bar can show an incorrect interval of loaded chunks/frames for a GT job (e.g. there were 2 frames 3 and 6 in the GT job):

Fixed

cvat/apps/engine/cache.py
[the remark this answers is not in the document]
Marishka17: Okay, but I guess if a model is renamed then the used cache key also should be updated. Otherwise, I think it will be odd to have a renamed model `RemoteStorage` and use the `cloudstorage_<id>` key.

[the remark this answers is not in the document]
zhiltsov-max: Maybe, but it not follows from the model name.

cvat/apps/engine/cache.py
```diff
@@ -1,349 +1,700 @@
...
+        else:
+            requested_frame_iter = iter(frame_ids)
+            next_requested_frame_id = next(requested_frame_iter, None)
+            if next_requested_frame_id is None:
+                return
+
+            # TODO: find a way to use prefetched results, if provided
+            db_images = (
+                db_data.images.order_by("frame")
+                .filter(frame__gte=frame_ids[0], frame__lte=frame_ids[-1])
```
Marishka17, line 328: Probably we should set chunk_size max limit and use here `frame__in` (in that case we do not need to iterate over `requested_frame_iter` and compare a frame with `next_requested_frame_id`). What do you think? (can be changed in a separate PR)

zhiltsov-max: I don't really like using `__in`, as it can result in very big queries, which are processed slowly. And it is possible here, as `chunk_size` is only limited by the task size now. I think it can be done once we introduce a limit on the `chunk_size`, which is planned to be done in another PR.

Marishka17: > I don't really like using __in, as it can result in very big queries, which are processed slowly.

I agree, but I also said that `frame__in` should be used only when we add a `chunk_size` limit.

zhiltsov-max: Ok, I thought you're speaking about order by + limit in the resulting sql.

Marishka17 (approved):

PR #8275, opened 2024-08-07
Optimize DB requests in task list, quality conflicts list and cloudstorages

zhiltsov-max:
### Motivation and context

- Optimized task list requests
- Optimized quality conflicts requests
- Optimized calls to related manifests in cloudstorages api (however, manifests should be either moved to a separate endpoint or should be limited in maximum number per CS)

Notes:
- [x] `/api/quality/conflicts/` - will be further changed in https://github.com/cvat-ai/cvat/pull/9116, but improved for now
- [ ] `/api/requests/` - long processing for getting a list of requests
- [x] <s>Try to optimize `count()` requests for list endpoints. Because of the [filter.distinct used](https://github.com/cvat-ai/cvat/blob/9ac1ba2ee25a9310a2ff74a14ec5ac886be4b39f/cvat/apps/iam/permissions.py#L204), we get slow requests. Potentially can be optimized inside pagination</s> So far the chosen approach is to provide an optimized implementation in a dedicated `list_serializer_class.to_representation()`.
- [x] Try to use prefetch cache for IAM checks, they request orgs and memberships several times - improved in https://github.com/cvat-ai/cvat/pull/9275, used here. Memberships fetches still can probably be improved

### Checklist

### License


bsekachev:
/api/jobs now takes longer time (on queries metric)

Each /api/jobs/id/preview also now takes longer time (on queries metric) [**significantly**]
In the second request we do not need many JOINS, so, I would suggest to reduce number of selected_related arguments.

zhiltsov-max:
I don't think these time measurements are reliable or relevant to the patch.

zhiltsov-max:
I've increased the page_size to get more representative output:
`http://localhost:7000/api/jobs?page_size=1000` (in my case it returned first 1000 of 2250 jobs).

Baseline:

Screenshot from 2024-08-07 19-29-46


Select_related (source_storage, target_storage, organization, owner), 12 joins:

Screenshot from 2024-08-07 19-29-53


Select_related (source_storage, target_storage) + prefetch_related(organization, owner), 10 joins:

Screenshot from 2024-08-07 19-29-58


And the same, but for 12 jobs (as UI presents):

Screenshot from 2024-08-07 19-28-42
Screenshot from 2024-08-07 19-29-31
Screenshot from 2024-08-07 19-29-37

I think, it makes sense to choose one of the 2 optimized variants. I think, the version with 10 joins is a good tradeoff for now, it provides a good balance between the two others.

cvat/apps/engine/views.py
```diff
@@ -1705,8 +1705,17 @@ class JobViewSet(viewsets.GenericViewSet, mixins.ListModelMixin, mixins.CreateMo
...
+        'assignee',
+        'segment__task__data',
+        'segment__task__project',
+        'segment__task__annotation_guide',
+        'segment__task__project__annotation_guide',
+        'segment__task__source_storage',
+        'segment__task__target_storage',
+    ).prefetch_related(
+        'segment__task__organization',
+        'segment__task__owner',
```
bsekachev, line 1718: May you explain why using prefetch_related is better here?

From my opinion it just makes extra requests (what in general requires more time) to the database and does not have benefits in comparison with select_related

IMHO we just need to define queryset "per view", leaving in defaults only those fields, necessary to IAM logic (e.g.  organization, owner)

zhiltsov-max: > May you explain why using prefetch_related is better here?
> From my opinion it just makes extra requests (what in general requires more time) to the database and does not have benefits in comparison with select_related

Yes, it does extra requests and does joining on the server side instead of doing it on the DB side, like in `select_related`. There are 2 "opposite" sides of prefetching - prefetch and join everything on the DB side and prefetch everything separately and join on the server side. Basically yes, we'd always prefer joining on the DB side, but separation can help in the following case: there is a small table, and its results are expected to be reused in many results. In this case it may be beneficial to fetch this small table separately and join it on the server side. Doing it on the DB side would significantly increase traffic by populating it with repeated entries. Probably, it's the case in my setup - I have just several users and orgs. I think the best strategy for optimizing this is actually measure the changes on the production server. Or, at least, take the statistics from it, to make the split decisions informed.

> IMHO we just need to define queryset "per view", leaving in defaults only those fields, necessary to IAM logic (e.g. organization, owner)

Probably, you meant "per endpoint usecase"? Ok, it makes sense, I agree.

cvat/apps/engine/serializers.py
```diff
@@ -2275,6 +2275,35 @@ def _create_files(self, instance, files):
...
+            # Optimized prefetch only for the current page
+            page: list[models.Task] = data
+
+            # Annotate page objects
+            # We do it explicitly here and not in the LIST queryset to avoid
+            # doing the same DB computations twice - one time for the page retrieval
+            # and another one for the COUNT(*) request to get the total count
+            page_task_ids = set(t.id for t in page)
+            job_summary_fields = [
+                m.value for m in models.TaskQuerySet.JobSummaryFields.__members__.values()
```
SpecLad, line 2290:
```suggestion
                m.value for m in models.TaskQuerySet.JobSummaryFields
```

cvat/apps/engine/views.py
```diff
@@ -903,16 +903,14 @@ class TaskViewSet(viewsets.GenericViewSet, mixins.ListModelMixin,
...
         'assignee',
         'owner',
         'target_storage',
         'source_storage',
         'annotation_guide',
     ).prefetch_related(
-        'segment_set__job_set',
-        'segment_set__job_set__assignee',
-    ).with_job_summary()
+        'data__validation_layout',
```
SpecLad, line 912: Why did you move this from `select_related` to `prefetch_related`?

cvat/apps/engine/views.py
```diff
@@ -2750,7 +2755,7 @@ class CloudStorageViewSet(viewsets.GenericViewSet, mixins.ListModelMixin,
     mixins.RetrieveModelMixin, mixins.CreateModelMixin, mixins.DestroyModelMixin,
     PartialUpdateModelMixin
 ):
-    queryset = CloudStorageModel.objects.prefetch_related('data').all()
+    queryset = CloudStorageModel.objects
```
SpecLad, line 2758:
```suggestion
    queryset = CloudStorageModel.objects.all()
```

`CloudStorageModel.objects` is a manager rather than a queryset. TBH, I don't know if this actually makes a difference, but the examples in DRF docs use `.all`, so I think it's safer to follow them.

cvat/apps/engine/views.py
zhiltsov-max: The reason is that validation layout has the heavy field `data`. We have a `DISTINCT` clause in the [IAM list filter](https://github.com/cvat-ai/cvat/blob/14d7eb4e69e77b5635351626c285d0e99410c654/cvat/apps/iam/permissions.py#L239), which by default (without `ON`, [which is a Postgres-specific feature](https://www.postgresql.org/docs/current/sql-select.html#SQL-DISTINCT)) compares all columns in all rows to deduplicate rows. For small fields it's not a problem, but for big JSON fields it can become a performance bottleneck. Everything in `select_related` gets into the `COUNT` query as well, so if we load this field in `select_related`, the DB will use this field in `DISTINCT` comparisons. `prefetch_related` fields do not get into `COUNT`, they're fetched by django after the page is retrieved.

SpecLad: I see, very interesting. Could you summarize this in a comment?

cvat/apps/engine/serializers.py
```diff
@@ -2275,6 +2275,35 @@ def _create_files(self, instance, files):
...
+            ]
+            job_counts = {
+                task["id"]: task
+                for task in models.Task.objects
+                .filter(id__in=page_task_ids)
```
SpecLad, line 2295: I just thought of something WRT this and similar optimizations - how well does this behave when `page_size` is `all` and there are a lot of tasks?

zhiltsov-max: `page_size=all` must be allowed only in development deployments. It can crash the server because of OOM.

cvat/apps/engine/cache.py
```diff
@@ -412,7 +413,11 @@ def get_or_set_task_chunk(
             self._make_chunk_key(db_task, chunk_number, quality=quality),
             set_callback,
         )
-        db_task.refresh_from_db(fields=["segment_set"])
+
+        if is_field_cached(db_task, "segment_set"):
+            # Refresh segments to report actual dates if they were fetched previously
+            # Doing so without a check leads to an error if the related object is not prefetched
```
SpecLad, line 419: FWIW, I think that's a bug, so I filed a Django bug report: https://code.djangoproject.com/ticket/36372.

SpecLad (approved):

PR #8321, opened 2024-08-19
Job validations (public part)

zhiltsov-max:
### Motivation and context

- Added `max_validations_per_job`, `target_metric`, `target_metric_threshold` fields in quality settings
- Added `assignee_last_updated` in quality reports
- Added separate `accuracy`, `precision`, and `recall` fields in quality summary
- Refactored some IAM code

### Checklist

- [ ] I submit my changes into the `develop` branch
- [ ] I have created a changelog fragment 
- [ ] I have updated the documentation accordingly
- [ ] I have added tests to cover my changes
- [ ] I have linked related issues (see [GitHub docs](
  https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword))
- [ ] I have increased versions of npm packages if it is necessary
  ([cvat-canvas](https://github.com/cvat-ai/cvat/tree/develop/cvat-canvas#versioning),
  [cvat-core](https://github.com/cvat-ai/cvat/tree/develop/cvat-core#versioning),
  [cvat-data](https://github.com/cvat-ai/cvat/tree/develop/cvat-data#versioning) and
  [cvat-ui](https://github.com/cvat-ai/cvat/tree/develop/cvat-ui#versioning))

### License

- [ ] I submit _my code changes_ under the same [MIT License](
  https://github.com/cvat-ai/cvat/blob/develop/LICENSE) that covers the project.
  Feel free to contact the maintainers if that's a concern.

## Summary by CodeRabbit

- **New Features**
  - Enhanced permission handling for jobs, tasks, and projects, streamlining the creation process.
  - Introduced new quality control metrics and settings, enabling more granular tracking and reporting.

- **Bug Fixes**
  - Improved validation logic to enforce constraints on job validation attempts.

- **Documentation**
  - Added descriptive help texts for new fields in quality report serializers, clarifying their purpose.

- **Chores**
  - Updated method signatures to improve flexibility and maintainability across permission handling and reporting functionalities.

PR #8348, opened 2024-08-26
Validation API core

zhiltsov-max:
### Motivation and context

Depends on #8272
Depends on #8321

- Added server API for creation of a GT job on task creation
- Added server support for task creation with GT pool (aka Honeypot)
- Added new GT job frame selection method `random_per_job`, which guarantees each annotation job gets the specified GT overlap, making each annotation job validatable
- Added new GT job frame count selection options based on task size % and segment size %
- Changed GT job creation parameter "frames" to accept relative frame ids instead of absolute (source data) ones
- Allowed frame deletion in GT jobs. Deleted GT frames are considered excluded from validation, so should not appear in quality reports. Frame removal from a simple GT job (in tasks without honeypots) doesn't remove task frames, only the GT job frames.

Server API changes:
- GET /api/tasks/{id}/ got a new `validation_mode` field, reflecting the current validation configuration (immutable)
- POST /api/tasks/{id}/data got a new `validation_params` field, which allow to enable GT / GT_POOL validation for a task on its creation

#### Tasks with Honeypots
This validation mode affects task creation, so can only be used in task creation. It cannot be disabled or changed after the task is created. When honeypots are configured, each job in the task gets several extra validation frames. 
The pool of available frames and the number of validation frames per job are specified by the user at task creation. 

Limitations:
- This validation mode can only be used with `random` frame ordering. 
- Inherently, this assumes that `job_frame_mapping` and `overlap` cannot be used in such tasks. 
- Track annotations are prohibited in tasks with honeypots enabled. 

Honeypot frames and GT annotations are accessible via the GT job, as in the case with regular GT jobs. However, unlike regular tasks with GT jobs, task annotation import affects the GT job as well in tasks with honeypots. Task annotation export contains only GT annotations on validation frames (so, only the GT copy of validation frames is included).

### Checklist

### License

## Summary by CodeRabbit

- **New Features**
	- Introduced a server setting to disable media chunks on the local filesystem, enhancing configurability.
	- Added tracking for the last assignee update date in quality reports, improving task management.
	- Enhanced job chunk identifiers for better clarity and uniqueness.

- **Bug Fixes**
	- Resolved memory management issues and refined job assignment logic in video processing.

- **Documentation**
	- Updated API schema with new enhancements related to job management and validation processes.

- **Chores**
	- Updated package dependencies and added new configuration settings for Redis in the Helm chart.


cvat/apps/engine/model_utils.py
```diff
@@ -0,0 +1,5 @@
+# Copyright (C) 2024 CVAT.ai Corporation
+#
+# SPDX-License-Identifier: MIT
+
+from __future__ import annotations
```
klakhov, line 5: Seems like we dont need this file

klakhov:
- There is a corner case when Im trying to create a task with 100% honeypots. So it gives me the task with just 1 gt job and 0 regular jobs. Is there a usecase for that? maybe we shoud prohibit such case?
- I dont see any changes in `.rego`. Will non-admin users be able to compute quality reports after this pr?
- Would it be hard to improve formatting for the validation error? Like return only the stuff thats insize `string=`?. I suppose we have lots of cases now when we can get it. On the second thought I belive its easier to UI to parse it for better UX, we already have some handling for such cases. Maybe not really relevant for now

Overall pr seems to be working fine

cvat/apps/engine/views.py
```diff
@@ -849,7 +849,7 @@ class TaskViewSet(viewsets.GenericViewSet, mixins.ListModelMixin,
         'subset', 'mode', 'dimension', 'tracker_link'
```
zhiltsov-max, line 849: Add `select_related("data__validation_layout")`

tests/python/rest_api/test_jobs.py
```diff
@@ -86,74 +89,130 @@ def _test_create_job_fails(
...
+            ),
+            ("manual", {}),
+        ],
+        idgen=lambda **args: "-".join([args["frame_selection_method"], *args["method_params"]]),
+    )
     @pytest.mark.parametrize("task_mode", ["annotation", "interpolation"])
-    def test_can_create_gt_job_with_manual_frames(self, admin_user, tasks, jobs, task_mode):
-        user = admin_user
-        job_frame_count = 4
+    def test_can_gt_job_in_a_task(
```
zhiltsov-max, line 104:
```suggestion
    def test_can_create_gt_job_in_a_task(
```

cvat/apps/dataset_manager/task.py
```diff
@@ -768,46 +794,137 @@ def __init__(self, pk):
             Prefetch('data__images', queryset=models.Image.objects.order_by('frame'))
         ).get(id=pk)
 
-        # Postgres doesn't guarantee an order by default without explicit order_by
-        self.db_jobs = models.Job.objects.select_related("segment").filter(
-            segment__task_id=pk, type=models.JobType.ANNOTATION.value,
-        ).order_by('id')
+        requested_job_types = [models.JobType.ANNOTATION]
+        if hasattr(self.db_task.data, 'validation_layout') and (
+            self.db_task.data.validation_layout.mode == models.ValidationMode.GT_POOL
```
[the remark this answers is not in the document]
zhiltsov-max: Basically, it's just a equality check for mode. I can try to add a `@property` in `Data` with this.

cvat/apps/dataset_manager/task.py
```diff
@@ -768,46 +794,137 @@ def __init__(self, pk):
...
+            data = AnnotationIR(self.db_task.dimension, data)
+
+        if data.tracks:
+            # Only tags and shapes are supported in tasks with GT pool
+            raise ValidationError("Tracks are not supported when task validation mode is {}".format(
+                models.ValidationMode.GT_POOL
+            ))
+
+        gt_job = next(
+            db_job for db_job in self.db_jobs if db_job.type == models.JobType.GROUND_TRUTH
```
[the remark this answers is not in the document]
zhiltsov-max: I think it will be an assertion error, if the GT job doesn't exist.

zhiltsov-max:
@klakhov ,

> There is a corner case when Im trying to create a task with 100% honeypots. So it gives me the task with just 1 gt job and 0 regular jobs. Is there a usecase for that? maybe we shoud prohibit such case?

Hm, I think yes, it should produce a validation error during task creation.

> I don't see any changes in .rego. Will non-admin users be able to compute quality reports after this pr?

I think it will be ether in a separate PR or in the PR with allocation reports, it seems to be the most relevant place.

cvat/apps/dataset_manager/task.py
```diff
@@ -768,46 +794,137 @@ def __init__(self, pk):
...
+            self.db_task.data.images
+            .filter(is_placeholder=True, real_frame__in=gt_pool_frames)
+            .values_list('frame', 'real_frame')
+            .iterator(chunk_size=1000)
+        ):
+            task_validation_frame_ids.add(frame)
+            task_validation_frame_groups.setdefault(real_frame, []).append(frame)
+
+        assert sorted(gt_pool_frames) == list(range(min(gt_pool_frames), max(gt_pool_frames) + 1))
+        gt_annotations = data.slice(min(gt_pool_frames), max(gt_pool_frames))
```
zhiltsov-max, line 885: Need to make sure the annotations actually use absolute frame ids

changelog.d/20240819_210200_mzhiltso_validation_api.md
```diff
@@ -0,0 +1,4 @@
+### Added
+
+- Last assignee update date in quality reports, new options in quality settings
```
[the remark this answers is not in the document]
zhiltsov-max: Thanks, updated.

cvat/apps/engine/views.py
zhiltsov-max: Updated

cvat/apps/dataset_manager/task.py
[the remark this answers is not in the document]
zhiltsov-max: Added assertion

cvat/apps/dataset_manager/task.py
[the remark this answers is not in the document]
zhiltsov-max: Added property

cvat/apps/engine/model_utils.py
zhiltsov-max: Fixed

zhiltsov-max:
@coderabbitai review

cvat/apps/dataset_manager/task.py
```diff
@@ -431,13 +439,29 @@ def put(self, data):
...
+        if not isinstance(data, AnnotationIR):
+            data = AnnotationIR(self.db_job.segment.task.dimension, data)
+
+        db_data = self.db_job.segment.task.data
+
+        if data.tracks and db_data.validation.mode == models.ValidationMode.GT_POOL:
+            # Only tags and shapes can be used in tasks with GT pool
+            raise ValidationError("Tracks are not supported when task validation mode is {}".format(
+                models.ValidationMode.GT_POOL
+            ))
```
klakhov, line 461: So, as far as I understand we cant use tracks in tasks with HP. 
But when I try to add a track in UI I get this error:
image

Seems error message is wrong

zhiltsov-max: Thanks, fixed

cvat/apps/dataset_manager/task.py
zhiltsov-max: Was incorrect indeed, fixed.

cvat/apps/dataset_manager/task.py
bsekachev: @zhiltsov-max  What is the nature of such restriction?
I understand that tracks with interpolation do not make sense in honeypot jobs, but is it the only reason?

cvat-ui/src/components/quality-control/task-quality/allocation-table.tsx
```diff
@@ -44,12 +45,49 @@ function AllocationTable(props: Readonly<Props>): JSX.Element {
         selectedRows: [],
     });
 
-    const data = gtJobMeta.includedFrames.map((frameID: number) => ({
-        key: frameID,
-        frame: frameID,
-        name: gtJobMeta.frames[frameID]?.name ?? gtJobMeta.frames[0].name,
-        active: !(frameID in gtJobMeta.deletedFrames),
-    }));
+    const data = (() => {
```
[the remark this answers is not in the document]
zhiltsov-max: Extracted a separate function. Probably, after https://github.com/cvat-ai/cvat/pull/8471 it can be simplified.

cvat/apps/dataset_manager/task.py
zhiltsov-max: Frames are in the random order, tracks make no sense in such case.

bsekachev: It makes sense if each frame of the track is a keyframe.

bsekachev: Just do not understand the reason why we forced this restriction.
If a user does it, why not. It is the user's decision.

cvat-ui/src/components/create-task-page/create-task-content.tsx
```diff
@@ -442,6 +452,21 @@ class CreateTaskContent extends React.PureComponent<Props & RouteComponentProps,
             return;
         }
 
+        try {
+            this.validateQualityParams(quality);
```
bsekachev, line 456: Why not return true/false like in other validation methods here?
Do we need to pass quality as a param if the method is reading from the class instance state anyway? And again it was not implemented the same way as in other validators.

cvat-ui/src/components/create-task-page/create-task-content.tsx
```diff
@@ -442,6 +452,21 @@ class CreateTaskContent extends React.PureComponent<Props & RouteComponentProps,
...
+            if (e instanceof Error) {
+                notification.error({
+                    message: 'Could not create a task',
+                    description: e.message,
+                    className: 'cvat-notification-create-task-fail',
+                });
+                reject();
+                return;
+            }
+            throw e;
```
bsekachev, line 467: Why do you throw the error outside of the async function? There is not such statement in other checks

cvat-ui/src/components/create-task-page/create-task-content.tsx
```diff
@@ -442,6 +452,21 @@ class CreateTaskContent extends React.PureComponent<Props & RouteComponentProps,
...
+        } catch (e) {
+            if (e instanceof Error) {
+                notification.error({
+                    message: 'Could not create a task',
+                    description: e.message,
+                    className: 'cvat-notification-create-task-fail',
+                });
+                reject();
+                return;
+            }
```
bsekachev, line 466: Do not we reject/return if because of some reason e is not instance of Error? 
```suggestion
            notification.error({
                message: 'Could not create a task',
                description: e instanceof Error ? e.message : '',
                className: 'cvat-notification-create-task-fail',
            });
            reject();
            return;
        }
```

cvat-ui/src/components/quality-control/task-quality/allocation-table.tsx
```diff
@@ -32,6 +33,55 @@ interface RowData {
     active: boolean;
 }
 
+interface TableRowData extends RowData {
+    key: Key;
+}
+
+export function getAllocationTableContents(gtJobMeta: FramesMetaData, gtJob: Job): TableRowData[] {
```
bsekachev, line 40: Let's at least add comment that this function is going to be used of application code by plugins.

cvat/apps/engine/migrations/0084_honeypot_support.py
```diff
@@ -0,0 +1,171 @@
...
+        min(data_start_frame + db_segment.stop_frame * step, data_stop_frame) + step,
+        step,
+    )
+
+    if db_segment.type == "range":
+        frame_set = frame_range
+    elif db_segment.type == "specific_frames":
+        frame_set = set(frame_range).intersection(db_segment.frames or [])
+    else:
+        raise ValueError(f"Unknown segment type: {db_segment.type}")
```
bsekachev, line 38: I am not sure that raising uncaught exception is good in migration file

bsekachev: On our prod we only have `specific_frames` defined, so, it will not be a problem

cvat/apps/engine/models.py
```diff
@@ -219,7 +220,61 @@ class FloatArrayField(AbstractArrayField):
...
+    frames_per_job_count = models.IntegerField(null=True)
+    frames_per_job_share = models.FloatField(null=True)
+
+class ValidationFrame(models.Model):
+    validation_params = models.ForeignKey(
+        ValidationParams, on_delete=models.CASCADE, related_name="frames"
+    )
+    path = models.CharField(max_length=1024, default='')
+
+class ValidationLayout(models.Model):
```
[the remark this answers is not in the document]
zhiltsov-max: Layout:

1. [Merriam-Webster](https://www.merriam-webster.com/dictionary/layout) 
> the plan or design or arrangement of something laid out
2. [Wiki](https://en.wikipedia.org/wiki/Layout) 
> In general terms, a layout is a structured arrangement of items within certain limits, or a plan for such arrangement. Specifically, layout may refer to: Page layout, the arrangement of visual elements on a page.

It's used to describe validation frames in tasks, both for simple GT and for Honeypots. That's why it doesn't have `pool` in the name.

> What is the reason to bind ValidationParams and ValidationLayout to Data instead of Task model?

This is done to be the same as storing `deleted_frames` in the `Data` model. Basically, it describes task data.

cvat/apps/engine/migrations/0084_honeypot_support.py
zhiltsov-max: The reason is to fail the migration, if the DB contains invalid entries. We don't know why they are there and what to do with them.

cvat-ui/src/components/create-task-page/create-task-content.tsx
bsekachev: The method only checks one problem. So, the error message is anyway clear

bsekachev: >  I'd like to see some automatic switching to random frame ordering if honeypots are selected in the quality section

Sounds like something easy to implement

cvat/apps/engine/models.py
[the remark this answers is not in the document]
bsekachev: Layout sounds like some set of elements and their relation to each other. 
But in our case is just couple of sets. Pool would sound good to describe and it is applicable in general for both GT job and Honeypot job. 

Hovewer if you do not want to use the word **Pool** -> it is okay, up to you.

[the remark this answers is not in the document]
bsekachev: > This is made to be the same as storing deleted_frames in the Data model.

I can't really understand the explanation. Hovewer in the future this design may be a problem if we want to use the same **Data** object to create multiple tasks (this is not a fact that we will do this, but anyway).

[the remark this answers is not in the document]
bsekachev: Never mind, considering existing database layout feature like: **import raw data, select them and create tasks based on them** already not implementable without new database classes.

bsekachev:
can't create a task with honeypot job and context images:

```
  File "/home/bsekachev/app.cvat.ai/cvat_enterprise/.env/lib/python3.10/site-packages/rq/worker.py", line 1431, in perform_job
    rv = job.perform()
  File "/home/bsekachev/app.cvat.ai/cvat_enterprise/.env/lib/python3.10/site-packages/rq/job.py", line 1280, in perform
  File "/home/bsekachev/app.cvat.ai/cvat_enterprise/.env/lib/python3.10/site-packages/rq/job.py", line 1317, in _execute
    return func(*args, **kwds)
  File "/home/bsekachev/app.cvat.ai/cvat_enterprise/cvat/cvat/apps/engine/task.py", line 1347, in _create_thread
    models.RelatedFile.objects.bulk_create(db_related_files)
  File "/home/bsekachev/app.cvat.ai/cvat_enterprise/.env/lib/python3.10/site-packages/django/db/models/manager.py", line 87, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/home/bsekachev/app.cvat.ai/cvat_enterprise/.env/lib/python3.10/site-packages/django/db/models/query.py", line 803, in bulk_create
    returned_columns = self._batched_insert(
  File "/home/bsekachev/app.cvat.ai/cvat_enterprise/.env/lib/python3.10/site-packages/django/db/models/query.py", line 1831, in _batched_insert
    self._insert(
  File "/home/bsekachev/app.cvat.ai/cvat_enterprise/.env/lib/python3.10/site-packages/django/db/models/query.py", line 1805, in _insert
    return query.get_compiler(using=using).execute_sql(returning_fields)
  File "/home/bsekachev/app.cvat.ai/cvat_enterprise/.env/lib/python3.10/site-packages/django/db/models/sql/compiler.py", line 1822, in execute_sql
    cursor.execute(sql, params)
  File "/home/bsekachev/app.cvat.ai/cvat_enterprise/.env/lib/python3.10/site-packages/django/db/backends/utils.py", line 102, in execute
    return super().execute(sql, params)
  File "/home/bsekachev/app.cvat.ai/cvat_enterprise/.env/lib/python3.10/site-packages/django/db/backends/utils.py", line 67, in execute
    return self._execute_with_wrappers(
  File "/home/bsekachev/app.cvat.ai/cvat_enterprise/.env/lib/python3.10/site-packages/django/db/backends/utils.py", line 80, in _execute_with_wrappers
    return executor(sql, params, many, context)
  File "/home/bsekachev/app.cvat.ai/cvat_enterprise/.env/lib/python3.10/site-packages/django/db/backends/utils.py", line 84, in _execute
    with self.db.wrap_database_errors:
  File "/home/bsekachev/app.cvat.ai/cvat_enterprise/.env/lib/python3.10/site-packages/django/db/utils.py", line 91, in __exit__
    raise dj_exc_value.with_traceback(traceback) from exc_value
  File "/home/bsekachev/app.cvat.ai/cvat_enterprise/.env/lib/python3.10/site-packages/django/db/backends/utils.py", line 89, in _execute
    return self.cursor.execute(sql, params)
django.db.utils.IntegrityError: duplicate key value violates unique constraint "engine_relatedfile_data_id_path_a7223d1e_uniq"
DETAIL:  Key (data_id, path)=(5, /home/bsekachev/app.cvat.ai/cvat_enterprise/cvat/data/data/5/raw/context_images example/related_images/3Z2A3692_jpg/3Z2A3692.jpg) already exists.
```

bsekachev (approved):

PR #8426, opened 2024-09-10
`*Permission.get_scopes`: don't tolerate unknown actions

SpecLad:
### Motivation and context

With almost all of the `get_scopes` methods, an unknown (action, method) combination will result in an array like `[None]` being returned (sometimes with other elements as well). If that happens, the OPA input will then have `"scope": null`, and so the policy evaluation will fail, unless the user is an admin.

Because of this, it's really easy to accidentally make a view admin-only, by forgetting to add/update an entry in `get_scopes` when making changes.

`TaskPermission`, `MembershipPermission` and `WebhookPermission` are even worse, because they will just return an empty list of scopes, which will later translate to an empty list of permissions, which means that everyone will be permitted to perform the action. This can lead to vulnerabilities like CVE-2024-45393.

Fix this by replacing all `.get` calls with indexing, which will cause a crash if the (action, method) combo is unknown. This breaks one endpoint (`/api/webhooks/events`), which is supposed to be publicly accessible; fix that by disabling authorization for it.

### Checklist

### License

### DIFF
--- a/cvat/apps/analytics_report/permissions.py
+++ b/cvat/apps/analytics_report/permissions.py
@@ -70,7 +70,7 @@ class AnalyticsReportPermission(OpenPolicyAgentPermission):
             {
                 "list": Scopes.LIST,
                 "create": Scopes.CREATE,
-            }.get(view.action, None)
+            }[view.action]
         ]
 
     def get_resource(self):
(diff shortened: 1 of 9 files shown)


nmanovic (approved):
LGTM

PR #8488, opened 2024-09-30
Added readiness and liveness probes

azhavoro:
### Checklist

### License

## Release Notes

- **New Features**
	- Introduced a management command `workerprobe` to check the health of workers in specified queues.
	- Enhanced health monitoring with readiness and liveness probes across multiple components including backend, workers, frontend, OPA, and kvrocks.

- **Chores**
	- Updated Helm chart version from `0.13.2` to `0.14.0`.
	- Modified configuration files to improve process management and health checks.


helm-chart/templates/cvat_backend/worker_analyticsreports/deployment.yml
```diff
@@ -61,6 +61,26 @@ spec:
           {{- with concat .Values.cvat.backend.additionalEnv $localValues.additionalEnv }}
           {{- toYaml . | nindent 10 }}
           {{- end }}
+          {{- if .Values.cvat.backend.worker.readinessProbe.enabled }}
+          readinessProbe:
```
SpecLad, line 65: I don't think it makes sense to define a readiness probe for a worker process. According to the [documentation](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/),

> The kubelet uses readiness probes to know when a container is ready to start accepting traffic. [...] One use of this signal is to control which Pods are used as backends for Services.

Workers do not accept traffic or provide a service, so the concept of readiness is meaningless for them.

helm-chart/templates/cvat_backend/server/deployment.yml
```diff
@@ -65,6 +65,20 @@ spec:
...
+              port: 8080
+            {{- toYaml (omit $localValues.readinessProbe "enabled") | nindent 12 }}
+          {{- end }}
+          {{- if $localValues.livenessProbe.enabled }}
+          livenessProbe:
+            httpGet:
+              path: /api/server/health/
+              port: 8080
+            {{- toYaml (omit $localValues.livenessProbe "enabled") | nindent 12 }}
+          {{- end }}
```
SpecLad, line 81: This liveness probe is IMO not suitable. `/api/server/health/` returns the aggregate status of the whole system, which includes services that the server has no control over (like the database and Redis). The purpose of the liveness probe is to tell Kubernetes when to restart the pod, but it doesn't make sense to restart the server when, e.g., the database is down.

I think that we should follow the documentation here:

> A common pattern for liveness probes is to use the same low-cost HTTP endpoint as for readiness probes, but with a higher failureThreshold. This ensures that the pod is observed as not-ready for some period of time before it is hard killed.

helm-chart/templates/cvat_backend/worker_analyticsreports/deployment.yml
```diff
@@ -61,6 +61,16 @@ spec:
...
+          {{- if .Values.cvat.backend.worker.livenessProbe.enabled }}
+          livenessProbe:
+            exec:
+              command:
+              - python
+              - manage.py
+              - workerprobe
+              - analytics_reports
+            {{- toYaml (omit .Values.cvat.backend.worker.livenessProbe "enabled") | nindent 12 }}
+          {{- end }}
```
SpecLad, line 73: This section is duplicated everywhere. Can you factor it out into a template?

helm-chart/templates/cvat_kvrocks/statefulset.yml
```diff
@@ -59,6 +59,25 @@ spec:
...
           {{- toYaml . | nindent 10 }}
           {{- end }}
+          #https://github.com/apache/kvrocks/blob/unstable/Dockerfile
+          {{- if .Values.cvat.kvrocks.readinessProbe.enabled }}
+          readinessProbe:
+            exec:
+              command:
+              - /bin/sh
+              - -c
+              - ./bin/redis-cli -p 6666 PING | grep -E '(PONG|NOAUTH)' || exit 1
```
SpecLad, line 69: I don't think `|| exit 1` does anything useful here. It just overwrites the previous error code.

helm-chart/templates/cvat_opa/deployment.yml
```diff
@@ -53,6 +53,18 @@ spec:
           env:
           {{- toYaml . | nindent 10 }}
           {{- end }}
+          {{- if .Values.cvat.opa.readinessProbe.enabled }}
+          readinessProbe:
+            tcpSocket:
```
SpecLad, line 58: OPA has an actual [health endpoint](https://www.openpolicyagent.org/docs/latest/rest-api/#health-api), so perhaps you could use that?

helm-chart/values.yaml
```diff
@@ -172,6 +186,10 @@ cvat:
     # -   mountPath: /tmp
     #     name: tmp
     #     subPath: test
+    readinessProbe:
+      enabled: true
+    livenessProbe:
+      enabled: true
```
SpecLad, line 192: Why don't you have other options here, like in the other probe settings?

tests/python/shared/fixtures/init.py
```diff
@@ -240,11 +240,26 @@ def kube_restore_clickhouse_db():
 
 
 def docker_restore_redis_inmem():
-    docker_exec_redis_inmem(["redis-cli", "-e", "flushall"])
+    docker_exec_redis_inmem(
+        [
+            "sh",
+            "-c",
+            "for p in rq:finished:* rq:job:* rq:wip:* rq:finished:* rq:failed:*; "
+            'do redis-cli -e --scan --pattern "$p" | xargs -r redis-cli -e del ; done',
```
SpecLad, line 248: Could you explain what you did here?

SpecLad (changes requested):

tests/python/shared/fixtures/init.py
azhavoro: Previously we reset the entire Redis DBs including RQ related keys about registered workers, which would cause the worker to restart on the next heartbeat, but in fact to pass the tests we only need to clear the keys related with jobs and statuses.
Here I find all keys related to jobs and their status and then delete them without deleting the whole DB

SpecLad: In that case, I don't think you got all the keys that need to be removed. E.g. there's the [scheduled jobs key](https://github.com/rq/rq-scheduler/blob/656898e31c9da2747639173fd5b672506b957f29/rq_scheduler/scheduler.py#L27), the [export cache lock keys](https://github.com/cvat-ai/cvat/blob/5f279771c7c245ec3fefea1778662a10580fbc33/cvat/apps/dataset_manager/util.py#L104), the [blocker keys](https://github.com/cvat-ai/cvat/blob/5f279771c7c245ec3fefea1778662a10580fbc33/cvat/utils/background_jobs.py#L33). There could be others I missed.

In general, I don't think this approach is workable, because it's very easy to forget to update this function when writing code that uses new Redis keys. I think you should do the opposite; keep certain keys and delete everything else. And please add a comment to explain what the code is doing.

cvat/apps/health/management/__init__.py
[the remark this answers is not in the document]
SpecLad: Ping?

SpecLad:
Please add a changelog entry.

SpecLad (approved):

PR #8496, opened 2024-10-01
fix nightly tests

azhavoro:
### How has this been tested?

https://github.com/cvat-ai/cvat/actions/runs/11124390649

### Checklist

### License

## Summary by CodeRabbit

- **Chores**
	- Updated CI-nightly workflow configuration for improved efficiency and organization.
	- Changed image tags for server and UI to reflect development versions.
	- Enhanced logging and error handling in testing processes.

### DIFF
--- a/.github/workflows/schedule.yml
+++ b/.github/workflows/schedule.yml
@@ -158,15 +158,21 @@ jobs:
           docker compose run --rm -v "$PWD:/mnt/src:ro" -w /mnt/src \
             cvat_opa test cvat/apps/*/rules
 
-      - name: REST API and SDK tests
+      - name: Generate SDK
         run: |
           pip3 install -r cvat-sdk/gen/requirements.txt
           ./cvat-sdk/gen/generate.sh
 
-          pip3 install -r ./tests/python/requirements.txt
-          pip3 install -e ./cvat-sdk
-          pip3 install -e ./cvat-cli
+      - name: Install test dependencies, SDK and CLI
+        run: |
+          pip3 install \
+            -r ./tests/python/requirements.txt \
+            -e './cvat-sdk[pytorch]' \
+            -e ./cvat-cli \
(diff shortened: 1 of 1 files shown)


.github/workflows/schedule.yml
```diff
@@ -127,7 +127,7 @@ jobs:
           context: .
           file: ./Dockerfile.ui
           cache-from: type=local,src=/tmp/cvat_cache_ui
-          tags: cvat/ui:latest
+          tags: cvat/ui:dev
```
SpecLad, line 130: Is this image actually used for anything? It seems like the UI doesn't actually get started at any point in this job.

And if it _is_ needed, wouldn't it make more sense to pull the image that's already been built, instead of building it anew?

.github/workflows/schedule.yml
```diff
@@ -147,9 +147,9 @@ jobs:
       - name: Pull CVAT server image
         run: |
           docker pull ${{ steps.meta-server.outputs.tags }}
+          docker tag ${{ steps.meta-server.outputs.tags }} cvat/server:dev
+          #TODO fix tag in Dockerfile.ci
```
SpecLad, line 151: Could you elaborate? What do you want to fix?

azhavoro: tag `local` is only used in this workflow and it seems to only complicate the payplan. Previously prepared images were labeled with `latest` tag and it already led to using wrong images during testing. I only suggest to remove the use of local tag, I think there is no need in it. But I don't want to do it now and suggest to defere it until workflow refactoring in the future, which should solve many problems such as duplicates and others.

SpecLad: FWIW, I don't think using `local` actually needs to complicate the pipeline. We have some code to retag images, but I don't think it's necessary; see #8562.

SpecLad:
I submitted a more radical fix here: #8596. Mind taking a look?

PR #8499, opened 2024-10-01
Use exceptions instead of HTML responses with error statuses

SpecLad:
### Checklist

### License

### DIFF
--- a/cvat/apps/engine/background.py
+++ b/cvat/apps/engine/background.py
@@ -12,7 +12,6 @@ from typing import Any, Callable, Dict, Optional, Union
 import django_rq
 from attrs.converters import to_bool
 from django.conf import settings
-from django.http.response import HttpResponseBadRequest
 from django.utils import timezone
 from django_rq.queues import DjangoRQ
 from rest_framework import serializers, status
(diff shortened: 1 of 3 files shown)


zhiltsov-max:
@SpecLad, if you're working on this, please check if you can improve the situation with output data in the error, asked here: https://github.com/cvat-ai/cvat/pull/8348#issuecomment-2378797877

> Would it be hard to improve formatting for the validation error? Like return only the stuff thats insize string=?. I suppose we have lots of cases now when we can get it. On the second thought I belive its easier to UI to parse it for better UX, we already have some handling for such cases. Maybe not really relevant for now

SpecLad:
@zhiltsov-max I don't understand what that comment is talking about. Is it some specific error? What's `string=`?

klakhov:
@SpecLad 
When I was testing the honeypots patch Ive seen different validation errors. For example this one, when Im trying to create honeypots task from video file:
image

The Idea was to improve the UX of the error. It seems we need to show only stuff that is written inside `string=` part of the error. As everything else seems as unnecessary information to the user

SpecLad:
Ah, I see. I think that exception is coming from inside an RQ job. I do think we have some deficiencies in how those exceptions are presented to the user, but fixing that would be out of scope for this PR.

That said, I find it odd that that particular error would be thrown from an RQ job. It seems like it's just validating input parameters; it would be better to do that in the view _before_ starting the RQ job. Not only is that more efficient, but it should also fix the presentation problem.

zhiltsov-max:
> It seems like it's just validating input parameters; it would be better to do that in the view before starting the RQ job.

Not everything can be validated before processing. For instance, in this case, you'd need to know that the input is video (it's possible to check, sure). It can only be checked after the `Upload-Finished` data request. There are other checks that depend on the actual data provided, like depending on frame count.

SpecLad:
> For instance, in this case, you'd need to know that the input is video (it's possible to check, sure). It can only be checked after the `Upload-Finished` data request.

Okay, that's a fair point.

SpecLad:
I don't have time to finish this PR right now; I'll try to come back to it later.

PR #8551, opened 2024-10-16
Fixed quality reports and immediate feedback on jobs with start frame and step

bsekachev:
### Motivation and context
To address the identified issue, we likely need to store relative frame numbers in self._required_frames.

Explanation:
The only place where self._required_frames is used is in the method:

```
def _is_frame_required(self, frame):
    return self._required_frames is None or frame in self._required_frames
```

This method is called in:

```
def get_included_frames(self):
    return set(
        i for i in self.rel_range
        if not self._is_frame_deleted(i)
        and not self._is_frame_excluded(i)
        and self._is_frame_required(i)
    )

```
Currently, the implementation tries to check whether a relative frame number exists in a set that contains absolute frame numbers, which doesn't work if start_frame or frame_step is defined when the task is created. This causes failures in both quality reports and immediate feedback, as well as possibly other functionalities that rely on get_included_frames.

Another issue with the current code is a logical inconsistency:

```
abs_range = self.abs_range
self._required_frames = set(
     self.abs_frame_id(frame) for frame in self._required_frames
     if frame in abs_range
)
```

Here, abs_frame_id expects a relative frame number and converts it to an absolute frame number, but we first check if the relative frame number from self._required_frames is already in the absolute range, which is incorrect.

### Checklist

### License

## Summary by CodeRabbit

- **New Features**
	- Enhanced logic for managing required frames within job data.
	- Added properties for absolute and relative frame ranges.
	- Implemented a method to return the number of frames in a job.

- **Bug Fixes**
	- Improved frame matching logic for better accuracy.

- **Refactor**
	- Cleaned up code for better readability and maintainability.

### DIFF
--- a/cvat/apps/dataset_manager/bindings.py
+++ b/cvat/apps/dataset_manager/bindings.py
@@ -812,7 +812,8 @@ class JobData(CommonData):
                 self._excluded_frames.update(self.db_data.validation_layout.disabled_frames)
 
         if self._required_frames:
-            self._required_frames = set(frame for frame in self._required_frames if frame in self.rel_range)
+            rel_range = self.rel_range
+            self._required_frames = set(frame for frame in self._required_frames if frame in rel_range)
 
     def __len__(self):
         segment = self._db_job.segment
--- a/cvat/apps/dataset_manager/bindings.py
+++ b/cvat/apps/dataset_manager/bindings.py
@@ -812,11 +812,7 @@ class JobData(CommonData):
                 self._excluded_frames.update(self.db_data.validation_layout.disabled_frames)
 
         if self._required_frames:
-            abs_range = self.abs_range
-            self._required_frames = set(
-                self.abs_frame_id(frame) for frame in self._required_frames
-                if frame in abs_range
-            )
+            self._required_frames = set(frame for frame in self._required_frames if frame in self.rel_range)
 
     def __len__(self):
         segment = self._db_job.segment
--- a/cvat/apps/quality_control/quality_reports.py
+++ b/cvat/apps/quality_control/quality_reports.py
@@ -2317,7 +2317,7 @@ class QualityReportUpdateManager:
                 job.id: JobDataProvider(
                     job.id,
                     queryset=job_queryset,
-                    included_frames=set(job.segment.frame_set) & active_validation_frames,
+                    included_frames=active_validation_frames,
                 )
                 for job in jobs
             }
--- a/cvat/apps/quality_control/quality_reports.py
+++ b/cvat/apps/quality_control/quality_reports.py
@@ -2303,13 +2303,13 @@ class QualityReportUpdateManager:
             if validation_layout.mode == ValidationMode.GT_POOL:
                 task_frame_provider = TaskFrameProvider(task)
                 active_validation_frames = set(
-                    task_frame_provider.get_rel_frame_number(frame)
-                    for frame, real_frame in (
+                    task_frame_provider.get_rel_frame_number(abs_frame)
+                    for abs_frame, abs_real_frame in (
                         Image.objects.filter(data=task.data, is_placeholder=True)
                         .values_list("frame", "real_frame")
                         .iterator(chunk_size=10000)
                     )
-                    if real_frame in active_validation_frames
+                    if task_frame_provider.get_rel_frame_number(abs_real_frame) in active_validation_frames
                 )
 
             jobs: List[Job] = [j for j in job_queryset if j.type == JobType.ANNOTATION]
--- a/cvat/apps/quality_control/quality_reports.py
+++ b/cvat/apps/quality_control/quality_reports.py
@@ -2309,7 +2309,8 @@ class QualityReportUpdateManager:
                         .values_list("frame", "real_frame")
                         .iterator(chunk_size=10000)
                     )
-                    if task_frame_provider.get_rel_frame_number(abs_real_frame) in active_validation_frames
+                    if task_frame_provider.get_rel_frame_number(abs_real_frame)
+                    in active_validation_frames
                 )
 
             jobs: List[Job] = [j for j in job_queryset if j.type == JobType.ANNOTATION]
--- a/changelog.d/20241016_180804_sekachev.bs.md
+++ b/changelog.d/20241016_180804_sekachev.bs.md
@@ -1,0 +1,4 @@
+### Fixed
+
+- Incorrect quality reports and immediate feedback with non default start frame or frame step
+  (<https://github.com/cvat-ai/cvat/pull/8551>)


cvat/apps/dataset_manager/bindings.py
```diff
@@ -812,11 +812,7 @@ def _init_frame_info(self):
                 self._excluded_frames.update(self.db_data.validation_layout.disabled_frames)
 
         if self._required_frames:
-            abs_range = self.abs_range
-            self._required_frames = set(
-                self.abs_frame_id(frame) for frame in self._required_frames
-                if frame in abs_range
-            )
+            self._required_frames = set(frame for frame in self._required_frames if frame in self.rel_range)
```
SpecLad, line 815: Are you sure that `self._required_frames` initially contains relative numbers? It seems that it comes from the `included_frames` constructor parameter, supplied here:

https://github.com/cvat-ai/cvat/blob/49ec1d175d9a9ce106158a7a7936e3c6622d3672/cvat/apps/quality_control/quality_reports.py#L2320

And `job.segment.frame_set` seems to return absolute frame numbers.

bsekachev: Looks like one more bug .. 

On `develop` branch in this place we are trying to find intersection between set with absolute frame numbers `job.segment.frame_set` and relative frame numbers `active_validation_frames`:
image

bsekachev: And the exception occurs:

```
[2024-10-16 17:34:44,154] ERROR rq.worker: [Job quality-check-task-4375-user-1]: exception raised while executing (_check_task_quality)
  File "/home/bsekachev/app.cvat.ai/cvat_enterprise/.env/lib/python3.10/site-packages/rq/worker.py", line 1431, in perform_job
    rv = job.perform()
  File "/home/bsekachev/app.cvat.ai/cvat_enterprise/.env/lib/python3.10/site-packages/rq/job.py", line 1280, in perform
  File "/home/bsekachev/app.cvat.ai/cvat_enterprise/.env/lib/python3.10/site-packages/rq/job.py", line 1317, in _execute
  File "/home/bsekachev/app.cvat.ai/cvat_enterprise/.env/lib/python3.10/site-packages/silk/profiling/profiler.py", line 161, in wrapped_target
    result = target(*args, **kwargs)
  File "/home/bsekachev/app.cvat.ai/cvat_enterprise/cvat/cvat/apps/quality_control/quality_reports.py", line 2259, in _check_task_quality
    return cls()._compute_reports(task_id=task_id)
  File "/home/bsekachev/app.cvat.ai/cvat_enterprise/cvat/cvat/apps/quality_control/quality_reports.py", line 2316, in _compute_reports
    job_data_providers = {
  File "/home/bsekachev/app.cvat.ai/cvat_enterprise/cvat/cvat/apps/quality_control/quality_reports.py", line 2317, in <dictcomp>
    job.id: JobDataProvider(
    return func(*args, **kwds)
  File "/home/bsekachev/app.cvat.ai/cvat_enterprise/cvat/cvat/apps/quality_control/quality_reports.py", line 566, in __init__
    self.job_data = JobData(
  File "/home/bsekachev/app.cvat.ai/cvat_enterprise/cvat/cvat/apps/dataset_manager/bindings.py", line 755, in __init__
    super().__init__(annotation_ir, self._db_task, **kwargs)
  File "/home/bsekachev/app.cvat.ai/cvat_enterprise/cvat/cvat/apps/dataset_manager/bindings.py", line 292, in __init__
    self._init_frame_info()
  File "/home/bsekachev/app.cvat.ai/cvat_enterprise/cvat/cvat/apps/dataset_manager/bindings.py", line 816, in _init_frame_info
    self._required_frames = set(
  File "/home/bsekachev/app.cvat.ai/cvat_enterprise/cvat/cvat/apps/dataset_manager/bindings.py", line 817, in <genexpr>
    self.abs_frame_id(frame) for frame in self._required_frames
  File "/home/bsekachev/app.cvat.ai/cvat_enterprise/cvat/cvat/apps/dataset_manager/bindings.py", line 313, in abs_frame_id
    raise ValueError("Unknown internal frame id %s" % relative_id)
ValueError: Unknown internal frame id 226
```

bsekachev: Additionally, it looks like in case with `gt_pool`, `job.segment.frame_set` returns relative frame numbers
But in this case relative = absolute, because `frame_step` and `start_frame` are lost somewhere :/ 
Maybe it was done intentionally (as does not really make sense during random ordering), probably only @zhiltsov-max knows 

image

image

zhiltsov-max: In the case of gt pool:

- input start frame and frame step make sense only during task creation, so they are dropped after the task is created (otherwise these values would be invalid)
- the gt job segment has the "range" type, opposed to "specific frames" in simple gt jobs. It means it's a regular section of the whole task frame range [0; size), but it has irregular segment size (includes all the pool frames)
- realtive frames are the same as absolute frames (start frame= 0, step = 1 regardless of the user input)

zhiltsov-max (commented):
Probably, this code fragment should be changed instead:

Specifically, https://github.com/cvat-ai/cvat/blob/49ec1d175d9a9ce106158a7a7936e3c6622d3672/cvat/apps/quality_control/quality_reports.py#L2320 seems to be incorrect for simple gt jobs (absolute frames from frame_set are compared with relative numbers from the active set). I think it's better to fix that line instead - probably, by adding some more conversions (e.g. `get_included_frames` result to absolute, then maybe back to relative in `included_frames=`). Overall, it looks like the code here assumes the active frames are absolute numbers.

Not a big problem yet, but it will be better to make sure `real_frame` (which are absolute) is aligned with acive set frames in comparison here: https://github.com/cvat-ai/cvat/blob/49ec1d175d9a9ce106158a7a7936e3c6622d3672/cvat/apps/quality_control/quality_reports.py#L2312 after changes.

bsekachev:
> e.g. get_included_frames result to absolute

`get_included_frames` is implemented in the way that it can't return absolute frames. 
Because it takes frames from `self.rel_range`:

```
def get_included_frames(self):
    return set(
        i for i in self.rel_range
        if not self._is_frame_deleted(i)
        and not self._is_frame_excluded(i)
        and self._is_frame_required(i)
    )
```

Additionally, I think extra conversions only make code more difficult to understand. Thus, both `get_included_frames` and `included_frames` should be absolute or relative (need to decide), hovewer I believe that based on my findings I wrote in the PR descriptions, it should be relative.

bsekachev:
@zhiltsov-max 

One more reason to keep relative is that `included_frames` then used in annotation manager that knows nothing about absolute frames at all

image

zhiltsov-max:
Yes, I am just saying it's used later incorrectly, like of it returned absolute frames. So the suggestion here is to maybe convert the returned values to absolute ones instead of changing the function, in the line https://github.com/cvat-ai/cvat/blob/49ec1d175d9a9ce106158a7a7936e3c6622d3672/cvat/apps/quality_control/quality_reports.py#L2300.

I agree to use relative frames as much as possible.

zhiltsov-max:
> Hovewer we may add rel to abs conversion into corresponding condition just to make logic better. Do I understand you correct?

Yes.

bsekachev:
@zhiltsov-max Updated

cvat/apps/quality_control/quality_reports.py
```diff
@@ -2303,21 +2303,22 @@ def _compute_reports(self, task_id: int) -> int:
...
+                    in active_validation_frames
                 )
 
             jobs: List[Job] = [j for j in job_queryset if j.type == JobType.ANNOTATION]
             job_data_providers = {
                 job.id: JobDataProvider(
                     job.id,
                     queryset=job_queryset,
-                    included_frames=set(job.segment.frame_set) & active_validation_frames,
+                    included_frames=active_validation_frames,
```
SpecLad, line 2321: I don't understand what you did here; why did you remove the intersection?

bsekachev: Intersection is implemented inside JobDataProvider class
This intesection was incorrect because one set is with absolute frames, another one with relative frames

bsekachev: https://github.com/cvat-ai/cvat/blob/63e43cad48a1d8da8e861a8555a924a173c0fbe0/cvat/apps/dataset_manager/bindings.py#L816

SpecLad (approved):

zhiltsov-max (commented):
Lgtm, but I can't really test this at the moment

PR #8554, opened 2024-10-17
Hide mask during editing

klakhov:
### Motivation and context

Added ability to hide a mask during editing
hide-mask

### Checklist

### License

## Release Notes

- **New Features**
  - Added the ability to hide edited objects in the canvas configuration.
  - Enhanced drawing functionality with visibility controls for shapes.
  - Introduced new methods for managing the state of edited annotations.

- **Bug Fixes**
  - Improved handling of drawing actions based on visibility states.

- **Documentation**
  - Updated interfaces and action types to reflect new features and state management capabilities.

- **Chores**
  - Refined component logic to support new visibility state management.


cvat-core/src/annotations-collection.ts
```diff
@@ -1045,6 +1046,7 @@ export default class Collection {
                             z_order: 0,
                             outside: element.outside || false,
                             occluded: element.occluded || false,
+                            hidden: state.hidden || false,
```
bsekachev, line 1049: element.hidden?

cvat-core/src/annotations-collection.ts
```diff
@@ -1026,6 +1026,7 @@ export default class Collection {
                         label_id: state.label.id,
                         outside: state.outside || false,
                         occluded: state.occluded || false,
+                        hidden: state.hidden || false,
```
bsekachev, line 1029: I am not sure that creating a new object already hidden makes sense as it confuses the user 
They press N and expect a new object to be appear, I think.
If the object is hidden, they may think that something is going wrong and start drawing it again  (especially when there are a lot of objects in the sidebar, newly added is not visible in the list)

cvat-ui/src/actions/annotation-actions.ts
```diff
@@ -1608,3 +1611,84 @@ export function restoreFrameAsync(frame: number): ThunkAction {
...
+        if (canvas && editedStateHidden) {
+            (canvas as Canvas).configure({
+                hideEditedObject: false,
+            });
+        }
+
+        dispatch({
+            type: AnnotationActionTypes.RESET_EDITED_STATE,
+            payload: {},
+        });
```
bsekachev, line 1660: `dispatch(resetEditedState())`?

cvat-ui/src/components/annotation-page/canvas/views/canvas2d/brush-tools.tsx
```diff
@@ -103,6 +104,10 @@ function BrushTools(): React.ReactPortal | null {
         }
     }, [setCurrentTool, blockedTools['polygon-minus']]);
 
+    const hideMask = useCallback((hide: boolean) => {
+        dispatch(changeHideEditedStateAsync(hide));
+    }, [dispatch]);
```
bsekachev, line 109: `dispatch` is a const, so, I am not sure we need this wrapper

cvat-ui/src/components/annotation-page/canvas/views/canvas2d/brush-tools.tsx
```diff
@@ -75,6 +75,7 @@ function BrushTools(): React.ReactPortal | null {
     const config = useSelector((state: CombinedState) => state.annotation.canvas.brushTools);
     const canvasInstance = useSelector((state: CombinedState) => state.annotation.canvas.instance);
     const labels = useSelector((state: CombinedState) => state.annotation.job.labels);
+    const editedState = useSelector((state: CombinedState) => state.annotation.editing);
```
bsekachev, line 78: Please, combine useSelector's together with `shallowEqual`

cvat-ui/src/reducers/index.ts
```diff
@@ -694,6 +694,12 @@ export enum NavigationType {
     EMPTY = 'empty',
 }
 
+export interface EditingState {
+    shapeType: ShapeType | null;
+    editedStateInstance: ObjectState | null;
+    editedStateHidden: boolean;
```
bsekachev, line 700: I will propose a little renaming, like: 
```suggestion
    objectState: ObjectState | null;
    isHidden: boolean;
```

cvat-ui/src/components/annotation-page/canvas/views/canvas2d/canvas-wrapper.tsx
```diff
@@ -146,13 +149,16 @@ interface DispatchToProps {
     onGetDataFailed(error: Error): void;
     onCanvasErrorOccurred(error: Error): void;
     onStartIssue(position: number[]): void;
+    onUpdateEditedObject(objectType: ShapeType | null, editedState: ObjectState | null): void;
```
bsekachev, line 152: It seems, the first argument can't be null

```suggestion
    onUpdateEditedObject(objectType: ShapeType, editedState: ObjectState | null): void;
```

cvat-ui/src/actions/annotation-actions.ts
```diff
@@ -1608,3 +1611,84 @@ export function restoreFrameAsync(frame: number): ThunkAction {
...
+                });
+            }
+
+            dispatch({
+                type: AnnotationActionTypes.HIDE_EDITED_STATE,
+                payload: {
+                    hide: editedStateInstance.hidden,
+                },
+            });
+        }
```
bsekachev, line 1635: I think we have three actions `changeHideEditedStateAsync`, `updateEditedStateAsync`, `resetEditedStateAsync` doing the same thing (updating the same key in store).

Probably we may combine them to reduce the code duplication. Let it accept `shapeType`, `objectState` and `isHidden`. And use these arguments to update store and objectState properly

cvat-ui/src/components/annotation-page/canvas/views/canvas2d/canvas-wrapper.tsx
bsekachev: At least now you do not call it with `null` here.

cvat-ui/src/components/annotation-page/canvas/views/canvas2d/canvas-wrapper.tsx
```diff
@@ -649,6 +664,7 @@ class CanvasWrapperComponent extends React.PureComponent<Props> {
         state.rotation = state.rotation || 0;
         state.occluded = state.occluded || false;
         state.outside = state.outside || false;
+        state.hidden = state.hidden || (editedStateHidden && workspace !== Workspace.SINGLE_SHAPE);
```
bsekachev, line 667: see my comment in `cvat-core`

cvat-ui/src/components/annotation-page/single-shape-workspace/single-shape-sidebar/single-shape-sidebar.tsx
```diff
@@ -185,6 +188,12 @@ const componentShortcuts = {
         sequences: ['del', 'shift+del'],
         scope: ShortcutScope.SINGLE_SHAPE_ANNOTATION_WORKSPACE,
     },
+    HIDE_MASK_SINGLE_SHAPE: {
+        name: 'Hide mask',
+        description: 'Hide edited mask',
```
bsekachev, line 193:
```suggestion
        description: 'Hide currently edited mask',
```

cvat-ui/src/containers/annotation-page/standard-workspace/objects-side-bar/object-buttons.tsx
```diff
@@ -44,6 +46,7 @@ function mapStateToProps(state: CombinedState, own: OwnProps): StateToProps {
             player: {
                 frame: { number: frameNumber },
             },
+            editing: editedState,
```
bsekachev, line 49: It is better to pass exact boolean property of the object you only use in this component. Potentially it will cause less renderings

cvat-ui/src/containers/annotation-page/standard-workspace/objects-side-bar/object-buttons.tsx
```diff
@@ -29,11 +29,13 @@ interface StateToProps {
     outsideDisabled: boolean;
     hiddenDisabled: boolean;
     keyframeDisabled: boolean;
+    editedState: EditingState,
```
bsekachev, line 32: As it comes from global store and we anyway use `mapStateToProps` here, I do not see the reason to pass as own props

bsekachev (commented):
I think we should remove `shapeType` and `editedStateHidden` from `editing` in global store, because the implementation only was requested for masks and it works only for masks. 

You already have `shapeType` in `drawing` state (when we speak about initial creating of the mask), do not need to duplicate it. You may use existing value. 

Also, small bug was found:
1. In brush tools switch to polygon drawing
2. Draw part of polygon
3. Press hide, press show
4. Drawn part of polygon lost

cvat-ui/src/reducers/index.ts
bsekachev: Renaming to `isHidden` is not relevant considering my general comment on review

cvat-ui/src/actions/annotation-actions.ts
bsekachev: Not really relevant considering my other comment with proposal to combine actions

cvat-ui/src/actions/annotation-actions.ts
bsekachev: Considering my general review comment it should only accept `objectState: ObjectState | null`

klakhov:
> I think we should remove `shapeType` and `editedStateHidden` from `editing` in global store, because the implementation only was requested for masks and it works only for masks.
> 
> You already have `shapeType` in `drawing` state (when we speak about initial creating of the mask), do not need to duplicate it. You may use existing value.
> 

I dont think we can really remove `shapeType` and `editedStateHidden`. There are cases when we need them:
1. `shapeType` - the problem with `shapeType` in `drawing` is that it is not reset after finishing drawing something. I suppose this value is used for re-draw feature. In shortcut handler we want to know if we need to dispatch changeHideEditedObject by this value. We can try to rely on `editedState` but we dont have it in case if new mask is drawn. 
2. `editedStateHidden` - when we have a mask we can use `objectState.hidden` to keep the state of current object,  but when we draw a new mask we dont have a objectState, so we dont know state of `hidden` variable. 

It seems we can create `dummy` ObjectState to keep the state when mask is not yet drawn, but I think it feels less convinient. Also when we are updating the hidden state of edited mask, we want to dispatch `updateAnnotationsAsync` to update the original objectState. In case of dummy object we dont need that so we need to know the difference somehow

cvat-core/src/annotations-collection.ts
klakhov: Hm, maybe it is for new objects. 
But what about edited ones? Should this hide feature affect state.hidden at all then?

bsekachev:
> shapeType - the problem with shapeType in drawing is that it is not reset after finishing drawing something

The store also have information about whether drawing is on or off currently, is not it? 
Just putting any information related to drawing to `editing` state looks not exactly logically correct, because drawing and editing are different actions. Additionally, as I mentioned, it is just duplication as we have shapeType in both places now. 

> editedStateHidden - when we have a mask we can use objectState.hidden to keep the state of current object, but when we draw a new mask we dont have a objectState, so we dont know state of hidden variable.

Partially I agree with that, but if we are talking about **editing**, we always have ObjectState (something what we edit). But when we are talking about drawing, as mentioned before, we should not keep related information about drawing process in `editing` store.

cvat-ui/src/actions/annotation-actions.ts
klakhov: Now we have two actions: `updateEditedStateAsync` for manipulation of edited state object, another - `changeHideActiveObjectAsync` to change the hide flag. I put the hide flag in `canvas` part of the store.

cvat-ui/src/containers/annotation-page/standard-workspace/objects-side-bar/object-buttons.tsx
klakhov: We dont pass new value as own prop, Im not sure what you mean here

bsekachev (approved):
LGTM

PR #8575, opened 2024-10-22
Fix UI crush on failed request status check

klakhov:
### Motivation and context

When a status check for `request` is failing for any reason, eg something happened on the server and the request failed with 50X error. The requests page crushes with an error: `Cannot read property 'target' of undefined` :
image

Now the error message is shown:
image

### Checklist

### License

## Release Notes

- **New Features**
	- Enhanced request handling with improved initialization and error management for tasks and jobs.
	- Added `listenToCreate` method in the Task class for better task creation management.
	- Introduced `generateInitialRequest` function to standardize request initialization across imports and exports.

- **Bug Fixes**
	- Improved clarity of status messages for failed requests in the RequestCard component.

- **Documentation**
	- Updated method signatures across various action files for better clarity and consistency.

- **Chores**
	- Refined import statements and type exports in the core wrapper to streamline functionality.


bsekachev (commented):
The logic surrounding generateInitialRequest seems a bit unclear.

From what I understand, it was introduced to handle cases where no response is received from the server, making it difficult to properly initialize the Request class.

However, I believe this approach may not be the best solution. It seems we are duplicating logic that the server already handles. Thus, we should ensure that generateInitialRequest mirrors the server API’s behavior.

Additionally, this logic is implemented outside of cvat-core and is duplicated across various actions in the cvat-ui module. It also introduces options like getInitialRequest in the Task.save() method, which makes the implementation less intuitive.

Instead, I’d suggest a simpler approach for handling situations where we can’t retrieve the status:

1. We can use initialRequest when available to initialize certain fields, which is fine.
2. If it’s unavailable and an exception occurs, we should properly handle the error, something is not handled in rendering logic if currently we see the error (we do not need to display Request object in the list because we do not actually have it)
3. As a temporary workaround for server issues, we could attempt to retry the request on the server-proxy level, perhaps with a timeout and a limited number of retries.

bsekachev:
Yes, maybe:

- [502 Bad Gateway](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/502)
- [503 Service Unavailable](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503)
- [504 Gateway Timeout](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/504)

klakhov:
> 1. We can use `initialRequest` when available to initialize certain fields, which is fine.

Implemented.

> 2. If it’s unavailable and an exception occurs, we should properly handle the error. Something is not handled in rendering logic if currently we see the error (we do not need to display `Request` object in the list because we do not actually have it).

The issue was actually with `cvat-core`, where the `Request` object was being created without the required fields. Now, the UI will only update when we receive a `Request` object from the server with all necessary fields. If an error occurs and we don’t have the `Request` object, an error notification will be displayed, and it won’t appear in the requests list.

> 3. As a temporary workaround for server issues, we could attempt to retry the request on the server-proxy level, perhaps with a timeout and a limited number of retries.

Added

bsekachev (commented):
LGTM

bsekachev (approved):

PR #8634, opened 2024-11-01
Add a quality setting for avoiding using bbox in point group matching

zhiltsov-max:
### Motivation and context

In some cases point groups do not represent a single object or a bounding box attached to the group does not represent the object. A new quality check option is added for such cases. The new option allows to control whether to use the bbox for point group comparisons or not.

- Added a new quality setting to enable or disable using of group bbox for point group comparisons
- Improved OKS sigma description

### Checklist

### License

## Release Notes

- **New Features**
  - Introduced a quality setting allowing users to compare point groups without bounding boxes, enhancing flexibility in point comparison methods.
  - Added a new checkbox in the Quality Settings Form for toggling the use of image space in point group comparisons, complete with a tooltip for guidance.

- **Bug Fixes**
  - Improved handling of point comparisons by integrating the new image space setting into existing logic, ensuring accurate distance calculations.

- **Documentation**
  - Updated changelog to reflect new features and settings related to point group comparisons.


cvat/apps/quality_control/serializers.py
```diff
@@ -81,6 +81,7 @@ class Meta:
             "max_validations_per_job",
             "iou_threshold",
             "oks_sigma",
+            "use_image_space_for_point_group_comparisons",
```
nmanovic, line 84: can we find a better name for the parameter? 43 symbols and the whole sentence looks too much.

cvat/apps/quality_control/models.py
```diff
@@ -205,6 +205,8 @@ class QualitySettings(models.Model):
 
     low_overlap_threshold = models.FloatField()
 
+    use_bbox_size_for_points = models.BooleanField()
```
SpecLad, line 208: Have you considered using an enum for this? E.g. call it `points_size_method` and have the values be `bbox_size` and `image_size`. Neither of these methods seems particularly "true" or "false" to me, so perhaps an enum would be more understandable?

zhiltsov-max: It's hard to say currently if there are going to be more methods in future. Basically, you can represent any boolean as a enum. I think currently it's premature.

SpecLad: > It's hard to say currently if there are going to be more methods in future.

I didn't say anything about more methods. I'm saying that the semantics don't seem boolean-like here.

> Basically, you can represent any boolean as a enum.

Certainly, but sometimes a boolean makes the meaning clear, and sometimes not. E.g., for `compare_groups` a boolean makes sense: if it's true, you compare groups, and if it's false, you don't.

Here you do something in either case, so why should the "bbox size" method be true and "image size" be false? It seems arbitrary. Furthermore, it's impossible to determine what `use_bbox_size_for_points=False` means without consulting the documentation.

zhiltsov-max: What do you think of `point_size_base` or `point_distance_base` for the enum name? The members are `image_size`, `group_bbox_size`.

zhiltsov-max: Updated to `point_size_base`. This determines the scale factor for the relative radius (configured in `oks_sigma`) of each point in the points annotation or in a group of such.

SpecLad (approved):

cvat/apps/quality_control/serializers.py
```diff
@@ -115,10 +116,25 @@ class Meta:
             """,
             "oks_sigma": """
                 Like IoU threshold, but for points.
-                The percent of the bbox area, used as the radius of the circle around the GT point,
+                The percent of the bbox side, used as the radius of the circle around the GT point,
```
SpecLad, line 119: FWIW, I'm not sure "side" is appopriate here, seeing as a box has two different side lengths. Maybe something like "geometric mean of bbox sides" would be better? Up to you.

zhiltsov-max: Ok, sounds good.

PR #8652, opened 2024-11-06
Allow matching empty frames in quality checks

zhiltsov-max:
### Motivation and context

Depends on #8634

Added a quality check option to consider frames matching, if both GT and job annotations have no annotations on a frame. This affects quality metrics and total counts in reports, but confusion matrices stay unchanged. This allows to use both positive and negative validation frames.

### How has this been tested?

Unit tests

### Checklist

### License

## Release Notes

- **New Features**
	- Introduced a quality setting for comparing point groups without bounding boxes.
	- Added an option to consider empty frames as matching in quality checks.
	- Enhanced quality settings form with new options for `matchEmptyFrames` and `useBboxSizeForPoints`.

- **Bug Fixes**
	- Corrected property name from `peoject_id` to `project_id` in the API quality reports filter.

- **Documentation**
	- Updated API schemas to include new quality settings properties. 

These changes improve the flexibility and accuracy of quality assessments within the application.


changelog.d/20241101_140759_mzhiltso_compare_point_groups_in_image_space.md
```diff
@@ -0,0 +1,4 @@
+### Added
+
+- A quality setting to compare point groups without using bbox
+  (<https://github.com/cvat-ai/cvat/pull/8634>)
```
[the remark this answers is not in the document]
zhiltsov-max: This is added in 20241106_170626_mzhiltso_match_empty_frames.md

cvat/apps/quality_control/migrations/0004_qualitysettings_use_bbox_size_for_points.py
```diff
@@ -0,0 +1,19 @@
...
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="qualitysettings",
+            name="use_bbox_size_for_points",
+            field=models.BooleanField(default=True),
+            preserve_default=False,
+        ),
+    ]
```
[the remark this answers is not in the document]
zhiltsov-max: So, after the existing records get an initial value during migration, they will have the value set, aren't they? New records will have the value initialized on the server side.

cvat/apps/quality_control/models.py
```diff
@@ -205,6 +205,8 @@ class QualitySettings(models.Model):
 
     low_overlap_threshold = models.FloatField()
 
+    use_bbox_size_for_points = models.BooleanField()
+
```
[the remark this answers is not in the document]
zhiltsov-max: The default value for the field is provided in the model constructor.

tests/python/shared/assets/quality_settings.json
```diff
@@ -14,13 +14,15 @@
       "line_orientation_threshold": 0.1,
       "line_thickness": 0.01,
       "low_overlap_threshold": 0.8,
+      "match_empty_frames": false,
```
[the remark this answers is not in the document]
zhiltsov-max: The option is checked in one of the `test_settings_affect_metrics` test parametrizations.

cvat/apps/quality_control/quality_reports.py
```diff
@@ -214,6 +217,13 @@
     panoptic_comparison: bool = True
     "Use only the visible part of the masks and polygons in comparisons"
 
+    match_empty_frames: bool = False
+    """
+    Consider unannotated (empty) frames as matching. If disabled, quality metrics, such as accuracy,
+    will be 0 if both GT and DS frames have no annotations. When enabled, they will be 1 instead.
+    This will also add virtual annotations to empty frames in the comparison results.
+    """
```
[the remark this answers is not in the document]
zhiltsov-max: The existing code already initializes all the known values, including this one.

cvat/apps/quality_control/quality_reports.py
```diff
@@ -1525,6 +1544,7 @@
             panoptic_comparison=settings.panoptic_comparison,
             iou_threshold=settings.iou_threshold,
             oks_sigma=settings.oks_sigma,
+            use_bbox_size_for_points=settings.use_bbox_size_for_points,
```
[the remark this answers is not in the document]
zhiltsov-max: There is a comment about this in the line 1551, it's handled differently.

cvat-core/src/quality-settings.ts
```diff
@@ -197,6 +217,7 @@ export default class QualitySettings {
         const result: SerializedQualitySettingsData = {
             iou_threshold: this.#iouThreshold,
             oks_sigma: this.#oksSigma,
+            use_bbox_size_for_points: this.#useBboxSizeForPoints,
```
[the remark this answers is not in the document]
zhiltsov-max: The schema is updated.

cvat/apps/quality_control/quality_reports.py
```diff
@@ -1972,8 +1979,16 @@ def _find_closest_unmatched_shape(shape: dm.Annotation):
             gt_label_idx = label_id_map[gt_ann.label] if gt_ann else self._UNMATCHED_IDX
             confusion_matrix[ds_label_idx, gt_label_idx] += 1
 
+        if self.settings.match_empty_frames and not gt_item.annotations and not ds_item.annotations:
+            # Add virtual annotations for empty frames
```
SpecLad, line 1983: What's the purpose of these virtual annotations?

zhiltsov-max: Basically, to break math and code less and get the expected values from `accuracy`, `precision`, and `recall` here and in aggregated reports (job report, task report).

cvat/apps/quality_control/quality_reports.py
```diff
@@ -2015,9 +2030,8 @@ def _make_zero_confusion_matrix(self) -> Tuple[List[str], np.ndarray, Dict[int,
 
         return label_names, confusion_matrix, label_id_idx_map
 
-    @classmethod
-    def _generate_annotations_summary(
-        cls, confusion_matrix: np.ndarray, confusion_matrix_labels: List[str]
+    def _compute_annotation_summary(
```
SpecLad, line 2033:
```suggestion
    def _compute_annotations_summary(
```

This would be consistent with the result type name.

cvat/apps/quality_control/quality_reports.py
```diff
@@ -2082,19 +2108,52 @@ def generate_report(self) -> ComparisonReport:
...
+        if self.settings.match_empty_frames and empty_frame_count:
+            # Add virtual annotations for empty frames,
+            # they are not included in the confusion matrix
+            annotation_summary.valid_count += empty_frame_count
+            annotation_summary.total_count += empty_frame_count
+            annotation_summary.ds_count += empty_frame_count
+            annotation_summary.gt_count += empty_frame_count
+
+        # Cannot be computed in accumulate()
+        annotation_components.shape.mean_iou = np.mean(mean_ious)
```
SpecLad, line 2140: This doesn't seem right. A mean of means is not necessarily the same as the mean of all underlying values.

(It also seems unrelated to the topic of the PR.)

cvat/apps/quality_control/quality_reports.py
```diff
@@ -2082,19 +2108,52 @@ def generate_report(self) -> ComparisonReport:
...
+            confusion_matrix, confusion_matrix_labels
+        )
+
+        if self.settings.match_empty_frames and empty_frame_count:
+            # Add virtual annotations for empty frames,
+            # they are not included in the confusion matrix
+            annotation_summary.valid_count += empty_frame_count
+            annotation_summary.total_count += empty_frame_count
+            annotation_summary.ds_count += empty_frame_count
+            annotation_summary.gt_count += empty_frame_count
```
SpecLad, line 2137: I'm not sure if this is really a useful thing to do here. Imagine a situation where there are 100 frames, frame 1 has 1 valid annotations and 2 total annotations, every other frame is empty.

In this case, with `match_empty_frames` off:

* frame 1 has accuracy 50%, others have 0%.
* total accuracy is 50%.

With `match_empty_frames` on:

* frame 1 has accuracy 50%, others have 100%.
* total accuracy is (1+99)/(2+99) = 99%.

Do you think this jump in total accuracy would be expected by the user?

cvat/apps/quality_control/quality_reports.py
SpecLad: I see, but it seems like bad UX. You choose to match empty frames, and you inexplicably get fake totals. Moreover, if I understand the code correctly, with this change the totals no longer match the confusion matrix, which seems like it could cause more confusion and math errors down the line.

Is this solution really that much less disruptive than fudging the metric formulas to return 1 instead of 0?

> It's not totally nonsense, as an empty annotation can be considered a frame annotation by itself.

This is a valid point, but you're not implementing that consistently. In your implementation, this "empty annotation" only appears when both GT and DS frames are empty. To do this consistently, you'd need to increase the GT count whenever the GT frame is empty, and the DS count whenever the DS frame is empty (and the valid count when both are).

You could also resolve the inconsistency between the confusion matrix and the totals by adding another row/column to the matrix specifically for these "empty annotations".

This would resolve the consistency issues, and slightly improve the UX issue, since you could see in the report where the extra annotation is coming from. But frankly, it still seems easier to me to just fudge the metrics.

cvat/apps/quality_control/quality_reports.py
zhiltsov-max: It's how it worked before this PR, I didn't plan to change anything here. Also, we don't have too many options here on how to compute the value.

cvat/apps/quality_control/quality_reports.py
zhiltsov-max: It's hard to say about the expectations, to me the user just works in one of the modes - they either consider empty frames as annotated or not. But the second value seems to be more correct, so maybe something should be done for the case with the option disabled.

cvat/apps/quality_control/quality_reports.py
```diff
@@ -1972,8 +1979,16 @@ def _find_closest_unmatched_shape(shape: dm.Annotation):
             gt_label_idx = label_id_map[gt_ann.label] if gt_ann else self._UNMATCHED_IDX
             confusion_matrix[ds_label_idx, gt_label_idx] += 1
 
+        if self.settings.match_empty_frames and not gt_item.annotations and not ds_item.annotations:
+            # Add virtual annotations for empty frames
+            valid_labels_count = 1
+            total_labels_count = 1
+
+            valid_shapes_count = 1
+            total_shapes_count = 1
```
SpecLad, line 1988: Shouldn't `ds_shapes_count` and `gt_shapes_count` be updated too?

cvat/apps/quality_control/quality_reports.py
SpecLad: Ah, okay, I didn't see it was moved from a different section.

> Also, we don't have too many options here on how to compute the value.

FWIW, I would imagine you could change the `mean_iou` member of `ComparisonReportAnnotationShapeSummary` to `sum_iou`, and compute the mean on-demand instead.

cvat/apps/quality_control/quality_reports.py
zhiltsov-max: Thanks, updated. I guess the idea was that these are not shapes exactly, but if we add precision and recall in the shape summary section, the metrics will be broken.

SpecLad (approved):

Issue #8606, opened 2024-10-29
Restrict access to auto-labelling models


rragundez:
Hi, I am interested on segregating the Nuclio models that appear in CVAT to some organizations only. I have already changed some permissions in the CVAT source code so I'm kinda familiar with it, but some help on where to look would be very much appeciated.

In short, I need to be able to deploy models in Nuclio but make them accesible to only some organization and not to everyone using the application. 

Thanks in advace!

rragundez:
any type of context or initial idea I can follow up on would be greatly appreciated

rragundez:
Solved by #8663

PR #8663, opened 2024-11-07
Filter nuclio functions based on organization

rragundez:
Filter nuclio functions based on organization. This allows organizations to produce their models and making only available to their CVAT org. 
This is quite helpful when a company has different departments for data science and one of them is related to R&D for example and their models might not be allowed to be shared while in-development as they might expose secret projects that a team might be working on.

### How has this been tested?
I tested in my own local deployment of CVAT and a colleague also tested the changes. By deploying nuclio models with the respective labels as mocked in the tests.

Solves #8606 

## Summary by CodeRabbit

- **New Features**
	- Introduced an optional organization filter for listing lambda functions, enhancing the ability to view functions associated with specific organizations.
	- Added structured "labels" fields to various models for better categorization and identification.

- **Bug Fixes**
	- Improved API response validation for organization-specific requests, ensuring correct status codes and function ID matching.

- **Tests**
	- Added a new test case to verify the API behavior when listing lambda functions with an organization ID, enhancing test coverage.

### DIFF
--- a/cvat/apps/lambda_manager/tests/assets/functions.json
+++ b/cvat/apps/lambda_manager/tests/assets/functions.json
@@ -26,7 +26,8 @@
           "name": "Person reidentification",
           "spec": "",
           "type": "reid"
-        }
+        },
+        "labels": {}
       },
       "spec": {
         "description": "Person reidentification model for a general scenario"
@@ -44,6 +45,9 @@
           "name": "Person reidentification",
           "spec": "",
           "type": "reid"
+        },
+        "labels": {
+          "test-label": "test-label-value"
         }
       },
       "spec": {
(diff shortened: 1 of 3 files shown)


bsekachev:
Hello, 

- The proposed approach does not use common authorization system (open policy agent and rego rules).
- Storing organization slug in labels specification in general is not reliable and not convenient. 
- You only protected list method, but I still can use other HTTP methods. 

And currently we are not going to add authorization to use serverless functions. They are global

rragundez:
@bsekachev hi, thanks for the feedback, can you point out on how I will go about using the auth with the rego rules? Any documentation or tips at what to look at will help.
I'm a little bit familiar with the files as we needed to modify them for internal purposes for our R&D teams.

Also which other methods are available to call the model? You mean the invoke one? Are there others?

About your point about just being global, I think you can consider that this can help the community especially in those cases where CVAT is used by various teams that need separation due to "secret" projects. If not the only answer is to maintain an instance of CVAT per team, which I don't see why not try to make a non-breaking change that enables the functionality.

PR #8676, opened 2024-11-11
not prefetching images when not needed

Eldies:
### Motivation and context

While importing annotations to task, all jobs of the task are loaded from db to ram. Related data is prefetched, specifically all image models which belong to the task.
As a result, each job holds its own copy of all the image models.

If there are a lot of jobs and a lot of images in the task, a lot of memory can be occupied.
And images are not utilised on annotations import/delete. Hence - do not prefetch images in these cases.

### Checklist

### License

## Summary by CodeRabbit

- **New Features**
	- Enhanced job retrieval process with improved error handling.
	- Introduced a mechanism for custom querysets in job initialization.
  
- **Bug Fixes**
	- Improved robustness in job fetching to prevent failures when jobs are not found.

- **Refactor**
	- Updated logic in the `JobAnnotation` class for clearer control flow and initialization.


Marishka17:
@Eldies, Could you please provide the difference in memory usage and number of db queries (before/after the patch)?

Marishka17 (commented):
Generally, it works well for me :thumbsup:
I have only a few small comments.

cvat/apps/dataset_manager/task.py
```diff
@@ -103,18 +110,15 @@ def add_prefetch_info(cls, queryset):
...
-                    Prefetch('images', queryset=models.Image.objects.order_by('frame'))
-            )),
+            Prefetch('segment__task__data', queryset=task_data_queryset),
 
             Prefetch('segment__task__label_set', queryset=label_qs),
             Prefetch('segment__task__project__label_set', queryset=label_qs),
         )
 
-    def __init__(self, pk, *, is_prefetched=False, queryset=None):
+    def __init__(self, pk, *, is_prefetched: bool = False, queryset: QuerySet = None, prefetch_images: bool = True):
```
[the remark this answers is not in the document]
zhiltsov-max: I can't say, this conditional logic was added in https://github.com/cvat-ai/cvat/commit/ba74709c40c08d31db59814bd4f9abec51ee3282.

> For instance, why don't we lock the row when updating job annotations directly?

I'm not sure what you mean here. Could you phrase it in more detail? I can guess that the whole update was supposed to be a single request, so that no lock is needed. Maybe there was some deadlock somewhere.

cvat/apps/dataset_manager/task.py
```diff
@@ -1018,7 +1022,7 @@ def put_job_data(pk, data):
 @plugin_decorator
 @transaction.atomic
 def patch_job_data(pk, data, action):
```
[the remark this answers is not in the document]
Eldies: checked them and set prefetch_images=True when needed

cvat/apps/dataset_manager/task.py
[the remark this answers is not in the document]
Eldies: set prefetch_images=False by default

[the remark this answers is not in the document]
Eldies: as far as I understand from https://github.com/cvat-ai/cvat/pull/5160/files#diff-ed7ab63c7c54f5d87f982240b298e7830de8e01da28b819779b44bc601db6f7bR74
`is_prefetched` actually is related to my `prefetch_images` - it was to determine whether images should be prefetched. But somewhere along the way this behaviour was broken.

select_for_update was there earlier, from (at least) https://github.com/cvat-ai/cvat/commit/ae6a4891e3fd874331dda661572aceb09a49fdb7 (in cvat/apps/engine/annotation.py) and was removed for the case when images are to be prefetched

cvat/apps/dataset_manager/task.py
```diff
@@ -103,18 +110,15 @@ def add_prefetch_info(cls, queryset):
...
-                    Prefetch('images', queryset=models.Image.objects.order_by('frame'))
-            )),
+            Prefetch('segment__task__data', queryset=task_data_queryset),
 
             Prefetch('segment__task__label_set', queryset=label_qs),
             Prefetch('segment__task__project__label_set', queryset=label_qs),
         )
 
-    def __init__(self, pk, *, is_prefetched=False, queryset=None):
+    def __init__(self, pk, *, is_prefetched: bool = False, queryset: QuerySet = None, prefetch_images: bool = False):
```
zhiltsov-max, line 119:
```suggestion
    def __init__(self, pk, *, is_prefetched: bool = False, queryset: QuerySet | None = None, prefetch_images: bool = False):
```

cvat/apps/dataset_manager/task.py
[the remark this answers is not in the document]
Eldies: Since in most cases jobs are not locked, and `is_prefetched` is only  used in `TaskAnnotation.init_from_db`, I believe `is_prefetched` can be dropped.

Eldies:
Measuring consumed memory with this [memray.patch](https://github.com/user-attachments/files/17815206/memray.patch)
With the task and annotations attached to the issue (100 jobs, 4098 images)

1. run locally (in docker)
2. import task from backup
3. apply patch
4. `python manage.py shell`
```
In [1]: import cvat
In [2]: from django.db import reset_queries, connections
In [3]: cvat.apps.dataset_manager.task.delete_task_data(<task id>)
In [4]: reset_queries()
In [5]: cvat.apps.dataset_manager.task.import_task_annotations(<path to annotations archive>, <task id>, 'Segmentation mask 1.1', True)
In [6]: sum([len(conn.queries) for conn in connections.all()])
```

On develop:
Peak memory usage: 2.5 GiB
Number of queries: 12287

With this PR:
Peak memory usage: 2.0 GiB
Number of queries: 11287

cvat/apps/dataset_manager/task.py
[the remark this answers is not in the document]
Marishka17: > Since is_prefetched is only used in TaskAnnotation.init_from_db and in all the other cases lock was removed two years ago and no problems emerged (?), I believe that the lock is not required and is_prefetched can be removed.

I do not agree with that. Maybe such problems were. Imagine the situation when 2 parallel PATCH requests come to update task annotations. If we just remove `select_for_update` then the result will be unexpected (At first glance, the same problem exists now for jobs). Probably we should not lock the database job or task row when updating annotations but use lock per annotations update action.

cvat/apps/dataset_manager/task.py
```diff
@@ -103,25 +110,17 @@ def add_prefetch_info(cls, queryset):
             'segment__task__project__owner',
             'segment__task__project__assignee',
```
Marishka17, line 111: @zhiltsov-max, @Eldies,
Is this information really used in cases except for export?
```
'segment__task__owner',
'segment__task__assignee',
'segment__task__project__owner',
'segment__task__project__assignee'
```

Marishka17 (commented):
Okay, it seems there are still some places that could be optimized (e.g., fetching the task data queryset once for the entire task and then passing it to each JobAnnotation), but it can be addressed in a future PRs.

If other guys @zhiltsov-max, @azhavoro, @SpecLad  agree that we could delete `select_for_update` here and fix potentiall issues with parallel imports in a separate PR - LGTM.
@Eldies, please, don't forget to check if any changes are required in private repositories.

cvat/apps/dataset_manager/task.py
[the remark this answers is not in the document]
Eldies: I investigated a bit more, and there only two cases when this lock happens on some changes in db: when a task is added to project and when a dataset is imported as a project. In all the other cases, this lock happens when no changes are applied - e.g. task export. 
But the lock is required on export, according to https://github.com/cvat-ai/cvat/blob/develop/cvat/apps/dataset_manager/task.py#L1087
therefore I am returning the option, but with the other, more descriptive name

Eldies:
I returned `select_for_update` because apparently it is needed on export. But I changed the option name to more descriptive and removed unnecessary differences in the conditional logic for locking

cvat/apps/dataset_manager/task.py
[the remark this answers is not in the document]
zhiltsov-max: I guess we actually need to figure out when we don't need to lock. There must be at least the following clients of this class:

- job, task, project import, export
- job, task annotation changes (patch, put), honeypot reroll in a job (can remove annotations on the changed honeypot frames)
- quality computation (~export), backup (~export), autoannotation (~import)

It looks to me that in most, if not all, of the cases we do need to lock data. 
- For exporting, the lock is needed to retrieve the relevant data from the DB. If there are multiple DB tables or if there is a batched load from a single table, we do need a lock. We also need image meta
- For importing, the lock is needed because the current logic is to overwrite the existing data. This logic was asked to be changed many times, however, this how it is supposed to be now. Image meta is needed as well, as we need to know image sizes for some of the formats and we need to match image names from the uploaded annotations
- For annotation changes in general, there is a writing step, for sure. But there is also a reading step, which is needed to compute the diff, and record it for analytics

We don't need to hold the lock for the whole operation though. The lifetime of a lock is bound to the transaction, and they are used to control it in the code. For exporting, we typically need it only to prefetch everything in the beginning. For importing, we probably need to lock 2 times - to read and to (over-)write.

cvat/apps/dataset_manager/task.py
```diff
@@ -1028,26 +1032,29 @@ def patch_job_data(pk, data, action):
...
+
 def export_job(job_id, dst_file, format_name, server_url=None, save_images=False):
     # For big tasks dump function may run for a long time and
     # we dont need to acquire lock after the task has been initialized from DB.
     # But there is the bug with corrupted dump file in case 2 or
     # more dump request received at the same time:
     # https://github.com/cvat-ai/cvat/issues/217
     with transaction.atomic():
-        job = JobAnnotation(job_id)
+        job = JobAnnotation(job_id, prefetch_images=True)
```
zhiltsov-max, line 1050:
```suggestion
        job = JobAnnotation(job_id, prefetch_images=True, lock_job_in_db=True)
```

zhiltsov-max (changes requested):
- What I think should be done in the PR is that the cache utilization should be improved. If it's possible to reuse the cache for multiple jobs, it should be reused. Please check if some manual "joins" can be useful in relevant cases (like [here](https://github.com/cvat-ai/cvat/blob/707087564b9d62408679251cc50f1fb3996a34eb/cvat/apps/quality_control/quality_reports.py#L2351-L2357)).
- Please review the locking logic, according to https://github.com/cvat-ai/cvat/pull/8676#discussion_r1853684979.
- Thank you for the profiling results here https://github.com/cvat-ai/cvat/pull/8676#issuecomment-2485623466. The number of requests seems quite big to me, do you have some breakdown on what is being queried? It's approximately 100 requests per job or ~3 per image.

cvat/apps/dataset_manager/task.py
zhiltsov-max:
```
'segment__task__project__owner',
'segment__task__project__assignee'
```
I also couldn't find where these 2 could be used.

cvat/apps/dataset_manager/task.py
```diff
@@ -93,6 +94,12 @@ def add_prefetch_info(cls, queryset):
         ])
         label_qs = JobData.add_prefetch_info(label_qs)
 
+        task_data_queryset = models.Data.objects.all()
```
zhiltsov-max, line 97: I'm pretty sure you need only 1 Data object, as there is at most 1 related to a Task.

Eldies:
> do you have some breakdown on what is being queried? It's approximately 100 requests per job or ~3 per image.

```
In: Counter([
    m.groups()
    for query in connection.queries
    for m in [re.search('FROM "([^"]+)"', query['sql'])]
    if m and len(m.groups()) >= 1
])
Out: 
Counter({('engine_label',): 2002,
         ('engine_job',): 1526,
         ('engine_data',): 1001,
         ('engine_image',): 1001,
         ('engine_skeleton',): 1001,
         ('engine_attributespec',): 1001,
         ('auth_user',): 1001,
         ('organizations_organization',): 1000,
         ('webhooks_webhook',): 216,
         ('engine_labeledimage',): 200,
         ('engine_labeledshape',): 200,
         ('engine_labeledtrack',): 200,
         ('engine_task',): 109,
         ('engine_trackedshapeattributeval',): 100,
         ('engine_trackedshape',): 100,
         ('engine_labeledtrackattributeval',): 100,
         ('engine_labeledshapeattributeval',): 100,
         ('engine_labeledimageattributeval',): 100,
         ('engine_validationlayout',): 1,
         ('engine_video',): 1,
         ('engine_segment',): 1})
```

zhiltsov-max:
@Eldies, please try to enable the silk profiler, you'll get a picture like this:

image

Note that you might need to use `@silk_profile` or call the import or export function directly from some API endpoint to collect the metrics, instead of calling it in an rq job.

Eldies:
For every case when JobAnnotation is called from TaskAnnotation, passed a db_task so that all JobAnnotations could use the same, shared db_task.

Added prefetching to TaskAnnotation. But now TaskAnnotation.add_prefetch_info kinda duplicates JobAnnotation.add_prefetch_info. Something like this could remove duplication:
```
-        return queryset.select_related(
-            'segment',
-            'segment__task',
-        ).prefetch_related(
-            'segment__task__project',
-            'segment__task__owner',
-             ....
-       )
+        return queryset.select_related(
+            'segment',
+        ).prefetch_related(
+            Prefetch('segment__task', queryset=TaskAnnotation.add_prefetch_info(models.Task.objects))
+        )
```
but it will add one request, so I am not sure is it a good idea or not

cvat/apps/dataset_manager/task.py
```diff
@@ -93,35 +94,47 @@ def add_prefetch_info(cls, queryset):
...
+        if prefetch_images:
+            task_data_queryset = task_data_queryset.select_related('video').prefetch_related(
+                Prefetch('images', queryset=models.Image.objects.order_by('frame'))
+            )
+
         return queryset.select_related(
             'segment',
             'segment__task',
```
zhiltsov-max, line 105: FYI, I don't think this impacts memory use heavily at the moment. It seems that using `select_related` results in different `Segment` and `Task` objects in Python, even if they are actually the same DB line. `prefetch_related`, however, results in the same objects with the same ids.

cvat/apps/dataset_manager/task.py
```diff
@@ -93,35 +94,47 @@ def add_prefetch_info(cls, queryset):
...
+        queryset: QuerySet | None = None,
+        prefetch_images: bool = False,
+        db_task: models.Task | None = None
+    ):
         if queryset is None:
-            queryset = self.add_prefetch_info(models.Job.objects)
+            if db_task is None:
+                queryset = self.add_prefetch_info(models.Job.objects, prefetch_images=prefetch_images)
+            else:
+                queryset = models.Job.objects.select_related("segment", "segment__task")
```
zhiltsov-max, line 130: Maybe it's not necessary to request `segment__task` at all in this case? You're going to overwrite it anyway.

cvat/apps/dataset_manager/task.py
```diff
@@ -786,11 +799,34 @@ def import_annotations(self, src_file, importer, **options):
 
         self.create(job_data.data.slice(self.start_frame, self.stop_frame).serialize())
 
+
 class TaskAnnotation:
```
zhiltsov-max, line 803: Please check `ProjectAnnotationAndData` as well in `project.py`.

cvat/apps/dataset_manager/task.py
```diff
@@ -1006,19 +1042,21 @@ def get_job_data(pk):
 
     return annotation.data
 
+
 @silk_profile(name="POST job data")
 @transaction.atomic
-def put_job_data(pk, data):
-    annotation = JobAnnotation(pk)
+def put_job_data(pk, data, db_task: models.Task | None = None):
```
zhiltsov-max, line 1048:
```suggestion
def put_job_data(pk, data, *, db_task: models.Task | None = None):
```
It will be better to make it kw-only, as it's not a "natural" parameter of the operation.

zhiltsov-max, line 1048:
```suggestion
def put_job_data(pk, data: AnnotationIR | dict, db_task: models.Task | None = None):
```

As the line was changed, we need to add annotation.

zhiltsov-max:
> It is 5396 db requests now,

Looks great! However, I still can see some suspicious numbers in these lines:

```
('engine_task',): 109,
('engine_job',): 1419,
```
From the regex, I can guess tasks and jobs can be mixed into some other requests, but still, the numbers are big. Could you check it please?

And here:
`('webhooks_webhook',): 216,`

Probably, it should be like 100 (job updated) + 1 (task updated) + 1 (project updated), if I'm not missing something.

> But now TaskAnnotation.add_prefetch_info kinda duplicates JobAnnotation.add_prefetch_info.

Actually, maybe we could call some functions from JobAnnotation in TaskAnnotation prefetch? I think we know there that we will need jobs at some point later.

> but it will add one request, so I am not sure is it a good idea or not.

The proposed code snippet doesn't look right. We don't need all the task prefetch, if we're working with just 1 job. But we know that we're going to work with jobs, if we're working with TaskAnnotations.

Eldies:
> And here: ('webhooks_webhook',): 216, Probably, it should be like 100 (job updated)

On every job update there is also a task update - in `JobAnnotation._set_updated_date` both the job and its task are touched.
There are 100+ job updates - therefore, 100+ task updates.
Not exactly 100, because there are a lot of annotations and T`askAnnotation._patch_data` is called several times, and some jobs are updated twice.
So, 200+ webhook reads.

> ('engine_task',): 109,

On every job update, the it's task is touched and when the task is touched, it is read from db - `pre_save_resource_event` reads old instance.

> ('engine_job',): 1419,

On each of 100+ job updates:
1 read in `pre_save_resource_event`
1 read in `__save_job_handler`
1 read in `__save_job__update_quality_metrics`

200 reads because in `TaskData.meta_for_task` jobs were not ordered when prefetced for segments, and each `db_segment.job_set.first()` caused a db query. Fixed it.

All the other reads are in `JobAnnotation` initializations.
Redesigned it - now `TaskAnnotations` loads all the jobs it needs along with all needed data once, and then passes jobs to `JobAnnotation`

Now there are 4401 db queries, 
```
Counter({('engine_job',): 318,
         ('webhooks_webhook',): 216,
         ('engine_labeledimage',): 200,
         ('engine_labeledshape',): 200,
         ('engine_labeledtrack',): 200,
         ('engine_task',): 109,
         ('engine_trackedshapeattributeval',): 100,
         ('engine_trackedshape',): 100,
         ('engine_labeledtrackattributeval',): 100,
         ('engine_labeledshapeattributeval',): 100,
         ('engine_labeledimageattributeval',): 100,
         ('organizations_organization',): 100,
         ('engine_label',): 4,
         ('engine_data',): 2,
         ('engine_skeleton',): 2,
         ('engine_attributespec',): 2,
         ('auth_user',): 2,
         ('engine_image',): 1,
         ('engine_validationlayout',): 1,
         ('engine_video',): 1,
         ('engine_segment',): 1})
```
For some reason when I try to use silk, the ui does not show me my tasks. I'll try to investigate it

cvat/apps/dataset_manager/task.py
Eldies: done

zhiltsov-max:
Ok, it feels like a good point to stop optimizing tasks for now. Do you have updated memory measurements for the [import use case](https://github.com/cvat-ai/cvat/pull/8676#issuecomment-2485623466)?

Eldies:
Memory consumption is the same as earlier, ~0.5 Gb less then on develop branch

zhiltsov-max:
> For some reason when I try to use silk, the ui does not show me my tasks. I'll try to investigate it

It's only configured for working with development setup (VS Code debug tasks), so maybe this is the reason. Make sure you're connecting to the right DB and server. You'll need to start `docker compose -f docker-compose.yml -f docker-compose.dev.yml up -d --build   cvat_opa cvat_db cvat_redis_inmem cvat_redis_ondisk cvat_server cvat_vector`, then launch the server debug task in VS code, launch the UI manually with `(cd cvat-ui && yarn run start)`, then go to `localhost:3000` in the browser.

Marishka17 (approved):

zhiltsov-max (approved):

PR #8708, opened 2024-11-15
If in LDAP mode, users not being assign a default role at login

rragundez:
Currently if using LDAP auth IAM_TYPE, a user might have access to the application to login but no role is assigned to the user. This creates confusion to users as they are not able to perform any actions in CVAT even though they can login.

In comparison, this behaviour is not encountered if using the BASIC auth IAM_TYPE, as the `IAM_DEFAULT_ROLE`, is assigned to the user when logs in for the first time.

To explain a bit further the flow that we encountered, maintainers of an organization use the email service to invite users to their org, then users click on the email link, login to the application and accept the invite to the organization. Even though they might be even `maintainers` in the organization they cannot perform any action as the rego rules for organization need users to at least have the "worker" role in the application. The same goes for sending invitation to the organization if they have been made `mantainers`. Here the links to the rego files:

- https://github.com/cvat-ai/cvat/blob/develop/cvat/apps/organizations/rules/organizations.rego
- https://github.com/cvat-ai/cvat/blob/develop/cvat/apps/organizations/rules/invitations.rego

### How has this been tested?
Yes, I have a local deployment with LDAP enabled. WIth these changes, if I log in then I get assigned the default role and I can perform actions in the application.

## Summary by CodeRabbit

- **New Features**
	- Enhanced LDAP user management by refining group assignment logic, ensuring default roles are only assigned to non-superuser and non-staff users.

- **Bug Fixes**
	- Improved control flow for user group assignments to prevent unnecessary role assignments for elevated privilege users.


azhavoro:
@rragundez Hi, thanks for the contrubution.
I have a question: why not just set up mapping between ldap and cvat groups using  `AUTH_LDAP_USER_GROUPS` ?https://docs.cvat.ai/docs/administration/advanced/ldap/

rragundez:
@azhavoro we do use that functionality for admins. But a group on the Active Directory that encompasses all users in the company does not exist. I also checked all groups in the response from the LDAP using my user. 
On the other hand the search when providing access to the application using the _BASE can be done at the domain level hence providing access to anyone without having to create a dedicated group and manage adding people, etc.

This will basically enable having the admins do all the setup but once that is done the owners and maintainers of organizations can manage user access themselves without us, the admins, having to deal with any extra LDAP group membership or assigning roles via the admin panel. 

Hope that helps with the explanation!

rragundez:
@azhavoro we do use that functionality for admins. But a group on the Active Directory that encompasses all users in the company does not exist. I also checked all groups in the response from the LDAP using my user. 
On the other hand the search when providing access to the application using the _BASE can be done at the domain level hence providing access to anyone without having to create a dedicated group and manage adding people, etc.

This will basically enable having the admins do all the setup but once that is done the owners and maintainers of organizations can manage user access themselves without us, the admins, having to deal with any extra LDAP group membership or assigning roles via the admin panel. 

Hope that helps with the explanation!

rragundez:
Hi @azhavoro, what do you think? does it makes sense? Sorry to push it but we have this bug now and we would prefer to give back to the source code and not having to fork the repo to add this change.

rragundez:
Hi @azhavoro and @Marishka17 a quick ping to get your feedback. Thanks.

Marishka17:
@rragundez, @azhavoro, Generally, I guess it makes sense, but I would prefer to add the `IAM_DEFAULT_ROLE` group only if no groups have been added after processing `DJANGO_AUTH_LDAP_GROUPS`.

rragundez:
Hi @Marishka17  thanks for the feedback. I agree with you and have made the necessary adjustment to the `if` condition.

rragundez:
Hi @Marishka17 @azhavoro maybe you have time to look at the changes before the weekend? they are pretty straight forward. 
If you think it will take longer can you let us know, is just that we are onboarding users as we speak and we need to make a decision if to continue to have this more manual addition to AD groups until this is merged, or fork the code and maintain the change. Thanks again and apologies for being so insistent but we are being currently affected by this.

cvat/apps/iam/signals.py
```diff
@@ -42,6 +42,9 @@ def create_user(sender, user=None, ldap_user=None, **kwargs):
                     if role == settings.IAM_ADMIN_ROLE:
                         user.is_staff = user.is_superuser = True
                     break
+        # add default group if no other group has been assigned and the user is not a superuser
+        if len(user_groups) == 0 and not (user.is_superuser or user.is_staff):
+            user_groups.append(Group.objects.get(name=settings.IAM_DEFAULT_ROLE))
```
Marishka17, line 47:
```suggestion
        if not len(user_groups):
            user_groups.append(Group.objects.get(name=settings.IAM_DEFAULT_ROLE))
```
Why do we need to check here whether a user is a superuser?
If I remember correctly, `settings.IAM_TYPE == 'LDAP'` does not assume a superuser to be created, e.g. via python manage.py createsuperuser. In this case, if a user is created with `is_staff`/`is_superuser`==True, the `user_groups` will already include `IAM_ADMIN_ROLE`.

cvat/apps/iam/signals.py
```diff
@@ -42,6 +42,9 @@ def create_user(sender, user=None, ldap_user=None, **kwargs):
                     if role == settings.IAM_ADMIN_ROLE:
                         user.is_staff = user.is_superuser = True
                     break
+        # add default group if no other group has been assigned and the user is not a superuser
```
Marishka17, line 45:
```suggestion
        # add default group if no other group has been assigned
```

Marishka17 (approved):
LGTM, 
@azhavoro, Could you please merge PR if you have no comments?

azhavoro:
@rragundez LGTM, but could you please add a changelog entry? The instrcutions can be found here https://github.com/cvat-ai/cvat/blob/develop/CHANGELOG.md?plain=1#L9-L14

rragundez:
Done @azhavoro, please check

azhavoro (approved):

rragundez:
Awesome @azhavoro thanks! One last question, how often are the releases onto master? in order to get this change

PR #8721, opened 2024-11-19
Fix an issue of releasing lock for rq export job when the worker subprocess is killed

Marishka17:
### Motivation and context

The main problem fixed by this PR is as follows:
In the previous implementation, "long" locks were used when exporting a resource or deleting an export cache.
If the export process was killed (e.g., by the OOM killer with 9 signal), the acquired lock was not released and remained active until the auto-release timeout expired (e.g., 4 hours). A subsequent user request to export a dataset could not acquire the lock, causing the job to be scheduled for execution after 60 seconds (default value). When the scheduled job ran again, it still could not acquire the lock, and the entire process was repeated. Additionally, if a user initiated the export process after the job was marked as scheduled, they were unable to re-initiate the process and received an error because the RQ job status was not set and handled correctly (it was remaining `STARTED`).

One more found and fixed problem is that 2 users that have rights to export a resource could not make export in parallel (with the same options like format, save_images) and one of them received a `LockNotAvailableError` error.
#### How it was fixed:
- Short locks are used now in worker processes with default lock TTL 30s and lock acquisition timeout 50s. 
- Export process acquires lock 2 times:
   - when checking whether a file exists or not and changing the last modification date if a cache file exists
   - when creating a file after the export process is finished
- scheduled jobs now have `SCHEDULED` status and are removed from scheduler jobs (`rq:scheduler:scheduled_jobs` set)
### Checklist

### License

## Release Notes

- **New Features**
  - Enhanced dataset export functionality with improved threading and error handling.
  - Introduced new constants for dataset caching and locking mechanisms, providing clearer defaults.

- **Bug Fixes**
  - Improved error handling for job management in the export process, ensuring better responses for canceled or failed jobs.

- **Documentation**
  - Updated function signatures and added detailed docstrings for clarity and type safety.

- **Tests**
  - Increased the default `max_retries` for export-related functions to enhance reliability during export processes.


cvat/apps/dataset_manager/views.py
```diff
@@ -30,10 +31,12 @@
...
 
+
 slogger = ServerLogManager(__name__)
 
 _MODULE_NAME = __package__ + '.' + osp.splitext(osp.basename(__file__))[0]
-def log_exception(logger=None, exc_info=True):
+
+def log_exception(logger: logging.Logger | None = None, exc_info: bool = True):
     if logger is None:
         logger = slogger
```
Marishka17, line 41:
```suggestion
        logger = slogger.glob
```

cvat/apps/dataset_manager/util.py
```diff
@@ -111,14 +111,17 @@ def get_export_cache_lock(
     *,
     ttl: int | timedelta,
     block: bool = True,
-    acquire_timeout: Optional[int | timedelta] = None,
+    acquire_timeout: int | timedelta,
 ) -> Generator[Lock, Any, Any]:
+    if acquire_timeout is None:
+        raise ValueError("Endless waiting for the lock should be avoided")
+
```
zhiltsov-max, line 118: `None` is not expected, according to the argument annotations. Maybe the check should be removed.

cvat/apps/dataset_manager/tests/test_rest_api_formats.py
```diff
@@ -1598,17 +1629,11 @@ def _clear(*_, file_path: str, file_ctime: str):
...
         # terminate() may break the locks, don't try to acquire
         # https://docs.python.org/3/library/multiprocessing.html#multiprocessing.Process.terminate
         self.assertTrue(export_checked_the_file.get())
-        self.assertTrue(export_created_the_file.get())
-
-        self.assertFalse(clear_removed_the_file.get())
-
-        new_export_path = export_file_path.get()
-        self.assertGreater(len(new_export_path), 0)
-        self.assertTrue(osp.isfile(new_export_path))
```
zhiltsov-max, line 1611: Maybe checks for the clear process activities should be restored? I'd expect it to check the exported file.

cvat/apps/dataset_manager/tests/test_rest_api_formats.py
```diff
@@ -1397,185 +1437,176 @@ def test_concurrent_export_and_cleanup(self):
...
-                    side_effect(wait_condition, export_created_the_file),
-                    original_remove,
-                    side_effect(set_condition, clear_removed_the_file),
-                )
-
                 mock_rq_get_current_job.return_value = MagicMock(timeout=5)
 
-                exited_by_timeout = False
+                set_condition(clear_process_has_run)
+                file_is_been_used_error_raised = False
```
zhiltsov-max, line 1521:
```suggestion
                file_is_being_used_error_raised = False
```

cvat/apps/dataset_manager/views.py
```diff
@@ -60,6 +64,14 @@ def get_export_cache_ttl(db_instance: str | Project | Task | Job) -> timedelta:
 
     return TTL_CONSTS[db_instance.lower()]
 
+def _patch_scheduled_job_status(job: rq.job.Job):
+    # NOTE: rq scheduler < 0.14 does not set the appropriate
+    # job status (SCHEDULED). This has been fixed in the 0.14 version.
+    # https://github.com/rq/rq-scheduler/blob/f7d5787c5f94b5517e209c612ef648f4bfc44f9e/rq_scheduler/scheduler.py#L148
+    # FUTURE-TODO: delete manual status setting after upgrading to 0.14
+    if job.get_status(refresh=False) != rq.job.JobStatus.SCHEDULED:
+        job.set_status(rq.job.JobStatus.SCHEDULED)
```
zhiltsov-max, line 73: Is it possible for us to update the version now? Are there any blockers?

cvat/apps/dataset_manager/views.py
```diff
@@ -133,41 +154,52 @@ def export(dst_format, project_id=None, task_id=None, job_id=None, server_url=No
...
-                            db_instance, (Project, Task)
-                        ) else db_instance.id,
-                        dst_format, output_path, cache_ttl,
-                        cleaning_job.id
-                    )
-                )
+            if osp_exists(output_path):
+                # Update last modification time to prolong the export lifetime
+                # and postpone the file deleting by the cleaning job
+                os.utime(output_path, None)
```
zhiltsov-max, line 168: Now as we call it from several places, it makes sense to factor out a function for this to encapsulate the logic. https://github.com/cvat-ai/cvat/blob/2bf9a0bae0e39fabdee9e9f4ca82b63cb3be7174/cvat/apps/engine/background.py#L334

cvat/apps/dataset_manager/views.py
```diff
@@ -133,41 +154,52 @@ def export(dst_format, project_id=None, task_id=None, job_id=None, server_url=No
...
+            file_ctime=instance_update_time.timestamp(),
+            logger=logger,
+        )
+        _patch_scheduled_job_status(cleaning_job)
+        logger.info(
+            "The {} '{}' is exported as '{}' at '{}' "
+            "and available for downloading for the next {}. "
+            "Export cache cleaning job is enqueued, id '{}'".format(
+                db_instance.__class__.__name__.lower(),
+                db_instance.name if isinstance(db_instance, (Project, Task)) else db_instance.id,
```
zhiltsov-max, line 196:
```suggestion
                db_instance.id,
```
I think we don't really need task names in logs.

cvat/apps/dataset_manager/tests/test_rest_api_formats.py
```diff
@@ -1397,185 +1437,176 @@ def test_concurrent_export_and_cleanup(self):
...
                     side_effect(set_condition, export_checked_the_file),
-                )
-
-                mock_os_replace.side_effect = chain_side_effects(
-                    original_replace,
-                    os_replace_dst_recorder,
-                    side_effect(set_condition, export_created_the_file),
-                    side_effect(wait_condition, clear_removed_the_file),
+                    side_effect(wait_condition, clear_process_has_run),
+                    side_effect(sleep, 1),
```
zhiltsov-max, line 1486: What is the point of this `sleep`?

cvat/apps/dataset_manager/tests/test_rest_api_formats.py
```diff
@@ -1397,185 +1437,176 @@ def test_concurrent_export_and_cleanup(self):
...
-                mock_os_remove.side_effect = chain_side_effects(
-                    side_effect(wait_condition, export_created_the_file),
-                    original_remove,
-                    side_effect(set_condition, clear_removed_the_file),
-                )
-
                 mock_rq_get_current_job.return_value = MagicMock(timeout=5)
 
-                exited_by_timeout = False
+                set_condition(clear_process_has_run)
```
zhiltsov-max, line 1520: Probably, this line should be moved into a call inside of `clear_export_cache` for correctness.

cvat/apps/dataset_manager/tests/test_rest_api_formats.py
```diff
@@ -1397,185 +1437,176 @@ def test_concurrent_export_and_cleanup(self):
...
+                )
+            )
 
             export_process.start()
 
-            wait_condition(export_checked_the_file) # ensure the expected execution order
+            wait_condition(export_checked_the_file)  # ensure the expected execution order
             clear_process.start()
 
             # A deadlock (interrupted by a timeout error) is the positive outcome in this test,
```
zhiltsov-max, line 1612: Probably, the comment in the line 1618 is not true anymore.

cvat/apps/dataset_manager/tests/test_rest_api_formats.py
```diff
@@ -1895,25 +1904,110 @@ def test_export_can_reuse_older_file_if_still_relevant(self):
...
+                patch("cvat.apps.dataset_manager.views.os.replace") as mock_os_replace,
+                patch(
+                    "cvat.apps.dataset_manager.views.rq.get_current_job"
+                ) as mock_rq_get_current_job,
+                patch("cvat.apps.dataset_manager.views.django_rq.get_scheduler"),
+                patch("cvat.apps.dataset_manager.views.task.export_task") as mock_export_task_func,
+            ):
+                mock_export_task_func.side_effect = chain_side_effects(
+                    original_export_task,
+                    side_effect(sleep, lock_acquire_timeout + 1 if sleep_on_export else 0),
```
zhiltsov-max, line 1943: What is the point of this `sleep`? Consider using conditions instead. Using simple sleeps implies unnecessary performance requirements on the executing system, it leads to unstable tests.

cvat/apps/dataset_manager/tests/test_rest_api_formats.py
```diff
@@ -1895,25 +1904,110 @@ def test_export_can_reuse_older_file_if_still_relevant(self):
...
+                process_closing(
+                    multiprocessing.Process(
+                        target=_export,
+                        kwargs=dict(task_id=task["id"], result_queue=result_queue),
+                    )
+                )
+            )
+
+            export_process_1.start()
+            export_process_2.start()
```
zhiltsov-max, line 1978: It's better to start the second export after the first one entered the critical section.

cvat/apps/dataset_manager/views.py
zhiltsov-max: Ok, got you.

cvat/apps/dataset_manager/util.py
Marishka17: > None is not expected, according to the argument annotations. 

But it was expected in the previous implementation+ I think we must prevent endless waiting for lock acquisition. Updated.

cvat/apps/dataset_manager/views.py
Marishka17: done

cvat/apps/dataset_manager/views.py
Marishka17: Agree, updated

cvat/apps/dataset_manager/tests/test_rest_api_formats.py
```diff
@@ -1517,16 +1516,16 @@ def _clear(*_, file_path: str, file_ctime: str):
             ):
                 mock_rq_get_current_job.return_value = MagicMock(timeout=5)
 
-                set_condition(clear_process_has_run)
-                file_is_been_used_error_raised = False
+                file_is_being_used_error_raised = False
                 try:
+                    set_condition(clear_process_has_run)
```
zhiltsov-max, line 1521: At this point you still can't say the function has been called. It would be correct in the except block below or after it, or in a side effect after the lock acquisition attempt (need to catch the `FileIsBeingUsedError` exception there, actually, so it should be a wrapper).

cvat/apps/dataset_manager/tests/test_rest_api_formats.py
Marishka17: I do not understand why these checks should be restored.
> self.assertFalse(clear_removed_the_file.get())

Custom _clear function checks that `mock_os_remove` is not called

> self.assertTrue(export_created_the_file.get())

export_created_the_file is not set anymore since we do not create a file, only update last modification date

> new_export_path = export_file_path.get()
> self.assertGreater(len(new_export_path), 0)
> self.assertTrue(osp.isfile(new_export_path))

we check that the initial export creates a file and then the file modification date is changed

zhiltsov-max: I don't insist on restoring the same checks that were there before the PR changes, but, as I said, it would be nice to check that all the conditions supposed by the test were actually used. The implementation of the test callbacks and of the functions being tested can change, and we just want to make sure these tests can spot a problem.

Marishka17: :crying_cat_face:

cvat/apps/dataset_manager/tests/test_rest_api_formats.py
Marishka17: >  It would be correct in the except block below or after it

Then there will be deadlock because the export process waits with an acquired lock when `clear_process_has_run` is set. At the same time, the clear process raises `FileIsBeingUsedError` only after it acquires a lock and checks that a file should not be deleted.

> or in a side effect after the lock acquisition attempt

I've tried to wrap the lock.acquire method to add a side effect and set `clear_process_has_run`, but this approach has 2 problems:
1. it's a bad idea to patch a shared lock that is used by another process from a  _clear process
2. in such a case, the second process will not be able to acquire a lock after releasing the first process

zhiltsov-max: > Then there will be deadlock because the export process waits with an acquired lock when clear_process_has_run is set. At the same time, the clear process raises FileIsBeingUsedError only after it acquires a lock and checks that a file should not be deleted.

I mean, yes, the cleanup function is expected to wait for the export, if both export and cleanup functions actually use locks. The deadlock is broken by the lock acquisition timeout exception. The `FileIsBeingUsedError` should not be thrown in this scenario. This is how the test worked before these changes. Now this exception can correctly occur after the export process checked the file and before the new export is finished. But the point of the test was to check that the exported file exists after the export, even if cleanup runs immediately after the export.

I think, maybe it should not even be changed, or it can be changed to check 2 cases: if the export exited after file check and the cleanup function runs exactly after this (now it's possible) and the previous case (as nothing changed there).

cvat/apps/dataset_manager/default_settings.py
```diff
@@ -3,12 +3,34 @@
...
+from django.core.exceptions import ImproperlyConfigured
+
 
 DATASET_CACHE_TTL = int(os.getenv("CVAT_DATASET_CACHE_TTL", 60 * 60 * 24))
 "Base lifetime for cached exported datasets, in seconds"
 
-DATASET_CACHE_LOCK_TIMEOUT = int(os.getenv("CVAT_DATASET_CACHE_LOCK_TIMEOUT", 10))
+default_dataset_export_lock_ttl = 30
+
+DATASET_EXPORT_LOCK_TTL = int(os.getenv("CVAT_DATASET_EXPORT_LOCK_TTL",  default_dataset_export_lock_ttl))
```
Marishka17, line 15: @zhiltsov-max, Since these settings will be used for backups also, do you mind if we rename them?
`EXPORT_CACHE_LOCK_TTL`, `EXPORT_CACHE_LOCK_ACQUISITION_TIMEOUT`, `EXPORT_LOCKED_RETRY_INTERVAL`

zhiltsov-max: Ok, but in this case they should probably be moved into the engine, as backups are implemented there.

Marishka17:
@klakhov, Could you please take a look at the failed cypress test?
https://github.com/cvat-ai/cvat/actions/runs/12391822641/job/34590012655
Looks like need to close the `Export is finished` notification.

cvat/apps/dataset_manager/default_settings.py
```diff
@@ -6,9 +6,3 @@
 
 DATASET_CACHE_TTL = int(os.getenv("CVAT_DATASET_CACHE_TTL", 60 * 60 * 24))
```
zhiltsov-max, line 7: Why this variable is left here? Maybe it's better to implement the same logic for backups as well? Either way, it doesn't seem to be the topic of this PR.

changelog.d/20241218_120228_maria_fix_export_job_lock_releasing.md
```diff
@@ -0,0 +1,9 @@
+### Fixed
+
+- Acquiring a long lock with a worker TTL could result in the lock not being released,
+  blocking further exports with the same parameters (<https://github.com/cvat-ai/cvat/pull/8721>)
+- Scheduled RQ jobs could not be restarted due to incorrect RQ job status
+  updating and handling (<https://github.com/cvat-ai/cvat/pull/8721>)
+- Parallel exports with the same parameters by different users caused
+  a LockNotAvailableError to be raised for all users except the one
+  who initiated the export first (<https://github.com/cvat-ai/cvat/pull/8721>)
```
zhiltsov-max, line 9:
```suggestion
- Exporting datasets could start significantly later than expected, both for 1 and several users in the same project/task/job 
  (<https://github.com/cvat-ai/cvat/pull/8721>)
- Scheduled RQ jobs could not be restarted due to incorrect RQ job status
  updating and handling (<https://github.com/cvat-ai/cvat/pull/8721>)
```

It feels like these messages are too detailed for regular users.

cvat/apps/dataset_manager/default_settings.py
Marishka17: If I remember correctly, backup logic also used this setting

cvat/apps/dataset_manager/default_settings.py
zhiltsov-max, line 1: The file can probably be removed now, as it is empty.

zhiltsov-max (approved):

cvat/apps/dataset_manager/default_settings.py
Marishka17: I'm not sure about it since we have the code in app.py that updates Django settings based on settings from default_settings.py. Of course, I can delete this file and remove this logic, but I think it can still be used in the future if there are some specific settings related to dataset functionality.

zhiltsov-max: We can restore it when it's needed.

PR #8724, opened 2024-11-20
SDK/CLI: improve mask support in the auto-annotation functionality

SpecLad:
### Motivation and context

While it is already possible to output mask shapes from AA functions, which the driver _will_ accept, it's not convenient to do so. Improve the practicalities of it in several ways:

* Add `mask` and `polygon` helpers to the interface module.

* Add a helper function to encode masks into the format CVAT expects.

* Add a built-in torchvision-based instance segmentation function.

* Add an equivalent of the `conv_mask_to_poly` parameter for Nuclio functions.

### How has this been tested?

Manual and unit tests.

### Checklist

### License


cvat-sdk/cvat_sdk/auto_annotation/functions/torchvision_instance_segmentation.py
```diff
@@ -0,0 +1,68 @@
...
+
+        if context.conv_mask_to_poly:
+            # Since we treat mask values of exactly LEVEL as true, we'd like them
+            # to also be considered high by find_contours. And for that, the level
+            # parameter must be slightly less than LEVEL.
+            contours = measure.find_contours(
+                mask[0].detach().numpy(), level=math.nextafter(LEVEL, 0)
+            )
+            if not contours:
+                return []
```
zhiltsov-max, line 48: 1. Aren't the masks binary here? Typically, masks are either binary or indexed, where the index is the instance class id or number
1. A single mask can consist of several polygons, so it makes sense to support this. Currently I can see only the first contour is used
1. It seems that it extracts contours >= level as negative- and the opposite as positive-oriented, so maybe you need to use only the negative ones to avoid producing duplicated annotations for holes, we had such problem in the original datumaro implementation

tests/python/sdk/test_masks.py
```diff
@@ -0,0 +1,66 @@
...
+        # 1100
+
+        assert encode_mask(bitmap, bbox) == [1, 5, 2, 1, 0, 4, 1]
+
+        # Edge case: full image
+        bbox = [0, 0, 6, 2]
+        assert encode_mask(bitmap, bbox) == [2, 3, 2, 2, 3, 0, 0, 5, 1]
+
+    def test_encode_mask_invalid(self):
+        with pytest.raises(ValueError):
```
zhiltsov-max, line 51: Consider using a regex to match the error message text

tests/python/sdk/test_masks.py
```diff
@@ -0,0 +1,66 @@
...
+        # 0111
+        # 1100
+
+        assert encode_mask(bitmap, bbox) == [1, 5, 2, 1, 0, 4, 1]
+
+        # Edge case: full image
+        bbox = [0, 0, 6, 2]
+        assert encode_mask(bitmap, bbox) == [2, 3, 2, 2, 3, 0, 0, 5, 1]
+
+    def test_encode_mask_invalid(self):
```
zhiltsov-max, line 50: Consider splitting the test into 3 separate ones

cvat-sdk/cvat_sdk/masks.py
```diff
@@ -0,0 +1,42 @@
...
+# SPDX-License-Identifier: MIT
+
+import math
+from collections.abc import Sequence
+
+import numpy as np
+from numpy.typing import ArrayLike
+
+
+def encode_mask(bitmap: ArrayLike, /, bbox: Sequence[float]) -> list[float]:
```
zhiltsov-max, line 12: Why the bbox is required?

SpecLad: Mainly because I thought I could get away with it. 🙂 Torchvision provides a bounding box with each mask for free, and so does Ultralytics, so I figured it was common enough that I didn't need to implement automatic bbox determination.

If you think it's warranted, I could do it, though.

tests/python/sdk/test_masks.py
SpecLad: Sure, done.

cvat-sdk/cvat_sdk/auto_annotation/functions/torchvision_instance_segmentation.py
SpecLad: > Aren't the masks binary here?

I'm not sure which masks you're talking about, but `mask` is a float tensor. The torchvision recommends converting it to boolean by comparing it to 0.5, which is what I do in the non-polygon branch.

> A single mask can consist of several polygons, so it makes sense to support this.

Yeah, you're right. I originally tried this and I thought `find_contours` returned duplicate contours sometimes, which is why I only used the first contour. But I can't reproduce this problem anymore, so I must've made some mistake before. 🤷‍♂️ 

I changed the logic back to process all contours.

> maybe you need to use only the negative ones to avoid producing duplicated annotations for holes

Yep, implemented.

cvat-sdk/cvat_sdk/auto_annotation/functions/torchvision_instance_segmentation.py
```diff
@@ -15,6 +17,38 @@
...
+        # parameter must be slightly less than LEVEL.
+        contours = measure.find_contours(mask[0].detach().numpy(), level=math.nextafter(LEVEL, 0))
+
+        for contour in contours:
+            if len(contour) < 3 or _is_positively_oriented(contour):
+                continue
+
+            contour = measure.approximate_polygon(contour, tolerance=2.5)
+
+            yield cvataa.polygon(label.item(), contour[:, ::-1].ravel().tolist())
```
zhiltsov-max, line 46: Probably, these polygons should be grouped, if there are more than 1 of them.

zhiltsov-max: As an option, assign a group, equal to the index, to each annotation, if no group is assigned already.

SpecLad: I probably didn't describe the problem clearly enough: the problem is selecting group IDs that do not exist in the **remote** dataset. The user has the option to add the auto-generated annotations to existing annotations in the task, and it's important that new shapes are not added to existing groups.

As far as I can see, your suggestions only make sure that the auto-generated group IDs don't conflict with each other, which is good, but not sufficient.

I could solve this problem by fetching existing annotations, finding the maximum existing group ID and assigning new group IDs starting from that, but I don't really want to do that. It's very inefficient, and is susceptible to race conditions.

IMO, a proper solution lies beyond the scope of this PR and would involve server-side code. (E.g. the server could support special group values that would mean "assign a new group".)

zhiltsov-max: Ok, I thought the annotations are supposed to be overwritten in the result of the call. If not, then it indeed requires a more complicated solution.

SpecLad: The overwrite is optional (it's controlled by the `clear_existing` flag). This mirrors the "Clean previous annotations" checkbox in the UI.

zhiltsov-max (approved):

PR #8746, opened 2024-11-27
Make SDK models pickleable

SpecLad:
### Motivation and context

Currently, they aren't (or rather, they can be pickled, but unpickling fails). This is due to a small quirk of how the model classes work, and is easily worked around.

In addition, don't create a new Configuration object for each model. These objects are pretty beefy, and they increase the size of each pickle by a full kilobyte (and of course they increase memory usage even when pickle is not involved). AFAICS, these objects are only used when assigning values to file-type fields, and it's easy enough to rewrite the logic so that it still works when the model's `_configuration` field is None.

### How has this been tested?

Unit tests.

### Checklist

### License

## Summary by CodeRabbit

- **New Features**
	- Model instances within the SDK can now be pickled, allowing for easier serialization and transmission.
  
- **Tests**
	- Introduced a new test to verify the pickling and unpickling functionality of the `PatchedLabelRequest` model.

### DIFF
--- a/cvat-sdk/gen/templates/openapi-generator/model_utils.mustache
+++ b/cvat-sdk/gen/templates/openapi-generator/model_utils.mustache
@@ -354,6 +354,13 @@ class OpenApiModel(object):
         new_inst = new_cls._new_from_openapi_data(*args, **kwargs)
         return new_inst
 
+    def __setstate__(self, state):
+        # This is the same as the default implementation. We override it,
+        # because unpickling attempts to access `obj.__setstate__` on an uninitialized
+        # object, and if this method is not defined, it results in a call to `__getattr__`.
+        # This fails, because `__getattr__` relies on `self._data_store`, which doesn't
+        # exist in an uninitialized object.
+        self.__dict__.update(state)
 
 class ModelSimple(OpenApiModel):
     """the parent class of models whose type != object in their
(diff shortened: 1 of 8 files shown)


cvat-sdk/gen/templates/openapi-generator/model_utils.mustache
```diff
@@ -1263,27 +1270,21 @@ def validate_and_convert_types(input_value, required_types_mixed, path_to_item,
     input_class_simple = get_simple_class(input_value)
     valid_type = is_valid_type(input_class_simple, valid_classes)
     if not valid_type:
-        if (configuration
-                or (input_class_simple == dict
-                    and dict not in valid_classes)):
```
zhiltsov-max, line 1268: Why this part of the condition was removed?

cvat-sdk/gen/templates/openapi-generator/model_utils.mustache
```diff
@@ -1263,27 +1270,21 @@ def validate_and_convert_types(input_value, required_types_mixed, path_to_item,
     input_class_simple = get_simple_class(input_value)
     valid_type = is_valid_type(input_class_simple, valid_classes)
     if not valid_type:
-        if (configuration
```
zhiltsov-max, line 1266: I can still see configuration in the function args, why this check was removed?

SpecLad: I want this logic to work the same way as before. Before, `configuration` was always true, so I substituted it with `True`. `True or ...` is equivalent to `True`, so I just removed the condition and the `else` branch.

I reviewed the code of `attempt_convert_item`, and the only place where I found `configuration` to actually be used was inside `deserialize_file`. So I updated the code of `deserialize_file` such that when `configuration` is None, it behaves the same way as if it was `Configuration()`.

zhiltsov-max (approved):

PR #8753, opened 2024-11-28
Dockerfile.ui: improve cacheability

SpecLad:
### Motivation and context

Since Docker doesn't know which programs use which environment variables, any `RUN` steps that follow an `ARG` step will not be loaded from the cache if the value of the `ARG` changes. In this case, the `yarn` command will not be loaded from the cache if any of the `ARG` values change, even though (to my knowledge) it does not use any of those values.

Fix this by moving the `ARG` steps as far down as possible.

### How has this been tested?

I ran `docker build` with various values of `SOURCE_MAPS_TOKEN` and observed which steps were loaded from cache.

### Checklist

### License

## Summary by CodeRabbit

- **Chores**
	- Updated the Dockerfile for improved handling of build arguments, enhancing the build process for the UI.
	- Streamlined the declaration and usage of build arguments to ensure better control during the UI build execution.


Dockerfile.ui
```diff
@@ -29,6 +23,13 @@ COPY cvat-core/ /tmp/cvat-core/
 COPY cvat-canvas3d/ /tmp/cvat-canvas3d/
 COPY cvat-canvas/ /tmp/cvat-canvas/
 COPY cvat-ui/ /tmp/cvat-ui/
+
+ARG WA_PAGE_VIEW_HIT
```
bsekachev, line 27: I think any mentioning of WA_PAGE_VIEW_HIT may be removed as we dropped google web analytics long time ago 

@azhavoro I am correct?

SpecLad: FWIW, this variable is still used in the code.

bsekachev: It is used in .ts code that also needs to be removed

bsekachev: This code will work if `WA_PAGE_VIEW_HIT`, but it is not defined anywhere in saas repo as I can see

bsekachev: Okay, I will send one more PR to solve it

bsekachev (approved):

PR #8769, opened 2024-12-04
Fix infinite lock for chunk preparing

azhavoro:
### Checklist

### License


cvat/apps/engine/default_settings.py
```diff
@@ -24,3 +24,8 @@
 """
 Sets the frequency of checking the readiness of the chunk
 """
+
+CVAT_CHUNK_LOCK_TIMEOUT = 50
```
Marishka17, line 28: Lock timeout is lock ttl, but here it is a timeout for lock acquisition.
I also think that you need to change the default value of `blocking_timeout` inside `get_rq_lock_for_job` because IMHO it is a bad practice to allow endless waiting for a lock at least by default.

cvat/apps/engine/utils.py
```diff
@@ -210,7 +210,7 @@ def get_rq_lock_by_user(queue: DjangoRQ, user_id: int, *, timeout: Optional[int]
         )
     return nullcontext()
 
-def get_rq_lock_for_job(queue: DjangoRQ, rq_id: str, *, timeout: Optional[int] = 60, blocking_timeout: Optional[int] = None) -> Lock:
+def get_rq_lock_for_job(queue: DjangoRQ, rq_id: str, *, timeout: Optional[int] = 60, blocking_timeout: Optional[int] = 60) -> Lock:
```
Marishka17, line 213: Shouldn't `blocking_timeout` be less than the request timeout? (to minimize 504 responses)

cvat/apps/engine/default_settings.py
```diff
@@ -24,3 +24,8 @@
 """
 Sets the frequency of checking the readiness of the chunk
 """
+
+CVAT_CHUNK_LOCK_ACQUISITION_TIMEOUT = 50
```
Marishka17, line 28: Could you please clarify the benefit of introducing this constant? It looks like this timeout could be different depending on where the code is executed (but now the same value is used). For instance, in the case of request handling, this value should be less than request timeout (otherwise, we get 504), but in the case of running inside a worker process, I guess this could be more than lock TTL.
So, if we are not planning now to have the different values depending on where a code is running, I guess it's better just to use default values from `get_rq_lock_for_job`.

cvat/apps/engine/utils.py
```diff
@@ -210,7 +210,7 @@ def get_rq_lock_by_user(queue: DjangoRQ, user_id: int, *, timeout: Optional[int]
         )
     return nullcontext()
 
-def get_rq_lock_for_job(queue: DjangoRQ, rq_id: str, *, timeout: Optional[int] = 60, blocking_timeout: Optional[int] = None) -> Lock:
+def get_rq_lock_for_job(queue: DjangoRQ, rq_id: str, *, timeout: Optional[int] = 60, blocking_timeout: Optional[int] = 50) -> Lock:
```
Marishka17, line 213:
```suggestion
def get_rq_lock_for_job(queue: DjangoRQ, rq_id: str, *, timeout: int = 60, blocking_timeout: int = 50) -> Lock:
```
\+ assert timeout and blocking_timeout

PR #8778, opened 2024-12-05
Memory optimization for image chunk preparation

azhavoro:
### Checklist

### License

## Release Notes

- **New Features**
  - Enhanced image loading functionality with a new `load_image` feature, improving handling of various image formats and EXIF rotation.
  - Improved segment preview generation, specifically for 3D segments.

- **Bug Fixes**
  - Resolved issues with TIFF image handling, ensuring correct image rotation and processing.

- **Improvements**
  - Refined task creation process with better validation checks and progress tracking for long-running tasks.
  - Updated utility functions for better type safety and error handling.


cvat/apps/engine/media_extractors.py
```diff
@@ -101,6 +101,21 @@ def image_size_within_orientation(img: Image.Image):
...
+
+def load_image(image: tuple[str, str, str])-> tuple[Image.Image, str, str]:
+    with Image.open(image[0]) as pil_img:
+        pil_img.load()
+
+        # current version of Pillow applies exif rotation immediately when TIFF image opened
+        # and it removes rotation tag after that
+        # so, has_exif_rotation(image) will return False for TIFF images even if they were actually rotated
+        # and original files will be added to the archive (without applied rotation)
+        # that is why we need the second part of the condition
```
SpecLad, line 113: I still don't understand. If loading the TIFF image removes the Exif orientation tag, then wouldn't `ImageOps.exif_transpose` do nothing anyway? It uses the same tag to determine whether it needs to do anything:

https://github.com/python-pillow/Pillow/blob/10.4.0/src/PIL/ImageOps.py#L687

For that matter, it doesn't seem like we need the first part of the condition either, since `exif_transpose` effectively contains the same condition.

cvat/apps/engine/media_extractors.py
```diff
@@ -868,16 +882,17 @@ def _compress_image(source_image: av.VideoFrame | io.IOBase | Image.Image, quali
             image = Image.fromarray(image, mode="L") # 'L' := Unsigned Integer 8, Grayscale
             image = ImageOps.equalize(image)         # The Images need equalization. High resolution with 16-bit but only small range that actually contains information
 
-        converted_image = image.convert('RGB')
+        if image.mode != 'RGB':
```
SpecLad, line 885: Perhaps we could save even more memory by not converting images of mode `L`? If an image is grayscale, we could just encode it as-is.

cvat/apps/engine/media_extractors.py
```diff
@@ -905,31 +920,20 @@ def save_as_chunk(self, images: Iterator[tuple[Image.Image|io.IOBase|str, str, s
...
-                    # current version of Pillow applies exif rotation immediately when TIFF image opened
-                    # and it removes rotation tag after that
-                    # so, has_exif_rotation(image) will return False for TIFF images even if they were actually rotated
-                    # and original files will be added to the archive (without applied rotation)
-                    # that is why we need the second part of the condition
-                    if isinstance(image, Image.Image) and (
-                        has_exif_rotation(image) or image.format == 'TIFF'
-                    ):
+                    if isinstance(image, Image.Image):
                         output = io.BytesIO()
```
SpecLad, line 924: An assert here would be similarly useful.

SpecLad (changes requested):

cvat/apps/engine/media_extractors.py
```diff
@@ -905,31 +920,20 @@ def save_as_chunk(self, images: Iterator[tuple[Image.Image|io.IOBase|str, str, s
...
                 if self._dimension == DimensionType.DIM_2D:
-                    # current version of Pillow applies exif rotation immediately when TIFF image opened
-                    # and it removes rotation tag after that
-                    # so, has_exif_rotation(image) will return False for TIFF images even if they were actually rotated
-                    # and original files will be added to the archive (without applied rotation)
-                    # that is why we need the second part of the condition
-                    if isinstance(image, Image.Image) and (
-                        has_exif_rotation(image) or image.format == 'TIFF'
-                    ):
+                    if isinstance(image, Image.Image):
```
SpecLad, line 923: This seems to change the chunk writing logic quite a bit. Previously, this would only re-encode images if they had to be re-oriented; now it always does it. In effect, the `else` branch of this `if` will never be reached now.

Is this really what we want? Seems like this will waste CPU time on unnecessary re-encoding, and cause quality degradation if the target format is lossy.

azhavoro: Oh yeah, I wanted to optimize by rotating only once when loading the image and indeed the logic of the function changed, I reverted this change.

cvat/apps/engine/media_extractors.py
azhavoro: done

cvat/apps/engine/media_extractors.py
```diff
@@ -868,16 +876,17 @@ def _compress_image(source_image: av.VideoFrame | io.IOBase | Image.Image, quali
...
-            converted_image.save(buf, format='JPEG', quality=quality, optimize=True)
+            image.save(buf, format='JPEG', quality=quality, optimize=True)
             buf.seek(0)
-            width, height = converted_image.size
-            return width, height, buf
+            return image.width, image.height, buf
         finally:
-            converted_image.close()
+            if image is not source_image:
+                image.close()
```
SpecLad, line 889: Is this `close` even needed at all anymore? As far as I can see, it's impossible for `image` to have an attached open file at this point (given the changes to `load_image`).

SpecLad (approved):

PR #8804, opened 2024-12-09
Introduce cron jobs to clean up the export cache (`data/cache/export/`) and temporary files/directories (`data/tmp/`)

Marishka17:
### Motivation and context

~~Depends on https://github.com/cvat-ai/cvat/pull/8721~~
**PR introduces the following changes:**
- [_Export datasets|backups_] Cache cleaning moved into a separate cron job. A separate RQ job is no longer enqueued after each export process
- [_Export datasets|backups_] Temporary directory is created inside the common CVAT tmp directory (`data/tmp/`) instead of `project|task|job/id/tmp/export_cache/`
- [_Export datasets_] One tmp directory (instead of 2) is created during the export process
- Added the `exportcachecleanup` management command to remove outdated `project|task|job/id/tmp/export_cache/` directories
- Added a cron job to clean up the `data/tmp/` directory. The new setting `TMP_FILE_OR_DIR_RETENTION_DAYS` is used to determine whether a file or directory should be removed
- Locks are now used when working with backups as they are used for export

**Notes:**
- `data/project|task|job/id/tmp/` is still used during uploading annotations/datasets, but this should be fixed in a separate PR.

**Breaking changes:**
- Enqueuing RQ jobs to backup a project/task: A filename argument was removed from the background function signature. Previously enqueued jobs will fail.
- Cleaning cache after exporting backups or datasets: Scheduled RQ jobs will fail due to `cvat.apps.dataset_manager.views.clear_export_cache` was moved and `cvat.apps.engine.backup._clear_export_cache` was deleted.
### Checklist

### License

## Summary by CodeRabbit

- **New Features**
	- Introduced periodic tasks for automatic cleanup of export caches for projects, tasks, and jobs.
	- Added new functionality for managing and cleaning up export cache files.

- **Bug Fixes**
	- Enhanced error handling for expired files in export processes.

- **Refactor**
	- Restructured export and cleanup functionalities for improved clarity and maintainability.
	- Updated test methods for better focus on export and cleanup operations.

- **Documentation**
	- Added deprecation warnings for outdated environment variables related to caching settings.

- **Chores**
	- Expanded formatting script to include new files for consistent code style.


cvat/settings/base.py
```diff
@@ -353,6 +353,20 @@ class CVAT_QUEUES(Enum):
         'func': 'cvat.apps.iam.utils.clean_up_sessions',
         'cron_string': '0 0 * * *',
     },
+    *(
+        {
+            'queue': CVAT_QUEUES.CLEANING.value,
+            'id': f'cron_{model.lower()}_export_cache_cleanup',
+            'func': 'cvat.apps.engine.cron.cron_export_cache_cleanup',
+            # Run once a day at midnight
```
SpecLad, line 361: This is only applicable to projects. You've configured the other jobs to run at 6:00 and noon.

That said, what's the purpose of staggering the jobs like that? Why not just have one job that runs at midnight and cleans up everything?

Marishka17: > This is only applicable to projects. You've configured the other jobs to run at 6:00 and noon.

That's because previously 3 jobs ran at midnight and I forgot to delete the comment.

> That said, what's the purpose of staggering the jobs like that? Why not just have one job that runs at midnight 
and cleans up everything?

I thought it would be better to split the "long" potential job into several to avoid clogging the queue since there are other jobs that execute in this queue.

cvat/apps/engine/cron.py
```diff
@@ -0,0 +1,104 @@
...
+def clear_export_cache(file_path: str, logger: logging.Logger) -> None:
+    try:
+        with get_export_cache_lock(
+            file_path,
+            block=True,
+            acquire_timeout=EXPORT_CACHE_LOCK_ACQUISITION_TIMEOUT,
+            ttl=EXPORT_CACHE_LOCK_TTL,
+        ):
+            if not osp.exists(file_path):
+                raise FileNotFoundError(f"Export cache file {file_path} doesn't exist")
```
SpecLad, line 44: This seems like a redundant check. If the file does not exist, then `osp.getmtime` will fail with the same exception.

cvat/apps/engine/cron.py
```diff
@@ -0,0 +1,104 @@
...
+
+            parsed_filename = ExportCacheManager.parse_file_path(file_path)
+            cache_ttl = get_export_cache_ttl(parsed_filename.instance_type)
+
+            if timezone.now().timestamp() <= osp.getmtime(file_path) + cache_ttl.total_seconds():
+                logger.info("Cache file '{}' is recently accessed".format(file_path))
+                raise FileIsBeingUsedError
```
SpecLad, line 51: What's the purpose of this exception? Seems like you could just return.

cvat/apps/engine/cron.py
```diff
@@ -0,0 +1,104 @@
...
+
+            os.remove(file_path)
+            logger.debug(f"Export cache file {file_path!r} successfully removed")
+    except LockNotAvailableError:
+        logger.info(f"Failed to acquire export cache lock for the file: {file_path}.")
+        raise
```
SpecLad, line 57: Does it make sense to raise here? You've already logged the message, so what is propagating the exception going to achieve?

cvat/apps/engine/cron.py
```diff
@@ -0,0 +1,104 @@
...
+
+        for child in export_cache_dir_path.iterdir():
+            # export cache dir may contain temporary directories
+            if not child.is_file():
+                logger.debug(
+                    f"The {child.relative_to(instance_dir_path)} is not a file, skipping..."
+                )
+                continue
+
+            with suppress(Exception):
```
SpecLad, line 97: IMO, it would be clearer to both catch and log here, instead of catching here and logging in `clear_export_cache`.

cvat/apps/engine/cron.py
```diff
@@ -0,0 +1,104 @@
...
+                clear_export_cache(child, logger)
+
+    finished_at = timezone.now()
+    logger.info(
+        f"Clearing the {model_name.lower()}'s export cache has been successfully "
```
SpecLad, line 102:
```suggestion
        f"Clearing the {model_name.lower()} export cache has been successfully "
```

The possessive doesn't make sense here, because you're not talking about an individual task's cache (or project's, etc.).

cvat/apps/engine/cron.py
```diff
@@ -0,0 +1,104 @@
...
+        f"completed after {int((finished_at - started_at).total_seconds())} seconds..."
```
SpecLad, line 103:
```suggestion
        f"completed after {int((finished_at - started_at).total_seconds())} seconds."
```

An ellipsis makes sense when you're starting to do something that will take a while. Here everything is already done.

cvat/apps/engine/management/commands/syncperiodicjobs.py
```diff
@@ -71,6 +71,8 @@ def handle(self, *args, **options):
                     cron_string=job_definition['cron_string'],
                     func=job_definition['func'],
                     id=job_id,
+                    args=job_definition.get('args'),
+                    kwargs=job_definition.get('kwargs'),
```
SpecLad, line 75: You need to update the comparison logic to make sure that the job is recreated if its args are changed.

cvat/settings/base.py
SpecLad: Do you think it would take a long time to run all three? I'd assume it would be pretty quick, since it's just deleting some files.

cvat/apps/engine/cron.py
Marishka17: This exception was introduced earlier and was used at least in tests, but I agree that we can remove it.

cvat/apps/engine/cron.py
Marishka17: Fixed

cvat/settings/base.py
Marishka17: > Do you think it would take a long time to run all three? 

I'll check with the real amount of data

cvat/apps/engine/cron.py
```diff
@@ -0,0 +1,87 @@
...
+    get_export_cache_ttl,
+    log_exception,
+)
+from cvat.apps.engine.log import ServerLogManager
+from cvat.apps.engine.models import Job, Project, Task
+
+logger = ServerLogManager(__name__).glob
+
+
+def clear_export_cache(file_path: str, logger: logging.Logger) -> None:
```
SpecLad, line 28:
```suggestion
def clear_export_cache(file_path: str) -> None:
```

cvat/settings/base.py
Marishka17: Okay, I’ve checked it on a queryset of 15k items, and it takes about 10 seconds. I guess that the actual numbers will be around 100–300k for tasks, less than 100k for projects, and well over 300k for jobs. As a result, this cron job could take significant time when a large number of tasks with many associated jobs are created within the last 30 days. This problem remains the same if there are 3 separate cron jobs. Based on this, I decided to merge separate jobs into one.

Marishka17: We could probably select tasks/projects/jobs that have been exported within the last 2 weeks instead of 1 month

cvat/apps/engine/cron.py
```diff
@@ -0,0 +1,104 @@
...
+        log_exception(logger)
+        raise
+
+
+def cron_export_cache_cleanup(path_to_model: str) -> None:
+    assert isinstance(path_to_model, str)
+
+    started_at = timezone.now()
+    module_name, model_name = path_to_model.rsplit(".", 1)
+    module = importlib.import_module(module_name)
```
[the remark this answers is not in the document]
Marishka17: removed

cvat/settings/base.py
zhiltsov-max: How about limiting 1 function run just by a fixed number of oldest entries? The next run will continue from where the previous one finished,

cvat/apps/engine/migrations/0087_job_last_export_date_project_last_export_date_and_more.py
```diff
@@ -0,0 +1,28 @@
...
+            model_name="project",
+            name="last_export_date",
+            field=models.DateTimeField(null=True),
+        ),
+        migrations.AddField(
+            model_name="task",
+            name="last_export_date",
+            field=models.DateTimeField(null=True),
+        ),
+    ]
```
zhiltsov-max, line 28: Should there be something like a one-off cleanup job to clean old exports? As a workaround, set last export date to creation date, otherwise objects will be skipped by the current cleanup condition.

Marishka17: You are right, but I thought we'd run such a one-off cleanup job/script manually after deploying this patch. I don't think that we need to add a data migration here, since such a migration can take a lot of time. Instead, we could specify in the documentation how to clean up the outdated cache.

> As a workaround, set the last export date to the creation date

This patch tries to uniform the naming of files saved in the export cache (specifically backups) without backward compatibility. Since a file could have been stored in the cache for less than 1 day, I don't think that we need to support files with old name patterns. So, this workaround is not fully robust.

cvat/settings/base.py
SpecLad: Here's a crazy idea. Instead of mucking around with last export dates, can we change the file layout so that all exports are saved in a dedicated directory? That way, the cleanup job would just have to iterate over that directory. This way, it'll be impossible to miss any stale exports.

SpecLad: > I guess that the actual numbers will be around 100–300k for tasks, less than 100k for projects, and well over 300k for jobs.

FWIW, I'm pretty sure we have analytics events for exports, so you could see what the actual numbers are (although there might not be events for backups).

> As a result, this cron job could take significant time when a large number of tasks with many associated jobs are created within the last 30 days. 

I might be misunderstanding something, but why does it matter how many tasks/jobs are _created_? I thought this was about exports.

site/content/en/docs/administration/advanced/upgrade_guide.md
```diff
@@ -56,6 +56,52 @@ To upgrade CVAT, follow these steps:
...
+
+if __name__ == "__main__":
+    cleanup_outdated_cache()
+
+```
+
+### How to run the script
+
+1. Save the script as `cleanup_script.py` in a directory where `manage.py` is located
+1. Run Django shell command: `python manage.py shell`
```
Marishka17, line 98: To run such one-time scripts, we probably need to add in the future a custom management command like `runscript` from the `django-extensions` package.

cvat/apps/dataset_manager/project.py
```diff
@@ -139,7 +138,7 @@ def export(self, dst_file: str, exporter: Callable, host: str='', **options):
         )
 
         temp_dir_base = self.db_project.get_tmp_dirname()
-        os.makedirs(temp_dir_base, exist_ok=True)
```
zhiltsov-max, line 142: Why these `makedirs` were removed? They have to create parent dirs. `TemporaryDirectory` fails, if these dirs do not exist.

Marishka17: `makedirs` calls were moved into `model.get_tmp_dirname` method

cvat/apps/dataset_manager/project.py
```diff
@@ -156,7 +155,7 @@ def import_dataset(self, dataset_file, importer, **options):
         project_data.soft_attribute_import = True
 
         temp_dir_base = self.db_project.get_tmp_dirname()
-        os.makedirs(temp_dir_base, exist_ok=True)
+
```
zhiltsov-max, line 158: Why this `makedirs` call was removed? It has to create parent dirs. `TemporaryDirectory` fails, if these dirs do not exist. But what I think is that we don't even need the parent dir here, it's actually better to create a regular temporary directory in `/tmp`.

cvat/apps/engine/cron.py
```diff
@@ -0,0 +1,133 @@
...
+
+        for child in os.scandir(export_cache_dir_path):
+            # stop clean up process correctly before rq job timeout is ended
+            if self._stop_event.is_set():
+                return
+
+            # export cache directory may contain temporary directories
+            if not child.is_file():
+                logger.debug(f"The {child.name} is not a file, skipping...")
+                continue
```
zhiltsov-max, line 92: These directories should also be removed, if they are outdated. Consider using `shutil.rmtree` for this.

cvat/apps/engine/cron.py
```diff
@@ -0,0 +1,133 @@
...
+    ):
+        parsed_filename = ExportCacheManager.parse_file_path(file_path)
+        cache_ttl = get_export_cache_ttl(parsed_filename.instance_type)
+
+        if timezone.now().timestamp() <= osp.getmtime(file_path) + cache_ttl.total_seconds():
+            logger.debug(f"Export cache file {file_path!r} was recently accessed")
+            return
+
+        os.remove(file_path)
+        logger.debug(f"Export cache file {file_path!r} was successfully removed")
```
zhiltsov-max, line 57: You should either move this into the `dataset_manager` app or move test from that app into `engine`. Currently we have some mix with no clear responsibility separation between components.

cvat/apps/engine/models.py
```diff
@@ -430,6 +432,26 @@ class Meta:
     def touch(self) -> None:
         self.save(update_fields=["updated_date"])
 
+class ABCModelMeta(ABCMeta, ModelBase):
+    pass
+
+class _FileSystemRelatedModel(models.Model, metaclass=ABCModelMeta):
```
zhiltsov-max, line 438: Why this class has an underscore prefix? Models inherit it, and it's functions are supposed to be used by model clients.

cvat/apps/engine/models.py
```diff
@@ -430,6 +432,26 @@ class Meta:
...
+    class Meta:
+        abstract = True
+
+    @abstractmethod
+    def get_dirname(self) -> str:
+        ...
+
+    def get_tmp_dirname(self, create: bool = True) -> str:
```
zhiltsov-max, line 446:
```suggestion
    def get_tmp_dirname(self, *, create: bool = True) -> str:
```

cvat/settings/base.py
```diff
@@ -413,6 +420,9 @@ class CVAT_QUEUES(Enum):
 CACHE_ROOT = os.path.join(DATA_ROOT, 'cache')
 os.makedirs(CACHE_ROOT, exist_ok=True)
 
+EXPORT_CACHE_ROOT = os.path.join(CACHE_ROOT, 'export')
+os.makedirs(EXPORT_CACHE_ROOT, exist_ok=True)
```
zhiltsov-max, line 424: This declaration should be duplicated in testing settings, otherwise the real local cache dir will be used in tests.

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -3089,30 +3091,47 @@ def test_api_v2_tasks_id_export_no_auth(self):
...
+                "cvat.apps.engine.cron.clear_export_cache",
+                side_effect=clear_export_cache,
+            ) as mock_clear_export_cache,
+            mock.patch(
+                "cvat.apps.engine.cron.get_current_job",
+            ) as mock_rq_get_current_job,
+        ):
+            mock_rq_job = mock.MagicMock(timeout=100)
+            mock_rq_get_current_job.return_value = mock_rq_job
+            cron_export_cache_cleanup()
```
zhiltsov-max, line 3113: With 5 seconds as export cache TTL the test will be working unreliable on slow machines. It's better to either patch the expiration check or use some significant number as TTL, e.g. 1 hour.

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -3089,30 +3091,47 @@ def test_api_v2_tasks_id_export_no_auth(self):
...
+            rq_job_ids = queue.finished_job_registry.get_job_ids()
+            self.assertEqual(len(rq_job_ids), 1)
+            job: RQJob | None = queue.fetch_job(rq_job_ids[0])
+            self.assertFalse(job is None)
+            file_path = job.return_value()
+            self.assertTrue(os.path.isfile(file_path))
 
-        from cvat.apps.engine.backup import _clear_export_cache
-        _clear_export_cache(**cleanup_job.kwargs)
+            sleep(TASK_CACHE_TTL.total_seconds() + 1)
```
zhiltsov-max, line 3130: It's better to patch expiration check or TTL one more time instead of using sleep.

cvat/apps/dataset_manager/project.py
```diff
@@ -139,7 +138,7 @@ def export(self, dst_file: str, exporter: Callable, host: str='', **options):
         )
 
         temp_dir_base = self.db_project.get_tmp_dirname()
```
zhiltsov-max, line 140: It makes sense to move these temp dirs to /tmp or the new shared cache as well. Just keep in mind that the temp output archive should be on the same disk as the resulting shared cache file, otherwise `os.rename` can't handle it. `os.rename` can be replaced to `shutil.move`, but this function can't overwrite files.

cvat/apps/engine/cron.py
```diff
@@ -0,0 +1,133 @@
...
+        sleep(sleep_interval)
+
+    if not stop_event.is_set():
+        stop_event.set()
+
+    cleanup_export_cache_thread.join()
+    if isinstance(
+        (exception := cleanup_export_cache_thread.exception), Exception
+    ):
+        raise exception
```
zhiltsov-max, line 126: Consider putting all this exception and event handling logic behind the class API.

cvat/settings/base.py
Marishka17: Thanks, I forgot about this

cvat/apps/engine/models.py
Marishka17: I missed that, thanks

cvat/apps/engine/models.py
Marishka17: > Models inherit it, and it's functions are supposed to be used by model clients.

That's true, but I named this class with an underscore to indicate that it will not be imported into other modules.

cvat/apps/engine/cron.py
Marishka17: done

cvat/apps/engine/cron.py
Marishka17: Moved `cron.py` into the `dataset_manager` app. I don't really think that it is the purpose of this PR to refactor the functionality layout.

cvat/apps/engine/models.py
zhiltsov-max: Do you mean will not be imported if someone uses `from models import *`? It's already discouraged in the project by pylint.

cvat/apps/dataset_manager/project.py
Marishka17: I don't think we need to move this cache into `/tmp`. If I remember correctly, there was a decision to use a custom temporary directory inside the CVAT folder.

cvat/apps/engine/models.py
Marishka17: Okay, I've removed the underscore from the class name.

> Do you mean will not be imported if someone uses from models import *?

Not only this, but it also indicates to developers that it is a private class intended for use within `models`. I thought about this point again, we should not prohibit using this class in other django apps, so it's really a public API.

cvat/apps/engine/cron.py
Marishka17: > Temp directories from failed exports should also be removed

I'm not sure what cases you are talking about. We use `TemporaryDirectory` during export/backup processes, so if any exception occurs, the temp directory is removed with all its content. If such a directory exists in the cache, then this directory will be removed after the export is finished. It is probably only possible if the worker process is killed during the export.

zhiltsov-max: Yes, I was speaking about cases when the export worker is killed for any reason, it happens quite frequently.

site/content/en/docs/administration/advanced/upgrade_guide.md
SpecLad: Why not just turn the script into a management command?

cvat/apps/engine/models.py
```diff
@@ -430,6 +432,26 @@ class Meta:
     def touch(self) -> None:
         self.save(update_fields=["updated_date"])
 
+class ABCModelMeta(ABCMeta, ModelBase):
+    pass
+
+class FileSystemRelatedModel(models.Model, metaclass=ABCModelMeta):
```
SpecLad, line 438: Does this actually need to inherit from `Model`? It's got no fields.

cvat/apps/engine/models.py
```diff
@@ -430,6 +432,26 @@ class Meta:
...
+    class Meta:
+        abstract = True
+
+    @abstractmethod
+    def get_dirname(self) -> str:
+        ...
+
+    def get_tmp_dirname(self, *, create: bool = True) -> str:
```
SpecLad, line 446: Does it ever make sense to use `create=False`? Presumably if you're getting the directory path, you're going to access it.

site/content/en/docs/administration/advanced/upgrade_guide.md
Marishka17: Are you talking about creating a general command to run such scripts or a specific command that handles this logic?

SpecLad: The latter.

cvat/apps/dataset_manager/util.py
```diff
@@ -190,6 +192,80 @@ class ParsedDatasetFilename(_ParsedExportFilename):
...
+    SPLITTER = "-"
+    INSTANCE_PREFIX = "instance"
+    TMP_ROOT = settings.TMP_FILES_ROOT
+
+    @classmethod
+    @contextmanager
+    def get_tmp_dir(
+        cls,
+        *,
+        prefix: str | object = _not_set,
```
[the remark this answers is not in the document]
Marishka17: PR is not ready for review yet

site/content/en/docs/administration/advanced/upgrade_guide.md
SpecLad: > Because we need to run this script only once.

I don't think that's relevant. The point of including this script is so that _others_ may run it, and we might as well make it as easy on them as possible. It'll also be easier to maintain the script if it's an actual Python module and not just a code snippet.

site/content/en/docs/administration/advanced/upgrade_guide.md
```diff
@@ -56,6 +56,52 @@ To upgrade CVAT, follow these steps:
...
+## Upgrade CVAT after v2.25.0
+
+In version 2.25.0, CVAT changed the location where the export cache is stored.
+The following Python script can be used to remove outdated files from the previous location:
+
+```python
+import shutil
+from pathlib import Path
+from django.utils import timezone
+from tqdm import tqdm
```
SpecLad, line 68: `tqdm` is not included in the server dependencies, so I don't think you should use it here.

cvat/apps/engine/models.py
Marishka17: > Does this actually need to inherit from Model? It's got no fields.

There were several new fields, but during fixing comments, those fields were deleted and I forgot to delete the base class

Marishka17:
@zhiltsov-max, @SpecLad,  okay, I've made several changes regarding the working with tmp dirs (including the export process):
1. The `data/tmp/` directory is used now during the export process, and a specific prefix (`export-<task|project|job>-...`) is added to the directories created during the export process. I've initially included the resource type in a directory name to be able to use `TTL_CONSTS`. However, this is not used now due to the third point.  
2. in the previous implementation, 2 tmp dirs were created during export. This has been fixed.
3. added a separate cron job to clear the `data/tmp/` directory. This cron job iterates over all files and directories from `data/tmp/` and deletes all items that were not accessed in the last N days (the default value is set to 2 weeks). This approach can guarantee that we won't have files/directories that should have been deleted but remained.

zhiltsov-max:
> and deletes all items that were not accessed in the last N days (the default value is set to 2 weeks)

Do you know why any value other than 1 day may be needed? Is there anything potentially useful in that directory?

> The `data/tmp/` directory is used now during the export process

The only concern I have about this is that we're still using `os.replace()` to finalize export, and it just an alias of `os.rename()` on linux, which doesn't support moving different disks. Probably, it's not a bit problem so far, but it can fire if someone tries to use a different mount for `data/tmp`. They will need to change `os.replace` to `shutil.move`.

Marishka17:
> Do you know why any value other than 1 day may be needed? Is there anything potentially useful in that directory?

No, I don't. I set this value to make sure I don't delete something we need too early.
As I can see, there are several cases when common tmp directory is used (not taking into account changes from this PR):
- to store some data when exporting events (in that case file should be removed after 1h) https://github.com/cvat-ai/cvat/blob/f8f40e22b08d62d263e159015b54a35c3656d8cf/cvat/apps/events/export.py#L162
- to store files when uploading by TUS (backups) and not(annotations|datasets|backups) https://github.com/cvat-ai/cvat/blob/f8f40e22b08d62d263e159015b54a35c3656d8cf/cvat/apps/engine/views.py#L3394 https://github.com/cvat-ai/cvat/blob/f8f40e22b08d62d263e159015b54a35c3656d8cf/cvat/apps/engine/backup.py#L1160
- to store some TUS-related files https://github.com/cvat-ai/cvat/blob/f8f40e22b08d62d263e159015b54a35c3656d8cf/cvat/apps/engine/mixins.py#L109 (in case of uploading backups)

site/content/en/docs/administration/advanced/upgrade_guide.md
```diff
@@ -56,6 +56,12 @@ To upgrade CVAT, follow these steps:
   docker logs cvat_server -f
   ```
 
+## Upgrade CVAT after v2.25.0
+
+In version 2.25.0, CVAT changed the location where the export cache is stored.
+To clean up the outdated cache, run the following command: `python manage.py exportcachecleanup`.
+
```
zhiltsov-max, line 63: > Enqueuing RQ jobs to backup a project/task: A filename argument was removed from the background function signature. Previously enqueued jobs will fail.

What do you think about adding a one-off migration script to avoid losing requested RQ jobs? It should probably just remove 1 parameter from the existing jobs.

cvat/settings/base.py
```diff
@@ -745,3 +764,6 @@ class CVAT_QUEUES(Enum):
 
 CLOUD_DATA_DOWNLOADING_MAX_THREADS_NUMBER = 4
 CLOUD_DATA_DOWNLOADING_NUMBER_OF_FILES_PER_THREAD = 1000
+
+# Indicates the maximum number of days a file or directory is retained in the temporary directory
+TMP_FILE_OR_DIR_RETENTION_DAYS = 14
```
zhiltsov-max, line 769:
```suggestion
# Indicates the maximum number of days a file or directory is retained in the temporary directory
TMP_FILE_RETENTION_DAYS = 14
```

- Probably, there is no point in distinguishing between files and dirs
- Consider moving this variable to `default_settings.py`
- As we discussed, probably there is no point in such a long default period. consider changing to 1-2 days instead.

cvat/apps/engine/views.py
```diff
@@ -531,17 +531,18 @@ def get_upload_dir(self):
...
             format_name = request.query_params.get("format", "")
             filename = request.query_params.get("filename", "")
             conv_mask_to_poly = to_bool(request.query_params.get('conv_mask_to_poly', True))
             tmp_dir = self._object.get_tmp_dirname()
-            uploaded_file = None
-            if os.path.isfile(os.path.join(tmp_dir, filename)):
-                uploaded_file = os.path.join(tmp_dir, filename)
+            uploaded_file = os.path.join(tmp_dir, filename)
+            if not os.path.isfile(uploaded_file):
+                uploaded_file = None
```
zhiltsov-max, line 544: - It looks like it should be a 400 error instead. 
- There should be some path validation, like one in `is_valid_uploaded_file_name` of `mixins.py`. For backups as well.

cvat/apps/engine/backup.py
```diff
@@ -542,7 +559,7 @@ def _export_task(self, zip_obj, target_dir=None):
         self._write_annotations(zip_obj, target_dir)
         self._write_annotation_guide(zip_obj, target_dir)
 
-    def export_to(self, file, target_dir=None):
+    def export_to(self, file: str | ZipFile, target_dir: str | None = None):
```
zhiltsov-max, line 562: Not an error, but such annotations in inherited functions tend to become outdated quickly. Typically it's better to just use annotations from the original function instead of duplicating them everywhere.

cvat/apps/engine/backup.py
```diff
@@ -1036,37 +1053,65 @@ def _import_project(filename, user, org_id):
     db_project = project_importer.import_project()
     return db_project.id
 
-def create_backup(db_instance, Exporter, output_path, logger, cache_ttl):
+
+def create_backup(
+    # FUTURE-FIXME: there db_instance_id should be passed
+    db_instance: models.Project | models.Task,
```
zhiltsov-max, line 1059: Why not change this parameter if the function args are already changed in this PR?

cvat/apps/dataset_manager/management/commands/exportcachecleanup.py
```diff
@@ -0,0 +1,46 @@
...
+
+class Command(BaseCommand):
+    help = "Cleanup outdated export cache"
+
+    def handle(self, *args, **options):
+        def update_progress():
+            progress = (i + 1) / objects_count
+            done = int(progress_bar_len * progress)
+            progress_bar = "#" * done + "-" * (progress_bar_len - done)
+            self.stdout.write(f"\rProgress: |{progress_bar}| {progress:.0%}", ending="")
```
zhiltsov-max, line 23: This might not work on mac and windows. Why not use tqdm?

Marishka17: Ask Roman about it

cvat/apps/dataset_manager/management/commands/exportcachecleanup.py
```diff
@@ -0,0 +1,46 @@
...
+                f"The {objects_count} folders are going to be checked"
+                if objects_count > 1
+                else "The 1 folder is going to be checked"
+            )
+            self.stdout.write(msg)
+
+            for i, obj in enumerate(queryset.iterator()):
+                update_progress()
+                export_cache_dir = Path(obj.get_dirname()) / "export_cache"
+                if export_cache_dir.exists():
```
zhiltsov-max, line 45:
```suggestion
                suppress(FileNotFoundError)::
```

The directory can be removed already after this check.

cvat/apps/dataset_manager/management/commands/exportcachecleanup.py
zhiltsov-max: @SpecLad, I saw you were saying it's not installed, but why not install it? Do you see any problems with that?

cvat/apps/dataset_manager/cron.py
```diff
@@ -0,0 +1,191 @@
...
+                    self._number_of_removed_objects += 1
+            except CacheFileOrDirPathParseError:
+                logger.warning(f"Cannot parse {child.name}, skipping...")
+                continue
+
+            except Exception:
+                log_exception(logger)
+
+
+def cleanup(thread_class_path: str) -> None:
```
zhiltsov-max, line 159: `python manage.py runperiodicjob cron_export_cache_cleanup`

`TypeError: cleanup() missing 1 required positional argument: 'thread_class_path'`

site/content/en/docs/administration/advanced/upgrade_guide.md
```diff
@@ -56,6 +56,12 @@ To upgrade CVAT, follow these steps:
   docker logs cvat_server -f
   ```
 
+## Upgrade CVAT after v2.25.0
+
+In version 2.25.0, CVAT changed the location where the export cache is stored.
+To clean up the outdated cache, run the following command: `python manage.py exportcachecleanup`.
```
zhiltsov-max, line 62:
```suggestion
To clean up the outdated cache, run the following command: `docker exec -it cvat_server python manage.py exportcachecleanup`.
```

Marishka17: how about a case when CVAT is deployed with kubernetes?

cvat/apps/dataset_manager/cron.py
Marishka17: The `runperiodicjob` command was added this week. This command should support passing args rather than setting the default value for the `cleanup` function.

> AttributeError: 'NoneType' object has no attribute 'timeout'

because this command should be executed only from worker

cvat/apps/dataset_manager/management/commands/exportcachecleanup.py
Marishka17: thanks, I've missed that

site/content/en/docs/administration/advanced/upgrade_guide.md
zhiltsov-max: Both variants are incorrect, need to find a better one.

zhiltsov-max: Maybe just add 3 options for simplicity. Organize them in tabs.

cvat/settings/base.py
Marishka17: > Consider moving this variable to default_settings.py

Are you also voting to move `TmpDirManager` into the `engine` in that case?

Marishka17: > Probably, there is no point in distinguishing between files and dirs

I prefer to keep the current variable name since it handles dis as well.

cvat/apps/engine/views.py
Marishka17: I think this comment should be addressed in a separate PR since it's not the purpose of this PR. I just changed a bit of code to not call `os.path.join` twice, the logic has not been changed.

site/content/en/docs/administration/advanced/upgrade_guide.md
Marishka17: > Maybe just add 3 options for simplicity. Organize them in tabs.

Sounds good, okay

site/content/en/docs/administration/advanced/upgrade_guide.md
Marishka17: In that case, we'll have to make this PR dependent on https://github.com/cvat-ai/cvat/pull/8898

cvat/settings/base.py
zhiltsov-max: > Are you also voting to move TmpDirManager into the engine in that case?

I just want each variable to have a specific owning component, which will be responsible for it's management and usage. But answering to the question - probably, in the current implementation, yes, because there are 2 clients to this functionality - events and exports. Anyway, I don't think it really has to be done in this PR to linger it more, we just need to separate `engine` and `dataset_manager` apps in further updates better.

site/content/en/docs/administration/advanced/upgrade_guide.md
```diff
@@ -56,6 +56,24 @@ To upgrade CVAT, follow these steps:
...
+  docker exec -it cvat_server python manage.py exportcachecleanup
+  {{< /tab >}}
+  {{< tab header="Kubernetes" >}}
+  cvat_backend_pod=$(kubectl get pods -l component=server -o 'jsonpath={.items[0].metadata.name}')
+  kubectl exec -it ${cvat_backend_pod} -- python manage.py exportcachecleanup
+  {{< /tab >}}
+  {{< tab header="Development" >}}
+  python manage.py exportcachecleanup
+  {{< /tab >}}
+{{< /tabpane >}}
```
zhiltsov-max, line 75:
```suggestion
<!--lint disable no-undefined-references-->

{{< tabpane lang="shell" >}}
  {{< tab header="Docker" >}}
  docker exec -it cvat_server python manage.py exportcachecleanup
  {{< /tab >}}
  {{< tab header="Kubernetes" >}}
  cvat_backend_pod=$(kubectl get pods -l component=server -o 'jsonpath={.items[0].metadata.name}')
  kubectl exec -it ${cvat_backend_pod} -- python manage.py exportcachecleanup
  {{< /tab >}}
  {{< tab header="Development" >}}
  python manage.py exportcachecleanup
  {{< /tab >}}
{{< /tabpane >}}

<!--lint enable no-undefined-references-->
```

It looks like there is some bug in remarklint.

cvat/apps/dataset_manager/management/commands/exportcachecleanup.py
Marishka17: > This might not work on mac and windows

I've checked this on mac and it works as expected. I also do not see the reason why it might not work on different terminals. I thought that carriage return should work similar on different operating systems/terminals.

> Frankly, I don't see a reason to get fancy with the output at all. If it was me, I would just print a message for every processed task/job/project and be done with it

I don't see the benefit of printing a message for every processed project/task/job when the number of objects is quite large. It's just a useless output.
So, if we are not going to add `tqdm` as a new dependency right now, I prefer to keep the current implementation.

cvat/apps/engine/management/commands/runperiodicjob.py
```diff
@@ -10,14 +10,15 @@ class Command(BaseCommand):
...
+        parser.add_argument("job_args", nargs="*", help="Arguments to pass to the job")
 
     def handle(self, *args, **options):
         job_id = options["job_id"]
 
         for job_definition in settings.PERIODIC_RQ_JOBS:
             if job_definition["id"] == job_id:
                 job_func = import_string(job_definition["func"])
-                job_func()
+                job_func(*options["job_args"])
```
SpecLad, line 21: I don't think this is right - you should be passing args from the job definition, not from the command line. Also, what about kwargs?

cvat/apps/dataset_manager/cron.py
```diff
@@ -0,0 +1,207 @@
...
+        if CleanupType.EXPORT_CACHE == self.value:
+            ThreadClass = CleanupExportCacheThread
+        elif CleanupType.TEMP_DIRECTORY == self.value:
+            ThreadClass = CleanupTmpDirThread
+        else:
+            raise ValueError(f"Unknown cleaning type: {self.value}")
+        return ThreadClass
+
+
+def cleanup(cleanup_type: CleanupType) -> None:
```
SpecLad, line 173:
```suggestion
def cleanup(cleanup_type: str) -> None:
```

cvat/apps/dataset_manager/cron.py
```diff
@@ -0,0 +1,207 @@
...
+    started_at = timezone.now()
+    stop_event = Event()
+    ThreadClass = CleanupType(cleanup_type).to_thread_class()
+    cleanup_thread = ThreadClass(stop_event=stop_event)
```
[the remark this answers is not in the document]
Marishka17: I agree that from the practical point of view, there is no reason for that. However, I decided to handle job TTL this way to finish the job correctly. If a cron job is not completed after set timeout (2h by default for the `cleaning` queue ), it will be moved to the `FailedJobRegistry`. This is not correct, because the job was done partially.

cvat/apps/dataset_manager/management/commands/exportcachecleanup.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def handle(self, *args, **options):
+        def update_progress():
+            progress = (i + 1) / objects_count
+            done = int(progress_bar_len * progress)
+            progress_bar = "#" * done + "-" * (progress_bar_len - done)
+            self.stdout.write(f"\rProgress: |{progress_bar}| {progress:.0%}", ending="")
+
+        now = timezone.now()
+        progress_bar_len = os.get_terminal_size().columns // 2
```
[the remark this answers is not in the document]
Marishka17: fixed

cvat/apps/dataset_manager/util.py
```diff
@@ -147,94 +154,206 @@ def get_export_cache_lock(
         if acquired:
             lock.release()
 
+class OperationType(str, Enum):
+    EXPORT = "export"
 
-EXPORT_CACHE_DIR_NAME = 'export_cache'
+    @classmethod
+    def values(cls) -> list[str]:
```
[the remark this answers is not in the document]
Marishka17: yes, it was used before, but I forgot to delete the method after refactoring :eyes:

cvat/apps/dataset_manager/cron.py
```diff
@@ -0,0 +1,207 @@
...
+
+    if rq_job := get_current_job():
+        seconds_left = rq_job.timeout - 60
+        sleep_interval = 10
+        assert seconds_left > sleep_interval
+        finish_before = started_at + timedelta(seconds=seconds_left)
+        cleanup_thread.start()
+
+        while timezone.now() < finish_before:
```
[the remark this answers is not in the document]
Marishka17: You're right, fixed

cvat/apps/dataset_manager/cron.py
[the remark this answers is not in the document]
SpecLad: I would argue that a job that was only done partially is not successful. Perhaps it's not a problem if it only happens on an occasion, but what if the cleanup job repeatedly fails to process the entire directory? That indicates some problem with the system, and marking the job as successful would mask that problem.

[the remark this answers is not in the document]
SpecLad: Oh, I just realized... even if you want to implement the logic to stop before the timeout, it still seems unnecessary to use threads. You can run the logic in the main thread, and instead of `if self._stop_event.is_set()` you check if you're out of time. This would make the code much simpler.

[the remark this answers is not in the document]
Marishka17: Okay, let's not mask that

site/content/en/docs/administration/advanced/upgrade_guide.md
Marishka17: I'm not sure that this is a real problem. I don't think there will be many such RQ jobs with `Queued` status when upgrading to the next release.  Additionally, I don't think that someone uses frequently the django admin panel to re-enqueue the job manually. In other cases, everything should work.
But, if @SpecLad and @zhiltsov-max are voting to add such a migration (e.g. as an example of the first Redis migration, I can do that)

Marishka17: Probably, we should also extract these documentation changes into a separate PR that will be merged on the same day of release.

SpecLad: I'm not voting for that (and I also don't want to make this PR dependent on another one...).

I also doubt that it's a big problem, and even if it is, it seems like there's a much simpler solution - just add the parameter back (and ignore it).

SpecLad (approved):

zhiltsov-max (approved):

PR #8807, opened 2024-12-10
Save drawn shape on submit in `single shape` mode

### LINKED ISSUE #8713: Finish current annotation on the last frame in single shape mode on Submit click

### Steps to Reproduce

1. Create a task with 1 label of type "points"
2. Create a gt job
2. Try to annotate a normal job in Single Shape mode (uncheck the "Predefined number of points" checkbox)
3. Go to the last frame
4. Put a point 
5. Click Submit

The last frame annotation will not be saved.

### Expected Behavior

_No response_

### Possible Solution

In single shape mode, when multiple points should be put, when the last frame frame is annotated, it would be nice to finish and save the current annotation, even without clicking N or Done, if the Submit button is clicked.

### Context

_No response_

### Environment

_No response_

klakhov:
### Motivation and context

Resolved #8713

It would make sense to save the currently drawn shape (e.g., points, polygon, polyline) when clicking `Submit` on the last frame. On non-last frames, we click `Done` to finish drawing, but on the last frame, we need to click `Done` and then `Submit`. If `Done` is not clicked, the shape is lost, which is inconvenient for users.

### Checklist

### License

## Summary by CodeRabbit

- **New Features**
	- Enhanced drawing functionality in the sidebar for more flexible shape annotation.
	- Added a "Done" button that dynamically appears based on current drawing state.
  
- **Bug Fixes**
	- Streamlined logic for determining drawing completion and button visibility.

- **Chores**
	- Introduced utility functions to improve drawing operations and interaction handling.


bsekachev:
The patch does not seem to work.

First `activeControl` is not dependency of `finishOnThisFrame`. So, it is probably will be outdated.
Secondly, `cancelCurrentCanvasOp` is called before from another hook (probably because `annotations` array gets updated)

bsekachev:
demo

klakhov:
I’ve fixed the issue mentioned above. The problem was with the `activeControl` dependency, and `activeShapeType` was incorrectly determined based on the label type in the global state. If the label type is `any`, `activeShapeType` should correspond to the current label type selected.

@zhiltsov-max Could you check if fix works for you?

zhiltsov-max:
Looks ok, but it probably works incorrectly if there is predefined number of points used. If set to > 1, the shape is saved automatically even with less number of points. I think maybe it will be better to disable skip/submit if there are not enough points placed.

klakhov:
As discussed earlier, the issue Maxim pointed out falls outside the scope of this patch. There are multiple ways to break the predefined number of points in a shape in the UI—beyond the single shape mode saving on the last frame. For example, deleting a point in standard mode can also cause issues.

If we want to make the predefined number of points feature more robust, we’d need to implement a more sophisticated solution that binds the number of points directly to the label model.

For this patch, however, it’s simply a small UX improvement that removes the need for an extra click. I believe we can proceed with reviewing and merging it.

cvat-ui/src/utils/drawing.ts
```diff
@@ -0,0 +1,29 @@
+// Copyright (C) 2024 CVAT.ai Corporation
```
bsekachev, line 1:
```suggestion
// Copyright (C) CVAT.ai Corporation
```

PR #8898, opened 2025-01-03
Support redis migrations

Marishka17:
### Motivation and context

Sometimes we need to modify the data stored in Redis (for instance, in the near future we will change the identifiers of RQ jobs). This PR introduces a common mechanism for handling Redis migrations.
### Checklist

### License


cvat/apps/engine/management/commands/migrateredis.py
```diff
@@ -0,0 +1,50 @@
...
+        raise Exception(f"Invalid migration: {file_path}")
+
+    return MigrationClass
+
+
+class Command(BaseCommand):
+    help = "Applies Redis migrations and records them in the database"
+
+    def handle(self, *args, **options) -> None:
+        migrations_dir = Path(settings.REDIS_MIGRATIONS_ROOT)
```
SpecLad, line 32: This seems a bit inflexible, because it only allows for one migration root. What will we do if we need to add migrations to the Enterprise repository, for example?

The database migration functionality in Django loads migrations from each app; perhaps we could do the same here?

cvat/apps/engine/management/commands/migrateredis.py
```diff
@@ -0,0 +1,50 @@
...
+        applied_migrations = RedisMigration.objects.all().values_list("name")
+
+        for migration_file in sorted(migrations_dir.glob("[0-9]*.py")):
+            migration_name = migration_file.stem
+            if migration_name in applied_migrations:
+                continue
+
+            migration = get_migration_class(module_name=migration_name, file_path=migration_file)
+            try:
+                migration.run()
```
SpecLad, line 42: For reviewing purposes, it would be useful to see an example of a migration class.

cvat/apps/engine/management/commands/migrateredis.py
```diff
@@ -0,0 +1,50 @@
...
+                RedisMigration.objects.create(name=migration_name)
+                self.stdout.write(
+                    self.style.SUCCESS(f"Successfully applied migration: {migration_name}")
+                )
+            except Exception as e:
+                self.stderr.write(self.style.ERROR(f"Failed to apply migration: {migration_name}"))
+                self.stderr.write(str(e))
```
SpecLad, line 49: A full traceback would be useful here, and not just the message.

cvat/apps/engine/management/commands/migrateredis.py
```diff
@@ -0,0 +1,50 @@
...
+                break
```
SpecLad, line 50: Shouldn't an exception cause the command to exit with a failure status? Right now it seems to be exiting successfully.

cvat/apps/engine/models.py
```diff
@@ -1345,3 +1345,9 @@ class RequestSubresource(TextChoices):
     ANNOTATIONS = "annotations"
     DATASET = "dataset"
     BACKUP = "backup"
+
+
+class RedisMigration(models.Model):
+    # todo: redis_inmem/redis_ondisk
```
SpecLad, line 1351: What do you mean by this?

Marishka17: Initially, I thought that it could be useful to store to which Redis instance a migration was applied

cvat/apps/engine/management/commands/migrateredis.py
Marishka17: of course, it was a draft code

cvat/apps/engine/management/commands/migrateredis.py
Marishka17: I've added a migration to remove `clear_export_cache` scheduled jobs

cvat/apps/engine/management/commands/migrateredis.py
Marishka17: Good point, fixed

cvat/apps/redis_handler/migration_loader.py
```diff
@@ -0,0 +1,96 @@
+# Copyright (C) 2025 CVAT.ai Corporation
+#
+# SPDX-License-Identifier: MIT
+
+import importlib
+from pathlib import Path
+from typing import Any, Generator, Mapping
+
+from django.apps import AppConfig, apps
+from django.utils.module_loading import module_has_submodule
```
SpecLad, line 10: This is not a documented function, so I would advise against using it.

cvat/apps/redis_handler/migration_loader.py
```diff
@@ -0,0 +1,96 @@
...
+            return {app_config.label: app_config for app_config in self._app_configs}
+
+        def __getitem__(self, label: str) -> AppConfig:
+            return self._app_configs_mapping[label]
+
+        def __iter__(self) -> Generator[AppConfig, Any, Any]:
+            yield from self._app_configs
+
+    def __init__(self) -> None:
+        self._app_configs = self.AppConfigs()
```
SpecLad, line 51: Does this need to be a specialized object? It seems like it's only used once, as a map, so it could just be a regular dict.

dev/format_python_code.sh
```diff
@@ -37,6 +37,7 @@ for paths in \
     "cvat/apps/events/signals.py" \
     "cvat/apps/engine/management/commands/syncperiodicjobs.py" \
     "cvat/apps/dataset_manager/management/commands/cleanuplegacyexportcache.py" \
+    "cvat/apps/redis_handler/**/*.py" \
```
SpecLad, line 40:
```suggestion
    "cvat/apps/redis_handler/" \
```

bash doesn't know about the `**` syntax.

cvat/apps/redis_handler/models.py
```diff
@@ -0,0 +1,15 @@
+from django.db import models
+
+
+class RedisMigration(models.Model):
```
SpecLad, line 4: Have you considered storing information about Redis migrations in Redis? It seems to me that storing them in the database could lead to unwanted side effects. For example, if the DB is rolled back to an earlier snapshot, it could lead to some migrations being applied again.

cvat/apps/redis_handler/migration_loader.py
```diff
@@ -0,0 +1,96 @@
...
+        for app_config in self._app_configs:
+            migrations_dir = Path(app_config.path) / self.REDIS_MIGRATIONS_DIR_NAME
+            for migration_file in sorted(migrations_dir.glob("[0-9]*.py")):
+                migration_name = migration_file.stem
+                (self._disk_migrations_per_app.setdefault(app_config.label, [])).append(
+                    migration_name
+                )
+
+    def _init_unapplied_migrations(self):
+        applied_migrations = DBRedisMigration.objects.all().values_list("name", "app_label")
```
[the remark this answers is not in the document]
Marishka17: I thought that `values_list` is the optimization that can reduce the memory usage, but okay, here we can do without it.

cvat/apps/redis_handler/migration_loader.py
```diff
@@ -0,0 +1,96 @@
...
+from pathlib import Path
+from typing import Any, Generator, Mapping
+
+from django.apps import AppConfig, apps
+from django.utils.module_loading import module_has_submodule
+
+from cvat.apps.redis_handler.models import RedisMigration as DBRedisMigration
+from cvat.apps.redis_handler.redis_migrations import BaseMigration as BaseRedisMigration
+
+_MigrationsPerApp = dict[int, list[str]]
```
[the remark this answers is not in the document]
Marishka17: > Shouldn't the key type be str?

I've used Indexes before and forgot to update type annotations after refactoring..

cvat/apps/redis_handler/management/commands/migrateredis.py
```diff
@@ -0,0 +1,44 @@
...
+                    )
+                )
+            except Exception as ex:
+                self.stderr.write(
+                    self.style.ERROR(
+                        f"[{migration.app_label}] Failed to apply migration: {migration.name}"
+                    )
+                )
+                self.stderr.write(self.style.ERROR(f"\n{traceback.format_exc()}"))
+                raise CommandError(str(ex))
```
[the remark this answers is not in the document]
Marishka17: From my point of view, it makes traceback more readable:
image
vs
image

cvat/apps/redis_handler/migration_loader.py
```diff
@@ -0,0 +1,96 @@
...
+
+import importlib
+from pathlib import Path
+from typing import Any, Generator, Mapping
+
+from django.apps import AppConfig, apps
+from django.utils.module_loading import module_has_submodule
+
+from cvat.apps.redis_handler.models import RedisMigration as DBRedisMigration
+from cvat.apps.redis_handler.redis_migrations import BaseMigration as BaseRedisMigration
```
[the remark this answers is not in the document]
Marishka17: I just wanted to make these class names less similar. Probably right now it is not quite obvious which class should be used to retrieve applied migrations from DB and which as a base class for Redis migration.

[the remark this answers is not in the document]
SpecLad: This suggests to me that one of them should be renamed. Perhaps `RedisMigration` to `AppliedMigration`?

cvat/apps/redis_handler/models.py
Marishka17: > Have you considered storing information about Redis migrations in Redis? 

I thought about that, but the same problem exists if Redis is reset. + I assumed that we could use these migrations to modify data stored in Redis on disk (since this service also supports Redis API), for instance, when we will need to update chunks/previews keys.

cvat/apps/redis_handler/migration_loader.py
Marishka17: removed

cvat/apps/redis_handler/management/commands/migrateredis.py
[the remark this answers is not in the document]
SpecLad: Okay, fair enough.

cvat/apps/redis_handler/models.py
SpecLad: > I thought about that, but the same problem exists if Redis is reset.

Does it? If Redis is reset to nothing, it should be safe to apply all migrations again - after all, that's what you would do when creating a fresh CVAT instance.

> I assumed that we could use these migrations to modify data stored in Redis on disk (since this service also supports Redis API), for instance, when we will need to update chunks/previews keys.

I'm not sure it's practical to implement such migrations. On-disk Redis is used as a cache, so if we need to change the schema in some way, we should be able to just ignore old cache entries and let them expire.

backend_entrypoint.sh
```diff
@@ -20,6 +20,7 @@ cmd_init() {
     ~/manage.py migrate
 
     wait-for-it "${CVAT_REDIS_INMEM_HOST}:${CVAT_REDIS_INMEM_PORT:-6379}" -t 0
+    ~/manage.py migrateredis
```
SpecLad, line 23: The logic in this script also waits for migrations to complete before starting the backend services. Perhaps we should implement something similar for Redis migrations?

Marishka17: I'll check

cvat/apps/engine/redis_migrations/001_cleanup_scheduled_jobs.py
```diff
@@ -0,0 +1,20 @@
...
+from django.conf import settings
+from rq_scheduler import Scheduler
+
+from cvat.apps.redis_handler.redis_migrations import BaseMigration
+
+
+class Migration(BaseMigration):
+    @staticmethod
+    def run():
+        scheduler: Scheduler = django_rq.get_scheduler(settings.CVAT_QUEUES.EXPORT_DATA.value)
```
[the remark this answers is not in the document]
Marishka17: I've added the ability to share a connection, but it looks like it isn't possible to pass connection to `get_scheduler`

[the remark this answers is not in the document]
SpecLad: Ah, right. I actually added this parameter to django-rq, but it looks like it's only been released in 3.0. Perhaps we could make use of it later.

Marishka17:
I have no time to investigate what is going on with Helm tests

cvat/apps/redis_handler/models.py
Marishka17: done

backend_entrypoint.sh
Marishka17: done

cvat-ui/src/components/export-backup/styles.scss
Marishka17, line 1: It looks like these changes were made by pre-commit hook
@bsekachev

cvat/apps/redis_handler/management/commands/migrateredis.py
```diff
@@ -0,0 +1,76 @@
...
+
+                    # add migration to applied ones
+                    applied_migration = AppliedMigration(
+                        name=migration.name,
+                        app_label=migration.app_label,
+                        applied_date=timezone.now(),
+                    )
+                    applied_migration_key = applied_migration.get_key()
+                    pipe.hset(applied_migration_key, mapping=applied_migration.to_dict())
+                    pipe.zadd(applied_migration.SORTED_SET_KEY, {applied_migration_key: 1})
```
SpecLad, line 60: Seems like a good candidate for a method - e.g. `applied_migration.save()`. There's a lot of implementation details here that the management command probably doesn't need to know about.

cvat/apps/redis_handler/migration_loader.py
```diff
@@ -0,0 +1,134 @@
...
+class AppliedMigration:
+    SORTED_SET_KEY: ClassVar[str] = "cvat:applied_migrations:"
+    KEY_PREFIX: ClassVar[str] = "cvat:applied_migration:"
+
+    name: str = field(validator=[validators.instance_of(str), validators.max_len(128)])
+    app_label: str = field(validator=[validators.instance_of(str), validators.max_len(128)])
+    applied_date: datetime = field(
+        validator=[validators.instance_of(datetime)], converter=to_datetime
+    )
+    identifier: UUID = field(factory=uuid4, converter=to_uuid)
```
SpecLad, line 45: Why do we need an extra identifier? Shouldn't `<app_label>.<name>` already be a unique key?

cvat/apps/redis_handler/migration_loader.py
```diff
@@ -0,0 +1,134 @@
...
+def to_uuid(value: str | UUID) -> UUID:
+    if isinstance(value, UUID):
+        return value
+
+    return UUID(value)
+
+
+@frozen
+class AppliedMigration:
+    SORTED_SET_KEY: ClassVar[str] = "cvat:applied_migrations:"
```
SpecLad, line 37:
```suggestion
    SORTED_SET_KEY: ClassVar[str] = "cvat:applied_migrations"
```

cvat/apps/redis_handler/migration_loader.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def _init_applied_migrations(self):
+        conn = Redis(
+            host=settings.REDIS_INMEM_SETTINGS["HOST"],
+            port=settings.REDIS_INMEM_SETTINGS["PORT"],
+            db=settings.REDIS_INMEM_SETTINGS["DB"],
+            password=settings.REDIS_INMEM_SETTINGS["PASSWORD"],
+        )
+        applied_migration_keys = [
+            i.decode("utf-8") for i in conn.zrange(AppliedMigration.SORTED_SET_KEY, 0, -1)
```
SpecLad, line 100: Does the set actually need to be sorted? Doesn't seem like the order matters.

cvat/apps/redis_handler/management/commands/migrateredis.py
```diff
@@ -0,0 +1,76 @@
...
+            if not loader:
+                return
+
+            sys.exit(1)
+
+        if not loader:
+            self.stdout.write("No migrations to apply")
+            return
+
+        conn = Redis(
```
SpecLad, line 40: Seems like the loader could reuse this connection.

cvat/apps/redis_handler/migration_loader.py
```diff
@@ -0,0 +1,134 @@
...
+                AppliedMigration(
+                    **{k.decode("utf-8"): v.decode("utf-8") for k, v in conn.hgetall(key).items()}
+                )
+            )
+
+    def _init_unapplied_migrations(self):
+        for app_label, migration_names in self._disk_migrations_per_app.items():
+            app_config = self._app_config_mapping[app_label]
+            app_applied_migrations = {
+                m.name for m in self._applied_migrations if m.app_label == app_config.label
```
SpecLad, line 113: It seems inefficient to search the entire list of applied migrations for every app. Given that this is the only use of `_applied_migrations`, couldn't you build `_applied_migrations` as a map from app names to the set of applied migration names?

cvat/apps/redis_handler/migration_loader.py
Marishka17: in the current use case - yes, changed to `set`

cvat/apps/redis_handler/migration_loader.py
Marishka17: Yes, it should, probably I did so initially because I would not keep only the applied date in a hash item

Marishka17:
It seems that [these](https://github.com/cvat-ai/cvat/actions/runs/13054237659/job/36421277781) tests are unstable. I've compared the latest logs with logs where there was a problem with helm tests related to resetting applied migrations from the Redis and waiting in workers for them. Workers are initialized correctly and start running.
https://github.com/cvat-ai/cvat/actions/runs/12833455817
https://github.com/cvat-ai/cvat/actions/runs/13054237659

SpecLad (approved):

Issue #8894, opened 2024-12-31
"Tracking error" in OpenCV TrackerMIL tracker when tracking many concurrently moving objects


archibald1418:
### Actions before raising this issue

- [X] I searched the existing issues and did not find anything similar.
- [X] I read/searched [the docs](https://docs.cvat.ai/docs/)

### Steps to Reproduce

On Google Chrome:

- Create a dataset with several objects moving in frame
- Load OpenCV and start tracking it with TrackerMIL as per [docs](https://docs.cvat.ai/docs/manual/advanced/ai-tools/#opencv-annotate-with-trackers)
- Repeat frame by frame
- For each new tracked object add a TrackerMIL tracker and continue frame by frame
- Error starts appearing from around 6 objects being tracked simultaneously (maybe it's my Chrome though, could happen later)
- Also, already with a smaller number of objects, TrackerMIL gets buggy while an object is leaving the frame or some shapes start lagging behind their objects (this was also an issue in https://github.com/cvat-ai/cvat/pull/7399)

![telegram-cloud-photo-size-2-5449446186269405091-y](https://github.com/user-attachments/assets/c11b6f01-73e7-4da6-8018-087a92a695c8)
![telegram-cloud-photo-size-2-5449446186269404791-y](https://github.com/user-attachments/assets/36b4a064-8c8e-4c98-a866-f14eafc7e499)

### Expected Behavior

No tracker lags and no 'Tracking errors' pop up

PR #8934, opened 2025-01-13
Testing OpenCV TrackerMIL 'Tracking error' on high resolution

archibald1418:
This contributes a regression test for reproducing a bug from #8894 which was fixed in #8942

### Motivation and context
A 'Tracking error' message notification popping up when tracking shapes with OpenCV's TrackerMIL.
Its cause was shown to be memory issues when using an OpenCV tracker on frames with high resolution (ca. 5000x5000)

### How has this been tested?

Work in progress

### Checklist

### License


.vscode/tasks.json
archibald1418, line 1: Here the pattern didn't work - the vscode task would go into an endless loop

bsekachev:
The test will not be relevant after the fix https://github.com/cvat-ai/cvat/pull/8942.
It needs to be updated slightly.

But anyway, my general comment is that we do not need two tests doing basically the same thing (but with different resolution).
I will recommend just to update existing test to use high resolution images there.

tests/cypress/plugins/imageGenerator/addPlugin.js
```diff
@@ -4,8 +4,11 @@
 
 // eslint-disable-next-line no-use-before-define
 exports.imageGenerator = imageGenerator;
+// eslint-disable-next-line no-use-before-define
```
bsekachev, line 7: /* eslint no-use-before-define: 0 */   to remove this rule across the file

tests/cypress/plugins/imageGenerator/addPlugin.js
```diff
@@ -43,3 +55,18 @@ async function imageGenerator(args) {
...
+    const {
+        directory, fileName, extension, buffer,
+    } = args;
+    let file = null;
+    try {
+        fs.mkdirp(directory);
+        file = path.join(directory, `${fileName}.${extension}`);
+        const image = await createImageFromBuffer(Buffer.from(buffer.data));
+        image.write(file);
+        // eslint-disable-next-line no-empty
```
bsekachev, line 69: Why did you make  exception silent here?

archibald1418: If you remove it, it throws a warning on empty `catch (e) {}` block, same principle as on line 54

.vscode/launch.json
```diff
@@ -152,7 +152,7 @@
                 "run",
                 "--headless",
                 "--browser",
-                "chrome"
+                "chrome",
```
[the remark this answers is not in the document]
archibald1418: It's auto eslint fix on save, I will be careful with it

tests/cypress/plugins/imageGenerator/addPlugin.js
archibald1418: Oh, you mean the exception. I thought the author of `imageGenerator()` had his reasons

bsekachev: The test should fail in case of any errors. Otherwise it will be more difficult to debug potential issues. 
Imagine that the image or directory were not created here and then you see strange error in a test which needs to be investigated additionally to finally find that the root cause was here

archibald1418: Valid point, applying

archibald1418: Also applied to `imageGenerator` in the same manner

bsekachev (approved):

PR #8953, opened 2025-01-16
[Dependent] Consensus simple merging

zhiltsov-max:
### Checklist

### License


cvat-ui/src/containers/actions-menu/actions-menu.tsx
```diff
@@ -36,6 +38,7 @@ interface DispatchToProps {
     openRunModelWindow: (taskInstance: any) => void;
     deleteTask: (taskInstance: any) => void;
     openMoveTaskToProjectWindow: (taskInstance: any) => void;
+    mergeConsensusJobs: (taskInstance: any) => void;
```
klakhov, line 41:
```suggestion
    mergeConsensusJobs: (taskInstance: Task) => void;
```

cvat-ui/src/reducers/consensus-reducer.ts
```diff
@@ -0,0 +1,104 @@
...
+
+import { ConsensusActions, ConsensusActionTypes } from 'actions/consensus-actions';
+import { ConsensusState } from '.';
+
+const defaultState: ConsensusState = {
+    taskInstance: null,
+    jobInstance: null,
+    fetching: true,
+    consensusSettings: null,
+    mergingConsensus: {},
```
klakhov, line 13: Usually we do `activities: { action: {} }` `merging` in our case

cvat-ui/src/reducers/consensus-reducer.ts
```diff
@@ -0,0 +1,104 @@
...
+};
+
+function makeKey(id: number, instance: string): string {
```
klakhov, line 16: Better to pass instance itself here

cvat-ui/src/reducers/index.ts
```diff
@@ -170,6 +171,16 @@ export interface ImportState {
     instanceType: 'project' | 'task' | 'job' | null;
 }
 
+export interface ConsensusState {
+    fetching: boolean;
+    consensusSettings: ConsensusSettings | null;
+    taskInstance: Task | null;
+    jobInstance: Job | null;
+    mergingConsensus: {
+        [taskId: string]: boolean;
```
klakhov, line 180:
```suggestion
        [instanceId: string]: boolean;
```

cvat-core/src/server-proxy.ts
```diff
@@ -767,6 +768,41 @@ async function deleteTask(id: number, organizationID: string | null = null): Pro
...
+    };
+
+    if (instanceType === 'task') requestBody.task_id = id;
+    else requestBody.job_id = id;
+
+    return new Promise<void>((resolve, reject) => {
+        async function request() {
+            try {
+                const response = await Axios.post(url, requestBody, { params });
+                params.rq_id = response.data.rq_id;
```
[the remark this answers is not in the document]
zhiltsov-max: Currently, the reason is that Requests API v2 is not implemented yet. It's hard to integrate anything except import or export into the same server code.

cvat-ui/src/components/jobs-page/job-card.tsx
```diff
@@ -95,7 +95,12 @@ function JobCardComponent(props: Props): JSX.Element {
             <Dropdown
                 trigger={['click']}
                 destroyPopupOnHide
-                overlay={<JobActionsMenu job={job} />}
+                overlay={(
+                    <JobActionsMenu
+                        job={job}
+                        consensusJobsPresent={false}
```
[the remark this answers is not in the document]
zhiltsov-max: Probably, yes, but here it's intentionally always false, because on the Jobs page you can't meaningfully use this merge button without seeing nested jobs.

supervisord/worker.consensus.conf
SpecLad, line 1: Do we really need another worker for this? Can't an existing worker, like the quality report worker, handle this?

cvat/schema.yml
```diff
@@ -1049,6 +1049,176 @@ paths:
...
+    post:
+      operationId: consensus_create_merge
+      summary: Create a consensus merge
+      parameters:
+      - in: query
+        name: rq_id
+        schema:
+          type: string
+        description: |
+          The consensus merge request id. Can be specified to check operation status.
```
SpecLad, line 1062: We should be using the request API for status checking.

cvat/apps/consensus/models.py
```diff
@@ -0,0 +1,50 @@
...
+
+        existing_fields = {f.name for f in cls._meta.fields}
+        return {k: v for k, v in default_settings.items() if k in existing_fields}
+
+    def to_dict(self):
+        return model_to_dict(self)
+
+    @property
+    def organization_id(self):
+        return getattr(self.task.organization, "id", None)
```
SpecLad, line 50:
```suggestion
        return self.task.organization_id
```

cvat/apps/engine/models.py
```diff
@@ -453,7 +453,7 @@ def get_tmp_dirname(self) -> str:
 
 
 @transaction.atomic(savepoint=False)
-def clear_annotations_in_jobs(job_ids):
+def clear_annotations_in_jobs(job_ids: Iterable[int] | Iterator[int]):
```
SpecLad, line 456: `Iterator` is a subclass of `Iterable`, so this is equivalent to `Iterable[int]`.

cvat/apps/engine/models.py
```diff
@@ -14,7 +14,7 @@
 from collections.abc import Collection, Sequence
 from enum import Enum
 from functools import cached_property
-from typing import Any, ClassVar, Optional
+from typing import Any, ClassVar, Iterable, Iterator, Optional
```
SpecLad, line 17: Please use `collections.abc`.

cvat/schema.yml
zhiltsov-max: Currently, the reason is that Requests API v2 is not implemented yet. It's hard to integrate anything except import or export into the same server code. Supporting other use cases is one of the key goals of v2.

supervisord/worker.consensus.conf
zhiltsov-max: Currently, we add a worker almost for any new activity, so this PR does the same. Probably, it can be optimized.

cvat/schema.yml
zhiltsov-max: This will create one more API style. This one is used for quality reports, job validations, and analytics.

cvat/schema.yml
```diff
@@ -7446,6 +7616,33 @@ components:
...
+          readOnly: true
+        iou_threshold:
+          type: number
+          format: double
+          description: Pairwise annotation matching IoU threshold
+        quorum:
+          type: number
+          format: double
+          description: |
+            Minimum required percent of sources having an annotation for it to be accepted
```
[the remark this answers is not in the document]
zhiltsov-max: It's a number in the range [0;1], actually. This description is a little bit more UI friendly, but it's possible to change the word and replace it in UI settings instead.

cvat/apps/consensus/rules/consensus_settings.rego
```diff
@@ -0,0 +1,104 @@
...
+    org := input.auth.organization
+    qobject := [
+        {"task__organization": org.id},
+        {"task__project__organization": org.id}, "|",
+    ]
+} else := qobject if {
+    utils.is_sandbox
+    user := input.auth.user
+    qobject := [
+        {"task__owner_id": user.id},
```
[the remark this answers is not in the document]
zhiltsov-max: It's hard to say, it's just a list endpoint for a model. Maybe someone needs to list settings in a project or in an org.

cvat/apps/consensus/serializers.py
```diff
@@ -0,0 +1,53 @@
...
+            """,
+        }.items():
+            extra_kwargs.setdefault(field_name, {}).setdefault(
+                "help_text", textwrap.dedent(help_text.lstrip("\n"))
+            )
+
+    def validate(self, attrs):
+        for k, v in attrs.items():
+            if (k.endswith("_threshold") or k == "quorum") and not 0 <= v <= 1:
+                raise serializers.ValidationError(f"{k} must be in the range [0; 1]")
```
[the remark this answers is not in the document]
zhiltsov-max: Ok, good idea

cvat/apps/consensus/signals.py
```diff
@@ -0,0 +1,22 @@
...
+from cvat.apps.engine.models import Task
+
+
+@receiver(
+    post_save,
+    sender=Task,
+    dispatch_uid=__name__ + ".save_task-initialize_consensus_settings",
+)
+def __save_task__initialize_consensus_settings(instance: Task, created, **kwargs):
+    # Initializes default quality settings for the task
```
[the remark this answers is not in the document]
zhiltsov-max: Good idea, added.

cvat/schema.yml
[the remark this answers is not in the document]
SpecLad: I mean... the description is for the API, not the UI. So it should describe what the API accepts.

cvat/apps/consensus/serializers.py
```diff
@@ -0,0 +1,53 @@
...
+            "id",
+            "task_id",
+        )
+
+        extra_kwargs = {k: {"required": False} for k in fields}
+
+        for field_name, help_text in {
+            "iou_threshold": "Pairwise annotation matching IoU threshold",
+            "quorum": """
+                Minimum required percent of sources having an annotation for it to be accepted
```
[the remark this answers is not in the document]
SpecLad: I accidentally left basically the same comment twice, so I'll close this one.

cvat/apps/consensus/models.py
```diff
@@ -0,0 +1,44 @@
...
+class ConsensusSettings(models.Model):
+    task = models.ForeignKey(Task, on_delete=models.CASCADE, related_name="consensus_settings")
+    quorum = models.FloatField(default=0.5)
+    iou_threshold = models.FloatField()
+
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        defaults = deepcopy(self.get_defaults())
+        for field in self._meta.fields:
+            if field.name in defaults:
+                field.default = defaults[field.name]
```
SpecLad, line 25: Does this change the class metadata every time an instance is created? It seems unnecessarily convoluted to me.

Why couldn't you just put the defaults in the model fields, e.g. `default=qc.DatasetComparator.DEFAULT_SETTINGS.iou_threshold`?

cvat/apps/consensus/models.py
```diff
@@ -0,0 +1,44 @@
...
+    def get_defaults(cls) -> dict:
+        # Here we use the same annotation matching settings as in quality control
+        import cvat.apps.quality_control.quality_reports as qc
+
+        default_settings = qc.DatasetComparator.DEFAULT_SETTINGS.to_dict()
+
+        existing_fields = {f.name for f in cls._meta.fields}
+        return {k: v for k, v in default_settings.items() if k in existing_fields}
+
+    def to_dict(self):
```
SpecLad, line 39: Where is this used?

cvat/apps/engine/utils.py
```diff
@@ -148,13 +148,16 @@ def parse_exception_message(msg: str) -> str:
         pass
     return parsed_msg
 
-def process_failed_job(rq_job: Job):
+def process_failed_job(rq_job: Job, *, logger: logging.Logger | None = None) -> str:
```
SpecLad, line 151: Why do this? I think we should just let each module log with its own logger.

supervisord/worker.consensus.conf
SpecLad: Well, it has to stop somewhere. We already have CI straining to support all existing workers.

cvat/apps/engine/utils.py
zhiltsov-max: Yes, exactly to allow each module to use the corresponding logger. It was a fixed logger before.

supervisord/worker.consensus.conf
zhiltsov-max: Ok, so your proposal is to use the quality reports worker and rename it? It will go in a separate PR then.

cvat/apps/consensus/models.py
zhiltsov-max: The idea was to decouple db-related stuff from the domain logic. Currently, the default values are stored in a domain class.

cvat/apps/consensus/models.py
zhiltsov-max: It's a part of reports, should be cleaned from this PR.

cvat/apps/consensus/models.py
SpecLad: > The idea was to decouple db-related stuff from the domain logic. Currently, the default values are stored in a domain class.

But how does it decouple it? You're still setting the field defaults, just in a roundabout way.

I don't think the default values specified in the field definition are stored in the database, if that's what you're saying.

> Yes, probably.

Well, can we not do that? Creating a class instance should not modify the class, that seems really counter-intuitive.

I could see at least two alternatives (not counting my proposal in the initial comment):

1. Replace the line this conversation is attached to with `kwargs.setdefault(field.name, defaults[field.name])`.
2. Keep the code as-is, but do it only once (after the class definition).

> Probably, this variant requires more code in a long run (see quality models).

By "more code", do you mean repeating `default=qc.DatasetComparator.DEFAULT_SETTINGS.xxxx` for every field? It doesn't seem too bad to me, and it has the advantage of being straightforward.

cvat/apps/engine/utils.py
SpecLad: I mean... the module is also fixed. It's `cvat.apps.engine.utils`. 🙂

zhiltsov-max: The function is currently used from consensus and from engine. Consensus passes it's own logger. How would you like it to be changed?

cvat/apps/consensus/models.py
zhiltsov-max: Yes, the defaults declared this way are not stored in the DB, but they still generate migrations. Probably, initially I wanted to avoid migration changes on each default change, if they are any. I can see there are both positive and negative sided in the current solution, so I decided to use the simple variant for now and come back to this later.

> I could see at least two alternatives (not counting my proposal in the initial comment):
> Keep the code as-is, but do it only once (after the class definition).

I  think I can try to introduce a way for this later using a metaclass or some other solution later.

cvat/apps/consensus/serializers.py
```diff
@@ -0,0 +1,51 @@
...
+            "task_id",
+            "iou_threshold",
+            "quorum",
+        )
+        read_only_fields = (
+            "id",
+            "task_id",
+        )
+
+        extra_kwargs = {k: {"required": False} for k in fields}
```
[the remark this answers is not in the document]
zhiltsov-max: Ok, updated.

cvat/apps/engine/utils.py
SpecLad: I think we should keep the function the same as it was before. (`'cvat.server.engine'` ought to be replaced by `__name__`, but that's unrelated to this PR.)

zhiltsov-max: Then consensus will print errors into the engine logger.

SpecLad: What's the problem with that? All the logger does is identify the module that emitted the message. In this case, the module is `cvat.apps.engine.utils`.

zhiltsov-max: The problem is that the produced log messages are both misleading (the module name) and uninformative (no full stacktrace to the message source):

```
[2025-02-21 14:53:11,412] ERROR cvat.server.engine: Traceback (most recent call last):
  File "/opt/venv/lib/python3.10/site-packages/rq/worker.py", line 1431, in perform_job
    rv = job.perform()
  File "/opt/venv/lib/python3.10/site-packages/rq/job.py", line 1280, in perform
  File "/opt/venv/lib/python3.10/site-packages/rq/job.py", line 1317, in _execute
  File "/home/django/cvat/apps/profiler.py", line 12, in wrapped
    return f(*args, **kwargs)
  File "/home/django/cvat/apps/consensus/merging_manager.py", line 249, in _merge
    return _TaskMerger(task=job.get_task_id()).merge_single_consensus_job(target_id)
    return func(*args, **kwds)
  File "/home/django/cvat/apps/consensus/merging_manager.py", line 157, in merge_single_consensus_job
    self._merge_consensus_jobs(parent_job_id)
  File "/home/django/cvat/apps/consensus/merging_manager.py", line 102, in _merge_consensus_jobs
    raise Exception("test")
Exception: test
```

Another problem I see is that they also may expose sensitive information, as they are returned from the API for any exception type and show the full stacktrace. 

I decided to revert the change, because improvements require a separate PR and fixing just the logger name doesn't change the situation significantly.

SpecLad (approved):

Issue #8859, opened 2024-12-23
Why does the position of an already marked annotation box change?


jaffe-fly:
### Actions before raising this issue

- [X] I searched the existing issues and did not find anything similar.
- [X] I read/searched [the docs](https://docs.cvat.ai/docs/)

### Steps to Reproduce

1,  fix a labeled  object box  use rectangle shape ,For example, changing the size of the box
2, Press the `F`  key and then press `D` key back to image
3, The box changes to its original size instead of maintaining the changed size

why ？and how to fix it

klakhov:
Hi, does this issue occur consistently? 
I couldn’t reproduce it on my end. Could there be additional steps in the reproduction scenario? 

Also, could you let me know which version of CVAT you are using?

jaffe-fly:
```
Server version: 2.22.0
Core version: 15.2.0
Canvas version: 2.20.9
UI version: 1.66.1
```
if you have a complete a task , you can re open it,  change one of images anno box,  sometimes your changes will be preserved, but sometimes ，When you switch images and go back to the previous image, it is not saved according to your changes.

when  anno a small datasets,  no such problem

my  datasets about 6-10G  2560*1440

jaffe-fly:
i found only when `event` request appear， the updated box will save
![image](https://github.com/user-attachments/assets/a7ff78c8-6436-42ec-98fd-0554cde06e8a)

But this `event` request does not appear immediately after you change the box, but waits for several seconds after update anno box. Why?

klakhov:
Hi @jaffe-fly,

I’m still trying to reproduce the error but haven’t had any luck so far.

I don’t think it’s related to the events request, as it only sends analytics data and doesn’t interact with annotations. Have you figured out a consistent way to reproduce the issue? Does it occur after resizing or dragging an object? Are there any errors in your browser console?

Thank you for helping with the investigation—any additional information would be greatly appreciated!

jaffe-fly:
![image](https://github.com/user-attachments/assets/1cab793b-9046-4738-9178-1430004bcd47)

jaffe-fly:
> Hi @jaffe-fly,
> 
> I’m still trying to reproduce the error but haven’t had any luck so far.
> 
> I don’t think it’s related to the events request, as it only sends analytics data and doesn’t interact with annotations. Have you figured out a consistent way to reproduce the issue? Does it occur after resizing or dragging an object? Are there any errors in your browser console?
> 
[...]

PR #8966, opened 2025-01-20
Fix: Cannot read properties of undefined (reading 'startPoints')

klakhov:
### Motivation and context

Related: #8859
One of the possible reasons to the issue is `Cannot read properties of undefined (reading 'startPoints')` error. For some unknown reason `handler` object inside of svg js event becomes undefined making an error in dragend hander. This leads to not dispatching `canvas.dragend` event so new object state coords are not saved. When the error occurs, object is visually moved but if we change frame and go back the annotation will be on the previous place.

annotaion-jump

In this fix, we will use `shape` as source of start point coordinates

### Checklist

### License


bsekachev (approved):
LGTM

bsekachev:
Just please add changelog entry

PR #8969, opened 2025-01-20
Refactor IAM id access

zhiltsov-max:
The existing object id access pattern for foreign keys produces unnecessary dependencies on the whole foreign object. It can be optimized to using just the object id, which is stored in the same db row in most cases (except m2m fields). This change doesn't immediately lead to optimized DB requests, but it opens optimization opportunities in the future.

- Optimized object id access for foreign key model fields in IAM checks
- Removed 1 extra DB request in task read serializer

### Checklist

### License


cvat/apps/engine/permissions.py
```diff
@@ -456,7 +462,7 @@ def create(cls, request, view, obj, iam_context):
     def create_scope_view(cls, request, task: Union[int, Task], iam_context=None):
         if isinstance(task, int):
             try:
-                task = Task.objects.get(id=task)
+                task = Task.objects.select_related("organization").get(id=task)
```
SpecLad, line 465: Could you explain this? It seems to go against the purpose of this PR, as it adds a _new_ join.

cvat/apps/engine/permissions.py
```diff
@@ -750,53 +753,53 @@ def get_scopes(request, view, obj):
     def get_resource(self):
         data = None
         if self.obj:
+            organization_id = None
```
SpecLad, line 756: This assignment is unnecessary, since both branches of the `if` below assign a new value.

cvat/apps/engine/serializers.py
```diff
@@ -2465,10 +2465,14 @@ class Meta:
...
     def to_representation(self, instance):
         response = super().to_representation(instance)
-        task_subsets = {task.subset for task in instance.tasks.all()}
-        task_subsets.discard('')
+
+        task_subsets = {task.subset for task in instance.tasks.all() if task.subset}
+        task_dimension = next(
+            (task.dimension for task in instance.tasks.all() if task.dimension),
+            None
+        )
```
SpecLad, line 2473: Seems like this calculation should be possible with SQL. Any reason to do it on the backend?

(Also, this doesn't seem related to the rest of the PR...)

cvat/apps/webhooks/permissions.py
```diff
@@ -108,9 +112,9 @@ def get_resource(self):
                     {
                         "owner": (
                             {
-                                "id": project.owner.id,
+                                "id": project.owner_id,
                             }
-                            if project.owner
+                            if project.owner_id
```
SpecLad, line 117:
```suggestion
                            if project.owner_id is not None
```

cvat/apps/engine/permissions.py
```diff
@@ -196,15 +200,17 @@ def get_resource(self):
         elif self.obj:
             data = {
                 'id': self.obj.id,
-                'owner': { 'id': getattr(self.obj.owner, 'id', None) },
+                'owner': { 'id': self.obj.owner_id },
                 'organization': {
-                    'id': self.obj.organization.id
-                } if self.obj.organization else None
+                    'id': self.obj.organization_id
+                } if self.obj.organization_id else None
```
SpecLad, line 206: I would suggest to add a common function for this pattern, since it seems to be repeating a lot:

```python
def id_dict(id):
    if id is None: return None
    return {'id': id}
```

cvat/apps/engine/permissions.py
SpecLad: Did you mean `select_related("project")`?

cvat/apps/engine/serializers.py
zhiltsov-max: Sure, it can be done with `.annotate()`, but it will increase DB response size and processing time. Previously it was sending an extra DB request for `.first()`, but we already have all the data needed at this point.

SpecLad: > Previously it was sending an extra DB request for .first(), but we already have all the data needed at this point.

Ah, okay, that explains it.

cvat/apps/webhooks/permissions.py
zhiltsov-max: `_id`s are either positive numbers or `None`, this change doesn't seem to add much. Do you suggest that this check patter is updated everywhere?

SpecLad (approved):

PR #9008, opened 2025-01-29
[Helm] Fixed export when disableDistinctCachePerService is false (by default)

azhavoro:
This PR removes the `disableDistinctCachePerService` setting.
Previously this cache was used to cache chunks of data on storage, which is not relevant now.
Currently only events need to be stored before being sent to Vector, and this will always be cached for each service, so these settings are no longer needed.

### Checklist

### License


nmanovic:
@Gabriellgpc , it will, but we need to deploy the fix. Please give us some time. CVAT Online will work as expected already today.

Gabriellgpc:
Got it. Thank you for the reply @nmanovic

PR #9010, opened 2025-01-29
Fix honeypot changes in tasks with cloud data

zhiltsov-max:
### Motivation and context

- Added missing updates of manifests on honeypot changes in tasks with cloud storage data. This change affects chunks and backups

Note: if tasks contained invalid honeypots before the update, they will require an extra manual honeypot update to become correct. There is no automatic fix for this issue.

### Checklist

### License

### DIFF
--- a/cvat/apps/engine/serializers.py
+++ b/cvat/apps/engine/serializers.py
@@ -53,6 +53,7 @@ from cvat.apps.engine.utils import (
     reverse,
     take_by,
 )
+from utils.dataset_manifest import ImageManifestManager
 
 slogger = ServerLogManager(__name__)
 
@@ -1029,6 +1030,8 @@ class JobValidationLayoutWriteSerializer(serializers.Serializer):
                 f"Honeypots cannot exist in {models.JobType.GROUND_TRUTH} jobs"
             )
 
+        assert not hasattr(db_data, 'video')
+
         frame_step = db_data.get_frame_step()
 
         def _to_rel_frame(abs_frame: int) -> int:
@@ -1166,6 +1169,12 @@ class JobValidationLayoutWriteSerializer(serializers.Serializer):
 
                 # Remove annotations on changed validation frames
                 self._clear_annotations_on_frames(db_segment, updated_honeypots)
+
+                # Update manifest
+                manifest_path = db_data.get_manifest_path()
+                if os.path.isfile(manifest_path):
+                    manifest = ImageManifestManager(manifest_path)
+                    manifest.reorder([db_frame.path for db_frame in db_frames.values()])
 
             # Update chunks
             job_frame_provider = JobFrameProvider(db_job)
@@ -1463,6 +1472,8 @@ class TaskValidationLayoutWriteSerializer(serializers.Serializer):
         if not frame_selection_method:
             return instance
 
+        assert not hasattr(instance.data, 'video')
+
         # Populate the prefetch cache for required objects
         prefetch_related_objects([instance],
             Prefetch('data__images', queryset=models.Image.objects.order_by('frame')),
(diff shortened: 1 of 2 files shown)


cvat/apps/engine/serializers.py
```diff
@@ -1179,6 +1182,12 @@ def _to_abs_frame(rel_frame: int) -> int:
                 # Remove annotations on changed validation frames
                 self._clear_annotations_on_frames(db_segment, updated_honeypots)
 
+                # Update manifest
+                manifest_path = db_data.get_manifest_path()
+                if os.path.isfile(manifest_path):
+                    manifest = ImageManifestManager(manifest_path)
+                    manifest.reorder([db_frame.path for db_frame in db_frames.values()])
```
Marishka17, line 1189: Changing the order of the whole task manifest when updating honeypots for a single job is probably not a good idea.

tests/python/rest_api/test_tasks.py
```diff
@@ -3024,6 +3042,51 @@ def fxt_uploaded_images_task_with_honeypots_and_changed_real_frames(
...
+            request.addfinalizer(partial(s3_client.remove_file, filename=image.name))
+
+        server_files = [f.name for f in image_files]
+
+        for image in image_files:
+            image.seek(0)
+
+        yield from self._images_task_with_honeypots_and_changed_real_frames_base(
+            request,
+            image_files=image_files,
```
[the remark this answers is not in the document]
zhiltsov-max: Server files are passed to make this a separate fixture, which can be marked and which can manage these files itself. 

Images are passed to be able to compare their pixels further in tests and to avoid adding logic to retrieve these images from the cloud in the task creation function. Another option would be to change `task_spec._get_frame` like `_images_task_with_honeypots_and_changed_real_frames_base` does.

cvat/apps/engine/serializers.py
zhiltsov-max: Do you see a better solution? We still need to pass all the images, because the manifest is a single file.

tests/python/rest_api/test_tasks.py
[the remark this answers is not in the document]
Marishka17: >  I didn't want to put this CS management logic into the base function.

I did not assume that, my comment was that you "duplicate" a bit passing data to `_image_task_fxt_base`. + There is no check that passed `server_files` corresponds to the `imgae_data` (because now it's possible to pass just `server_files` and generate `image_files` in the `_image_task_fxt_base`)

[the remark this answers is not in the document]
zhiltsov-max: Ok, it makes sense to add a check that if `server_files` is used, images should also be provided.

cvat/apps/engine/serializers.py
Marishka17: > Which problem do you see in the current implementation?

I'm not sure whether large tasks will cause problems. This action can probably take more time than 60 seconds. + Perhaps we need to be sure that the other manifest content corresponds to `db_frames`, otherwise we may not be able to understand why some frames contain unmatched annotations. (I know that such a change will fix problems with unmatching frames from manifest and frame sizes from db, but just logging will be useful, I guess). Up to you.

zhiltsov-max: > I'm not sure whether large tasks will cause problems. This action can probably take more time than 60 seconds.

Why it can take a big amount of time? From what I've seen in the implementation, it should be a file construction and writing, and the file itself is at most several kb in size. Am I missing something?

> Perhaps we need to be sure that the other manifest content corresponds to `db_frames`
I suppose there are 2 possible reasons for this:
- it's an old task, created before the update and which had some layout updates
- the file was updated in another thread

The first case tasks should just become correct after the next honeypot update.

The second case is described in the endpoint description. There are several places in this function where things can break in such situation. A lock or some other protection mechanism is needed, but we aren't using them for endpoint access currently.

Marishka17: okay

Marishka17 (approved):

PR #9031, opened 2025-01-30
Test for interpolation issues after deleting keyframes

archibald1418:
Regression test for 2 bugs that were fixed in #8951 and #8952 
The fix dealt with incorrect frame interpolation when deleting keyframes
Also, opening the task afterwards could crash with application error

### How has this been tested?

This consists of two test cases reflecting two different bugs + `before()` and `beforeEach()` hooks which unifies conditions before both cases.

##### Before all tests:
- Login
- Create a zip archive with a dataset consisting of 5 frames
- Create task from archive using UI 
(NB: `headless..` commands only work with `mounted_file_share` which being regularly cleared in Helm CI)
- Open task job and save `jobID` into a variable
- Send headless request  to cvat-core's `JobsApI` to create a rectangle track with 3 keyframes: 0, 2 and 4. After this, the rectangle moves clockwise along the image's perimeter (topLeft -> topCenter -> topRight -> centerRight -> bottomRight)

#### Before each test
- Restore all frames using headless `PATCH` request to send `deleted_frames: []` to `/data/meta`. This recovers frame deletions that were done under test. Wait for response
- Reload the page to hydrate annotations that were added previously. Wait for `GET` to `/data/meta` which will load the created rectangle track onto the `#cvat_canvas_background`. 
- Save the job
- Go to first frame to start the test

### Case 1
This checks that the job can be opened without any crashes or error notification messages after deleting keyframes and saving the job. Interpolation is terminated and the tracked rectangle should not exist in the DOM

- Flick through all frames and click 'Delete' on frames 0, 2, 4 (keyframes)
- Save, reload the job. Error should not appear.
- Object should not exist. Error notifications should not exist

### Case 2
- Go to frame 2
- Delete frame 2. Confirm that frame 3 is open and that previous frame is frame 1
- Remember shape positions on frames 1 and 3 (=interpolated positions)
- Save the job, reload the job
- Validate that the shapes are on the same positions as before the reload

### Checklist

### License


.vscode/launch.json
```diff
@@ -152,7 +152,7 @@
                 "run",
                 "--headless",
                 "--browser",
-                "chrome"
+                "chrome",
```
bsekachev, line 155: Trailing commas are not a part of JSON spec
Apparently vscode supports them, however it would be more correct to remove them, rather then add.

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
```diff
@@ -0,0 +1,179 @@
+// Copyright (C) 2025 CVAT.ai Corporation
```
bsekachev, line 1: We do not add year anymore

tests/cypress/support/commands.js
```diff
@@ -348,6 +348,32 @@ Cypress.Commands.add('headlessCreateObjects', (objects, jobID) => {
...
+    });
+});
+
+Cypress.Commands.add('headlessGetJobMeta', (jobID) => {
+    cy.window().then(async ($win) => {
+        await $win.cvat.server.request(`/api/jobs/${jobID}/data/meta`, {
+            method: 'GET',
+        });
+    });
+});
```
bsekachev, line 375: Where do you use this command?

tests/cypress/support/commands.js
```diff
@@ -348,6 +348,32 @@ Cypress.Commands.add('headlessCreateObjects', (objects, jobID) => {
...
 });
 
+Cypress.Commands.add('headlessRemoveAnnotations', (jobID) => {
+    cy.window().then(async ($win) => {
+        const job = (await $win.cvat.jobs.get({ jobID }))[0];
+        await job.annotations.clear({ reload: true });
+        await job.annotations.save();
+        return cy.wrap();
+    });
+});
```
bsekachev, line 358: Where do you use this command?

tests/cypress/support/commands.js
```diff
@@ -1778,3 +1826,11 @@ Cypress.Commands.add('clickSaveAnnotationView', () => {
     cy.get('button').contains('Save').click();
     cy.get('button').contains('Save').trigger('mouseout');
 });
+
+Cypress.Commands.add('clickUndo', () => {
+    cy.contains('.cvat-annotation-header-button', 'Undo').click();
+});
```
bsekachev, line 1832: Unused command

tests/cypress/support/commands.js
```diff
@@ -1778,3 +1826,11 @@ Cypress.Commands.add('clickSaveAnnotationView', () => {
...
+
+Cypress.Commands.add('clickFirstFrame', () => {
+    cy.get('.cvat-player-first-button').click();
+});
```
bsekachev, line 1836: I do not think one-line commands make sense. 

It may be considered as encapsulation approach, what is basically might be a good idea, however if you add it here, we must replace all corresponding occurrences and implement the same for other buttons, what is definitely not related to this pull request.

tests/cypress/support/commands.js
```diff
@@ -1767,6 +1793,28 @@ Cypress.Commands.overwrite('reload', (orig, options) => {
...
+    cy.get('[placeholder="Label name"]').type(labelName);
+    cy.contains('button', 'Continue').click();
+    // Archive name is relative to cypress/fixtures
+    cy.get('input[type="file"]').attachFile(archiveName, { subjectType: 'drag-n-drop' });
+    cy.contains('button', 'Submit & Continue').click();
+    cy.get('.cvat-notification-create-task-success').should('exist').within(() => {
+        cy.get('.anticon-close').click();
+    });
+    cy.get('.cvat-notification-create-task-fail').should('not.exist');
+});
```
bsekachev, line 1816: I do not think we need one more command to create a task.
Why existing one can't be used? Why can't we modify/refactor it?

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
```diff
@@ -0,0 +1,179 @@
...
+const posY = 50;
+const color = 'white';
+const imageFileName = `image_${issueId}`;
+const archiveName = `${imageFileName}.zip`;
+const archivePath = `cypress/fixtures/${archiveName}`;
+const imagesFolder = `cypress/fixtures/${imageFileName}`;
+const dirToArchive = imagesFolder;
+
+const [fX, fY] = [30, 30];
+const [w, h] = [34, 23];
```
bsekachev, line 23: Do you really need this declaration? What is wrong with `xtl`, `ytl`, `xbr`, `ybr`, commonly used across CVAT code? 

- x top left
- y top left
- x bottom right
- y bottom right

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
```diff
@@ -0,0 +1,179 @@
...
+const width = 400;
+const height = 400;
+const posX = 50;
+const posY = 50;
+const color = 'white';
+const imageFileName = `image_${issueId}`;
+const archiveName = `${imageFileName}.zip`;
+const archivePath = `cypress/fixtures/${archiveName}`;
+const imagesFolder = `cypress/fixtures/${imageFileName}`;
+const dirToArchive = imagesFolder;
```
bsekachev, line 20: Is there any reason to duplicate variable with another name?

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
```diff
@@ -0,0 +1,179 @@
+// Copyright (C) 2025 CVAT.ai Corporation
+//
+// SPDX-License-Identifier: MIT
+
+/// <reference types="cypress" />
+
+const taskName = '5frames';
+const labelName = 'label';
+const issueId = 'xxxx';
```
bsekachev, line 9: xxxx?

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
```diff
@@ -0,0 +1,179 @@
...
+const color = 'white';
+const imageFileName = `image_${issueId}`;
+const archiveName = `${imageFileName}.zip`;
+const archivePath = `cypress/fixtures/${archiveName}`;
+const imagesFolder = `cypress/fixtures/${imageFileName}`;
+const dirToArchive = imagesFolder;
+
+const [fX, fY] = [30, 30];
+const [w, h] = [34, 23];
+const createRectangleShape2Points = {
```
bsekachev, line 24: 2Points / 4Points make sense when you create using UI elements. 
In this case everything created headlessly.

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
```diff
@@ -0,0 +1,179 @@
...
+                ], 0, labelName);
+                cy.headlessCreateObjects([track], jobID);
+            });
+        });
+
+        beforeEach(() => {
+            // Restore all frames by patching deleted_frames to []
+            cy.intercept('PATCH', '/api/jobs/**/data/meta**').as('patchMeta');
+            cy.headlessRestoreAllFrames(jobID);
+            cy.wait('@patchMeta');
```
bsekachev, line 124: I believe interception should be inside `headlessRestoreAllFrames`.
Otherwise external code knows implementation details of this command.

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
```diff
@@ -0,0 +1,179 @@
...
+                    wrap(shape0, 0),
+                    wrap(shape1, 2),
+                    wrap(shape2, 4),
+                ], 0, labelName);
+                cy.headlessCreateObjects([track], jobID);
+            });
+        });
+
+        beforeEach(() => {
+            // Restore all frames by patching deleted_frames to []
```
bsekachev, line 121: I think it is clear from the command name, is not it?

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
```diff
@@ -0,0 +1,179 @@
...
+            cy.intercept('PATCH', '/api/jobs/**/data/meta**').as('patchMeta');
+            cy.headlessRestoreAllFrames(jobID);
+            cy.wait('@patchMeta');
+
+            // Get job meta updates from the server and reload page to bring changes to UI
+            cy.intercept('GET', '/api/jobs/**/data/meta**').as('getMeta');
+            cy.reload();
+            cy.wait('@getMeta');
```
bsekachev, line 129: What will be changed if we remove this interception?

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
```diff
@@ -0,0 +1,179 @@
...
+                // Remove all annotations and draw a track rect
+                const wrap = (shape, frame) => shapeToPayload(shape, frame, 'rectangle');
+                const shape0 = createRectangleShape2Points;
+                const shape1 = translateShape(shape0, delta, 'x');
+                const shape2 = translateShape(shape1, delta, 'y'); // TODO: fix coords, rect flies away
+                const track = makeTrack([
+                    wrap(shape0, 0),
+                    wrap(shape1, 2),
+                    wrap(shape2, 4),
+                ], 0, labelName);
```
bsekachev, line 115: Please, just change initial object structure as `headlessCreateObjects` expects. Making this thing dynamic is not obvious and looks too complex.

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
```diff
@@ -0,0 +1,179 @@
...
+    }
+    function compareShape(num) {
+        const { x, y } = stash[num];
+        cy.get('.cvat_canvas_shape').invoke('attr', 'x').then((xVal) => {
+            expect(parseFloat(xVal)).to.be.closeTo(x, precision);
+        });
+        cy.get('.cvat_canvas_shape').invoke('attr', 'y').then((yVal) => {
+            expect(parseFloat(yVal)).to.be.closeTo(y, precision);
+        });
+    }
```
bsekachev, line 87: Please make these functions pure. 
Thus, `storeShape` should be probably renamed to `readShapeCoordinates` or something like this and return { x, y }.

`compareShape` should be probably renamed to `validateShapeCoodrinates` or something like this and  accept { x, y } as argument, implementing comparison.

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
```diff
@@ -0,0 +1,179 @@
...
+
+            cy.saveJob();
+            cy.clickFirstFrame();
+        });
+        it('Delete interpolated frames 0, 2, 4. Error should not appear', () => {
+            cy.on('uncaught:exception', (err) => {
+                const expectedMsg = 'No one left position or right position was found. Interpolation impossible. Client ID:';
+                expect(err.message).to.include(expectedMsg); // Exclude familiar error
+                throw err;
+            });
```
bsekachev, line 139: Error should not appear as you written in the test description. So, why do you need this code?

tests/cypress/support/commands.js
archibald1418: Tasks can't be created with a zip archive headlessly because headless functions only work with mounted_file_share which gets copied during build in helm CI, not mounted, freshly created tasks can be lost. So it's either adding this command or add changes into helm CI

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
```diff
@@ -0,0 +1,179 @@
...
+
+            // Delete frames 0, 2, 4. Watch out for errors
+            cy.clickFirstFrame();
+            cy.checkFrameNum(0);
```
bsekachev, line 143: There is command `goCheckFrameNum`  doing both things

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
```diff
@@ -0,0 +1,179 @@
...
+            cy.checkFrameNum(3);
+            cy.goToNextFrame(4);
+            cy.clickDeleteFrameAnnotationView();
+
+            // There should be no objects on the deleted frame
+            cy.get('.cvat_canvas_shape').should('not.exist');
+            cy.clickSaveAnnotationView();
+
+            // This might pop up after deleting or saving (on firefox e.g.)
+            cy.get('.ant-notification-notice-error').should('not.exist');
```
bsekachev, line 157: May you explain why you added it here?

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
archibald1418: Makes sense

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
```diff
@@ -0,0 +1,179 @@
...
+
+            // Reopening a task with bad metadata might throw an exception that we can catch
+            cy.goToTaskList();
+            cy.openTaskJob(taskName);
```
bsekachev, line 161: cy.reload() and corresponding interceptor (e.g. annotations receiving) to make sure that the page was updated?

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
```diff
@@ -0,0 +1,179 @@
...
+            cy.get('.ant-notification-notice-error').should('not.exist');
+        });
+        it('Change track positions on frames 2 and 4. Delete frame. Confirm same shape positions', () => {
+            cy.goToNextFrame(1);
+            cy.goToNextFrame(2);
```
bsekachev, line 166: `goCheckFrameNumber(2)`?

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
archibald1418: This ensures that getMeta was sent and that we get the correct UI state. If this is not satisfied, the test should fail now. In theory this request can crash with network error (which was being checked in `issue_8872_delete_frame.js`, for example)

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
```diff
@@ -0,0 +1,179 @@
...
+            cy.clickDeleteFrameAnnotationView();
+            cy.clickSaveAnnotationView();
+            cy.checkFrameNum(3);
```
bsekachev, line 169: it would be more logical to swipe this two commands

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
archibald1418: Makes sense, applying

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
```diff
@@ -0,0 +1,179 @@
...
+            storeShape(3);
+            cy.goToPreviousFrame(1);
+            storeShape(1);
+            cy.reload({ forceReload: false }).then(() => {
```
bsekachev, line 173: Why force reload needed after saving the job?

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
archibald1418: It is, actually, applying

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
archibald1418: applied

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
archibald1418: No reason indeed, applied

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
bsekachev: Successful request is not equal the correct UI state. So, if you want to check this state, it should be implemented somehow else. 

There may be different issues with the code, what is much closer to real than failing with network error in test environment.

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
```diff
@@ -0,0 +1,179 @@
...
+                cy.goToNextFrame(1).then(() => compareShape(1));
+                cy.goToNextFrame(3).then(() => compareShape(3));
```
bsekachev, line 175: `cy.reload().then({ instructions })` does not seem reliable to me as `instructions` may be executed before the page is actually reloaded. But it will be simpler to explain on the call, if you are interested

tests/cypress/support/commands.js
bsekachev: I did not mentioned headless functions here. 

We already have `cy.createAnnotationTask` which is not headless.
I am asking why we need one more non-headless command

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
archibald1418: Makes sense, any error should fail the test, applying

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
archibald1418: I added it here because on firefox the error can pop up even before the reload

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
bsekachev: Okay, I checked additionally. 
If cy.reload() actually expects for `load` event to be triggerred, it should work.

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
archibald1418: I guess no need to reload cache if it doesn't interfere with sending `GET /data/meta` request

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
bsekachev: > the error can pop up even before the reload

Which one?

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
archibald1418: Let's do it

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
bsekachev: 1. REST API requests are never cached
2. If, in theory, by mistake, they become cached, it is better to know about it if tests are failed

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
bsekachev: Can such a function return a promise?

bsekachev: > the best practice for this 

May you share a link?

archibald1418: Assigning return values:
https://docs.cypress.io/app/core-concepts/best-practices#Assigning-Return-Values

Workarounds that cypress proposes instead:
https://docs.cypress.io/app/core-concepts/variables-and-aliases

archibald1418: I actually found an example in our codebase where you use promises as a retval, I'll try that. There's still the issue of  too much nested code

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
archibald1418: Just checked on firefox, it's still a crash, so you're correct, no need to check notifications

tests/cypress/support/commands.js
archibald1418: Oh I got you, you're right, applying changes

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
```diff
@@ -0,0 +1,170 @@
...
+            cy.reload();
+
+            cy.saveJob();
+            cy.get('.cvat-player-first-button').click();
+        });
+        it('Delete interpolated frames 0, 2, 4. Error should not appear', () => {
+            cy.on('uncaught:exception', (err) => {
+                console.error(err);
+                throw err;
+            });
```
bsekachev, line 128: Why do you need this fragment?

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
```diff
@@ -0,0 +1,170 @@
...
+                const shape0 = rect;
+                const shape1 = translateShape(shape0, delta, 'x');
+                const shape2 = translateShape(shape1, delta, 'y');
+                const track = {
+                    shapes: [
+                        {
+                            frame: 0,
+                            type: 'rectangle',
+                            points: Object.values(shape0),
+                            // ECMAScript guarantees chronological order of keys
```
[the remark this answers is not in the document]
archibald1418: I mean that Object.values will preserve the order of coordinates. I can delete it if it's redundant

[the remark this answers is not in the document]
bsekachev: It would be simpler if we just use Array instead of dictionary, to store points. 
We may trust esma specs, however understanding of the code requires knowledge of the spec from the reader as this is not really obvious thing.

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
archibald1418: Solved with local variables

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
```diff
@@ -0,0 +1,170 @@
...
+const archivePath = `cypress/fixtures/${archiveName}`;
+const imagesFolder = `cypress/fixtures/${imageFileName}`;
+
+const [xtl, ytl, xbr, ybr] = [
+    30, 30,
+    30 + 34, 30 + 23,
+];
+const rect = {
+    xtl, ytl, xbr, ybr,
+};
```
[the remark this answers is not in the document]
archibald1418: applied

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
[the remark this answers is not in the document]
archibald1418: Makes sense, applying

bsekachev:
Please, update the code first. Then, resolve corresponding comments.

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
```diff
@@ -0,0 +1,169 @@
...
+    }
+    if (axis === 'y') {
+        return [
+            points[0],
+            points[1] + delta,
+            points[2],
+            points[3] + delta,
+        ];
+    }
+    return null;
```
bsekachev, line 50:
```suggestion
    return points;
```

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
```diff
@@ -0,0 +1,169 @@
...
+        beforeEach(() => {
+            cy.headlessRestoreAllFrames(jobID);
+
+            // Get job meta updates from the server and reload page to bring changes to UI
+            cy.reload();
+
+            cy.saveJob();
+            cy.get('.cvat-player-first-button').click();
+        });
+        it('Delete interpolated frames 0, 2, 4. Error should not appear', () => {
```
bsekachev, line 128: Please add one space between `it`

tests/cypress/e2e/issues_prs2/issue_8952_interpolation_impossible.js
```diff
@@ -0,0 +1,169 @@
...
+}
+
+context('Create any track, check if track works correctly after deleting some frames', () => {
+    function readShapeCoords() {
+        return cy.get('.cvat_canvas_shape').then(($shape) => ({
+            x: +$shape.attr('x'),
+            y: +$shape.attr('y'),
+        }));
+    }
+    function validateShapeCoords({ x, y }) {
```
bsekachev, line 60: Please, add one space between functions declaration

tests/cypress/support/commands.js
```diff
@@ -348,6 +348,17 @@ Cypress.Commands.add('headlessCreateObjects', (objects, jobID) => {
     });
 });
 
+Cypress.Commands.add('headlessRestoreAllFrames', (jobID) => {
+    cy.intercept('PATCH', '/api/jobs/**/data/meta**').as('patchMeta');
```
bsekachev, line 352: Inteception may be more specific here:

```
cy.intercept('PATCH', `/api/jobs/${jobID}/data/meta**`)
```

bsekachev (approved):

tests/cypress/support/commands.js
archibald1418: applied

PR #9047, opened 2025-02-03
Refactor bulk_create calls

zhiltsov-max:
### Motivation and context

- Added default batch size in bulk create calls
- Removed SQLite workaround in `bulk_create()` calls

### Checklist

### License


SpecLad:
What's the purpose of this?

zhiltsov-max:
As the description states - use chunk size by default. It's often unintentionally forgotten.

SpecLad:
> As the description states - use chunk size by default. It's often unintentionally forgotten.

Sure, but why do we want this?

cvat/apps/engine/model_utils.py
```diff
@@ -18,3 +22,86 @@ class Undefined:
...
+
+
+def bulk_create(
+    db_model: type[_ModelT],
+    objs: Iterable[_ModelT],
+    batch_size: int | None = ...,
+    ignore_conflicts: bool = ...,
+    update_conflicts: bool | None = ...,
+    update_fields: Sequence[str] | None = ...,
+    unique_fields: Sequence[str] | None = ...,
```
SpecLad, line 37: Can't this be replaced with `**kwargs`?

cvat/apps/engine/model_utils.py
```diff
@@ -18,3 +22,86 @@ class Undefined:
...
+    if not objs:
+        return []
+
+    flt_param = flt_param or {}
+
+    if flt_param:
+        if "postgresql" in settings.DATABASES["default"]["ENGINE"]:
+            return db_model.objects.bulk_create(objs, batch_size=batch_size, **kwargs)
+        else:
+            # imitate RETURNING
```
SpecLad, line 65: I wonder if this workaround can be removed now? The only relevant database that this would be used for is SQLite; according to the Django docs, SQLite 3.35 supports returning the primary key; and Ubuntu 22.04 already has SQLite 3.37.

cvat/apps/engine/views.py
```diff
@@ -1153,7 +1153,7 @@ def _append_upload_info_entries(self, client_files: list[dict[str, Any]]):
         task_data.client_files.bulk_create([
             ClientFile(file=self._prepare_upload_info_entry(cf['file'].name), data=task_data)
             for cf in client_files
-        ])
+        ], batch_size=settings.DEFAULT_DB_BULK_CREATE_BATCH_SIZE)
```
SpecLad, line 1156: Why not use the helper function here?

cvat/apps/engine/model_utils.py
SpecLad, line 1: Consider moving this to `cvat/utils`. Doesn't seem like it has much to do with `engine`.

cvat/apps/engine/default_settings.py
```diff
@@ -95,3 +95,5 @@
 MAX_CONSENSUS_REPLICAS = int(os.getenv("CVAT_MAX_CONSENSUS_REPLICAS", 11))
 if MAX_CONSENSUS_REPLICAS < 1:
     raise ImproperlyConfigured(f"MAX_CONSENSUS_REPLICAS must be >= 1, got {MAX_CONSENSUS_REPLICAS}")
+
+DEFAULT_DB_BULK_CREATE_BATCH_SIZE = int(os.getenv("CVAT_DEFAULT_DB_BULK_CREATE_BATCH_SIZE", 1000))
```
SpecLad, line 99: Does this really need to be a setting, and an environment variable? Under what circumstances would one want to adjust this?

cvat/apps/engine/model_utils.py
```diff
@@ -18,3 +22,86 @@ class Undefined:
...
+
+def bulk_create(
+    db_model: type[_ModelT],
+    objs: Iterable[_ModelT],
+    batch_size: int | None = ...,
+    ignore_conflicts: bool = ...,
+    update_conflicts: bool | None = ...,
+    update_fields: Sequence[str] | None = ...,
+    unique_fields: Sequence[str] | None = ...,
+    *,
```
SpecLad, line 38: I would suggest to move this above `batch_size`.

cvat/apps/engine/model_utils.py
zhiltsov-max: It can, I just wanted it to be a "drop-in" replacement with the same signature. It can't be a pure drop-in, because now we require the first model argument, but this design offers good enough compatibility.

cvat/apps/engine/model_utils.py
zhiltsov-max: Yes, probably.

cvat/apps/engine/model_utils.py
zhiltsov-max: I used API from the [documentation](https://docs.djangoproject.com/en/5.1/ref/models/querysets/#bulk-create) (and from the VS code django stub). Do you think the argument order should be changed?

`bulk_create(objs, batch_size=None, ignore_conflicts=False, update_conflicts=False, update_fields=None, unique_fields=None)`

cvat/apps/engine/default_settings.py
zhiltsov-max: Basically, it's just a named constant. I can't say why this may require changes so far, so I'm open to removing configurability. Typically, it's better have such performance constants configurable.

cvat/apps/engine/views.py
zhiltsov-max: Updated

cvat/apps/engine/model_utils.py
SpecLad: By "this", I meant the asterisk. Basically, I'm suggesting to make all arguments with defaults kw-only.

cvat/apps/engine/model_utils.py
SpecLad: Okay, but then why use the ellipses? Seems like you could copy the default values from the original signature as well.

cvat/apps/engine/default_settings.py
SpecLad: Alright, I suppose we can keep it.

cvat/apps/engine/model_utils.py
zhiltsov-max: I used annotations from https://github.com/sbdchd/django-types/blob/062b1bfe5313b7e9f1a0a165fa5bf549e6b4addd/django-stubs/db/models/query.pyi#L52 that are [used by VS Code](https://github.com/microsoft/pylance-release/issues/4562#issuecomment-1615049922), which is our default development environment. The [more official django-stubs repo](https://github.com/typeddjango/django-stubs/blob/008c6e6b2534d9cf444a9650330611ca2248a4ee/django-stubs/db/models/query.pyi#L78) has the same annotations. Ellipses are supposed to mean "unchanged", and this is enforced by the implementation. If the official defaults ever change, we will not have to do any changes in this function.

cvat/apps/engine/model_utils.py
zhiltsov-max: We can make the parameters kw-only, but in the original function they are positional.

cvat/apps/engine/model_utils.py
SpecLad: > and what most developers will typically see in the editor help.

I mean... would you rather see ellipses or actual values?

> The [more official django-stubs repo](https://github.com/typeddjango/django-stubs/blob/008c6e6b2534d9cf444a9650330611ca2248a4ee/django-stubs/db/models/query.pyi#L78) has the same annotations.

No it doesn't, check again. 🙂 

> Ellipses are supposed to mean "unchanged", and this is enforced by the implementation. If the official defaults ever change, we will not have to do any changes in this function.

I think the official defaults changing would imply a backwards compatibility break, so wouldn't it actually be useful for us to have hardcoded them?

cvat/apps/engine/model_utils.py
zhiltsov-max: Currently, it uses the setting owned by `engine`. The setting can be moved just to `base.py`, but it makes ownership unknown and requires it to always be specified. I consider `engine` a base component establishing common logic for other components, so it defines a lot of utilities used everywhere.

cvat/apps/engine/model_utils.py
zhiltsov-max: > No it doesn't, check again. 🙂

Ok, probably I missed this.

> I think the official defaults changing would imply a backwards compatibility break, so wouldn't it actually be useful for us to have hardcoded them?

I think it doesn't matter. Time spent on discussing this question is probably bigger than for changing the code in such case.

It seems Django is quite confident about the values, so I copied them from the "official" stub.

cvat/apps/engine/model_utils.py
SpecLad: Okay, works for me.

SpecLad (approved):

PR #9050, opened 2025-02-04
Use `any` as the default function label type

SpecLad:
### Motivation and context

Currently, the default is `unknown`, but allowing `unknown` as a function label type value creates some complications.

* The sets of allowed label types of a task and of a function are not the same, so we have to use different enums to represent them.

* `any` and `unknown` have similar meanings, and it's not clear what the difference between them is unless you search the code (AFAIK it's not documented anywhere).

The sole benefit of having a separate `unknown` label type seems to be that function labels of type `any` are only allowed to be mapped to task labels of type `any`, while function labels of type `unknown` can be mapped to task labels of any type. But it doesn't seem like a useful distinction. Functions with both `unknown` and `any` label types can produce any shape types, so if we allow one of them, it makes sense to also allow the other.

In addition, functions that can produce _multiple_ shape types for a single label are probably going to be rare, so I think it's reasonable to assume that a function with an `any`-typed label will still produce a single shape type and therefore can be mapped to a label with a specific type (although the user is responsible for ensuring that the types actually match).

To sum up, I don't think the additional complexity introduced by the `unknown` type is worth it. So let's remove it and use `any` instead.

This PR keeps some `unknown`-related logic in the UI, since there's some code in the Enterprise backend that also uses `unknown` as a label type. If this patch is accepted, I'll replace those uses with `any` too, then go back and remove the remaining logic here.

### How has this been tested?

Manual testing.

### Checklist

### License

### DIFF
--- a/changelog.d/20250204_183545_roman_unknown_any.md
+++ b/changelog.d/20250204_183545_roman_unknown_any.md
@@ -1,0 +1,5 @@
+### Changed
+
+- When invoking Nuclio functions, labels of type `any` can now be mapped to
+  labels of all types except `skeleton`
+  (<https://github.com/cvat-ai/cvat/pull/9050>)
--- a/cvat-ui/src/components/model-runner-modal/labels-mapper.tsx
+++ b/cvat-ui/src/components/model-runner-modal/labels-mapper.tsx
@@ -39,7 +39,7 @@ function labelsCompatible(modelLabel: LabelInterface, jobLabel: LabelInterface):
     const compatibleTypes = [[LabelType.MASK, LabelType.POLYGON]];
     return modelLabelType === jobLabelType ||
         (jobLabelType === 'any' && modelLabelType !== LabelType.SKELETON) ||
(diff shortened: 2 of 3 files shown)


bsekachev:
The idea behind "unknown" was that it is illogical that model returns "any" as it is always something specific. 
Did you decide to update it for unification?

SpecLad:
I'm against removing it, for a few reasons:

1. I'd like to require as little configuration as possible when creating functions, to lower the entry barrier. The label types in a model are not necessary for it to function, they're just extra information that the UI can use to help the user, so it makes sense for them to be optional.

   The auto-annotation function interface in the SDK also does not require label types, and I'd like to keep it that way.

2. I think it's premature to say that a model can _never_ output multiple shape types. The models we have in the repo don't do that, but we don't know what kind of model we (or our users) will implement next. To not support this seems like an unnecessary, artificial restriction.

bsekachev (approved):

PR #9052, opened 2025-02-05
Added ability to customize logo and title

klakhov:
### Motivation and context

This simple PR moves logo management to `/about` server endpoint. It would make easier to change it if needed by replacing `logo.svg` asset on the server. Additinally, there is a possibility to edit subtitle using settings

CVAT logo is `64x32`px in svg format. It is reccomended to use  the same format and proportions.
Customization process is the following: update `logo.svg` asset in `cvat/apps/engine/static` (if the logo has different name, update `LOGO_FILENAME` in `settings/base.py`), update `subtitle` in `ABOUT_INFO` to change text on the sign-in/sign-up pages. 

### Checklist

### License


cvat/apps/engine/views.py
```diff
@@ -189,6 +189,12 @@
 _DATA_UPDATED_DATE_HEADER_NAME = 'X-Updated-Date'
 _RETRY_AFTER_TIMEOUT = 10
 
+ICON_FILE = "assets/logo.svg"
+
+def get_logo() -> str:
```
nmanovic, line 194: Please check that we don't read it on every about request.

cvat/settings/base.py
```diff
@@ -769,3 +769,15 @@ class CVAT_QUEUES(Enum):
...
+ABOUT_INFO = {
+    "name": "Computer Vision Annotation Tool",
+    "title": "Open Data Annotation Platform",
+    "description": "CVAT is completely re-designed and re-implemented " +
+                "version of Video Annotation Tool from Irvine, California " +
+                "tool. It is free, online, interactive video and image annotation " +
+                "tool for computer vision. It is being used by our team to " +
+                "annotate million of objects with different properties. Many UI " +
+                "and UX decisions are based on feedbacks from professional data " +
+                "annotation team.",
```
bsekachev, line 782: I would suggest to remove title and put its content to description

cvat/apps/engine/views.py
```diff
@@ -189,6 +189,12 @@
...
+    with open(osp.join(osp.dirname(__file__), ICON_FILE)) as f:
+        return f.read()
```
bsekachev, line 196: I will propose to support .png files also
Thus it may be encoded to base64 on the server and displayed as is on client

bsekachev (commented):
In this implementation it is still not customizable, as was requested by the customer,  as icon is a part of docker image

cvat/apps/engine/views.py
```diff
@@ -210,15 +216,11 @@ def get_serializer(self, *args, **kwargs):
     def about(request):
         from cvat import __version__ as cvat_version
         about = {
-            "name": "Computer Vision Annotation Tool",
+            "name": settings.ABOUT_INFO["name"],
```
bsekachev, line 219: Will the response be cached by the browser?

cvat-ui/src/components/common/cvat-logo.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import { useSelector } from 'react-redux';
+
+function CVATLogo(): JSX.Element {
+    const SVGLogo = useSelector((state: any) => state.about.server.logo);
+
+    return (
+        <div className='cvat-logo-icon'>
+            <img
```
bsekachev, line 9: What if the server returns logo (let's say 1024x256px)? Will it be fit into the component?

cvat/apps/engine/views.py
```diff
@@ -210,15 +216,11 @@ def get_serializer(self, *args, **kwargs):
...
+            "description": settings.ABOUT_INFO["description"],
             "version": cvat_version,
-            "description": "CVAT is completely re-designed and re-implemented " +
-                "version of Video Annotation Tool from Irvine, California " +
-                "tool. It is free, online, interactive video and image annotation " +
-                "tool for computer vision. It is being used by our team to " +
-                "annotate million of objects with different properties. Many UI " +
-                "and UX decisions are based on feedbacks from professional data " +
-                "annotation team."
+            "logo": get_logo(),
```
SpecLad, line 223: Do we really need to include the whole logo into the API response?

I think it would be more sensible to include the URL of the logo, and stick the actual logo in the `static` directory. Then:

1. The API response won't be as bloated.
2. The usual HTTP caching mechanisms will apply, so the clients won't fetch the logo more often than necessary.
3. It'll be more efficient if the logo is in a non-text format (e.g. PNG), since no encoding will be necessary.

SpecLad:
> In this implementation it is still not customizable, as was requested by the customer, as icon is a part of docker image

That seems normal to me - where else would it be? I'm assuming that the customization process will involve mounting an alternate icon file into the container.

Did you have something else in mind?

cvat/apps/engine/views.py
```diff
@@ -210,15 +216,11 @@ def get_serializer(self, *args, **kwargs):
     def about(request):
         from cvat import __version__ as cvat_version
         about = {
-            "name": "Computer Vision Annotation Tool",
+            "name": settings.ABOUT_INFO["name"],
+            "title": settings.ABOUT_INFO["title"],
+            "description": settings.ABOUT_INFO["description"],
```
SpecLad, line 221: Why did you make the name and description customizable? I don't think that was in the customer's request. It's also not mentioned in the changelog.

cvat/settings/base.py
```diff
@@ -769,3 +769,15 @@ class CVAT_QUEUES(Enum):
 
 # Indicates the maximum number of days a file or directory is retained in the temporary directory
 TMP_FILE_OR_DIR_RETENTION_DAYS = 3
+
+ABOUT_INFO = {
+    "name": "Computer Vision Annotation Tool",
+    "title": "Open Data Annotation Platform",
```
SpecLad, line 775: I think calling this a "title" risks confusion with the "name" field. Maybe it should be called something like "subtitle", or "tagline"?

cvat/settings/base.py
klakhov: This text is still used in 'about' modal window that we have. I agree it may be outdated, but I dont think we want to remove it in terms of this PR.
If we are talking about customization of `/about` endpoint it makes sense so we can edit the description too.

cvat/apps/engine/views.py
```diff
@@ -189,6 +189,12 @@
 _DATA_UPDATED_DATE_HEADER_NAME = 'X-Updated-Date'
 _RETRY_AFTER_TIMEOUT = 10
 
+ICON_FILE = "assets/logo.svg"
```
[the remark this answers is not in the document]
klakhov: Moved to settings

cvat/apps/engine/views.py
klakhov: Implemented solution with static file, so different image formats are supported now

cvat/apps/engine/views.py
klakhov: Yes, static content will be cached

cvat-ui/src/components/common/cvat-logo.tsx
klakhov: Changed styles a bit. By default logo component is `64x32` so everything bigger will be scaled using `object-fit: contain` to maintain the size

cvat/apps/engine/views.py
klakhov: For me it makes sense if we want to customize the `/about` response via settings we can also customize those fields.

klakhov:
From my perspective, the logo customization process should be straightforward: place the new file in correct folder and configure a setting for its name if necessary. The same goes for text info in `/about`.
I dont think we want implement something like uploading new logo from UI

cvat/apps/engine/views.py
klakhov: Changed implementation to use static logo file as Roman suggested. Now it should be cached correctly

cvat-ui/src/components/signing-common/styles.scss
```diff
@@ -34,14 +34,18 @@ $social-google-background: #4286f5;
...
+
 .cvat-signing-header {
     background: transparent;
     position: fixed;
     padding: $grid-unit-size * 2 0 0 0;
     width: 100%;
 
     .cvat-logo-icon {
-        fill: white;
+        filter: brightness(0) invert(1);
```
bsekachev, line 48: As I understand it makes any logo on sign-in page white, however what if a color logo provided? 
This approach will break such logo.

cvat/apps/engine/serializers.py
```diff
@@ -2627,6 +2627,8 @@ class AboutSerializer(serializers.Serializer):
     name = serializers.CharField(max_length=128)
     description = serializers.CharField(max_length=2048)
     version = serializers.CharField(max_length=64)
+    logo = serializers.CharField()
```
SpecLad, line 2630:
```suggestion
    logo_url = serializers.CharField()
```

cvat/apps/engine/views.py
```diff
@@ -189,6 +189,9 @@
 _DATA_UPDATED_DATE_HEADER_NAME = 'X-Updated-Date'
 _RETRY_AFTER_TIMEOUT = 10
 
+def get_logo_uri(request) -> str:
+    return request.build_absolute_uri(f'{settings.STATIC_URL}{settings.LOGO_FILENAME}')
```
SpecLad, line 193:
```suggestion
    return request.build_absolute_uri(storages["staticfiles"].url(settings.LOGO_FILENAME))
```

This will correctly handle URL escaping and concatenation.

cvat/apps/engine/views.py
```diff
@@ -189,6 +189,9 @@
 _DATA_UPDATED_DATE_HEADER_NAME = 'X-Updated-Date'
 _RETRY_AFTER_TIMEOUT = 10
 
+def get_logo_uri(request) -> str:
```
SpecLad, line 192: This seems like it should be a private method inside `ServerViewSet` (or just inlined).

cvat/apps/engine/views.py
SpecLad: I don't think we should make things customizable without a clear use case.

cvat/settings/base.py
```diff
@@ -769,3 +769,16 @@ class CVAT_QUEUES(Enum):
 
 # Indicates the maximum number of days a file or directory is retained in the temporary directory
 TMP_FILE_OR_DIR_RETENTION_DAYS = 3
+
+LOGO_FILENAME = 'logo.png'
```
SpecLad, line 773:
```suggestion
LOGO_FILENAME = 'logo.svg'
```

cvat/apps/engine/views.py
klakhov: Ok, removed. Lets keep only subtitle for now

cvat/apps/engine/views.py
klakhov: Feels like inlined would be simplier. Added.

cvat/apps/engine/views.py
klakhov: Updated

SpecLad (approved):

bsekachev (approved):

PR #9063, opened 2025-02-06
Enhance Annotation Layer Changes: Relative zOrder Updates

arianizadi:
### Checklist

### License

### DIFF
--- a/cvat-ui/src/containers/annotation-page/standard-workspace/objects-side-bar/object-item.tsx
+++ b/cvat-ui/src/containers/annotation-page/standard-workspace/objects-side-bar/object-item.tsx
@@ -419,8 +419,6 @@ class ObjectItemContainer extends React.PureComponent<Props, State> {
                 switchOrientation={this.switchOrientation}
                 toBackground={this.toBackground}
                 toForeground={this.toForeground}
-                zIndexForwards={this.zIndexForwards}
-                zIndexBackwards={this.zIndexBackwards}
                 changeColor={this.changeColor}
                 changeLabel={this.changeLabel}
                 edit={this.edit}
--- a/cvat-ui/src/containers/annotation-page/standard-workspace/objects-side-bar/object-item.tsx
+++ b/cvat-ui/src/containers/annotation-page/standard-workspace/objects-side-bar/object-item.tsx
@@ -274,7 +274,7 @@ class ObjectItemContainer extends React.PureComponent<Props, State> {
         const { objectState, readonly, minZLayer } = this.props;
 
         if (!readonly) {
-            objectState.zOrder = minZLayer - 1;
+            objectState.zOrder = Math.max(objectState.zOrder - 1, minZLayer);
             this.commit();
         }
     };
@@ -283,7 +283,7 @@ class ObjectItemContainer extends React.PureComponent<Props, State> {
         const { objectState, readonly, maxZLayer } = this.props;
 
         if (!readonly) {
-            objectState.zOrder = maxZLayer + 1;
+            objectState.zOrder = Math.min(objectState.zOrder + 1, maxZLayer);
             this.commit();
         }
     };
@@ -419,6 +419,8 @@ class ObjectItemContainer extends React.PureComponent<Props, State> {
                 switchOrientation={this.switchOrientation}
                 toBackground={this.toBackground}
                 toForeground={this.toForeground}
+                zIndexForwards={this.zIndexForwards}
+                zIndexBackwards={this.zIndexBackwards}
                 changeColor={this.changeColor}
                 changeLabel={this.changeLabel}
                 edit={this.edit}
--- a/cvat-ui/src/containers/annotation-page/standard-workspace/objects-side-bar/objects-list.tsx
+++ b/cvat-ui/src/containers/annotation-page/standard-workspace/objects-side-bar/objects-list.tsx
@@ -577,7 +577,7 @@ class ObjectsListContainer extends React.PureComponent<Props, State> {
                 preventDefault(event);
                 const state = activatedState(true);
                 if (state && !readonly && state.objectType !== ObjectType.TAG) {
-                    state.zOrder = minZLayer - 1;
+                    state.zOrder = Math.max(state.zOrder - 1, minZLayer);
                     updateAnnotations([state]);
                 }
             },
@@ -585,7 +585,7 @@ class ObjectsListContainer extends React.PureComponent<Props, State> {
                 preventDefault(event);
                 const state = activatedState(true);
                 if (state && !readonly && state.objectType !== ObjectType.TAG) {
-                    state.zOrder = maxZLayer + 1;
+                    state.zOrder = Math.min(state.zOrder + 1, maxZLayer);
                     updateAnnotations([state]);
                 }
             },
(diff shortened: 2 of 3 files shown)


arianizadi:
Would this kind of change need a test? I did not find any tests for the previous implementation.

klakhov:
Hi,
Moving objects between layers could be a useful feature. However, I don’t think we should completely remove the `To Background` and `To Foreground` functions. What do you think about adding additional shortcuts or buttons to support moving objects between layers?

An `e2e` test would be much appreciated. You can find basic tests for z order [here](https://github.com/cvat-ai/cvat/blob/develop/tests/cypress/e2e/actions_objects2/case_16_z_order_features.js)

arianizadi:
That makes sense. I will try to fix that and write the tests. Thank you!

klakhov:
Any updates on this?

arianizadi:
Sorry for the delay, school had gotten busy. I updated the code. I will work on the e2e tests.

arianizadi:
I’m having trouble writing a test for the toForeground button. When clicked, the button creates a new layer on top, but I can’t figure out how to move the slider in Cypress to reveal that layer.

I’ve tried:
	•	Simulating button clicks to move the slider
	•	Setting the current z-value using JavaScript
	•	Clicking the bottom of the slider

None of these approaches have worked. If anyone has suggestions, I’d really appreciate the help!

arianizadi:
Everything should be working now. All tests are passing.

arianizadi:
What icons would be preferred for the two new buttons? Also, I forgot to mention I added E2E tests for the existing toForeground and toBackground buttons.

cvat-ui/src/containers/annotation-page/standard-workspace/objects-side-bar/object-item.tsx
```diff
@@ -288,6 +288,30 @@ class ObjectItemContainer extends React.PureComponent<Props, State> {
...
+        const { objectState, readonly, maxZLayer } = this.props;
+
+        if (!readonly) {
+            const newZOrder = objectState.zOrder + 1;
+            if (newZOrder >= maxZLayer) {
+                objectState.zOrder = newZOrder;
+                this.commit();
+            }
+        }
+    };
```
klakhov, line 313: I'm not entirely sure about this logic. We apply this action only when the target layer is less than `minZLayer` or greater than `maxZLayer`—why? I would expect this action to work even when, for example, `minZLayer` is -1, `maxZLayer` is 5, and the current layer is 2. In that case, pressing 'next' should set it to 3, and pressing 'previous' should set it to 1.

cvat-ui/src/containers/annotation-page/standard-workspace/objects-side-bar/objects-list.tsx
```diff
@@ -589,6 +601,24 @@ class ObjectsListContainer extends React.PureComponent<Props, State> {
                     updateAnnotations([state]);
                 }
             },
+            MOVE_TO_PREVIOUS_LAYER: (event: KeyboardEvent | undefined) => {
+                preventDefault(event);
+                const state = activatedState(true);
+                if (state && !readonly && state.objectType !== ObjectType.TAG) {
+                    const newZOrder = state.zOrder - 1;
+                    state.zOrder = newZOrder;
+                    updateAnnotations([state]);
```
klakhov, line 610: It seems shortcut logic is different compared to logic that happens in menu clicks

cvat-ui/src/containers/annotation-page/standard-workspace/objects-side-bar/objects-list.tsx
```diff
@@ -143,6 +143,18 @@ const componentShortcuts = {
         sequences: ['+', '='],
         scope: ShortcutScope.OBJECTS_SIDEBAR,
     },
+    MOVE_TO_PREVIOUS_LAYER: {
+        name: 'Move to previous layer',
+        description: 'Move the active object one layer backward (decrease z-order value)',
+        sequences: ['u'],
```
klakhov, line 149: I would purpose to use different shortcuts. As our current operations work with `-` and `+`. It would be better to use the same buttons but with `shift` modifier

tests/cypress/e2e/actions_objects2/case_18_z_order_buttons.js
```diff
@@ -0,0 +1,57 @@
+/// <reference types="cypress" />
+import { taskName, labelName } from '../../support/const';
+
+describe('Z-order button actions', () => {
+    const performZOrderAction = (actionSelector, expectedZOrder) => {
+        cy.wait(500);
```
klakhov, line 6: Please, avoid using `wait`

klakhov:
As for the icons, I would suggest using something like a box with an arrow pointing up and down, respectively. Or maybe just an arrow. However, I believe the first option would be better.

cvat-ui/src/containers/annotation-page/standard-workspace/objects-side-bar/object-item.tsx
arianizadi: Yes, you are correct. Do you think it is more useful for the functionality to be clamped to max and min layer or do you think it should just move to the next or previous without clamping? Like in the implementation for the shortcut logic?

arianizadi: I will just use the non clamped method. The clamping causes extra button presses for creating new layers.

cvat-ui/src/containers/annotation-page/standard-workspace/objects-side-bar/objects-list.tsx
arianizadi: I tried implementing it using the shift modifier, but it triggers both. I instead just used the open bracket and close bracket.

klakhov: Hi, sorry for misleading you. `shift+plus` wont really work as to press `plus` we need shift modifier.

```
    TO_BACKGROUND: {
        name: 'To background',
        description: 'Put an active object "farther" from the user (decrease z axis value)',
        sequences: ['-'],
        scope: ShortcutScope.OBJECTS_SIDEBAR,
    },
    TO_FOREGROUND: {
        name: 'To foreground',
        description: 'Put an active object "closer" to the user (increase z axis value)',
        sequences: ['='],
        scope: ShortcutScope.OBJECTS_SIDEBAR,
    },
    MOVE_TO_PREVIOUS_LAYER: {
        name: 'Move to previous layer',
        description: 'Move the active object one layer backward (decrease z-order value)',
        sequences: ['_'],
        scope: ShortcutScope.OBJECTS_SIDEBAR,
    },
    MOVE_TO_NEXT_LAYER: {
        name: 'Move to next layer',
        description: 'Move the active object one layer forward (increase z-order value)',
        sequences: ['+'],
        scope: ShortcutScope.OBJECTS_SIDEBAR,
    },
```

It seems we need to do something like this to achive the behavior I've thought about. Like we press usual buttons `-=` to move all layers and the same ones with shift to move it one by one

klakhov: Regarding the keybinding issue — it occurs because CVAT stores settings (including shortcuts) in `localStorage` under the `clientSettings` key. When the app is rebuilt, it restores from this local storage and overwrites the updated settings.

There are two ways to fix this:

1. Manually clear the `clientSettings` key from `localStorage` via browser devtools.
2. Use the **Restore defaults** button in the settings window:
   image

arianizadi: No worries, thank you! I will try this and test.

arianizadi:
The proper key bindings should be working now. Is there anything else I should update? @klakhov

arianizadi:
@klakhov @bsekachev @nmanovic — Is anything else needed before final review and merge?

bsekachev:
Hello, I updated PR a little bit, thank you for the contribution!

PR #9075, opened 2025-02-07
[WIP] API to download export results

Marishka17:
### Checklist

### License


cvat/apps/engine/mixins.py
```diff
@@ -474,14 +450,30 @@ def export_dataset_v1(
...
+    @extend_schema(summary='Download a prepared dataset file',
+        parameters=[
+            OpenApiParameter('rq_id', description='Request ID',
+                location=OpenApiParameter.QUERY, type=OpenApiTypes.STR, required=True),
+        ],
+        responses={
+            '200': OpenApiResponse(description='Download of file started'),
+            '204': OpenApiResponse(description='No prepared dataset file related with provider request ID'),
+        },
+        exclude=True, # private API endpoint that should be used only as result_url
```
SpecLad, line 470: Given this, does it make sense to set any other parameters? Seems like they're just going to be ignored.

cvat/apps/engine/mixins.py
```diff
@@ -474,14 +450,30 @@ def export_dataset_v1(
...
     )
     @action(detail=True, methods=['POST'], serializer_class=None, url_path='dataset/export')
-    def export_dataset_v2(self, request: ExtendedRequest, pk: int):
+    def initiate_dataset_export(self, request: ExtendedRequest, pk: int):
         self._object = self.get_object() # force call of check_object_permissions()
 
-        save_images = is_dataset_export(request)
-        callback = self.get_export_callback(save_images)
+        export_manager = DatasetExportManager(self._object, request)
+        export_manager.initialize_export_args()
```
SpecLad, line 457: Any reason why `initialize_export_args` is not part of `export`?

cvat/apps/engine/permissions.py
```diff
@@ -288,15 +298,19 @@ def get_scopes(request: ExtendedRequest, view: ViewSet, obj: Project | None):
...
+            ('initiate_dataset_export', 'POST'): Scopes.EXPORT_DATASET if is_dataset_export(request) else Scopes.EXPORT_ANNOTATIONS,
+            ('initiate_backup_export', 'POST'): Scopes.EXPORT_BACKUP,
             ('import_backup', 'POST'): Scopes.IMPORT_BACKUP,
             ('append_backup_chunk', 'PATCH'): Scopes.IMPORT_BACKUP,
             ('append_backup_chunk', 'HEAD'): Scopes.IMPORT_BACKUP,
             ('preview', 'GET'): Scopes.VIEW,
+            ('download_dataset', 'GET'): Scopes.DOWNLOAD_EXPORTED_FILE,
+            ('download_backup', 'GET'): Scopes.DOWNLOAD_EXPORTED_FILE,
+            # FUTURE-TODO: delete this after dropping support for deprecated API
+            ('annotations', 'GET'): Scopes.EXPORT_ANNOTATIONS,
```
SpecLad, line 310: `GET annotations` is not supposed to be deleted, though, is it? It's still used for fetching raw annotations.

cvat/apps/engine/permissions.py
```diff
@@ -404,6 +426,7 @@ class Scopes(StrEnum):
         EXPORT_BACKUP = 'export:backup'
         VIEW_VALIDATION_LAYOUT = 'view:validation_layout'
         UPDATE_VALIDATION_LAYOUT = 'update:validation_layout'
+        DOWNLOAD_EXPORTED_FILE = 'retrieve:exported_file'
```
SpecLad, line 429: The variable name is inconsistent with the value.

cvat/apps/engine/permissions.py
```diff
@@ -657,6 +698,12 @@ def create(cls, request: ExtendedRequest, view: ViewSet, obj: Job | None, iam_co
                             request, task, iam_context=iam_context
                         ))
 
+                elif scope == cls.Scopes.DOWNLOAD_EXPORTED_FILE:
+                    rq_id = request.query_params.get("rq_id")
+                    if not rq_id:
+                        raise ValidationError("Missing request id in query parameters")
+                    scope_params["rq_job_id"] = RQId.parse(rq_id)
```
SpecLad, line 705: This logic is repeated multiple times; could you factor it out?

cvat/apps/engine/permissions.py
```diff
@@ -776,6 +825,13 @@ def get_resource(self):
                     "assignee": { "id": self.obj.segment.task.project.assignee_id }
                 } if self.obj.segment.task.project else None
             }
+            if __class__.Scopes.DOWNLOAD_EXPORTED_FILE == self.scope:
+                # extend data with rq job owner
+                data["rq_job"] = {
+                    "owner": {
+                        "id": self.rq_job_id.user_id if self.rq_job_id else None
+                    }
+                }
```
SpecLad, line 834: This is repeated multiple times as well.

cvat/apps/engine/rules/jobs.rego
```diff
@@ -273,3 +274,8 @@ allow if {
     organizations.has_perm(organizations.MAINTAINER)
     utils.has_perm(utils.USER)
 }
+
+allow if {
+    input.scope == utils.DOWNLOAD_EXPORTED_FILE
+    input.auth.user.id == input.resource.rq_job.owner.id
+}
```
SpecLad, line 281: Add the missing newline, please.

cvat/apps/engine/utils.py
```diff
@@ -313,12 +312,12 @@ def build_annotations_file_name(
     timestamp: str,
     format_name: str,
     is_annotation_file: bool = True,
-    extension: str = "{}",
+    extension: str = "zip",
```
SpecLad, line 315: I don't think the default is really needed here - it's never used.

cvat/apps/engine/views.py
```diff
@@ -398,51 +386,52 @@ def perform_create(self, serializer, **kwargs):
...
-            To check the status of the process of importing a project dataset from a file:
-
-            After initiating the dataset upload, you will receive an rq_id parameter.
-            Make sure to include this parameter as a query parameter in your subsequent
-            GET /api/projects/id/dataset requests to track the status of the dataset import.
-            Also you should specify action parameter: action=import_status.
+            Utilizing this endpoint:
+            - to export project dataset in a specific format
+            - to check the status of the process of importing a project dataset from a file
+            is deprecated.
```
SpecLad, line 394: This doesn't seem like a correct description. Checking the import status is deprecated, but initiating an export is _removed_, right?

cvat/apps/engine/views.py
```diff
@@ -579,87 +584,44 @@ def upload_finished(self, request: ExtendedRequest):
...
-            ), description='Download of file started'),
-            '201': OpenApiResponse(description='Annotations file is ready to download'),
-            '202': OpenApiResponse(description='Dump of annotations has been started'),
-            '401': OpenApiResponse(description='Format is not specified'),
-            '405': OpenApiResponse(description='Format is not available'),
-        })
-    @action(detail=True, methods=['GET'],
-        serializer_class=LabeledDataSerializer,
-        csrf_workaround_is_needed=csrf_workaround_is_needed_for_export)
+            '410': OpenApiResponse(description="API endpoint is no longer handles exporting process"),
```
SpecLad, line 598: Do we even need to keep this endpoint in the schema if it doesn't do anything useful? I can see the value of keeping the endpoint itself, so that we can return a useful status to any existing client, but what purpose does keeping it in the schema serve?

changelog.d/20250304_120308_maria_api_to_download_export_results.md
```diff
@@ -0,0 +1,29 @@
+### Deprecated
+
+- Utilizing `GET /api/projects/id/dataset?action=import_status` API endpoint
+  to check the status of the import process (<https://github.com/cvat-ai/cvat/pull/9075>)
+
+### Changed
```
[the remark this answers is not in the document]
Marishka17: I was thinking about the more appropriate section for these lines but decided to put them into `changed` because API endpoints have not been deleted.

cvat/apps/engine/mixins.py
Marishka17: If I understand you correctly, you are talking about defining the whole API endpoint schema here, right? I think it would still be better to explicitly define the endpoint schema because it would be helpful for developers.

cvat/apps/engine/permissions.py
Marishka17: These are project permissions, there is no API to retrieve project annotations.

changelog.d/20250304_120308_maria_api_to_download_export_results.md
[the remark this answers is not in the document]
SpecLad: Seems like a distinction without a difference. If an endpoint always returns 410, then it effectively _has_ been deleted.

cvat/apps/engine/permissions.py
SpecLad: Ah, my bad.

cvat/apps/engine/mixins.py
SpecLad: Okay, suit yourself.

IMO, since you don't need to provide a full OpenAPI schema, it would be more readable to summarize the behavior of the endpoint in a concise freeform comment (you could even reduce duplication by putting that comment into `download_file`). But you're more likely to edit this part of the code than me, so I'll defer to your judgement.

Marishka17:
@zhiltsov-max, Could you please check that deprecated API is not used in Human Protocol oracles (and update if needed)?

cvat/apps/dataset_manager/views.py
```diff
@@ -235,3 +235,13 @@ def get_all_formats():
         'importers': get_import_formats(),
         'exporters': get_export_formats(),
     }
+
+
+def get_export_callback(db_instance: Project | Task | Job, save_images: bool):
+    if isinstance(db_instance, Project):
+        return export_project_as_dataset if save_images else export_project_annotations
```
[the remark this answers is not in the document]
Marishka17: > change export to take a db_instance in the first place

I'm not sure that this is a good approach, IMHO scheduled background job should not take some model objects because:
1. the job might start with an outdated object or when it is gone
2. it requires more space in Redis

If I understood you correctly :eyes: and you are talking about keeping only [this](https://github.com/cvat-ai/cvat/blob/9247d7aee3ede762339a2b30ff21c654139745a2/cvat/apps/dataset_manager/views.py#L114) function

cvat/apps/engine/views.py
Marishka17: Probbaly not to break SDK?

cvat/apps/dataset_manager/views.py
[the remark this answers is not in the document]
SpecLad: Ah, I see. That's a fair point.

There's still a possible simplification that could be made (`export_project_as_dataset` could be merged with `export_project_annotations`, etc.), but it seems that it would require another migration, so it doesn't seem worth doing.

cvat/apps/engine/views.py
SpecLad: But you deleted the query parameters from the schema, which would break it anyway.

(Also, just to clarify: what SDK usage scenario are you talking about? Someone using the low-level API with the current version of the SDK to interact with an old version of CVAT?)

cvat/apps/engine/background.py
```diff
@@ -117,41 +106,115 @@ def _handle_rq_job_v2(self, rq_job: Optional[RQJob], queue: DjangoRQ) -> Optiona
...
+            file_path = rq_job.return_value()
+
+            if not file_path:
+                return (
+                    Response(
+                        "A result for exporting job was not found for finished RQ job",
+                        status=status.HTTP_500_INTERNAL_SERVER_ERROR,
+                    )
+                    if rq_job_meta.result_url
+                    else Response(status=status.HTTP_204_NO_CONTENT)
```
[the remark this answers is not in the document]
Marishka17: handling cases when this API is used to download a final result of an rq job when the export process was initiated to some cloud storage. I've changed the response status code from 204 to 400.

cvat/apps/engine/views.py
Marishka17: > Also, just to clarify: what SDK usage scenario are you talking about? 

New SDK version + new server version. But okay, I've deleted these endpoints from the server schema.

Marishka17: now it's done

SpecLad: > New SDK version + new server version.

When used with a new server version, the SDK functions corresponding to the removed endpoints would not work anyway.

cvat/apps/engine/views.py
```diff
@@ -1483,50 +1387,50 @@ def append_data_chunk(self, request: ExtendedRequest, pk: int, file_id: str):
...
-                serializers=[LabeledDataSerializer, OpenApiTypes.BINARY],
-                resource_type_field_name=None
-            ), description='Download of file started'),
-            '201': OpenApiResponse(description='Annotations file is ready to download'),
-            '202': OpenApiResponse(description='Dump of annotations has been started'),
-            '400': OpenApiResponse(description='Exporting without data is not allowed'),
-            '405': OpenApiResponse(description='Format is not available'),
+            '200': OpenApiResponse(LabeledDataSerializer),
+            '400': OpenApiResponse(description="Exporting without data is not allowed"),
+            '410': OpenApiResponse(description="API endpoint is no longer handles exporting process"),
```
SpecLad, line 1433:
```suggestion
            '410': OpenApiResponse(description="API endpoint no longer handles exporting process"),
```

cvat/apps/engine/tests/utils.py
```diff
@@ -125,6 +128,252 @@ def setUp(self):
...
+        with ForceLogin(user, self.client):
+            response = self.client.post(path, data=data)
+        return response
+
+    def _put_request(self, url: str, user: str, *, data: dict[str, Any] | None = None):
+        with ForceLogin(user, self.client):
+            response = self.client.put(url, data=data)
+        return response
+
+    def _wait_request_to_be_finished(
```
SpecLad, line 151: It's odd to see this here. IIRC, for unit tests we run RQ jobs in synchronous mode, so no waiting should be necessary - the job should be complete immediately after the initiating request.

cvat/apps/engine/tests/utils.py
```diff
@@ -125,6 +128,252 @@ def setUp(self):
...
+            request_status = response.json()["status"]
+            if request_status in ("finished", "failed"):
+                break
+
+            sleep(sleep_interval)
+        assert request_status == "finished", f"The last request status was {request_status}"
+        return response
+
+    def _query_params_to_str(self, **params: dict[str, Any]) -> str:
+        return "?" + "&".join([f"{k}={v}"for k, v in params.items()])
```
SpecLad, line 176:
```suggestion
        return "?" + "&".join([f"{k}={quote(v)}" for k, v in params.items()])
```

cvat/apps/engine/tests/utils.py
```diff
@@ -125,6 +128,252 @@ def setUp(self):
...
+        if query_params:
+            api_path += self._query_params_to_str(**query_params, save_images=True)
+        return api_path
+
+    def _get_api_path_to_export_job_annotations(self, job_id: int, *, query_params: dict[str, Any] | None = None) -> str:
+        api_path = f"/api/jobs/{job_id}/dataset/export"
+        if query_params:
+            api_path += self._query_params_to_str(**query_params, save_images=False)
+        return api_path
+
```
SpecLad, line 270: I'm not sure these functions are warranted. Each of them is used exactly once, so you're not reducing any duplication (and IMO, it decreases readability because each function is far away from its caller).

Also, I think it would be better to move the `query_params_to_str` call to `_export`, which _would_ reduce duplication.

cvat/apps/engine/tests/utils.py
```diff
@@ -125,6 +128,252 @@ def setUp(self):
...
+    def _get_api_path_to_export_project_annotations(self, project_id: int, *, query_params: dict[str, Any] | None = None) -> str:
+        api_path = f"/api/projects/{project_id}/dataset/export"
+        if query_params:
+            api_path += self._query_params_to_str(**query_params, save_images=False)
+        return api_path
+
+    def _get_api_path_to_export_task_dataset(self, task_id: int, *, query_params: dict[str, Any] | None = None) -> str:
+        api_path = f"/api/tasks/{task_id}/dataset/export"
+        if query_params:
+            api_path += self._query_params_to_str(**query_params, save_images=True)
```
SpecLad, line 250: If `query_params` is `None`, then the `save_images` parameter will never be added - presumably that's not the desired effect?

Every export requires at least the `format` parameter, though, so it seems like you could make `query_params` required.

cvat/apps/engine/tests/utils.py
```diff
@@ -125,6 +128,252 @@ def setUp(self):
...
+        result_url = response.json().get("result_url")
+        assert result_url, "The result_url param was not found in the server response"
+
+        response = self._get_request(result_url, user)
+        self.assertEqual(response.status_code, expected_4xx_status_code or status.HTTP_200_OK)
+
+        if not expected_4xx_status_code and file_path:
+            content = BytesIO(b"".join(response.streaming_content))
+            with open(file_path, "wb") as f:
+                f.write(content.getvalue())
```
SpecLad, line 219:
```suggestion
            with open(file_path, "wb") as f:
                f.write(response.content)
```

cvat/apps/engine/views.py
```diff
@@ -1483,50 +1387,50 @@ def append_data_chunk(self, request: ExtendedRequest, pk: int, file_id: str):
...
+            ),
             OpenApiParameter('cloud_storage_id', description='Storage id',
-                location=OpenApiParameter.QUERY, type=OpenApiTypes.INT, required=False),
-            OpenApiParameter('use_default_location', description='Use the location that was configured in the task to export annotation',
-                location=OpenApiParameter.QUERY, type=OpenApiTypes.BOOL, required=False,
-                default=True, deprecated=True),
+                location=OpenApiParameter.QUERY, type=OpenApiTypes.INT, required=False,
+                deprecated=True
+            ),
+            # --- Deprecated params section ---
```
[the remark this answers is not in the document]
Marishka17: No, this is the end of the section

cvat/apps/engine/views.py
```diff
@@ -1483,50 +1387,50 @@ def append_data_chunk(self, request: ExtendedRequest, pk: int, file_id: str):
...
-            - GET /api/requests/<rq_id> to check process status,
-                where rq_id is request id returned on initializing request
+            - `POST /api/tasks/<task_id>/dataset/export?save_images=False` to initiate export process
+            - `GET /api/requests/<rq_id>` to check process status,
+                where `rq_id` is request id returned on initializing request
+            - `GET result_url` to download a prepared file,
+                where `result_url` can be found in the response on checking status request
         """),
         parameters=[
+            # --- Deprecated params section ---
```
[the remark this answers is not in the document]
Marishka17: I did not remove these parameters from the specification because they are still used to determine whether a request is made to export the dataset or just to get raw annotations. So, I think it would be better to remove them from the schema once the endpoint no longer returns a 410 response.

cvat/apps/engine/tests/utils.py
Marishka17: it's true, but I believe there are still several advantages of adding a waiting step:
1. We do not need to update related tests if we run these tests in asynchronous mode
2. This way we test that `/api/requests/rq_id` API returns the expected code (200, 401, 403)

cvat/apps/engine/tests/utils.py
Marishka17: You are right, thanks

cvat/apps/engine/views.py
```diff
@@ -197,6 +191,15 @@
 _DATA_UPDATED_DATE_HEADER_NAME = 'X-Updated-Date'
 _RETRY_AFTER_TIMEOUT = 10
 
+def get_410_response_for_export_api(path: str) -> HttpResponseGone:
+    return HttpResponseGone(textwrap.dedent(f"""\
+        This endpoint is no longer supported.
+        To initiate the export process, use POST {path}.
+        To check the process status, use GET /api/requests/<rq_id>,
```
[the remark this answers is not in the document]
SpecLad: The paths you pass to this function still contain angle brackets.

cvat/apps/engine/views.py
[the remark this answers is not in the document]
SpecLad: I disagree. By listing these parameters in the schema you are making the claim that they are supported, whereas in fact they are not. I can see the utility in keeping them around for a short while, so that people who read the schema can understand what happened to them; but in this case you should change the description to something accurate, e.g. "this parameter is no longer supported".

FWIW, I also think we should keep the logic that returns 410 if these parameters are present indefinitely, or at least for a long time. It costs very little, and it ensures that clients that use the old API get an error. OTOH, we probably don't need to keep these parameters in the schema for very long.

cvat/apps/engine/tests/utils.py
SpecLad: >  if we run these tests in asynchronous mode

Is it possible?

> This way we test that /api/requests/rq_id API returns the expected code (200, 401, 403)

I've nothing against calling this API, I just don't think we need to do it in a loop.

SpecLad: > Warum nicht?

Well... how would you do it? AFAICS, `cvat.settings.testing` forces all queues to use sync mode. Moreover, it also uses a fake Redis implementation, so it wouldn't be possible to run RQ workers anyway.

SpecLad (approved):

PR #8989, opened 2025-01-24
Added a button to frame player to copy filename

klakhov:
### Motivation and context

Added a simple button to quickly copy frame name.

Resolved #7237

### Checklist

- [x] I submit my changes into the `develop` branch
- [x] I have created a changelog fragment 
- [x] I have updated the documentation accordingly
- ~~[ ] I have added tests to cover my changes~~
- [x] I have linked related issues (see [GitHub docs](
  https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword))

### License

- [x] I submit _my code changes_ under the same [MIT License](
  https://github.com/cvat-ai/cvat/blob/develop/LICENSE) that covers the project.
  Feel free to contact the maintainers if that's a concern.

PR #9081, opened 2025-02-10
Test for button on frame player that copies the current frame's filename

archibald1418:
This checks that the frame filename displayed in the annotation view UI is copied to clipboard correctly
Regression test for #8989 

### How has this been tested?
Use the existing Main task to trigger UI events that change clipboard contents.
##### before the test suite
- Create main task
- Open main task job

##### before each test case
- Add a spy on the window's clipboard (has to be new every case)
_Note: Unlike mocks, spies call the real function to gain information on how this function was called_

#### Case 1: Check that frame filename can be copied to clipboard
- Scrape active filename
- Press on button, confirm that both filenames are equal

#### Case 2: Check clipboard after switching frames
- Switch to next frame
- Repeat actions from Case 1

### Checklist

### License


tests/cypress/e2e/actions_objects2/case_copy_frame_filename.js
```diff
@@ -0,0 +1,90 @@
...
+            .then((clipboard) => {
+                cy.spy(clipboard, 'writeText').as('copyTextToClipboard');
+            });
+    });
+    describe('Open job, trigger events that change clipboard content', () => {
+        function scrapeFrameFilename() {
+            return cy.get('.cvat-player-filename-wrapper').invoke('text');
+        }
+
+        function clipboard() {
```
klakhov, line 33: It's better to name this `copyToClipboard` since it not only retrieves the clipboard contents but also triggers the action by pressing the button.

tests/cypress/e2e/actions_objects2/case_copy_frame_filename.js
```diff
@@ -0,0 +1,90 @@
...
+        });
+
+        it('Check clipboard after switching frames', () => {
+            cy.goToNextFrame(1);
+            scrapeFrameFilename().then((fileName) => {
+                clipboard().should('equal', fileName);
+            });
+        });
+
+        it('Check clipboard after deleting frame, after Undo', () => {
```
klakhov, line 55: I'm not entirely sure if we need tests that mix copying and deleting, as these are separate functionalities. Deleting a frame shouldn't affect copying. If treated separately, we are essentially repeating the same operations as in the previous two tests: copy → change frame → copy again. Even if there were a potential intersection, it would be considered a regression and should be covered by a dedicated regression test. Since we haven't encountered such a regression, I suggest we remove these test cases for now.

archibald1418: Ok makes sense

tests/cypress/e2e/actions_objects2/case_copy_frame_filename.js
archibald1418: applying

klakhov (approved):

PR #9082, opened 2025-02-10
Refactor way of working with RQ meta

Marishka17:
### Motivation and context

This PR changes the way we work with RQ meta. The previous implementation tried to standardize the fields saved in the meta and how to access them, however, we had to remember the structure of nested objects. This approach explicitly defines the structure of the meta being used.
Extracted from https://github.com/cvat-ai/cvat/pull/9075
Partially depends on https://github.com/cvat-ai/cvat/pull/9077

### Checklist

### License


cvat/apps/engine/rq_job_handler.py
```diff
@@ -4,80 +4,301 @@
...
+from cvat.apps.engine.types import ExtendedRequest
+
 from .models import RequestAction, RequestSubresource, RequestTarget
 
+str_validator = attrs.validators.instance_of(str)
+int_validator = attrs.validators.instance_of(int)
+optional_str_validator = attrs.validators.optional(attrs.validators.instance_of(str))
+optional_int_validator = attrs.validators.optional(attrs.validators.instance_of(int))
+optional_bool_validator = attrs.validators.optional(attrs.validators.instance_of(bool))
+optional_float_validator = attrs.validators.optional(attrs.validators.instance_of(float))
```
SpecLad, line 28: These should probably be underscore-prefixed.

cvat/apps/engine/rq_job_handler.py
```diff
@@ -4,80 +4,301 @@
...
-    USER = 'user'
-    PROJECT_ID = 'project_id'
-    TASK_ID = 'task_id'
-    JOB_ID = 'job_id'
-    ORG_ID = 'org_id'
-    ORG_SLUG = 'org_slug'
-    STATUS = 'status'
-    PROGRESS = 'progress'
-    TASK_PROGRESS = 'task_progress'
+    REQUEST = "request"
```
SpecLad, line 260: Did you run this file through black? If so, could you delete it from the exceptions in `pyproject.toml`?

cvat/apps/engine/serializers.py
```diff
@@ -3545,17 +3557,23 @@ class RequestSerializer(serializers.Serializer):
...
 
+    def __init__(self, *args, **kwargs):
+        self._base_rq_job_meta: BaseRQMeta | None = None
+        super().__init__(*args, **kwargs)
+
     @extend_schema_field(UserIdentifiersSerializer())
     def get_owner(self, rq_job: RQJob) -> dict[str, Any]:
-        return UserIdentifiersSerializer(rq_job.meta[RQJobMetaField.USER]).data
+        assert self._base_rq_job_meta
+        return UserIdentifiersSerializer(self._base_rq_job_meta.user.to_dict()).data
```
SpecLad, line 3567: Seems like `to_dict` shouldn't be necessary here.

cvat/apps/engine/serializers.py
```diff
@@ -3573,29 +3591,28 @@ def get_expiry_date(self, rq_job: RQJob) -> Optional[str]:
...
         representation = super().to_representation(rq_job)
 
         # FUTURE-TODO: support such statuses on UI
         if representation["status"] in (RQJobStatus.DEFERRED, RQJobStatus.SCHEDULED):
             representation["status"] = RQJobStatus.QUEUED
 
-        if representation["status"] == RQJobStatus.FINISHED:
-            if result_url := rq_job.meta.get(RQJobMetaField.RESULT_URL):
-                representation["result_url"] = result_url
+        if  representation["status"] == RQJobStatus.FINISHED:
```
SpecLad, line 3613: I think you accidentally added a space here.

cvat/apps/engine/task.py
```diff
@@ -592,10 +594,13 @@ def _create_thread(
     slogger.glob.info("create task #{}".format(db_task.id))
 
     job = rq.get_current_job()
+    rq_job_meta = ImportRQMeta.from_job(job)
 
-    def _update_status(msg: str) -> None:
-        job.meta['status'] = msg
-        job.save_meta()
+    def _update_status(rq_job_meta: ImportRQMeta, msg: str) -> None:
```
SpecLad, line 599: Seems like you could do:

```suggestion
    def update_status(msg: str) -> None:
```

cvat/apps/engine/rq.py
```diff
@@ -0,0 +1,440 @@
...
+
+@attrs.define(kw_only=True)
+class RQMetaWithFailureInfo(AbstractRQMeta):
+    # mutable && optional fields
+    formatted_exception: str | None = attrs.field(
+        validator=[_optional_str_validator],
+        default=None,
+        on_setattr=on_setattr,
+    )
+    exc_type: type[Exception] | None = attrs.field(
```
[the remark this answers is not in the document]
Marishka17: I don't think so

cvat/apps/engine/rq.py
```diff
@@ -0,0 +1,440 @@
...
+    _job: RQJob | None = attrs.field(init=False, default=None)
+
+    def update_value(self, attribute: attrs.Attribute, value: Any):
+        self._job.meta[attribute.name] = value
+
+    def to_dict(self) -> dict:
+        return asdict(self, filter=lambda k, _: not k.name.startswith("_"))
+
+    @classmethod
+    def from_job(cls, rq_job: RQJob):
```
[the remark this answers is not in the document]
Marishka17: renamed to `for_job`

cvat/apps/engine/rq.py
```diff
@@ -0,0 +1,440 @@
...
+        keys_to_keep = [k.name for k in attrs.fields(cls) if not k.name.startswith("_")]
```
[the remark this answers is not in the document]
Marishka17: is no longer used

cvat/apps/engine/rq.py
```diff
@@ -0,0 +1,440 @@
...
+    def _get_resettable_fields() -> list[RQJobMetaField]:
+        """Return a list of fields that must be reset on retry"""
+
+    def reset_meta_on_retry(self) -> dict[RQJobMetaField, Any]:
+        resettable_fields = self._get_resettable_fields()
+
+        return {k: v for k, v in self._job.meta.items() if k not in resettable_fields}
+
+
+on_setattr = attrs.setters.pipe(attrs.setters.validate, AbstractRQMeta.update_value)
```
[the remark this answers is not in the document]
Marishka17: > but I don't think save is that useful anyway - it's just a one-line method.

I've changed the implementation a bit to support initialization from either a provided meta dictionary or an RQ job. I initially added the `save` method because, otherwise, it felt odd to construct a meta object for a job, modify some fields, and then update the data in Redis by calling a method from the original job. Like this:
```python
job = rq.get_current_job()
rq_job_meta = ImportRQMeta.for_job(job)
rq_job_meta.status = msg
job.save_meta()
```

cvat/apps/engine/rq.py
```diff
@@ -0,0 +1,527 @@
...
+
+class ImmutableRQMetaAttribute(_GettableRQMetaAttribute):
+    pass
+
+
+class MutableRQMetaAttribute(_GettableRQMetaAttribute, _SettableRQMetaAttribute):
+    pass
+
+
+class UserRQMetaAttribute(ImmutableRQMetaAttribute):
```
SpecLad, line 100: I don't think you're actually getting anything from the parent class here. This class could be reduced to:

```python
class UserRQMetaAttribute:
    def __get__(self, instance: WithMeta, objtype: type[WithMeta] | None = None):
        return UserMeta(instance.meta[RQJobMetaField.USER])
```

(In fact, I think it could even be replaced with a property in `BaseRQMeta`.)

cvat/apps/engine/rq.py
```diff
@@ -0,0 +1,527 @@
...
+    def __set__(self, instance: WithMeta, value: Any):
+        self.validate(value)
+        instance.meta[self._key] = value
+
+
+class ImmutableRQMetaAttribute(_GettableRQMetaAttribute):
+    pass
+
+
+class MutableRQMetaAttribute(_GettableRQMetaAttribute, _SettableRQMetaAttribute):
```
SpecLad, line 96: It doesn't seem like 5 classes are needed here. They could be combined into two - `ImmutableRQMetaAttribute` (merged `_AbstractRQMetaAttribute` and `_GettableRQMetaAttribute`), and `MutableRQMetaAttribute` (same as `_SettableRQMetaAttribute`, but inheriting from `ImmutableRQMetaAttribute`).

cvat/apps/engine/rq.py
```diff
@@ -0,0 +1,527 @@
...
+        RQJobMetaField.UserField.ID, validator=lambda x: isinstance(x, int)
+    )
+    username: str = ImmutableRQMetaAttribute(
+        RQJobMetaField.UserField.USERNAME, validator=lambda x: isinstance(x, str)
+    )
+    email: str = ImmutableRQMetaAttribute(
+        RQJobMetaField.UserField.EMAIL, validator=lambda x: isinstance(x, str)
+    )
+
+    def __init__(self, meta: dict[RQJobMetaField.UserField, Any]) -> None:
```
SpecLad, line 129: This type annotation is incorrect, because e.g. `UserField.ID` is not an instance of `UserField`.

cvat/apps/engine/rq.py
```diff
@@ -0,0 +1,527 @@
...
+        self._meta = meta
+
+    @property
+    def meta(self) -> dict[RQJobMetaField.UserField, Any]:
+        return self._meta
+
+    def to_dict(self):
+        return self.meta
```
SpecLad, line 137: Aren't these redundant?

cvat/apps/engine/rq.py
```diff
@@ -0,0 +1,527 @@
...
+    meta: dict[str, Any]
+
+
+class _AbstractRQMetaAttribute:
+    def __init__(
+        self, key: str, *, optional: bool = False, validator: Callable | None = None
+    ) -> None:
+        assert validator is None or callable(validator), "validator must be callable"
+        self._key = key
+        self._validator = validator
```
SpecLad, line 68: The `validator` parameter ought to be moved to `_SettableRQMetaAttribute`, since it doesn't do anything in an immutable attribute.

cvat/apps/engine/rq.py
```diff
@@ -0,0 +1,527 @@
...
+
+        return instance.meta[self._key]
+
+
+class _SettableRQMetaAttribute(_AbstractRQMetaAttribute):
+    def validate(self, value):
+        if value is None and not self._optional:
+            raise ValueError(f"{self._key} is required")
+        if value is not None and self._validator and not self._validator(value):
+            raise ValueError("Wrong type")
```
SpecLad, line 85: IMO, if the validator can only signal that the type is wrong, then you might as well replace the `validator` parameter with something like `expected_type`.

cvat/apps/engine/rq.py
```diff
@@ -0,0 +1,527 @@
...
+    @property
+    def meta(self) -> dict[RQJobMetaField, Any]:
+        return self._job.meta if self._job else self._meta
+
+    def to_dict(self):
+        return self.meta
+
+    @classmethod
+    def for_job(cls, job: RQJob):
+        return cls(job=job)
```
SpecLad, line 176:
```suggestion
        return cls(job=job, meta=job.meta)
```

This would simplify the rest of the class: no need for the `if` in `def meta`, no need for the `assert` in `__init__`.

cvat/apps/engine/rq.py
```diff
@@ -0,0 +1,527 @@
...
+
+        user = request.user
+
+        return cls.for_meta(
+            {
+                RQJobMetaField.USER: UserMeta(
+                    {
+                        RQJobMetaField.UserField.ID: user.id,
+                        RQJobMetaField.UserField.USERNAME: user.username,
+                        RQJobMetaField.UserField.EMAIL: getattr(user, "email", ""),
```
SpecLad, line 285: How could `email` be a missing attribute? Isn't it part of the `User` model?

cvat/apps/engine/rq.py
```diff
@@ -0,0 +1,527 @@
...
+                    }
+                ).to_dict(),
+                RQJobMetaField.REQUEST: RequestMeta(
+                    {
+                        RQJobMetaField.RequestField.UUID: request.uuid,
+                        RQJobMetaField.RequestField.TIMESTAMP: timezone.localtime(),
```
SpecLad, line 291: Why local time?

cvat/apps/engine/rq.py
```diff
@@ -0,0 +1,527 @@
...
+
+        oid = organization_id(db_obj)
+        oslug = organization_slug(db_obj)
+        pid = project_id(db_obj)
+        tid = task_id(db_obj)
+        jid = job_id(db_obj)
+
+        user = request.user
+
+        return cls.for_meta(
```
SpecLad, line 279: Isn't the class redundant here? You're creating a dictionary, wrapping it, then immediately unwrapping it.

cvat/apps/engine/rq.py
Marishka17: you are right, replaced with property

cvat/apps/engine/rq.py
Marishka17: I prefer to keep `validator`, changed exception message

cvat/apps/engine/rq.py
Marishka17: Hmm, because it was used previously when defining meta: https://github.com/cvat-ai/cvat/blame/4f3bc1007b0efb0836b1ac4bb6482887227205eb/cvat/apps/engine/utils.py#L259C18-L259C18
Fixed

cvat/apps/engine/rq.py
Marishka17: Okay, right now there is no reason to do so. In the previous implementation, I thought it would be better to check that the generated meta matches the expected structure.

cvat/apps/engine/rq.py
Marishka17: fixed

cvat/apps/engine/rq.py
SpecLad: Okay, works for me.

cvat/apps/engine/rq.py
Marishka17: Okay, I also think so, but I made this way because:
1. `ImmutableRQMetaAttribute` expects objects with a `meta` attribute and I would not prefer to use `obj.to_dict()[SomeField]`
2. in some places we need to have the "dict" representation of an user or request (`BaseRQMeta.for_job(rq_job).user.to_dict()`). In such cases, I also don't like calling `BaseRQMeta.for_job(rq_job).user.meta` because it is less informative.
So, I kept `to_dict` only for `UserMeta` and `RequestMeta`.

SpecLad: Okay, works for me.

PR #9084, opened 2025-02-10
Streaming export for YOLO and COCO formats

Eldies:
### Motivation and context

When we export a dataset, all annotations are kept in RAM. It may be a problem if they are large.
Enabling streaming export for task or job data, for yolo formats

depends on https://github.com/cvat-ai/datumaro/pull/81, https://github.com/cvat-ai/datumaro/pull/90, https://github.com/cvat-ai/datumaro/pull/94

### Checklist

### License


zhiltsov-max:
I've noticed export performance degradation in streaming mode, about 2-3x in my cases. As I understand it, the problem is in `coco/exporter.py:142`, where the regular json library is used instead of the optimized orjson. I suppose orjson isn't compatible directly with json stream used in the implementation now, but I could find that there is other relevant functionality in orjson - https://github.com/ijl/orjson?tab=readme-ov-file#fragment. Please check this issue.

cvat/apps/dataset_manager/bindings.py
```diff
@@ -1679,46 +1676,64 @@ def __init__(
...
 
+            # do not keep parsed lazy list data after this iteration
+            frame_data = frame_data._replace(
+                labeled_shapes=[
+                    (
+                        shape._replace(points=shape.points.lazy_copy())
+                        if isinstance(shape.points, LazyList) and not shape.points.is_parsed
+                        else shape
+                    )
+                    for shape in frame_data.labeled_shapes
```
zhiltsov-max, line 1719: This copying should probably include skeleton elements as well

cvat/apps/engine/lazy_list.py
```diff
@@ -268,3 +268,19 @@ def __setstate__(self, state):
         self._parsed = state["parsed"]
         if self._parsed:
             self.extend(state["parsed_elements"])
+
+    def lazy_copy(self) -> list[T]:
+        """
+        Makes a copy without parsing elements.
+        Only works if elements has not been not parsed yet.
```
zhiltsov-max, line 275:
```suggestion
        Only works if elements has not been parsed yet.
```

cvat/apps/dataset_manager/bindings.py
Eldies: As far as I understand, skeleton elements are just points? Then it seems as not really profitable to include them - size of points field is miniscule compared to all the other fields in LabeledShape, and extra computations needed to copy it, will not result in significant improvement in memory usage.

Also, I tried to export a task with 100000 skeletons with 17 elements each (on develop brunch) and 5.5 of 5.7 Gb of consumed memory is in `init_from_db`. So, not much room for improvement until I start to streamify `init_from_db`

zhiltsov-max: > As far as I understand, skeleton elements are just points?

They are single point shapes. Basically, most other shape types are also stored as points. Ok, the changes made are already significant improvement. Maybe we'll return to this later.

zhiltsov-max (approved):

PR #9090, opened 2025-02-11
Add attribute support for auto-annotation functions

SpecLad:
### Motivation and context

Remove one of the long-standing limitations on auto-annotation functions by adding the necessary validation and remapping logic to support attribute specifications and values. Add a utility module for attributes with functionality I needed, but felt didn't belong in the auto-annotation layer.

Adds the necessary code to support using functions with attributes via agents, as well. I will submit the necesssary server-side code will be submitted to the private repository later; until that is merged, attempts to create native functions with attributes will be rejected.

### How has this been tested?

Unit tests and manual testing.

### Checklist

### License


cvat-sdk/cvat_sdk/auto_annotation/driver.py
```diff
@@ -16,13 +16,39 @@
...
     name: str
+    attributes: Optional[Mapping[str, _AttributeNameMapping]] = attrs.field(
+        kw_only=True, default=None
+    )
+
+    def map_attribute(self, name: str) -> _AttributeNameMapping:
+        if self.attributes is None:
+            return _AttributeNameMapping(name)
+
+        return self.attributes.get(name)
```
zhiltsov-max, line 40: Probably, this call can return `None`, which contradicts the `map_attribute()` return type annotation.

cvat-sdk/cvat_sdk/auto_annotation/driver.py
```diff
@@ -16,13 +16,39 @@
 from cvat_sdk.core.progress import NullProgressReporter, ProgressReporter
 from cvat_sdk.datasets.task_dataset import TaskDataset
 
+from ..attributes import attribute_value_validator
 from .exceptions import BadFunctionError
 from .interface import DetectionFunction, DetectionFunctionContext, DetectionFunctionSpec
 
 
+@attrs.frozen
+class _AttributeNameMapping:
```
zhiltsov-max, line 25: Shouldn't these classes called `*Mapping` be `collections.abc.Mapping`-compatible?

cvat-sdk/cvat_sdk/auto_annotation/driver.py
```diff
@@ -78,39 +126,119 @@ def _get_expected_function_output_type(self, fun_label, ds_label):
...
+        # we'll just use exact comparisons for now.
+        if ds_attr.input_type != fun_attr.input_type:
+            raise BadFunctionError(
+                f"{attr_desc} has input type {fun_attr.input_type!r} in the function,"
+                f" but {ds_attr.input_type!r} in the dataset"
+            )
+
+        if (
+            ds_attr.input_type.value not in {"text", "checkbox"}
+            and ds_attr.values != fun_attr.values
```
zhiltsov-max, line 143: Maybe relax the requirements a little and allow reordering?

tests/python/sdk/test_attributes.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+def test_number_attribute_values_good():
+    assert number_attribute_values(0, 0, 1) == ["0", "0", "1"]
+    assert number_attribute_values(0, 10, 1) == ["0", "10", "1"]
+    assert number_attribute_values(0, 10, 10) == ["0", "10", "10"]
+    assert number_attribute_values(0, 10, 5) == ["0", "10", "5"]
+
+
+def test_number_attribute_values_bad():
```
zhiltsov-max, line 17:
```suggestion
def test_number_attribute_values_can_reject_bad_values():
```

cvat-sdk/cvat_sdk/auto_annotation/driver.py
SpecLad: Thanks, fixed.

cvat-sdk/cvat_sdk/auto_annotation/driver.py
SpecLad: Yeah, okay. Updated.

tests/python/sdk/test_attributes.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+def test_attribute_value_validator_checkbox():
+    validator = attribute_value_validator(
+        models.AttributeRequest(name="a", mutable=False, input_type="checkbox", values=[])
+    )
+
+    assert validator("true")
+    assert validator("false")
+    assert not validator("maybe")
```
zhiltsov-max, line 38: In many cases it's convenient to also support automatic value projection in validators, for instance in a checkbox validator we could accept and automatically convert the same values as in `attrs.converters.to_bool()`. Consider supporting it here as well to provide more convenient interface. I can see it could be useful for checboxes and numbers so far.

tests/python/sdk/test_attributes.py
```diff
@@ -0,0 +1,102 @@
...
+
+@pytest.mark.parametrize(
+    ["values", "exc_match"],
+    [
+        (["0", "1"], "wrong number of values"),
+        (["0", "10", "1", "1"], "wrong number of values"),
+        # the exception is raised by external code, so we don't want to rely on the message
+        (["a", "10", "1"], ""),
+        (["0", "a", "1"], ""),
+        (["0", "10", "a"], ""),
```
zhiltsov-max, line 65: Shouldn't these cases be wrapped by the sdk code?

cvat-sdk/cvat_sdk/auto_annotation/driver.py
SpecLad: Well... it's a more abstract notion of mapping. The idea is that each of the `XNameMapping` classes defines how names within `X` should be mapped, so it parallels the structure of `X`. So, for example, if you want to map `attribute.name`, you look up `attribute_nm.name` (where `attribute_nm` is an `_AttributeNameMapping` object).

Maybe they shouldn't be called `*Mapping`, but I couldn't think of anything better when I created them.

tests/python/sdk/test_attributes.py
SpecLad: I don't know. Why bother?

tests/python/sdk/test_auto_annotation.py
```diff
@@ -304,6 +346,88 @@ def detect(context, image: PIL.Image.Image) -> list[models.LabeledShapeRequest]:
...
+                                cvataa.attribute_val(2, "forward"),  # orientation
+                            ],
+                        ),
+                        # tail
+                        cvataa.keypoint(30, [30, 30]),
+                    ],
+                    attributes=[
+                        cvataa.attribute_val(1, "calico"),  # color
+                        cvataa.attribute_val(2, "McFluffy"),  # name
+                    ],
```
zhiltsov-max, line 396: I think it could be convenient to support passing attributes as a dict with attr name as a key, what do you think?

tests/python/sdk/test_auto_annotation.py
```diff
@@ -304,6 +346,88 @@ def detect(context, image: PIL.Image.Image) -> list[models.LabeledShapeRequest]:
...
+                                ),
+                                cvataa.text_attribute_spec("orientation (should be ignored)", 2),
+                            ],
+                        ),
+                        cvataa.keypoint_spec("tail", 30),
+                    ],
+                    attributes=[
+                        cvataa.select_attribute_spec("color", 1, ["gray", "calico"]),
+                        cvataa.text_attribute_spec("name (should be ignored)", 2),
+                    ],
```
zhiltsov-max: `*_attribute_spec` - probably, id should be optional. Probably, they are unique within a label.

tests/python/sdk/test_attributes.py
zhiltsov-max: It's an error message that is supposed to be understood by the user, so it's a part of the library api.

tests/python/sdk/test_attributes.py
zhiltsov-max: So far I can see 2.5 use cases:
- checkbox: allows "true", ..., "TRUE", "yes", ..., "YES", True, False, 1, 0 -> converts to "true" / "false"
- number: accepts a number or a string, converts to a string

In future, if we ever change something about the value representation, we can also add some conversion without the need to change user code.

tests/python/sdk/test_attributes.py
zhiltsov-max: Then it should probably be checked in these tests

tests/python/sdk/test_auto_annotation.py
SpecLad: > I think it would be convenient to pass attributes as a dict with attr name as a key, what do you think?

I think it's irrelevant. `detect` outputs `LabeledShapeRequest`s, and attribute values inside a `LabeledShapeRequest` are identified by IDs.

Furthermore, I'm not sure this would be a meaningful improvement. The attribute identifier (be it name or number) has to occur at least twice in a function (in the spec and in `detect`), so in a real-life function I would expect it to be factored out as a named constant. And once you have a constant, it doesn't matter if its value is a name or a number.

> The same idea is valid for labels as well

NNs usually output labels as numbers, so I think requiring names in `detect` would be a downgrade in convenience.

tests/python/sdk/test_attributes.py
```diff
@@ -0,0 +1,102 @@
+# Copyright (C) CVAT.ai Corporation
+#
+# SPDX-License-Identifier: MIT
+
+import pytest
+from cvat_sdk import models
+from cvat_sdk.attributes import attribute_value_validator, number_attribute_values
+
+
+def test_number_attribute_values_good():
```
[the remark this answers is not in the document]
SpecLad: Applied.

tests/python/sdk/test_attributes.py
SpecLad: I dunno, it seems a bit too magical to me. One of the principles of the detection function interface is that if you were to create a task based on `spec`, then the output of `detect` should consist of valid annotations for that task. "No" is not a valid value for a boolean attribute (technically I think you can still upload it, but the UI won't display it correctly...). I don't want to be in a situation where using "no" in an AA function will silently convert it to "false", but using it in, e.g. `client.tasks.api.update_annotations` results in it being uploaded as-is.

Also note that converting from non-string values wouldn't be useful (at least as far the auto-annotation API is concerned), because `AttributeValRequest` will reject non-string values upon construction. So any numbers or booleans will not survive long enough to be handed to the validator.

-----

While implementing this PR, I had an idea to add helper functions for constructing the values, which would be similar to what you're proposing here. With these helper functions, you could construct an attribute value like so:

```
cvataa.attribute_val(<id>, number_attribute_value(<val>))
```

But it seemed impractical, since it's not that hard to write `str(<val>)` or `"true" if <val> else "false"`.

tests/python/sdk/test_attributes.py
SpecLad: But we can't be sure that the message won't be changed in a future Python version.

tests/python/sdk/test_auto_annotation.py
SpecLad: > So, basically, this is the problem I'm talking about. If we make attribute ids optional in the attr spec, we can leave only names in the user code.

How are you proposing to do that? `AttributeValRequest` has a mandatory `spec_id` field (and no `spec_name` field).

> Looks significantly better than

I agree, but I don't think it's the whole picture (also, I don't think you'd need anything like `self.attr_name_map`). Here's what a full function with attributes would look like, in my view:

```python
ATTR_ID_COLOR = 1

spec = DetectionFunctionSpec(... attributes=[cvaaa.radio_attribute_spec("color", ATTR_ID_COLOR, ...)] ...)

def detect(...):
    ... attributes=[cvataa.attribute_val(ATTR_ID_COLOR, ...)] ...
```

And here's what it would look like with attribute names only:

```python
ATTR_NAME_COLOR = "color"

spec = DetectionFunctionSpec(... attributes=[cvaaa.radio_attribute_spec(ATTR_NAME_COLOR, ...)] ...)

def detect(...):
    ... attributes={ATTR_NAME_COLOR: ...} ...
```

which is nicer, sure, but I don't think it's a substantial enough improvement.

tests/python/sdk/test_attributes.py
zhiltsov-max: > I don't want to be in a situation where using "no" in an AA function will silently convert it to "false", but using it in, e.g. client.tasks.api.update_annotations results in it being uploaded as-is.

Why this is a question? The scope of the aa component should be to provide a wrapper for models with convenient user interface, not a shortcut for server api. Basically, users should never know about the server api used internally.

> "No" is not a valid value for a boolean attribute

I mean, neither are `"true"` and `"false"`, because they are strings. The user also doesn't specify the actual list of valid values for checkbox attributes.

For me, the only question is how simple it is for the user to integrate their model. Anything done behind the API is behind. If we can do lots of magic to provide better interface, let's do this.

SpecLad: > Why this is a question?

Because I want there to be a single mental model that people can use both for writing AA functions and for using the API to upload annotations directly. These two use cases are very similar, so I'd like the interfaces to be as similar as possible.

> I mean, neither are `"true"` and `"false"`, because they are strings.

What do you mean? All attribute values are strings.

tests/python/sdk/test_attributes.py
SpecLad: I'm confident that the message will remain understandable, I just can't be sure of the exact words. 🤷 

Anyway, I don't want to die on this hill, so I went ahead and added a custom message.

tests/python/sdk/test_attributes.py
zhiltsov-max: > What do you mean? All attribute values are strings.

All attribute values are strings in the server API, but they don't have to be like this in the model output parsing code. Numbers and booleans are good candidates to become first non-string values acceptable from a model, because it's their natural representation.

tests/python/sdk/test_auto_annotation.py
SpecLad: Okay, I tried to find a solution by adding a more complicated helper function. It takes a dictionary and converts it to a list of `AttributeValRequest`s. This function also does type conversions, so it should cover your concerns from another thread.

However, the attributes are still identified by number. I don't think I can support string identifiers without making everything more complicated than it's worth.

zhiltsov-max: Ok, I think we can stop at this point. It still can be improved, but it's better to do this in a separate PR.

zhiltsov-max (approved):

PR #9095, opened 2025-02-12
Fix: Cannot read properties of undefined (reading 'width') in Video GT job

klakhov:
### Motivation and context

It seems that a regression was introduced during recent refactoring related to frame metadata. In a GT job created from a video, navigating through frames results in an error: *Cannot read properties of undefined (reading 'width').
gt_error

### Checklist

### License


cvat-core/src/server-response-types.ts
```diff
@@ -460,18 +460,21 @@ export interface SerializedCloudStorage {
     manifests?: string[];
 }
 
+export interface SerializedFrameMetaData {
+    width: number;
+    height: number;
+    name: string;
+    related_files: number;
+    has_related_context: boolean;
```
bsekachev, line 468: This field is not used anyhow in client code, I do not think we need to add it

cvat-core/src/frames.ts
```diff
@@ -89,11 +89,8 @@ export class FramesMetaData {
     public includedFrames: number[] | null;
     public frameFilter: string;
     public frames: {
-        width: number;
-        height: number;
-        name: string;
-        related_files: number;
-    }[];
+        [index: number]: SerializedFrameMetaData;
```
bsekachev, line 92: Changing interface of public member in that way is a dangerous update 
I see this field may be accessed in some places, including enterprise code. I suspect all this code must be updated correspondingly. 

Do we have any other options?

bsekachev (approved):

PR #9116, opened 2025-02-18
Project quality

zhiltsov-max:
### Motivation and context

Obsoletes https://github.com/cvat-ai/cvat/pull/6367

Depends on: https://github.com/cvat-ai/cvat/pull/9275, https://github.com/cvat-ai/cvat/pull/9272

- Added quality estimation for projects
- Updated quality report contents and quality report summary server API
  - `frame_count` field - deprecated in favor of the new `validation_frames` field
  - `frame_share` field - deprecated in favor of the new `validation_frame_share` field
- Added support for job filters in quality settings. This allows to include or exclude jobs based on filter criteria
- Added support for quality estimation in consensus replicas
- Task quality checks now can use individual settings or inherit ones from the parent project
- Project quality updates now trigger task quality updates for all nested tasks. Relevant task quality reports are reused
- \[Server API\] Optimized `GET api/quality/reports`, `GET api/quality/conflicts` requests
- \[Server API\] Optimized permission checks in `api/quality/*` requests
- \[Server API\] Improved endpoint descriptions in the schema

### Checklist

### License


cvat-ui/src/components/quality-control/task-quality/jobs-filter-configuration.ts
zhiltsov-max, line 1: @klakhov,

I think, it makes sense to add some extra conditions:
- parent object filters, such as task id, task name (contains)
- job type (consensus replicas are excluded by default)

image

zhiltsov-max:
@klakhov, merged with develop, added job filters and settings filters in server API.

cvat-ui/src/components/quality-control/task-quality/jobs-filter-configuration.ts
```diff
@@ -0,0 +1,79 @@
...
+        },
+        type: {
+            label: 'Job Type',
+            type: 'select',
+            operators: ['select_equals'],
+            valueSources: ['value'],
+            fieldSettings: {
+                listValues: [
+                    { value: 'annotation', title: 'Annotation' },
+                    { value: 'ground_truth', title: 'Ground truth' },
```
zhiltsov-max, line 70:
```suggestion
```
GT jobs are not allowed for quality checks.

cvat-ui/src/components/quality-control/task-quality/jobs-filter-configuration.ts
```diff
@@ -0,0 +1,79 @@
...
+            type: 'text',
+            valueSources: ['value'],
+            operators: ['like'],
+        },
+        project_name: {
+            label: 'Project name',
+            type: 'text',
+            valueSources: ['value'],
+            operators: ['like'],
+        },
```
zhiltsov-max, line 61: As soon as we can launch quality checks no more than in just 1 project, project-level conditions seem to be extra.

cvat-ui/src/components/quality-control/task-quality/jobs-filter-configuration.ts
zhiltsov-max, line 1: + (no) job assignee

cvat-ui/src/components/quality-control/task-quality/jobs-filter-configuration.ts
klakhov: Removed

cvat-ui/src/components/quality-control/task-quality/jobs-filter-configuration.ts
klakhov: Added common filter `assignee == X`

cvat-core/src/api-implementation.ts
zhiltsov-max, line 1: Please update job_filter type handling on UI - it's now a string with json inside. If I unset the filter, I get an error `ServerError: **job_filter**: This field may not be null.`, an empty string should be sent instead.

zhiltsov-max, line 1: If there is an error while saving quality settings update for the `inherit` field, the UI shows the error, but displays the updated variant for the button switch. Probably, it should revert back to what was before the error for this switch.

cvat-ui/src/components/common/filters/job-filter-configuration-base.ts
```diff
@@ -0,0 +1,84 @@
...
+        type: 'select',
+        operators: ['select_equals'],
+        valueSources: ['value'],
+        fieldSettings: {
+            listValues: [
+                { value: '2d', title: '2D' },
+                { value: '3d', title: '3D' },
+            ],
+        },
+    },
```
zhiltsov-max, line 47: There is no point in this filter so far:
- 3d isn't supported in quality
- tasks in projects can't have different dimensions

klakhov: Removed

cvat-core/src/api-implementation.ts
klakhov: Makes sense, added reset

klakhov: When I send `job_filter: ""` I still get an error: `filter: can't parse filter expression: Expecting value: line 1 column 1 (char 0)`

zhiltsov-max: Fixed in https://github.com/cvat-ai/cvat/pull/9116/commits/b25369fd9e22166ac9f295b34cf3bc2d298c6df0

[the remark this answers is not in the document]
klakhov: Fixed

cvat/apps/quality_control/quality_reports.py
```diff
@@ -2330,19 +2666,20 @@ def schedule_custom_quality_check_job(
...
+            job_callback, job_params = self._make_rq_job_params(target=target)
+            queue.enqueue(
+                job_callback,
+                **job_params,
+                job_id=rq_id,
+                meta=BaseRQMeta.build(request=request, db_obj=target),
+                result_ttl=self._JOB_RESULT_TTL,
+                failure_ttl=self._JOB_RESULT_TTL,
+                depends_on=dependency,
+            )
```
Marishka17, line 2678: Why did you move job enqueuing outside a lock? It must be done inside a critical section. Otherwise, 2 parallel requests from the same user to compute quality reports for different resources may have the same dependency.

zhiltsov-max: It looks like an invalid automatic merge. Thanks for noticing

klakhov:
@zhiltsov-max Could you please look on failing Regal? It seems it doesnt like that rules for confilicts are too long. Maybe we can change the config for max-rule-length setting.

changelog.d/20250218_192952_mzhiltso_project_quality.md
```diff
@@ -0,0 +1,4 @@
+### Added
+
+- Quality analytics for projects
```
bsekachev, line 3: For the target user this patch does not introduce quality analytics for projects.
It only adds quality settings on a project level

cvat-core/src/api.ts
```diff
@@ -402,17 +402,21 @@ function build(): CVATCore {
                 },
             },
             quality: {
-                async reports(filter = {}) {
-                    const result = await PluginRegistry.apiWrapper(cvat.analytics.quality.reports, filter);
+                async reports(filter = {}, aggregate = false) {
```
bsekachev, line 405: missing type definition in `index.ts` for `aggregate` parameter

cvat-core/src/api-implementation.ts
```diff
@@ -430,7 +424,7 @@ export default function implementAPI(cvat: CVATCore): CVATCore {
         return new ConsensusSettings({ ...settings, descriptions });
     });
 
-    implementationMixin(cvat.analytics.quality.reports, async (filter: QualityReportsFilter) => {
+    implementationMixin(cvat.analytics.quality.reports, async (filter: QualityReportsFilter, aggregate?: boolean) => {
```
bsekachev, line 427: Let's re-use types from core definition here:
e.g.

Parameters<CVATCore['analytics']['quality']['reports]>[0],
Parameters<CVATCore['analytics']['quality']['reports]>[1]

bsekachev: Additionally let's add return types

cvat-core/src/api-implementation.ts
```diff
@@ -529,19 +523,27 @@ export default function implementAPI(cvat: CVATCore): CVATCore {
 
         return mergedConflicts;
     });
-    implementationMixin(cvat.analytics.quality.settings.get, async (filter: QualitySettingsFilter) => {
-        checkFilter(filter, {
-            taskID: isInteger,
-        });
+    implementationMixin(
+        cvat.analytics.quality.settings.get,
+        async (filter: QualitySettingsFilter, aggregate?: boolean) => {
```
bsekachev, line 528: The same comments as above about types definition

cvat-core/src/index.ts
```diff
@@ -152,7 +152,7 @@ export default interface CVATCore {
             reports: (filter: QualityReportsFilter) => Promise<PaginatedResource<QualityReport>>;
             conflicts: (filter: QualityConflictsFilter) => Promise<QualityConflict[]>;
             settings: {
-                get: (filter: QualitySettingsFilter) => Promise<QualitySettings>;
+                get: (filter: QualitySettingsFilter, aggregate?: boolean) => Promise<QualitySettings[]>;
```
bsekachev, line 155: Should it be `PaginatedResource<QualitySettings>`?

cvat-core/src/api-implementation.ts
```diff
@@ -529,19 +523,27 @@ export default function implementAPI(cvat: CVATCore): CVATCore {
...
+            if (settingsList) {
+                const schema = await getServerAPISchema();
+                const descriptions = convertDescriptions(schema.components.schemas.QualitySettings.properties);
 
-        return new QualitySettings({ ...settings, descriptions });
-    });
+                const settings = settingsList.map((setting) => new QualitySettings({ ...setting, descriptions }));
+                return settings;
+            }
+            return null;
```
bsekachev, line 545: `null` is not declared as possible return type for this method
Additionally the endpoint returns list, so, I do not think it can be null.

cvat-core/src/quality-report.ts
```diff
@@ -91,15 +105,16 @@ export default class QualityReport {
 
     get summary(): QualitySummary {
         return {
-            frameCount: this.#summary.frame_count,
-            frameSharePercent: this.#summary.frame_share * 100,
+            totalFrames: this.#summary.total_frames,
+            validationFrames: this.#summary.validation_frames,
+            validationFrameSharePercent: this.#summary.validation_frame_share * 100,
```
bsekachev, line 110: Is there any reason to convert to `percent`? 
In accuracy/precision/recall we also multiply to 100, but we do not add any `percent` prefix

I would vote to keep these names and values API-consistent.

cvat-core/src/quality-settings.ts
```diff
@@ -19,12 +20,34 @@ export enum PointSizeBase {
...
+    objectVisibilityThreshold: 'object_visibility_threshold',
+    panopticComparison: 'panoptic_comparison',
+    compareAttributes: 'compare_attributes',
+    targetMetric: 'target_metric',
+    targetMetricThreshold: 'target_metric_threshold',
+    maxValidationsPerJob: 'max_validations_per_job',
+    emptyIsAnnotated: 'empty_is_annotated',
+    jobFilter: 'job_filter',
+    inherit: 'inherit',
+};
```
bsekachev, line 43: Feels like it can be utility function to convert snake_case to camelCase
We already have `fieldsToSnakeCase` and Camelized<Type>, so, we may add reverse function and type.

cvat-core/src/quality-settings.ts
```diff
@@ -61,151 +88,171 @@ export default class QualitySettings {
         this.#panopticComparison = initialData.panoptic_comparison;
         this.#compareAttributes = initialData.compare_attributes;
         this.#emptyIsAnnotated = initialData.empty_is_annotated;
+        this.#jobFilter = initialData.job_filter || '';
+        this.#inherit = initialData.inherit;
         this.#descriptions = initialData.descriptions;
+
+        this.#updateTrigger = new FieldUpdateTrigger();
```
bsekachev, line 95: I would deprecate FieldUpdateTrigger for further using.
It leads to uncertain state of the object when user just changes the fields, but not saves them on the server, or when settings were not saved successfully because of a reason. Also update behaviour can not be modified in plugin with this approach as updateTrigger is a private member. So, many potential problems. 

IMHO API should be:
settings.update({ ..fieldsToUpdate }) 

as we implemented for jobs.

cvat-core/src/server-proxy.ts
```diff
@@ -373,6 +373,50 @@ function setAuthData(response: AxiosResponse): void {
...
+    params = {},
+): (
+        filter: FilterType,
+        aggregate?: boolean,
+    ) => Promise<PaginatedResource<SerializedDataType>> {
+    return async function fetchAggregatedResourceInner(
+        filter: FilterType,
+        aggregate = false,
+    ): Promise<PaginatedResource<SerializedDataType>> {
+        const id = (filter as any)?.id || null;
```
bsekachev, line 387:
```suggestion
        const id = (filter as any)?.id ?? null;
```

cvat-core/src/server-proxy.ts
```diff
@@ -373,6 +373,50 @@ function setAuthData(response: AxiosResponse): void {
     }
 }
 
+function fetchAggregatedResource<FilterType, SerializedDataType>(
```
bsekachev, line 376: 1. fetchAggregatedResource does not seem to be a correct name for function fetching all resources (the same about parameter `aggregate`. `aggregation` is another operation, when many values are converted to a certain value)

2. It is not really logical to handle single `id` logic in this function

3. I also do not think we should handle case when aggregate===false in this function

You probably did that to reduce possible code duplication, but I think the code seems more difficult to understand not and it is not worthy of profit.  

Why not call `fetchAll` from corresponding functions when `aggregate` (or another, more suitable name, is `true`? 

e.g. 

if (aggregate) {
    return fetchAll(url, filter);
}

Just three lines of code.

cvat-core/src/session.ts
```diff
@@ -768,7 +768,12 @@ export class Task extends Session {
     public readonly sourceStorage: Storage;
     public readonly targetStorage: Storage;
     public readonly organization: number | null;
-    public readonly progress: { count: number; completed: number };
+    public readonly progress: {
```
bsekachev, line 771: Is this change related to this pull request?

cvat-core/src/server-proxy.ts
```diff
@@ -713,33 +757,7 @@ async function createProject(projectSpec: SerializedProject): Promise<Serialized
...
-            },
-        });
-    } catch (errorData) {
-        throw generateError(errorData);
-    }
-
-    response.data.results.count = response.data.count;
-    return response.data.results;
-}
+const getTasks = fetchAggregatedResource<TasksFilter, SerializedTask>(`${config.backendAPI}/tasks`, { page_size: 10 });
```
bsekachev, line 760: `page_size` is a part of `TasksFilter`
Thus, it seems this setup filter will not have effect.

It is already in develop, I just emphasized it

cvat-ui/src/components/consensus-management-page/task-consensus/consensus-settings-form.tsx
```diff
@@ -75,7 +75,7 @@ export default function ConsensusSettingsForm(props: Readonly<Props>): JSX.Eleme
                 <Text strong>General</Text>
                 <CVATTooltip
                     title={generalTooltip}
-                    className='cvat-settings-tooltip'
+                    className='cvat-management-page-tooltip'
```
bsekachev, line 78: We still have many `cvat-settings-tooltip` in the code in this patch.

But, I think it should actually stay as is.
As not all tooltips on different management pages should follow such styles.
It is only about settings.

bsekachev: And for some cards.. 
But for color maybe just define a common class e.g. `.cvat-white-color-tooltip`?

cvat-ui/src/components/common/filters/job-filter-configuration-base.ts
```diff
@@ -0,0 +1,72 @@
+// Copyright (C) CVAT.ai Corporation
+//
+// SPDX-License-Identifier: MIT
+
+import { Fields } from '@react-awesome-query-builder/antd';
+import asyncFetchUsers from 'components/resource-sorting-filtering/request-users';
+
+export const fields: Fields = {
```
bsekachev, line 8: I am not sure we need a dedicated file

We may just use lodash.pick to get only specific filter properties from jobs page filter in quality control filter.

cvat-ui/src/components/consensus-management-page/consensus-management-page.tsx
```diff
@@ -261,7 +261,7 @@ function ConsensusManagementPage(): JSX.Element {
                 activeKey={activeTab}
                 defaultActiveKey={DEFAULT_TAB}
                 onChange={onTabKeyChange}
-                className='cvat-task-control-tabs'
+                className='cvat-tabs'
```
bsekachev, line 264: cvat-consensus-management-page-tabs

cvat-ui/src/components/quality-control/quality-control-page.tsx
```diff
@@ -411,7 +496,7 @@ function QualityControlPage(): JSX.Element {
                 activeKey={activeTab}
                 defaultActiveKey='Overview'
                 onChange={onTabKeyChange}
-                className='cvat-task-control-tabs'
+                className='cvat-tabs'
```
bsekachev, line 499: cvat-quality-control-page-tabs

cvat-ui/src/components/quality-control/quality-control-page.tsx
```diff
@@ -166,89 +208,133 @@ function QualityControlPage(): JSX.Element {
...
             dispatch(reducerActions.setError(error instanceof Error ? error : new Error('Unknown error')));
         } finally {
+            setActiveTab(getTabFromHash(supportedTabs));
             dispatch(reducerActions.setFetching(false));
         }
     };
 
-    const onSaveQualitySettings = useCallback(async (values) => {
+    const onSaveQualitySettings = useCallback(async (settingsList: QualitySettings[], onError?: () => void) => {
+        const { settings, childrenSettings } = state.qualitySettings;
```
bsekachev, line 310: `childrenSettings` is not mentioned in useCallback dependencies

cvat-ui/src/components/quality-control/quality-control-page.tsx
```diff
@@ -166,89 +208,133 @@ function QualityControlPage(): JSX.Element {
...
+                message: 'Could not save quality settings',
+                description: typeof Error === 'object' ? (error as object).toString() : '',
+            });
+            if (onError) {
+                onError();
             }
-            return settings;
-        } catch (e) {
-            return false;
+            throw error;
```
bsekachev, line 335: Why do we need both `onError` and re-throw?
3rdparty code calling this function may .catch() error or pass `onError` callback. Just need to choose something one.

cvat-ui/src/components/quality-control/quality-control-page.tsx
```diff
@@ -166,89 +208,133 @@ function QualityControlPage(): JSX.Element {
...
-                        message: 'Could not save quality settings',
-                        description: typeof Error === 'object' ? (error as object).toString() : '',
-                    });
-                    throw error;
-                } finally {
-                    dispatch(reducerActions.setQualitySettingsFetching(false));
-                }
+            dispatch(reducerActions.setQualitySettingsFetching(true));
+            const updatedSettings = await Promise.all(settingsList.map((setting) => setting.save()));
+            const updatedInstanceSettings = updatedSettings.find((setting) => setting.id === settings.id) || settings;
```
bsekachev, line 319: Is it possible do not find own settings across all updated settings? 
If yes, let's use ?? operator.
If not, why do we need "default" value here?

cvat-ui/src/components/quality-control/quality-control-page.tsx
```diff
@@ -166,89 +208,133 @@ function QualityControlPage(): JSX.Element {
...
+            const updatedChildrenSettings = childrenSettings?.map((childSetting) => {
+                const updatedSetting = updatedSettings.find((setting) => setting.id === childSetting.id);
+                return updatedSetting || childSetting;
+            }) || null;
```
bsekachev, line 323: The same two questions as above

cvat-ui/src/components/quality-control/quality-control-page.tsx
```diff
@@ -350,14 +437,7 @@ function QualityControlPage(): JSX.Element {
...
     if (instance) {
-        title = (
-            <Col className='cvat-quality-page-header'>
-                <Title level={4} className='cvat-text-color'>
-                    Quality control for
-                    <Link to={`/tasks/${instance.id}`}>{` Task #${instance.id}`}</Link>
-                </Title>
-            </Col>
-        );
+        title = setupTitle(instance);
```
bsekachev, line 440: for your information, you may just create anon function and call it immediately

(() => {
    // do some logic
    // return some jsx
})()

cvat-ui/src/components/quality-control/quality-control-page.tsx
```diff
@@ -366,39 +446,44 @@ function QualityControlPage(): JSX.Element {
...
-                            onDeleteFrames={onDeleteFrames}
-                            onRestoreFrames={onRestoreFrames}
-                        />
-                    ),
-                });
-            }
+        const isTaskWithGT = instance instanceof Task && gtJobInstance && gtJobMeta && qualitySettings;
+        const isProject = instance instanceof Project && qualitySettings;
+
+        if (isTaskWithGT && validationLayout) {
```
bsekachev, line 460:
```suggestion
        const isTaskWithGT = instance instanceof Task && gtJobInstance && gtJobMeta;
        const isProject = instance instanceof Project;

        if (isTaskWithGT && validationLayout && qualitySettings) {
```

cvat-ui/src/components/quality-control/quality-control-page.tsx
```diff
@@ -366,39 +446,44 @@ function QualityControlPage(): JSX.Element {
...
+                        qualitySettings={qualitySettings}
+                        onDeleteFrames={onDeleteFrames}
+                        onRestoreFrames={onRestoreFrames}
+                    />
+                ),
+            });
         }
 
-        if (qualitySettings) {
+        if (isTaskWithGT || isProject) {
```
bsekachev, line 478:
```suggestion
        if (qualitySettings && (isTaskWithGT || isProject)) {
```

bsekachev: Shall we also check `childrenQualitySettings`?

cvat-ui/src/components/quality-control/quality-settings-tab.tsx
```diff
@@ -2,41 +2,93 @@
...
+import React, { useCallback, useEffect, useState } from 'react';
 import Text from 'antd/lib/typography/Text';
 import Form from 'antd/lib/form';
-import notification from 'antd/lib/notification';
-import { QualitySettings } from 'cvat-core-wrapper';
+import Switch from 'antd/lib/switch';
+import { Project, QualitySettings, Task } from 'cvat-core-wrapper';
 import CVATLoadingSpinner from 'components/common/loading-spinner';
-import { formFieldsError } from 'utils/validation';
+import { Col, Row } from 'antd/lib/grid';
```
bsekachev, line 11: Please put 3rdparty imports prior internal imports

cvat-ui/src/components/quality-control/quality-settings-tab.tsx
```diff
@@ -2,41 +2,93 @@
...
-        } catch (error) {
-            notification.error({
-                message: 'Could not save quality settings',
-                description: formFieldsError(error).map((text: string): JSX.Element => <div>{text}</div>),
-                className: 'cvat-notification-save-quality-settings-failed',
+        if (settings) {
+            await setupSettingsFromForm(settings);
+            setQualitySettings([settings]);
+        }
+    }, [form, settings, setQualitySettings]);
```
bsekachev, line 74:
```suggestion
    }, [setupSettingsFromForm, settings, setQualitySettings]);
```

bsekachev: It would look more nicer like this:

```
const settings = await createSettingsFromForm();
setQualitySettings([settings]);
```

to avoid objects mutability

cvat-ui/src/components/quality-control/quality-settings-tab.tsx
```diff
@@ -2,41 +2,93 @@
...
+
+    const onInheritChange = useCallback(async (value: boolean) => {
+        if (settings) {
+            settings.inherit = value;
+            setQualitySettings([settings], () => {
+                settings.inherit = !value;
```
bsekachev, line 80: Avoiding such handlers is one of reasons why I suggested to rework qualitySettings.save method

cvat-ui/src/components/quality-control/quality-settings-tab.tsx
```diff
@@ -2,41 +2,93 @@
...
             });
         }
-    }, [form, setQualitySettings]);
+    }, [form, settings, setQualitySettings]);
```
bsekachev, line 83: How it depends on form?

cvat-ui/src/components/quality-control/quality-settings-tab.tsx
```diff
@@ -2,41 +2,93 @@
...
+
+    const [nonInheritedChildSettings, setNonInheritedChildSettings] = useState<QualitySettings[]>([]);
+    useEffect(() => {
+        if (childrenSettings) {
+            const filteredSettings = childrenSettings.filter((child) => !child.inherit);
+            setNonInheritedChildSettings(filteredSettings);
+        }
+    }, [childrenSettings]);
```
bsekachev, line 91: Do we really need state and use effect to handle it?

cvat-ui/src/components/quality-control/quality-settings-tab.tsx
```diff
@@ -48,14 +100,74 @@ function QualitySettingsTab(props: Readonly<Props>): JSX.Element | null {
         );
     }
 
+    let header: JSX.Element | null = null;
+    if (instance instanceof Task && instance.projectId !== null) {
+        header = (
+            <div className='cvat-quality-control-settings-header'>
+                <Switch checked={settings?.inherit} onChange={onInheritChange} />
```
bsekachev, line 107: `settings` may be null in case of fetching error as I understand
But should we render anything in this case if rendered elements will not work? 

E.g. if there is no settings, this switcher won't do anything. And probably other uncertain issues appear. So, maybe just not to render the component?

cvat-ui/src/components/resource-sorting-filtering/filtering.tsx
```diff
@@ -240,16 +240,18 @@ export default function ResourceFilterHOC(
                                 </div>
                             )}
                         >
-                            <Button
-                                className='cvat-quick-filters-button'
-                                type='default'
-                                onClick={() => onPredefinedVisibleChange(!predefinedVisible)}
-                            >
+                            {predefinedFilterValues ? (
```
bsekachev, line 243: May you please explain this change?
Above there is conditional rendering. If `predefinedFilters` is null, nothing will be rendered

cvat-ui/src/utils/hooks.ts
```diff
@@ -151,6 +151,28 @@ export function useResetShortcutsOnUnmount(componentShortcuts: Record<string, Ke
     }, []);
 }
 
+export function usePageSizeData(ref: any): any {
```
bsekachev, line 154: I do not really understand what problem you are trying to solve with this hook.

klakhov:
> Just created a new project which is EMPTY
> 
> image

Can you look on what is returned from server and how requests for settings  are sent in dev tools? When I create empty project I dont see this component.

> This is probably not "custom" settings. Custom are any non-default settings, from this view project settings also may be custom. Maybe use "Own" instead? image
> 
Updated to "Own"


> This field needs a tooltip to clarify why it needed image
Added tooktip info into `general` section

cvat-core/src/quality-settings.ts
klakhov: Removed this mapping, as new save method was implemented

cvat-core/src/server-proxy.ts
klakhov: Ok, added.

cvat-core/src/session.ts
klakhov: Yes, partially. We will need progress for Tasks table on Project quality page. But interface provided here is not correct. Actual correct value is [already being used](https://github.com/cvat-ai/cvat/blob/develop/cvat-core/src/session.ts#L852). So this is just type fix

cvat-ui/src/utils/instance-helper.ts
```diff
@@ -0,0 +1,27 @@
...
+    if (location.pathname.includes('projects')) {
+        return 'project';
+    }
+    if (location.pathname.includes('jobs')) {
+        return 'job';
+    }
+    return 'task';
+}
+
+export function readInstanceId(type: InstanceType): number {
```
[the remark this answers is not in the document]
klakhov: Used hooks from merged patch

cvat-ui/src/components/consensus-management-page/task-consensus/consensus-settings-form.tsx
klakhov: So, as some cards are using this too, why we need 2 different places to specify white text?
Its easier to have one class

cvat-ui/src/components/quality-control/quality-control-page.tsx
```diff
@@ -166,89 +208,133 @@ function QualityControlPage(): JSX.Element {
...
-
     const [activeTab, setActiveTab] = useState(getTabFromHash(supportedTabs));
-    const [instance, setInstance] = useState<Task | null>(null);
 
-    const initializeData = async (id: number): Promise<void> => {
+    const location = useLocation();
+    const requestedInstanceType: InstanceType = readInstanceType(location);
+    const requestedInstanceID = readInstanceId(requestedInstanceType);
+
+    const isMounted = useIsMounted();
```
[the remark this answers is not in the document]
klakhov: Removed

cvat-ui/src/utils/instance-helper.ts
```diff
@@ -0,0 +1,27 @@
+// Copyright (C) CVAT.ai Corporation
+//
+// SPDX-License-Identifier: MIT
+
+import { useParams, useLocation } from 'react-router';
+
+export type InstanceType = 'project' | 'task' | 'job';
```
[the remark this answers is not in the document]
klakhov: Used it from merged patch

cvat-ui/src/components/quality-control/quality-control-page.tsx
```diff
@@ -166,89 +208,133 @@ function QualityControlPage(): JSX.Element {
...
-            const taskInstance = await fetchTask(id);
-
-            setInstance(taskInstance);
-            try {
-                dispatch(reducerActions.setQualitySettingsFetching(true));
-                const settings = await core.analytics.quality.settings.get({ taskID: taskInstance.id });
-                dispatch(reducerActions.setQualitySettings(settings));
-            } finally {
-                dispatch(reducerActions.setQualitySettingsFetching(false));
+            switch (type) {
```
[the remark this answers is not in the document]
klakhov: Done.

cvat-ui/src/components/quality-control/quality-control-page.tsx
```diff
@@ -166,89 +208,133 @@ function QualityControlPage(): JSX.Element {
...
+            });
+        } finally {
+            dispatch(reducerActions.setQualitySettingsFetching(false));
+        }
+    };
+
+    const initializeData = async (): Promise<void> => {
+        try {
+            await receiveInstance(requestedInstanceType, requestedInstanceID);
+            await receiveSettings(requestedInstanceType, requestedInstanceID);
```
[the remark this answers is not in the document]
klakhov: Added re-throw

cvat-ui/src/components/quality-control/quality-control-page.tsx
```diff
@@ -166,89 +208,133 @@ function QualityControlPage(): JSX.Element {
...
         } catch (error: unknown) {
             dispatch(reducerActions.setError(error instanceof Error ? error : new Error('Unknown error')));
         } finally {
+            setActiveTab(getTabFromHash(supportedTabs));
```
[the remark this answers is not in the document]
klakhov: Removed

cvat-ui/src/components/quality-control/quality-control-page.tsx
klakhov: Removed callback, only re-throw left

cvat-ui/src/components/quality-control/quality-control-page.tsx
klakhov: Yes, it is possible that there wont be `instance settings` in the provided array.
Used `??`.

cvat-ui/src/components/quality-control/quality-control-page.tsx
klakhov: Used code from merged patch

cvat-ui/src/components/quality-control/quality-settings-tab.tsx
klakhov: Removed

cvat-ui/src/components/quality-control/quality-settings-tab.tsx
klakhov: Removed

cvat-ui/src/components/quality-control/quality-settings-tab.tsx
klakhov: Removed.

cvat-ui/src/components/quality-control/quality-settings-tab.tsx
klakhov: There is a conditional rendering that will render simple text if there is no settings [below](https://github.com/cvat-ai/cvat/blob/develop/cvat-ui/src/components/quality-control/quality-settings-tab.tsx#L53)

cvat-ui/src/components/resource-sorting-filtering/filtering.tsx
klakhov: It was unnecessary, removed

cvat-ui/src/components/quality-control/quality-settings-tab.tsx
klakhov: I've changed the setQualitySettings implementation. Added so now we are passing settings instance and an object with new fields. As it implemented in updateJobAsync

cvat-ui/src/components/consensus-management-page/task-consensus/consensus-settings-form.tsx
bsekachev: After adding `cvat-management-page-tooltip` we also have 2 different places where white text specified. 
Not any tooltip from any management page should have these styles, so, it is more about the name. The previous was good enough

cvat-ui/src/components/quality-control/quality-settings-tab.tsx
bsekachev: So, if request has failed, why "No quality settings found"? That does not seem like truth message

klakhov: Removed. Now component returns `null`

cvat-ui/src/components/consensus-management-page/task-consensus/consensus-settings-form.tsx
klakhov: Ok. Switched back

cvat/apps/iam/permissions.py
```diff
@@ -78,7 +84,7 @@ def build_iam_context(
         "group_name": request.iam_context["privilege"],
         "org_id": getattr(organization, "id", None),
         "org_slug": getattr(organization, "slug", None),
-        "org_owner_id": getattr(organization.owner, "id", None) if organization else None,
+        "org_owner_id": getattr(organization, "owner_id", None) if organization else None,
```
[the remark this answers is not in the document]
zhiltsov-max: Also implemented in https://github.com/cvat-ai/cvat/pull/9275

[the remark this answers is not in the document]
bsekachev: This PR should be dependent, I do not know how review several PRs changing the same lines in parallel

cvat/apps/quality_control/views.py
```diff
@@ -62,34 +66,29 @@
...
+        "task_id",
+        "project_id",
+        "severity",
+    ]
     simple_filters = set(filter_fields) - {"id"}
     lookup_fields = {
         "job_id": "report__job__id",
-        "task_id": "report__job__segment__task__id",  # task reports do not contain own conflicts
+        "task_id": "report__job__segment__task__id",  # task reports do not have own conflicts
+        "project_id": "report__job__segment__task__project__id",  # project reports do not have own conflicts
```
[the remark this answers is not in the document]
zhiltsov-max: I can see it keeps the same joins, but also adds an extra request.

cvat/apps/quality_control/views.py
```diff
@@ -155,31 +187,31 @@ class QualityReportViewSet(
...
 ):
-    queryset = QualityReport.objects.prefetch_related(
-        "job",
-        "job__segment",
-        "job__segment__task",
+    @staticmethod
+    def _add_prefetch_params(queryset: QuerySet[QualityReport]) -> QuerySet[QualityReport]:
+        return queryset.prefetch_related("assignee")
+
+    queryset = _add_prefetch_params(QualityReport.objects.get_queryset())
```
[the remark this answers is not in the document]
zhiltsov-max: No, it was only to match the function signature. `objects` is `Manager`, not a `QuerySet`.

cvat/apps/iam/permissions.py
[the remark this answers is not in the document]
zhiltsov-max: It was in the PR description

cvat/apps/quality_control/views.py
```diff
@@ -405,7 +505,41 @@ def get_queryset(self):
         queryset = super().get_queryset()
 
         if self.action == "list":
-            permissions = QualitySettingPermission.create_scope_list(self.request)
+            iam_context = None
```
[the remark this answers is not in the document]
zhiltsov-max: Thanks, should be fixed in [27c2bcd](https://github.com/cvat-ai/cvat/pull/9116/commits/27c2bcd288d312ff791dbbfeea6bc13d79dc02c7)

cvat/apps/quality_control/models.py
```diff
@@ -209,8 +247,24 @@ def choices(cls):
...
+        pass
+
+    task = models.OneToOneField(
+        Task, on_delete=models.CASCADE, related_name="quality_settings", null=True, blank=True
+    )  # OneToOneField implies unique
+    project = models.OneToOneField(
+        Project, on_delete=models.CASCADE, related_name="quality_settings", null=True, blank=True
+    )  # OneToOneField implies unique
+
+    inherit = models.BooleanField(default=True)
```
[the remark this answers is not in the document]
zhiltsov-max: I don't think `False` will be a better default. We want all new tasks to inherit by default, because it is natural and so expected. The existing tasks are handled by the migration: https://github.com/cvat-ai/cvat/blob/57d6765a8065faae96135cd2c629ae59f3ccd110/cvat/apps/quality_control/migrations/0007_project_quality.py#L71. 

If all the new tasks didn't inherit by default, users would have to do extra actions for all the tasks to enable inheritance. For instance, when a dataset with multiple subsets is imported into a project as several tasks.

> New task settings also may not inherit in 2 of 3 cases

What are these cases?

For projects this option doesn't mean anything so far, so any value is correct.

bsekachev (commented):
@klakhov 

It would be more natural to use filter value `page_size: 'all'` instead of extra `aggregate` parameter (additionally, as I wrote somewhere in comments `aggregation` is not good name to fetch all resources). I do not suggest updating it now, however I would propose to reconsider it during refactoring.

zhiltsov-max:
@bsekachev, using `page_size: all` should only be allowed in development mode. It can easily kill the server.

bsekachev:
> @bsekachev, using page_size: all should only be allowed in the development mode. It can easily kill the server. This 'all' should be handled by the UI in this case.

Yes, I just talking about cvat-core interface.

PR #9122, opened 2025-02-19
Add skeleton support for native functions

SpecLad:
### Motivation and context

All of the support code in the auto-annotation SDK layer is already there, we just need to send sublabel information when creating a native function, and validate it when running an agent.

In addition, force keypoint sublabel types to be "points", since this is how the UI creates them for tasks/projects.

### How has this been tested?

Tests will be added in the private repository.

### Checklist

### License

### DIFF
--- a/cvat-cli/src/cvat_cli/_internal/agent.py
+++ b/cvat-cli/src/cvat_cli/_internal/agent.py
@@ -147,39 +147,48 @@ class _Agent:
             label_desc = f"label {remote_label['name']!r}"
             label = labels_by_name.get(remote_label["name"])
 
-            if not label:
-                raise CriticalError(incompatible_msg + f"{label_desc} is not supported.")
+            self._validate_sublabel_compatibility(remote_label, label, incompatible_msg, label_desc)
 
-            if (
-                remote_label["type"] not in {"any", "unknown"}
-                and remote_label["type"] != label.type
-            ):
+            sublabels_by_name = {sl.name: sl for sl in getattr(label, "sublabels", [])}
+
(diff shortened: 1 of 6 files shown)


cvat-cli/src/cvat_cli/_internal/agent.py
```diff
@@ -147,39 +147,48 @@ def _validate_detection_function_compatibility(self, remote_function: dict) -> N
...
+                raise CriticalError(
+                    incompatible_msg + f"{attr_desc} has input type {remote_attr['input_type']!r},"
+                    f" but the function object declares input type {attr.input_type.value!r}."
+                )
+
+            if remote_attr["values"] != attr.values:
+                raise CriticalError(
+                    incompatible_msg + f"{attr_desc} has values {remote_attr['values']!r},"
+                    f" but the function object declares values {attr.values!r}."
+                )
```
zhiltsov-max, line 191: Is there a way for the users to get notified that their function definition uses some unknown labels or attributes? If some annotations are silently ignored, it doesn't look really convenient to me. Maybe it's better to show some warning for ignored labels and attributes on agent startup?

SpecLad: Do you mean the case where the local function has a label/sublabel/attribute that the remote function doesn't? At the moment I don't have any notifications for this case. My thinking was that since the end user only sees the remote function's labels, from their perspective everything will work as expected, even if the behind-the-scenes function object supports some extra labels.

That said, I'm not opposed to adding the warnings you propose; but I think it goes beyond the scope of this PR.

zhiltsov-max (approved):

PR #9152, opened 2025-02-26
Make the dump_objects.py script easier to use in the private repo

SpecLad:
This involves two things:

1. Make the output directory configurable.
2. Instead of hardcoding resource types, rewrite every file that already exists.

### How has this been tested?

Manual testing.

### Checklist

### License

### DIFF
--- a/tests/python/shared/utils/dump_objects.py
+++ b/tests/python/shared/utils/dump_objects.py
@@ -2,9 +2,11 @@
 #
 # SPDX-License-Identifier: MIT
 
+import argparse
 import json
 from datetime import timezone
 from http import HTTPStatus
+from pathlib import Path
 from typing import Any
 
 from config import ASSETS_DIR, get_method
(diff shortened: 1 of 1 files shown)


tests/python/shared/utils/dump_objects.py
```diff
@@ -34,41 +36,39 @@ def clean_list_response(data: dict[str, Any]) -> dict[str, Any]:
...
+        if endpoint == "annotations":
+            continue  # this will be handled at the end
+
+        response = get_method("admin1", endpoint, page_size="all")
+
+        with open(dump_path, "w") as f:
+            json.dump(clean_list_response(response.json()), f, indent=2, sort_keys=True)
+
+        if endpoint in ["jobs", "tasks"]:
+            obj = endpoint.removesuffix("s")
```
zhiltsov-max, line 60: It looks like it's better to pass some config file instead. Something like a `{ url: filename }` dict.

SpecLad: I don't think I understand your proposal. Could you elaborate?

zhiltsov-max: I the point is to extend the usability of the script, it's better to provide a way to use it with some user-passed configuration instead of hardcoding behavior for specific endpoints.

SpecLad: I don't think so. The less configuration is needed, the more usable it is.

You could say that the set of JSON dump files in the asset directory is the "configuration".

zhiltsov-max: Ok, so if the script depends on its outputs, how a new endpoint should be added?

SpecLad: So, what do you think?

zhiltsov-max: As I said, it's better to support configuration files instead. I'm not sure I've seen the current design anywhere else. The most similar thing that comes to my mind is drop-in files in /etc, but in that case most of them are either configuration files, scripts, or files with a specific format. If you want to keep the current design, at least add documentation - in the script and, probably, in `README.md` of python tests.

SpecLad: > As I said, it's better to support configuration files instead.

But why is it better?

> I'm not sure I've seen the current design anywhere else. The most similar thing that comes to my mind is drop-in files in /etc, but in that case most of them are either configuration files, scripts, or files with a specific format.

It's an ad-hoc script, so I'm not too concerned about using ad-hoc techniques. 🤷

> If you want to keep the current design, at least add documentation

Sure, added.

PR #9155, opened 2025-02-27
Fix included frames handling on UI

### LINKED ISSUE #9097: Incorrect data returned in frames meta request for a ground truth job

### Steps to Reproduce

1. Create a task with Ground Truth job with attached archive.

Content of the image corresponds to the image number in this archive). Each image resolution is random. 

E.g. this is image_10.jpg. Its real resolution is 800x900
Image

2. Open GT job and check meta response. It has resolution 1500x900. The image distorsed

Image

Of course, image name also may be not correct: 

Image

[images.zip](https://github.com/user-attachments/files/18766802/images.zip)

### Expected Behavior

_No response_

### Possible Solution

_No response_

### Context

_No response_

### Environment

```Markdown

```

zhiltsov-max:
### Motivation and context

Fixes #9097

- Added a temporary emulation in UI for the upcoming server changes in `/api/task|job/data/meta` `.frames`
- Updated related GT allocation table code
- Fixed type for the `included_frames` field in the related meta API response class

### How has this been tested?

[data_with_related_images.zip](https://github.com/user-attachments/files/19028777/data_with_related_images.zip)
[random_200.zip](https://github.com/user-attachments/files/19028782/random_200.zip)
[random_video_10s.mp4](https://github.com/user-attachments/assets/2e229289-13dc-43e5-9baa-c6361f4e312d)

### Checklist

### License

### DIFF
--- a/cvat-core/src/frames.ts
+++ b/cvat-core/src/frames.ts
@@ -208,7 +208,23 @@ export class FramesMetaData {
             // it may be a videofile or one image
             framesInfo = frameNumbers.map(() => initialData.frames[0]);
         } else {
-            framesInfo = initialData.frames;
+            framesInfo = (() => {
+                if (initialData.frames.length > 1 &&
+                    initialData.included_frames &&
+                    initialData.included_frames.length &&
+                    initialData.included_frames.length < initialData.frames.length
+                ) {
+                    // Drop placeholder frames for simple GT jobs.
+                    // This should be supported on the server side at some point.
+                    // So far, just pretend it's already implemented there.
+                    //
+                    // It's a heuristic that can accidentally drop normal frames called
+                    // "placeholder.jpg", we minimize such cases by the extra checks above.
+                    return initialData.frames.filter((f) => f.name !== 'placeholder.jpg');
+                }
+
+                return initialData.frames;
+            })();
         }
 
         Object.defineProperties(
--- a/cvat-core/src/frames.ts
+++ b/cvat-core/src/frames.ts
@@ -207,24 +207,16 @@ export class FramesMetaData {
         if (initialData.frames.length === 1) {
             // it may be a videofile or one image
             framesInfo = frameNumbers.map(() => initialData.frames[0]);
+        } else if (this.includedFrames?.length) {
+            // Only simple GT jobs have includedFrames, so this condition works only for them
+            const framesNumbers = new Set(this.includedFrames.map((dataFrameNumber: number) => (
+                Math.floor((dataFrameNumber - this.startFrame) / this.frameStep)
+            )));
+            // Frames must contain placeholders on positions out of includedFrames.
+            // That is, we can use frames with indices corresponding to frame numbers
+            framesInfo = initialData.frames.filter((frame, idx) => framesNumbers.has(idx));
         } else {
-            framesInfo = (() => {
-                if (initialData.frames.length > 1 &&
-                    initialData.included_frames &&
-                    initialData.included_frames.length &&
-                    initialData.included_frames.length < initialData.frames.length
-                ) {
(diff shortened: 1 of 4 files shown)


cvat-core/src/frames.ts
```diff
@@ -208,7 +208,23 @@ export class FramesMetaData {
             // it may be a videofile or one image
             framesInfo = frameNumbers.map(() => initialData.frames[0]);
         } else {
-            framesInfo = initialData.frames;
+            framesInfo = (() => {
+                if (initialData.frames.length > 1 &&
```
bsekachev, line 212: there is condition `initialData.frames.length === 1` above

bsekachev: The solution with frame names looks like a dirty hack
What to you think about this one?

```ts
            if (this.includedFrames?.length) {
                // only gt jobs have this.includedFrame, so this condition works only for them
                const framesIndexes = new Set(this.includedFrames.map((dataFrameNumber: number) => {
                    return Math.floor((dataFrameNumber - this.startFrame) / this.frameStep);
                }));
                framesInfo = initialData.frames.filter((_, idx) => framesIndexes.has(idx))
            } else {
                framesInfo = initialData.frames;
            }
```

cvat-core/src/frames.ts
```diff
@@ -728,7 +744,9 @@ export async function getContextImage(jobID: number, frame: number): Promise<Rec
     const meta = await frameData.getMeta();
     const requestId = frame;
     const { jobStartFrame } = frameData;
-    const { related_files: relatedFiles } = meta.frames[frame - jobStartFrame];
+    const dataFrameNumber = meta.getDataFrameNumber(frame - jobStartFrame);
+    const frameIndex = meta.getFrameIndex(dataFrameNumber);
+    const { related_files: relatedFiles } = meta.frames[frameIndex];
```
bsekachev, line 749: Seems like a fix not related to the main PR, I will propose to create a separate pull request

cvat-core/src/frames.ts
zhiltsov-max: > The solution with frame names looks like a dirty hack

Why it looks this way, if it's a constant defined by the server specifically for this?

> What to you think about this one?

The formula part `dataFrameNumber - this.startFrame` is not correct in cases with nonzero start frame and frame step. In meta `start_frame`, `stop_frame`, and `included_frames` contain input data frame numbers (absolute ones). The correct way would be to obtain the original data frame number, then find the corresponding `included_frames` index.

cvat-core/src/frames.ts
zhiltsov-max: Ok, https://github.com/cvat-ai/cvat/pull/9162

cvat-core/src/frames.ts
bsekachev: > Why it looks this way, if it's a constant defined by the server specifically for this?

Because as you mentioned, user may have files named `placeholder.jpg`, what potentially may lead to bugs
We should not consider such solutions

bsekachev: > The formula part dataFrameNumber - this.startFrame is not correct in cases with nonzero start frame and frame step

I tested it in this case, and it seems to work in cases with non-zero start frame and frame step !== 1. Not really understand from your explanation why it should not. 

> In meta `start_frame`, `stop_frame`, and `included_frames` contain input data frame numbers (absolute ones)

Yes. We want to map absolute frame number to segment frame number, using the formula above. 
Then, as for GT jobs meta returns info about all frames in the task, from 1st one, we do assumption, that frame number == frame index.

zhiltsov-max: Ok, looks good to me. Added some explanatory comments why it works.

bsekachev (approved):

PR #9166, opened 2025-03-03
Adjusted customizable logo size

bsekachev:
### Motivation and context
- Max width set to 128px instead of fixed 64px
- Logo is not resized anymore to max dimensions by CVAT

### Checklist

### License

### DIFF
--- a/cvat-ui/src/components/header/styles.scss
+++ b/cvat-ui/src/components/header/styles.scss
@@ -34,6 +34,8 @@
 
     img {
         object-fit: contain;
+        max-width: 100%;
+        max-height: 100%;
     }
 }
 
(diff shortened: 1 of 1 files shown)


cvat-ui/src/components/header/styles.scss
```diff
@@ -34,8 +34,6 @@
 
     img {
         object-fit: contain;
-        width: 100%;
```
klakhov, line 37: It seems that without a static container width and width: 100% on the image, object-fit: contain doesn't work as expected. As a result, logos larger than 128px won't be scaled down to fit but will be clipped instead.
image

bsekachev: Added max-width, max-height css props

bsekachev: However it breaks the implementation.

bsekachev: @klakhov 
Try now

klakhov (approved):
Works for me

PR #9189, opened 2025-03-10
add platform: Add ${PLATFORM:-} to docker-compose.yml

sc303030:
### Motivation and context

- I hope I get this error when I use it on my Mac, so I can add a feature.
```shell
The requested image's platform (linux/amd64) does not match 
the detected host platform (linux/arm64/v8) and no specific platform was requested 
```
-If no environmental variable is entered, it operates as it is, and if an environmental variable is entered from the Mac, it operates as the corresponding environmental variable

### How has this been tested?

I ran both commands and then tested them. based on macos
- `PLATFORM=linux/amd64 docker compose up -d` -> Success

### Checklist

https://github.com/cvat-ai/cvat/issues/9188
### License

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -112,6 +112,7 @@ services:
       cvat:
         aliases:
           - cvat-server
+    platform: ${PLATFORM:-}
 
   cvat_utils:
     container_name: cvat_utils
@@ -129,6 +130,7 @@ services:
       - cvat_logs:/home/django/logs
     networks:
       - cvat
+    platform: ${PLATFORM:-}
 
   cvat_worker_import:
     container_name: cvat_worker_import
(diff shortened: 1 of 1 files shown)


SpecLad:
It seems that there's already a built-in environment variable, `DOCKER_DEFAULT_PLATFORM`, that does the same thing.

hagar-dolev:
I think this should be at least mentioned in the doc's that this might be needed for macos installations.
adding 'export DOCKER_DEFAULT_PLATFORM=linux/amd64' as a recommendation.
I do think this could be nice not to need to change the global local platform config just for this software or alternatively to not need to change the code directly each time - as this is bad practice here.

hagar-dolev:
Additionally - simply adding the setting for default platform results in additional installation and compatibility issues, for instance:
```
Error response from daemon: image with reference apache/kvrocks:2.7.0 was found but does not provide the specified platform (linux/amd64)
```

PR #9200, opened 2025-03-12
Rework startup of container processes

azhavoro:
- Reworked supervisord config files to reduce code duplication
- No need to add new configuration files for new worker services
- [Helm] Backend pods support overriding container arguments
- Used 1 worker pod with 1 process for tests to reduce resource usage
- `utils` -> `worker_utils` (changed helm and docker compose accordingly)

### Checklist

### License


backend_entrypoint.sh
```diff
@@ -50,7 +80,24 @@ cmd_run() {
         sleep 10
     done
 
-    exec supervisord -c "supervisord/$1.conf"
+    supervisord_includes=$(_get_supervisord_includes "${@:2}")
```
SpecLad, line 83: It seems to me that, instead of accepting the list of includes as a parameter, this script should determine it automatically based on the list of workers. Otherwise, you're going to duplicate the includes everywhere the worker containers are configured (e.g. in the Compose file and the Helm chart), which could lead to desynchronization over time.

supervisord/clamav_reusable.conf
```diff
@@ -0,0 +1,5 @@
+[program:clamav-update]
+command=bash -c "if [ \"${CLAM_AV}\" = 'yes' ]; then /usr/bin/freshclam -d \
```
SpecLad, line 2: As long as you're doing this patch, perhaps you could move the handling of `CLAM_AV` to the backend entrypoint script? If `CLAM_AV` is not `yes`, then the script could avoid including this file at all.

helm-chart/test.values.yaml
```diff
@@ -12,26 +12,62 @@ cvat:
...
+          - name: CLICKHOUSE_USER
+            valueFrom:
+              secretKeyRef:
+                name: cvat-analytics-secret
+                key: CLICKHOUSE_USER
+          - name: CLICKHOUSE_PASSWORD
+            valueFrom:
+              secretKeyRef:
+                name: cvat-analytics-secret
+                key: CLICKHOUSE_PASSWORD
```
SpecLad, line 53: Can we just set these variables for all workers? I don't see a reason to micromanage them.

supervisord/server.conf
```diff
@@ -1,25 +1,5 @@
...
-logfile_backups=10          ; number of backed up logfiles
-loglevel=debug              ; info, debug, warn, trace
-pidfile=/tmp/supervisord/supervisord.pid ; pidfile location
-
-[program:clamav_update]
-startsecs=0
-command=bash -c "if [ \"${CLAM_AV}\" = 'yes' ]; then /usr/bin/freshclam -d \
-    -l %(ENV_HOME)s/logs/freshclam.log --foreground=true; fi"
+[include]
+files = reusable/supervisord.conf reusable/clamav.conf reusable/smokescreen.conf
```
SpecLad, line 2:
```suggestion
files = reusable/supervisord.conf %(ENV_CVAT_SUPERVISORD_INCLUDES)s
```

The includes have to be supplied by the entrypoint script, because clamav may not need to be included.

backend_entrypoint.sh
```diff
@@ -50,7 +93,28 @@ cmd_run() {
         sleep 10
     done
 
-    exec supervisord -c "supervisord/$1.conf"
+    supervisord_includes=""
+    postgres_app_name="cvat:$component"
+
+    if [ "$component" = "worker"  ]; then
+        if [ "$#" -eq 1 ]; then
+            fail "run worker: expected at least 1 worker name"
```
SpecLad, line 101: FWIW, they're really _queue_ names, not worker names.

helm-chart/values.yaml
```diff
@@ -67,6 +72,11 @@ cvat:
         additionalEnv: []
         additionalVolumes: []
         additionalVolumeMounts: []
+        args:
+        - "run"
+        - "worker"
+        - "export"
+        - "include=smokescreen"
```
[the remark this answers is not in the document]
SpecLad: You haven't answered me. It doesn't really make sense to me to customize, for example, the export worker's arguments - why would it be anything other than `run worker export`?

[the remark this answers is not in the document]
azhavoro: The main reason is that I don't want to add an additional component that will allow to run a worker handling all queues. I don't think it's a good idea to include new component in Helm configuration that is not needed in usual deployment. The ability to overwrite arguments allows to run such worker in an already defined deployment kind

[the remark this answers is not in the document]
SpecLad: That's fine, but:

a. you don't need to make _every_ worker customizable - utils is enough, and
b. even for utils, it only makes sense to customize the list of queues, not all arguments.

backend_entrypoint.sh
```diff
@@ -50,7 +83,29 @@ cmd_run() {
...
+    supervisord_includes=""
+    postgres_app_name="cvat:$component"
+    if [ "$component" = "server" ]; then
+        supervisord_includes=$(_get_includes "$component")
+    elif [ "$component" = "worker"  ]; then
+        if [ "$#" -eq 1 ]; then
+            fail "run worker: expected at least 1 queue name"
+        fi
+
+        worker_list="${@:2}"
```
SpecLad, line 95:
```suggestion
        queue_list="${@:2}"
```

helm-chart/test.values.yaml
```diff
@@ -12,26 +12,62 @@ cvat:
...
+      - name: CLICKHOUSE_USER
+        valueFrom:
+          secretKeyRef:
+            name: cvat-analytics-secret
+            key: CLICKHOUSE_USER
+      - name: CLICKHOUSE_PASSWORD
+        valueFrom:
+          secretKeyRef:
+            name: cvat-analytics-secret
+            key: CLICKHOUSE_PASSWORD
```
SpecLad, line 37: Why set this only for the test? I think we should just add the Clickhouse variables to `cvat.sharedBackendEnv`.

backend_entrypoint.sh
```diff
@@ -50,7 +83,29 @@ cmd_run() {
...
+
+        queue_list="${@:2}"
+        echo "Workers to run: $queue_list"
+        export CVAT_QUEUES=$queue_list
+
+        postgres_app_name+=":${queue_list// /+}"
+
+        supervisord_includes=$(_get_includes $queue_list)
+    fi
+    echo "Additional supervisor configs that will be includes: $supervisord_includes"
```
SpecLad, line 103:
```suggestion
    echo "Additional supervisor configs that will be included: $supervisord_includes"
```

SpecLad (approved):
LGTM, with one nitpick. Also, SonarCloud has some cosmetic suggestions as well, I'd suggest looking into it.

IIUC, this is tied to a private PR, so I'm leaving it to you to merge when everything is ready.

docker-compose.yml
```diff
@@ -115,16 +115,16 @@ services:
         aliases:
           - cvat-server
 
-  cvat_utils:
-    container_name: cvat_utils
```
PercipioCorey, line 119: This change breaks `docker compose docker-compose.external_db.yml up`.  Will open issue...

PR #9214, opened 2025-03-14
Helm: support adding backend environment variables from supercharts

SpecLad:
### Motivation and context

I have a need to add an environment variable for backend containers in the Enterprise version. `additionalEnv` doesn't work very well for this, because:

* It is an array, so if a user specifies a different value for it with `--set`, any defaults set by the superchart will disappear.

* It is inserted into the resource definition verbatim, so you can't refer to things like the release name.

This patch introduces a separate setting for this purpose. It takes a map, and runs the variable values through `tpl` to support interpolation.

### How has this been tested?

Manual testing.

### Checklist

### License

### DIFF
--- a/changelog.d/20250314_163649_roman_superchart_env.md
+++ b/changelog.d/20250314_163649_roman_superchart_env.md
@@ -1,0 +1,5 @@
+### Added
+
+- \[Helm\] Added a new value, `cvat.backend.extensionEnv`, to support
+  supercharts adding environment variables to backend containers
+  (<https://github.com/cvat-ai/cvat/pull/9214>)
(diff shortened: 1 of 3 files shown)


SpecLad:
@azhavoro Could you review? Note that I added a backlink from a PR in the private repository to show how I plan to use this.

azhavoro:
Please update the chart version

SpecLad:
I had already bumped it in #9202, which is going to be in the same release. So I don't think I need to do it again.

azhavoro:
> I had already bumped it in #9202, which is going to be in the same release. So I don't think I need to do it again.

Ok

azhavoro (approved):

PR #9233, opened 2025-03-20
Do not reset zoom when switching between workspaces

klakhov:
### Motivation and context

An improvement requested by the data annotation team: switching between workspaces (standard and review) caused the zoom level to reset, which was not user-friendly. This PR addresses the issue.

### Checklist

### License


bsekachev (approved):
LGTM as fix

However I would propose to get rid of triggering resize event on windows as potentially it may lead to unnecessary layout changes also in other UI elements.  

My idea here is to wrap onResize with useCallback depending on fitCanvas. 
And wrap setLayoutConfig calling this callback in zero setTimeout after setting the state. Not sure about if we really need setTimeout. As this is not related directly to this PR is up to you to implement or not.

PR #9234, opened 2025-03-20
do not keep in memory task annotations on import

Eldies:
### Motivation and context

When annotations are imported into a task, they are being kept in RAM, even though they are not used.
Stop doing it

### Checklist

### License

### DIFF
--- a/cvat/apps/dataset_manager/task.py
+++ b/cvat/apps/dataset_manager/task.py
@@ -781,10 +781,11 @@ class JobAnnotation:
 
 
 class TaskAnnotation:
-    def __init__(self, pk):
+    def __init__(self, pk, write_only: bool = False):
         self.db_task = models.Task.objects.prefetch_related(
             Prefetch('data__images', queryset=models.Image.objects.order_by('frame'))
         ).get(id=pk)
+        self._write_only = write_only
 
         # TODO: maybe include consensus jobs except for task export
         requested_job_types = [models.JobType.ANNOTATION]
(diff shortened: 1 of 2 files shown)


cvat/apps/dataset_manager/task.py
```diff
@@ -781,10 +781,11 @@ def import_annotations(self, src_file, importer, **options):
 
 
 class TaskAnnotation:
-    def __init__(self, pk):
+    def __init__(self, pk, write_only: bool = False):
         self.db_task = models.Task.objects.prefetch_related(
             Prefetch('data__images', queryset=models.Image.objects.order_by('frame'))
         ).get(id=pk)
+        self._write_only = write_only
```
SpecLad, line 788: IMO, if we implement a write-only mode, we should actually _make_ the object write-only. Otherwise, there's a risk of misuse, where we (for example) create the object with `write_only=True` and then call `export`, with questionable results.

What if we only initialized `self.ir_data` if `write_only` is false? That should prevent all attempts to read the data.

cvat/apps/dataset_manager/task.py
```diff
@@ -781,10 +781,11 @@ def import_annotations(self, src_file, importer, **options):
 
 
 class TaskAnnotation:
-    def __init__(self, pk):
+    def __init__(self, pk, write_only: bool = False):
```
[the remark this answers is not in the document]
Eldies: done

cvat/apps/dataset_manager/task.py
Eldies: done

PR #9256, opened 2025-03-25
do not keep deleted annotations in ram

Eldies:
### Motivation and context

When annotations are imported into a job, previous annotations are deleted. But after deletion they are kept in memolry for no reason, which can be a problem if they are large.
Dont keep deleted annotations in memory anymore.

### Checklist

### License

### DIFF
--- a/cvat/apps/dataset_manager/task.py
+++ b/cvat/apps/dataset_manager/task.py
@@ -489,18 +489,10 @@ class JobAnnotation:
             self.init_from_db()
             deleted_data = self.data
             models.clear_annotations_in_jobs([self.db_job.id])
-            self.reset()
         else:
             labeledimage_ids = [image["id"] for image in data["tags"]]
             labeledshape_ids = [shape["id"] for shape in data["shapes"]]
             labeledtrack_ids = [track["id"] for track in data["tracks"]]
-
-            # It is not important for us that data had some "invalid" objects
-            # which were skipped (not actually deleted). The main idea is to
-            # say that all requested objects are absent in DB after the method.
-            self.ir_data.tags = data['tags']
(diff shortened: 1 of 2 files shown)


cvat/apps/dataset_manager/task.py
```diff
@@ -761,6 +761,7 @@ def import_annotations(self, src_file, importer, **options):
             create_callback=self.create,
         )
         self.delete()
+        self.reset()
```
SpecLad, line 764: It seems to me that this call belongs in `_delete`. Is there any circumstance in which we would delete the data from the DB but wish to keep the local copy in `JobAnnotation`?

Eldies: Well, I tried to put it there first but then some tests failed, so it seems we use the local copy after delete in some cases.

https://github.com/cvat-ai/cvat/actions/runs/14061841262/job/39374864549

SpecLad: In your previous attempt, the `reset` was unconditional, which is wrong. We can only reset if _all_ annotations are deleted, rather than only some. Try putting the `reset` call into the `if data is None` branch.

Eldies: Tried it, and no tests failed this time.
Though I believe it adds inconsistency to _delete method: it's not obvious why in some cases ir_data is empty after _delete, and in some other cases it contains deleted annotations

Eldies: changed it a bit: resetting on _delete unconditionally, using return value of the method instead

SpecLad: Okay, that seems like it should work.

SpecLad (approved):

PR #9259, opened 2025-03-25
Added CVAT table component

bsekachev:
### Checklist

### License

### DIFF
--- a/cvat-ui/src/components/common/cvat-table.tsx
+++ b/cvat-ui/src/components/common/cvat-table.tsx
@@ -1,0 +1,254 @@
+// Copyright (C) CVAT.ai Corporation
+//
+// SPDX-License-Identifier: MIT
+
+import React, { useCallback, useEffect, useState } from 'react';
+import { Row, Col } from 'antd/lib/grid';
+import Table, { TableProps } from 'antd/lib/table';
+import Input from 'antd/lib/input';
+import Text from 'antd/lib/typography/Text';
+import Button from 'antd/lib/button';
+import Space from 'antd/lib/space';
+import Popover from 'antd/lib/popover';
+import Checkbox, { CheckboxChangeEvent } from 'antd/lib/checkbox';
+import { DownloadOutlined, MoreOutlined } from '@ant-design/icons';
+import { Config } from '@react-awesome-query-builder/antd';
+import jsonLogic from 'json-logic-js';
+
+import { ResourceFilterHOC, defaultVisibility } from 'components/resource-sorting-filtering';
+
+type Props = TableProps & {
+    onFilterDataSource?(data: TableProps['dataSource']): void;
+    onChangeColumnVisibility?(id: number, isHidden: boolean): void;
+    queryBuilderConfig?: Partial<Config>;
+    queryBuilderPredefinedQueries?: Record<string, string>;
+    queryBuilderCache?: {
+        key: string;
+        capacity: number;
+    };
+    csvExport?: {
+        filename: string;
+    };
+    tableTitle?: string;
+    searchDataIndex?: (string | string[])[];
+};
+
+/*
+    Enhanced table component
+    Supports:
+    * titles
+    * advanced filtration with json query
+    * overall search
+    * csv export
+    * show/hide columns
+    Current restrictions:
+    * queryBuilderConfig, queryBuilderPredefinedQueries, queryBuilderCache are supposed to be static
+*/
+function CVATTable(props: Props): JSX.Element {
+    const {
+        onChangeColumnVisibility,
+        onFilterDataSource,
+        queryBuilderConfig,
+        queryBuilderCache,
+        queryBuilderPredefinedQueries,
+        searchDataIndex,
+        tableTitle,
+        dataSource,
+        csvExport,
+        columns,
+        ...rest
+    } = props;
+
+    const [FilteringComponent, setFilteringComponent] = useState<ReturnType<typeof ResourceFilterHOC> | null>(null);
+    const [filterValue, setFilterValue] = useState<string | null>(null);
+    const [searchPhrase, setSearchPhrase] = useState<string | null>(null);
+    const [visibility, setVisibility] = useState(defaultVisibility);
+    const [filteredDataSource, setFilteredDataSource] = useState<typeof dataSource>(dataSource);
+    const [modifiedColumns, setModifiedColumns] = useState<typeof columns>([]);
+
+    const downloadCSV = useCallback(() => {
+        if (csvExport && columns) {
+            const header = columns.map((column) => ({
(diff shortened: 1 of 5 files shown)


cvat-ui/src/components/quality-control/task-quality/allocation-table.tsx
```diff
@@ -159,63 +170,33 @@ function AllocationTable(props: Readonly<Props>): JSX.Element | null {
...
+                onChange: (selectedRowKeys: Key[], selectedRows: RowData[]) => {
+                    setSelection({
+                        ...selection,
+                        selectedRowKeys,
+                        selectedRows,
+                    });
+                },
+            }}
+            size='small'
+            pagination={{ showSizeChanger: true, defaultPageSize }}
```
klakhov, line 197: We should move pagination to the center as discussed

cvat-ui/src/components/common/cvat-table.tsx
```diff
@@ -0,0 +1,274 @@
...
+    } = props;
+
+    const [FilteringComponent, setFilteringComponent] = useState<ReturnType<typeof ResourceFilterHOC> | null>(null);
+    const [filterValue, setFilterValue] = useState<string | null>(null);
+    const [searchPhrase, setSearchPhrase] = useState<string | null>(null);
+    const [visibility, setVisibility] = useState(defaultVisibility);
+    const [filteredDataSource, setFilteredDataSource] = useState<typeof dataSource>(dataSource);
+    const [modifiedColumns, setModifiedColumns] = useState<typeof columns>([]);
+
+    const downloadCSV = useCallback(() => {
```
klakhov, line 71: I think we should add a way to manipulate csvExport externally such as: rename existing column, add new colums for data rows.
It can be useful to: 1. Add a column with a link to frame/job (which is now combined in 1 column in column with ID) 2. Rename 'Actions' to 'Active' in management table for more clarity

cvat-ui/src/components/common/cvat-table.tsx
```diff
@@ -0,0 +1,274 @@
...
+            })).filter((column) => !!column.title);
+
+            let csv = '';
+            if (dataSource) {
+                const rows = dataSource
+                    .map((dataItem) => header.map(({ dataIndex }) => {
+                        let value = null;
+                        if (Array.isArray(dataIndex)) {
+                            value = dataIndex.reduce((acc: unknown, path: string) => (
+                                typeof acc === 'object' && acc ? (acc as Record<string, unknown>)[path] : null
```
klakhov, line 86: This operation seems duplicated on line 134

cvat-ui/src/components/common/cvat-table.tsx
```diff
@@ -0,0 +1,274 @@
...
+                            ), dataItem);
+                        } else {
+                            value = dataItem[dataIndex];
+                        }
+
+                        if (typeof value === 'string') {
+                            return `"${value.replace(/"/g, '""')}"`;
+                        }
```
klakhov, line 94: Why do we need such manipulation with quotes? 
Also we can use `replaceAll` so we dont need to specify global flag.

cvat-ui/src/components/common/cvat-table.tsx
```diff
@@ -0,0 +1,274 @@
...
+
+import { ResourceFilterHOC, defaultVisibility } from 'components/resource-sorting-filtering';
+
+type Props = TableProps & {
+    onFilterDataSource?(data: TableProps['dataSource']): void;
+    onChangeColumnVisibility?(id: number, isHidden: boolean): void;
+    renderExtraActions?(): JSX.Element;
+    queryBuilderConfig?: Partial<Config>;
+    queryBuilderPredefinedQueries?: Record<string, string>;
+    queryBuilderCache?: {
```
klakhov, line 26: Maybe its better to be inside of a object `query: {buiderConfig, builderPredifinedQueries, etc }`

yarn.lock
```diff
@@ -3318,10 +3318,10 @@ character-reference-invalid@^2.0.0:
   resolved "https://registry.yarnpkg.com/character-reference-invalid/-/character-reference-invalid-2.0.1.tgz#85c66b041e43b47210faf401278abf808ac45cb9"
   integrity sha512-iBZ4F4wRbyORVsu0jPV7gXkOsGYjGHPmAyv+HiHG8gi5PtC9KI2j1+v8/tlibRvjoWX027ypmG/n0HtO5t7unw==
 
-chart.js@^4.3.0:
-  version "4.4.2"
-  resolved "https://registry.yarnpkg.com/chart.js/-/chart.js-4.4.2.tgz#95962fa6430828ed325a480cc2d5f2b4e385ac31"
-  integrity sha512-6GD7iKwFpP5kbSD4MeRRRlTnQvxfQREy36uEtm1hzHzcOqwWx0YEHuspuoNlslu+nciLIB7fjjsHkUv/FzFcOg==
+chart.js@^4.4.8:
```
klakhov, line 3321: Seems like excessive change

cvat-ui/src/components/quality-control/styles.scss
```diff
@@ -119,6 +103,8 @@ $excluded-background: #d9d9d973;
 .cvat-frame-allocation-table {
     @include cvat-table-dynamic-size;
 
+    height: calc(100vh - $grid-unit-size * 56);
```
klakhov, line 106: We shouldnt add a static height to table as we have `cvat-table-dynamic-size` now, it will break flex-grow property.
Instead, we should add `flex-grow:1` to `.cvat-table-wrapper` to ensure it takes all available place. And add a padding-bottom so pagination is not at the edge

yarn.lock
bsekachev: Not really, I bumped the dependency intentionally
We may create one more pr for that

cvat-ui/src/components/common/cvat-table.tsx
bsekachev: to support quotes in any fields.
See answer here: https://stackoverflow.com/questions/17808511/how-to-properly-escape-a-double-quote-in-csv

bsekachev: also, in the past we had issues with replaceAll (its browser support low).
Maybe that is not a case anymore for us as in some cases we anyway use replaceAll now

cvat-ui/src/components/common/cvat-table.tsx
bsekachev: > I think we should add a way to manipulate csvExport externally such as: rename existing column, add new colums for data rows.

The component may be enhanced by demand
Now there is not need to implement it as it covers existing features.

klakhov: Without it we will lose the correct column name('Actions` -> `Active`) for management page table

PR #9275, opened 2025-03-27
Optimize permission checks in jobs list

zhiltsov-max:
### Motivation and context

Job serializer checks task visibility for each job, this creates extra db requests. If called for a list endpoint, extra requests are being sent for all the page elements.

Changes:
- Added prefetch for task visibility checks for the requesting user
- Added an option to reuse input object `organization` in IAM context retrieval, if available. Make this working with caching for jobs and quality reports
- Optimized out 1 extra db request to retrieve org owner in IAM context retrieval

Test setup:
Call `api/jobs/` in swagger, page_size = 100 as an org owner, ~3300 available jobs

Before:
image

After:
image

### Checklist

### License


cvat/apps/engine/serializers.py
```diff
@@ -667,16 +667,55 @@ def to_representation(self, instance):
...
         return data
 
+    @classmethod
+    def many_init(cls, *args, **kwargs):
+        if len(args) == 1 and isinstance(args[0], list):
+            from .views import JobViewSet  # avoid circular import
+
+            ctx_request = kwargs.get("context", {}).get("request")
+            ctx_view = kwargs.get("context", {}).get("view")
+            if isinstance(ctx_view, JobViewSet) and ctx_view.action == "list" and ctx_request:
```
SpecLad, line 691: Are the first two conditions here necessary? In what other circumstances would this method be called? Would it be inappropriate to use the same logic in them?

cvat/apps/engine/models.py
```diff
@@ -856,6 +859,9 @@ def _validate_constraints(self, obj: dict[str, Any]):
         ).count() != 0:
             raise TaskGroundTruthJobsLimitError()
 
+    def with_issue_counts(self):
+        return self.annotate(issues__count=models.Count('issues'))
```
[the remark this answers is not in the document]
zhiltsov-max: Previously it was
```
    .annotate(
        django_models.Count('issues', distinct=True),
    )
```

It does the same, but implicitly. This is used in the serialized to return the `issues.count` field

cvat/apps/engine/serializers.py
zhiltsov-max: I can say that Django does a 0 arg call during init, and `JobViewSet` can be uninitialized at that point. Maybe the checks can be simplified somehow, they just ensure we're in the supposed use case. We don't need them in 1-object calls, only in massive ones, like `list`.

cvat/apps/iam/permissions.py
```diff
@@ -53,8 +53,14 @@ def get_organization(request, obj):
 
             raise exc
 
+        if not org_id:
+            return None
+
         try:
-            return Organization.objects.select_related("owner").get(id=organization_id)
+            if hasattr(obj, "organization"):
```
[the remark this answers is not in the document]
zhiltsov-max: Added an explanatory comment. Example - the `Job` model prior this PR.

cvat/apps/engine/models.py
```diff
@@ -911,10 +922,14 @@ def get_task_id(self):
...
-    def organization_id(self):
+    def organization_id(self) -> int | None:
         return self.segment.task.organization_id
 
-    def get_organization_slug(self):
+    @property
+    def organization(self) -> Organization | None:
+        return self.segment.task.organization
+
+    def get_organization_slug(self) -> str | None:
```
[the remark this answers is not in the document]
zhiltsov-max: Updated

cvat/apps/iam/permissions.py
```diff
@@ -53,8 +53,17 @@ def get_organization(request, obj):
...
+            return None
+
         try:
-            return Organization.objects.select_related("owner").get(id=organization_id)
+            # If the object belongs to an organization transitively via the parent object
+            # there might be no organization field, because it has to be defined and implemented
+            # manually
+            if hasattr(obj, "organization"):
+                return obj.organization
+            else:
```
SpecLad, line 65:
```suggestion
            try:
                return obj.organization
            except AttributeError:
```

cvat/apps/engine/serializers.py
SpecLad: > We don't need this logic in 1-object calls, only in massive ones, like list.

This is the `many_init` method, which is only invoked when the serializer is created with `many=True`. So it seems like it should only get invoked in the scenario you want.

Also, I think the `len(args)` check could be avoided too by moving this logic to a more fitting place; you can give this serializer a custom `list_serializer_class`, and in that class, override `to_representation`.

zhiltsov-max: Ok, using `list_serializer_class` looks good to me.

SpecLad (approved):
LGTM. IMO, a changelog entry would be useful to add, but I'll leave that up to you.

PR #9285, opened 2025-03-28
Lambda server postprocessing

SpecLad:
### Motivation and context

With Nuclio detector functions, we have code on the server that converts the output of the function into CVAT API format annotations that is used when the user runs a function on the whole task. But we also have roughly the same code on the client that is used when the user runs a function on a frame. This is unnecessary code duplication, and it means that there could be subtle differences with how the function output is interpreted in batch vs interactive requests.

In addition, this creates an annoying situation with agents, which already create annotations in the CVAT API format. To make agents work with per-frame annotation (which I would like to do), we'd have to convert those annotations to the Nuclio format, send them to the UI, and have the UI convert them again, which is just silly.

Therefore, in this patch I make the server do conversion for both batch and interactive requests. Note that the UI doesn't quite expect the results in the API format (it wants `ObjectState`s), so a small amount of conversion logic in the UI remains; but it's much simpler than before, since the server does the bulk of the work. In the future it would be nice to refactor this to reuse the logic from `Collection.import`, but I don't feel confident enough in the frontend codebase to do it myself.

This patch also moves the responsibility of handling the "Convert masks to polygons" parameter from the client to the server.

### How has this been tested?

Manual testing with the `openvino-omz-intel-face-detection-0205`, `openvino-omz-public-mask-rcnn-inception-resnet-v2-atrous-coco` and `pth-mmpose-hrnet32` functions.

### Checklist

### License


cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/tools-control.tsx
```diff
@@ -1197,173 +1196,69 @@ export class ToolsControlComponent extends React.PureComponent<Props, State> {
...
-                                case 'checkbox':
-                                    return {
-                                        ...acc,
-                                        [attributeSpec.id as number]: `${value}`.toLowerCase() === 'true' ? 'true' : 'false',
-                                    };
-                                default:
-                                    return acc;
-                            }
-                        }, {} as Record<number, string>);
+                        return Object.fromEntries(attributes.map((a) => [a.spec_id, a.value]));
```
bsekachev, line 1201: We lost min/max/step validation for numeric attributes for interactive requests
Server only checks .isnumeric() during validation

BTW, one more potential issue in related code is: "0.5".isnumeric() returns **False**, thus such values returned from a serverless functions are ignored (e.g. confidence)

cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/tools-control.tsx
```diff
@@ -1197,173 +1196,69 @@ export class ToolsControlComponent extends React.PureComponent<Props, State> {
...
+                        // The function call endpoint doesn't support the cleanup parameter.
+                        const { cleanup, ...restOfBody } = body;
 
                         const result = await core.lambda.call(jobInstance.taskId, model, {
                             ...restOfBody, frame, job: jobInstance.id,
-                        }) as DetectedShapes;
+                        }) as DetectorResults;
+
+                        const tagStates = result.tags.map((tag) => {
+                            const jobLabel = jobInstance.labels
```
bsekachev, line 1215: `!` finding corresponding label is not really guaranteed, as it may be added after opening the job, thus, passing validation on the server and failing here

bsekachev: but maybe that is not a thing we should be worried

cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/tools-control.tsx
```diff
@@ -1197,173 +1196,69 @@ export class ToolsControlComponent extends React.PureComponent<Props, State> {
...
-                                const jobLabel = jobInstance.labels
-                                    .find((jLabel: Label): boolean => jLabel.name === data.label);
+                        const shapeStates = result.shapes.map((shape) => {
+                            const jobLabel = jobInstance.labels
+                                .find((jLabel) => jLabel.id === shape.label_id)!;
 
-                                if (!jobLabel) return null;
+                            return new core.classes.ObjectState({
+                                attributes: loadAttributes(shape.attributes),
+                                elements: shape.elements?.map?.((element) => {
```
bsekachev, line 1233:
```suggestion
                                elements: shape.elements?.map((element) => {
```

bsekachev:
Overall LGTM
Concerning moment is that deploying this patch requires updating all browsers before running any detectors. Perhaps many users do not update them often and will not understand what happened when see client errors.

cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/tools-control.tsx
SpecLad: I don't think that's possible. If the client doesn't know about a label, it will not include it in the mapping, so the server should not return any shapes with that label.

cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/tools-control.tsx
SpecLad: I don't think that's right. If `elements` is undefined, then `elements?.map` is also undefined, so calling it will throw an exception.

bsekachev: Not really
If elements is undefined, the expression will return `undefined` immediately. It will not try to go further

```
test = {
    a: {
        b: {
            
        }
    }
}
{a: {…}}
test.a.b?.c?.d?.e?.f
    
>> undefined
```

SpecLad: Oh, I see. Updated.

cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/tools-control.tsx
SpecLad: > We lost min/max/step validation for numeric attributes for interactive requests
> Server only checks .isnumeric() during validation

Ah, good point. I added range checking to the server validation logic. I also found out that the step checking was incorrect (it checked "value divisible by step" instead of "(value - min) divisible by step"), so I fixed that as well.

> BTW, one more potential issue in related code is: "0.5".isnumeric() returns False, thus such values returned from a serverless functions are ignored (e.g. confidence)

I'm pretty sure that confidence is ignored regardless (I don't even know why Nuclio functions return it...). Aside from that, I don't think attributes can have fractional values anyway, can they? Such values would fail the step check.

SpecLad:
> Concerning moment is that deploying this patch requires updating all browsers before running any detectors.

That's true. Do we have any general approach for handling such situations?

PR #9294, opened 2025-04-01
fix: explicit docker-compose read mode

marmal88:
### Motivation and context
Making the mode explicit for the container cvat_server to retrieve the external_db password would help to avoid the issue as below. link to docker defaults [link](https://docs.docker.com/reference/compose-file/services/#long-syntax-4)

Screenshot 2025-04-01 at 5 58 46 PM

### How has this been tested?
Running container with the following changes in the docker file

### Checklist
- [ x ] I submit my changes into the `develop` branch

### License

### DIFF
--- a/docker-compose.external_db.yml
+++ b/docker-compose.external_db.yml
@@ -12,7 +12,8 @@ x-backend-settings: &backend-settings
     CVAT_POSTGRES_USER:
     CVAT_POSTGRES_PASSWORD_FILE: /run/secrets/postgres_password
   secrets:
-    - postgres_password
+    - source: postgres_password
+      mode: '0444'
 
 services:
   cvat_db:


docker-compose.external_db.yml
```diff
@@ -12,7 +12,8 @@ x-backend-settings: &backend-settings
     CVAT_POSTGRES_USER:
     CVAT_POSTGRES_PASSWORD_FILE: /run/secrets/postgres_password
   secrets:
-    - postgres_password
+    - source: postgres_password
+      mode: '0444'
```
SpecLad, line 16: From your link, the default mode already seems to be 0444 - so what does this change?

marmal88: Hey @SpecLad this makes it explicit. I have this flaky experience with screenshot shown when the container is unable to access the secret location.

SpecLad: But how would making it explicit help the situation?

marmal88: I've not had issues with it after the change to make explicit.

SpecLad: Okay, I think I see what's going on. You're probably affected by docker/compose#12658. Since it's already fixed upstream, I don't think we need to work around it on our side. Until they release a new Docker Compose version, you can downgrade to 2.33.1 as a workaround.

PR #9297, opened 2025-04-02
Fix deferred jobs handling

Marishka17:
### Motivation and context

~~depends on https://github.com/cvat-ai/cvat/pull/9303~~
The following issue is also fixed:
```
  File "/opt/venv/lib/python3.10/site-packages/rq/worker.py", line 1443, in perform_job
    self.handle_job_success(job=job, queue=queue, started_job_registry=started_job_registry)
  File "/opt/venv/lib/python3.10/site-packages/rq/worker.py", line 1382, in handle_job_success
    queue.enqueue_dependents(job, pipeline=pipeline)
  File "/opt/venv/lib/python3.10/site-packages/rq/queue.py", line 1203, in enqueue_dependents
    jobs_to_enqueue = [
  File "/opt/venv/lib/python3.10/site-packages/rq/queue.py", line 1209, in <listcomp>
    and dependent_job.dependencies_are_met(
  File "/opt/venv/lib/python3.10/site-packages/rq/job.py", line 1577, in dependencies_are_met
    connection.watch(*[self.key_for(dependency_id) for dependency_id in self._dependency_ids])
  File "/opt/venv/lib/python3.10/site-packages/redis/client.py", line 2138, in watch
    return self.execute_command("WATCH", *names)
  File "/opt/venv/lib/python3.10/site-packages/redis/client.py", line 1901, in execute_command
    return self.immediate_execute_command(*args, **kwargs)
  File "/opt/venv/lib/python3.10/site-packages/redis/client.py", line 1940, in immediate_execute_command
    return conn.retry.call_with_retry(
  File "/opt/venv/lib/python3.10/site-packages/redis/retry.py", line 46, in call_with_retry
    return do()
  File "/opt/venv/lib/python3.10/site-packages/redis/client.py", line 1941, in <lambda>
    lambda: self._send_command_parse_response(
  File "/opt/venv/lib/python3.10/site-packages/redis/client.py", line 1246, in _send_command_parse_response
    return self.parse_response(conn, command_name, **options)
  File "/opt/venv/lib/python3.10/site-packages/redis/client.py", line 2060, in parse_response
    result = Redis.parse_response(self, connection, command_name, **options)
  File "/opt/venv/lib/python3.10/site-packages/redis/client.py", line 1286, in parse_response
    response = connection.read_response()
  File "/opt/venv/lib/python3.10/site-packages/redis/connection.py", line 905, in read_response
    raise response
redis.exceptions.ResponseError: wrong number of arguments for 'watch' command
```
How can it be tested manually?
Follow these steps  on the develop branch and here:
1. set `ONE_RUNNING_JOB_IN_QUEUE_PER_USER` env
1. be sure that the export process takes some time, e.g, run sleep inside the `export` func
1. initiate 3 exports consistently (X1, X2, X3)
1. cancel X2 when X1 is running: POST /api/requests/rq_id/cancel
1. initiate X2 export one more time (be sure that X1 is still running)
1. check that X1 is finished successfully and `redis.exceptions.ResponseError: wrong number of arguments for 'watch' command` is not raised
 
### Checklist

### License


Marishka17:
Upstream fix: https://github.com/rq/rq/pull/2241

cvat-sdk/cvat_sdk/core/proxies/tasks.py
```diff
@@ -156,7 +157,7 @@ def upload_data(
                     break
 
                 elif status.value == models.RequestStatus.allowed_values[("value",)]["FAILED"]:
-                    raise exceptions.ApiException(status=status, reason=message, http_resp=response)
+                    raise BackgroundRequestException(message)
```
zhiltsov-max, line 160: Consider replacing this code block with a `self.client.wait_for_completion()` call and making this change in that function instead.

cvat-sdk/cvat_sdk/core/exceptions.py
```diff
@@ -13,3 +13,9 @@ class InvalidHostException(CvatSdkException):
 
 class IncompatibleVersionException(CvatSdkException):
     """Indicates server and SDK version mismatch"""
+
+
+class BackgroundRequestException(Exception):
```
zhiltsov-max, line 18: Shouldn't it be inherited from `CvatSdkException` or `ApiException`?

cvat/rq_patching.py
```diff
@@ -77,7 +80,72 @@ def custom_started_job_registry_cleanup(self, timestamp: Optional[float] = None)
     return job_ids
 
 
-def update_started_job_registry_cleanup() -> None:
+def cancel_job(self: Job, pipeline: Pipeline | None = None, enqueue_dependents: bool = False):
```
zhiltsov-max, line 83: It makes sense to add a comment about what's changed, why, and when it can be removed from the codebase.

cvat/rq_patching.py
```diff
@@ -77,7 +80,72 @@ def custom_started_job_registry_cleanup(self, timestamp: Optional[float] = None)
...
+                # exception as it is the responsibility of the caller to
+                # handle it
+                raise
+
+
+def patch_rq() -> None:
     # don't forget to check if the issue https://github.com/rq/rq/issues/2006 has been resolved in upstream
     assert VERSION == "1.16.0"
     rq.registry.StartedJobRegistry.cleanup = custom_started_job_registry_cleanup
+    rq.job.Job.cancel = cancel_job
```
zhiltsov-max, line 151: Probably, the comment from the line 148 is not related to the `cancel` patching, and `cancel` needs it's own version check.

cvat-sdk/cvat_sdk/core/proxies/tasks.py
Marishka17: I thought about this, but did not want to refactor this code in the current PR. Okay, I'll do it.

cvat/settings/base.py
```diff
@@ -762,3 +762,13 @@ class CVAT_QUEUES(Enum):
...
     "subtitle": "Open Data Annotation Platform",
 }
+
+if ONE_RUNNING_JOB_IN_QUEUE_PER_USER:
+    PERIODIC_RQ_JOBS.append(
+        {
+            "queue": CVAT_QUEUES.CLEANING.value,
+            "id": "cleanup_deferred_job_registry",
+            "func": "cvat.apps.redis_handler.cron.cleanup_deferred_job_registry",
+            "cron_string": "0 8 * * *",
```
zhiltsov-max, line 772: Do you have information how often we get such errors? Shouldn't this job be executed more often (e.g. every 1 or 2 hours) to reduce maximum waiting time for users?

tests/python/rest_api/test_tasks.py
```diff
@@ -5794,7 +5796,7 @@ def _init_tasks(cls):
 
             cls.data[key] = (task, dataset_file)
 
-    @pytest.mark.skip("Fails sometimes, needs to be fixed")
+    # @pytest.mark.skip("Fails sometimes, needs to be fixed")
```
zhiltsov-max, line 5799: Should this comment be removed now?

cvat/rq_patching.py
Marishka17: I believe a link to the PR with description is enough

cvat/rq_patching.py
Marishka17: > cancel needs its own version check

Not exactly, but I’ve added a comment with a link to the upstream fix

cvat/settings/base.py
Marishka17: > Do you have information how often we get such errors?

I would say very rarely, so 1 run per day is a good start. In any case, we can increase the number of runs when needed.

tests/python/rest_api/test_tasks.py
Marishka17: thanks

cvat/apps/engine/background.py
```diff
@@ -91,15 +91,17 @@ def handle_existing_rq_job(
...
 
-        if rq_job_status in {RQJobStatus.STARTED, RQJobStatus.QUEUED}:
+        if rq_job_status in {
+            RQJobStatus.STARTED,
+            RQJobStatus.QUEUED,
+            RQJobStatus.DEFERRED,
+        }:
             return Response(
-                data="Export request is being processed",
+                RqIdSerializer({"rq_id": rq_job.id}).data,
```
zhiltsov-max, line 100: Doing so for `started` jobs is probably not entirely correct, as the export results in this case might be outdated. I don't think it's a big issue so far to be fixed immediately, but it makes sense to add a comment about this decision. Our export implementation doesn't lock the whole object during the export, so there are multiple ways to get outdated exported data.

changelog.d/20250507_101308_maria_fix_handling_deferred_jobs.md
zhiltsov-max, line 1: 1. `wrong number of arguments for 'watch' command` seems to be fixed. I reproduced it on develop and haven't seen such errors during testing on this PR.
2. I still could reproduce the issue with multiple jobs (more than 2) from 1 user being queued / started in parallel, even without canceling any jobs. With canceling, in turn, I could get 3 more API issues:
  2.1. It's possible to get `HTTP response body: {"detail":"You don't have permission to perform this action"}` when trying to retrieve a request if it was cancelled
  2.2. It's possible to get empty status in a cancelled request
  2.3. It's possible to get an RQ job parsing error `rq.exceptions.NoSuchJobError: Unexpected job format: {'last_heartbeat': b'2025-05-08T15:40:48.317914Z', 'status': b'started', 'started_at': b'2025-05-08T15:40:48.317914Z', 'worker_name': b'390dc20f6f8b40bca0c37fdaa733401d'}` during retrieveing a cancelled request

I don't think the newly found errors should be addressed in this PR. I'm not sure if the issue with jobs from one user was supposed to be fixed, but if so, it seems like it deserves a separate PR and a related deep investigation.

Several insights on 2: 
- I tested repeated parallel export requests in a loop for 3 jobs, and, probably, I could find all the known errors we met, even with 1 server and 2 export workers.
- It's possible to get 2 jobs in the started state (or 1 started + 1 queued) from 1 user, because of the operation order during job finish in the RQ workers. They first put dependencies into the queue and only then report finishing to RQ. This is a kind of a false positive case.
- It's possible to get multiple jobs from 1 user in the started state if the job fails because the worker has failed or there was an exception in the job. In this case, the started job is kept for some time, then becomes abandoned. Also a false positive.
- If the job takes some time to be completed, errors can be reproduced much more reliably.
- The job enqueueing in the dataset export endpoint seems to be correct overall.
- I still could get hanging deferred jobs, even if their dependency has completed and there was no loop. I'm not sure if the added periodic job could resolve this problem entirely.
- The error seems to happen only during job finishes with multiple workers. Export works fine in most cases, I couldn't reliably pollute the export queue.

Hypotheses:
- The job dependency analysis can potentially be an issue, I couldn't discard this hypothesis so far.
  - There is the call to `is_deferred`, which does a status refresh. Changing this to `get_status()` doesn't solve the problem immediately.
  - Maybe dependencies can be missed if they change state during job dependency analysis, the code doesn't get notified about this.
  - Maybe there is some logic conflict with dependency pushing into the queue by RQ workers and jobs with same names

For further analysis, I'd enable job status logging from each worker (add a job UUID into meta for tracking different jobs with same rq id), then find if we really have cases where several jobs are incorrectly processed in parallel (taking into account the order of job finish operations in RQ workers). Currently, I was checking it manually in django queues in the admin panel, and it has some delay.

Marishka17: > With canceling, in turn, I could get 3 more API issues:

I've created a separate issue for this

> I still could get hanging deferred jobs, even if their dependency has completed and there was no loop. I'm not sure if the added periodic job could resolve this problem entirely.

Could you please comment on why you have such doubts?

> For further analysis, I'd enable job status logging from each worker (add a job UUID into meta for tracking different jobs with same rq id), then find if we really have cases where several jobs are incorrectly processed in parallel (taking into account the order of job finish operations in RQ workers). Currently, I was checking it manually in django queues in the admin panel, and it has some delay.

This also deserves a separate issue/PR

zhiltsov-max: > Could you please comment on why you have such doubts?

I just didn't test calling the new cleanup function for that situation. I double checked it now, it seems to be working.

zhiltsov-max (approved):

PR #9298, opened 2025-04-02
Merge the `_xxx_request` helper methods in unit tests

SpecLad:
### Motivation and context

There are two different versions of these methods, in `ApiTestBase` and in `_LambdaTestCaseBase`. Since the latter inherits from the former, Pylint warns that the overridden methods are incompatible (and they are!). To fix this, merge these two sets of methods into one, with all needed capabilities.

Some notes on the decisions I made WRT the merged methods:

1. In `_post_request` I emulated the `query_params` parameter from Django 5.1.

2. In `_get_request` I renamed `data` to `query_params` for consistency.

3. In `_patch_request` I decided not to add a `query_params` parameter for now, since it wasn't needed. While some lambda tests called `_patch_request` with an org ID, this parameter was unnecessary, so I just removed it.

4. I set the default format to "json" everywhere, since CVAT's API is mostly JSON-based.

### How has this been tested?

It tests itself. 🙂 

### Checklist

### License

### DIFF
--- a/cvat/apps/engine/tests/utils.py
+++ b/cvat/apps/engine/tests/utils.py
@@ -11,7 +11,7 @@ from contextlib import contextmanager
 from io import BytesIO
 from pathlib import Path
 from typing import Any, Callable, TypeVar
-from urllib.parse import quote
+from urllib.parse import urlencode
 
 import av
 import django_rq
@@ -128,9 +128,9 @@ class ApiTestBase(APITestCase):
         super().setUp()
         self.client = self.client_class()
 
-    def _get_request(self, path: str, user: str, *, data: dict[str, Any] | None = None) -> Response:
+    def _get_request(self, path: str, user: str, *, query_params: dict[str, Any] | None = None) -> Response:
         with ForceLogin(user, self.client):
-            response = self.client.get(path, data=data)
+            response = self.client.get(path, data=query_params)
         return response
 
     def _delete_request(self, path: str, user: str):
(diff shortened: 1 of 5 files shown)


cvat/apps/engine/tests/utils.py
```diff
@@ -128,24 +128,49 @@ def setUp(self):
...
+        self,
+        path: str,
+        user: str,
+        *,
+        format: str = "json",  # pylint: disable=redefined-builtin
+        query_params: dict[str, Any] = None,
+        data: dict[str, Any] | None = None
+    ):
+        if query_params:
+            # Note: once we upgrade to Django 5.1+, this should be changed to pass query_params
```
Marishka17, line 151: I guess we need to raise an exception here if the django version is 5.1+, because otherwise this comment will be forgotten (since [changes](https://github.com/django/django/blob/5773bc9cf929f4648da30cb9c472b816bb4db771/django/test/client.py#L674) are backward compatible).

Marishka17 (approved):

cvat/apps/engine/tests/utils.py
SpecLad: I don't really want to force this change to be done _immediately_ upon upgrading Django. Upgrading can be complicated, so I'd rather not pile more work on whoever's doing it.

The note is just a reminder in case anyone touches this in the future. If it's forgotten, it's not a big deal. The `query_params` emulation is not broken, it's just redundant.

PR #9303, opened 2025-04-03
Fix `test_tasks.py::TestImportTaskAnnotations::test_check_import_error_on_wrong_file_structure`

Marishka17:
### Motivation and context

There were several issues with this test, like:
- the archive being created was not closed
- exception capture details were never checked
- not all formats return expected error format
### Checklist

### License

### DIFF
--- a/cvat/apps/dataset_manager/formats/cvat.py
+++ b/cvat/apps/dataset_manager/formats/cvat.py
@@ -1674,8 +1674,8 @@ def _import(src_file, temp_dir, instance_data, load_data_callback=None, **kwargs
     if is_zip:
         zipfile.ZipFile(src_file).extractall(temp_dir)
 
+        detect_dataset(temp_dir, format_name="cvat", importer=_CvatImporter)
         if isinstance(instance_data, ProjectData):
-            detect_dataset(temp_dir, format_name="cvat", importer=_CvatImporter)
             dataset = Dataset.import_from(temp_dir, "cvat", env=dm_env)
             if load_data_callback is not None:
                 load_data_callback(dataset, instance_data)
--- a/tests/python/rest_api/test_tasks.py
+++ b/tests/python/rest_api/test_tasks.py
@@ -5628,38 +5628,38 @@ class TestImportTaskAnnotations:
 
         assert compare_annotations(original_annotations, updated_annotations) == {}
 
-    @pytest.mark.parametrize(
-        "format_name",
+    @parametrize(
+        "format_name, specific_info_included",
         [
-            # FIXME: RejectionReason.detection_unsupported
-            # "COCO 1.0",
-            # "COCO Keypoints 1.0",
-            "CVAT 1.1",
-            "LabelMe 3.0",
-            "MOT 1.1",
-            "MOTS PNG 1.0",  # RejectionReason.unmet_requirements, but specific requirement information unavailable
-            # FIXME: looks like it desn't fail
-            # "PASCAL VOC 1.1",
-            "Segmentation mask 1.1",
-            "YOLO 1.1",
-            "WiderFace 1.0",
-            "VGGFace2 1.0",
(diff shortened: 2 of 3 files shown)


cvat/apps/dataset_manager/formats/cvat.py
```diff
@@ -1674,8 +1674,8 @@ def _import(src_file, temp_dir, instance_data, load_data_callback=None, **kwargs
     if is_zip:
         zipfile.ZipFile(src_file).extractall(temp_dir)
 
+        detect_dataset(temp_dir, format_name="cvat", importer=_CvatImporter)
```
Marishka17, line 1677: I don't see the reason why we should check the uploaded file structure only when importing into a project

tests/python/rest_api/test_tasks.py
```diff
@@ -5631,21 +5631,23 @@ def test_can_import_datumaro_json(self, admin_user, tasks, dimension):
...
+            # "COCO 1.0",
+            # "COCO Keypoints 1.0",
             "CVAT 1.1",
             "LabelMe 3.0",
             "MOT 1.1",
-            "MOTS PNG 1.0",
-            "PASCAL VOC 1.1",
+            "MOTS PNG 1.0",  # RejectionReason.unmet_requirements, but specific requirement information unavailable
+            # FIXME: looks like it desn't fail
+            # "PASCAL VOC 1.1",
```
SpecLad, line 5642:
```suggestion
            pytest.param("PASCAL VOC 1.1", marks=pytest.mark.xfail),
```

This will make sure that if the behavior is fixed, the test is updated.

tests/python/rest_api/test_tasks.py
```diff
@@ -5631,21 +5631,23 @@ def test_can_import_datumaro_json(self, admin_user, tasks, dimension):
...
             "Segmentation mask 1.1",
             "YOLO 1.1",
             "WiderFace 1.0",
             "VGGFace2 1.0",
-            "Market-1501 1.0",
+            "Market-1501 1.0",  # RejectionReason.unmet_requirements, but specific requirement information unavailable
```
SpecLad, line 5647: These comments are liable to become outdated. IMO, if you care _why_ the dataset is rejected, you should explicitly check it in the test. (And if not, then you shouldn't need the comments.)

tests/python/rest_api/test_tasks.py
```diff
@@ -5667,16 +5669,20 @@ def test_check_import_error_on_wrong_file_structure(self, tasks_with_shapes, for
...
+                zip_file.write(self.tmp_dir / path, path)
+
         task = self.client.tasks.retrieve(task_id)
 
         with pytest.raises(exceptions.ApiException) as capture:
             task.import_annotations(format_name, source_archive_path)
 
-            assert b"Check [format docs]" in capture.value.body
-            assert b"Dataset must contain a file:" in capture.value.body
+        error_message = str(capture.value)
```
SpecLad, line 5681:
```suggestion
        error_message = capture.value.body.decode()
```

IMO, we should make sure that the error is in the place where we expect it to be.

cvat/apps/dataset_manager/formats/cvat.py
SpecLad: The change looks reasonable to me, but since it's user-facing, you should add a changelog entry.

tests/python/rest_api/test_tasks.py
Marishka17: You are right, I'll fix it. I’ve changed the logic of the SDK `Client.wait_for_completion` method to raise a custom-defined exception instead of `ApiException`. It seemed incorrect to me that `ApiException` was raised with a 200 status code (but a 200 status code is the desired status in the requests API, even when a background job fails). That's why `.body` is gone.

SpecLad (approved):

PR #9320, opened 2025-04-11
Re-use common env variables in dev env

bsekachev:
### Motivation and context
Webpack uses `CVAT_UI_HOST` and `CVAT_UI_PORT`
Server uses `UI_HOST` and `UI_PORT`.

This patch removes the duplication. 

Additionally webpack dev server may be initialized by `CVAT_HOST` now. 

### Checklist

### License

### DIFF
--- a/cvat-ui/package.json
+++ b/cvat-ui/package.json
@@ -5,7 +5,7 @@
   "main": "src/index.tsx",
   "scripts": {
     "build": "webpack --config ./webpack.config.js",
-    "start": "webpack serve --env API_URL=http://localhost:7000 --config ./webpack.config.js --mode=development",
+    "start": "webpack serve --env API_URL=\"http://${CVAT_HOST:-'localhost'}:7000\" --config ./webpack.config.js --mode=development",
     "type-check": "tsc --noEmit",
     "type-check:watch": "yarn run type-check --watch",
     "lint": "eslint './src/**/*.{ts,tsx}'",
--- a/cvat/settings/development.py
+++ b/cvat/settings/development.py
@@ -19,14 +19,14 @@ ALLOWED_HOSTS.append("testserver")
 SENDFILE_BACKEND = "django_sendfile.backends.development"
 
 # Cross-Origin Resource Sharing settings for CVAT UI
-UI_SCHEME = os.environ.get("UI_SCHEME", "http")
-UI_HOST = os.environ.get("UI_HOST", "localhost")
-UI_PORT = os.environ.get("UI_PORT", 3000)
+CVAT_UI_SCHEME = os.environ.get("CVAT_UI_SCHEME", "http")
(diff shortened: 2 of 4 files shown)


cvat/settings/development.py
```diff
@@ -19,14 +19,14 @@
 SENDFILE_BACKEND = "django_sendfile.backends.development"
 
 # Cross-Origin Resource Sharing settings for CVAT UI
-UI_SCHEME = os.environ.get("UI_SCHEME", "http")
-UI_HOST = os.environ.get("UI_HOST", "localhost")
-UI_PORT = os.environ.get("UI_PORT", 3000)
+CVAT_UI_SCHEME = os.environ.get("CVAT_UI_SCHEME", "http")
+CVAT_UI_HOST = os.environ.get("CVAT_UI_HOST", "localhost")
+CVAT_UI_PORT = os.environ.get("CVAT_UI_PORT", 3000)
```
SpecLad, line 24: The environment variable renaming looks good to me, but I think the Python variables should keep their old name. We usually don't use the `CVAT_` prefix for setting names (since they all belong to CVAT).

cvat-ui/package.json
```diff
@@ -5,7 +5,7 @@
   "main": "src/index.tsx",
   "scripts": {
     "build": "webpack --config ./webpack.config.js",
-    "start": "webpack serve --env API_URL=http://localhost:7000 --config ./webpack.config.js --mode=development",
+    "start": "webpack serve --env API_URL=\"http://${CVAT_HOST:-'localhost'}:7000\" --config ./webpack.config.js --mode=development",
```
SpecLad, line 8: When is the backend not on localhost (in a development environment)?

bsekachev: I setup dev env on VPS and trying to adjust settings to this case

bsekachev: If leaving API_URL as `localhost`, it generally works, but at the same time leads to some issues:
e.g. logo URL or dataset export URLs returned with `localhost:7000/../..` what will not work in case with remote development

cvat/settings/development.py
bsekachev: As these vars are development specific, it is not really convenient to have just `UI_HOST` and `UI_PORT` as they are quite abstract for shared environment. 

> We usually don't use the CVAT_ prefix for setting names

I would not say about "usual" here.

- CVAT_SERVERLESS
- CVAT_POSTGRES_PASSWORD
- CVAT_DATASET_CACHE_TTL
- CVAT_VERSION
- CVAT_NUCLIO_HOST

and many more may be found.

SpecLad: Of these, only `CVAT_BASE_URL` and `CVAT_HOST` are setting names (and I'd argue they shouldn't be...).

bsekachev: Oh, you told only about setting names
I though about env names

SpecLad (approved):

PR #9324, opened 2025-04-11
Logging unhandled rejection, refactored exception logger code

bsekachev:
### Motivation and context
Now only errors from "sync" stacks are caught and logged.
However we also have sometimes unhandled rejected promises which are never logged.
It is a good signal for us that something does not work as expected and this is why I would like to log them. 

### Checklist

### License


cvat-ui/src/cvat-logger.ts
```diff
@@ -1,12 +1,59 @@
...
+import ErrorStackParser from 'error-stack-parser';
 import { getCore } from 'cvat-core-wrapper';
+import { getCVATStore } from 'cvat-store';
+import { CombinedState } from 'reducers';
 
 const core = getCore();
 const { logger } = core;
 const { EventScope } = core.enums;
 
+export function logError(error: unknown, save: boolean, extras = {}): void {
```
klakhov, line 15: It seems that `type` is now always specified in `extras`, so we could consider adding type definitions for this parameter.
Or maybe add a default `type: error` if its not specified

PR #9345, opened 2025-04-21
Fix 500 status code returned when an API method is not allowed

Marishka17:
### Motivation and context

By default, DRF returns a 405 status code when an unexpected API method is used for an existing URL. However, the code that handles this is not called since permissions are checked before that, and our code did not assume that unsupported actions could be used.

### How has this been tested?

Manually
### Checklist

### License

### DIFF
--- a/cvat/apps/iam/permissions.py
+++ b/cvat/apps/iam/permissions.py
@@ -9,7 +9,6 @@ import importlib
 import operator
 from abc import ABCMeta, abstractmethod
 from collections.abc import Sequence
-from contextlib import suppress
 from enum import Enum
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Optional, TypeVar
--- a/cvat/apps/iam/permissions.py
+++ b/cvat/apps/iam/permissions.py
@@ -113,14 +113,19 @@ class OpenPolicyAgentPermission(metaclass=ABCMeta):
 
     @classmethod
     def get_scopes(cls, request: ExtendedRequest, view: ViewSet, obj: Any):
-        with suppress(KeyError):
+        # rest_framework.viewsets.ViewSetMixin.initialize_request implementation
+        if view.action is None:
+            return view.http_method_not_allowed(request)
+
+        try:
             scopes = cls._get_scopes(request, view, obj)
-
             # prevent code bugs when _get_scope defines scopes "softly"
-            if all(scopes):
-                return scopes
-
-        return view.http_method_not_allowed(request)
+            assert all(scopes)
+            return scopes
+        except KeyError:
+            assert (
+                False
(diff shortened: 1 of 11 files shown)


cvat/apps/iam/permissions.py
```diff
@@ -100,6 +106,22 @@ class OpenPolicyAgentPermission(metaclass=ABCMeta):
...
     obj: Optional[Any]
 
+    @classmethod
+    @abstractmethod
+    def _get_scopes(cls, request: ExtendedRequest, view: ViewSet, obj: Any) -> list:
+        """Method to override to define scopes based on the request"""
+
+    @classmethod
+    def get_scopes(cls, request: ExtendedRequest, view: ViewSet, obj: Any):
+        with suppress(KeyError):
```
SpecLad, line 116: The problem with this approach is that it will also suppress the exception when a view _does_ support the method, but the corresponding permission class doesn't define a scope for it. In this case, the code is broken, so we _should_ propagate the exception (and consequently return a 500 error).

Do you think you could implement this behavior?

cvat/apps/iam/permissions.py
```diff
@@ -100,6 +106,22 @@ class OpenPolicyAgentPermission(metaclass=ABCMeta):
...
+            scopes = cls._get_scopes(request, view, obj)
+
+            # prevent code bugs when _get_scope defines scopes "softly"
```
SpecLad, line 119: What do you mean by "softly"?

cvat/apps/iam/permissions.py
Marishka17: I think so

cvat/apps/iam/permissions.py
Marishka17: image

SpecLad: Ah, okay. In that case, `assert all(scopes)` seems more appropriate, since returning a `None` scope is a bug.

cvat/apps/iam/permissions.py
```diff
@@ -100,6 +105,27 @@ class OpenPolicyAgentPermission(metaclass=ABCMeta):
...
+    @classmethod
+    @abstractmethod
+    def _get_scopes(cls, request: ExtendedRequest, view: ViewSet, obj: Any) -> list:
+        """Method to override to define scopes based on the request"""
+
+    @classmethod
+    def get_scopes(cls, request: ExtendedRequest, view: ViewSet, obj: Any):
+        # rest_framework.viewsets.ViewSetMixin.initialize_request implementation
+        if view.action is None:
+            return view.http_method_not_allowed(request)
```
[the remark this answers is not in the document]
Marishka17: :scream:

SpecLad (approved):

PR #9346, opened 2025-04-21
Fix testing formats in TaskAnnotationAPI test suite

archibald1418:
Removed an erroneous break statement after the for-loop looping through dataset export formats.
Fixed the failures with annotations and proposed fixes for some bugs. 

### How has this been tested?
Failures with annotations: incorrect type names, label attributes and points etc.
Fixed minor issue in handling the 'source' attribute on import, updated changelog
Added skeletons to the test
Removed redundant checks and annotations for formats that convert polygons to masks
Included formats previously left unchecked
Provided descriptions of bugs for review
Minor improvements in logging

Filed format issues to Datumaro:
- https://github.com/cvat-ai/datumaro/issues/99
- https://github.com/cvat-ai/datumaro/issues/103

### Checklist

### License


cvat/apps/dataset_manager/bindings.py
```diff
@@ -2193,6 +2192,7 @@ def import_dm_annotations(dm_dataset: dm.Dataset, instance_data: Union[ProjectDa
         'yolo_ultralytics_oriented_boxes',
         'yolo_ultralytics_pose',
     ]
+    import_source = str(SourceType.FILE) # safe to assume this on any import from a file
```
SpecLad, line 2195: I can't say it's wrong... but why was this change needed?

tests/cypress/e2e/features/consensus.js
```diff
@@ -214,7 +214,7 @@ context('Basic manipulations with consensus job replicas', () => {
                 cy.get('#cvat_canvas_text_content').should('be.visible')
                     .invoke('text')
                     .should('include', `${labelName}`)
-                    .and('include', 'consensus');
+                    .and('include', 'Consensus');
```
SpecLad, line 217: Is this really related to this patch?

cvat/apps/dataset_manager/bindings.py
archibald1418: The importer didn't account for the case when 'source' key was at the root level of a shape/track. It would only pick the 'source' field if attributes or elements had it. Otherwise it was setting 'source' value to default (='manual') - which is incorrect. 

Before the fix , comparing annotations with COCO 1.0 did not crash the test - because 'source' was wrongly being set to 'manual', it passed equality test but wasn't correct since all formats on import were assuming source=='file'

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -4967,10 +4941,49 @@ def _create_task(self, owner, assignee, annotation_format=""):
...
+                    "color": "#000000",
+                    "attributes": []
+                }
+            ]
 
         with ForceLogin(owner, self.client):
             response = self.client.post('/api/tasks', data=data, format="json")
-            assert response.status_code == status.HTTP_201_CREATED
+            self.assertEqual(response.status_code, status.HTTP_201_CREATED,
+                f"Reason:\n{pformat(dict(response.data))}"
```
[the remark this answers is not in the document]
archibald1418: true, removing

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -5541,13 +5554,25 @@ def _get_formats(self, user):
...
-    def _check_response(self, response, data):
+    def _check_response(self, response, data, expected_source='manual'):
+        IGNORE_KEYS = ["id", "version"]
         if not response.status_code in [
             status.HTTP_401_UNAUTHORIZED, status.HTTP_403_FORBIDDEN]:
             try:
-                compare_objects(self, data, response.data, ignore_keys=["id", "version"])
+                # Special checks for some fileds
+                # 1. 'source' is always 'file' after importing anno from file
+                if (source := response.data.get('source')):
```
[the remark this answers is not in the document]
archibald1418: Many annotations in the test and src had this so I figured this should be a valid input. Tracks [may](https://docs.cvat.ai/docs/api_sdk/sdk/reference/models/labeled-track/) have it

tests/cypress/e2e/features/consensus.js
archibald1418: no but the CI was crashing because of this
I can merge this in separate request

cvat/apps/dataset_manager/formats/mot.py
```diff
@@ -8,13 +8,15 @@
...
 from cvat.apps.dataset_manager.util import make_zip_archive
+from cvat.apps.engine.models import SourceType
 
 from .registry import dm_env, exporter, importer
 
 
 def _import_to_task(dataset, instance_data):
     tracks = {}
     label_cat = dataset.categories()[dm.AnnotationType.label]
+    import_source = str(SourceType.FILE)
```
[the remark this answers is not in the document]
archibald1418: applying

cvat/apps/dataset_manager/bindings.py
SpecLad: > The importer didn't account for the case when 'source' key was at the root level of a shape/track.

What would that look like? AFAIK, Datumaro's `Annotation` class doesn't have a `source` attribute.

Well, either way, this seems like a reasonable fix; but it is user-visible, so it needs a changelog entry.

cvat/apps/engine/tests/test_rest_api.py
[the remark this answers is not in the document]
SpecLad: But `response.data` is not an annotation, it is the result of an API call to `/api/.../annotations`. There should be no `source` field in there.

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -5541,13 +5554,25 @@ def _get_formats(self, user):
...
+                    self.assertEquals(source, expected_source)
+                if (anns := (response.data['shapes'] or response.data['tracks'])):
```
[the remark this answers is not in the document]
SpecLad: > This field is the only one that can change on import/export so should be checked separately.

Which field?

> As for tags, Isn't `assertDictEqual` enough to check them?

Which `assertDictEqual`? There isn't one in this function.

cvat/apps/engine/tests/test_rest_api_3D.py
```diff
@@ -208,6 +210,15 @@ def etree_to_dict(t):
...
 
 class Task3DTest(_DbTestBase):
+
+    def compare_annotations(self, a, b):
+        assert a.get('source') in ('manual', None)
+        assert b.get('source') in ('file', None)
+        compare_objects(self, a, b, ignore_keys=["source"])
+
+    def assertDictEqual(self, a, b):
+        self.compare_annotations(a, b)
```
SpecLad, line 220: This function seems to be pointless - it just calls another function with the same arguments.

cvat/apps/engine/tests/utils.py
```diff
@@ -455,3 +456,33 @@ def filter_dict(
     d: dict[str, Any], *, keep: Sequence[str] = None, drop: Sequence[str] = None
 ) -> dict[str, Any]:
     return {k: v for k, v in d.items() if (not keep or k in keep) and (not drop or k not in drop)}
+
+
+def compare_objects(self, obj1, obj2, ignore_keys, fp_tolerance=0.001, current_key=None):
+    key_info = "{}: ".format(current_key or "")
```
SpecLad, line 462: This doesn't seem right - it'll set `key_info` to `: ` if `current_key` is `None`.

cvat/apps/engine/tests/utils.py
```diff
@@ -455,3 +456,33 @@ def filter_dict(
...
+    error_msg = "{}{} != {}"
+
+    if isinstance(obj1, dict):
+        self.assertTrue(isinstance(obj2, dict), error_msg.format(key_info, obj1, obj2))
```
SpecLad, line 466: Seems like the formatted error message is always the same - so perhaps it would make sense to create it once at the beginning?

cvat/apps/engine/tests/test_rest_api_3D.py
```diff
@@ -208,6 +210,15 @@ def etree_to_dict(t):
 
 
 class Task3DTest(_DbTestBase):
+
+    def compare_annotations(self, a, b):
+        assert a.get('source') in ('manual', None)
+        assert b.get('source') in ('file', None)
```
SpecLad, line 216:
```suggestion
        assert b["source"] == "file"
```

I don't think it makes sense to allow `None` - if we want all loaded annotations to have source `"file"`, then that's what we should check.

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -6554,15 +6663,16 @@ def _get_initial_annotation(annotation_format):
 
                 # 7. check annotation
                 if export_format in {"Segmentation mask 1.1", "MOTS PNG 1.0",
-                        "CamVid 1.0", "ICDAR Segmentation 1.0"}:
+                        "CamVid 1.0", "ICDAR Segmentation 1.0",
+                        "Cityscapes 1.0"}:
```
SpecLad, line 6667: Why isn't it possible to predict the result for Cityscapes?

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -5541,13 +5554,25 @@ def _get_formats(self, user):
             )
         return response
 
-    def _check_response(self, response, data):
+    def _check_response(self, response, data, expected_source='manual'):
```
SpecLad, line 5557: IMO, "manual" isn't a good default here. This function is used to (among other things) check that when we save an annotation to CVAT, we get the same annotation back; in that case, the source should not be changed to manual, but be preserved. What you should do is default this to None, and only put `source` in `IGNORE_KEYS` if it's not None.

cvat/apps/engine/tests/test_rest_api.py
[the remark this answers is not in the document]
archibald1418: > Which field?
source

> Which assertDictEqual? There isn't one in this function.

Sorry I meant, assertEqual. Under the hood it uses `assertDictEqual` to compare dicts

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -6497,7 +6595,17 @@ def _get_initial_annotation(annotation_format):
...
+        formats.pop('Market-1501 1.0')
+        formats.pop('Kitti Raw Format 1.0') # temporary until review
+        # FIXME: Kitti Raw Format 1.0 bug in Datumaro Exporter
+        # when track_ids are absent, negative indices are assigned by the exporter
+        # if annotations contain shapes
+        # results in incorrect order of shapes inside dataset item subsets
+        # https://github.com/cvat-ai/datumaro/blob/4f6748e4f9eac038453e8f03344cdd9c73b00bc4/src/datumaro/plugins/data_formats/kitti_raw/exporter.py#L423C17-L423C18
+
         for export_format, import_format in formats.items():
+            print(f"{export_format} :: {import_format}")
```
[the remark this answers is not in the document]
archibald1418: ok, applying

cvat/apps/engine/tests/test_rest_api.py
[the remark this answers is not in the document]
archibald1418: Yes you're right, removing this check

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -6206,7 +6233,6 @@ def _get_initial_annotation(annotation_format):
                         }
                     ],
                 }]
-
```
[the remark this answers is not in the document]
archibald1418: You mean the line break?

cvat/apps/engine/tests/test_rest_api.py
archibald1418: It converts polygons to masks, so the points can be very different

cvat/apps/engine/tests/test_rest_api_3D.py
archibald1418: Makes sense, applying

cvat/apps/engine/tests/test_rest_api_3D.py
archibald1418: ok, removing

cvat/apps/engine/tests/utils.py
archibald1418: makes sense

cvat/apps/engine/tests/test_rest_api_3D.py
```diff
@@ -548,7 +563,11 @@ def test_api_v2_rewrite_annotation(self):
                     task_ann_prev.data["shapes"][0].pop("id")
                     task_ann.data["shapes"][0].pop("id")
                     self.assertEqual(len(task_ann_prev.data["shapes"]), len(task_ann.data["shapes"]))
-                    self.assertEqual(task_ann_prev.data["shapes"], task_ann.data["shapes"])
+                    for shape_prev, shape_ann in zip(
+                        task_ann_prev.data["shapes"],
+                        task_ann.data["shapes"]
+                    ):
+                        self.assertDictEqual(shape_prev, shape_ann)
```
archibald1418, line 570: I'll probably do a function for this

cvat/apps/engine/tests/test_rest_api.py
archibald1418: Good call, applying

cvat/apps/engine/tests/test_rest_api.py
zhiltsov-max: Yeah, the reason is that segmentation formats that store masks as images can add or remove boundary pixels during saving.

cvat/apps/engine/tests/test_rest_api.py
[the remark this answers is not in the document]
zhiltsov-max: I guess it should be applied to annotations inside `data`, but not to `data` itself.

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -5541,13 +5554,25 @@ def _get_formats(self, user):
...
             try:
-                compare_objects(self, data, response.data, ignore_keys=["id", "version"])
+                # Special checks for some fileds
+                # 1. 'source' is always 'file' after importing anno from file
+                if (source := response.data.get('source')):
+                    self.assertEquals(source, expected_source)
+                if (anns := (response.data['shapes'] or response.data['tracks'])):
+                    for ann in anns:
+                        self.assertEquals(ann.get('source'), expected_source)
+                IGNORE_KEYS.append('source')
```
zhiltsov-max, line 5569: `IGNORE_KEYS` is spelled in upper case suggesting that this is a constant, but then it's modified here

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -6272,7 +6296,9 @@ def _get_initial_annotation(annotation_format):
                                       + polygon_shapes_with_attrs
                 annotations["tags"] = tags_with_attrs + tags_wo_attrs
 
-            elif annotation_format == "ImageNet 1.0":
+            elif annotation_format in [
+                "ImageNet 1.0",
+                "Ultralytics YOLO Classification 1.0"]:
```
zhiltsov-max, line 6301:
```suggestion
                "Ultralytics YOLO Classification 1.0"
            ]:
```
This reads better

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -6288,20 +6314,43 @@ def _get_initial_annotation(annotation_format):
...
-                                      + tags_wo_attrs
+                annotations["shapes"] = points_wo_attrs
+                annotations["tags"] = tags_wo_attrs
 
             elif annotation_format == "KITTI 1.0":
                 annotations["shapes"] = rectangle_shapes_wo_attrs \
-                                            + polygon_shapes_wo_attrs
+                                            # + polygon_shapes_wo_attrs
+                # FIXME: polygons disappear after export in KITTI
+                # although documentation says the opposite
```
zhiltsov-max, line 6353: Polygons should be converted to masks in KITTI

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -6497,7 +6595,17 @@ def _get_initial_annotation(annotation_format):
             print("The following export formats have no pair:",
                 set(export_formats) - set(import_formats))
 
+        # Rare and buggy formats that are not crucial for testing
+        formats.pop('Market-1501 1.0')
```
zhiltsov-max, line 6599: Probably, it's better to skip this format with an issue link in a comment.

cvat/apps/engine/tests/test_rest_api.py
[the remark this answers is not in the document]
SpecLad: Yes.

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -6193,7 +6220,7 @@ def _get_initial_annotation(annotation_format):
                 tags_with_attrs = [{
                     "frame": 1,
                     "label_id": task["labels"][0]["id"],
-                    "group": 3,
+                    "group": 0,
```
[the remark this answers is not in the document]
SpecLad: Do you know why that happens?

cvat/apps/engine/tests/test_rest_api.py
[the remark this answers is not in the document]
SpecLad: > source

But I'm not talking about `source`, I'm talking about `shapes` and `tracks` (and `tags`).

cvat/apps/engine/tests/test_rest_api.py
archibald1418: removed the constant altogether, in favour of arrays with string literals

cvat/apps/engine/tests/test_rest_api.py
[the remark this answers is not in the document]
archibald1418: added checking tags, shapes and tracks

cvat/apps/engine/tests/utils.py
SpecLad: Currently the big useless strings are created anyway, when arguments to `assertTrue` are evaluated.

cvat/apps/engine/tests/test_rest_api_3D.py
```diff
@@ -208,6 +210,12 @@ def etree_to_dict(t):
 
 
 class Task3DTest(_DbTestBase):
+    def compare_shapes(self, shapes_orig: list[dict], shapes_imported: list[dict]):
+        for a, b in zip(shapes_orig, shapes_imported):
```
SpecLad, line 214: It would make sense to also move the length assertion to this function.

archibald1418: applying

cvat/apps/dataset_manager/bindings.py
zhiltsov-max, line 1: It seems this PR misses some changes from https://github.com/cvat-ai/cvat/pull/9265

cvat/apps/engine/tests/test_rest_api.py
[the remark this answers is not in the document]
archibald1418: Found it, it happens because the importer [doesn't](https://github.com/cvat-ai/cvat/blob/3480d7729f485d199a7f8fed9cfa0f74c0cec3c3/cvat/apps/dataset_manager/bindings.py#L2235) store one-item groups. Later the tag-importer takes the group value from the `group_map`

So I guess, it's not a bug but rather bad input. So group 3 -> group 0 seems like a valid change, what do you think?

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -6462,6 +6513,56 @@ def _get_initial_annotation(annotation_format):
                 annotations["shapes"] = rectangle_shapes_with_attrs \
                                       + polygon_shapes_with_attrs
 
+            elif annotation_format in [
```
zhiltsov-max, line 6516: `CVAT for *` and `Datumaro` also support skeletons

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -6497,6 +6598,12 @@ def _get_initial_annotation(annotation_format):
             print("The following export formats have no pair:",
                 set(export_formats) - set(import_formats))
 
+        # Rare and buggy formats that are not crucial for testing
+        formats.pop('Market-1501 1.0') # Issue: https://github.com/cvat-ai/datumaro/issues/99
+        formats.pop('Kitti Raw Format 1.0')
+        # can change order of shapes
```
zhiltsov-max, line 6604:
```suggestion
```
It's not a bug without special conditions.

cvat/apps/engine/tests/test_rest_api_3D.py
zhiltsov-max: We need to check len(a) == len(b)

cvat/apps/engine/tests/test_rest_api.py
archibald1418: So on checks, the test should ignore the order for this format only?

zhiltsov-max: If all the other fields match, it's fine for any format.

cvat/apps/engine/tests/utils.py
```diff
@@ -455,3 +456,49 @@ def filter_dict(
...
+def freeze_object(obj, ignore_keys=None) -> frozenset:
+    if isinstance(obj, dict):
+        obj = filter_dict(obj, drop=ignore_keys)
+        for k, v in obj.items():
+            obj[k] = freeze_object(v, ignore_keys)
+        return frozenset(obj.items())
+    elif isinstance(obj, list):
+        for i, elem in enumerate(obj):
+            obj[i] = freeze_object(elem, ignore_keys)
+        return frozenset(obj)
```
zhiltsov-max, line 470: Converting any list to a set doesn't seem correct, as it may drop repeated values. Consider using `tuple` instead.

cvat/apps/engine/tests/utils.py
```diff
@@ -455,3 +456,49 @@ def filter_dict(
     d: dict[str, Any], *, keep: Sequence[str] = None, drop: Sequence[str] = None
 ) -> dict[str, Any]:
     return {k: v for k, v in d.items() if (not keep or k in keep) and (not drop or k not in drop)}
+
+
+def freeze_object(obj, ignore_keys=None) -> frozenset:
```
zhiltsov-max, line 461: Maybe it's better to switch to https://pypi.org/project/deepdiff/? We already use it in integration tests.

cvat/apps/engine/tests/utils.py
archibald1418: Put a tuple here

cvat/apps/engine/tests/utils.py
archibald1418: I tried but couldn't get it to do excatly what I want here without hampering readability, so I stuck to the freeze_object solution which I would say is clearer and works

cvat/apps/dataset_manager/bindings.py
archibald1418: applied #9265, #9354

cvat/apps/engine/tests/test_rest_api.py
archibald1418: added skeleton inputs to these formats

cvat/apps/dataset_manager/bindings.py
zhiltsov-max: Please clean up 2 extra changelog fragments.

.vscode/launch.json
zhiltsov-max, line 1: Please remove extra changes in this file

changelog.d/20250501_231631_oleg.valiulin_fix_test_annotation_api_stops_early.md
```diff
@@ -0,0 +1,5 @@
+### Added
+
+- Set 'file' as default source of annotations imported from files
```
zhiltsov-max, line 3:
```suggestion
- Set `file` as default source of annotations imported from files
```

cvat/apps/engine/tests/test_rest_api.py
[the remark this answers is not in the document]
zhiltsov-max: Yes, single-item groups are meaningless and will be dropped on export.

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -5541,13 +5555,50 @@ def _get_formats(self, user):
...
+            IGNORE_KEYS = ["id", "version"]
+            TAG_DEFAULT_VALUES = dict(
+                source='manual',
+                attributes=[],
+            )
+            TRACK_DEFAULT_VALUES = dict(
+                **TAG_DEFAULT_VALUES,
+            )
+            SHAPE_DEFAULT_VALUES = dict(
+                **TRACK_DEFAULT_VALUES,
```
zhiltsov-max, line 5572: What is the logic behind these substitutions? Shapes are not supposed to include track fields, neither do tracks with tag fields. It would be better to extract common annotation fields to be used in all these structures instead. Also, tracks can include elements and track elements shapes can't have elements.

cvat/apps/engine/tests/utils.py
```diff
@@ -455,3 +459,78 @@ def filter_dict(
...
+
+
+def check_optional_fields(self: TestCase,
+                        obj: dict[str, int | float | str],
+                        optional_values: dict[str, int | float | str],
+                        expected_values: dict[str, int | float | str]
+                        ) -> None | NoReturn:
+    for k in optional_values.keys():
+        if obj.get(k) and not optional_values[k] == obj[k]:
+            self.assertEqual(obj[k], expected_values.get(k) or obj[k]) # coalesce
```
zhiltsov-max, line 499:
```suggestion
    for k, default in optional_values.items():
        self.assertEqual(obj.get(k, default), expected_values.get(k, default))
```

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -5541,13 +5555,50 @@ def _get_formats(self, user):
...
+                outside=False,
+                z_order=0,
+                rotation=0,
+                elements=[],
+            ) # if omitted, are set by the server
+              # https://docs.cvat.ai/docs/api_sdk/sdk/reference/models/labeled-shape/
+            _default_values = {
+                AnnotationType.TAG: TAG_DEFAULT_VALUES,
+                AnnotationType.SHAPE: SHAPE_DEFAULT_VALUES,
+                AnnotationType.TRACK: TRACK_DEFAULT_VALUES
```
zhiltsov-max, line 5583: I wouldn't recommend using this enum for this, because it's not required to match the annotation request fields. It will be better to simply use string constants here.

cvat/apps/engine/tests/test_rest_api.py
archibald1418: Now comparing only common default fields

cvat/apps/engine/tests/test_rest_api.py
archibald1418: removed this in favour ['shapes', 'tracks', 'tags']

cvat/apps/engine/tests/utils.py
```diff
@@ -497,7 +497,8 @@ def check_optional_fields(self: TestCase,
                         ) -> None | NoReturn:
     for k in optional_values.keys():
         # if obj.get(k) and not optional_values[k] == obj[k]:
-        self.assertEqual(obj.get(k, optional_values[k]), optional_values[k]) # coalesce
+        self.assertEqual(obj.get(k, optional_values[k]), optional_values[k],
+                         f"key '{k}' has incorrect expected optional value {obj.get(k)} instead of {optional_values[k]}") # coalesce
```
zhiltsov-max, line 501: - Please remove commented code.
- It's not totally clear what this function is supposed to check now. Optional values have to be present always, because they have default values, but the values doesn't have to be the defaults.

cvat/apps/engine/tests/utils.py
```diff
@@ -455,3 +459,80 @@ def filter_dict(
...
+
+
+def compare_objects(self: TestCase, obj1, obj2, ignore_keys, fp_tolerance=0.001, current_key=None, order=True):
+    key_info = "{}: ".format(current_key) if current_key else ""
+    error_msg = "{}{} != {}"
+
+    def is_annotation_type(k):
+        return any(k.startswith(_type) for _type in [AnnotationType.SHAPE, AnnotationType.TAG])
```
zhiltsov-max, line 509: - Why not `k == _type`?
- Again, using `AnnotationType` is not totally correct here. This enum doesn't belong to the `engine` component as isn't supposed to reflect annotation response fields

cvat/apps/engine/tests/utils.py
```diff
@@ -455,3 +459,80 @@ def filter_dict(
...
+            v1 = obj1[k]
+            v2 = obj2[k]
+            if k == "attributes":
+                key = lambda a: (a.get("spec_id") or a["id"])
+                v1.sort(key=key)
+                v2.sort(key=key)
+            elif not order and is_annotation_type(k):
+                v1 = frozenset(freeze_object(v1, ignore_keys))
+                v2 = frozenset(freeze_object(v2, ignore_keys))
+            compare_objects(self, v1, v2, ignore_keys, current_key=k)
```
zhiltsov-max, line 524:
```suggestion
            compare_objects(self, v1, v2, ignore_keys, current_key=k, fp_tolerance=fp_tolerance, order=order)
```
And in the call below as well.

cvat/apps/engine/tests/utils.py
archibald1418: removed altogether now simply using `key in {'shapes', 'tags'}`

cvat/apps/engine/tests/utils.py
```diff
@@ -455,3 +459,80 @@ def filter_dict(
...
+) -> list | dict:
+    if isinstance(obj, dict):
+        obj = filter_dict(obj, keep=keep, drop=drop)
+        for k, v in obj.items():
+            obj[k] = filter_object(v, keep=keep, drop=drop)
+        return obj
+    elif isinstance(obj, list):
+        for i, val in enumerate(obj):
+            obj[i] = filter_object(val, keep=keep, drop=drop)
+        obj = list(filter(lambda x: x != dict(), obj))
```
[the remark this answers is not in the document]
archibald1418: removed the function altogether, in favour of check_optional_fields

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -6657,6 +6656,8 @@ def _get_initial_annotation(annotation_format):
         # Rare and buggy formats that are not crucial for testing
         formats.pop('Market-1501 1.0') # Issue: https://github.com/cvat-ai/datumaro/issues/99
 
+        formats = {'COCO Keypoints 1.0': 'COCO Keypoints 1.0'}
```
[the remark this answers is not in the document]
archibald1418: removed

cvat/requirements/testing.txt
```diff
@@ -12,3 +12,5 @@ lupa==2.4
     # via fakeredis
 sortedcontainers==2.4.0
     # via fakeredis
+ipdb==0.13.13
+    # a nicer debugger with --pdb
```
zhiltsov-max, line 16: Why this was added? Our default workflow is to use VS Code and it's debugger, if needed.

cvat/apps/engine/tests/test_rest_api_3D.py
```diff
@@ -208,6 +211,24 @@ def etree_to_dict(t):
 
 
 class Task3DTest(_DbTestBase):
+    def compare_annotations(self, annos_orig: list[dict], annos_imported: list[dict]):
+
+        for _type in set(AnnotationType):
```
zhiltsov-max, line 216: This enum should not be used in this file.

cvat/apps/engine/tests/test_rest_api_3D.py
```diff
@@ -208,6 +211,24 @@ def etree_to_dict(t):
...
+            key = f"{_type}s"
+            anns = annos_imported.data[key]
+            self.assertEqual(
+                len(annos_orig.data[key]),
+                len(annos_imported.data[key]),
```
zhiltsov-max, line 221: `annos_orig` and `annos_imported` are both supposed to be `list`s, and lists do not have a `data` member.

cvat/apps/engine/tests/test_rest_api_3D.py
```diff
@@ -208,6 +211,24 @@ def etree_to_dict(t):
 
 
 class Task3DTest(_DbTestBase):
+    def compare_annotations(self, annos_orig: list[dict], annos_imported: list[dict]):
+
+        for _type in set(AnnotationType):
+            key = f"{_type}s"
+            anns = annos_imported.data[key]
+            self.assertEqual(
```
zhiltsov-max, line 219: This check duplicates the checks in the `compare_objects`. Probably, it can be removed.

cvat/requirements/testing.txt
archibald1418: Yes, but it doesn't work with `--pdb `when launching a debug config. It should drop to pdb on failed test assertions but instead it only stops at the end of manage.py, and at this point the stack is already lost. Instead you would have to manually debug step by step to find the failure point, disable breakpoints manually, add conditions and so on. So I have found it that debugging through terminal with `python manage.py ... --pdb` is much more convenient than doing this across our massive test suites.

Possibly, you have managed to make it work with VSCode in this case, I would be happy to hear more about your workflow

cvat/apps/engine/tests/test_rest_api_3D.py
archibald1418: yes, removing from the 3d, too

cvat/apps/engine/tests/test_rest_api_3D.py
archibald1418: applying

cvat/requirements/testing.txt
SpecLad: I'd like to point out that you are free to install whatever tools you want to use; however, this file is for test suite _requirements_.

Moreover, it is autogenerated, so your addition will disappear next time someone updates it.

cvat/apps/engine/tests/test_rest_api_3D.py
```diff
@@ -208,6 +211,24 @@ def etree_to_dict(t):
...
+                len(annos_orig.data[key]),
+                len(annos_imported.data[key]),
+                "Different number of shapes"
+            )
+            for ann in anns:
+                self.assertEquals(ann.get('source'), 'file')
+        compare_objects(self,
+                        annos_orig.data, annos_imported.data,
+                        ignore_keys=["source", "id", "version"],
```
zhiltsov-max, line 228: There are many instances of id field cleanups in the tests in this file. They can probably be removed now.

```
task_ann_prev.data["shapes"][0].pop("id")
task_ann.data["shapes"][0].pop("id")
```

cvat/requirements/testing.txt
SpecLad: > I don't think anyone used pdb to debug unit tests in the project. 

FWIW, I use it all the time. :slightly_smiling_face: But I'm quite satisfied with vanilla pdb.

> this step should probably be enabled on the CI to speed up problem search, if it's not

It is, I added it back in #8338.

cvat/apps/engine/tests/utils.py
```diff
@@ -455,3 +457,84 @@ def filter_dict(
...
+            obj[k] = filter_object(v, keep=keep, drop=drop)
+        return obj
+    elif isinstance(obj, list):
+        for i, val in enumerate(obj):
+            obj[i] = filter_object(val, keep=keep, drop=drop)
+        return obj
+    return obj
+
+
+def compare_objects(self: TestCase, obj1, obj2, ignore_keys, fp_tolerance=0.001, current_key=None, order=True):
```
zhiltsov-max, line 504: Here's my take on the comparison: 

<details>

```python
from scipy.optimize import linear_sum_assignment


def _match_lists(
    a_objs: list[Any],
    b_objs: list[Any],
    *,
    distance: Callable[[Any, Any], bool],
) -> tuple[list[tuple[Any, Any]], list[Any], list[Any]]:
    """
    Matches two lists of objects using a distance function.
    The distance function should return True if the objects are equal, and False otherwise.

    Returns: (matches, a_unmatched, b_unmatched)
    """

    if len(a_objs) != len(b_objs):
        raise ValueError("The number of objects in the lists should be equal")

    distances = np.asarray([1 - int(distance(a, b)) for a in a_objs for b in b_objs])
    distances = distances.reshape((len(a_objs), len(b_objs)))

    # O(n^3) complexity or better
    a_matches, b_matches = linear_sum_assignment(distances)

    matches = []
    a_unmatched = []
    b_unmatched = []

    for a_idx, b_idx in zip(a_matches, b_matches):
        dist = distances[a_idx, b_idx]
        if dist == 1:
            if a_idx < len(a_objs):
                a_unmatched.append(a_objs[a_idx])
            if b_idx < len(b_objs):
                b_unmatched.append(b_objs[b_idx])
        else:
            matches.append((a_objs[a_idx], b_objs[b_idx]))

    return matches, a_unmatched, b_unmatched


def _format_key(key: list[str]) -> str:
    return ".".join([] + key) or ""


def compare_objects(
    self: TestCase,
    obj1: Any,
    obj2: Any,
    ignore_keys: Collection[str],
    *,
    fp_tolerance: float = 0.001,
    current_key: list[str] | str | None = None,
    order: bool = True,
) -> bool:
    if isinstance(current_key, str):
        current_key = [current_key]
    elif not current_key:
        current_key = []

    key_info = f"{_format_key(current_key)}: "
    error_msg = "{}{} != {}"

    if isinstance(obj1, dict):
        self.assertTrue(isinstance(obj2, dict), error_msg.format(key_info, obj1, obj2))
        for k in (obj1.keys() | obj2.keys()) - set(ignore_keys):
            compare_objects(
                self,
                obj1[k],
                obj2[k],
                ignore_keys,
                current_key=current_key + [k],
                fp_tolerance=fp_tolerance,
                order=order,
            )
    elif isinstance(obj1, list):
        self.assertTrue(isinstance(obj2, list), error_msg.format(key_info, obj1, obj2))
        self.assertEqual(
            len(obj1),
            len(obj2),
            error_msg.format(key_info, pformat(obj1, compact=True), pformat(obj2, compact=True)),
        )

        if order:
            for v1, v2 in zip(obj1, obj2):
                compare_objects(
                    self,
                    v1,
                    v2,
                    ignore_keys,
                    current_key=current_key,
                    fp_tolerance=fp_tolerance,
                    order=order,
                )
        else:

            def _compare(a, b) -> bool:
                try:
                    compare_objects(
                        self,
                        a,
                        b,
                        ignore_keys,
                        current_key=current_key,
                        fp_tolerance=fp_tolerance,
                        order=order,
                    )
                    return True
                except AssertionError:
                    return False

            _, a_unmatched, b_unmatched = _match_lists(obj1, obj2, distance=_compare)

            if a_unmatched or b_unmatched:
                self.fail(
                    "Failed to match lists. "
                    + error_msg.format(
                        key_info, pformat(obj1, compact=True), pformat(obj2, compact=True)
                    )
                )

    elif isinstance(obj1, float) or isinstance(obj2, float):
        self.assertAlmostEqual(obj1, obj2, delta=fp_tolerance, msg=current_key)
    else:
        self.assertEqual(obj1, obj2, msg=current_key)
```

It needs a little tuning for the output in the case of error and for supporting key-based ordering to be fully correct. For instance, we need to compare annotation coordinated as they are, but the annotation order can be arbitrary. It can be done via, e.g., a custom callable or a prefix filter in the `order` argument.

</details>

cvat/requirements/testing.txt
archibald1418: removed the ipdb dep and the pyproject.toml edit

cvat/apps/engine/tests/_types.py
```diff
@@ -0,0 +1,9 @@
+from typing import Protocol
+
+
+class OrderStrategy(Protocol):
+    def __call__(self, key_path: str) -> bool: ...
+
+
+always_check_order: OrderStrategy = lambda _: True
+never_check_order: OrderStrategy = lambda _: False
```
zhiltsov-max, line 9: These 2 modes could probably be represented simply by the `True` and `False` values

cvat/apps/engine/tests/utils.py
```diff
@@ -533,3 +554,237 @@ def filter_dict(
...
+                "shape_elements": [shape.get("elements", []) for shape in data["shapes"]],
+                "track_elements": [track.get("elements", []) for track in data["tracks"]],
+            }
+
+        response_elements = _get_elements(response_data)
+        for elements in response_elements.values():
+            for elem in elements:
+                for ann in elem:
+                    check_optional_fields(self, ann, COMMON_DEFAULT_FIELDS)
+        response_elements = _get_elements(response_data)
```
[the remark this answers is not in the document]
archibald1418: true, deleted this

cvat/apps/engine/tests/_types.py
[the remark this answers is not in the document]
archibald1418: These are custom types and there will probably be more of them

cvat/apps/engine/tests/_types.py
zhiltsov-max: I mean, why not define `check_order: bool | OrderStrategy = True,` instead?

cvat/apps/engine/tests/utils.py
```diff
@@ -533,3 +597,244 @@ def filter_dict(
...
+        self.assertEqual(obj1, obj2, msg=current_key)
+
+
+def check_annotation_response(
+    self: TestCase,
+    response: dict,
+    data: dict,
+    *,
+    expected_source: str | None = None,
+    ignore_keys: list[str] = ["id", "version"],
```
zhiltsov-max, line 778:
```suggestion
    ignore_keys: list[str] = ("id", "version"),
```
Using mutable objects as function defaults is unsafe

cvat/apps/engine/tests/utils.py
```diff
@@ -533,3 +554,237 @@ def filter_dict(
     d: dict[str, Any], *, keep: Sequence[str] = None, drop: Sequence[str] = None
 ) -> dict[str, Any]:
     return {k: v for k, v in d.items() if (not keep or k in keep) and (not drop or k not in drop)}
+
+
+def freeze_object(obj, ignore_keys=None) -> frozenset:
```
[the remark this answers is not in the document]
archibald1418: removed

cvat/apps/engine/tests/test_rest_api.py
zhiltsov-max, line 1: I agree that this code has to be formatted, but mixing formatting in the whole big file with other changes is not a good idea in most cases. Could you please separate these changes into different PRs?

btw, we have a script to format only what's needed: `dev/format_python_code.sh`.

cvat/apps/engine/tests/utils.py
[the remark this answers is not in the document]
zhiltsov-max: It's still here.

cvat/apps/engine/tests/test_rest_api.py
archibald1418: Reverted range of commits after blacking, applied the the script `format_python_code.sh`, re-applied changes related to linting

cvat/apps/engine/tests/utils.py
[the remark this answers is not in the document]
zhiltsov-max:
```suggestion
```

cvat/apps/engine/tests/utils.py
```diff
@@ -533,3 +556,228 @@ def filter_dict(
...
+    for k in optional_values.keys():
+        right = obj.get(k)
+        left = right if right is not None else optional_values[k]  # accounting for bool fields
+        if right is None:
+            right = optional_values[k]
+        self.assertEqual(
+            left,
+            right,
+            f"key {k!s} has incorrect actual optional value {left!s} instead of expected {right!s}",
+        )
```
zhiltsov-max, line 575: - could you rename left and right to something more self-explanatory? The values seem to be `actual` and `expected`, not simply elements of 2 objects.
- I can't find where the optional values are compared between 2 annotations
- consider using `{}.get(key, default)`

cvat/apps/engine/tests/utils.py
```diff
@@ -533,3 +557,234 @@ def filter_dict(
...
+                    )
+                )
+
+    elif isinstance(obj1, float) or isinstance(obj2, float):
+        self.assertAlmostEqual(obj1, obj2, delta=fp_tolerance, msg=current_key)
+    else:
+        self.assertEqual(obj1, obj2, msg=current_key)
+
+
+def check_annotation_response(
```
zhiltsov-max, line 722: How about this?

<details>

```
def filter_dict(
    d: dict[str, Any], *, keep: Sequence[str] = None, drop: Sequence[str] = None
) -> dict[str, Any]:
    return {
        k: v for k, v in d.items()
        if (keep is None or k in keep) and (drop is None or k not in drop)
    }

def compare_objects(
    self: TestCase,
    obj1: Any,
    obj2: Any,
    ignore_keys: Collection[str],
    *,
    defaults: dict[str, Any] | Callable[[list[str]], dict | None] | None = None,
    fp_tolerance: float = 0.001,
    current_key: list[str] | str | None = None,
    check_order: bool | OrderStrategy = True,
) -> bool | NoReturn:
    if isinstance(current_key, str):
        current_key = [current_key]
    elif not current_key:
        current_key = []

    key_info = f"{_format_key(current_key)}: "
    error_msg = "{}{} != {}"

    if isinstance(obj1, dict):
        self.assertTrue(isinstance(obj2, dict), error_msg.format(key_info, obj1, obj2))

        current_key_defaults = {}
        if defaults and isinstance(defaults, dict):
            current_key_defaults = defaults
        elif defaults and callable(defaults):
            current_key_defaults = defaults(current_key) or {}

        for k in (obj1.keys() | obj2.keys() | current_key_defaults.keys()) - set(ignore_keys):
            compare_objects(
                self,
                obj1[k] if not k in current_key_defaults else obj1.get(k, current_key_defaults[k]),
                obj2[k] if not k in current_key_defaults else obj2.get(k, current_key_defaults[k]),
                ignore_keys,
                defaults=defaults,
                current_key=current_key + [k],
                fp_tolerance=fp_tolerance,
                check_order=check_order,
            )
    elif isinstance(obj1, list):
        self.assertTrue(isinstance(obj2, list), error_msg.format(key_info, obj1, obj2))
        self.assertEqual(
            len(obj1),
            len(obj2),
            error_msg.format(key_info, pformat(obj1, compact=True), pformat(obj2, compact=True)),
        )

        if check_order is True or check_order(current_key):
            for v1, v2 in zip(obj1, obj2):
                compare_objects(
                    self,
                    v1,
                    v2,
                    ignore_keys,
                    defaults=defaults,
                    current_key=current_key,
                    fp_tolerance=fp_tolerance,
                    check_order=check_order,
                )
        else:

            def _compare(a, b) -> bool:
                try:
                    compare_objects(
                        self,
                        a,
                        b,
                        ignore_keys,
                        defaults=defaults,
                        current_key=current_key,
                        fp_tolerance=fp_tolerance,
                        check_order=check_order,
                    )
                    return True
                except AssertionError:
                    return False

            _, a_unmatched, b_unmatched = _match_lists(obj1, obj2, distance=_compare)

            if a_unmatched or b_unmatched:
                self.fail(
                    "Failed to match lists. "
                    + error_msg.format(
                        key_info, pformat(obj1, compact=True), pformat(obj2, compact=True)
                    )
                )

    elif isinstance(obj1, float) or isinstance(obj2, float):
        self.assertAlmostEqual(obj1, obj2, delta=fp_tolerance, msg=current_key)
    else:
        self.assertEqual(obj1, obj2, msg=current_key)


def check_annotation_response(
    self: TestCase, response: dict, data: dict,
    *,
    expected_values: dict | None = None,
    ignore_keys: Sequence[str] = frozenset(("id", "version"))
) -> None | NoReturn:
    OPTIONAL_FIELDS = dict(
        source="manual",
        occluded=False,
        outside=False,
        z_order=0,
        rotation=0,
        attributes=[],
        elements=[],
    )  # if omitted, are set by the server
    # https://docs.cvat.ai/docs/api_sdk/sdk/reference/models/labeled-shape/
    # elements are handled separately

    if expected_values is not None:
        def put_expected_values(v: Any) -> Any:
            if isinstance(v, dict):
                v.update(filter_dict(expected_values, keep=v.keys() & expected_values.keys()))

                for k, vv in v.items():
                    v[k] = put_expected_values(vv)
            if isinstance(v, list):
                v = [put_expected_values(item) for item in v]
            if isinstance(v, tuple):
                v = tuple(put_expected_values(item) for item in v)

            return v

        data = put_expected_values(deepcopy(data))

    def _check_order_in_annotations(key_path: list[str]) -> bool:
        return "points" in key_path

    def _key_defaults(key_path: list[str]) -> dict | None:
        if key_path and key_path[-1] == "tags":
            return filter_dict(OPTIONAL_FIELDS, keep=["group", "source", "attributes"])
        if key_path and key_path[-1] == "shapes":
            return filter_dict(OPTIONAL_FIELDS, keep=["occluded", "outside", "z_order", "rotation", "group", "source", "attributes", "elements"])
        if key_path and key_path[-1] == "tracks":
            return filter_dict(OPTIONAL_FIELDS, keep=["group", "source", "attributes", "elements"])
        if key_path and _format_key(key_path).endswith("tracks.elements"):
            return filter_dict(OPTIONAL_FIELDS, keep=["group", "source", "attributes"])
        if key_path and _format_key(key_path).endswith("tracks.elements.shapes"):
            return filter_dict(OPTIONAL_FIELDS, keep=["occluded", "outside", "z_order", "rotation", "attributes"])

        return None

    try:
        compare_objects(
            self,
            response.data,
            data,
            ignore_keys=ignore_keys,
            defaults=_key_defaults,
            check_order=_check_order_in_annotations,
        )
    except AssertionError as e:
        print(
            "Objects are not equal:",
            pformat(response.data, compact=True),
            "!=",
            pformat(data, compact=True),
            sep="\n",
        )
        print(e)
        raise

```

</details>

archibald1418: Couple of cases that were missed here:
- root-level 'source' is padded by the server (added `if key_path == []:` to _key_defaults)
- 'attribute.source' is not added by the server (added `k != 'attributes'` to `put_expected_values`)

zhiltsov-max: What are these extra `source` fields? They're not described in the API.

archibald1418: + IIRC, 'group' doesn't have a default value and is not padded by the server
e.g. [here](https://docs.cvat.ai/docs/api_sdk/sdk/reference/models/labeled-track/)

archibald1418: > What are these extra `source` fields? They're not described in the API.

Yes, this was my mistake. I removed the if for root-level 'source' and it works fine. No other extra 'source' fields are found anywhere else.
As for 'attributes.source', the function added optional fields to an attribute so I just made it ignore 'attributes' altogether

zhiltsov-max (approved):

PR #9354, opened 2025-04-22
less ram on export to cvat

Eldies:
### Checklist

### License

### DIFF
--- a/cvat/apps/dataset_manager/bindings.py
+++ b/cvat/apps/dataset_manager/bindings.py
@@ -1761,7 +1761,7 @@ class CvatTaskOrJobDataExtractor(StreamingSubsetBase, CvatDataExtractorBase):
         self._grouped_by_frame = list(self._instance_data.group_by_frame(include_empty=True))
 
     @staticmethod
-    def replace_lazy_lists(frame_data: CommonData.Frame) -> CommonData.Frame:
+    def copy_frame_data_with_replaced_lazy_lists(frame_data: CommonData.Frame) -> CommonData.Frame:
         return frame_data._replace(
             labeled_shapes=[
                 (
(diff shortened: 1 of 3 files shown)


cvat/apps/dataset_manager/bindings.py
```diff
@@ -1760,19 +1760,23 @@ def __init__(self, *args, **kwargs):
 
         self._grouped_by_frame = list(self._instance_data.group_by_frame(include_empty=True))
 
+    @staticmethod
+    def replace_lazy_lists(frame_data: CommonData.Frame) -> CommonData.Frame:
```
Marishka17, line 1764: I think the current method name doesn't make it clear that the replacement is not done in place. Would copy_frame_data_with_replaced_lazy_lists be a better name?

Additionally, I believe this method should be declared in the base class.

Eldies: renamed it.
I dont think it should be declared in base class - types are different, and it is only used in successors

Marishka17: I'm not so familiar with this code base, but it looks like:
- there are only 2 subclasses of `CvatDataExtractorBase`: `CvatTaskOrJobDataExtractor` and `CVATProjectDataExtractor` and both of them require to have this logic.
- `CommonData.Frame` and `ProjectData.Frame` can be refactored to reduce code duplication, can inherit one another -> in this case, the implementation of the `copy_frame_data_with_replaced_lazy_lists` child methods will be the same. I think it's a standard architecture pattern when a base class declare some methods that must be implemented in each child class.

But okay, let's postpone the method moving until related refactoring.

Marishka17 (approved):

PR #9356, opened 2025-04-23
Size controls for cuboid items in the 3d workspace

### LINKED ISSUE #7562: 3D cuboid sizes as text input

We're using cvat to do annotation of pointclouds, in our ODD we have quite a lot of objects with one or a few dimensions having a fixed metric size. During labelling the cuboids produced by human-labelers have variance, due to how accurately a labeler can adjust the annotation dimensions. This problem is especially pronounced on objects with low density of the pointcloud, which naturally happens for objects that are far away from the sensor.

As an improvement I see that the variation in annotations can be reduced by providing a text-field allowing an annotator to type-in a known size for certain dimensions of an object.

alternative can be that at least it is possible to see that metric size of dimensions whilest adjusting the size of the cuboid.

### Additional context

image

klakhov:
### Motivation and context

Resolves #7562

This is a qucik implementation of use case when objects sizes are known before annotation.
Annotators can create one(or several) objects  with correct sizes and then copy/paste them in correct places
size-controls

### Checklist

### License


cvat-ui/src/components/annotation-page/standard-workspace/objects-side-bar/object-item-details.tsx
```diff
@@ -4,10 +4,12 @@
 // SPDX-License-Identifier: MIT
 
 import React from 'react';
-import { Row } from 'antd/lib/grid';
+import { Row, Col } from 'antd/lib/grid';
 import Text from 'antd/lib/typography/Text';
 import Collapse from 'antd/lib/collapse';
 
+import { SizeType } from 'reducers';
+import InputNumber from 'antd/lib/input-number';
```
bsekachev, line 12: 3rdparty imports prior local

cvat-ui/src/components/annotation-page/standard-workspace/objects-side-bar/object-item-details.tsx
```diff
@@ -52,26 +63,67 @@ function ItemAttributesComponent(props: Props): JSX.Element {
...
+                                    <InputNumber
+                                        value={sizeParams.w || ''}
+                                        onChange={(value) => {
+                                            if (typeof value === 'number') {
+                                                changeSize(SizeType.WIDTH, value);
+                                            }
+                                        }}
+                                        disabled={readonly}
+                                    />
+                                </Col>
```
bsekachev, line 80: Seems like code duplication may be reduced for these 3 elements with Array.map

cvat-ui/src/reducers/index.ts
```diff
@@ -854,6 +854,12 @@ export enum ColorBy {
     LABEL = 'Label',
 }
 
+export enum SizeType {
+    WIDTH = 'width',
+    HEIGHT = 'height',
+    LENGTH = 'length',
+}
```
bsekachev, line 861: We only use this enum in two component files (like SizeParams interface), I am not sure it will be necessary in other components, and thus, I do not think we need to keep it in reducers. 

Maybe it is a good place for types used somehow in the app storage, but not for any

klakhov: Moved to `object-item-details`

nmanovic:
@klakhov , will it be possible to set a size as default for all cuboids?

klakhov:
No, the default cuboid size will still be (1, 1, 1). However, as described in the issue, there may be multiple valid "default" sizes depending on the use case. The idea here is to allow users to configure several cuboids using input fields and then copy them. Previously, copying wasn’t a reliable solution because the UI didn’t indicate the current cuboid size—so users couldn’t be sure that the copied cuboid met the intended size constraints.

PR #9364, opened 2025-04-24
Add project ID display in projects page

### LINKED ISSUE #9347: Show project ID next to it's name like task or job

It'd be great to see the project ID with the name. Complex datasets may require multiple projects inside the CVAT to organize them properly. Since the ID is unique, it would be great to see to eliminate any misunderstandings.

_No response_

_No response_

### Additional context

_No response_

omerferhatt:
Fixes #9347

### Checklist

- [*] I submit my changes into the `develop` branch
- [*] I have linked related issues (see [GitHub docs](

### License

- [*] I submit _my code changes_ under the same [MIT License](

### DIFF
--- a/cvat-ui/src/components/projects-page/project-item.tsx
+++ b/cvat-ui/src/components/projects-page/project-item.tsx
@@ -93,6 +93,7 @@ export default function ProjectItemComponent(props: Props): JSX.Element {
                             className='cvat-projects-project-item-title'
                             aria-hidden
                         >
+                            <Text strong type='secondary' className='cvat-item-project-id'>{`#${instance.id}: `}</Text>
                             {instance.name}
                         </Text>
                     )}


klakhov:
I don't think the current implementation will be very useful. Right now, the project is represented as a card, which is already compact. Displaying both the ID and name on a single line would make the project name even smaller and harder to read.

Personally, I believe we should navigate between projects primarily using their name and preview. While the ID can be useful in some cases, it should be shown explicitly and not crammed onto the same line as the name—especially if the ID is something long like `#12345678`.

A better approach might be to display the ID as a ribbon component, similar to how it's shown on the jobs page. However, I also think this should be configurable, as not all users will need to see the ID.

klakhov:
I will close the PR for now, we can re-open it if we find the display of id useful. Lets first discuss the implementation scenario in the issue.

PR #9367, opened 2025-04-25
Fixed exception when paste polygon with ctrl hold

bsekachev:
### Motivation and context
Scenario
1. Draw a 2D cuboid
2. Copy it
3. Paste it with ctrl several times

UI crashes. 

Reason?

So, different things are not looks good. For example canvas should not be called .configure method on any change in annotations. `configure` probably should not call `transform` when unnecessary. Or unnecessary `paintHandler` objects should not be created when transform called for different shapes.

But the core reason for this error is that we call `transform` on any drawInstance when it only defined for `rect`, `polygon`, `points`. I tried to understand why we added these custom methods 5 years ago, unsuccessfully. Then I tried to remove them and everything works in latest Chrome and Mozilla. So, maybe we just do not need them. 

### Checklist

### License

### DIFF
--- a/changelog.d/20250425_104058_sekachev.bs_fixed_exception.md
+++ b/changelog.d/20250425_104058_sekachev.bs_fixed_exception.md
@@ -1,0 +1,4 @@
+### Fixed
+
+- UI crashes when paste cuboids with hold Ctrl key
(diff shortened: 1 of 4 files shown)


klakhov (approved):
Fix works for me

PR #9371, opened 2025-04-28
Implement support for interactive detection via agents

SpecLad:
### Motivation and context

This fills one of the feature gaps between agent-based and Nuclio functions.

### Checklist

### License

### DIFF
--- a/cvat-cli/src/cvat_cli/_internal/agent.py
+++ b/cvat-cli/src/cvat_cli/_internal/agent.py
@@ -494,11 +494,14 @@ class _Agent:
             self._process_available_ars(category)
 
     def _process_available_ars(self, category) -> None:
-        if self._potential_work_per_category[category]:
+        with self._potential_work_condition:
+            if not self._potential_work_per_category[category]:
+                return
+
             self._potential_work_per_category[category] = False
 
-            while ar_assignment := self._poll_for_ar(category):
-                self._process_ar(ar_assignment)
+        while ar_assignment := self._poll_for_ar(category):
+            self._process_ar(ar_assignment)
 
     def _process_ar(self, ar_assignment: dict) -> None:
         self._client.logger.info("Got annotation request assignment: %r", ar_assignment)
--- a/cvat-cli/src/cvat_cli/_internal/agent.py
+++ b/cvat-cli/src/cvat_cli/_internal/agent.py
@@ -13,7 +13,7 @@ import secrets
 import shutil
 import tempfile
 import threading
-from collections.abc import Iterator, Sequence
+from collections.abc import Generator, Iterator, Sequence
 from datetime import datetime, timedelta, timezone
 from pathlib import Path
 from typing import TYPE_CHECKING, Optional, Union
@@ -28,6 +28,7 @@ from cvat_sdk.auto_annotation.driver import (
     _DetectionFunctionContextImpl,
     _SpecNameMapping,
 )
+from cvat_sdk.datasets.caching import make_cache_manager
 from cvat_sdk.exceptions import ApiException
 
 from .common import CriticalError, FunctionLoader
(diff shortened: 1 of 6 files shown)


cvat-cli/src/cvat_cli/_internal/agent.py
```diff
@@ -108,6 +112,73 @@ class _NewReconnectionDelay:
...
+    This class deletes least-recently used tasks from the dataset cache,
+    so that at any time the cache contains at most _MAX_TASKS_WITH_CHUNKS
+    tasks with downloaded chunks, and at most _MAX_TASKS_WITHOUT_CHUNKS without.
+
+    This helps manage disk usage, since agents may run indefinitely, and
+    we don't want the dataset cache to keep growing.
+    """
+
+    _MAX_TASKS_WITH_CHUNKS = 1
+    _MAX_TASKS_WITHOUT_CHUNKS = 10
```
zhiltsov-max, line 126: If we're trying to manage disk space, why not limit the maximum amount of memory occupied? If we limit the number or tasks, we don't provide memory guarantees. We have implicit server-side restrictions on chunk size (Redis' 512MB), so if we limit the storage used by, e.g., 1 GB by default, we will have limited cache space, yet we'll always be able to serve a request (by cleaning the cache until it can fit the requested data). I'd always try to download the missing requested chunk, even though it can be bigger than the maximum cache size.

cvat-cli/src/cvat_cli/_internal/agent.py
```diff
@@ -413,6 +489,17 @@ def run(self, *, burst: bool) -> None:
 
                 watcher.join()
 
+    def _process_all_available_ars(self):
+        for category in REQUEST_CATEGORIES_WITH_DECREASING_PRIORITY:
+            self._process_available_ars(category)
+
+    def _process_available_ars(self, category) -> None:
+        if self._potential_work_per_category[category]:
+            self._potential_work_per_category[category] = False
```
zhiltsov-max, line 498: Shouldn't this access be protected by the c. v. lock acquisition? If I'm not mistaken, it's possible to miss a new request notification currently. I can see only one thread locks the c. v., so the lock doesn't control the c. v. access.

SpecLad: I thought about it, but I couldn't find a scenario in which anything bad would happen. The watcher thread can only set `_potential_work_per_category[category]` to `True`, and there are four cases:

1. It happens before the `if`. In this case the `if` condition evaluates to true, and we process all outstanding requests.
2. It happens between the evaluation of the condition and the assignment. Since we're already within the `if` body, we process all outstanding requests.
3. It happens after the assignment, but within the `if` body. Since we keep acquiring requests until we the server runs out of them, the new request will be processed too. `_potential_work_per_category[category]` will remain set to True, so `_wait_between_polls` will terminate immediately and `_process_all_available_ars` will be called again - which is redundant, but harmless.
4. It happens after the `if` body. In this case we exit `_process_all_available_ars` without processing the new request, but now `_wait_between_polls` will terminate immediately, and `_process_all_available_ars` will be called again and this time it will process the new request.

If you see a scenario in which a new request isn't processed ASAP, please let me know.

> I can see only one thread locks the c. v., so the lock doesn't control the c. v. access.

FWIW, the main thread _does_ lock the CV in `_wait_between_polls`.

cvat-cli/src/cvat_cli/_internal/agent.py
SpecLad: I was actually thinking about adding another media policy, where the dataset would only keep one chunk downloaded at a time. (Or maybe two, one current and one prefetched.) That would solve the media storage problem and simplify the cache cleanup code. I don't think caching all chunks is really that useful for an agent, because the entire cache is invalidated when a task's `updated_at` increases.

(However, that only concerns batch requests, so I left it out of this PR.)

cvat-cli/src/cvat_cli/_internal/agent.py
zhiltsov-max: Ok, maybe then it makes sense to leave an explanatory comment, it's up to you. I'd personally prefer just use the default solution without having to keep all this analysis in mind.

cvat-cli/src/cvat_cli/_internal/agent.py
zhiltsov-max: We can already use `/api/tasks|jobs/{id}/meta .chunks_updated_date` for finer cache management.

cvat-cli/src/cvat_cli/_internal/agent.py
SpecLad: Yeah, I suppose you're right. I placed the access under the lock.

cvat-cli/src/cvat_cli/_internal/agent.py
SpecLad: Hmm, interesting. That could be quite useful for the dataset API.

However, I still would like to move away from caching chunks on the agent in the future. If we want to stay within a certain storage limit (and I think we do), then we won't be able to store all chunks for a task if it's big enough. And then we'd kinda have to fetch chunks on-demand anyway. I suppose it's possible to cache _some_ chunks within a given limit and load others on-demand, but it feels too complicated.

zhiltsov-max (approved):

PR #9376, opened 2025-04-30
Automatically set a keyframe when enabling a previously disabled tracker

klakhov:
### Motivation and context

When we are using AI tracker, we have ability to stop tracking for several frames. Then, when we enable it back tracking starts only if we manually adjusted track position (the keyframe was set). For some cases - when object doesnt really move it may be tedious, so this pr adds enabling keyframe automatically on tracking enabling.
tracking-stop2

### Checklist

### License


bsekachev (approved):
Maybe it is a good idea to call `fetchAnnotations();` in objectState.save().then(() => { ... })
Otherwise the object may not be saved if there are "real" async processes behind.

PR #9384, opened 2025-05-01
SDK: Deprecate server URL scheme autodetection

SpecLad:
### Motivation and context

This feature is fundamentally insecure. If a user makes use of this, then a MitM could trivially block the HTTPS connection attempt, causing the SDK to downgrade to HTTP. Even in the absence of an active attacker, a transient error could cause a downgrade, leading to request details and credentials being exposed in cleartext.

In addition, this feature creates a usability problem, because if the autodetection fails, the user is given no information about the cause. This could be fixed, but given the main issue, I don't think it's worthwhile.

I don't think we can just remove this feature immediately, given that the SDK has worked this way for a long time, and there are certainly people relying on it. Therefore, in this patch I just add a big warning, and recommend specifying the scheme explicitly. In a future release, we can drop the autodetection.

### How has this been tested?

Manual testing.

### Checklist

### License

### DIFF
--- a/cvat-sdk/cvat_sdk/core/client.py
+++ b/cvat-sdk/cvat_sdk/core/client.py
@@ -77,10 +77,10 @@ class Client:
         config: Optional[Config] = None,
         check_server_version: bool = True,
     ) -> None:
-        url = self._validate_and_prepare_url(url)
-
         self.logger = logger or logging.getLogger(__name__)
         """The root logger"""
+
+        url = self._validate_and_prepare_url(url)
 
         self.config = config or Config()
         """Configuration for this object"""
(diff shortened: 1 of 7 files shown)


cvat-sdk/cvat_sdk/core/client.py
```diff
@@ -167,7 +165,22 @@ def _detect_schema(cls, base_url: str) -> str:
...
+                        return True
+            return False
+
+        if attempt("https"):
+            return "https"
+
+        self.logger.warning(
+            "Failed to connect to the server using HTTPS; will attempt HTTP instead"
+        )
+        self.logger.warning(
```
zhiltsov-max, line 177: Can we keep it only for localhost?

SpecLad: I'd rather not; it would make the behavior too magical for my taste. Moreover, for SDK usage I don't think it makes that much difference - if you're writing code, it's not too hard to add `http:`.

I will make the default CLI URL `http://localhost`, though, which should make it easier to use the CLI (I meant to do this to begin with, I just forgot 🙂).

zhiltsov-max (approved):

PR #9395, opened 2025-05-06
Documentation: Add Google Analytics tag and cookie consent banner

paveltovchigrechko:
### Motivation and context

We would like to gather user statistics for the documentation website. This PR add Google Analytics script and a cookie banner to comply with GDPR requirements. This PR is a proposal to discuss. 

How it should work in general:

1. When a user comes to the documentation website for the first time, we must display a cookie banner with options to accept and reject the analytics cookie.
2. While the banner is present (user yet to make their choice), we must not inject the tracking script on the page.
3. If user accepts tracking, we can proceed to add the script for analytics for all documentation pages for this user. The next time, the same user comes to the website we recognize them via cookie (if they did not use incognito mode) and we don't display the banner.
4. If user rejects cookie, we don't add analytics script. When such a user returns, we treat them as a newcomer, and we display the banner again.
5. We have a new link in the footer that allows calling the banner. If user selects `Reject`, we clear the previously set cookie and remove the script from the page. If user selects `Accept` we do the same as on the step 3.

Some notes about the solution, potential topics for discussion, and doubts:

- Hugo allows adding Google analytics via configuration file (see [Hugo docs](https://gohugo.io/methods/site/config/#services)), but this leads to the script being injected on the website build. This way when a user comes to the website for the first time, the script is already on the page. This is not correct. So, the solution is to have a separate template with all code related to GA tag and the banner. It allows to control the script injection basing on the user choice about cookies. We also move the GA tag to Github workflow as a variable. It is used on the production site. For local development, we have a default tag in the configuration file (`config.toml`).
- <strike>Following the previous: the GA tag is hard coded in the template. It may be not good for the code maintenance.</strike>UPDATE: now the GA tag is hardcoded in the Github workflow only (`docs.yml`).
- The template with cookie banner is called through the footer template assuming we have footer on every documentation page. Subject for discussion.
- All CSS are subjects for discussions.
- To avoid analytics pollution when running the website locally, the cookie banner HTML leverages Hugo native production check. In order to test this PR locally, you should run **hugo server -e production**. This will use the default GA tag value from the `config.toml`.
- <strike>The solution does not provide a way to reset the cookies via website interface. I assume users can do it through the browser tools if necessary. </strike>UPDATE: the solution add a "Cookie settings" link in the footer. When user selects the link, cookie banner appears. If user selects `Reject`, we clear the previously set cookies.
- The solution shows banner for all website visitors regardless of their location. I don't think we should differentiate them, but it's also a subject for discussion.

### How has this been tested?

1. Run `GA_TRACKING_ID="UA-000000000-0" HUGO_ENV=production hugo server --environment production --bind 0.0.0.0` locally. 
2. Open the site

    - Check browser console for any JavaScript errors: should not be any
    - Check if GA_TRACKING_ID is set: Open browser dev tools → Console → type window.GA_TRACKING_ID (should return 'UA-000000000-0'
    - Check if element exists: In console type document.getElementById('cookie-consent-banner'): should return the element
    - Verify localStorage: Check Application tab → Local Storage for cookieConsent: should be no cookies

3. Select `Accept`: Application => localStorage should show cookieConsent: "true"
4. Clear the application data, select `Reject`: Application => localStorage should show cookieConsent: "rejected"

To test it with network requests: in `injectGAScript` function (`cookie-consent.js`) comment lines 28-31, in HTML (`cookie-consent.html`) comment lines 5 and 27. Run the server again locally with `GA_TRACKING_ID="UA-000000000-0" HUGO_ENV=production hugo server --environment production --bind 0.0.0.0`.

### Checklist

### License


SpecLad:
> Following the previous: the GA tag is hard coded in the template. It may be not good for the code maintenance.

> To avoid analytics pollution when running the website locally, [...]

It seems to me that there's a simple way to solve both of these problems:

1. Use [`os.Getenv`](https://gohugo.io/functions/os/getenv/) to get the tag from an environment variable instead of hardcoding it in the source.
2. If the environment variable is empty, skip all the cookie-related code.
3. Define the environment variable in the GitHub workflow.
4. For development, set the variable to a fake tag value.

site/layouts/partials/cookie-consent.html
```diff
@@ -0,0 +1,193 @@
+<style>
+#cookieConsentBanner {
```
SpecLad: I would suggest to move the inline CSS/JS code to separate files; that way, our linters can check them.

site/layouts/partials/cookie-consent.html
```diff
@@ -0,0 +1,193 @@
...
+  #cookieConsentBanner button {
+    flex: 1 1 48%;
+    margin: 0;
+    padding: 0.6em 1em;
+  }
+}
+</style>
+<div id="cookieConsentBanner">
+  <p>
+    We use cookies to improve functionality and analyze traffic. Select "Accept" to consent or "Reject" to use only essential cookies.
```
SpecLad: I'm not sure, but I think we have to inform the users of what 3rd-party analytics provider(s) we're using here.

site/layouts/partials/cookie-consent.html
```diff
@@ -0,0 +1,193 @@
...
+  document.head.appendChild(s);
+
+  var inline = document.createElement('script');
+  inline.innerHTML = `
+    window.dataLayer = window.dataLayer || [];
+    function gtag(){dataLayer.push(arguments);}
+    gtag('js', new Date());
+    gtag('config', 'G-GVSBK1DNK5');
+  `;
+  document.head.appendChild(inline);
```
SpecLad: It seems that all this does is run the code in the `script` element. But then why do you even need the element? You can just run the code directly.

SpecLad:
> The solution does not provide a way to reset the cookies via website interface. I assume users can do it through the browser tools if necessary.

I'm fairly sure that users have to be able to change their cookie preferences without going through the browser tools. And it seems doable - you could add a link to the footer that shows the dialog again.

site/layouts/partials/cookie-consent.html
paveltovchigrechko: Moved CSS to `site/assets/scss/_cookie-consent.scss`; this is imported via `_variables-project.scss` (just as `_custom.scss`).
Also, moved JavaScript code to `site/assets/js/cookie-consent.js`.

site/layouts/partials/cookie-consent.html
paveltovchigrechko: Rewrote the `injectGAScript` function (now in the `site/assets/js/cookie-consent.js`; now it runs the code directly without creating an element.

paveltovchigrechko:
> It seems to me that there's a simple way to solve both of these problems:

> Use [os.Getenv](https://gohugo.io/functions/os/getenv/) to get the tag from an environment variable instead of hardcoding it in the source.
> If the environment variable is empty, skip all the cookie-related code.
> Define the environment variable in the GitHub workflow.
> For development, set the variable to a fake tag value.

So, here is my updated solution:
1. GitHub Workflow (docs.yml): Sets `GA_TRACKING_ID=G-GVSBK1DNK5` environment variable
2. Hugo Config (config.toml): Provides `ga_tracking_id` parameter for the tag default value (empty line)
3. Cookie Consent HTML (line 4): Only shows banner if both production mode AND GA tracking ID are configured

> I'm fairly sure that users have to be able to change their cookie preferences without going through the browser tools. And it seems doable - you could add a link to the footer that shows the dialog again.

Added a footer link ("Cookie settings") that invokes the banner and allows changing user preferences regarding cookie usage.

site/layouts/partials/cookie-consent.html
paveltovchigrechko: Let's add `We use Google Analytics cookies...`

site/assets/js/cookie-consent.js
```diff
@@ -0,0 +1,149 @@
...
+        // Clear dataLayer
+        if (window.dataLayer) {
+            window.dataLayer = [];
+        }
+
+        // Remove GA script if present
+        var gaScript = document.getElementById('ga-script');
+        if (gaScript) {
+            gaScript.remove();
+        }
```
SpecLad: I'm not sure this would actually do anything if the script had already been loaded... IMO, if the script already exists, the only safe option is to reload the page. (And if it doesn't exist, we don't need to do anything.)

site/assets/js/cookie-consent.js
```diff
@@ -0,0 +1,149 @@
...
+
+    /**
+     * Handle cookie acceptance
+     */
+    window.acceptCookies = function() {
+        var previousConsent = localStorage.getItem('cookieConsent');
+        localStorage.setItem('cookieConsent', 'true');
+        document.getElementById('cookie-consent-banner').style.display = 'none';
+
+        // Only inject GA if not already injected or if this is a fresh consent
```
SpecLad: Isn't this redundant? `injectGAScript` already checks if the script tag exists.

site/assets/js/cookie-consent.js
```diff
@@ -0,0 +1,149 @@
...
+            injectGAScript();
+        }
+    };
+
+    /**
+     * Handle cookie rejection
+     */
+    window.rejectCookies = function() {
+        var previousConsent = localStorage.getItem('cookieConsent');
+        localStorage.setItem('cookieConsent', 'rejected');
```
SpecLad: Your values for `cookieConsent` are `true` and `rejected`. Please make them consistent.

site/assets/js/cookie-consent.js
```diff
@@ -0,0 +1,149 @@
...
+        document.getElementById('cookie-consent-banner').style.display = 'none';
+
+        // If user previously accepted but now rejects, clear analytics cookies
```
SpecLad, line 117: This check seems redundant too.

.github/workflows/docs.yml
```diff
@@ -56,6 +56,7 @@ jobs:
           PATH="$PWD/hugo:$PATH" python site/build_docs.py
         env:
           HUGO_ENV: production
+          GA_TRACKING_ID: G-GVSBK1DNK5
```
SpecLad: I think the proper term for this is "[Google tag ID](https://developers.google.com/tag-platform/gtagjs#add_the_google_tag_to_your_website)".

site/config.toml
```diff
@@ -94,6 +99,10 @@ section = ["HTML", "print"]
 cvat_ai_terms_of_use = "https://www.cvat.ai/terms-of-use"
 cvat_ai_privacy_notice = "https://www.cvat.ai/privacy"
 
+# Google Analytics configuration
+# GA_TRACKING_ID environment variable takes precedence
+ga_tracking_id = "UA-000000000-0"  # Test tracking ID for local development
```
SpecLad: Hugo already has a setting dedicated to this (https://gohugo.io/templates/embedded/#configuration-google-analytics), so why not use that?

site/assets/js/cookie-consent.js
```diff
@@ -0,0 +1,149 @@
+/**
+ * Cookie Consent Banner JavaScript
+ * Copyright (C) 2025 CVAT.ai Corporation
```
SpecLad:
```suggestion
 * Copyright (C) CVAT.ai Corporation
```

We don't put years in copyright notices.

site/layouts/partials/cookie-consent.html
```diff
@@ -0,0 +1,27 @@
+{{/* Cookie Consent Banner */}}
+{{/* Only show in production and if GA tracking ID is configured */}}
+{{/* Styles are imported via _custom.scss */}}
+{{ $gaTrackingId := getenv "GA_TRACKING_ID" | default .Site.Params.ga_tracking_id -}}
+{{ if and hugo.IsProduction $gaTrackingId -}}
```
SpecLad: It doesn't seem like the second check is necessary here, since the default tracking ID always exists.

site/assets/scss/_variables_project.scss
```diff
@@ -9,6 +9,7 @@ Add styles or override variables from the theme here.
 */
 
 @import 'custom';
+@import 'cookie-consent';
```
SpecLad: It seems to me like this should be in `_styles_project.scss`, since it's not a variable. See [Project style files](https://www.docsy.dev/docs/adding-content/lookandfeel/#project-style-files).

site/assets/js/cookie-consent.js
paveltovchigrechko: True. I removed the check in this function.

site/assets/js/cookie-consent.js
paveltovchigrechko: Changes the values for `cookieConsent` to `true` and `false`.

site/assets/js/cookie-consent.js
paveltovchigrechko: I don't think so. This check prevents unnecessary work: if the user never accepted cookies, there's no point in trying to clear analytics cookies that were never set in the first place.

.github/workflows/docs.yml
paveltovchigrechko: Renamed the variable to `GOOGLE_TAG_ID`

site/layouts/partials/cookie-consent.html
paveltovchigrechko: Correct. We don't need this check, I removed it.

site/assets/js/cookie-consent.js
paveltovchigrechko: Not sure if I understand your concerns correctly, but here's the idea.
 
The goal here is not to "unload" JavaScript that's already run (which is impossible anyway), but to stop future tracking events (by clearing window.dataLayer), remove tracking cookies (cookie clearing), and prevent script re-injection (element removal). 

Already loaded script doesn't need "stopping". Once a Google Analytics script loads, it registers itself globally. Removing the script element doesn't "unload" the already-executed code, but that's intentional. We handle the loaded state by clearing `window.dataLayer` and cookies - that's the actual cleanup needed.

As for page reload, it looks as poor UX decision. Users would lose their current page position, the reload may be slow.

site/assets/js/cookie-consent.js
SpecLad: I mean... sure, but the work is so miniscule that it seems pointless to prevent it.

site/assets/js/cookie-consent.js
SpecLad: Are you certain that clearing `dataLayer` will prevent further requests to the Google Analytics servers? If so, then sure, this implementation is fine. (Although in that case removing the script seems unnecessary anyway.)

paveltovchigrechko: I searched for Google Analytics documentation, and found this: https://developers.google.com/tag-platform/security/guides/privacy#turn-off-analytics. So, answering your question: clearing just `dataLayer` is not enough. According to the documentation, we have to set `window['ga-disable-GA_MEASUREMENT_ID']` to `true`. I updated the `clearAnalyticsCookie` function.

site/assets/js/cookie-consent.js
paveltovchigrechko: Let's leave this check as it serves its small purpose.

SpecLad: Okay.

site/assets/js/cookie-consent.js
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+        // Clear dataLayer
+        if (window.dataLayer) {
+            window.dataLayer = [];
+        }
+
+        // Properly disable Google Analytics tracking
+        if (window.GOOGLE_TAG_ID) {
+            window['ga-disable-' + window.GOOGLE_TAG_ID] = true;
```
SpecLad, line 88: What if the user disables cookies and then enables again? We should delete this property in `injectGAScript`.

site/assets/js/cookie-consent.js
```diff
@@ -0,0 +1,154 @@
...
+        // Check if GA tracking ID is available
+        if (!window.GOOGLE_TAG_ID) {
+            console.log('GA tracking ID not configured');
+            return;
+        }
+
+        if (document.getElementById('ga-script')) return;
+
+        // Re-enable Google Analytics tracking if it was previously disabled
+        if (window.GOOGLE_TAG_ID) {
```
SpecLad: This condition is redundant - you have an earlier `if` that returns from the function if the tag ID is not set.

SpecLad (approved):

PR #9032, opened 2025-01-31
Added automatic saving for settings and persistent storage for image filter settings

klakhov:
### Motivation and context

- Removed `save` button on the settings modal, now settings are auto saved
- Added serialization for image processing filters so they can be restored automatically

### Checklist

- [x] I submit my changes into the `develop` branch
- [ ] I have created a changelog fragment 
- [ ] I have updated the documentation accordingly
- [ ] I have added tests to cover my changes
- [x] I have linked related issues (see [GitHub docs](
  https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword))

### License

- [x] I submit _my code changes_ under the same [MIT License](
  https://github.com/cvat-ai/cvat/blob/develop/LICENSE) that covers the project.
  Feel free to contact the maintainers if that's a concern.

PR #9397, opened 2025-05-06
[Refactoring] Improve shourtcuts action

klakhov:
### Motivation and context

The `restoreSettingsAsync` function has become quite complex, especially the section with four nested loops. In this PR, I’ve refactored the logic by extracting parts into separate functions to improve readability and maintainability.

Related #9032

### Checklist

- ~~[ ] I have created a changelog fragment ~~

### License

### DIFF
--- a/cvat-ui/src/utils/conflict-detector.ts
+++ b/cvat-ui/src/utils/conflict-detector.ts
@@ -2,7 +2,7 @@
 //
 // SPDX-License-Identifier: MIT
 
-import { isEqual } from 'lodash';
+import { isEqual, cloneDeep } from 'lodash';
 import { registerComponentShortcuts } from 'actions/shortcuts-actions';
 import { KeyMap, KeyMapItem } from './mousetrap-react';
 import { ShortcutScope } from './enums';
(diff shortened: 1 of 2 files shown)


bsekachev:
It seems that something might have been affected by the recent refactoring.
Could you try adding an intentionally conflicting shortcut to the local storage (for example, another F1 sequence that would conflict with the global SHOW SHORTCUTS)?
In the develop branch, one of the F1 shortcuts is removed in such cases, but with this patch, it appears that isn't happening.

cvat-ui/src/utils/conflict-detector.ts
```diff
@@ -161,3 +161,36 @@ export function unsetExistingShortcuts(
...
+        const conflicts = conflictDetector({ [key]: currValue }, shortcutsKeyMap);
+        if (!conflicts) return;
+
+        Object.keys(conflicts).forEach((conflictingKey) => {
+            resultMap[conflictingKey].sequences = removeConflictingSequences(
+                resultMap[conflictingKey].sequences,
+                conflicts[conflictingKey].sequences,
+            );
+        });
+    });
```
bsekachev, line 193: Let's consider:

- wrongShortcut to be a shortcut which has conflicts with any other shortcuts
- conflictingShortcut to be a shortcut which conflicts with wrong shortcut

E.g. I set `TOGGLE_ANNOTATION_PAGE` to `f1` in local storage
And we have `SWITCH_SHORTCUTS` also having `f1` sequence

`TOGGLE_ANNOTATION_PAGE` in this scheme is `wrongShortcut`
`SWITCH_SHORTCUTS` is conflicting shortcut

The problem I see: we are filtering sequences of the conflictingShortcut (SWITCH_SHORTCUTS), however it would be more correct to filter sequences of wrongShortcut (TOGGLE_ANNOTATION_PAGE)

bsekachev (approved):
I see several problems in the current approach which probably not related to this refactoring, but please, check it in advance. 

The first one has described in the code. 

The second one: when we uploading shortcuts from local storage, we are looking for conflicts between **local storage shortcuts** and **current keymap** in the application. The problem here is that actually shortcuts from local storage may not have any conflicts between each other, when there may be conflicts between local storage shortcuts and current keymap shortcuts, which are basically just default values defined in the code (however these conflicts just do not make sense as we are going to update all sequences in current keymap from local storage values)

Speaking "update all sequences" I mean, in general, that all sequences are usually re-written, however there are also corner cases when application may have new shortcuts not saved in the local storage, or part shortcuts from local storage have been removed. This case will not work correctly as well.

As consequence, we may see several warnings in the browser console: 
image

First conflict detector removes sequence for SWITCH_SHORTCUTS, leaving TOGGLE_ANNOTATION_PAGE equal to `f1`. 
Then in reducer it again runs conflict detector and find the conflict again.


Refactoring LGTM, however the feature still needs some improvements.

PR #9429, opened 2025-05-16
Added date picker and service changes for backend

Obaid005:
### Checklist

### License

### DIFF
--- a/cvat-ui/package.json
+++ b/cvat-ui/package.json
@@ -5,7 +5,7 @@
   "main": "src/index.tsx",
   "scripts": {
     "build": "webpack --config ./webpack.config.js",
-    "start": "webpack serve --env API_URL=http://localhost:7000 --config ./webpack.config.js --mode=development",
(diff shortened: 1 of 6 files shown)


klakhov (commented):
Hi @Obaid005,

I see the goal here is to enable filtering tasks by `created_date`, which makes sense.

However, I’m wondering why we need separate `created_from` and `created_to` fields. We can already achieve the same behavior using comparison operators:

* `created_date >= ...` works as `created_from`
* `created_date <= ...` works as `created_to`

Also, is there a particular reason for introducing a separate date picker component? The existing filter UI already includes date picker support—see the `updated_date` filter as an example:

image

It might be more consistent (and simpler) to integrate `created_date` using the same approach.

klakhov:
It seems we already have such integration with `created_date` support: #9429 8992

klakhov:
I will close this PR as there is working implementation in another one

PR #9439, opened 2025-05-19
Add support for AWS environmental credential chain for AWS storage providers

snelson44:
### Motivation and context
Currently there is no way to use the environmental credentials chain for aws cloud storages. You must either explicitly set your credentials or cvat will assume you are attempting to do anonymous access and signing requests will be disabled entirely. I added the ability for the credentials type to be specified and passed on to the aws provider. If anonymous is requested, then signing requests will still be disabled. If credentials are given, they will be used to access aws, if credentials are not explicitly given, boto3 will look for them in the environment.

### How has this been tested?
Added rest api tests for adding a cloud storage. One test to show anonymous access was still supported, and a second test to show that the aws provider could pull creds from the environment

### Checklist

### License

### DIFF
--- a/ai-models/agents_deployment/helm/Chart.yaml
+++ b/ai-models/agents_deployment/helm/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
-name: cvat-sam2-agent
-description: A chart to deploy CVAT SAM2 Agent.
+name: cvat-agent
+description: A chart to deploy CVAT Agents.
 type: application
-version: 0.1.0
+version: 0.2.0
 appVersion: "1.0.4"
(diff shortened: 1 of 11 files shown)


snelson44:
> Hi @snelson44, thank you for the contribution.
> 
> There are a few concerns with this approach:
> 
> 1. You haven't introduced a new credential type. This means it will not work when configuring cloud storage via UI (there are 2 credential types for AWS S3 compatible buckets: key/secret key pair that would not work without key/secret key on UI, anonymous access - env variables are not used by the newly introduced server logic).
> 2. There should be one more setting parameter that enables/disables this functionality on the server (by default it should be disabled). Since it is a common code we cannot allow by default the use of the shared env variables/AWS profile configured on the host.

Added credentials type and setting parameter @Marishka17

UzairQamarxyz:
Is there any update on this?

PR #9448, opened 2025-05-22
Measure content_size for Data and Assets

bsekachev:
### Motivation and context
- Updated models to store content_size
- Added measure function
- Added command to apply on existing objects

### Checklist

### License


cvat/apps/engine/management/commands/init_content_size.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def handle(self, *args, **options):
+        batch_size = 1000
+        for Model, get_directory_path, get_queryset in [
+            (Asset, lambda asset: asset.get_asset_dir(), lambda: Asset.objects),
+            (Data, lambda data: data.get_upload_dirname(), lambda: Data.objects.exclude(size=0)),
+        ]:
+            total = 0
+            self.stdout.write(f"Started for {Model.__name__}")
+            while True:
```
SpecLad, line 22:
```suggestion
            for db_objects in take_by(get_queryset().filter(content_size=None).iterator(), batch_size):
```

(`take_by` is in `cvat.apps.engine.utils`)

cvat/apps/engine/management/commands/init_content_size.py
```diff
@@ -0,0 +1,42 @@
...
+                        self.stdout.write(f"Finished for {Model.__name__}")
+                        break
+                    paths = [get_directory_path(db_object) for db_object in db_objects]
+                    sizes = get_paths_sizes(paths)
+                    for i, path in enumerate(paths):
+                        if isinstance(sizes[path], int):
+                            db_objects[i].content_size = sizes[path]
+                        else:
+                            self.stderr.write(f"Failed to get size of {path}: {str(sizes[path])}")
+                            db_objects[i].content_size = 0
```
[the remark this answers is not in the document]
bsekachev: Alright

cvat/apps/engine/serializers.py
```diff
@@ -3499,6 +3505,51 @@ class AssetWriteSerializer(WriteOnceMixin, serializers.ModelSerializer):
...
+                    for chunk in bytes_or_file.chunks():
+                        destination.write(chunk)
+
+        try:
+            av_scan_paths(dirname)
+            size_or_error = get_paths_sizes([dirname])[dirname]
+            if not isinstance(size_or_error, int):
+                raise size_or_error
+            serializer = AssetWriteSerializer(data=data | {"guide_id": guide_id})
+            serializer.is_valid(raise_exception=True)
```
SpecLad, line 3545: This usage of `AssetWriteSerializer` makes no sense. You've already saved the file to the disk, why would you only validate now? The view should first validate the request body, and only then save.

cvat/apps/engine/serializers.py
```diff
@@ -3499,6 +3505,51 @@ class AssetWriteSerializer(WriteOnceMixin, serializers.ModelSerializer):
     filename = serializers.CharField(required=True, max_length=MAX_FILENAME_LENGTH)
     guide_id = serializers.IntegerField(required=True)
 
+    @staticmethod
+    def write_asset(
```
SpecLad, line 3509: This method seems like it should be a custom `save` implementation.

cvat/apps/engine/backup.py
```diff
@@ -114,26 +114,21 @@ def _read_annotation_guide(zip_object, guide_filename, assets_dirname):
...
-            markdown = markdown.replace(f'{name}', f'/api/assets/{asset_serializer.instance.pk}')
-            path = os.path.join(settings.ASSETS_ROOT, str(asset_serializer.instance.uuid))
-            os.makedirs(path)
-            with open(os.path.join(path, basename), 'wb') as destination:
-                destination.write(data)
+        asset_serializer = AssetWriteSerializer.write_asset(
+            owner,
+            guide_serializer.instance.id,
+            data,
+            os.path.basename(name)
```
SpecLad, line 129: I would suggest using `SimpleUploadedFile` here. That way, `AssetWriteSerializer` will only need to handle one input data type. In addition, `SimpleUploadFile` will automatically sanitize the name.

cvat/apps/engine/serializers.py
```diff
@@ -3499,6 +3505,51 @@ class AssetWriteSerializer(WriteOnceMixin, serializers.ModelSerializer):
...
+                if any(map(lambda x: x > settings.ASSET_MAX_IMAGE_SIZE, image.size)):
+                    scale_factor = settings.ASSET_MAX_IMAGE_SIZE / max(image.size)
+                    image = image.resize((map(lambda x: int(x * scale_factor), image.size)))
+                image.save(filename)
+            else:
+                with open(filename, "wb") as destination:
+                    for chunk in bytes_or_file.chunks():
+                        destination.write(chunk)
+
+        try:
```
SpecLad, line 3539: This `try` block covers too little. All code after `os.makedirs` should be covered, since the directory may need to be cleaned up.

cvat/apps/engine/serializers.py
```diff
@@ -3499,6 +3505,51 @@ class AssetWriteSerializer(WriteOnceMixin, serializers.ModelSerializer):
...
+            size_or_error = get_paths_sizes([dirname])[dirname]
+            if not isinstance(size_or_error, int):
+                raise size_or_error
+            serializer = AssetWriteSerializer(data=data | {"guide_id": guide_id})
+            serializer.is_valid(raise_exception=True)
+            serializer.save(content_size=size_or_error, owner=owner)
+            return serializer
+        except:
+            os.remove(filename)
+            os.rmdir(dirname)
```
SpecLad, line 3550: I would suggest `shutil.rmtree` here, since that'll work whether the directory is empty or not.

bsekachev: For me such approach seems a little safer then rmtree from shutil

cvat/apps/engine/utils.py
```diff
@@ -357,6 +358,35 @@ def take_by(iterable: Iterable[_T], chunk_size: int) -> Generator[list[_T], None
...
+    if not paths:
+        return {}
+
+    chunk_size = 100
+    sizes: Dict[str, int] = {}
+
+    for chunk in take_by(paths, chunk_size):
+        # the function must not be called with uncontrolled input
+        result = subprocess.run( # nosec
+            ["du", "-s"] + chunk,
```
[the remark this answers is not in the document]
bsekachev: The main reason why I chosen `du` is that `Path.glob` + syscall for each file is extremely slow in comparison with `du`. 

`du` is a default utility in many linux distributives, so, IMO it seems to be a good trade-off

And `du` may return size in bytes also, however I am not sure we really need it. 

Filesystem usage is exactly the thing we want to control, so, I am not sure that just raw file size is better.

cvat/apps/engine/serializers.py
bsekachev: Yes, reimplemented

cvat/apps/engine/task.py
```diff
@@ -779,6 +779,17 @@ def update_status(msg: str) -> None:
 
     av_scan_paths(upload_dir)
 
+    # If something was uploaded to the raw directory, update content_size in the Data model
+    # raw_data_dir and upload_dir are not always the same
+    raw_data_dir = db_data.get_upload_dirname()
+    if os.path.exists(raw_data_dir):
+        size_or_error = get_paths_sizes([raw_data_dir])[raw_data_dir]
+        if not isinstance(size_or_error, int):
+            slogger.glob.warning(size_or_error, exc_info=True)
```
SpecLad, line 788: Shouldn't we about the creation if we couldn't determine the size? It seems like a sign that something went wrong.

cvat/apps/engine/utils.py
```diff
@@ -357,6 +359,29 @@ def take_by(iterable: Iterable[_T], chunk_size: int) -> Generator[list[_T], None
         yield batch
 
 
+def get_paths_sizes(paths: list[str]) -> dict[str, int | OSError | ValueError]:
+    sizes: dict[str, int | Exception] = {}
+    for path in paths:
```
SpecLad, line 364: As far as I can see, there's no reason for this function to take a list of paths at this point. It should just take a path and return `int` (or raise an exception). It would make both the function and the calling code simpler.

cvat/apps/engine/views.py
```diff
@@ -2663,47 +2662,32 @@ def get_serializer_class(self):
...
+        guide_id = request.data.get("guide_id")
+        db_guide = AnnotationGuide.objects.prefetch_related("assets").get(pk=guide_id)
         if db_guide.assets.count() >= settings.ASSET_MAX_COUNT_PER_GUIDE:
-            raise ValidationError(f'Maximum number of assets per guide reached')
+            raise ValidationError(f"Maximum number of assets per guide reached")
 
-        serializer = self.get_serializer(data={
-            'filename': file.name,
-            'guide_id': guide_id,
+        serializer = AssetWriteSerializer(data={
```
SpecLad, line 2680: I might be missing something here, but why not use `data=request.data`? You'd need to rename the `asset_file` field to just `file` to match the input, but otherwise it should work. That would also make the `if not file` check unnecessary, since the serializer already checks that.

SpecLad: > request.data.guide_id is a string - validation failed

Isn't that a bug in the UI, then? It shouldn't be a string in the first place.

> owner_id shouldn't be a field in the first place - I think it should, as it is a model field.

I don't think that's relevant. The purpose of the serializer is to validate input data, and `owner_id` is not part of the input. So it doesn't need to be validated and therefore doesn't need to be a serializer field.

(It's more complicated when the serializer is used for read requests, but that's not the case here.)

bsekachev: > Isn't that a bug in the UI, then? It shouldn't be a string in the first place.

IDK how exactly this magic works in django, but after updating scheme it worked
I didn't touch UI, it sends multipart/form-data with integer

cvat/apps/engine/views.py
```diff
@@ -2644,7 +2633,7 @@ class AssetsViewSet(
     queryset = Asset.objects.select_related(
         'owner', 'guide', 'guide__project', 'guide__task', 'guide__project__organization', 'guide__task__organization',
     ).all()
-    parser_classes=_UPLOAD_PARSER_CLASSES
+    parser_classes = api_settings.DEFAULT_PARSER_CLASSES + [MultiPartParser]
```
SpecLad, line 2636:
```suggestion
    parser_classes = [MultiPartParser]
```

`FileField` doesn't work with JSON, and AFAICS there aren't any other endpoints that take request bodies, so you might as well disable JSON entirely.

cvat/apps/engine/views.py
SpecLad: Okay, nice!

cvat/apps/engine/task.py
SpecLad: It's true that this isn't critical for CVAT to function, but isn't the point of this feature to be able to accurately track users' use of storage space? If we allow tasks to be created without filling in `content_size`, then some storage will be unaccounted for.

SpecLad (approved):

cvat/apps/engine/views.py
SpecLad: > IDK how exactly this magic works in django

FWIW, `IntegerField` does work with strings (in fact, it will cast the input value to str before casting it back to int):

https://github.com/encode/django-rest-framework/blob/33d59fefaa5af04f4bed9312239eb1e5e6def2a2/rest_framework/fields.py

PR #9496, opened 2025-06-02
SDK: improve mask utilities

SpecLad:
### Motivation and context

* Add `decode_mask` (the reverse operation of `encode_mask`).

* Make the bounding box in `encode_mask` optional.

* Add a few more tests.

These additions are needed for a feature I'm developing (although they seem pretty useful on their own).

### How has this been tested?

Unit tests.

### Checklist

### License


changelog.d/20250602_163511_roman_decode_mask.md
```diff
@@ -0,0 +1,9 @@
+### Added
+
+- \[SDK\] Added `decode_mask`, a utility function that creates a bitmap
+  based on the `points` array in a mask shape
+  (<https://github.com/cvat-ai/cvat/pull/9496>)
+
+- \[SDK\] `encode_mask` may now be called without an explicit `bbox`,
+  in which case the bounding box is determined automatically
```
zhiltsov-max:
```suggestion
  in which case the whole mask is encoded
```
Maybe the docstring of `encode_mask` should be changed instead, it's a little bit unclear there.

cvat-sdk/cvat_sdk/masks.py
```diff
@@ -4,21 +4,24 @@
...
 
     bitmap must be a boolean array of shape (H, W), where H is the height and
     W is the width of the image that the mask applies to.
 
-    bbox must have the form [x1, y1, x2, y2], where (0, 0) <= (x1, y1) < (x2, y2) <= (W, H).
-    The mask will be limited to points between (x1, y1) and (x2, y2).
+    bbox, if specified, must have the form [x1, y1, x2, y2],
+    where (0, 0) <= (x1, y1) < (x2, y2) <= (W, H).
+    The encoded mask will be limited to points between (x1, y1) and (x2, y2).
+    If bbox is None, all points of the bitmap will be considered.
```
zhiltsov-max:
```suggestion
    If bbox is None, the bbox will be inferred automatically from the bitmap points.
```

It's a little bit unclear what "considered" means in practice.

SpecLad: Yeah, good point. I'm hesitant to imply that the bounding box is actually included _in_ the encoded mask (it doesn't seem like a detail that a regular user needs to know about), so I tried to rephrase it differently.

zhiltsov-max: From my point, but it's not a question of whether the bbox coordinates are included in the encoded representation, it's more about which part of the mask is to be included in the RLE counts - the whole image or only its "significant part". Anyway, the updated variant answers this question.

zhiltsov-max (approved):

PR #9517, opened 2025-06-09
Update the interface for tracker function calls to support non-rectangle shapes

SpecLad:
### Motivation and context

This is a preliminary step towards supporting agent-based tracking functions.

Note that this does not add the ability to actually use trackers with non-rectangles from the UI. That'll have to be added separately.

### How has this been tested?

Manual testing with TransT and SAM2.

### Checklist

### License

### DIFF
--- a/cvat-core/src/ml-model.ts
+++ b/cvat-core/src/ml-model.ts
@@ -5,7 +5,7 @@
 
 import PluginRegistry from './plugins';
 import {
-    LabelType, ModelProviders, ModelKind,
+    LabelType, ModelProviders, ModelKind, ShapeType,
 } from './enums';
 import {
     SerializedModel, ModelParams, MLModelTip, MLModelLabel,
@@ -28,6 +28,10 @@ export default class MLModel {
 
     public get labels(): MLModelLabel[] {
         return Array.isArray(this.serialized.labels_v2) ? [...this.serialized.labels_v2] : [];
+    }
+
+    public get supportedShapeTypes(): ShapeType[] {
+        return this.serialized.supported_shape_types;
     }
 
     public get version(): number {
--- a/cvat-core/src/ml-model.ts
+++ b/cvat-core/src/ml-model.ts
@@ -30,7 +30,7 @@ export default class MLModel {
         return Array.isArray(this.serialized.labels_v2) ? [...this.serialized.labels_v2] : [];
     }
 
-    public get supportedShapeTypes(): ShapeType[] {
+    public get supportedShapeTypes(): ShapeType[] | undefined {
         return this.serialized.supported_shape_types;
     }
 
--- a/cvat/apps/lambda_manager/tests/assets/functions.json
+++ b/cvat/apps/lambda_manager/tests/assets/functions.json
@@ -4,7 +4,6 @@
       "metadata": {
         "name": "test-openvino-omz-public-yolo-v3-tf",
         "annotations": {
-          "framework": "openvino",
           "name": "YOLO v3",
           "spec": "[\n  { \"id\": 0, \"name\": \"person\" },                              \n  { \"id\": 1, \"name\": \"bicycle\" },                              \n  { \"id\": 2, \"name\": \"car\" }                              \n]\n",
           "type": "detector"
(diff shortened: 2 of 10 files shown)


cvat/apps/lambda_manager/tests/assets/functions.json
```diff
@@ -90,6 +90,25 @@
         "httpPort": 49158
       }
     },
+    "test-tracker-with-supported-shape-types": {
+      "metadata": {
+        "name": "test-tracker-with-supported-shape-types",
+        "annotations": {
+          "framework": "pytorch",
```
bsekachev: As far as I remember, `framework` was removed long time before from specs

cvat/apps/lambda_manager/views.py
```diff
@@ -247,6 +247,22 @@ def parse_attributes(attrs_spec):
         self.help_message = meta_anno.get("help_message", "")
         self.gateway = gateway
 
+        if "supported_shape_types" in meta_anno:
```
bsekachev, line 250: Backward compatibility is a good thing, but I believe we anyway should update existing tracker specs to come to unified interface somewhere in the future (I mean adding "supported_shape_types" to all trackers)

bsekachev (commented):
If we introduce general interface for trackers, maybe it is time to think now, how such interface like 

{
    "type": string;
    "points": number[];
}

is compatible with skeletons tracking, especially when some skeleton points are outside. For regular shapes it may be interpreted as None object in response["shapes"], for a skeleton, it does not seem to work.

cvat-core/src/ml-model.ts
```diff
@@ -30,6 +30,10 @@ export default class MLModel {
         return Array.isArray(this.serialized.labels_v2) ? [...this.serialized.labels_v2] : [];
     }
 
+    public get supportedShapeTypes(): ShapeType[] {
+        return this.serialized.supported_shape_types;
```
bsekachev, line 34:
```suggestion
        return this.serialized.supported_shape_types ?? [ShapeType.RECTANGLE];
```

bsekachev, line 34: In `SerializedModel` interface supported_shape_types may be undefined

So, maybe we need a default value here
If for any tracker supported_shape_types are guaranteed, maybe empty array is a good default value. 

Alternatively need to specify that supportedShapeTypes may return undefined also.

bsekachev: Ideally we probably need several classes inherited from ML model (Detector, Tracker, Interactor)
But this is about refactoring and not related to this patch. Currently I only suggest providing correct types.

cvat/apps/lambda_manager/views.py
SpecLad: I don't really see a reason to do that. The compatibility code is simple enough that I think we can maintain it indefinitely.

SpecLad:
> skeletons tracking

IMO, we should have the "shape" structure used in this interface replicate the `LabeledShape` structure from the API to the greatest extent possible. So for skeletons, I'd expect it to be something like `{"type": "skeleton", "elements": [{"type": "points", "points": [1, 2]}, ...]}`.

cvat/apps/lambda_manager/views.py
bsekachev: I did not say compatibility code is complex. And we definitely want to have it for now. 

This is just one backward compatibility, then will be one more, then another one, etc. Thus, with time, we need to maintain too many backward compatibilities. The idea is that we need to update functions to correspond to the latest interface.

And one more reason to do that is these functions are often used as examples to write new.

bsekachev:
> > skeletons tracking
> 
> IMO, we should have the "shape" structure used in this interface replicate the `LabeledShape` structure from the API to the greatest extent possible. So for skeletons, I'd expect it to be something like `{"type": "skeleton", "elements": [{"type": "points", "points": [1, 2]}, ...]}`.

Agree

cvat/apps/lambda_manager/tests/assets/functions.json
SpecLad: Oh yeah, I guess I missed this file when I did that... I went ahead and removed all frameworks.

cvat-core/src/ml-model.ts
SpecLad: I added `undefined`. IMO, it's better than adding a default, since it might help catch this method being used on non-trackers.

bsekachev (approved):

PR #9523, opened 2025-06-10
Experimental adaptive zoom algorithm applicable for touchpads

bsekachev:
### Checklist

### License

### DIFF
--- a/cvat-canvas/src/typescript/canvasModel.ts
+++ b/cvat-canvas/src/typescript/canvasModel.ts
@@ -454,14 +454,14 @@ export class CanvasModelImpl extends MasterImpl implements CanvasModel {
 
     public zoom(x: number, y: number, deltaY: number): void {
         const basicZoomCoef = 6 / 5; // historical value
-        // the higher value, means smoother the zooming
-        // we need trade-off between speed and smoothness
-        const zoomSmoother = 8;
+        // less value of adjust coef, means zoomin/zoomout smoother
+        // we need a trade-off between speed and smoothness, value 1 / 8 is good enough
+        const adjustCoef = 1 / 8;
         const oldScale: number = this.data.scale;
-        let scaleFactor = basicZoomCoef ** (-deltaY / zoomSmoother);
+        let scaleFactor = basicZoomCoef ** (-deltaY * adjustCoef);
 
         if (!this.data.configuration.adaptiveZoom) {
-            // old alogithm, just multiply to 6/5 or 5/6
+            // old alogithm, just multiplies to 6/5 or 5/6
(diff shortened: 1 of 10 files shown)


cvat-canvas/src/typescript/canvasView.ts
```diff
@@ -1706,10 +1724,23 @@ export class CanvasViewImpl implements CanvasView, Listener {
...
+                // even when the ctrl key is not pressed actually
+                return;
+            }
+
+            let { deltaY } = event;
+            // clamp too high values to avoid strong zooming
+            // high values are usually applicable to mice
+            // 8 is a good experimental value to avoid strong zooming
+            const MAX_DELTA_Y = 8;
+            deltaY = Math.abs(deltaY) > MAX_DELTA_Y ? MAX_DELTA_Y * Math.sign(deltaY) : deltaY;
```
klakhov: I think we can use clamp function from shared here

cvat-canvas/src/typescript/canvasModel.ts
```diff
@@ -450,9 +452,19 @@ export class CanvasModelImpl extends MasterImpl implements CanvasModel {
         };
     }
 
-    public zoom(x: number, y: number, direction: number): void {
+    public zoom(x: number, y: number, deltaY: number): void {
+        const basicZoomCoef = 6 / 5; // historical value
+        // less value of adjust coef, means zoomin/zoomout smoother
+        // we need a trade-off between speed and smoothness, value 1 / 8 is good enough
+        const adjustCoef = 1 / 8;
```
klakhov: Does this coef depend on `MAX_DELTA_Y` from wheel event handler?
If yes, its probably better to not write one more literal but write something like `1/COEF`

bsekachev: Not really
MAX_DELTA_Y only stricts extremely high values (like 100, 300, 1000) (if to scroll the wheel to fast)
This is smoothing coefficient

bsekachev: If to think, they both are to make zoom smoother. 
Nevertheless, they are used in different operations (clamp, division) and not related to each other

bsekachev: hmm, maybe I want to try just reduce MAX_DELTA_Y to 1 and remove this coefficient

bsekachev: however not sure it will work the same way, as when I just scroll wheel slightly, deltaY is about 3.

bsekachev: Yes, it works totally different for touchpads

PR #9527, opened 2025-06-11
Add page size for different pages

klakhov:
### Motivation and context

- The PR adds page size selector for different pages.
- Also added optimization logic into preview fetching: request for preview will only be sent when the card is in user view
- Unified card views page with scrollbars like on task page
- Changed models card design a bit so 12 of them are fit on the page like projects
- Added hook for updatedQuery to reduce code duplication

### Checklist

### License


klakhov:
@bsekachev What do you think, should we add scrollbar into pages with card views?

bsekachev:
Layout on jobs page changed if there are less than 10 cards (the same for projects page)
image

Also cannot apply any filter or search on jobs page (need to check other pages)
image

cvat-ui/src/utils/hooks.ts
```diff
@@ -205,3 +205,32 @@ export function useInstanceId(type: InstanceType): number {
...
     return +(params.tid as string);
 }
+
+interface UseUpdatedQueryOptions {
+    page?: number;
+    pageSize?: number;
+}
+
+export function useUpdatedQuery<QueryType extends {
+    page: number, pageSize?: number
```
bsekachev:
```suggestion
    page: number;
    pageSize?: number;
```

cvat-ui/src/utils/hooks.ts
```diff
@@ -205,3 +205,32 @@ export function useInstanceId(type: InstanceType): number {
...
+}>(query: QueryType, options: UseUpdatedQueryOptions = {}): QueryType {
```
bsekachev:
```suggestion
}>(defaultQuery: QueryType, options: UseUpdatedQueryOptions = {}): QueryType {
```

cvat-ui/src/components/cloud-storages-page/cloud-storages-page.tsx
```diff
@@ -27,14 +28,7 @@ export default function StoragesPageComponent(): JSX.Element {
...
 
-    const queryParams = new URLSearchParams(history.location.search);
-    const updatedQuery = { ...query };
-    for (const key of Object.keys(updatedQuery)) {
-        (updatedQuery as Indexable)[key] = queryParams.get(key) || null;
-        if (key === 'page') {
-            updatedQuery.page = updatedQuery.page ? +updatedQuery.page : 1;
-        }
-    }
+    const updatedQuery = useUpdatedQuery<CloudStoragesQuery>(query, { pageSize: 12 });
```
bsekachev: Maybe `useUpdatedQuery` is not the best name. Just `useResourceQuery<Type>(default)`?
Is there a reason to add `options` argument when `pageSize` and `page` are required in QueryType?

cvat-ui/src/components/common/preview.tsx
```diff
@@ -45,55 +45,47 @@ export default function Preview(props: Props): JSX.Element {
...
-        } if (project !== undefined) {
-            return state.projects.previews[project.id];
-        } if (task !== undefined) {
-            return state.tasks.previews[task.id];
-        } if (cloudStorage !== undefined) {
-            return state.cloudStorages.previews[cloudStorage.id];
-        } if (model !== undefined) {
-            return state.models.previews[model.id];
-        }
+        if (job !== undefined) return state.jobs.previews[job.id];
```
bsekachev: IMHO with {} code looks prettier

cvat-ui/src/components/common/preview.tsx
```diff
@@ -45,55 +45,47 @@ export default function Preview(props: Props): JSX.Element {
...
+            <div ref={ref} className={loadingClassName || ''} aria-hidden>
                 <Spin size='default' />
             </div>
         );
     }
 
     if (preview.initialized && !preview.preview) {
         return (
-            <div className={emptyPreviewClassName || ''} onClick={onClick} aria-hidden>
+            <div ref={ref} className={emptyPreviewClassName || ''} onClick={onClick} aria-hidden>
```
bsekachev: Is there a reason to put this ref here and below? (preview already downloaded)

cvat-ui/src/components/task-page/job-list.tsx
```diff
@@ -126,7 +121,7 @@ function JobListComponent(props: Props): JSX.Element {
     const [query, setQuery] = useState<JobsQuery>(updatedQuery);
     const filteredJobs = setUpJobsList(jobs, query);
     const jobViews = filteredJobs
-        .slice((query.page - 1) * PAGE_SIZE, query.page * PAGE_SIZE)
+        .slice((query.page - 1) * query.pageSize, query.page * query.pageSize)
```
bsekachev: Should not it be a part of setUpJobsList like on other views? models/requests

cvat-ui/src/utils/hooks.ts
klakhov: The first parameter isn't the default query—it's the current query from the application state, which can be modified based on user actions (like navigating between pages). The `options` object is actually used to provide default values, maybe its not the best name, so I renamed it to `defaultParams` for clarity.

The logic works as follows:
If there's something in `location.search`, we use those parameters; otherwise, we fall back to `defaultParams`. `query` can have page num that doesnt exist for example, so we reset it here with default value, as empty `location.search` appears when we re-visit `Projects, Tasks, etc` page.

cvat-ui/src/components/cloud-storages-page/cloud-storages-page.tsx
klakhov: Renamed to `useResourceQuery`. Answered about `options` in previous comment.

cvat-ui/src/components/common/preview.tsx
klakhov: Reverted.

cvat-ui/src/components/common/preview.tsx
klakhov: Removed

cvat-ui/src/components/task-page/job-list.tsx
klakhov: The issue here is that we need to filter jobs first, since sorting and filtering are handled on the UI side. To keep `setupJobList` consistent with the other setup functions, we need to introduce a separate method like `filterJobs`.
So I extracted the logic into two functions.

bsekachev:
@klakhov Please add changelog entries

PR #9528, opened 2025-06-12
Allow tasks and projects to be transferred between organizations and sandboxes

Marishka17:
Depends on https://github.com/cvat-ai/cvat/pull/9543, https://github.com/cvat-ai/cvat/pull/9535
### Motivation and context

https://github.com/user-attachments/assets/ca286e97-4c14-4202-80f7-ea95e43b2ae8

### Checklist

### License


cvat/apps/engine/models.py
```diff
@@ -437,6 +438,36 @@ class Meta:
...
+        *,
+        assignee_id: int | None = None,
+        updated_date: datetime.datetime | None = None,
+        save: bool = True,
+    ) -> None:
+        self.assignee_id = assignee_id
+        self.assignee_updated_date = updated_date or timezone.now()
+
+        if save:
+            self.save(update_fields=["assignee, assignee_updated_date"])
```
SpecLad:
```suggestion
            self.save(update_fields=["assignee", "assignee_updated_date"])
```

Although you don't seem to use `save=False` anywhere, so maybe you could just remove the `save` parameter.

cvat/apps/engine/models.py
```diff
@@ -437,6 +438,36 @@ class Meta:
...
+    )
+    assignee_updated_date = models.DateTimeField(null=True, blank=True, default=None)
+
+    class Meta:
+        abstract = True
+
+    def update_assignee(
+        self,
+        *,
+        assignee_id: int | None = None,
```
SpecLad: IMO, you should a) turn this into a positional-only parameter, and b) remove the default. Currently, `x.update_assignee()` will remove the assignee, which seems surprising.

cvat/apps/engine/models.py
```diff
@@ -437,6 +438,36 @@ class Meta:
     def touch(self) -> None:
         self.save(update_fields=["updated_date"])
 
+class AssignableModel(models.Model):
+    assignee = models.ForeignKey(
+        User, null=True, blank=True, on_delete=models.SET_NULL, related_name="assigned_%(class)ss"
```
SpecLad:
```suggestion
        User, null=True, blank=True, on_delete=models.SET_NULL, related_name="assigned_%(class)ss", related_query_name="assigned_%(class)s"
```

cvat/apps/engine/models.py
```diff
@@ -437,6 +438,36 @@ class Meta:
...
+        updated_date: datetime.datetime | None = None,
+        save: bool = True,
+    ) -> None:
+        self.assignee_id = assignee_id
+        self.assignee_updated_date = updated_date or timezone.now()
+
+        if save:
+            self.save(update_fields=["assignee, assignee_updated_date"])
+
+class AssignableModelWithoutBackRelationship(AssignableModel):
```
SpecLad: Why is this class needed? You've used it for `Issue`, but why not let it inherit from `AssignableModel`?

cvat/apps/engine/models.py
```diff
@@ -586,10 +614,8 @@ class Task(TimestampedModel, FileSystemRelatedModel):
     name = SafeCharField(max_length=256)
     mode = models.CharField(max_length=32)
     owner = models.ForeignKey(User, null=True, blank=True,
-        on_delete=models.SET_NULL, related_name="owners")
-    assignee = models.ForeignKey(User, null=True,  blank=True,
-        on_delete=models.SET_NULL, related_name="assignees")
-    assignee_updated_date = models.DateTimeField(null=True, blank=True, default=None)
+        on_delete=models.SET_NULL, related_name="owned_tasks")
```
SpecLad: It would be more consistent to use "tasks" here - consider, for example, the equivalent field in `CloudStorage` or `Asset`.

Also, as long as you're updating this field, please add a `related_query_name`.

cvat/apps/engine/models.py
Marishka17: Thanks. Yes, I do not use this option, but I still think it should be added (but probably with another default value). For instance, previously, we had a code in the private application that reset a job assignee in some cases.

cvat/apps/engine/models.py
Marishka17: Okay, agree

cvat/apps/engine/models.py
Marishka17: Just because there was no backwards relation for issues. Removed.

cvat/apps/engine/models.py
SpecLad: IMO, we should apply YAGNI here. If we end up needing this option, it can easily be added - it's only like 3 lines of code.

(Also, shouldn't we update `updated_date` here too?)

cvat-core/src/api-implementation.ts
```diff
@@ -328,15 +328,23 @@ export default function implementAPI(cvat: CVATCore): CVATCore {
...
-        return organizations;
+        const organizationsData = await serverProxy.organizations.get(filter, fullResponseData);
+        if (fullResponseData) {
+            return {
+                ...organizationsData,
+                results: organizationsData.results.map((organizationData) => new Organization(organizationData)),
+            };
+        }
+
+        return organizationsData.map((organizationData) => new Organization(organizationData));
```
klakhov: Adding a parameter like `fullResponseData`  makes the `organizations.get` call inconsistent in terms of return values. Instead, I’d suggest always returning the organization data along with some additional metadata.

We already have the `PaginatedResource` type, which allows us to return something like `Object.assign(invitations, { count: invitationsData.count })`. We could create a new type extending from that, to include fields like `count` and `next`, if such metadata is needed on the CVAT UI side.

cvat-core/src/session-implementation.ts
```diff
@@ -753,6 +756,9 @@ export function implementTask(Task: typeof TaskClass): typeof TaskClass {
             if (fields.consensus_replicas) {
                 taskSpec.consensus_replicas = fields.consensus_replicas;
             }
+            if (typeof this.organizationId !== 'undefined') {
+                taskSpec.organization_id = this.organizationId;
+            }
```
klakhov: Let’s move those lines up, right after the `subset` assignement (line 746).

cvat-ui/src/actions/organization-actions.ts
```diff
@@ -251,4 +266,24 @@ export function updateOrganizationMemberAsync(
...
+    return async (dispatch): Promise<void> => {
+        dispatch(organizationActions.getOrganizations());
+
+        try {
+            const result = await core.organizations.get(query, withHasNextPage);
+            if (withHasNextPage) {
+                dispatch(organizationActions.getOrganizationsSuccess(result.results, result.count, result.next));
+            } else {
+                dispatch(organizationActions.getOrganizationsSuccess(result, result.length));
+            }
```
klakhov: Referring back to my comment about `api-implementation.ts`: it seems that _always_ returning an extended version of `PaginatedResource` would help us avoid these conditional checks and make the code cleaner

cvat-ui/src/actions/projects-actions.ts
```diff
@@ -69,7 +71,15 @@ const projectActions = {
...
-    updateProjectFailed: (error: any) => createAction(ProjectsActionTypes.UPDATE_PROJECT_FAILED, { error }),
+    updateProjectFailed: (projectId: number, error: any, message?: string) => (
+        createAction(ProjectsActionTypes.UPDATE_PROJECT_FAILED, { projectId, error, message })
+    ),
+    openLinkedCloudStorageUpdatingModal: (project: Project) => (
+        createAction(ProjectsActionTypes.OPEN_LINKED_CLOUD_STORAGE_UPDATING_MODAL, { project })
+    ),
+    closeLinkedCloudStorageUpdatingModal: () => (
+        createAction(ProjectsActionTypes.CLOSE_LINKED_CLOUD_STORAGE_UPDATING_MODAL, { })
+    ),
```
klakhov:
```suggestion
    closeLinkedCloudStorageUpdatingModal: () => (
        createAction(ProjectsActionTypes.CLOSE_LINKED_CLOUD_STORAGE_UPDATING_MODAL)
    ),
```

cvat-ui/src/actions/projects-actions.ts
```diff
@@ -171,8 +181,13 @@ export const getProjectsPreviewAsync = (project: any): ThunkAction => async (dis
     }
 };
 
+export enum ProjectUpdateTypes {
+    UPDATE_ORGANIZATION = 'UPDATE_ORGANIZATION',
+}
```
klakhov: I dont think we need such enum per-resource. We can just keep general `UpdateTypes`

cvat-ui/src/actions/projects-actions.ts
```diff
@@ -181,7 +196,11 @@ export function updateProjectAsync(
             dispatch(projectActions.updateProjectSuccess(updatedProject));
             return updatedProject;
         } catch (error) {
-            dispatch(projectActions.updateProjectFailed(error));
+            let message = '';
+            if (updateType === ProjectUpdateTypes.UPDATE_ORGANIZATION) {
+                message = `Could not transfer the project #${projectInstance.id} to the new workspace`;
```
klakhov: Instead of constructing the exception message directly in the action, I suggest delegating that responsibility to the `notification-reducer`.

We can pass the `updateType` to the corresponding action, and then build the appropriate message in the reducer where it belongs.

```ts
updateProjectFailed: (projectId: number, error: any, updateType?: ProjectUpdateTypes) =>
    createAction(ProjectsActionTypes.UPDATE_PROJECT_FAILED, { projectId, error, updateType }),
```

Then in the notifications reducer:

```ts
case ProjectsActionTypes.UPDATE_PROJECT_FAILED: {
    const { projectId, error, updateType } = action.payload;
    let message = `Could not update [project #${projectId}](/project/${projectId})`;

    if (updateType === UpdateTypes.UPDATE_ORGANIZATION) {
        message = `Could not transfer the [project #${projectId}](/project/${projectId}) to the new workspace`;
    }

    // ...existing code...
}
```

cvat-ui/src/actions/tasks-actions.ts
```diff
@@ -350,7 +358,11 @@ export function updateTaskAsync(
             dispatch(updateTaskInState(updated));
             return updated;
         } catch (error) {
-            dispatch(updateTaskFailed(taskInstance.id, error));
+            let message = '';
+            if (updateType === TaskUpdateTypes.UPDATE_ORGANIZATION) {
```
klakhov: Same thing about notification here

cvat-ui/src/components/projects-page/actions-menu-items.tsx
```diff
@@ -58,6 +58,12 @@ export default function ProjectActionsItems(
         label: <CVATMenuEditLabel>Assignee</CVATMenuEditLabel>,
     }, 30]);
 
+    menuItems.push([{
+        key: 'edit_organization',
+        onClick: () => startEditField('organization'),
+        label: <CVATMenuEditLabel>Organization</CVATMenuEditLabel>,
+    }, 31]);
```
klakhov: We usually try to reserve weight numbers in the `*1–*9` range for plugins, so it’s better to use `40` here.

cvat-ui/src/components/select-organization-modal/select-organization-modal.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+import React from 'react';
+import Modal from 'antd/lib/modal';
+
+import { useSelector, useDispatch } from 'react-redux';
+import { CombinedState } from 'reducers';
+
+import { organizationActions } from '../../actions/organization-actions';
+import OrganizationSelector from '../selectors/organization-selector';
+import { Organization } from '../../cvat-core-wrapper';
```
klakhov: We dont need `../` in those imports

cvat-ui/src/components/select-organization-modal/styles.scss
```diff
@@ -0,0 +1,6 @@
+// Copyright (C) CVAT.ai Corporation
+//
+// SPDX-License-Identifier: MIT
+
+@import 'base';
+@import '../../styles';
```
klakhov: Seems this file doesnt contain anything useful

cvat-ui/src/components/tasks-page/actions-menu-items.tsx
```diff
@@ -85,6 +87,14 @@ export default function TaskActionsItems(menuItemsData: MenuItemsData, taskMenuP
         label: <CVATMenuEditLabel>Assignee</CVATMenuEditLabel>,
     }, 50]);
 
+    if (!projectId) {
+        menuItems.push([{
+            key: 'edit_organization',
+            onClick: () => startEditField('organization'),
+            label: <CVATMenuEditLabel>Organization</CVATMenuEditLabel>,
+        }, 51]);
```
klakhov: Same comment about numbers here, `60`

cvat-ui/src/components/projects-page/actions-menu.tsx
```diff
@@ -64,6 +72,47 @@ function ProjectActionsComponent(props: Props): JSX.Element {
...
+        ) {
+            dispatch(projectActions.openLinkedCloudStorageUpdatingModal(projectInstance));
+        } else {
+            dispatch(updateProjectAsync(projectInstance, ProjectUpdateTypes.UPDATE_ORGANIZATION)).then(stopEditField);
+        }
+    }, [projectInstance]);
+
+    const onUpdateProjectOrganization = useCallback((dstOrganization: Organization | null) => {
+        const dstOrganizationId = (dstOrganization) ? dstOrganization.id : dstOrganization;
+        setDropdownOpen(false);
```
klakhov: Why dont we use `stopEditField` instead?

cvat-ui/src/components/tasks-page/actions-menu.tsx
```diff
@@ -112,6 +125,47 @@ function TaskActionsComponent(props: Props): JSX.Element {
...
+            dispatch(openLinkedCloudStorageUpdatingModal(taskInstance));
+        } else {
+            dispatch(updateTaskAsync(taskInstance, {}, TaskUpdateTypes.UPDATE_ORGANIZATION))
+                .then(stopEditField);
+        }
+    }, [taskInstance]);
+
+    const onUpdateTaskOrganization = useCallback((dstOrganization: Organization | null) => {
+        const dstOrganizationId = (dstOrganization) ? dstOrganization.id : dstOrganization;
+        setDropdownOpen(false);
```
klakhov: It seems we can use `stopEditField`

cvat/apps/engine/serializers.py
```diff
@@ -2432,7 +2480,24 @@ def create(self, validated_data):
...
+            and organization_id != instance.organization_id
+        )
+
+        if (
+            workspace_transferring and (set(validated_data.keys()) - {
+                'source_storage', 'target_storage', 'organization_id'
+            })
+        ):
+            raise serializers.ValidationError(
+                "Task attributes cannot be updated during transferring to another workspace"
```
zhiltsov-max: Consider specifying the fields in question, the current error message can be confusing to the users. For example, the `Transfer a storage linked task to an organization and detach storages` example changes 2 storage attributes, but it's allowed.

cvat/apps/engine/serializers.py
```diff
@@ -2502,10 +2566,51 @@ def update(self, instance, validated_data):
 
             instance.project = project
 
+        if workspace_transferring:
+            if organization_id is not None:
+                try:
+                    organization_slug = list(
+                        Organization.objects.filter(pk=organization_id).values_list('slug', flat=True)
+                    )[0]
```
```suggestion
                    organization_slug = (
                        Organization.objects.filter(pk=organization_id).values_list('slug', flat=True)
                    )[0]
```

Calling `list` is not needed here, the returned object allows indexing. Removing `list` will also result in a request with a `LIMIT 1` clause.

cvat/apps/engine/serializers.py
```diff
@@ -2432,7 +2480,24 @@ def create(self, validated_data):
...
+
+        if (
+            workspace_transferring and (set(validated_data.keys()) - {
+                'source_storage', 'target_storage', 'organization_id'
+            })
+        ):
+            raise serializers.ValidationError(
+                "Task attributes cannot be updated during transferring to another workspace"
+            )
+
```
zhiltsov-max: Given that most of the code below is irrelevant if a transfer is requested, can we extract this logic into a new function or at least make a big if-else in this one?

cvat/apps/engine/serializers.py
```diff
@@ -2502,10 +2566,51 @@ def update(self, instance, validated_data):
...
+            if instance.assignee_id is not None:
+                instance.update_assignee(None, updated_date=assignee_updated_date)
+
+            assigned_jobs_qs = models.Job.objects.filter(
+                segment__task__id=instance.pk,
+                assignee__isnull=False
+            )
+            if assigned_jobs_qs.exists():
+                assigned_jobs_qs.update(assignee=None, assignee_updated_date=assignee_updated_date)
```
zhiltsov-max: I think you can simply send 1 update request without an extra `.exists()` check here.

cvat/apps/engine/rq.py
```diff
@@ -426,3 +436,67 @@ def define_dependent_job(
...
+                if is_rq_job_related(BaseRQMeta.for_job(job)):
+                    raise RunningBackgroundProcessesError
+
+    for queue in queues:
+        job_ids = set(
+            queue.finished_job_registry.get_job_ids() + queue.failed_job_registry.get_job_ids()
+        )
+
+        for batched_job_ids in take_by(job_ids, chunk_size=1000):
+            with queue.connection.pipeline() as pipe:
```
zhiltsov-max: It looks like we can get no rollback for already modified batches if a pipeline for a new batch fails.

tests/python/shared/assets/annotations.json
```diff
@@ -6057,176 +6057,6 @@
     },
     "43": {
       "shapes": [
-        {
```
zhiltsov-max: Probably, changes in this file are just the reordering of annotations. Could you revert it?

cvat/apps/engine/permissions.py
```diff
@@ -259,7 +261,27 @@ def create(cls, request: ExtendedRequest, view: ViewSet, obj: Project | None, ia
...
             assignee_id = request.data.get('assignee_id') or request.data.get('assignee')
 
-            for scope in cls.get_scopes(request, view, obj):
+            scopes = cls.get_scopes(request, view, obj)
+
+            if cls.Scopes.UPDATE_ORGANIZATION in scopes:
+                # consider this case as deleting a project in the org A and creating a new one in the org B
+                permissions.append(cls.create_base_perm(request, view, cls.Scopes.DELETE, iam_context, obj))
+
+                if dst_org_id := request.data['organization_id']:
```
zhiltsov-max, line 271: If the field is missing in the request, accessing it will lead to the 500 (server) error instead of 400.

cvat/apps/engine/permissions.py
```diff
@@ -259,7 +261,27 @@ def create(cls, request: ExtendedRequest, view: ViewSet, obj: Project | None, ia
         if view.basename == 'project':
             assignee_id = request.data.get('assignee_id') or request.data.get('assignee')
 
-            for scope in cls.get_scopes(request, view, obj):
+            scopes = cls.get_scopes(request, view, obj)
+
+            if cls.Scopes.UPDATE_ORGANIZATION in scopes:
+                # consider this case as deleting a project in the org A and creating a new one in the org B
+                permissions.append(cls.create_base_perm(request, view, cls.Scopes.DELETE, iam_context, obj))
```
zhiltsov-max: Maybe we can optimize several IAM requests by checking if the new org id is different from the existing one

cvat/apps/engine/permissions.py
```diff
@@ -423,6 +445,7 @@ class Scopes(StrEnum):
         UPDATE_ASSOCIATED_STORAGE = 'update:associated_storage'
         UPDATE_DESC = 'update:desc'
         UPDATE_METADATA = 'update:metadata'
+        # complex scope consisting of a combination of other scopes
```
zhiltsov-max: It's not clear to which scope this comment is related to and it doesn't really add useful information. There are several composite scopes in this class. The same is in projects. If the intention was to say that it's a "proxy scope" with no direct rules, maybe it's better to change the wording.

cvat/apps/engine/permissions.py
```diff
@@ -442,17 +465,35 @@ def create(cls, request: ExtendedRequest, view: ViewSet, obj: Task | None, iam_c
...
 
-            for scope in cls.get_scopes(request, view, obj):
+            scopes = cls.get_scopes(request, view, obj)
+
+            if cls.Scopes.UPDATE_ORGANIZATION in scopes:
+                # consider this case as deleting a task in the org A and creating a new one in the org B
+                if obj.project is not None:
+                    raise ValidationError(
+                        'Cannot change the organization for a task inside a project'
+                    )
```
zhiltsov-max: It's better to check this in the serializer and check only permissions on this level.

cvat/apps/engine/permissions.py
```diff
@@ -442,17 +465,35 @@ def create(cls, request: ExtendedRequest, view: ViewSet, obj: Task | None, iam_c
             assignee_id = request.data.get('assignee_id') or request.data.get('assignee')
             owner = request.data.get('owner_id') or request.data.get('owner')
 
-            for scope in cls.get_scopes(request, view, obj):
+            scopes = cls.get_scopes(request, view, obj)
+
+            if cls.Scopes.UPDATE_ORGANIZATION in scopes:
+                # consider this case as deleting a task in the org A and creating a new one in the org B
+                if obj.project is not None:
```
zhiltsov-max:
```suggestion
                if obj.project_id:
```
Accessing `project` can lead to fetching from the DB, if the project is not prefetched.

cvat/apps/engine/permissions.py
Marishka17: This code is under `if cls.Scopes.UPDATE_ORGANIZATION in scopes` statement; scope `UPDATE_ORGANIZATION` is only added when the `organization_id ` exists in the request data. 👀

cvat/apps/engine/permissions.py
```diff
@@ -564,9 +605,10 @@ def _get_scopes(cls, request: ExtendedRequest, view: ViewSet, obj: Task | None)
                 'labels': Scopes.UPDATE_DESC,
                 'bug_tracker': Scopes.UPDATE_DESC,
                 'subset': Scopes.UPDATE_DESC,
-                'organization': Scopes.UPDATE_ORGANIZATION,
+                'organization_id': Scopes.UPDATE_ORGANIZATION,
                 'source_storage': Scopes.UPDATE_ASSOCIATED_STORAGE,
                 'target_storage': Scopes.UPDATE_ASSOCIATED_STORAGE,
+                # 'cloud_storage_id': Scopes.UPDATE_ASSOCIATED_STORAGE,
```
zhiltsov-max: Can we remove the comment?

cvat/apps/engine/rq.py
```diff
@@ -153,9 +157,13 @@ def for_job(cls, job: RQJob):
     def for_meta(cls, meta: dict[str, Any]):
         return cls(meta=meta)
 
-    def save(self) -> None:
+    def save(self, *, pipeline: Pipeline | None = None) -> None:
         assert isinstance(self._job, RQJob), "To save meta, rq job must be set"
-        self._job.save_meta()
+        # TODO: send PR to the upstream repo to support pipelines
+        # self._job.save_meta()
```
zhiltsov-max: Can we remove commented code?

cvat/apps/engine/serializers.py
```diff
@@ -2502,10 +2566,51 @@ def update(self, instance, validated_data):
...
+            dst_organization_id=organization_id,
+        )
 
-        instance.save()
+        if workspace_transferring:
+            try:
+                update_org_related_data_in_rq_jobs(organization_id, organization_slug, task_id=instance.pk)
+            except RunningBackgroundProcessesError:
+                raise serializers.ValidationError(
+                    "There are some running background processes related with the task"
```
zhiltsov-max: It's better to change the wording to mention something known to the users, e.g. requests, exports, imports etc. "Some background processes" is quite a broad class, which can be confusing to users.

cvat/apps/engine/serializers.py
```diff
@@ -2502,10 +2566,51 @@ def update(self, instance, validated_data):
 
             instance.project = project
 
+        if workspace_transferring:
```
zhiltsov-max: Please add a comment about possible race conditions.

cvat/apps/engine/serializers.py
```diff
@@ -2502,10 +2566,51 @@ def update(self, instance, validated_data):
...
+            instance,
+            validated_data=validated_data,
+            workspace_transferring=workspace_transferring,
+            dst_organization_id=organization_id,
+        )
 
-        instance.save()
+        if workspace_transferring:
+            try:
+                update_org_related_data_in_rq_jobs(organization_id, organization_slug, task_id=instance.pk)
```
zhiltsov-max: I'm not sure about performance of this function, maybe it's really better to move transfers into a queue.

cvat/apps/engine/serializers.py
```diff
@@ -3499,20 +3689,61 @@ class Meta:
...
+
+                similar_ones: list[models.CloudStorage] = []
+                msg_no_similar_cs = "Could not find a similar cloud storage in the new workspace"
+
+                qs_with_similar_ones = models.CloudStorage.objects.filter(
+                    provider_type=original_cs.provider_type,
+                    resource=original_cs.resource,
+                    specific_attributes=original_cs.specific_attributes,
+                    organization_id=dst_organization_id,
+                ).annotate(Count("manifests"))
```
zhiltsov-max: `original_cs` can be undefined here

cvat/apps/engine/serializers.py
```diff
@@ -3499,20 +3689,61 @@ class Meta:
...
-        # storage_instance maybe None
-        storage_instance = getattr(instance, storage_type)
+        if not new_conf:
+            if not workspace_transferring:
+                continue
+
+            if (
+                storage_instance and storage_instance.cloud_storage_id
+            ):
+                # generally should not occur
```
zhiltsov-max: This can happen because of a thread race, but it's not an invalid situation.

cvat/apps/engine/serializers.py
```diff
@@ -3499,20 +3689,61 @@ class Meta:
...
+                except models.CloudStorage.DoesNotExist:
+                    new_location = models.Location.LOCAL
+
+                similar_ones: list[models.CloudStorage] = []
+                msg_no_similar_cs = "Could not find a similar cloud storage in the new workspace"
+
+                qs_with_similar_ones = models.CloudStorage.objects.filter(
+                    provider_type=original_cs.provider_type,
+                    resource=original_cs.resource,
+                    specific_attributes=original_cs.specific_attributes,
```
zhiltsov-max: If this field has a flexible format (e.g. an unordered json object), it's better to check field matches separately after loading candidates from the DB.

cvat/apps/engine/serializers.py
```diff
@@ -3499,20 +3689,61 @@ class Meta:
...
+                ).annotate(Count("manifests"))
+
+                for cs in qs_with_similar_ones:
+                    if cs.manifests__count != original_cs.manifests__count:
+                        continue
+
+                    # we do not need to compare manifest contents since the content
+                    # will be synchronized with the original CS file on the next content request
+                    if (
+                        list(cs.manifests.values_list("filename", flat=True).order_by("filename"))
```
zhiltsov-max: Consider doing a bulk prefetch before the loop, there are several  examples of bulk requests in `*ReadListSerializer` classes in this file.

cvat/apps/engine/serializers.py
zhiltsov-max, line 1: In the line 3758: 

> `f"The configuration of {storage_type} is not full"`

Consider making the message more clear to the users. Also, it looks like these checks could be moved into `StorageSerializer` `validate()`.

cvat/apps/engine/serializers.py
```suggestion
                ).annotate(Count("manifests")).filter(manifests__count=original_cs.manifests__count)
```

cvat/apps/engine/permissions.py
zhiltsov-max: Ok, found it.

cvat-ui/src/components/projects-page/actions-menu.tsx
Marishka17: I thought we only needed to call `stopEditField` when an update is fully done, since it not only hides the dropdown but also changes the state of `editField`. However, in some cases—such as when the opened context is an organization, or when a resource is linked with cloud storage—we need to hide the initial dropdown, show several modals, and only after confirming the transfer and/or choosing the transfer scheme should we stop the edit operation. Am I missing something?

cvat-ui/src/components/projects-page/actions-menu-items.tsx
Marishka17: > We usually try to reserve weight numbers in the *1–*9 range for plugins

It makes sense, but changing this weight has 2 disadvantages, IMHO:
1. It requires updating all next weights + in private plugins too, am I right?
1. Is there any reason to insert some action between edit actions? I thought that all edit-related actions should be grouped.

cvat/apps/engine/permissions.py
Marishka17: I think this comment is relevant to other updates as well, and we currently don’t check whether a field’s value in the request data differs from the current field value. Additionally, I believe that at this level we should validate permissions regardless of whether the value has changed - this logic should be handled by the serializers.

cvat/apps/engine/rq.py
Marishka17: It's the previous "normal" implementation of this function, not just commented code

cvat/apps/engine/serializers.py
Marishka17: Let's take a look at whether it would really be a problem.

cvat/apps/engine/rq.py
zhiltsov-max: It there a specific reason to store the previous normal implementation in the comment? It is available in the version history.

cvat-ui/src/components/projects-page/actions-menu.tsx
klakhov: I don’t think we really need to call `stopEditField` when an update is fully completed. We should call it only when we actually want to stop editing the field. In this case, once we’ve selected an organization, the input operation is finished, and we should move on to the next step with the modal.

Using `stopEditField` hides the dropdown but keeps the state, and we then have to remember to reset it manually. For example, if I click *Cancel* in the modal, `editField` isn’t reset. As a result, when reopening the menu, it just opens the organization selector instead of the original menu, and there’s no way to go back.

cvat-ui/src/components/projects-page/actions-menu-items.tsx
klakhov: Updating the subsequent weights isn’t a big deal, and as far as I remember, we don’t currently have any active plugins that extend this menu.

In a plugin, we could probably add one more edit action that would fit well within this group.

cvat-ui/src/components/selectors/organization-selector.tsx
```diff
@@ -0,0 +1,108 @@
...
+    const dispatch = useDispatch();
+
+    useEffect(() => {
+        if (organizations.length) {
+            setSearchResults((prev) => (page === defaultPage ? [...organizations] : [...prev, ...organizations]));
+        }
+    }, [organizations]);
+
+    useEffect(() => {
+        setSearchResults([]);
```
klakhov: Do we really need this reset here? It causes the search dropdown to flicker on every change of the search phrase.

org-search-1

cvat-ui/src/components/selectors/organization-selector.tsx
```diff
@@ -0,0 +1,108 @@
...
+            page,
+            ...((searchPhrase) ? { search: searchPhrase } : {}),
+        };
+        dispatch(getOrganizationsAsync(filterQuery, true));
+    }, [page]);
+
+    const handlePopupScroll = (e: React.UIEvent<HTMLDivElement>) => {
+        const target = e.currentTarget;
+        if (
+            ((target.scrollTop + target.offsetHeight) >= (target.scrollHeight - 10)) &&
```
klakhov: Instead of relying on scroll events, I’d suggest using an Intersection Observer for the last element in the list. We can leverage `useInView` from `react-intersection-observer`.

One approach is to place a `ref` on the last label in the options, or alternatively, add a static invisible option at the end of the list to act as the observer target.

cvat-ui/src/components/update-linked-cs-modal/select-cs-updating-scheme-modal.tsx
```diff
@@ -0,0 +1,133 @@
...
+import { updateProjectAsync, ProjectUpdateTypes } from 'actions/projects-actions';
+import { updateTaskAsync, TaskUpdateTypes } from 'actions/tasks-actions';
+import { useDropdownEditField } from 'utils/hooks';
+import CVATTooltip from 'components/common/cvat-tooltip';
+import Space from 'antd/lib/space';
+import { QuestionCircleOutlined } from '@ant-design/icons';
+import {
+    closeLinkedCloudStorageUpdatingModal as closeTaskLinkedCloudStorageUpdatingModal,
+} from '../../actions/tasks-actions';
+import { projectActions } from '../../actions/projects-actions';
```
klakhov: Excessive `../../`

cvat-ui/src/components/update-linked-cs-modal/styles.scss
```diff
@@ -0,0 +1,6 @@
+// Copyright (C) CVAT.ai Corporation
```
klakhov: File isnt used

cvat-ui/src/reducers/organizations-reducer.ts
```diff
@@ -137,6 +145,52 @@ export default function (
         case AuthActionTypes.LOGOUT_SUCCESS: {
             return { ...defaultState };
         }
+        case OrganizationActionsTypes.GET_ORGANIZATIONS:
+            return {
+                ...state,
+                // initialized: false,
```
klakhov:
```suggestion
                 initialized: false,
```

cvat-ui/src/reducers/tasks-reducer.ts
```diff
@@ -36,6 +36,9 @@ const defaultState: TasksState = {
         deletes: {},
         updates: {},
     },
+    updateWorkspace: {
```
klakhov: Do we really need one more reducer field? We already have `activities: updates`

cvat-ui/src/components/update-linked-cs-modal/select-cs-updating-scheme-modal.tsx
```diff
@@ -0,0 +1,133 @@
...
+    const saveInstance = useCallback(() => {
+        if (instance instanceof core.classes.Project) {
+            dispatch(updateProjectAsync(instance, ProjectUpdateTypes.UPDATE_ORGANIZATION));
+        } else if (instance instanceof core.classes.Task) {
+            dispatch(updateTaskAsync(instance, {}, TaskUpdateTypes.UPDATE_ORGANIZATION));
+        }
+    }, [instance]);
+
+    const closeModal = useCallback(() => {
+        if (instance instanceof core.classes.Project) {
```
klakhov: I’d suggest using `activities.updates` to store the instance and a simple boolean field to represent the open/closed state of this modal. Then, in this case, we can simply dispatch an action to set the boolean to `false`.

When we call `saveInstance` with `update*Async`, it will reset the corresponding `activities.updates` fields automatically.

cvat/apps/engine/serializers.py
```diff
@@ -224,6 +228,80 @@ def to_representation(self, instance):
...
 
 
+class OrgTransferableMixin():
+    def _validate_org_transferring(self, attrs: dict[str, Any]):
+        """Method should be called explicitly"""
+
+        if set(attrs.keys()) - {'source_storage', 'target_storage', 'organization_id'}:
+            raise serializers.ValidationError(
+                "Resource attributes cannot be updated during transferring to another workspace"
+            )
```
```suggestion
        if disallowed_fields := set(attrs.keys()) - {
            'source_storage', 'target_storage', 'organization_id'
        }:
            raise serializers.ValidationError(
                f"Fields {disallowed_fields} cannot be updated when "
                f"transferring to another workspace"
            )
```

cvat/apps/engine/serializers.py
```diff
@@ -224,6 +228,80 @@ def to_representation(self, instance):
...
+                )[0]
+            except IndexError:
+                raise serializers.ValidationError("Invalid organization id")
+
+        cur_user_id = request.user.id
+        if instance.owner_id != cur_user_id:
+            instance.owner_id = cur_user_id
+            update_fields.append("owner_id")
+
+        assignee_updated_date = timezone.now()
```
```suggestion
        update_date = timezone.now()
```

Consider renaming this variable, it's used in several places and is not really related to assignee update.

cvat/apps/engine/serializers.py
```diff
@@ -2430,25 +2586,44 @@ def create(self, validated_data):
...
+                update_fields.append(field_name)
 
+    def update_labels(
+        self,
+        instance: models.Task,
+        *,
+        validated_data: dict[str, Any],
+        update_fields: list[str],
+    ):
+        # TODO: update_fields
```
zhiltsov-max: Could you explain what is supposed to be done for "update_fields"? I can see that if there are only labels in the input fields, the task `updated_date` will not be updated. It looks like a bug. I think you could just add `"updated_date"` to `update_fields` here if you want to stick to the current implementation.

cvat/apps/engine/serializers.py
```diff
@@ -2501,11 +2676,29 @@ def update(self, instance, validated_data):
...
-        instance.save()
+        if (
+            "organization_id" in validated_data
+            and validated_data.get("organization_id") != instance.organization_id
+        ):
+            self.update_organization(instance, validated_data=validated_data, update_fields=update_fields)
+        else:
+            self.update_base_properties(instance, validated_data=validated_data, update_fields=update_fields)
+            self.update_labels(instance, validated_data=validated_data, update_fields=update_fields)
+            # TODO: should be separate action
```
zhiltsov-max: Could you explain what is supposed to be done here?

cvat/apps/engine/serializers.py
```diff
@@ -2647,27 +2881,45 @@ def create(self, validated_data):
...
-            instance.assignee_id = validated_data.pop('assignee_id')
-            instance.assignee_updated_date = timezone.now()
+            self.update_organization(instance, validated_data=validated_data, update_fields=update_fields)
+        else:
+            self.update_base_properties(instance, validated_data=validated_data, update_fields=update_fields)
 
-        labels = validated_data.get('label_set', [])
-        LabelSerializer.update_labels(labels, parent_instance=instance)
+            labels = validated_data.get('label_set', [])
+            LabelSerializer.update_labels(labels, parent_instance=instance)
```
zhiltsov-max: Need to update project's `updated_date`.

cvat/apps/engine/serializers.py
```diff
@@ -3499,61 +3771,74 @@ class Meta:
...
-                and getattr(storage_instance, "location", None) == models.Location.CLOUD_STORAGE
-                and "cloud_storage_id" in new_conf
-            ):
-                raise serializers.ValidationError(
-                    "It is not allowed to reset a cloud storage id without explicitly resetting a location"
-                )
+                new_conf = {
+                    "location": models.Location.CLOUD_STORAGE,
+                    "cloud_storage_id": similar_ones[0].pk
+                }
```
zhiltsov-max, line 3852: Consider moving everything after `get()` into the `else` block of try-except-else, the try section seems to be too broad for the exception being catched.

cvat/apps/engine/serializers.py
```diff
@@ -3499,61 +3771,74 @@ class Meta:
...
+        # todo: check that CS exists (filter also by dst_organization_id or instance.organization_id)
 
-        storage_instance.location = new_location or storage_instance.location
-        storage_instance.cloud_storage_id = new_cloud_storage_id
-        storage_instance.save()
+        if storage_instance:
+            storage_serializer.save()
+        else:
+            storage_instance = storage_serializer.save()
+            setattr(instance, storage_type, storage_instance)
```
```suggestion
        storage_instance = storage_serializer.save()
        setattr(instance, storage_type, storage_instance)
```
- Why this condition is needed? 
- The updated existing object might be outdated in the instance, but will be present if there was no object

cvat/apps/engine/serializers.py
Marishka17: thanks, I fogot about this option

cvat/apps/engine/serializers.py
Marishka17: Updating other task properties should be forbidden when moving it to a project.

cvat/apps/engine/serializers.py
Marishka17: That comment was for me not to forget to update the updated_date field when modifying labels. I know that it was a bug.

cvat/apps/engine/serializers.py
Marishka17: you're right

cvat-core/src/core-types.ts
```diff
@@ -63,3 +63,4 @@ export interface UpdateStatusData {
 }
 
 export type PaginatedResource<T> = T[] & { count: number };
+export type PaginatedResourceWithNextUrl<T> = T[] & { count: number, next: string };
```
klakhov:
```suggestion
export type PaginatedResource<T> = T[] & { 
    count: number; 
    next?: string; 
};
```
Lets just extend current type, It looks a bit cleaner

cvat-core/src/server-proxy.ts
```diff
@@ -1961,13 +1961,14 @@ async function getOrganizations(filter) {
         response = await Axios.get(`${backendAPI}/organizations`, {
             params: {
                 ...filter,
+                page_size: filter.page_size || 10,
+                sort: filter.sort || 'slug',
```
klakhov, line 1969: Why do we need default sorting?

cvat-ui/src/actions/organization-actions.ts
```diff
@@ -91,6 +98,18 @@ const organizationActions = {
     updateOrganizationMemberFailed: (username: string, role: string, error: any) => createAction(
         OrganizationActionsTypes.UPDATE_ORGANIZATION_MEMBER_FAILED, { username, role, error },
     ),
+    getOrganizations: () => createAction(OrganizationActionsTypes.GET_ORGANIZATIONS, {}),
+    getOrganizationsSuccess: (array: Organization[], count: number, nextPageUrl?: string) => (
```
klakhov:
```suggestion
    getOrganizationsSuccess: (organizations: Organization[], count: number, nextPageUrl?: string) => (
```

cvat/apps/engine/serializers.py
```diff
@@ -2345,14 +2460,47 @@ def to_representation(self, instance):
         representation['consensus_enabled'] = self.get_consensus_enabled(instance)
         return representation
 
-
-class TaskWriteSerializer(WriteOnceMixin, serializers.ModelSerializer):
+@extend_schema_serializer(
+    examples=[
```
zhiltsov-max: Could you restore the original default example for regular field updates? Swagger only shows these new examples now.

cvat-ui/src/actions/organization-actions.ts
```diff
@@ -91,6 +98,18 @@ const organizationActions = {
...
+        createAction(OrganizationActionsTypes.GET_ORGANIZATIONS_SUCCESS, { array, count, nextPageUrl })
+    ),
+    getOrganizationsFailed: (error: any) => createAction(OrganizationActionsTypes.GET_ORGANIZATIONS_FAILED, { error }),
+    updateOrganizationsGettingQuery: (query: Partial<OrganizationsQuery>) => (
```
klakhov: Do we really need this action? Usually we would use getOrganizations to setup the query

cvat-ui/src/actions/organization-actions.ts
```diff
@@ -251,4 +270,19 @@ export function updateOrganizationMemberAsync(
     };
 }
 
+export function getOrganizationsAsync(query: Partial<OrganizationsQuery> = {}): ThunkAction {
+    return async (dispatch): Promise<void> => {
+        dispatch(organizationActions.getOrganizations());
+        dispatch(organizationActions.updateOrganizationsGettingQuery(query));
+
+        try {
+            const result = await core.organizations.get(query);
```
klakhov: Probably we need to use `filterNull` before passing it into cvat-core

cvat/apps/engine/serializers.py
Marishka17: By restoring, do you mean adding a corresponding example or deleting these examples?

cvat-ui/src/actions/organization-actions.ts
zhiltsov-max, line 1: I think it will be better to move this action from the "safe local actions" section to the "be cautious" section below:
move_to_org_placement

right next to "move to project" for tasks. This is more in line with the action nature than changing assignee.

The confirmation dialog should also be shown when moving from the local workspace as well, the action seems to be "too simple" now. It's also easy to missclick it when trying to change an assignee, because both actions use the same dialog and have no title during the selection.

For some reason, moving to project and moving to org use different destination selection layouts, probably they should be the same. Assigning is a little bit different in this regard, it's supposed to be a quick action.

cvat/apps/engine/serializers.py
Marishka17: So, I can add a default example as it was, but it will include all fields, and I do not really think it's right to show different operations in one example.

cvat-ui/src/reducers/index.ts
```diff
@@ -67,6 +67,8 @@ export interface ProjectsState {
         };
         updates: {
             [projectId: number]: boolean; // updated (updating if in dictionary)
+            current: Project | null;
+            visibleLinkedCloudStorageUpdatingModal: boolean;
```
klakhov: Sorry, I was mistaken about `activities:updates`—I thought we were storing full instances there, not just IDs. Adding an instance here makes things a bit confusing.

Let’s revert to the original approach, where we have a separate field:

```ts
updateWorkspace: {
    instance: null,
}
```

The only thing I’d change is to avoid duplicating this in both the task and project reducers/actions. Since this field is specifically related to the `LinkedCloudStorageUpdatingModal`, we can place it in the cloud storage reducer with two fields:

```ts
updateWorkspace: {
    instance: Task | Project | null,
    visibleLinkedCloudStorageUpdatingModal: boolean,
}
```

cvat/apps/engine/serializers.py
zhiltsov-max: It was convenient to have an example with all the possible fields in swagger. I'm just asking to add 1 more example for updating regular fields.

cvat-ui/src/components/selectors/organization-selector.tsx
```diff
@@ -0,0 +1,118 @@
...
+                    return {
+                        value: organization.slug,
+                        label: (
+                            <div ref={(isLastItem) ? ref : undefined}>
+                                {organization.slug}
+                                {organization.name === organization.slug ? '' : ` (${organization.name})`}
+                            </div>
+                        ),
+                    };
+                }),
```
klakhov:
```suggestion
                ...searchResults.map((organization, index, array) => ({
                    value: organization.slug,
                    label: (
                        <div ref={index === array.length - 1 ? ref : undefined}>
                            {organization.slug}
                            {organization.name === organization.slug ? '' : ` (${organization.name})`}
                        </div>
                    ),
                })),
```

cvat-ui/src/actions/organization-actions.ts
Marishka17: > For some reason, moving to project and moving to org use different layouts, 

It was discussed with the team to use this selector for updating an organization. Probably the `Move to project` action should also be re-designed, but it's out of the scope of this PR.

cvat-core/src/api-implementation.ts
zhiltsov-max, line 1: UI allows moving to the org the object currently belongs to. Maybe it will be better to exclude the current org from the list

cvat-ui/src/actions/organization-actions.ts
zhiltsov-max: @bsekachev , @klakhov , was it discussed?

cvat-core/src/server-proxy.ts
Marishka17: IMHO, it's better to use the `slug` sorting when searching for an organization in the list. I initially considered using default sorting for the "initial" list of organizations, but decided not to complicate the approach (and it might be unexpected for users anyway). Since we're calling this function only for searching organizations, I set slug as the default sorting here.

cvat-core/src/api-implementation.ts
Marishka17: okay, I will fix, thx

cvat-ui/src/actions/organization-actions.ts
Marishka17: @zhiltsov-max, great that you don't believe me

zhiltsov-max: I just think that maybe something was overlooked during decision making

cvat-ui/src/reducers/index.ts
Marishka17: Done, but I still find it questionable why an additional boolean field is needed when the instance should be enough.

cvat-ui/src/actions/organization-actions.ts
Marishka17: Now it's in the danger zone, and confirmation is required

cvat-core/src/api-implementation.ts
Marishka17: Now a current org is shown but disabled

cvat-ui/src/actions/organization-actions.ts
klakhov: The usage of this type of selector was explicitly discussed on one of the meetings, and we even integrated such a selector before implementing bulk actions to support this feature.

Its position in the menu is debatable, but I agree that placing it in the “danger zone” with a confirmation step is a better approach. That’s how it’s currently implemented in the PR, and it looks good to me.

cvat-ui/src/actions/organization-actions.ts
klakhov: Seems this issue is still not resolved

cvat-ui/src/actions/organization-actions.ts
```diff
@@ -251,4 +271,22 @@ export function updateOrganizationMemberAsync(
     };
 }
 
+export function getOrganizationsAsync(query: Partial<OrganizationsQuery> = {}): ThunkAction {
+    return async (dispatch): Promise<void> => {
+        dispatch(organizationActions.getOrganizations());
+        dispatch(organizationActions.updateOrganizationsGettingQuery(query));
+
+        // Clear query object from null fields
```
klakhov:
```suggestion
```

cvat-ui/src/components/tasks-page/actions-menu-items.tsx
```diff
@@ -112,20 +114,26 @@ export default function TaskActionsItems(menuItemsData: MenuItemsData, taskMenuP
...
             key: 'move_task_to_project',
             onClick: onMoveTaskToProject,
             label: 'Move to project',
         }, 90]);
+
+        menuItems.push([{
+            key: 'edit_organization',
+            onClick: () => startEditField('organization'),
+            label: <CVATMenuEditLabel>Organization</CVATMenuEditLabel>,
+        }, 100]);
```
klakhov, line 146: Lete move this after divider and before the `delete`

cvat-ui/src/utils/modals.ts
```diff
@@ -0,0 +1,32 @@
+import Modal from 'antd/lib/modal';
```
klakhov, line 5: Missing header here

cvat-ui/src/components/select-organization-modal/select-organization-modal.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+import React from 'react';
+import Modal from 'antd/lib/modal';
+
+import { useSelector, useDispatch } from 'react-redux';
+import { CombinedState } from 'reducers';
+
+import { Organization } from 'cvat-core-wrapper';
+import { organizationActions } from 'actions/organization-actions';
+import OrganizationSelector from '../selectors/organization-selector';
```
klakhov:
```suggestion
import OrganizationSelector from 'components/selectors/organization-selector';
```

cvat-ui/src/components/selectors/organization-selector.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+import './styles.scss';
+import _ from 'lodash';
+import React, { useEffect, useState } from 'react';
+import AutoComplete from 'antd/lib/auto-complete';
+import { useInView } from 'react-intersection-observer';
+
+import { Organization } from 'cvat-core-wrapper';
+import { getOrganizationsAsync } from 'actions/organization-actions';
+import { useSelector, useDispatch } from 'react-redux';
```
klakhov: Lets move 3dparty imports before our components/actions

cvat-ui/src/actions/organization-actions.ts
zhiltsov-max: Ok

zhiltsov-max: I would add a title to the selection box popup. I think what we have in the workspace switching dialog looks much more clear than what we have here, given that the actions menu disappears after selecting an option there.

select_workspace_dialog

select_something_dialog

cvat-ui/src/components/update-linked-cs-modal/select-cs-updating-scheme-modal.tsx
```diff
@@ -0,0 +1,126 @@
...
+            ]}
+        >
+            <p>
+                This
+                {' '}
+                {instanceType}
+                {' '}
+                is linked to cloud storage.
+                Please choose how you would like the transfer to be done.
+            </p>
```
klakhov:
```suggestion
    <p>
        {`This ${instanceType} is linked to cloud storage. `}
        Please choose how you would like the transfer to be done.
    </p>
```

cvat-ui/src/reducers/index.ts
klakhov: You’re right, it doesn’t seem to make sense. We should remove it.

cvat/apps/engine/serializers.py
zhiltsov-max, line 1: I tried to move a sandbox task with CS images as data source to an org and got the task in the org, but no new CS in the org (it had none). The CS remained as it was in the DB and the task.data also wasn't changed. Shouldn't the task be detached or the CS be copied in this case?

cvat-ui/src/actions/organization-actions.ts
klakhov: What do you think about changing the selector’s placeholder from **“Type to search”** to **“Select an organization”**?
The initial idea was to keep the selector minimalistic.

Adding a title, like **“New organization:”** is still an option though.

cvat/apps/engine/serializers.py
zhiltsov-max, line 1: Project webhooks have an explicit `organization` field. Apparently, it's not changed on project organization changes. I think, ideally we could make this field inferred from the target, the same way as in other models (e.g. `AnnotationGuide`), but it would be a big and unrelated change for this PR.

cvat-ui/src/actions/organization-actions.ts
zhiltsov-max: I think it would work as well.

klakhov: I would vote for updating placeholder

cvat/apps/engine/serializers.py
Marishka17: Thanks, I didn't think about it. So, I've fixed a problem with project webhooks, but there is still an issue with organizations' webhooks. Do you mind if task/project transfers are considered as deletions in one org and creations in another org for webhooks as well? (we'll inform about 2 events: project X is deleted in the org A and is created in the org B)

cvat-ui/src/actions/organization-actions.ts
Marishka17: We need this action because there are two components that rely on fetched organizations: the "Organization" menu item from the main header, and the organization selectors in the tasks/projects menus. So, if we don't keep data of the filters used to fetch a page of organizations, the list of organizations a user can switch between will become outdated

cvat/apps/engine/serializers.py
Marishka17: Updating CS linked with task data should be implemented in Dmitry's PR.

zhiltsov-max: I think the solution with 2 events is fine as the first implementation.

zhiltsov-max: In this case this PR should probably be marked dependent on that PR. But I'm not sure the changes there would really address the question here, because if we move a CS task from one org to another, the old cloud storage and chunks become available to the new (basically, unauthorized) users now, since there's no new CS in the new org. The task is not detached and there's no permission check on task data access after the task is moved, the CS is just available.

cvat-ui/src/actions/organization-actions.ts
klakhov: I understand that we need the `query` to make the UI work properly—I'm not suggesting we remove it.
What I’m saying is that we can set it directly in the state as part of the `getOrganizations` action, rather than using a separate action.

This is similar to how it's handled in [getJobsAsync](https://github.com/cvat-ai/cvat/blob/develop/cvat-ui/src/actions/jobs-actions.ts#L88)

To me, having two separate actions makes sense only when `updateQuery` is called from different places (like for projects).

Marishka17: Ok

Marishka17:
@zhiltsov-max, @klakhov, please review the recent changes ASAP

tests/python/rest_api/test_tasks.py
zhiltsov-max, line 1: Could you please add at least 1 test for changing a task and a project org?

tests/python/shared/assets/tasks.json
Marishka17, line 1: @bsekachev, probably need to update assets in the private repo as well

tests/python/rest_api/test_tasks.py
Marishka17: I've added simple tests. Unfortunately, I have no time to cover more cases.

PR #9535, opened 2025-06-16
Lightweight backups for cloud storage based tasks

Eldies:
### Checklist

### License


cvat/apps/engine/background.py
```diff
@@ -176,6 +176,21 @@ def get_result_endpoint_url(self) -> str:
...
     SUPPORTED_TARGETS = {RequestTarget.PROJECT, RequestTarget.TASK}
 
+    @dataclass
+    class ExportArgs(AbstractExporter.ExportArgs):
+        make_lightweight_backup: bool
+
+    def init_request_args(self) -> None:
+        super().init_request_args()
+        make_lightweight_backup = to_bool(
+            self.request.query_params.get("make_lightweight_backup", True)
```
Marishka17: Was the query parameter name agreed upon with the team? I think just `lightweight` will be better. (It is a query param inside the endpoint that handles backup process initialization, there is no need to add the`backup` suffix, IMHO)

cvat/apps/engine/background.py
```diff
@@ -214,6 +229,7 @@ def _init_callback_with_params(self):
             Exporter,
             logger,
             self.job_result_ttl,
+            self.export_args.make_lightweight_backup,
```
Marishka17: This change affects already created rq jobs. We need to remove them or add a redis migration. Additionally, I do not really like this approach because this option can only be used when task data is linked with a cloud storage; for other storage types, this option is useless. Probably we can use the rq job meta or introduce another exporter class that inherits `TaskExporter`.
@SpecLad, what do you think?

cvat/apps/engine/backup.py
```diff
@@ -556,7 +561,7 @@ def serialize_data():
                 ]
                 data['validation_layout'] = validation_params
 
-            if self._db_data.storage == StorageChoice.CLOUD_STORAGE:
+            if not self._make_lightweight_backup and self._db_data.storage == StorageChoice.CLOUD_STORAGE:
```
Marishka17:
```suggestion
            if self._db_data.storage == StorageChoice.CLOUD_STORAGE and not self._lightweight
```

cvat/apps/engine/backup.py
```diff
@@ -887,13 +892,19 @@ def _import_task(self):
...
+
+            manifest = ImageManifestManager(os.path.join(self._db_task.data.get_upload_dirname(), self.MEDIA_MANIFEST_FILENAME))
+            data['server_files'] = list(manifest.data)
+
         create_task(self._db_task.pk, data.copy(), is_backup_restore=True)
         self._db_task.refresh_from_db()
         db_data.refresh_from_db()
 
         db_data.deleted_frames = data_serializer.initial_data.get('deleted_frames', [])
-        db_data.storage = StorageChoice.LOCAL
```
Marishka17, line 895: Doesn't this change affect task backups created with the data from the share?

cvat/apps/dataset_manager/util.py
```diff
@@ -272,14 +272,15 @@ def make_backup_file_path(
...
+        full_fledged_backup: bool,
     ) -> str:
         instance_type = InstanceType(instance_type.lower())
         filename = cls.FILE_NAME_TEMPLATE_WITH_INSTANCE.format(
             instance_type=instance_type,
             instance_id=instance_id,
             file_type=ExportFileType.BACKUP,
             instance_timestamp=instance_timestamp,
-            optional_suffix="",
+            optional_suffix="_ff" if full_fledged_backup else "",
```
Marishka17: Consider adding a suffix only when it's a lightweight backup.

```suggestion
            optional_suffix=cls.SPLITTER + "lightweight" if lightweight else "",
```

cvat/schema.yml
```diff
@@ -9466,6 +9476,10 @@ components:
           allOf:
           - $ref: '#/components/schemas/StorageRequest'
           nullable: true
+        data_storage:
```
Marishka17: Did you discuss these API changes with the team? Why was the decision made to patch the Data-specific fields via the PATCH /api/tasks/id/ API endpoint and without DataSerializer? I think we need to introduce a common approach that can be extended in the future when we support other Data field changes. IMHO, it will be better to have some DataWriteSerializer with only one "cloud_storage_id" field that can be changed (for now). Since we do not have the data_storage related model, + we are not going to allow changing the storage type.

cvat/apps/engine/backup.py
Eldies: fixed

cvat/apps/dataset_manager/util.py
Eldies: done

cvat/apps/engine/backup.py
Eldies: done

cvat/apps/engine/background.py
Eldies: No, it was not agreed. renamed to `lightweight`

cvat/schema.yml
SpecLad: There is already precedent in the API for treating fields of the `Data` object as if they were logically on the task; e.g., see `data_chunk_size`, `data_compressed_chunk_type`, `data_original_chunk_type`. So it doesn't seem like a big stretch to add `data_storage` as well.

cvat/schema.yml
```diff
@@ -9466,6 +9476,10 @@ components:
...
+          allOf:
+          - $ref: '#/components/schemas/TaskDataStorageRequest'
```
SpecLad: Why not use the same type as `source_storage`/`target_storage`?

cvat/schema.yml
```diff
@@ -10791,6 +10824,10 @@ components:
           description: |
             The number of consensus replica jobs for each annotation job.
             Configured at task creation
+        data_storage:
```
SpecLad: Does this field need to be here? AFAIK, `TaskWriteRequest` is only used to create blank tasks, and the data storage for those is only set via the `POST /api/tasks/<id>/data` request.

cvat/apps/engine/background.py
SpecLad: It seems to me that this parameter has a similar role to the `format`/`save_images` parameters of dataset exports, so it should be handled in a similar way.

cvat/schema.yml
Marishka17: It's only about response schema (but I also think that these fields should be grouped). I see the difference between target_storage/source_storage and data_storage - data_storage is a part of a whole interface that describes a task data. It is not a separate entity. + Was there any discussion about adding an ability to change the data storage type(local/share/cloud)? If no, we should not mention in the API that it is possible to set a different location (but this will happen if we use the Storage schema)

cvat/apps/engine/background.py
Marishka17: It would only make sense for me if:
1. this option is handled for all storage types
2. there is an option in the UI to attach some data to the created empty tasks

cvat/schema.yml
SpecLad: > It's only about response schema (but I also think that these fields should be grouped). I see the difference between target_storage/source_storage and data_storage - data_storage is a part of a whole interface that describes a task data. It is not a separate entity. 

I don't quite understand what you're proposing. Could you show an example of what you would prefer the `GET /api/tasks/<id>` response to look like?

> Was there any discussion about adding an ability to change the data storage type(local/share/cloud)? If no, we should not mention in the API that it is possible to set a different location (but this will happen if we use the Storage schema)

I certainly don't think that changing the data storage type is in scope for this PR. We should make it so that if the user tries to change it, the request is rejected. That said, it's not inconceivable to me that in the future we'd want to allow at least some changes, e.g. between cloud and share.

cvat/apps/engine/background.py
SpecLad: > this option is handled for all storage types

Well... couldn't it be? I don't think there's anything actually preventing us from creating lightweight backups regardless of the current data storage type. We don't have to actually _implement_ this right now, but we might as well structure the code so that it could be implemented in the future.

> there is an option in the UI to attach some data to the created empty tasks

Sorry, I don't understand. This part of the code creates the backup, whereas this sounds like something you'd do when restoring the backup. How is it relevant?

cvat/schema.yml
```diff
@@ -3700,6 +3700,11 @@ paths:
           type: integer
         description: A unique integer value identifying this project.
         required: true
+      - in: query
+        name: lightweight
+        schema:
+          type: boolean
+        description: Make lightweight backup for cloud based tasks
```
SpecLad: 1. It would be useful to explain what "lightweight" means.
2. I only skimmed the PR so I could've missed it, but I didn't see any code that would reject this parameter when the task is _not_ cloud based. Please add it if it's not there.

Eldies: 1. Tried to write a better description.
2. I don't quite understand why it should be explicitly rejected instead of just doing nothing

cvat/schema.yml
Eldies: got rid of this in favor of using GET/PATCH `/api/tasks/<id>/data/meta`

cvat/schema.yml
Eldies: they have the `id` field, which is not needed here. Removed it, using other way to change data storage

cvat/schema.yml
Eldies: I totally missed that we already have a way to read and update data of the task. Removed all these changes, added fields to `DataMetaReadSerializer` and `DataMetaWriteSerializer` instead

cvat/apps/engine/background.py
Eldies: make `lightweight` parameter into a kwarg for `create_backup`. That way, existing rq backup jobs still can execute

cvat/apps/engine/background.py
```diff
@@ -177,6 +177,19 @@ def get_result_endpoint_url(self) -> str:
 class BackupExporter(AbstractExporter):
     SUPPORTED_TARGETS = {RequestTarget.PROJECT, RequestTarget.TASK}
 
+    @dataclass
+    class ExportArgs(AbstractExporter.ExportArgs):
+        lightweight: bool
+
+    def init_request_args(self) -> None:
+        super().init_request_args()
+        lightweight = to_bool(self.request.query_params.get("lightweight", True))
```
Marishka17: True by default may be unexpected behavior for users, as it changes the output result despite the fact that the usage of API parameters does not change on the user's side (in some scripts).

cvat/apps/engine/background.py
```diff
@@ -216,6 +229,9 @@ def _init_callback_with_params(self):
             logger,
             self.job_result_ttl,
         )
+        self.callback_kwargs = dict(
```
Marishka17: Consider using dictionary literal

cvat/apps/engine/backup.py
```diff
@@ -887,12 +892,20 @@ def _import_task(self):
         if validation_params:
             data['validation_params'] = validation_params
 
+        if data["storage"] == StorageChoice.CLOUD_STORAGE:
+            if data['client_files'] != [self.MEDIA_MANIFEST_FILENAME]:
+                raise ValidationError(f"Expected {self.MEDIA_MANIFEST_FILENAME} in backup files")
+
+            manifest = ImageManifestManager(os.path.join(self._db_task.data.get_upload_dirname(), self.MEDIA_MANIFEST_FILENAME))
+            data['server_files'] = list(manifest.data)
+
```
Marishka17, line 901: There are 2 file types (client_files/server_files) that are passed to the task creation logic. I don't remember that we're using somewhere more than one data type when creating tasks and I rather think that this should not be allowed.

cvat/apps/engine/backup.py
```diff
@@ -1100,6 +1118,8 @@ def create_backup(
     Exporter: Type[ProjectExporter | TaskExporter],
     logger: Logger,
     cache_ttl: timedelta,
+    *,
+    lightweight: bool = True,
```
Marishka17: I also don't think the lightweight should be True by default because it changes the logic for RQ jobs created before.

One more disadvantage of setting the default value in this func - the default value for this argument is defined 2 times now: here and when reading the request.query_params. IMHO, it's better to have only one place where the default value is defined (when handling query params). But I'm okay if it's okay for others.

cvat/apps/engine/backup.py
```diff
@@ -971,13 +984,14 @@ def _prepare_project_meta(self, project):
 class ProjectExporter(_ExporterBase, _ProjectBackupBase):
     ModelClass: ClassVar[models.Project] = models.Project
 
-    def __init__(self, pk, version=Version.V1):
+    def __init__(self, pk, version=Version.V1, *, lightweight: bool = True):
```
Marishka17: I suggest not defining the default value for the lightweight argument here (the same comment for the TaskExported class).
```suggestion
    def __init__(self, pk, *, lightweight: bool, version: Version = Version.V1):
```

cvat/apps/engine/cache.py
```diff
@@ -583,7 +587,8 @@ def _read_raw_images(
             reader = ImageReaderWithManifest(manifest_path)
             with ExitStack() as es:
                 db_cloud_storage = db_data.cloud_storage
-                assert db_cloud_storage, "Cloud storage instance was deleted"
+                if not db_cloud_storage:
+                    raise CloudStorageMissingError("Task is not connected to cloud storage")
```
Marishka17:
```suggestion
                    raise CloudStorageMissingError("Task is no longer connected to cloud storage")
```

cvat/apps/engine/cache.py
```diff
@@ -74,6 +74,10 @@ class CacheTooLargeDataError(Exception):
     pass
 
 
+class CloudStorageMissingError(Exception):
```
Marishka17: Probably it's better to define this exception in models.py or exceptions.py, but not here.

cvat/apps/engine/mixins.py
```diff
@@ -366,6 +366,9 @@ class BackupMixin:
                 enum=Location.list()),
             OpenApiParameter('cloud_storage_id', description='Storage id',
                 location=OpenApiParameter.QUERY, type=OpenApiTypes.INT, required=False),
+            OpenApiParameter('lightweight',
+                description='Make lightweight backup (backup which does not include media) for tasks whose media is located in cloud storage',
```
Marishka17:
```suggestion
                description='Makes a lightweight backup (without media files) for tasks whose media is located in cloud storage',
```

cvat-core/src/server-proxy.ts
```diff
@@ -995,12 +995,14 @@ async function backupTask(
...
     useDefaultSettings: boolean,
     fileName?: string,
+    lightweight: boolean,
 ): Promise<string | void> {
     const { backendAPI } = config;
     const params: Params = {
         ...enableOrganization(),
         ...configureStorage(targetStorage, useDefaultSettings),
         ...(fileName ? { filename: fileName } : {}),
+        ...(!lightweight ? { lightweight } : {}),
```
Marishka17: It looks strange that you send the lightweight query param only when it's False.

cvat-core/src/server-response-types.ts
```diff
@@ -483,6 +483,8 @@ export interface SerializedFramesMetaData {
     size: number;
     start_frame: number;
     stop_frame: number;
+    storage: StorageLocation;
```
Marishka17: In fact it can be also "share" type

cvat-core/src/session.ts
```diff
@@ -1214,13 +1241,19 @@ export class Task extends Session {
         return result;
     }
 
-    async backup(targetStorage: Storage, useDefaultSettings: boolean, fileName?: string): Promise<string | void> {
+    async backup(
+        targetStorage: Storage,
+        useDefaultSettings: boolean,
+        fileName?: string,
+        lightweight: bool,
```
Marishka17:
```suggestion
        lightweight: boolean,
```
But I also think it should be optional

cvat-ui/src/components/export-backup/export-backup-modal.tsx
```diff
@@ -149,6 +156,17 @@ function ExportBackupModal(): JSX.Element {
...
                 />
+                <Form.Item
+                    className='cvat-settings-switch-lightweight'
+                >
+                    <Space>
+                        <Switch
+                            checked={lightweight}
+                            onChange={setLightweight}
+                        />
+                        <Text strong>Make light-weight backup</Text>
```
Marishka17:
```suggestion
                        <Text strong>Make a lightweight backup</Text>
```

cvat-ui/src/components/export-backup/export-backup-modal.tsx
```diff
@@ -35,6 +38,7 @@ const initialValues: FormValues = {
         cloudStorageId: undefined,
     },
     useProjectTargetStorage: true,
+    lightweight: true,
```
Marishka17, line 45: I still think we shouldn't show this option at least in cases where we know that it's useless (e.g. when a task has no cloud data). Was it agreed upon with others?

cvat/apps/engine/background.py
Marishka17: > That way, existing rq backup jobs still can execute

But with another behavior since it's now True :)

bsekachev:
image

I do not think it makes sense to keep this switch for tasks, not created from a cloud storage 

Let's add a tooltip (like "Use default settings" has), as it is not obviously for a user what is difference between light weight backup and regular

bsekachev:
When I export regular backup and lightweight, I can't see both results on requests page, it seems they overwrite each other. Only the latest is visible. I believe it should work like for exports with different formats

image

bsekachev:
image

The layout here does not look well.

Are there any reasons why the field only gets rendered after fetching corresponding cloud storage?
May we show loading indicator instead while fetching?

bsekachev:
image

UI for view/edit is different, may we just always show selector and do this field similar to "Assigned to"? (maybe even align it on the right). 

In this form "Select cloud storage" has "*" mark, making this field obligatory, however I may left it empty. I do not think we need this mark

bsekachev:
For a project containing mixed cloud storage and non cloud storage data, the switcher "Make lightweight backup" seems a little uncertain. I would suggest to rename it, maybe like "Use lightweight backup whenever possible" (with corresponding hint)..? IDK, I would suggest to think on better message here

bsekachev:
@klakhov Could you please look at client changes here?

cvat-core/src/frames.ts
```diff
@@ -701,6 +715,22 @@ function saveJobMeta(meta: FramesMetaData, jobID: number): Promise<FramesMetaDat
     return frameMetaCache[jobID];
 }
 
+function saveTaskMeta(meta: FramesMetaData, taskID: number): Promise<FramesMetaData> {
```
klakhov: This method is complete dublicate of what we have for saveJobMeta, I think we can rewrite it so we have only one function

cvat-core/src/frames.ts
```diff
@@ -949,6 +979,14 @@ export async function patchMeta(jobID: number): Promise<FramesMetaData> {
     return newMeta;
 }
 
+export async function patchTaskMeta(taskID: number, meta: FramesMetaData): Promise<FramesMetaData> {
```
klakhov: Same comment as for saveTaskMeta

cvat-core/src/project.ts
```diff
@@ -235,13 +235,19 @@ export default class Project {
         return result;
     }
 
-    async backup(targetStorage: Storage, useDefaultSettings: boolean, fileName?: string): Promise<string | void> {
+    async backup(
+        targetStorage: Storage,
+        useDefaultSettings: boolean,
+        fileName?: string,
+        lightweight: bool,
```
klakhov:
```suggestion
        lightweight: boolean,
```

Also we cant have a required parameter follow optional parameter fileName

cvat-core/src/server-proxy.ts
```diff
@@ -995,12 +995,14 @@ async function backupTask(
     targetStorage: Storage,
     useDefaultSettings: boolean,
     fileName?: string,
+    lightweight: boolean,
```
klakhov: Same comment here, we cant have required param follow optional one

cvat-core/src/server-proxy.ts
klakhov: Probably the lightweight param should be optional and we should send it only if its specified 
```
...(typeof lightweight === 'boolean' ? { lightweight } : {}),
```

cvat-core/src/server-proxy.ts
```diff
@@ -1075,13 +1077,15 @@ async function backupProject(
...
 ): Promise<string | void> {
     const { backendAPI } = config;
     // keep current default params to 'freeze" them during this request
     const params: Params = {
         ...enableOrganization(),
         ...configureStorage(targetStorage, useDefaultSettings),
         ...(fileName ? { filename: fileName } : {}),
+        ...(!lightweight ? { lightweight } : {}),
```
klakhov: Same comment with optional here

cvat-ui/src/components/export-backup/export-backup-modal.tsx
klakhov: I would agree here, in case there is no `cs` for task we shouldnt even show the form item for this

cvat-ui/src/components/task-page/cloud-storage-editor.tsx
```diff
@@ -0,0 +1,130 @@
...
+        </span>
+    );
+}
+
+async function getCloudStorageById(id: number): Promise<CloudStorage | null> {
+    try {
+        const data = await getCore().cloudStorages.get({ id });
+        if (data.length === 1) {
+            return data[0];
+        }
```
klakhov:
```suggestion
    const [data] = await core.projects.get({ id });
    return data;
```

cvat-ui/src/components/task-page/cloud-storage-editor.tsx
```diff
@@ -0,0 +1,130 @@
...
+            getCloudStorageById(meta.cloudStorageId).then((_cloudStorage) => {
+                setCloudStorage(_cloudStorage);
+                setCloudStorageLoaded(true);
+            });
+        } else {
+            setCloudStorageLoaded(true);
+        }
+    }, [meta]);
+
+    const updateCloudStorage = (_cloudStorage: CloudStorage | null): Promise<void> => (
```
klakhov: The `updateCloudStorage` function updates task metadata, which for the user essentially means that the task itself is being updated. This should trigger the loading spinner.

I recommend merging `develop` and checking how task fields are updated. We use the `onUpdateTask` callback, which dispatches the `updateTaskAsync` action.

I think we should create a dedicated `updateTaskMetadataAsync` action and an `onUpdateTaskMetadata` callback, instead of handling the update logic directly here.

The reason is that within `updateTaskMetadataAsync`, we can dispatch `updateTask` / `updateTaskFailed`, which will update the `activities.updates` fields in the reducer—these are responsible for showing the loading spinner on the task page.

cvat-ui/src/components/task-page/cloud-storage-editor.tsx
```diff
@@ -0,0 +1,130 @@
...
+    useEffect(() => {
+        if (meta && meta.cloudStorageId) {
+            getCloudStorageById(meta.cloudStorageId).then((_cloudStorage) => {
+                setCloudStorage(_cloudStorage);
+                setCloudStorageLoaded(true);
+            });
+        } else {
+            setCloudStorageLoaded(true);
+        }
+    }, [meta]);
```
klakhov: Fetching task metadata and the corresponding cloud storage should be part of the `receiveTask` function for the `task-page`, which manages the loading spinner.
If we try to fetch this data here, the user will see the page before all the necessary data has been fully loaded.

cvat-ui/src/components/task-page/cloud-storage-editor.tsx
```diff
@@ -0,0 +1,130 @@
...
+                reject();
+            });
+        })
+    );
+
+    return (
+        meta && meta.storage === StorageLocation.CLOUD_STORAGE && (
+            <Row justify='space-between' align='middle'>
+                <Col span={12}>
+                    <Row className='cvat-cloud-storage'>
```
klakhov: Something is wrong with class names here

```suggestion
                    <Row className='cvat-task-cloud-storage'>
```

cvat-ui/src/components/task-page/cloud-storage-editor.tsx
```diff
@@ -0,0 +1,130 @@
...
+                        { !cloudStorageEditing && <Col>Connected cloud storage:</Col> }
+                        { !cloudStorageEditing && (
+                            <Col>
+                                { cloudStorageLoaded && cloudStorage && renderCloudStorageName(cloudStorage) }
+                                { cloudStorageLoaded && !cloudStorage && 'MISSING' }
+                                <Text
+                                    className='cvat-issue-tracker-value'
```
klakhov:
```suggestion
                                    className='cvat-task-cloud-storage-?'
```

cvat/apps/engine/task.py
```diff
@@ -1512,7 +1517,8 @@ def _to_abs_frame(rel_frame: int) -> int:
         _create_static_chunks(db_task, media_extractor=extractor, upload_dir=upload_dir)
 
     # Prepare the preview image and save it in the cache
-    TaskFrameProvider(db_task=db_task).get_preview()
+    if not (is_data_in_cloud and is_backup_restore):
+        TaskFrameProvider(db_task=db_task).get_preview()
```
zhiltsov-max, line 1521: Checking only for backup restore here looks inconsistent with other task modes, consider changing the condition to only skip the case with a detached CS from a lightweight backup.

cvat/apps/engine/serializers.py
```diff
@@ -2752,21 +2760,36 @@ class Meta:
...
-    deleted_frames = serializers.ListField(child=serializers.IntegerField(min_value=0))
+    deleted_frames = serializers.ListField(child=serializers.IntegerField(min_value=0), required=False)
+    cloud_storage_id = serializers.IntegerField(required=False, allow_null=True)
 
     class Meta:
         model = models.Data
-        fields = ('deleted_frames',)
+        fields = ('deleted_frames', 'cloud_storage_id')
+
+    def validate_cloud_storage_id(self, cloud_storage_id: int):
```
zhiltsov-max, line 2770: We might need to invalidate existing chunks or ensure that the task manifest works for the new cloud storage after the attached cloud storage is changed. We do check image hashes, so there will be an error if images are different, but it's checked when new chunks are created. The chunks will be invalidated after `CVAT_CHUNK_CACHE_TTL` by default. I think that checking the manifest contents on changing the attached storage is more correct in this situation, because it's the point where an invalid storage could be introduced into the task.

cvat/apps/engine/background.py
Eldies: Now false by default

cvat/apps/engine/background.py
Eldies: done

cvat/apps/engine/backup.py
Eldies: Well, the manifest comes from backup archive so it should be in client_files, and media is in a cloud storage, so it should be in server_files. I don't really have any ideas how to do it in another way

cvat/apps/engine/backup.py
Eldies: Set it to None here, which is functionally the same as False, but does not presume that it is a default value

cvat/apps/engine/cache.py
Eldies: done

cvat-core/src/server-proxy.ts
Eldies: done

cvat-core/src/server-response-types.ts
Eldies: created a new type specifically for this field

cvat-ui/src/components/export-backup/export-backup-modal.tsx
Eldies: Is it a good idea to make an additional request for meta every time backup dialog is showed? Also, it should work also for projects, and they could have a lot of tasks

cvat-core/src/frames.ts
Eldies: done

cvat-core/src/frames.ts
Eldies: done

cvat-core/src/server-proxy.ts
Eldies: fixed

cvat-core/src/server-proxy.ts
Eldies: fixed

cvat-ui/src/components/task-page/cloud-storage-editor.tsx
Eldies: created updateTaskMetadataAsync and update meta using it. Although, there is still a short loading without spinner, because I don't think loading new cloud storage can be done in updateTaskMetadataAsync

cvat-ui/src/components/task-page/cloud-storage-editor.tsx
Eldies: done

cvat-ui/src/components/task-page/cloud-storage-editor.tsx
Eldies: fixed

cvat/apps/engine/task.py
Eldies: The condition here already checks that the task is restored from backup and it has data in cloud, so it is the case with a detached CS from a lightweight backup

cvat-ui/src/actions/tasks-actions.ts
```diff
@@ -356,6 +356,23 @@ export function updateTaskAsync(
     };
 }
 
+export function updateTaskMetadataAsync(
+    taskInstance: Task,
+    taskMeta: FramesMetaData,
+): ThunkAction<Promise<Task>> {
+    return async (dispatch: ThunkDispatch): Promise<Task> => {
```
klakhov:
```suggestion
export function updateTaskMetadataAsync(
    taskInstance: Task,
    taskMeta: FramesMetaData,
): ThunkAction<Promise<FramesMetaData>> {
    return async (dispatch: ThunkDispatch): Promise<FramesMetaData> => {
```

cvat-ui/src/components/task-page/task-page.tsx
```diff
@@ -46,18 +51,29 @@ function TaskPageComponent(): JSX.Element {
 
     const receiveTask = (): Promise<Task[]> => {
```
klakhov: Lets rewrite this function using async/await. Its so confusing to read right now

klakhov (commented):
As mentioned before, lets move this under assignee and make the text just 'Cloud storage'. Also in 'secondary' so its gray
image

cvat-ui/src/components/task-page/cloud-storage-editor.tsx
klakhov: Why do we need to load cloud storage after update again? I suppose we already have a list of cloud storages when we search for new ones in SelectCloudStorage.

cvat-ui/src/components/task-page/cloud-storage-editor.tsx
```diff
@@ -0,0 +1,53 @@
...
+                searchPhrase={searchPhrase}
+                cloudStorage={cloudStorageInstance}
+                setSearchPhrase={setSearchPhrase}
+                onSelectCloudStorage={(_cloudStorage: CloudStorage | null) => {
+                    if (_cloudStorage) {
+                        taskMeta.cloudStorageId = _cloudStorage.id;
+                        onUpdateTaskMeta(taskMeta);
+                    }
+                }}
+                secondary
```
klakhov:
```suggestion
```

klakhov (commented):
Should this field be clearable, `cs=null`? If it is, when I clear it it should make update request. If not, then we shouldnt be able to do it. I reccomend implementing behavior just as it is now for assignee selector
cs-bug

cvat-ui/src/components/task-page/cloud-storage-editor.tsx
Eldies: I forgot to commit related changes to SelectCloudStorage. It is to make "Select cloud storage" text color the same as "Assigned to" text color

cvat-ui/src/components/task-page/task-page.tsx
Eldies: done

Eldies:
> lets move this under assignee and make the text just 'Cloud storage'. Also in 'secondary' so its gray

done

> Should this field be clearable, cs=null? If it is, when I clear it it should make update request. If not, then we shouldnt be able to do it.

done

cvat-ui/src/components/task-page/cloud-storage-editor.tsx
Eldies: removed reloading of cloud storage

cvat-ui/src/components/select-cloud-storage/select-cloud-storage.tsx
```diff
@@ -19,6 +19,8 @@ export interface Props {
     name?: string[];
     setSearchPhrase: (searchPhrase: string) => void;
     onSelectCloudStorage: (cloudStorageId: number | null) => void;
+    required?: boolean;
+    label? : string;
```
klakhov:
```suggestion
    label?: JSX.Element;
```

cvat-ui/src/components/select-cloud-storage/select-cloud-storage.tsx
```diff
@@ -50,7 +52,13 @@ const searchCloudStoragesWrapper = debounce((phrase, setList) => {
 
 function SelectCloudStorage(props: Props): JSX.Element {
     const {
-        searchPhrase, cloudStorage, name, setSearchPhrase, onSelectCloudStorage,
+        searchPhrase,
+        cloudStorage,
+        name,
+        setSearchPhrase,
+        onSelectCloudStorage,
+        required,
```
klakhov: As far as I understand now, it is not possible to change `cs` from `something` to `null`, right? Then there is no point in having `required` param. Now every usage of `SelectCloudStorage` specifies it to `true`

cvat-ui/src/components/task-page/cloud-storage-editor.tsx
```diff
@@ -0,0 +1,58 @@
...
+    taskMeta: FramesMetaData,
+    cloudStorageInstance: CloudStorage,
+    onUpdateTaskMeta: (meta: FramesMetaData) => Promise<void>;
+}
+
+export async function getCloudStorageById(id: number): Promise<CloudStorage | null> {
+    try {
+        const [data] = await getCore().cloudStorages.get({ id });
+        return data;
+    } catch (error) {
```
klakhov:
```suggestion
    } catch (error: any) {
```

cvat-ui/src/components/task-page/task-page.tsx
```diff
@@ -44,29 +49,37 @@ function TaskPageComponent(): JSX.Element {
...
+
+                const meta = await task.meta.get();
+                setTaskMeta(meta);
+
+                if (meta.cloudStorageId) {
+                    const cloudStorage = await getCloudStorageById(meta.cloudStorageId);
+                    setCloudStorageInstance(cloudStorage);
+                } else {
+                    setCloudStorageInstance(null);
+                }
```
klakhov: Do we need this code as its already `null` by default?

cvat-ui/src/components/task-page/task-page.tsx
```diff
@@ -44,29 +49,37 @@ function TaskPageComponent(): JSX.Element {
...
-            promise.then(([task]: Task[]) => {
-                if (task) {
-                    setTaskInstance(task);
-                }
-            }).catch((error: Error) => {
-                notification.error({
-                    message: 'Could not receive the requested task from the server',
-                    description: error.toString(),
-                });
+    const receiveTask = async (): void => {
```
klakhov:
```suggestion
    const receiveTask = async (): Promise<void> => {
```

cvat-ui/src/components/task-page/task-page.tsx
```diff
@@ -44,29 +49,37 @@ function TaskPageComponent(): JSX.Element {
...
+                setTaskMeta(meta);
+
+                if (meta.cloudStorageId) {
+                    const cloudStorage = await getCloudStorageById(meta.cloudStorageId);
+                    setCloudStorageInstance(cloudStorage);
+                } else {
+                    setCloudStorageInstance(null);
+                }
+            }
+        } catch (error: Error) {
```
klakhov:
```suggestion
        } catch (error: any) {
```

cvat-ui/src/components/task-page/cloud-storage-editor.tsx
```diff
@@ -0,0 +1,58 @@
...
+        notification.error({
+            message: 'Could not fetch a cloud storage',
+            description: error.toString(),
+        });
+    }
+    return null;
+}
+
+export default function CloudStorageEditorComponent(props: Props): JSX.Element {
```
klakhov:
```suggestion
export default function CloudStorageEditorComponent(props: Props): JSX.Element | null {
```

cvat-ui/src/components/task-page/cloud-storage-editor.tsx
```diff
@@ -0,0 +1,58 @@
...
+                        taskMeta.cloudStorageId = _cloudStorage.id;
+                        onUpdateTaskMeta(taskMeta);
+                    } else {
+                        setSearchPhrase(cloudStorageInstance ? cloudStorageInstance.displayName : '');
+                    }
+                }}
+                label={label}
+            />
+        )
+    );
```
klakhov:
```suggestion
    if (taskMeta.storage !== StorageLocation.CLOUD_STORAGE) {
        return null;
    }

    return (
        <SelectCloudStorage
            searchPhrase={searchPhrase}
            cloudStorage={cloudStorageInstance}
            setSearchPhrase={setSearchPhrase}
            onSelectCloudStorage={(_cloudStorage: CloudStorage | null) => {
                if (_cloudStorage) {
                    taskMeta.cloudStorageId = _cloudStorage.id;
                    onUpdateTaskMeta(taskMeta);
                } else {
                    setSearchPhrase(cloudStorageInstance ? cloudStorageInstance.displayName : '');
                }
            }}
            label={label}
        />
    );
```

klakhov (commented):
Seems like unit tests are failing

klakhov:
We need to add padding here inbetween fields and remove `:` for consistency
image

cvat-ui/src/components/task-page/task-page.tsx
Eldies: removed it

cvat-ui/src/components/select-cloud-storage/select-cloud-storage.tsx
Eldies: removed it

Eldies:
> We need to add padding here inbetween fields and remove `:` for consistency

Done. Added two spaces to `Assigned to  ` to align it with Cloud storage. Is there a better way to do it?

Eldies:
> When I export regular backup and lightweight, I can't see both results on requests page, it seems they overwrite each other. Only the latest is visible. I believe it should work like for exports with different formats

done
image

cvat/apps/engine/serializers.py
Eldies: invalidating existing task chunks now. Checking that the attached storage contains all the images and their checksum is the same can take some time

klakhov (approved):

SpecLad:
There's no changelog entry for this - could you add one?

PR #9548, opened 2025-06-19
UI: Don't pass the shapes to the lambda function when continuing tracking

SpecLad:
### Motivation and context

Currently, the UI always passes the current shape points when calling the tracker function. When initializing the tracking state, this makes sense.

However, when continuing tracking, the utility of this is dubious.

* The point coordinates passed in this way are those for the _previous_ frame; however, the function does not have access to that frame itself. Therefore, it cannot examine the pixel data that the shape covers.

* In every case, these shapes are data that the function has already seen: it's either the shape that the function received in the initial request, or the shape that the function has output in the previous request. Therefore, if the function really needs this data, it can already ensure access to it by including it into the tracking state.

OTOH, including the shapes has some downsides:

* It's unnecessary data bloating the request. For rectangles, this isn't a big deal, but for, say, masks the extra data could be significant.

* It complicates the mental model for the tracker API.

Because of all this, this patch removes the passing of the shapes.

This is a breaking change. However, none of the example trackers in this repository use the previous-frame shapes, and neither does the private SAM2 tracker, so chances are that most 3rd-party trackers won't either. I have included compatibility code that passes a dummy `shapes` array to the Nuclio function, so that code that relies on its _length_ will still work.

### How has this been tested?

I have tried using TransT and SiamMask (and they still work).

### Checklist

### License

### DIFF
--- a/changelog.d/20250619_171731_roman_either_states_or_shapes.md
+++ b/changelog.d/20250619_171731_roman_either_states_or_shapes.md
@@ -1,0 +1,5 @@
+### Changed
+
+- Nuclio tracker functions are no longer passed the previous frame's shapes
+  when continuing the tracking
+  (<https://github.com/cvat-ai/cvat/pull/9548>)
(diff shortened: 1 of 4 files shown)


SpecLad:
@bsekachev Could you review this?

PR #9550, opened 2025-06-19
Split rest_api test runs in two to improve performance

archibald1418:
This splits rest_api test execution between two groups of tests executed on two different github-runners.
rest_api_testing job now consists of 3 workers, runtime reduced from 50min to 25min

The two slowest test classes `tests/python/rest_api/test_tasks.py::{TestTaskData, TestPostTaskData}` are now run separately and parametrized by `CVAT_ALLOW_STATIC_CACHE` server parameter. 

Allure reports and coverage report routines were updated accordingly, similar to `e2e_testing` github workflow job

### How has this been tested?

This is a test refactoring PR, so the only test for this is that tests don't crash or fail unexpectedly. The three groups are executed separately each using their own instance of the current testing infra

### Checklist

### License


tests/python/rest_api/test_tasks_slow.py
SpecLad, line 1: Why do you want to move these tests to a separate file? Pytest has markers; you should be able to just mark the tests you want and filter by marker when running pytest.

archibald1418: I'll try that, this should work fine. But still, is it a bad idea to split an enormous test file and its massive dependencies?

SpecLad: I'm all for splitting large files, but it should be based on a criteria that is relatively static and easy to determine. E.g. when I'm adding a new test, it's usually clear whether it should go into `test_tasks` or, say, `test_projects`. But how do I know if it's slow or not? And if a test becomes slow over time, we shouldn't have to move the entire code into a different file.

archibald1418: That's actually a good point, and marks would describe a test better than a new file. They are also more extensible and easier to maintain

tests/python/rest_api/test_tasks.py
```diff
@@ -1004,6 +1004,7 @@ def test_datumaro_export_without_annotations_includes_image_info(
 @pytest.mark.usefixtures("restore_redis_ondisk_per_function")
 @pytest.mark.usefixtures("restore_redis_ondisk_after_class")
 @pytest.mark.usefixtures("restore_redis_inmem_per_function")
+@pytest.mark.slow
```
zhiltsov-max: Marking tests `slow` doesn't seem a scalable criteria to me. If we want to split the runs further into 3 or more runners, we're going to have `slow1`, `slow2` or `slow` and `very_slow`?

archibald1418: I agree, but for purpose of this pull request, and for just 2 classes, I would suggest this is enough. ASA we have a system of more detailed and scalable test performance categories (as one of the outputs of @orazvaliev 's work on perf testing) this will be more structured. I suppose such classification should be done in a different PR

This marks the only two massive outliers in the sample of test runtimes by classes, the rest of the distribution is much more contained (data from Allure reports)

archibald1418: image

orazvaliev: I still think that splitting test_tasks.py file is a good idea. In this case no execution performance tags are required.
Basically each test class that has long execution time or just too many code lines can be placed into separate test module

orazvaliev: @SpecLad raised a good point regarding placing new tests, but this won't be an issue if we have one module per class

tests/python/rest_api/test_tasks_slow.py
archibald1418: Moved them to a `test_task_data.py` instead of `test_task_slow.py`, I suppose this is clear enough

tests/python/rest_api/test_tasks.py
archibald1418: Revived commits with file splitting, created a file `test_task_data.py` instead

tests/python/shared/tasks/base.py
```diff
@@ -0,0 +1,37 @@
...
+
+import attrs
+from cvat_sdk.api_client import models
+
+from .interface import ITaskSpec
+from .utils import parse_frame_step
+
+
+@attrs.define
+class _TaskSpecBase(ITaskSpec):
```
orazvaliev: leading _ underscore should be removed here and in other places where you expect a name to be imported in another module

tests/python/rest_api/test_task_data.py
```diff
@@ -0,0 +1,2247 @@
...
+from shared.utils.config import get_method, make_api_client, patch_method, post_method
+from shared.utils.helpers import (
+    generate_image_file,
+    generate_image_files,
+    generate_manifest,
+    generate_video_file,
+    read_video_file,
+)
+
+from ._test_base import _TestTasksBase
```
orazvaliev: It's better to replace relative imports with absolute for consistency. As i see all other imports here are from the root

.github/workflows/main.yml
```diff
@@ -201,13 +219,28 @@ jobs:
...
+            }}
           path: "${{ github.workspace }}/rest_api_testing"
+
       - name: Upload allure results
         if: always()
         uses: actions/upload-artifact@v4
         with:
-          name: allure-results-rest-api
+          name: ${{
+            contains(matrix.pattern, 'not')
```
orazvaliev, line 238: IMO this pattern is unnecessary complicated. Basically the main issue is that in any case we end up having too many artifacts 
image
I'd suggest using merge action https://github.com/actions/upload-artifact/blob/main/merge/README.md. If you feel that it can be achieved in this PR, please do it. Otherwise we will create PR.

But in any case please simplify the filter. Smth like this would be enough 
`allure-results-tests-api-${{matrix.pattern}}-cache-${{matrix.allow_cache}}`

archibald1418: Simplifying the filter won't work, that way we will have spaces in folder names, since we have a 'not test_task_data' as one of the pytest filters.
Merging seems more correct but the only issue is that report will show redundant launches of the same tests. If only we had a way to reflect the server parameters in allure reports stats

orazvaliev: i don't think we should care about merging similar tests with cache / no_cache parameter. Let's sort this out in the next PR.

archibald1418:
TODO: add license on new files

PR #9553, opened 2025-06-20
WIP. Performance regression testing using K6

orazvaliev:
### Checklist

### License


tests/perf/docker-compose-perf.yml
```diff
@@ -0,0 +1,44 @@
+name: cvat-performance
+include:
+  - path:
+    - ../../docker-compose.yml
+    - ../../docker-compose.dev.yml
+    - ../../components/serverless/docker-compose.serverless.yml
+    - ../docker-compose.minio.yml
+    - ../docker-compose.file_share.yml
+    # - docker-compose-perf-cpuset.yml
```
archibald1418: Should this be commented or not?

tests/perf/README.md
archibald1418, line 1: @paveltovchigrechko could you please look at the manual and see if you like everything?

tests/perf/docker-compose-perf.yml
```diff
@@ -0,0 +1,44 @@
...
+services:
+  perf-k6:
+    image: grafana/k6:latest
+    container_name: perf_k6
+    entrypoint:
+    - k6
```
archibald1418:
```suggestion
      - k6
```
Please add indentation to other array values

tests/perf/perfkit/baselines.py
```diff
@@ -0,0 +1,62 @@
+import json
```
archibald1418, line 4:
```python
#

import json
...
```

Please add this licensing header to all of the new test code files

tests/perf/requirements.txt
```diff
@@ -0,0 +1,6 @@
+typer>=0.16.0
```
archibald1418: Why the `>=` ? Shouldn't we pin this?

tests/perf/scripts/libs/default-specs.js
```diff
@@ -0,0 +1,83 @@
+// Copyright (C) CVAT.ai Corporation
```
archibald1418, line 1: If it's the same vanilla js logic, could it work if we refactored this to `tests/share/default-specs.js` and then both `cypress` and `perf` could import them from there?

tests/perf/scripts/tests/regression/tasks.js
```diff
@@ -0,0 +1,42 @@
+import APITasks from '../../libs/api/tasks.js';
+import APIAuth from '../../libs/api/auth.js';
+import { ADMIN_PASSWORD, ADMIN_USERNAME } from '../../variables/variables.js';
+
+export const options = {
+  scenarios: {
+      contacts: {
+          executor: 'constant-arrival-rate',
+          // How long the test lasts
```
archibald1418: You wouldn't need most of the comments if this was a `typescript` file, typed with `@types/k6` that I just downloaded
https://grafana.com/docs/k6/latest/using-k6/javascript-typescript-compatibility-mode/

tests/perf/scripts/variables/variables.js
archibald1418, line 1: I suppose this should be named `constants.js` or `const.js` rather than `variables.js`

tests/perf/perfkit/baselines.py
```diff
@@ -0,0 +1,62 @@
...
+from plumbum import local
+from rich.console import Console
+
+from perfkit.config import BASELINE_FILE
+from perfkit.k6_summary import K6Summary
+
+
+git = local["git"]
+console = Console()
```
archibald1418: Can the `Console` be moved to `config.py` and shared between files or is it required that each file should have its own `Console` ?

tests/perf/perfkit/baselines.py
```diff
@@ -0,0 +1,62 @@
...
+        console.print(f"[yellow] Overwriting existing baseline for test '{test_key}'[/yellow]")
+
+    baseline_content[test_key][commit_id] = k6_summary_output.as_dict()
+    if alias is not None:
+        baseline_content["aliases"][alias] = commit_id
+    save_baselines(baseline_content)
+    console.print(f"[green] Baseline added for test '{test_key}' for commit '{commit_id}'[/green]")
+
+
+def resolve_commit_by_alias(baselines: dict, alias: str) -> str:
```
archibald1418, line 51:
```python
from typing import NoReturn
def resolve_commit_by_alias(baselines: dict, alias: str) -> str | NoReturn:
```

et sim., for all functions that raise

tests/perf/perfkit/cluster.py
```diff
@@ -0,0 +1,122 @@
...
+    exec_cmd = docker[
+        "exec",
+        "-i",
+        container_name,
+        "bash",
+        "-c",
+        f'echo "{admin_code}" | python3 ~/manage.py shell',
+    ]
+    exec_cmd()
+    time.sleep(10)
```
archibald1418:
```python
    time.sleep(WAIT_FOR_CLUSTER)
```

Putting the constant on top will signal to the reader that there is sleeping involved

tests/perf/perfkit/main.py
```diff
@@ -0,0 +1,237 @@
...
+        nonlocal test_invalid
+        stop_k6_docker()
+        test_invalid.set()
+
+    def run_k6_test(i) -> K6Summary | None:
+        if test_invalid.is_set():
+            return
+        print_info(f"run {i + 1}/{runs}...")
+        start_cluster()
+        time.sleep(1)
```
archibald1418: Why sleep again if you've just slept for 10 whole seconds in the end of `start_cluster()`?

tests/perf/perfkit/comparison_report.py
```diff
@@ -0,0 +1,76 @@
...
+            )
+        )
+
+    for metric_name, metric in k6_output_summary.metrics.items():
+        for stat_name, stat_value in metric.as_dict().items():
+            metric_stat = f"{metric_name}_{stat_name}"
+            try:
+                baseline_value = baseline[metric_name][stat_name]
+            except KeyError:
+                # add_report_row(metric_stat)
```
archibald1418: Should the comments be here?

tests/perf/perfkit/comparison_report.py
```diff
@@ -0,0 +1,76 @@
...
+                continue
+
+            allowed_delta: float | None = None
+            if metric_name in ALLOWED_DELTAS and stat_name in ALLOWED_DELTAS[metric_name]:
+                allowed_delta = ALLOWED_DELTAS[metric_name][stat_name]
+            else:
+                # add_report_row(metric_stat, baseline_value, stat_value)
+                continue
+
+            assert allowed_delta
```
archibald1418: Perhaps this is a bit cleaner? 

```suggestion
            if metric_name  not in ALLOWED_DELTAS and stat_name not in ALLOWED_DELTAS[metric_name]:
                continue
            allowed_delta = ALLOWED_DELTAS[metric_name][stat_name]
```

This maintains the fail fast approach that you used just before
And the assertion wouldn't be necessary

tests/perf/perfkit/comparison_report.py
```diff
@@ -0,0 +1,76 @@
...
+MISSING = "MISSING"
+OK = "✅"
+FAIL = "❌"
+
+ReportRow = tuple[str, str, str, str, str]
+
+
+def build_report(
+    k6_output_summary: K6Summary, baseline: dict[str, dict[str, float]]
+) -> tuple[list, bool]:
```
archibald1418:
```suggestion
) -> tuple[list[ReportRow], bool]:
```

tests/perf/perfkit/config.py
```diff
@@ -0,0 +1,28 @@
...
+
+ROOT_DIR = Path(__file__).parent.parent
+
+BASE_URL = "http://localhost:8080/api"
+PROMETHEUS_URL = "http://localhost:9190"
+URL_SERVER_ABOUT = f"{BASE_URL}/server/about"
+CVAT_SERVER_SERVICE = "cvat_server"
+K6_PERF_SERVICE = "k6-perf"
+
+BASELINE_FILE = Path.home() / ".perf_baselines.json"
```
archibald1418: This json will be stored on the instance? Do we also need it to run the perfkit locally?

tests/perf/perfkit/console_print.py
```diff
@@ -0,0 +1,26 @@
+from rich.console import Console
+import typer
+
+
+console = Console()
+
+
+def exit_with_error(msg: str, bold: bool = False) -> None:
```
archibald1418, line 10:
```suggestion
def exit_with_error(msg: str, bold: bool = False) -> NoReturn:
```
This does not return anything
https://docs.python.org/3/library/typing.html#typing.NoReturn

tests/perf/perfkit/k6_summary.py
```diff
@@ -0,0 +1,134 @@
...
+class K6MetricStats:
+    avg: float | None = None
+    max: float | None = None
+    min: float | None = None
+    med: float | None = None
+    p90: float | None = None
+    p95: float | None = None
+    count: float | None = None
+    value: float | None = None
+    rate: float | None = None
```
archibald1418, line 22:
```suggestion
    avg: Optional[float] = None
    max: Optional[float] = None
    min: Optional[float] = None
    med: Optional[float] = None
    p90: Optional[float] = None
    p95: Optional[float] = None
    count: Optional[float]= None
    value: Optional[float] = None
    rate: Optional[float] = None
```

looks nicer IMHO

archibald1418:
Please, add a small description for the pull request. Like, what it does and which features the perfkit introduces and how it measures the load on an endpoint. Can be done with e.g. relevant links to k6 docs as well

tests/perf/perfkit/k6_summary.py
```diff
@@ -0,0 +1,134 @@
...
+        stats.clear()
+        stats.update(filtered)
+        return stats
+
+
+@dataclass
+class K6Summary:
+    metrics: dict[str, K6MetricStats]
+
+    def __add__(self, other: "K6Summary") -> "K6Summary":
```
archibald1418, line 36: In modern Python it's usually done like so:

```python
from typing import Self
```

```suggestion
    def __add__(self, other: Self) -> Self | NoReturn:
```

tests/perf/perfkit/k6_summary.py
```diff
@@ -0,0 +1,134 @@
...
+            raise ValueError("Metric keys mismatch between summaries")
+
+        new_metrics: dict[str, K6MetricStats] = {}
+
+        for metric_name in self.metrics:
+            m1 = self.metrics[metric_name]
+            m2 = other.metrics[metric_name]
+
+            combined = K6MetricStats()
+            for field in K6MetricStats.__annotations__:
```
archibald1418, line 47:
```suggestion
            for field in dataclasses.fields(K6MetricStats):
            ...
            setattr(combined, field.name, ...)
```

tests/perf/perfkit/k6_profile.py
```diff
@@ -0,0 +1,62 @@
...
+            *self._build_env_list(),
+            "--out",
+            "experimental-prometheus-rw",
+            "--address",
+            "0.0.0.0:6565",
+            "--tag",
+            f"testid={self.test_file.name}",
+            "--summary-export=/output/summary.json",
+            # "--no-connection-reuse",
+            # "--no-vu-connection-reuse",
```
archibald1418: Are these needed?

tests/perf/perfkit/k6_summary.py
```diff
@@ -0,0 +1,134 @@
...
+                elif v1 is not None:
+                    setattr(combined, field, v1)
+                elif v2 is not None:
+                    setattr(combined, field, v2)
+                # else leave it as None
+            new_metrics[metric_name] = combined
+        return K6Summary(metrics=new_metrics)
+
+    def __truediv__(self, divisor: float) -> "K6Summary":
+        if divisor == 0:
```
archibald1418, line 67: It either can be a `float` or cannot

tests/perf/perfkit/k6_summary.py
```diff
@@ -0,0 +1,134 @@
...
+                    )
+                    consistent = False
+
+        return consistent
+
+    def as_dict(self) -> dict:
+        output = {}
+        for metric, stats in self.metrics.items():
+            output[metric] = stats.as_dict()
+        return output
```
archibald1418, line 117:
```suggestion
	return {metric: stats.as_dict() for metric, stats in self.metrics.items()}
```

tests/perf/perfkit/k6_summary.py
```diff
@@ -0,0 +1,134 @@
...
+
+        for metric_name, metric in self.metrics.items():
+            divided = K6MetricStats()
+            for field, value in metric.as_dict().items():
+                setattr(divided, field, value / divisor)
+            new_metrics[metric_name] = divided
+
+        return K6Summary(metrics=new_metrics)
+
+    def compare(self, other: "K6Summary", allowed_deltas: dict[str, dict[str, float]]) -> bool:
```
archibald1418, line 80: Can `allowed_deltas`  be something different than `ALLOWED_DELTAS` ?

If not, this could could probably be a:
```suggestion
    def __eq__(self, other: Self) -> bool:
```

Even if it can, we can inject them by constructor and lookup with `self.allowed_deltas`

tests/perf/perfkit/k6_summary.py
```diff
@@ -0,0 +1,134 @@
+from dataclasses import dataclass, asdict
+import json
+import pathlib
+from perfkit.console_print import print_error
+
+
+@dataclass
```
archibald1418:
```suggestion
@dataclass(KW_ONLY=True)
```

1. allows any order of args which are quite similar
1. forces the caller to spell them out, avoids confusion
1. allows to pass different `**dict`s

tests/perf/perfkit/main.py
```diff
@@ -0,0 +1,237 @@
...
+        stop_metrics = start_metrics_watcher(max_alerts=20, on_threshold_exceeded=interrupt_test)
+        print_info("starting warmup")
+        exit_code = run_k6_docker(warmup_profile)
+        if exit_code != 0:
```
archibald1418, line 60:
```suggestion
        if exit_code:
```

tests/perf/perfkit/main.py
```diff
@@ -0,0 +1,237 @@
...
+):
+    test_invalid = threading.Event()
+    k6_metrics_total: list[K6Summary] = []
+
+    def interrupt_test() -> None:
+        nonlocal test_invalid
+        stop_k6_docker()
+        test_invalid.set()
+
+    def run_k6_test(i) -> K6Summary | None:
```
archibald1418, line 52:
```suggestion
    def run_k6_test(i) -> K6Summary | None | NoReturn:
```

archibald1418, line 52: This function does many different things, I would suggest breaking it into stages:
`run_test`, `run_warmup`, `run_golden`

tests/perf/perfkit/main.py
```diff
@@ -0,0 +1,237 @@
...
+        elif alias:
+            try:
+                commit_value = resolve_commit_by_alias(baselines, alias)
+            except RuntimeError as err:
+                exit_with_error(f"Can't resolve alias '{alias}': {err}")
+            else:
+                print_info(f"Resolved alias '{alias}' → commit '{commit_value}'")
+        else:
+            exit_with_error(f"Must provide either --commit or --alias")
+        assert commit_value
```
archibald1418, line 124: Perhaps a message explaining what happened?

tests/perf/perfkit/main.py
```diff
@@ -0,0 +1,237 @@
...
+
+    test_key = Path(test_file).stem
+    baselines = load_baselines()
+
+    commit_value = resolve_commit()
+    # check test entry exists
+    try:
+        baseline = resolve_test_baseline(baselines, test_key, commit_value)
+    except RuntimeError:
+        exit_with_error(f"No baseline for test {test_key}'")
```
archibald1418:
```suggestion
    def exit_if_test_entry_not_exists() -> None | NoReturn:
	    try:
	        baseline = resolve_test_baseline(baselines, test_key, commit_value)
	    except RuntimeError:
	        exit_with_error(f"No baseline for test {test_key}'")
     
```

tests/perf/perfkit/main.py
```diff
@@ -0,0 +1,237 @@
...
+        f.write(f"Regression Report for test: {test_key} | {datetime.datetime.now()}\n\n")
+        f.write(
+            "{:<40} {:>10} {:>10} {:>10} {:>6}\n".format(
+                "Metric", "Baseline", "Actual", "Delta", "Status"
+            )
+        )
+        f.write("-" * 80 + "\n")
+        for row in comparison_report:
+            f.write("{:<40} {:>10} {:>10} {:>10} {:>6}\n".format(*row))
+        f.write("\n")
```
archibald1418:
```suggestion
   def write_report_to_file(report: , report_file_path,  test_key) -> None: ...
```

tests/perf/perfkit/main.py
```diff
@@ -0,0 +1,237 @@
...
+        exit_with_error(f"Alias '{alias}' not found")
+
+
+@golden_app.command("show")
+def show():
+    data = load_baselines()
+    if not data:
+        print_info("no golden baselines in a file. Try to `run-golden`")
+        return typer.Exit(0)
+    aliases = data.get("aliases", {})
```
archibald1418:
```suggestion
    aliases = data.pop("aliases", {})
```

and then remove the redundant `if continue` in the loop

tests/perf/perfkit/main.py
```diff
@@ -0,0 +1,237 @@
...
+@golden_app.command("remove-alias")
+def remove_alias(alias: str):
+    data = load_baselines()
+    aliases = data.get("aliases", {})
+    if alias in aliases:
+        del aliases[alias]
+        save_baselines(data)
+        print_success(f"Alias '{alias}' removed")
+    else:
+        exit_with_error(f"Alias '{alias}' not found")
```
archibald1418: Let's be consistent and tr failing early here as we did before

```suggestion
	if alias not in aliases:
		return exit_with_error(f"Alias '{alias}' not found")
	del aliases[alias]
    save_baselines(data)
    print_success(f"Alias '{alias}' removed")
```

tests/perf/perfkit/main.py
```diff
@@ -0,0 +1,237 @@
...
+app.add_typer(golden_app)
+
+
+@app.command(
+    "run-golden", help="Run and record golden test execution that will be used for comparison."
+)
+def run_golden(
+    test_file: str = typer.Argument(..., help="K6 test script to run"),
+    runs: int = typer.Option(DEFAULT_RUNS, help="Number of runs"),
+    # commit: Optional[str] = typer.Option(None, help="Product version or commit hash. Default is the last commit."),
```
archibald1418:
```suggestion
```

tests/perf/scripts/libs/api/tasks.js
```diff
@@ -0,0 +1,39 @@
+
+import { check } from 'k6';
+import http from 'k6/http';
+
+import { defaultTaskSpec } from '../default-specs.js';
+import { BASE_URL } from '../../variables/variables.js'
+
+function createTask(authKey) {
+    const { taskSpec, dataSpec, extras } = defaultTaskSpec({
```
archibald1418:
```suggestion
    const { taskSpec } = defaultTaskSpec({
```

archibald1418: actually, eslint catches this, there are logs

archibald1418:
@orazvaliev you are invited to fix the remaining linting errors locally

tests/perf/scripts/variables/variables.js
```diff
@@ -0,0 +1,4 @@
+export const BASE_URL = 'http://localhost:8080/api/';
+export const ADMIN_EMAIL = 'admin@localhost.company';
+export const ADMIN_USERNAME = 'admin';
+export const ADMIN_PASSWORD = '12qwaszx';
```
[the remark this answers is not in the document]
archibald1418: It's a test code run on a local or private instance, I don't think it's an issue

tests/perf/scripts/tests/regression/tasks.js
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export function setup() {
+    const token = APIAuth.login(ADMIN_USERNAME, ADMIN_PASSWORD);
+    const createdTasks = createTasks(token, 30);
+    return { token, tasksData: createdTasks };
+}
+
+export default function (data) {
+    const randomTask = data.tasksData[data.tasksData.length % __VU];
```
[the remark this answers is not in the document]
archibald1418: This is a good catch, I guess it's a bug

tests/perf/perfkit/comparison_report.py
orazvaliev: I like your code more actually. It's not about `fail fast` IMO (because it has no failure and it's fast enough already), but the logic is more straightforward.

tests/perf/perfkit/config.py
orazvaliev: It will create a file on a disk if it does not exist

tests/perf/perfkit/k6_summary.py
orazvaliev: The same code actually. Style issue

tests/perf/perfkit/k6_summary.py
orazvaliev: `float | int` should work

tests/perf/perfkit/k6_summary.py
orazvaliev: It's shorter but harder to read. I would prefer to be more explicit

tests/perf/perfkit/k6_summary.py
orazvaliev: I know that using magic methods is a prettier way but in my case this method can be extended in the future. `__eq__` does not give you any flexibility of comparing objects. If i make ALLOWED_DELTAS as a constant this will add another dependency in my method. Not good at all

tests/perf/perfkit/main.py
orazvaliev: I prefer to be more explicit in general. There are languages where 0 is true for example. So i don't want to waste time thinking about python's `bool()`

tests/perf/perfkit/main.py
orazvaliev: As i mentioned previously NoReturn can't be used here

orazvaliev: 20 lines of code it's not that much. Also it works as a glue here having log calls and high level methods

tests/perf/perfkit/main.py
orazvaliev: It's just a small self check. Protection against future changes

tests/perf/perfkit/main.py
orazvaliev: Replaced it with Optional baselines return value. Removed try except block

tests/perf/perfkit/main.py
orazvaliev: Good catch. In general i should rework json structure to not confuse `aliases` with test keys at all. Didn't have time for this

tests/perf/perfkit/k6_summary.py
orazvaliev: __annotations__ is a simple dict with class attributes and types. I don't see any reason to use a more complicated API

tests/perf/perfkit/k6_summary.py
orazvaliev: Fixed `Self`, `NoReturn` has different meaning

archibald1418: You're right, NoReturn is used for functions that always raise

archibald1418:
@orazvaliev Food for thought: wouldn't it be nicer to just let the exceptions propagate to the entrypoint instead of `exit()`-ing ? What is your take on this?

tests/perf/perfkit/main.py
archibald1418: Yeah I was mistaken, it is used on functions that always raise or always exit()

tests/perf/perfkit/k6_summary.py
archibald1418: Extending you mean in the children?

orazvaliev:
> @orazvaliev Food for thought: wouldn't it be nicer to just let the exceptions propagate to the entrypoint instead of `exit()`-ing ? What is your take on this?

`exit()` of Typer basically is a `raise Exception` but with fancy formatting.

tests/perf/scripts/libs/api/tasks.js
```diff
@@ -0,0 +1,41 @@
+// Copyright (C) CVAT.ai Corporation
+//
+// SPDX-License-Identifier: MIT
```
archibald1418, line 3:
```suggestion
// SPDX-License-Identifier: MIT

```

PR #9554, opened 2025-06-20
Added middleware to track last_activity of a user

bsekachev:
### Motivation and context
We need a robust way to define when user used CVAT service.

It seems `last_login` field in User model may not get updated for more than a year if user works every day and do not do any logins. It does not seem reliable. Also, different auth methods (basic, social, sso) and a lot of code around make it difficult to guarantee that this field gets updated always. 

Timestamps from analytics are not reliable also:
1. User may send client events with future timestamps (probably it is a bug, but currently we already have some). Thus his account always have "recent" activity from the future.
2. When user account scheduled for deletion, just opening CVAT account does not guarantee analytics will receive any logs from client (e.g. if browser closed immediately, or other issues in client code). Also, user may be authentificated via SDK/CLI what probably will not be displayed in analytics. 

Finally, currently to find non-active users it is necessary to make many queries to the clickhouse database, so, usage report generation is much slower than with just postgres. 

For me middleware with registering last activity once a day, seems the most reliable way.
 

### Checklist

### License


cvat/apps/engine/models.py
```diff
@@ -1172,6 +1172,7 @@ class TrackedShapeAttributeVal(AttributeVal):
 class Profile(models.Model):
     user = models.OneToOneField(User, on_delete=models.CASCADE)
     rating = models.FloatField(default=0.0)
+    last_activity = models.DateTimeField(null=True, blank=True, default=None)
```
SpecLad: I think for consistency with other similar fields, it would be better to call this `last_activity_date`.

cvat/settings/base.py
```diff
@@ -787,3 +787,5 @@ class CVAT_QUEUES(Enum):
             "cron_string": "0 8 * * *",
         }
     )
+
+USER_LAST_ACTIVITY_UPDATEP_PERIOD = timedelta(days=1)
```
SpecLad:
```suggestion
USER_LAST_ACTIVITY_UPDATE_PERIOD = timedelta(days=1)
```

Also, I'm doubtful about using "period" here, because we don't update it _periodically_. Maybe something like `MIN_INTERVAL` would be better?

cvat/apps/engine/middleware.py
```diff
@@ -24,3 +27,22 @@ def __call__(self, request):
         response.headers["X-Request-Id"] = request.uuid
 
         return response
+
+
+class LastActivityMiddleware:
```
SpecLad, line 32: I think you forgot to add this to the middleware list.

bsekachev: Yep, for sure

SpecLad (approved):

cvat/apps/engine/middleware.py
```diff
@@ -24,3 +27,22 @@ def __call__(self, request):
...
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        if request.user.is_authenticated:
+            user = request.user
```
SpecLad: Hmm... does this work with basic or token authentication? IIRC, DRF does those in the view function, so the user might not be available at this point.

bsekachev: Yep, it seems there may be a problem. 
I think if to perform logic after calling drf views, it solves the issue. Is not it?

SpecLad: That should work, but you need to do it in a `finally` block, so that requests that result in exceptions are still counted.

bsekachev: I am not sure we really need to care about failed requests 

It does not bring a lot of profit, however creates other issues. E.g. request may be failing because of the database related issues and in finally block we again trying to save something?

SpecLad: Okay, fair enough.

cvat/apps/engine/middleware.py
```diff
@@ -24,3 +28,27 @@ def __call__(self, request):
...
+            if not profile:
+                return response
+
+            last_activity_date = profile.last_activity_date
+            if (
+                not last_activity_date
+                or (now() - last_activity_date) > settings.USER_LAST_ACTIVITY_UPDATE_MIN_INTERVAL
+            ):
+                # such way we avoid failing and any db updates if the Profile was removed during the request
+                Profile = apps.get_model("engine", "Profile")
```
SpecLad: The `get_model` call seems unnecessary - why not just import `Profile` from `.models`?

bsekachev: circular import

SpecLad: You should be able to do a local import (just replace this line with an import statement).

PR #9563, opened 2025-06-23
Support api token auth in cli and sdk

zhiltsov-max:
### Checklist

### License

### DIFF
--- a/cvat-cli/src/cvat_cli/_internal/common.py
+++ b/cvat-cli/src/cvat_cli/_internal/common.py
@@ -10,13 +10,20 @@ import importlib.util
 import logging
 import os
 import sys
+import textwrap
 from http.client import HTTPConnection
 from pathlib import Path
 from typing import Any, Callable, Optional
 
 import attrs
 import cvat_sdk.auto_annotation as cvataa
-from cvat_sdk.core.client import Client, Config
+from cvat_sdk.core.client import (
+    ApiTokenAuthCredentials,
+    BasicAuthCredentials,
+    Client,
+    Config,
+    Credentials,
+)
 
 from ..version import VERSION
 from .parsers import BuildDictAction, parse_function_parameter
@@ -27,20 +34,34 @@ class CriticalError(Exception):
     pass
 
 
-def get_auth_factory(s: str) -> Callable[[str], tuple[str, str]]:
+CVAT_AUTH_ENV_VAR = "CVAT_AUTH"
+
+
+def get_auth_factory(s: Optional[str]) -> Callable[[str], Credentials]:
     """
     Parse a USER[:PASS] string and return a callable that takes the server URL
-    and returns a (user, pass) tuple for that URL.
+    and returns auth credentials for that URL. If the value is not provided,
+    it will try to read the CVAT_AUTH environment variable for a Personal Access Token (PAT).
     The callable will prompt the user for the password if none was initially supplied.
     """
+
+    if not s:
+        s = os.getenv(CVAT_AUTH_ENV_VAR)
+        if s is not None:
+            return lambda _: ApiTokenAuthCredentials(s)
+        else:
+            s = getpass.getuser()
+
     user, _, password = s.partition(":")
     if not password:
         password = os.environ.get("PASS")
 
     if password:
-        return lambda _: (user, password)
+        return lambda _: BasicAuthCredentials(user, password)
     else:
-        return lambda url: (user, getpass.getpass(f"Password for {user} at {url}: "))
+        return lambda url: BasicAuthCredentials(
+            user, getpass.getpass(f"Password for {user} at {url}: ")
+        )
 
 
 def configure_common_arguments(parser: argparse.ArgumentParser) -> None:
@@ -54,11 +75,18 @@ def configure_common_arguments(parser: argparse.ArgumentParser) -> None:
     parser.add_argument(
         "--auth",
         type=get_auth_factory,
-        metavar="USER:[PASS]",
-        default=getpass.getuser(),
-        help="""defaults to the current user and supports the PASS
-                environment variable or password prompt
-                (default user: %(default)s).""",
+        metavar="USER[:PASS]",
+        default=get_auth_factory(None),  # simple default=None doesn't trigger the type function
+        help=textwrap.dedent(
+            """\
+            User and password to use for authentication;
+            defaults to the current user and supports the PASS
+            environment variable or password prompt.
+            A Personal Access Token (PAT) can be generated on the server
+            and specified in the {} environment variable.
+            (default user: {}).
+        """
+        ).format(CVAT_AUTH_ENV_VAR, getpass.getuser()),
     )
     parser.add_argument(
         "--server-host", type=str, default="http://localhost", help="host (default: %(default)s)"
--- a/cvat-sdk/cvat_sdk/core/client.py
+++ b/cvat-sdk/cvat_sdk/core/client.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 
 import logging
 import urllib.parse
+from abc import ABCMeta
 from collections.abc import Generator, Sequence
 from contextlib import contextmanager, suppress
 from pathlib import Path
@@ -57,6 +58,33 @@ class Config:
 
     cache_dir: Path = attrs.field(converter=Path, default=_DEFAULT_CACHE_DIR)
     """Directory in which to store cached server data"""
+
+
+class Credentials(metaclass=ABCMeta):
+    pass
+
+
+@attrs.define
+class BasicAuthCredentials(Credentials):
+    """
+    Represents basic authentication credentials.
+    """
+
+    user: str
+    """Username for authentication"""
+
+    password: str
+    """Password for authentication"""
+
+
+@attrs.define
+class ApiTokenAuthCredentials(Credentials):
+    """
+    Represents API token authentication credentials.
+    """
+
+    token: str
+    """API token for authentication"""
 
 
 _VERSION_OBJ = pv.Version(VERSION)
@@ -201,20 +229,26 @@ class Client:
     def close(self) -> None:
         return self.__exit__(None, None, None)
 
-    def login(self, credentials: tuple[str, str]) -> None:
-        (auth, _) = self.api_client.auth_api.create_login(
+    def login(self, credentials: Credentials | tuple[str, str]) -> None:
+        if isinstance(credentials, BasicAuthCredentials):
+            credentials = (credentials.user, credentials.password)
+        elif isinstance(credentials, ApiTokenAuthCredentials):
+            self.api_client.default_headers["Authorization"] = "Bearer " + credentials.token
+            return
+        elif not isinstance(credentials, tuple) or len(credentials) != 2:
+            raise TypeError(f"Unsupported credentials type: {type(credentials).__name__}")
+
+        self.api_client.auth_api.create_login(
             models.LoginSerializerExRequest(username=credentials[0], password=credentials[1])
         )
-
         assert "sessionid" in self.api_client.cookies
         assert "csrftoken" in self.api_client.cookies
-        self.api_client.set_default_header("Authorization", "Token " + auth.key)
 
     def has_credentials(self) -> bool:
         return (
             ("sessionid" in self.api_client.cookies)
             or ("csrftoken" in self.api_client.cookies)
-            or bool(self.api_client.get_common_headers().get("Authorization", ""))
+            or bool(self.api_client.default_headers.get("Authorization"))
         )
 
     def logout(self) -> None:


site/content/en/docs/api_sdk/sdk/highlevel-api.md
```diff
@@ -139,6 +138,41 @@ an error can be raised or suppressed (controlled by `config.allow_unsupported_se
...
+```
+
+{{% /tab %}}
+
+{{%tab header="Personal Access Token authentication" %}}
+
+```python
+from cvat_sdk import make_client, ApiTokenCredentials
+
+with make_client('https://app.cvat.ai', credentials=ApiTokenCredentials('token')) as client:
```
SpecLad: Seeing as we're recommending this method, perhaps we should make it more ergonomic by interpreting a plain string as a token?

tests/python/cli/util.py
```diff
@@ -21,15 +22,24 @@
...
 
-    if isinstance(test, unittest.TestCase):
-        # Unittest
-        test.assertEqual(expected_code, main(args), str(args))
-    else:
-        # Pytest case
-        assert expected_code == main(args)
+    with contextlib.ExitStack() as es:
+        if env:
+            es.enter_context(process_env_changed(env))
```
SpecLad: It doesn't seem like this functionality needs to be inside this function - if a test needs to change the environment, it can just do that before calling `run_cli`.

Also, the `process_env_changed` decorator seems like overkill for the purpose you're using it for - it would be much simpler to assign `os.environ['CVAT_AUTH']` at the start of the test and `del` it in a `finally` block.

site/content/en/docs/api_sdk/sdk/highlevel-api.md
SpecLad, line 1: What about the low-level API?

tests/python/cli/util.py
```diff
@@ -130,5 +163,23 @@ def run_cli(
             cmd,
             *args,
             expected_code=expected_code,
+            env=env,
         )
         return self.stdout.getvalue()
+
+
+@contextlib.contextmanager
+def monkeypatch_object(obj: Any, attribute: str, value: Any):
```
SpecLad: Isn't this just a copy of Pytest's `monkeypatch` fixture?

site/content/en/docs/api_sdk/sdk/highlevel-api.md
```diff
@@ -139,6 +138,41 @@ an error can be raised or suppressed (controlled by `config.allow_unsupported_se
 If the error is suppressed, some SDK functions may not work as expected with this server.
 By default, a warning is raised and the error is suppressed.
 
+### Authentication
+
+High-level SDK supports 2 authentication options:
+- Basic authentication, with a username and a password
```
SpecLad: I don't think the HL SDK actually uses basic authentication at any point, does it? If you give it a password, it creates a session and uses session authentication.

cvat-sdk/cvat_sdk/core/client.py
```diff
@@ -200,7 +228,15 @@ def __exit__(self, exc_type, exc_value, traceback) -> None:
     def close(self) -> None:
         return self.__exit__(None, None, None)
 
-    def login(self, credentials: tuple[str, str]) -> None:
+    def login(self, credentials: Credentials | tuple[str, str]) -> None:
```
SpecLad, line 232: I think you have to handle the situation where a user calls `login` with credentials of one type, while the client already has credentials of another type. Seems like currently both sets of credentials will be used at the same time, which could lead to confusing results.

cvat-cli/src/cvat_cli/_internal/common.py
```diff
@@ -27,20 +34,34 @@ class CriticalError(Exception):
...
+    it will try to read the CVAT_AUTH environment variable for a Personal Access Token (PAT).
     The callable will prompt the user for the password if none was initially supplied.
     """
+
+    if not s:
+        s = os.getenv(CVAT_AUTH_ENV_VAR)
+        if s is not None:
+            return lambda _: ApiTokenCredentials(s)
+        else:
+            s = getpass.getuser()
```
SpecLad: IMO, instead of putting all of this into `get_auth_factory`, it would be better to have a function like this:

```python
def default_auth_factory(url):
    token = os.getenv(CVAT_AUTH_ENV_VAR)
    if token is not None:
        return ApiTokenCredentials(s)
    return ask_for_password(url, getpass.getuser())
```

and have this function be the default for `--auth`.

cvat-cli/src/cvat_cli/_internal/common.py
```diff
@@ -27,20 +34,34 @@ class CriticalError(Exception):
     pass
 
 
-def get_auth_factory(s: str) -> Callable[[str], tuple[str, str]]:
+CVAT_AUTH_ENV_VAR = "CVAT_AUTH"
```
SpecLad: Since this variable can only contain a token, why not call it `CVAT_TOKEN`?

site/content/en/docs/api_sdk/sdk/highlevel-api.md
```diff
@@ -139,6 +138,41 @@ an error can be raised or suppressed (controlled by `config.allow_unsupported_se
...
+- Personal Access Token authentication, with a token value
+
+Basic authentication doesn't require a special configuration, but for better security it's
+recommended to use a Personal Access Token (PAT) instead, if possible.
```
SpecLad, line 154: Probably somewhere around here we need to mention that PATs are available in the Enterprise and Cloud versions.

site/content/en/docs/api_sdk/sdk/highlevel-api.md
zhiltsov-max: I think we can introduce other credential variants later, e.g. just the `username` (a saved password is used in this case). Maybe this idea can be implemented in a new parameter like `api_token=`?

tests/python/cli/util.py
zhiltsov-max: Basically, I could pull it up into the test, but technically, the `run_cli` function is supposed to hide implementation details of the launch, including the fact that there is no separate process is used.

site/content/en/docs/api_sdk/sdk/highlevel-api.md
zhiltsov-max: It uses session authentication under the hood indeed, but the important part is that the user provides username and password. You're not configuring session tokens here.

cvat-sdk/cvat_sdk/core/client.py
zhiltsov-max: There are already 2 methods to check authentication and to logout, do you think there should be an extra protection in login?

site/content/en/docs/api_sdk/sdk/highlevel-api.md
zhiltsov-max: Yes, there should be a separate PR with documentation for this feature

cvat-cli/src/cvat_cli/_internal/common.py
zhiltsov-max: No issues with this, maybe `CVAT_API_TOKEN` will be a better choice then.

site/content/en/docs/api_sdk/sdk/highlevel-api.md
zhiltsov-max: I guess we can add a custom auth scheme directly into the template, because we'd need to put it into the server schema otherwise.

cvat-sdk/cvat_sdk/core/client.py
SpecLad: Yeah, I think so. We probably should just call `logout` at the beginning, so that any existing credentials get purged.

site/content/en/docs/api_sdk/sdk/highlevel-api.md
SpecLad: Yeah, but it's still not _basic_ authentication. I'd say just replace the word "Basic" with "Password", and it'll be fine.

site/content/en/docs/api_sdk/sdk/highlevel-api.md
SpecLad: > a saved password is used in this case

I'd rather discourage the use of passwords, not make it easier! :-P

> Maybe this idea can be implemented in a new parameter like `api_token=`?

That would work (although I would prefer `access_token`, seeing as this is Personal _Access Token_ authentication).

tests/python/cli/util.py
SpecLad: I think the `os.environ` method would work even if a separate process _was_ used, so that shouldn't be an issue.

site/content/en/docs/api_sdk/sdk/highlevel-api.md
SpecLad: Or maybe just `token`. I mean, how many types of tokens could we possibly have? :man_shrugging:

tests/python/cli/util.py
zhiltsov-max: It would fit, thanks

cvat-sdk/cvat_sdk/core/client.py
zhiltsov-max: Logout is an API call, it's going to be a different auth cleanup function then.

site/content/en/docs/api_sdk/sdk/highlevel-api.md
zhiltsov-max: I don't know. GitHub has 6 official ones, for example: https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/about-authentication-to-github#githubs-token-formats. They are all API access tokens, so I'd choose either `api_token` or `access_token`.

site/content/en/docs/api_sdk/sdk/highlevel-api.md
zhiltsov-max: When you specify a username and a password, it's basic authentication, isn't it?

SpecLad: I don't think so. To me, basic authentication is specifically what is described in [RFC 7617](https://datatracker.ietf.org/doc/html/rfc7617).

cvat-sdk/cvat_sdk/core/client.py
SpecLad: > Logout is an API call

I don't understand, what's the problem with this? If you're logged in, and you want to log in under different credentials, doesn't it make sense to log out first?

site/content/en/docs/api_sdk/sdk/highlevel-api.md
SpecLad: > They are all API access tokens

That's my point exactly; even if internally there are multiple types, from a client point of view they can be used interchangeably.

> so I'd choose either `api_token` or `access_token`.

Among those two options, my vote is for `access_token`. But the most important thing is to choose one terminology and stick with it everywhere.

zhiltsov-max: Updated

cvat-sdk/cvat_sdk/core/client.py
```diff
@@ -200,27 +229,44 @@ def __exit__(self, exc_type, exc_value, traceback) -> None:
...
+
+    def _clear_credentials(self):
+        self.api_client.cookies.pop("sessionid", None)
+        self.api_client.cookies.pop("csrftoken", None)
+        self.api_client.configuration.api_key.pop("csrfHeaderAuth", None)
+        self.api_client.configuration.access_token = None
 
     def logout(self) -> None:
         if self.has_credentials():
             self.api_client.auth_api.create_logout()
```
Eldies, line 271: as far as I can see, `create_login` does not happen if an access token is used. So, is `create_logout` necessary in this case?

changelog.d/20250623_185705_mzhiltso_sdk_cli_pat_support.md
```diff
@@ -0,0 +1,7 @@
+### Added
+
+- \[CLI\] `CVAT_ACCESS_TOKEN` env. variable can now be used for authentication with an API token
```
SpecLad:
```suggestion
- \[CLI\] `CVAT_ACCESS_TOKEN` environment variable can now be used for authentication with an API token
  (<https://github.com/cvat-ai/cvat/pull/9563>)
```

changelog.d/20250623_185705_mzhiltso_sdk_cli_pat_support.md
```diff
@@ -0,0 +1,7 @@
...
+  \[SDK\] `Client.login()` and `make_client()` can now be called with an API token
```
SpecLad:
```suggestion
- \[SDK\] `Client.login()` and `make_client()` can now be called with an API token
```

changelog.d/20250623_185705_mzhiltso_sdk_cli_pat_support.md
```diff
@@ -0,0 +1,7 @@
...
+  (<https://github.com/cvat-ai/cvat/pull/9563>)
+  \[SDK\] `make_client()` can now be called with a server URL that contains a port component
```
SpecLad:
```suggestion
- \[SDK\] `make_client()` can now be called with a server URL that contains a port component
```

cvat-cli/src/cvat_cli/_internal/common.py
```diff
@@ -27,20 +34,40 @@ class CriticalError(Exception):
     pass
 
 
-def get_auth_factory(s: str) -> Callable[[str], tuple[str, str]]:
+CVAT_ACCESS_TOKEN_ENV_VAR = "CVAT_ACCESS_TOKEN"  # nosec - a variable name declaration
```
SpecLad, line 37: What do you think about moving support for this variable to the SDK? Seems like it would be useful for custom scripts if the `Client` class would automatically fetch the token from the environment.

cvat-cli/src/cvat_cli/_internal/common.py
```diff
@@ -54,11 +81,18 @@ def configure_common_arguments(parser: argparse.ArgumentParser) -> None:
...
-                (default user: %(default)s).""",
+        metavar="USER[:PASS]",
+        default=default_auth_factory(),
+        help=textwrap.dedent(
+            """\
+            User and password to use for authentication;
+            defaults to the current user and supports the PASS
+            environment variable or password prompt.
+            A Personal Access Token (PAT) can be generated on the server
+            and specified in the {} environment variable.
```
SpecLad:
```suggestion
            and specified in the {} environment variable instead.
```

cvat-cli/src/cvat_cli/_internal/common.py
zhiltsov-max: It looks like a good feature in the scope of the Persistent Login functionality. It's going to be addressed via a user profile file though. I suppose, environment dependencies in SDK are going to be introduced in that scope - making client from a local profile, automatic authentication for a given server URL, etc.

SpecLad: Okay, let's leave this for later then.

cvat-sdk/cvat_sdk/core/client.py
```diff
@@ -200,27 +229,48 @@ def __exit__(self, exc_type, exc_value, traceback) -> None:
...
+        if self.has_credentials():
+            self.logout()
+
+        if isinstance(credentials, BasicAuthCredentials):
+            credentials = (credentials.user, credentials.password)
+        elif isinstance(credentials, AccessTokenCredentials):
+            self.api_client.configuration.access_token = credentials.token
+            return
+        elif not isinstance(credentials, tuple) or len(credentials) != 2:
+            raise TypeError(f"Unsupported credentials type: {type(credentials).__name__}")
```
SpecLad: If `credentials` is a tuple of length other than 2, this will say "Unsupported credentials type: tuple", which seems misleading.

cvat-sdk/cvat_sdk/core/client.py
```diff
@@ -354,13 +404,51 @@ def make_endpoint_url(
...
+    if credentials is not None and access_token is not None:
+        raise ValueError(
+            "'credentials' and 'access_token' cannot be used together. Please use only one."
+        )
+
     url = host.rstrip("/")
+
     if port:
+        parsed_url = urlsplit(("https://" if "://" not in url else "") + host)
+        if port and parsed_url.port:
```
SpecLad:
```suggestion
        if parsed_url.port:
```

cvat-sdk/cvat_sdk/core/uploading.py
```diff
@@ -28,14 +28,23 @@
 TUS_CHUNK_SIZE = 10 * 2**20
 
 
+def _make_request_headers(api_client: ApiClient) -> dict[str, str]:
```
SpecLad: This is the same function as in `downloading.py` - could you factor it out?

cvat-sdk/cvat_sdk/core/client.py
```diff
@@ -354,13 +404,51 @@ def make_endpoint_url(
 
 
 def make_client(
-    host: str, *, port: Optional[int] = None, credentials: Optional[tuple[str, str]] = None
+    host: str,
+    *,
+    port: Optional[int] = None,
+    credentials: Union[Credentials, tuple[str, str], None] = None,
+    access_token: Optional[str] = None,
```
SpecLad, line 411: How about interpreting `credentials="..."` as an access token? Then there's no need to add another parameter, and the signatures of `make_client` and `Client.login` can be consistent with each other.

cvat-sdk/gen/templates/openapi-generator/configuration.mustache
```diff
@@ -97,6 +97,13 @@ class Configuration:
         }
     )
 
+    Alternatively, you can authenticate with an API access token in the following way:
+
+    conf = {{{packageName}}}.Configuration(
+        ...
+        access_token= <access token value>,
```
SpecLad:
```suggestion
        access_token=<access token value>,
```

site/content/en/docs/api_sdk/sdk/highlevel-api.md
```diff
@@ -88,10 +88,9 @@ It is the starting point for using CVAT SDK.
 A `Client` instance allows you to:
 - configure connection options with the `Config` class
 - check server API compatibility with the current SDK version
-- deduce server connection scheme (`https` or `http`) automatically
 - manage user session with the `login()`, `logout()` and other methods
-- obtain Repository objects with the `users`, `tasks`, `jobs` and other members
-- reach to lower-level APIs with the corresponding members
+- obtain high-level server object wrappers with the `users`, `tasks`, `jobs` and other members
+- easily reach lower-level APIs to send raw requests, typically via the `api` member of the object
```
SpecLad:
```suggestion
- reach lower-level APIs to send raw requests, typically via the `api` member of the object
```

I don't think it makes sense to single out this feature as easy. The other ones aren't really supposed to be hard either. :slightly_smiling_face:

site/content/en/docs/api_sdk/sdk/lowlevel-api.md
```diff
@@ -134,15 +134,23 @@ Typically, the first thing you do with `ApiClient` is log in.
 
 CVAT supports 3 authentication options:
 - Basic authentication, with a username and a password
+- Personal Access Token (PAT) authentication, with an access token value
```
SpecLad, line 136: I would suggest to put it first, since it's what most (all?) users should be using.

site/content/en/docs/api_sdk/sdk/lowlevel-api.md
```diff
@@ -134,15 +134,23 @@ Typically, the first thing you do with `ApiClient` is log in.
...
 - Session authentication, with a session ID and a CSRF token
 - Token authentication, with an API key (deprecated)
```
SpecLad: Can we better differentiate old tokens from the new ones? I would suggest to rename these to "legacy tokens", or something along those lines.

tests/python/shared/utils/dump_objects.py
```diff
@@ -49,19 +50,36 @@ def clean_list_response(data: dict[str, Any]) -> dict[str, Any]:
...
     parser = argparse.ArgumentParser()
     parser.add_argument("--assets-dir", type=Path, default=ASSETS_DIR)
+    parser.add_argument(
+        "--asset-path",
+        dest="asset_urls",
+        default=[],
+        action="append",
+        type=_parse_asset_url,
+        help="Repeatable, an override for the default inferred path for an asset. "
+        "Format: '<asset filename without extension>:<url path after api/>'",
```
SpecLad, line 68: I would suggest using `nargs=2` instead of manual parsing.

tests/python/shared/utils/dump_objects.py
```diff
@@ -49,19 +50,36 @@ def clean_list_response(data: dict[str, Any]) -> dict[str, Any]:
...
+    asset_filename, url_path = s.lower().rsplit(":", maxsplit=1)
+    return asset_filename, url_path
+
+
 def main():
     parser = argparse.ArgumentParser()
     parser.add_argument("--assets-dir", type=Path, default=ASSETS_DIR)
+    parser.add_argument(
+        "--asset-path",
+        dest="asset_urls",
```
SpecLad: These are inconsistent - either the argument should be `--asset-url`, or the variable should be `asset_paths`.

cvat-sdk/cvat_sdk/core/client.py
```diff
@@ -58,6 +60,33 @@ class Config:
     """Directory in which to store cached server data"""
 
 
+class Credentials(metaclass=ABCMeta):
+    pass
+
+
+@attrs.define
+class BasicAuthCredentials(Credentials):
```
SpecLad:
```suggestion
class PasswordCredentials(Credentials):
```

This would be consistent with the documentation.

That said, maybe we don't need this class at all? It doesn't really change much compared to the tuple. Also, I'd like to eventually deprecate password authentication entirely, which would render the class unnecessary.

changelog.d/20250623_185705_mzhiltso_sdk_cli_pat_support.md
```diff
@@ -0,0 +1,8 @@
+### Added
+
+- \[CLI\] `CVAT_ACCESS_TOKEN` environment variable can now be used for authentication with an API token
+  (<https://github.com/cvat-ai/cvat/pull/9563>)
+- \[SDK\] `Client.login()` and `make_client()` can now be called with an API token
+  (<https://github.com/cvat-ai/cvat/pull/9563>)
+- \[SDK\] `make_client()` can now be called with a server URL that contains a port component
```
SpecLad, line 7: Hasn't this been possible before, actually? Looking at the source, `make_client` will just pass the URL (port included) to the `Client` constructor if the `port` argument is unspecified.

It seems that the only thing different is that now `make_client` will raise if `port` is specified while `host` already contains a port.

zhiltsov-max: Even if the previous implementation allowed it, it wasn't supposed by the function interface.

site/content/en/docs/api_sdk/sdk/lowlevel-api.md
zhiltsov-max: In all the places where the old tokens are mentioned, there is a deprecation notice, and all the new ones are either Access tokens or PAT. I think we can even remove the old tokens from the documentation already, because they were deprecated 3 months ago. Otherwise, they are not much different, only the key prefix is changed.

cvat-sdk/cvat_sdk/core/client.py
zhiltsov-max: I don't like this idea, because there are several possible interpretations for just a string. Now it could be token value. Later, it could also be username or token name for a remembered authentication. It's one of the reasons why I didn't do it for CLI args and env vars.

cvat-sdk/cvat_sdk/core/client.py
zhiltsov-max: This class isn't supposed to change anything for users, except adding a name to the tuple. What is the problem with removing this class later, when we deprecate this type of authentication?

tests/python/shared/utils/dump_objects.py
zhiltsov-max: I'm not sure it's a good idea. Have you seen any successful examples of such syntax?

tests/python/shared/utils/dump_objects.py
zhiltsov-max: Updated

site/content/en/docs/api_sdk/sdk/lowlevel-api.md
zhiltsov-max: Ok, updated.

cvat-sdk/cvat_sdk/core/uploading.py
zhiltsov-max: Ok, refactored

cvat-sdk/cvat_sdk/core/client.py
SpecLad: Okay, fair enough; but then how about adding an `access_token` parameter to `Client.login`?

site/content/en/docs/api_sdk/sdk/lowlevel-api.md
SpecLad: > In all the places where the old tokens are mentioned, there is a deprecation notice, and all the new ones are either Access tokens or PAT.

My point is that it's hard to understand what is deprecated, because "token authentication" is (grammatically) a superclass of "personal access token authentication". So it would be helpful to differentiate the old tokens somehow.

> I think we can even remove the old tokens from the documentation already, because they were deprecated 3 months ago.

We shouldn't remove them from the documentation until they are removed from the product.

cvat-sdk/cvat_sdk/core/client.py
SpecLad: Why add it in the first place?

tests/python/shared/utils/dump_objects.py
SpecLad: What do you mean by "successful"? The argparse feature works, so it'll be successful if you use it. :slightly_smiling_face: 

But it's just a suggestion, so I don't mind if you keep the code as-is.

changelog.d/20250623_185705_mzhiltso_sdk_cli_pat_support.md
SpecLad: But it did work, though. So it would be dishonest to claim that it had only been added now.

cvat-sdk/cvat_sdk/core/client.py
zhiltsov-max: I consider `make_client()` as a convenience factory function and the main end-user interface. It means, that only this function should include most of the QoL features. If the user uses the `Client` class directly, there is focus on control instead of simplification - you're possibly going to use the `Config` class, call `login()` separately etc. I'd removed the tuple support from `login()`, if not for backward compatibility.

tests/python/shared/utils/dump_objects.py
zhiltsov-max: By "successful" I mean any popular tools with such kind of parameter syntax, where it's known to perform without significant problems.

cvat-sdk/cvat_sdk/core/client.py
zhiltsov-max: For completeness of the provided API and for future extensibility - we support these credential types and here are the classes that express them. Would you prefer having `TokenCredentials` and a tuple in the API?

changelog.d/20250623_185705_mzhiltso_sdk_cli_pat_support.md
zhiltsov-max: It wasn't officially supported previously, now it is. If it worked, it was either a bug or an undocumented feature.

site/content/en/docs/api_sdk/sdk/lowlevel-api.md
zhiltsov-max: > My point is that it's hard to understand what is deprecated, because "token authentication" is (grammatically) a superclass of "personal access token authentication". So it would be helpful to differentiate the old tokens somehow.

Their similarity is not coincidental, as the only difference is which prefix is used in the header. Ok, I renamed them to "Legacy tokens" in the documentation, another option is "Authorization tokens".

tests/python/shared/utils/dump_objects.py
SpecLad: I dunno, here's autopep8: https://github.com/hhatto/autopep8/blob/4046ad49e25b7fa1db275bf66b1b7d60600ac391/autopep8.py#L3819

I'm pretty sure I've used it myself too. I don't know what problems you're expecting.

cvat-sdk/cvat_sdk/core/client.py
SpecLad: > For completeness of the provided API and for future extensibility - we support these credential types and here are the classes that express them.

How does a dedicated password credentials class help here? All of this is already expressible with the tuple.

I could understand doing this for consistency if these credential types were on equal footing, but we want to discourage the use of password authentication. So why would we add more functionality related to it?

> Would you prefer having `TokenCredentials` and a tuple in the API?

Yeah, pretty much.

zhiltsov-max: I think we (the sdk writers) are going to be the main users of this class. The existing users will continue using tuples where available, an extra class is only likely to encumber them. For us, this allows writing more generic and clear code. Introducing only `TokenCredentials` looks incomplete, because there are 2 officially supported credential types in high-level SDK. Not adding a class to discourage this authentication somewhere in the future, while supporting tuples, also doesn't make much sense. If we ever want to discourage this authentication, we should raise a warning when we get such a value in the API call.

SpecLad: Alright, let's keep it. What do you think about the name, though?

zhiltsov-max: It was renamed it to `PasswordCredentials`, do you want to change anything else?

SpecLad: Oh, my bad. I missed this change.

PR #9567, opened 2025-06-24
Fixed template to confirm email address

bsekachev:
### Motivation and context
From

image

To

image

### Checklist

### License

### DIFF
--- a/cvat/apps/iam/templates/account/email/email_confirmation_message.html
+++ b/cvat/apps/iam/templates/account/email/email_confirmation_message.html
@@ -134,7 +134,7 @@
             <td align="center" valign="bottom">
               <img
                 style="margin: 24px;"
-                src="https://raw.githubusercontent.com/cvat-ai/cvat/develop/cvat/apps/engine/static/logo.png"
+                src="https://raw.githubusercontent.com/cvat-ai/cvat/v2.40.0/cvat/apps/engine/static/logo.png"
                 alt="CVAT Logo"
                 width="128px"
               ></img>
--- a/cvat/apps/organizations/templates/invitation/invitation_message.html
+++ b/cvat/apps/organizations/templates/invitation/invitation_message.html
@@ -135,7 +135,7 @@
             <td align="center" valign="bottom">
               <img
                 style="margin: 24px;"
-                src="https://raw.githubusercontent.com/cvat-ai/cvat/develop/cvat/apps/engine/static/logo.png"
+                src="https://raw.githubusercontent.com/cvat-ai/cvat/v2.40.0/cvat/apps/engine/static/logo.png"
                 alt="CVAT Logo"
                 width="128px"
               ></img>


cvat/apps/iam/templates/account/email/email_confirmation_message.html
```diff
@@ -123,11 +133,10 @@
             <td align="center" valign="bottom">
               <a href="https://www.cvat.ai/" target="_blank" style="display: inline-block;">
                 <img
-                  src="https://github.com/cvat-ai/cvat/blob/develop/site/content/en/images/cvat_poster_with_name.png?raw=true"
+                  src="https://app.cvat.ai/static/logo.svg"
```
SpecLad: It doesn't seem appropriate to link back to `app.cvat.ai` for emails that could be coming from any CVAT instance... do you think you could change it to link to the domain of the current instance?

(Even better would be to attach the image to the email, but I don't know if that's possible in our implementation.)

cvat/apps/iam/templates/account/email/email_confirmation_message.html
SpecLad, line 1: Please add a changelog entry.

cvat/apps/iam/templates/account/email/email_confirmation_message.html
bsekachev: We may use `settings.CVAT_HOST`. But it will work only if the host is publicly available.

bsekachev: > Even better would be to attach the image to the email

Convertation of the logo to base64 should work this way

SpecLad: In fact, since this logo is a static file, you should be able to use the [`static` template tag](https://docs.djangoproject.com/en/5.2/howto/static-files/#configuring-static-files) to reference it.

bsekachev: It only works for HTML templates, but not for emails. 

In email it would be incorrect URL like this:
image

Moreover, context for this template passed not from our code, so
That means that we can't support dynamic logo file, configured in settings (LOGO_FILENAME) without overriding default implementation of `send_confirmation_mail` in `DefaultAccountAdapter`.

bsekachev: It may be just simpler to:

1. Refer to common logo 
2. Remove the logo

SpecLad: > That means that we can't support dynamic logo file, configured in settings (LOGO_FILENAME) without overriding default implementation of `send_confirmation_mail` in `DefaultAccountAdapter`.

I think we only need to override [`get_context_data`](https://codeberg.org/allauth/django-allauth/src/commit/4248e0532da93b26e0fcabb9faa5850f4dd7a62b/allauth/account/views.py#L320).

SpecLad: Oh, never mind that, I mistook the purpose of that function...

Well I don't know then, maybe we should just remove the logo.

cvat/apps/iam/templates/account/email/email_confirmation_message.html
```diff
@@ -121,15 +132,12 @@
...
-              <a href="https://www.cvat.ai/" target="_blank" style="display: inline-block;">
-                <img
-                  src="https://github.com/cvat-ai/cvat/blob/develop/site/content/en/images/cvat_poster_with_name.png?raw=true"
-                  alt="Logo"
-                  border="0"
-                  width="100%"
-                  style="display: block; width: 100%;"
-                >
-              </a>
+              <svg style="margin: 24px;" width="128" height="24" viewBox="0 0 80 15" xmlns="http://www.w3.org/2000/svg">
```
SpecLad: Have you checked if email clients can actually display this? There's some evidence that inline SVG support in various email clients is lacking: https://css-tricks.com/a-guide-on-svg-support-in-email/#svg-email-inline https://www.caniemail.com/features/html-svg/

changelog.d/20250624_130027_sekachev.bs_fixed_confirmation_email.md
```diff
@@ -0,0 +1,4 @@
+### Changed
+
+- Improved email templates for email vericiation and organization invitation
```
SpecLad: Some more detail would be good to see. At least the URL fix should be a separate item (in the "Fixed" category).

Also, vericiation -> verification (or maybe confirmation).

cvat/apps/iam/templates/account/email/email_confirmation_message.html
bsekachev: If some can't - so, they won't see the logo

bsekachev: Alternatively, it still may be encoded as base64 "src" attribute of image tag. This is less readable, but maybe alright

bsekachev: On the other side, it seems, src="base64" is also not well-supported..

SpecLad: Let's go back to inline SVG then, at least that's readable. :man_shrugging:

bsekachev: IDK, maybe it is better to keep static link to raw.githubusercontent.com
The previous was just to develop branch and gradually got broken.

SpecLad: Works for me.

PR #9568, opened 2025-06-24
Actualize api token auth schema

zhiltsov-max:
### Motivation and context

- \[Server API\] Updated API token auth schema - now only the API token is required
- \[Server API\] Updated sessionid + csrftoken auth schema to require only this pair

### Checklist

### License

### DIFF
--- a/cvat-sdk/cvat_sdk/core/client.py
+++ b/cvat-sdk/cvat_sdk/core/client.py
@@ -205,9 +205,6 @@ class Client:
         (auth, _) = self.api_client.auth_api.create_login(
             models.LoginSerializerExRequest(username=credentials[0], password=credentials[1])
         )
-
-        assert "sessionid" in self.api_client.cookies
-        assert "csrftoken" in self.api_client.cookies
         self.api_client.set_default_header("Authorization", "Token " + auth.key)
 
     def has_credentials(self) -> bool:
@@ -222,6 +219,7 @@ class Client:
             self.api_client.auth_api.create_logout()
             self.api_client.cookies.pop("sessionid", None)
             self.api_client.cookies.pop("csrftoken", None)
+            self.api_client.default_headers.pop("X-CSRFToken", None)
             self.api_client.default_headers.pop("Authorization", None)
 
     def wait_for_completion(
--- a/cvat/apps/iam/schema.py
+++ b/cvat/apps/iam/schema.py
@@ -39,22 +39,13 @@ class TokenAuthenticationScheme(TokenScheme):
     priority = 0
     match_subclasses = True
 
-    def get_security_requirement(self, auto_schema):
-        # These schemes must be used together
-        return {"sessionAuth": [], "csrfAuth": [], self.name: []}
-
     def get_security_definition(self, auto_schema):
         schema = super().get_security_definition(auto_schema)
         schema["x-token-prefix"] = self.target.keyword
         schema["description"] = textwrap.dedent(
-            f"""
-            To authenticate using a token (or API key), you need to have 3 components in a request:
-            - the 'sessionid' cookie
-            - the 'csrftoken' cookie or 'X-CSRFTOKEN' header
-            - the 'Authentication' header with the '{self.target.keyword} ' prefix
-
+            f"""\
             You can obtain an API key (the token) from the server response on
-            the basic auth request.
+            the basic auth request in the /api/auth/login/ endpoint.
         """
         )
         return schema
@@ -68,10 +59,6 @@ class CookieAuthenticationScheme(SessionScheme):
 
     name = ["sessionAuth", "csrfAuth"]
     priority = 0
-
-    def get_security_requirement(self, auto_schema):
-        # These schemes cannot be used separately
-        return None
 
     def get_security_definition(self, auto_schema):
         sessionid_schema = super().get_security_definition(auto_schema)
--- a/cvat/schema.yml
+++ b/cvat/schema.yml
@@ -25,10 +25,10 @@ paths:
               $ref: '#/components/schemas/AssetWriteRequest'
         required: true
       security:
+      - tokenAuth: []
+      - signatureAuth: []
       - sessionAuth: []
         csrfAuth: []
-        tokenAuth: []
-      - signatureAuth: []
       - basicAuth: []
       responses:
         '201':
@@ -52,10 +52,10 @@ paths:
       tags:
       - assets
       security:
+      - tokenAuth: []
+      - signatureAuth: []
       - sessionAuth: []
         csrfAuth: []
-        tokenAuth: []
-      - signatureAuth: []
       - basicAuth: []
       responses:
         '200':
@@ -74,10 +74,10 @@ paths:
       tags:
       - assets
       security:
+      - tokenAuth: []
+      - signatureAuth: []
       - sessionAuth: []
         csrfAuth: []
-        tokenAuth: []
-      - signatureAuth: []
       - basicAuth: []
       responses:
         '204':
@@ -104,10 +104,10 @@ paths:
               $ref: '#/components/schemas/LoginSerializerExRequest'
         required: true
       security:
+      - tokenAuth: []
+      - signatureAuth: []
       - sessionAuth: []
         csrfAuth: []
-        tokenAuth: []
-      - signatureAuth: []
       - basicAuth: []
       - {}
       responses:
@@ -128,10 +128,10 @@ paths:
       tags:
       - auth
       security:
+      - tokenAuth: []
+      - signatureAuth: []
       - sessionAuth: []
         csrfAuth: []
-        tokenAuth: []
-      - signatureAuth: []
       - basicAuth: []
       - {}
       responses:
@@ -158,10 +158,10 @@ paths:
               $ref: '#/components/schemas/PasswordChangeRequest'
         required: true
       security:
+      - tokenAuth: []
+      - signatureAuth: []
       - sessionAuth: []
         csrfAuth: []
-        tokenAuth: []
-      - signatureAuth: []
       - basicAuth: []
       responses:
         '200':
@@ -187,10 +187,10 @@ paths:
               $ref: '#/components/schemas/PasswordResetSerializerExRequest'
         required: true
       security:
+      - tokenAuth: []
+      - signatureAuth: []
       - sessionAuth: []
         csrfAuth: []
-        tokenAuth: []
-      - signatureAuth: []
       - basicAuth: []
       - {}
       responses:
@@ -219,10 +219,10 @@ paths:
               $ref: '#/components/schemas/PasswordResetConfirmRequest'
         required: true
       security:
+      - tokenAuth: []
+      - signatureAuth: []
       - sessionAuth: []
         csrfAuth: []
-        tokenAuth: []
-      - signatureAuth: []
       - basicAuth: []
       - {}
       responses:
@@ -244,10 +244,10 @@ paths:
               $ref: '#/components/schemas/RegisterSerializerExRequest'
         required: true
       security:
+      - tokenAuth: []
+      - signatureAuth: []
       - sessionAuth: []
         csrfAuth: []
-        tokenAuth: []
-      - signatureAuth: []
       - basicAuth: []
       - {}
       responses:
@@ -282,10 +282,10 @@ paths:
               $ref: '#/components/schemas/SigningRequest'
         required: true
       security:
+      - tokenAuth: []
+      - signatureAuth: []
       - sessionAuth: []
         csrfAuth: []
-        tokenAuth: []
-      - signatureAuth: []
       - basicAuth: []
       responses:
         '200':
@@ -396,10 +396,10 @@ paths:
       tags:
       - cloudstorages
       security:
+      - tokenAuth: []
+      - signatureAuth: []
       - sessionAuth: []
         csrfAuth: []
-        tokenAuth: []
-      - signatureAuth: []
       - basicAuth: []
       responses:
         '200':
@@ -483,10 +483,10 @@ paths:
               $ref: '#/components/schemas/CloudStorageWriteRequest'
         required: true
       security:
+      - tokenAuth: []
+      - signatureAuth: []
       - sessionAuth: []
         csrfAuth: []
-        tokenAuth: []
-      - signatureAuth: []
       - basicAuth: []
       responses:
         '201':
@@ -509,10 +509,10 @@ paths:
       tags:
       - cloudstorages
       security:
+      - tokenAuth: []
+      - signatureAuth: []
       - sessionAuth: []
         csrfAuth: []
-        tokenAuth: []
-      - signatureAuth: []
       - basicAuth: []
       responses:
         '200':
@@ -586,10 +586,10 @@ paths:
             schema:
               $ref: '#/components/schemas/PatchedCloudStorageWriteRequest'
       security:
+      - tokenAuth: []
+      - signatureAuth: []
       - sessionAuth: []
         csrfAuth: []
-        tokenAuth: []
-      - signatureAuth: []
       - basicAuth: []
       responses:
         '200':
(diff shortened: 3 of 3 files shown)


cvat/apps/iam/schema.py
```diff
@@ -39,22 +39,13 @@ class TokenAuthenticationScheme(TokenScheme):
...
-            f"""
-            To authenticate using a token (or API key), you need to have 3 components in a request:
-            - the 'sessionid' cookie
-            - the 'csrftoken' cookie or 'X-CSRFTOKEN' header
-            - the 'Authentication' header with the '{self.target.keyword} ' prefix
-
+            f"""\
             You can obtain an API key (the token) from the server response on
-            the basic auth request.
+            the basic auth request in the /api/auth/login/ endpoint.
```
SpecLad:
```suggestion
            the /api/auth/login/ endpoint.
```

This endpoint does not accept basic authentication.

cvat-sdk/cvat_sdk/core/client.py
```diff
@@ -205,9 +205,6 @@ def login(self, credentials: tuple[str, str]) -> None:
         (auth, _) = self.api_client.auth_api.create_login(
             models.LoginSerializerExRequest(username=credentials[0], password=credentials[1])
         )
-
-        assert "sessionid" in self.api_client.cookies
-        assert "csrftoken" in self.api_client.cookies
```
SpecLad, line 210: Why do you want to remove this?

cvat-sdk/cvat_sdk/core/client.py
```diff
@@ -222,6 +219,7 @@ def logout(self) -> None:
             self.api_client.auth_api.create_logout()
             self.api_client.cookies.pop("sessionid", None)
             self.api_client.cookies.pop("csrftoken", None)
+            self.api_client.default_headers.pop("X-CSRFToken", None)
```
SpecLad, line 223: I don't see anything that would add this header, so why remove it?

cvat-sdk/cvat_sdk/core/client.py
zhiltsov-max: These fields are not required by the authentication scheme used in SDK, so there is no reason to check for them here. There are tests for auth endpoints that do still check that the basic auth login endpoint returns such cookies.

cvat-sdk/cvat_sdk/core/client.py
zhiltsov-max: This function cleans up different authentication methods that could be used in the client, both manually crafted and builtin ones, so that the next possible login call was clean.

cvat/apps/iam/schema.py
zhiltsov-max: Do you mean it doesn't use the `basicAuth` schema defined nearby? You still have to use basic auth credentials to get a token via this endpoint.

SpecLad: > Do you mean it doesn't use the `basicAuth` schema defined nearby?

Yes. It takes credentials in the request body instead.

> You still have to use basic auth credentials to get a token via this endpoint.

Sure, the credentials are the same, but the protocol is different.

SpecLad:
There are a few other places that should be updated:

* The "Authentication" section in `lowlevel-api.md`.
* `TestTokenAuth`, which should probably be split into `TestTokenAuth` and `TestSessionAuth`, testing each method separately.

cvat-sdk/gen/templates/openapi-generator/api_client.mustache
```diff
@@ -678,19 +681,26 @@ class ApiClient(object):
...
         for auth in auth_settings:
             auth_setting = self.configuration.auth_settings().get(auth)
             if auth_setting:
-                self._apply_auth_params(
-                    headers, queries, resource_path, method, body, auth_setting)
+                self._apply_auth_params(headers, queries, resource_path, method, body, auth_setting)
+
+                # A custom extension for CSRF HTTPS headers - add browser headers automatically
+                if auth in ('csrfAuth', 'csrfHeaderAuth'):
+                    headers['Host'] = parse_url(request_host).netloc
```
SpecLad: Could you explain why this is needed? `Host` is not a browser header, it's mandatory in HTTP/1.1 - and I'm pretty sure urllib3 already adds it to every request.

cvat-sdk/gen/templates/openapi-generator/configuration.mustache
```diff
@@ -86,21 +86,29 @@ class Configuration:
 
     API Key Authentication Example.
 
-    You can authenticate with API token after doing the basic auth the following way:
+    You can authenticate with a token after logging in the following way:
```
SpecLad: Probably worth mentioning the deprecation here.

cvat-sdk/gen/templates/openapi-generator/api_client.mustache
```diff
@@ -678,19 +681,26 @@ class ApiClient(object):
...
 
         for auth in auth_settings:
             auth_setting = self.configuration.auth_settings().get(auth)
             if auth_setting:
-                self._apply_auth_params(
-                    headers, queries, resource_path, method, body, auth_setting)
+                self._apply_auth_params(headers, queries, resource_path, method, body, auth_setting)
+
+                # A custom extension for CSRF HTTPS headers - add browser headers automatically
+                if auth in ('csrfAuth', 'csrfHeaderAuth'):
```
SpecLad: If you're going to add custom logic here, then do we even need `csrfHeaderAuth`? You might as well automatically set the `X-CSRFToken` header when `auth == 'csrfAuth'`.

cvat-sdk/gen/templates/openapi-generator/api_client.mustache
```diff
@@ -678,19 +681,26 @@ class ApiClient(object):
...
+                    headers['Host'] = parse_url(request_host).netloc
+                    headers['Origin'] = request_host
 
     def _apply_auth_params(self, headers, queries, resource_path, method, body, auth_setting):
         if auth_setting['in'] == 'cookie':
-            headers['Cookie'] = auth_setting['key'] + "=" + auth_setting['value']
+            headers['Cookie'] = "; ".join(
+                (
+                    [headers.get("Cookie", "")] if headers.get("Cookie", "") else []
```
SpecLad:
```suggestion
                    [headers["Cookie"]] if headers.get("Cookie", "") else []
```

cvat-sdk/gen/templates/openapi-generator/api_client.mustache
```diff
@@ -678,19 +681,26 @@ class ApiClient(object):
...
+
+                # A custom extension for CSRF HTTPS headers - add browser headers automatically
+                if auth in ('csrfAuth', 'csrfHeaderAuth'):
+                    headers['Host'] = parse_url(request_host).netloc
+                    headers['Origin'] = request_host
 
     def _apply_auth_params(self, headers, queries, resource_path, method, body, auth_setting):
         if auth_setting['in'] == 'cookie':
-            headers['Cookie'] = auth_setting['key'] + "=" + auth_setting['value']
+            headers['Cookie'] = "; ".join(
```
SpecLad: This is just my opinion, so feel free to ignore this, but I would suggest this instead:

```
if headers.setdefault("Cookie", ""):
    headers["Cookie"] += "; "
headers["Cookie"] += auth_setting['key'] + "=" + auth_setting['value']
```

All the lists are IMO making this harder to understand.

cvat-sdk/gen/templates/openapi-generator/configuration.mustache
```diff
@@ -86,21 +86,29 @@ class Configuration:
...
             "tokenAuth": <auth key value>,
         }
     )
 
-    You need to specify all the 3 keys for this kind of auth.
-
     If your custom server uses another token prefix, use the 'api_key_prefix' parameter.
 
+
+    You can authenticate with session tokens after logging in the following way:
```
SpecLad: This should probably be listed first, so that we're starting with a non-deprecated method.

site/content/en/docs/api_sdk/sdk/lowlevel-api.md
```diff
@@ -132,47 +132,106 @@ Typically, the first thing you do with `ApiClient` is log in.
 
 ## Authentication
 
-CVAT supports 2 authentication options:
-- basic auth, with your username and password
-- token auth, with your API key
+CVAT supports 3 authentication options:
+- Basic authentication, with a username and a password
+- Token authentication, with an API access key
+- Session authentication, with a session id and a CSRF token
```
SpecLad:
```suggestion
- Session authentication, with a session ID and a CSRF token
```

site/content/en/docs/api_sdk/sdk/lowlevel-api.md
```diff
@@ -132,47 +132,106 @@ Typically, the first thing you do with `ApiClient` is log in.
 
 ## Authentication
 
-CVAT supports 2 authentication options:
-- basic auth, with your username and password
-- token auth, with your API key
+CVAT supports 3 authentication options:
+- Basic authentication, with a username and a password
+- Token authentication, with an API access key
```
SpecLad:
```suggestion
- Token authentication, with an API key
```

This is the only place where I see "API access key" being used, everywhere else is just "API key".

site/content/en/docs/api_sdk/sdk/lowlevel-api.md
```diff
@@ -132,47 +132,106 @@ Typically, the first thing you do with `ApiClient` is log in.
...
+- Session authentication, with a session id and a CSRF token
 
-Token auth requires a token, which can be obtained after performing the basic auth.
+Token authentication requires an API key, which can be obtained after logging in
```
SpecLad, line 140: The deprecation should be mentioned here.

site/content/en/docs/api_sdk/sdk/lowlevel-api.md
```diff
@@ -132,47 +132,106 @@ Typically, the first thing you do with `ApiClient` is log in.
...
+the `/api/auth/login` endpoint using the basic authentication credentials.
 
-The low-level API supports 2 ways of authentication.
-You can specify authentication parameters in the `Configuration` object:
+Session authentication requires a session id and a CSRF token, which can be obtained after
```
SpecLad:
```suggestion
Session authentication requires a session ID and a CSRF token, which can be obtained after
```

site/content/en/docs/api_sdk/sdk/lowlevel-api.md
```diff
@@ -132,47 +132,106 @@ Typically, the first thing you do with `ApiClient` is log in.
...
+logging in the `/api/auth/login` endpoint using the basic authentication credentials.
```
SpecLad:
```suggestion
logging in via the `/api/auth/login` endpoint using the basic authentication credentials.
```

cvat-sdk/gen/templates/openapi-generator/api_client.mustache
zhiltsov-max: I rechecked the reason. urllib3 indeed appends the `Host` header when it's not specified. It's [being constructed](https://github.com/python/cpython/blob/f297a2292cd3c3596f21ca5914310f1f8d5d8750/Lib/http/client.py#L1181) with the default HTTP and HTTPS ports truncated. In the Helm tests, where the problem was discovered, we are [using the server host with the default port](https://github.com/cvat-ai/cvat/blob/7485d1a03277d19e7e1e88e85141bd547af0c9fd/.github/workflows/main.yml#L485) specified explicitly. SDK sends the `Origin` header with the port included in the value and `Host` with the truncated port. Django checks them to match in the CSRF check, and they do not match:
- [_get_raw_host](https://github.com/django/django/blob/8293b0f52d086410bb480b7d6a79e614c8184761/django/http/request.py#L112) returns the passed `Host` header unchanged.
- [get_host()](https://github.com/django/django/blob/8293b0f52d086410bb480b7d6a79e614c8184761/django/http/request.py#L141) checks the host without the port, but then returns the passed `Host` header unchanged
- [csrf middleware](https://github.com/django/django/blob/8293b0f52d086410bb480b7d6a79e614c8184761/django/middleware/csrf.py#L283-L288) prepends the scheme to the host and checks the values for exact match

I think we actually have 2 options here, supposing it's an SDK-only problem:
- clean up the default port in the passed `Origin` header on the SDK side
- modify `Host` to include the port, overriding the autogenerated value

I think now that the first solution might be better.

cvat-sdk/gen/templates/openapi-generator/api_client.mustache
zhiltsov-max: The idea was that both these headers are mentioned in schema and they require a specific user-provided value. `Origin` and `Host` can and should be generated automatically based on the request being sent. I agree that we can add a custom code to automatically copy the passed `csrfAuth` into `csrfHeaderAuth` on the SDK side.

cvat-sdk/gen/templates/openapi-generator/api_client.mustache
SpecLad: I agree, let's do that. The `Origin` spec actually recommends omitting default ports: https://datatracker.ietf.org/doc/html/rfc6454#section-6.2.

site/content/en/docs/api_sdk/sdk/lowlevel-api.md
zhiltsov-max: Added deprecation notes in all the places

cvat-sdk/gen/templates/openapi-generator/api_client.mustache
zhiltsov-max: Added an automatic `X-CSRFToken` header

cvat-sdk/gen/templates/openapi-generator/api_client.mustache
zhiltsov-max: Ok, implemented

cvat/apps/iam/schema.py
```diff
@@ -39,49 +39,54 @@ class TokenAuthenticationScheme(TokenScheme):
...
-
     def get_security_definition(self, auto_schema):
         sessionid_schema = super().get_security_definition(auto_schema)
-        csrftoken_schema = {
+
+        csrf_token_description = textwrap.dedent(
+            """\
+            A CSRF protection token. Can be received in the 'csrftoken' cookie in the
+            server response on the /api/auth/login endpoint.
+            For HTTPS requests, the 'Origin' and 'Host' or 'Referer' and 'Host' headers
```
SpecLad:
```suggestion
            The 'Origin' header
```
Let's not complicate this. `Origin` is sufficient in all cases, so let's just pretend it's required.

cvat-sdk/gen/templates/openapi-generator/api_client.mustache
```diff
@@ -659,8 +661,19 @@ class ApiClient(object):
         else:
             return content_types[0]
 
+    def build_origin_header(self, request_host: typing.Optional[str] = None) -> str:
+        if request_host is None:
+            request_host = self.configuration.host
+
+        parsed_url = parse_url(request_host)
+        origin = f"{parsed_url.scheme}://{parsed_url.host}"
+        if parsed_url.port is not None and parsed_url.port not in (80, 443):
```
SpecLad:
```suggestion
        if parsed_url.port is not None and not (
            (parsed_url.scheme == 'http' and parsed_url.port == 80)
            or (parsed_url.scheme == 'https' and parsed_url.port == 443)
        ):
```

We should only omit the port if it matches the default port for the current scheme.

cvat/apps/iam/schema.py
zhiltsov-max: Ok

cvat-sdk/gen/templates/openapi-generator/api_client.mustache
zhiltsov-max: Updated

SpecLad (approved):

PR #9571, opened 2025-06-24
Fix zip handling of osx hidden dirs

kierstenarnold-fw:
### Motivation and context

It looks like zip files sometimes (always?) get categorized as 'archive' and are handled by the ArchiveReader rather than the ZipReader. Add osx hidden dir checking to ArchiveReader as well.

This is done through ArchiveReader's parent class DirectoryReader for ease of implementation. I can't think of a reason why we wouldn't want to ignore these dirs; so at best we catch junk directories that could show up in other archive types and at worst there's a small overhead of string comparisons.

### How has this been tested?

I didn't see any existing tests of ZipReader's filtering logic and I didn't see an easy way to add a test for this. I tested these changes locally in a running server and verified an osx compressed archive gets processed to all expected images. Also verified there's no regressions in existing tests.

### Checklist

### License

### DIFF
--- a/cvat/apps/engine/media_extractors.py
+++ b/cvat/apps/engine/media_extractors.py
@@ -387,7 +387,7 @@ class DirectoryReader(ImageListReader):
         for source in source_path:
             for root, _, files in os.walk(source):
                 paths = [os.path.join(root, f) for f in files]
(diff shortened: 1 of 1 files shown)


joelhess:
@nmanovic What can we do to get this merged?

PR #9579, opened 2025-06-26
Add support for agent-based tracking functions

SpecLad:
### How has this been tested?

This functionality is agent-only, so tests will be added to the private repository.

### Checklist

### License


cvat-cli/src/cvat_cli/_internal/agent.py
```diff
@@ -85,22 +90,127 @@ def result(self, future: concurrent.futures.Future):
...
 
+def _worker_job_init_tracking(
+    context: _TrackingFunctionContextImpl,
+    task_id: int,
+    image: PIL.Image.Image,
+    shapes: list[cvataa.TrackableShape],
+) -> list[str]:
+    _tracking_states.prune()
+
+    pp_image = _current_function.preprocess_image(context, image)
```
zhiltsov-max: Image preprocessing doesn't seem to be an essential step for every tracker or agent, I think it makes sense to make it optional for all agents instead.

cvat-cli/src/cvat_cli/_internal/agent.py
```diff
@@ -85,22 +90,127 @@ def result(self, future: concurrent.futures.Future):
...
+
+        while (
+            self._id_to_ext_state
+            and next(iter(self._id_to_ext_state.values())).last_accessed_at < cutoff
+        ):
+            self._id_to_ext_state.popitem(last=False)
+
+
 _current_function: cvataa.AutoAnnotationFunction
+_tracking_states: _TrackingStateContainer
```
zhiltsov-max, line 147: If I understand correctly, states are stored in each worker individually and will be destroyed on an agent stop or an agent worker fail. 

I also couldn't find any logic to store the state somewhere (in the parent agent?) if it's changed in the `_worker_job_track`, so I'm not sure it's supported.

cvat-sdk/cvat_sdk/auto_annotation/interface.py
```diff
@@ -236,6 +236,142 @@ def detect(
...
+        for st in value:
+            models.ShapeType(st)
+
+
+class TrackingFunctionContext(metaclass=abc.ABCMeta):
+    """
+    Information that is supplied to an auto-annotation tracking function.
+
+    Currently, this class is empty, but in the future some properties may be added.
+    """
```
zhiltsov-max, line 267: I can suggest adding the frame distance between the track start and the current frame.

cvat-cli/src/cvat_cli/_internal/agent.py
```diff
@@ -85,22 +90,127 @@ def result(self, future: concurrent.futures.Future):
...
+            raise _BadArError(f"Tracking state {state_id!r} not found - possibly expired")
+
+        if ext_state.original_task_id != task_id:
+            # This is a defense-in-depth measure. State IDs are supposed to be unguessable,
+            # but even if an attacker manages to obtain one, they will not be able to use it
+            # to get any information about a task they don't have access to.
+            raise _BadArError(f"Tracking state {state_id!r} is not for task #{task_id}")
+
+        if image_dims != ext_state.original_image_dims:
+            raise _BadArError(f"Image size changed between frames")
```
zhiltsov-max: I was confused when I got this error in an noop agent, consider changing the message like this:

```suggestion
            raise _BadArError(f"Image sizes of the start frame and the current frame is different. Perhaps, the task is not a video task?")
```

cvat-cli/src/cvat_cli/_internal/agent.py
SpecLad: Yeah, there's currently no persistence for states. I think it's a reasonable approach for the initial version, since I don't expect a user to keep tracking any one object for long enough where this becomes an issue. It also makes it easier to implement tracker functions, since there's no need to write serialization/deserialization code. And it makes things a little faster, since we don't have to _run_ that code.

Later on, if we decide we need it, I think we can implement optional state persistence support: if a function opts into it and the user defines a data store on the command line, we'll pickle the state object and save it to that store instead of using `_TrackingStateContainer`.

> I also couldn't find any logic to store the state somewhere (in the parent agent?) if it's changed in the `_worker_job_track`, so I'm not sure it's supported.

I'm not sure what you mean. The states are stored in the `_tracking_states` variable in the child process. If a function modifies a state object, then it just stays modified.

cvat-sdk/cvat_sdk/auto_annotation/interface.py
SpecLad: Do you have an example for how a function might make use of this?

cvat-cli/src/cvat_cli/_internal/agent.py
SpecLad: I think the second sentence there is not quite correct, because it implies that the task must be created from a video file, which is not the case. I'll think about how best to phrase it.

cvat-cli/src/cvat_cli/_internal/agent.py
zhiltsov-max: The main use case I can see for state persistence now is that if an agent worker fails on some of the frames, all the states in the worker will be destroyed. I can see that the states are local to each worker, so they can be modified inside the worker. It's not a problem if we always handle the same request in the same worker.

cvat-sdk/cvat_sdk/auto_annotation/interface.py
zhiltsov-max: I don't have much information on trackers, but the length of the current track can be used to accumulate some information about the tracked object (e.g. embeddings) over multiple frames or to produce some labels or attributes (e.g. depending on the track duration). Without knowing the current distance, it's difficult to reliably store some information in the state, unless it's something generic, because you don't have a meaningful frame identifier.

cvat-cli/src/cvat_cli/_internal/agent.py
SpecLad: Sure, I understand the use case for persistence. I just thought that tracking support is still usable enough without it that we could postpone it until another iteration.

cvat-sdk/cvat_sdk/auto_annotation/interface.py
SpecLad: I think I'd rather leave this to the function. I think all you're describing is already possible by adding attributes to the state object.

cvat-cli/src/cvat_cli/_internal/agent.py
SpecLad: I ended up just using the first sentence by itself.

IMO, it would make sense to prohibit creating tracks on the UI level when frame sizes are unequal, so that the user won't see this message in the first place.

cvat-cli/src/cvat_cli/_internal/agent.py
SpecLad: Yeah, I agree. I didn't do this originally, because you can't have optional methods in protocols without it being either a) incorrect with regards to typing, or b) confusingly documented (this would involve splitting the interface into multiple protocols).

I implemented this, choosing a).

cvat-sdk/cvat_sdk/auto_annotation/interface.py
zhiltsov-max: Do we have a guarantee that the tracking function is always called for track frames in order, so that it's correct to do [`state["frame_idx"] += 1`](https://gist.github.com/SpecLad/26f0b7feacd70261154a393a085f9c73#file-func_sam2_tracker-py-L115)?

SpecLad: I believe so. There's no point in calling a tracking function out of order - if you call it on frame 3 before frame 2, then you'll get worse results for frame 3.

SpecLad:
A last-minute change: I realized that a tracking function that supports multiple shape types would want to keep track of what shape it needs to return from `track`; and since we're storing that information already, I just added it to the context.

cvat-sdk/cvat_sdk/auto_annotation/interface.py
```diff
@@ -236,6 +236,166 @@ def detect(
...
+
+
+class TrackingFunctionShapeContext(TrackingFunctionContext):
+    """
+    Information that is supplied to shape-specific methods of an auto-annotation tracking function.
+    """
+
```
zhiltsov-max, line 274: Consider adding label into the context or `init_tracking_state()`'s shape argument as well. Different classes can be tracked differently

SpecLad: This is true in principle, but so far none of the tracking functions we've implemented for CVAT used label information (and the UI doesn't even send the label when invoking tracking functions). So it doesn't seem useful enough to expend effort on supporting right now.

PR #9580, opened 2025-06-26
spell-checking integration

archibald1418:
Typos: correction dictionaries

This PR is to demonstrate capabilities of a correction-based spell-checker

https://github.com/crate-ci/typos

### Motivation and context

This is an effort to standardize spelling of some words and to avoid obvious typos in code.

### Checklist

### License


_typos.toml
```diff
@@ -0,0 +1,57 @@
...
+inconvinient = "inconvenient"
+clicable = "clickable"
+convertor = "converter"
+perameters = "parameters"
+succefull = "successful"
+roration = "rotation"
+tupe = "type"
+onject = "object"
+concatening = "concatenating"
+perpective = "perspective"
```
SpecLad: Could you put these in alphabetical order?

cvat-data/src/ts/3rdparty/Decoder.worker.js
SpecLad, line 1: This file is 3rd-party and should not be modified.

cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/controls-side-bar.tsx
```diff
@@ -295,7 +295,7 @@ export default function ControlsSideBarComponent(props: Props): JSX.Element {
             }
 
             canvasInstance.cancel();
-            // repeateDrawShapes gets all the latest parameters
+            // repeatDrawShapes gets all the latest parameters
```
SpecLad:
```suggestion
            // repeatDrawShape gets all the latest parameters
```

cvat-ui/src/components/annotation-page/top-bar/filters-modal.tsx
```diff
@@ -156,7 +156,7 @@ function FiltersModalComponent(): JSX.Element {
                     type: '!struct',
                     subfields: getAttributesSubfields(labels),
                     fieldSettings: {
-                        treeSelectOnlyLeafs: true,
+                        treeSelectOnlyLeaves: true,
```
SpecLad: This is a dependency typo, we can't fix it: https://github.com/ukrbublik/react-awesome-query-builder/blob/1035fd7e861035a0c157d35907f56540685d562e/packages/core/modules/index.d.ts#L1336

cvat/apps/engine/migrations/0033_projects_adjustment.py
SpecLad, line 1: I don't think we can rename existing migrations. These names are recorded the database, so changing them could make Django think a migration hasn't been applied when it actually has.

serverless/pytorch/foolwood/siammask/nuclio/model_handler.py
```diff
@@ -66,7 +66,7 @@ def infer(self, image, shape, state):
             state = self.decode_state(state)
             state = siamese_track(state, image, mask_enable=True,
                 refine_enable=True, device=self.device)
-            shape = state['ploygon'].flatten().tolist()
+            shape = state['polygon'].flatten().tolist()
```
SpecLad: Another dependency typo: https://github.com/foolwood/SiamMask/blob/a2b3ce3dca23b3a911d96b791e007eec6d39a625/tools/test.py#L314

tests/cypress/e2e/actions_projects_models/case_103_project_export.js
```diff
@@ -16,7 +16,7 @@ context('Export project dataset.', { browser: '!firefox' }, () => {
         attrValue: 'Oak',
         nameSecond: `Case ${caseID} second`,
         labelSecond: 'Car',
-        attrNameSecons: 'Color',
+        attrNameSeconds: 'Color',
```
SpecLad:
```suggestion
        attrNameSecond: 'Color',
```

tests/cypress/e2e/issues_prs2/issue_1498_message_ui_raw_labels_wrong.js
```diff
@@ -107,7 +107,7 @@ context('Message in UI when raw labels are wrong.', () => {
                 });
         });
         it('Label "attributes input_type" is incorrect.', () => {
-            const inputTypes = ['select radio', 'textt', 'nnumber'];
+            const inputTypes = ['select radio', 'textt', 'number'];
```
SpecLad: I think these are supposed to be incorrect.

.github/workflows/pylint.yml
```diff
@@ -16,3 +16,14 @@ jobs:
...
           echo "Pylint version: "$(pylint --version | head -1)
           pylint -j0 .
+
+  SpellCheck:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@4
+
+      - name: Run spell check
+        uses: crate-ci/typos@v1.34.0
```
SpecLad: Can you set it to use the version from `dev/requirements.txt`?

cvat-ui/src/components/annotation-page/top-bar/filters-modal.tsx
archibald1418: Yeah these are a problem. I rebuilt everything every time and everything worked well but I guess it's not that simple
Probably a good idea would be to run the spell-check on the dep source code too to see where the dependency typos are and include them in the dict or just add file directives

serverless/pytorch/foolwood/siammask/nuclio/model_handler.py
archibald1418: We would have to probably ignore these files separately

.github/workflows/pylint.yml
archibald1418: Each new version they update the `VERSION` env in their github-action script (e.g. [1.34](https://github.com/crate-ci/typos/blob/208cd070927dc240b5ed4c1fe54436c1118ff153/action/entrypoint.sh#L23), [1.32](https://github.com/crate-ci/typos/blob/3be83342e28b9421997e9f781f713f8dde8453d2/action/entrypoint.sh#L23)). It is downloaded then later with `wget`

So we would just have to denote the desired version in the name of the action. Applying

tests/cypress/e2e/issues_prs2/issue_1498_message_ui_raw_labels_wrong.js
archibald1418: Added a line disabling comment

_typos.toml
archibald1418: sorted with vscode `Sort Lines` extension

cvat-data/src/ts/3rdparty/Decoder.worker.js
archibald1418: ignored `.js` files inside `3rdparty` in `_typos.toml`

cvat/apps/engine/migrations/0033_projects_adjustment.py
archibald1418: renamed the file back, now spell-checker ignores it during spell-check and ignores a line in another migration file that references this migration name

serverless/pytorch/foolwood/siammask/nuclio/model_handler.py
archibald1418: Ignored only the line. This code is 5yo and no other references of this anywhere else,  safe to say that disabling a line should be enough 🤞🏽 
If `'ploygon'` is needed again, dev can do `# spellchecker: disable-line`

tests/cypress/e2e/actions_projects_models/case_103_project_export.js
archibald1418: removed `secons` entry from dict, easier to fix this manually to avoid confusion between seconds and second

.github/workflows/pylint.yml
archibald1418: Changed version of the called action

cvat-ui/src/components/annotation-page/top-bar/filters-modal.tsx
archibald1418: added `leafs` in the dictionary. This incorrect plural is actually a convention in tree-related code

_typos.toml
```diff
@@ -0,0 +1,68 @@
...
+	# Taken from https://github.com/crate-ci/typos/blob/master/docs/reference.md#example-configurations
+
+]
+
+[default.extend-words]
+AVOD = "AVOD"
+NAM = "NAM"
+adjastment = "adjustment"
+analitics = "analytics"
+archieve = "archive"
```
SpecLad: I'm not sure about this one - it could also be supposed to be "achieve". Probably better to use "" as the correction to force manual resolution.

_typos.toml
```diff
@@ -0,0 +1,68 @@
...
+blok = "block"
+cheeck = "check"
+clicable = "clickable"
+concatening = "concatenating"
+configurate = "configure"
+convertation = "conversion"
+convertor = "converter"
+correciton = "correction"
+corrent = "correct"
```
SpecLad: This is also ambiguous - it could be "current".

_typos.toml
```diff
@@ -0,0 +1,68 @@
...
+cubiod = "cuboid"
+cuboiud = "cuboid"
+cuccessfully = "successfully"
+custome = "custom"
+cuttent = "current"
+debounding = "debouncing"
+firts = "first"
+inconvinient = "inconvenient"
+keupoints = "keypoints"
+lastest = "lastest"
```
SpecLad: Can't we fix this one? "Lastest" isn't really a word.

_typos.toml
```diff
@@ -0,0 +1,68 @@
...
+lavel = "label"
+leafs = "leafs"
+nd = "nd"
+occlud = "occluded"
```
SpecLad: This could also be "occlude".

_typos.toml
```diff
@@ -0,0 +1,68 @@
...
+occurreed = "occurred"
+onject = "object"
```
SpecLad: This could be "inject".

_typos.toml
```diff
@@ -0,0 +1,68 @@
...
+perameters = "parameters"
+performe = "perform"
+perpective = "perspective"
+puted = "input"
```
SpecLad: This is too big of a change of meaning to be applied automatically. The obvious correction for "puted" would be "put".

_typos.toml
```diff
@@ -0,0 +1,68 @@
...
+lastest = "lastest"
+lavel = "label"
+leafs = "leafs"
+nd = "nd"
+occlud = "occluded"
+occurreed = "occurred"
+onject = "object"
+perameters = "parameters"
+performe = "perform"
+perpective = "perspective"
```
SpecLad: A bunch of these words seem to already be in the upstream dictionary, like this one:

https://github.com/crate-ci/typos/blob/208cd070927dc240b5ed4c1fe54436c1118ff153/crates/typos-dict/assets/words.csv#L48200

Why did you add them again here?

.github/workflows/pylint.yml
```diff
@@ -16,3 +16,14 @@ jobs:
 
           echo "Pylint version: "$(pylint --version | head -1)
           pylint -j0 .
+
+  SpellCheck:
```
SpecLad: This should be in its own workflow.

_typos.toml
```diff
@@ -0,0 +1,68 @@
...
+	"*.log",
+	"*.min.js",
+	"**/node_modules",
+    "*.patch",
+]
+default.extend-ignore-re = [
+	# test passwords
+    "\"md5.*?\"",
+
+	# Line ignore with trailing "# spellchecker: disable-line":
```
SpecLad: It seems better to ignore all lines with "spellchecker: disable-line" anywhere in them, because different languages have different comment syntaxes.

_typos.toml
```diff
@@ -0,0 +1,68 @@
...
+	"(?Rm)^.*(#|//)\\s*spellchecker: disable-line$",
+
+	# Line block with # spellchecker:<on|off>:
```
SpecLad: Is this actually used anywhere?

_typos.toml
```diff
@@ -0,0 +1,68 @@
+files.extend-exclude = [
+	"cvat-ui/src/assets/opencv_4.8.0.js",
+	"cvat/apps/engine/migrations/0033_projects_adjastment.py", # legacy migration name, set for life
+	"cvat-data/src/ts/3rdparty/*.js",
+	"**/*.zip",
+	"*.log",
+	"*.min.js",
+	"**/node_modules",
```
SpecLad: Aren't Git-ignored files supposed to be excluded automatically? `node_modules` should be one of them.

.github/workflows/pylint.yml
SpecLad: That's not what I mean. The workflow should dynamically get the version from `dev/requirements.txt`, so that we don't have to update this version twice in the future.

SpecLad: Look at `black.yml`, for example.

_typos.toml
archibald1418: It's not, to think of it, `latest` was probably meant [there](https://github.com/cvat-ai/cvat/blob/5b41abd909dc7e975320ced642df0ccec03c1720/cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/tools-control.tsx#L224)

_typos.toml
archibald1418: - Some of them are attempts to exclude other suggestions that are much less likely to appear in our context

> blok -> block,~bloke~
> cheeck -> check,~cheek~
> convertation -> conversion,~conversation~
> custome -> custom,~customs~ ~costume~ ~customer~ (although last one appears 3 times, I can keep it)
> firts -> ~flirts~,first
> lavel -> ~level~, ~laravel~, ~laravel~, label (although maybe level can be kept)

- Others are incorrect guesses by the dictionary

> correciton -> ~corrections~ (fixed to correction)
> tpic -> ~topic~ (+tpic)
> 

- One was a spelling, specific only to a library used 

> trak -> track

archibald1418: But with `perpective`, just found out there seems to be a bug with tokens containing slashes

> Hide/unhide a cuboid via sidear. The cuboid on the perpective/top/side/front view be hidden/unhidden

doesn't find `perpective`. I thought it was a dict issue and mistakenly added this to the config

And they haven't fixed it yet: https://github.com/crate-ci/typos/issues/758

Codespell catches it though:
image

_typos.toml
archibald1418: This works fine with slashes, too
And easily extensible if we need this in e.g. `.sql` files

_typos.toml
archibald1418: No, but this is a recommendation from typos. There might be a need for a block disable/enable (like eslint has)

_typos.toml
archibald1418: It does, removing this. `typos --files` doesn't show it

_typos.toml
archibald1418: Agreed, applying

_typos.toml
archibald1418: I'll leave two suggestions

_typos.toml
archibald1418: Leaving both suggestions

_typos.toml
archibald1418: Removing

_typos.toml
archibald1418: Agreed, removing

_typos.toml
archibald1418: UPD: no option to have several suggestions, but they are working on it https://github.com/crate-ci/typos/issues/1041

Sadly, `codespell` can't update dictionaries but only [allows](https://github.com/codespell-project/codespell#updating-the-dictionaries) to replace them completely. This is not what we want, it will require a lot of configuration and we also would like to use their built-in dicts. In their issues I found no plans on adding this. Only option is to copy their big dicts manually and update them locally, I'm not sure this is more convenient than a much smaller `_typos.toml`

For now, we can only add false-negatives that can only have one suggestion. I would suggest we just ignore the ambiguous ones and let them slip.

archibald1418: They maintain a monthly _Non-critical corrections_ issues where they collect suggested corrections from the users, and merge them promptly

_typos.toml
SpecLad: > * Some of them are attempts to exclude other suggestions that are much less likely to appear in our context

I don't think we need to micro-manage our dictionary to that extent. You can't always predict in advance which suggestion would be correct, and even if you could, there will always be more ambiguous typos in the dictionary. Developers can always resolve the typos manually.

> * Others are incorrect guesses by the dictionary
> * One was a spelling, specific only to a library used

Yeah, those are fine. (Although note that `trak` is not "specific to a library", it's an identifier from the ISO base media file format spec.)

> But with `perpective`, just found out there seems to be a bug with tokens containing slashes

Okay, makes sense.

_typos.toml
SpecLad: > For now, we can only add false-negatives that can only have one suggestion. I would suggest we just ignore the ambiguous ones and let them slip.

I'm not sure we have to ignore them - I got the impression that you can write e.g. `onject = ""` to flag such typos without providing a suggestion.

> They maintain a monthly _Non-critical corrections_ issues where they collect suggested corrections from the users, and merge them promptly

Yeah, I went ahead and submitted the least ambiguous typos from your list upstream already.

_typos.toml
SpecLad: Seems like you can remove `*.zip` too, then. Looks like typos skips binary files by default.

_typos.toml
```diff
@@ -0,0 +1,59 @@
...
+perameters = "parameters"
+performe = "perform"
+roration = "rotation"
+segway = "segway"
+splitted = "splitted"
+succefull = "successful"
+tesk = "task"
+tpic = "tpic"
+trak = "trak"
+tupe = "type"
```
SpecLad: This one is also ambiguous, it could be "tuple".

_typos.toml
```diff
@@ -0,0 +1,59 @@
...
+	"*.min.js",
+    "*.patch",
+]
+default.extend-ignore-re = [
+	# test passwords
+    "\"md5.*?\"",
+
+	# Line ignore with trailing "# spellchecker: disable-line"
+	# or trailing "// spellchecker: disable-line"
+	"(?Rm)^.*(#|//)\\s*spellchecker: disable-line$",
```
SpecLad: It's inconsistent that there's a space after the colon here, but not in `spellchecker:on`/`off`. I see that there's no space in the original doc, so I'd suggest to remove it.

site/content/en/images/shapes-converter-action-run.jpg
SpecLad, line 1: It looks like you haven't updated the references to these images from Markdown files.

_typos.toml
archibald1418: > I got the impression that you can write e.g. onject = "" to flag such typos without providing a suggestion.

Great point, we can always just force the user to fix it manually

_typos.toml
archibald1418: Ok, removing the micro-management

_typos.toml
archibald1418: removing

_typos.toml
archibald1418: > You can't always predict in advance which suggestion would be correct

In case with `block vs bloke,` `check vs cheek,` `flirts vs first` I think you can (these suggestions are way off for a technical text). Removing others though

archibald1418:
🔽  expected fail, will be fixed with pre-commit integration

_typos.toml
```diff
@@ -0,0 +1,58 @@
+files.extend-exclude = [
+	"cvat-ui/src/assets/opencv_4.8.0.js",
+	"cvat/apps/engine/migrations/0033_projects_adjastment.py", # legacy migration name, set for life
+	"cvat-data/src/ts/3rdparty/*.js",
+	"**/*.zip",
+	"*.log",
+	"*.min.js",
+    "*.patch",
```
klakhov, line 7: I think we also should add `.onnx` decoders here. Or are they already ignored?

archibald1418: `typos` has a fixed list of source code file formats that it checks, and `.onnx` is simply not in it

https://github.com/crate-ci/typos/blob/208cd070927dc240b5ed4c1fe54436c1118ff153/crates/typos-cli/src/default_types.rs

UPD: assets ignore was related to the typos project itself, sorry for the confusion

klakhov: If it ignores `/assets`, why we have `opencv-js` in ignored list?

archibald1418: Consecutively, `typos --file-types` doesn't detect `.onnx` although they are not ignored by `.gitignore`

archibald1418: @klakhov I was wrong, it doesn't, it just doesn't detect `.onnx` files

_typos.toml
```diff
@@ -0,0 +1,58 @@
+files.extend-exclude = [
+	"cvat-ui/src/assets/opencv_4.8.0.js",
```
SpecLad: Please replace tabs with spaces.

archibald1418: done

_typos.toml
archibald1418: Applied

site/content/en/images/shapes-converter-action-run.jpg
archibald1418: https://github.com/cvat-ai/cvat/blob/6073cf07edda7433febbe5cf55cada1b720bf2b3/site/content/en/docs/manual/advanced/analytics-and-monitoring/manual-qa.md#L140

Renamed link from standart to standard

archibald1418: https://github.com/cvat-ai/cvat/blob/6073cf07edda7433febbe5cf55cada1b720bf2b3/site/content/en/docs/enterprise/shapes-converter.md#L79

Renamed link from coverter -> converter

SpecLad (approved):
Looks good to me. @klakhov Do you have any other comments?

klakhov (approved):
LGTM

PR #9581, opened 2025-06-27
Added Redis shared cache for throttling

azhavoro:
### Motivation and context

Changed the default Django cache backend from LocMem to Redis, so that the cache is shared by all server processes and used for throttling to work correctly.

### Checklist

### License

### DIFF
--- a/changelog.d/20250630_142609_andrey_shared_throttle_cache.md
+++ b/changelog.d/20250630_142609_andrey_shared_throttle_cache.md
@@ -1,0 +1,4 @@
+### Changed
+
+- Changed the default Django cache backend from LocMem to Redis
+  (<https://github.com/cvat-ai/cvat/pull/9581>)


cvat/settings/base.py
```diff
@@ -583,6 +583,10 @@ class CVAT_QUEUES(Enum):
         "LOCATION": f"redis://:{urllib.parse.quote(redis_ondisk_password)}@{redis_ondisk_host}:{redis_ondisk_port}",
         "TIMEOUT": CVAT_CHUNK_CACHE_TTL,
     },
+    "shared": {
+        "BACKEND": "django.core.cache.backends.redis.RedisCache",
```
SpecLad: Can we just use Redis for the default cache? Is there any use case where a local memory cache is preferable?

cvat/settings/base.py
```diff
@@ -583,6 +583,10 @@ class CVAT_QUEUES(Enum):
...
+        "LOCATION": f"redis://:{urllib.parse.quote(redis_inmem_password)}@{redis_inmem_host}:{redis_inmem_port}",
```
SpecLad: Can we store the cache in a different database number, so that it doesn't interfere with jobs?

cvat/utils/throttle.py
```diff
@@ -0,0 +1,14 @@
+# Copyright (C) CVAT.ai Corporation
+#
+# SPDX-License-Identifier: MIT
+
+from django.core.cache import caches
+from rest_framework.throttling import AnonRateThrottle
+
+
+class SharedCacheThrottle(AnonRateThrottle):
```
SpecLad: Is this supposed to inherit from `AnonRateThrottle`? I assume the point of this class is to be a mixin for different types of throttles.

cvat/settings/base.py
azhavoro: Hmm, that seems reasonable, I'll change the default entry.

cvat/utils/throttle.py
azhavoro: Oh, exactly, my bad

SpecLad:
Please add a changelog entry. Otherwise LGTM.

SpecLad:
And edit the description please, the current one doesn't reflect the contents of the PR anymore.

PR #9543, opened 2025-06-18
Add dynamic field selector to actions menu

klakhov:
### Motivation and context

This PR adds additional menu component into actions menu of different resources. With dynamic selector it is now possible to edit different fields using menu. 

Currently supported:
- Assignee for task, project, job
- State, Stage for job

selector-job-assignee

### Checklist

- [x] I submit my changes into the `develop` branch
- [x] I have created a changelog fragment 
- [ ] I have updated the documentation accordingly
- [ ] I have added tests to cover my changes
- ~~[ ] I have linked related issues (see [GitHub docs](
  https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword))~~

### License

- [x] I submit _my code changes_ under the same [MIT License](
  https://github.com/cvat-ai/cvat/blob/develop/LICENSE) that covers the project.
  Feel free to contact the maintainers if that's a concern.

PR #9582, opened 2025-06-27
Bulk actions on UI

klakhov:
### Motivation and context

The PR introduces a way to select multiple resources and make a bulk operation on full group.

Depends on #9543, #9527
### Checklist

### License


bsekachev:
When I selected 1,2,3 with Ctrl and then holding shift I click number 5. 
It should select 1,2,3,4,5, not 3,4,5 (previously selected should not reset)

bsekachev:
UX: Arrow up/down does not work in this selector

image

UX/BUG:  Selector is still visible while updating and I can edit value from keyboard

image

bsekachev:
BUG:  Any click here resets the selection and bulk export 

image

bsekachev:
Performance: Let's send preview requests one by one. Otherwise we are risky to ddos ourselves.

bsekachev:
On jobs/projects page we have dedicated "Reset assignee":

image

On task page / tasks page we do not:

image

Why behaviour is different?

bsekachev:
Missing padding before icon

image

bsekachev:
Not sure it is directly related to the PR, but after request cancellation, it still has active status "Queued" with loading icon. 
It confuses a bit. 

image

But related thing is that cancelled request still may be selected:

image

It then leads to errors related to repeated cancellation:

image

___
Please check the same thing for deleted task/project

changelog.d/20250708_134458_klakhov_bulk_actions.md
```diff
@@ -0,0 +1,5 @@
+### Added
+
+- Introduced bulk actions to perform single operations on multiple selected resources at once
```
bsekachev:
```suggestion
- Introduced bulk actions to perform operations on multiple selected resources at once
```

bsekachev: Let's specify more details about supported bulk operations

cvat-ui/src/components/task-page/task-page.tsx
```diff
@@ -37,19 +37,26 @@ function TaskPageComponent(): JSX.Element {
...
+    const isTaskUpdating = (updates[id] || jobsFetching) && !bulkFetching;
 
     const receiveTask = (): Promise<Task[]> => {
         if (Number.isInteger(id)) {
             const promise = core.tasks.get({ id });
             promise.then(([task]: Task[]) => {
                 if (task) {
                     setTaskInstance(task);
+                    if (task.jobs && Array.isArray(task.jobs)) {
+                        const jobsList = [...task.jobs] as JobsList;
```
bsekachev: task.jobs already have specified type: `Job[]` and it is not compatible with JobsList having also count. So, this cast does not seem to be correct. 

Maybe to rewrite it just as:

```
dispatch(jobsActions.getJobsSuccess(
    Object.defineProperty([...task.jobs], 'count', { value: ... })
));
```

cvat-ui/src/utils/platform.ts
```diff
@@ -0,0 +1,18 @@
+// Copyright (C) CVAT.ai Corporation
+//
+// SPDX-License-Identifier: MIT
+
+export function platform(): string {
+    let platformName = 'unknown';
+
+    if (window.isSecureContext && 'userAgentData' in navigator) {
+        // Use the new User-Agent Client Hints API when available in secure context
+        platformName = (navigator as any).userAgentData?.platform || navigator.platform;
```
bsekachev: With just ``if .. else`` it may be rewritten with two returns. 'unknown' is not necessary in the beginning.
Additionally, I believe .toLowerCase() should not be in utility function. If outside code needs lowercase, better to do it outside.

cvat-ui/src/index.tsx
```diff
@@ -123,10 +129,13 @@ function mapDispatchToProps(dispatch: any): DispatchToProps {
         resetErrors: (): void => dispatch(resetErrors()),
         resetMessages: (): void => dispatch(resetMessages()),
         loadOrganization: (): void => dispatch(activateOrganizationAsync()),
-        initInvitations: (): void => dispatch(getInvitationsAsync({ page: 1 }, true)),
-        initRequests: (): void => dispatch(getRequestsAsync({ page: 1 })),
+        initInvitations: (): void => dispatch(getInvitationsAsync({ page: 1, pageSize: 10 }, true)),
+        initRequests: (): void => dispatch(getRequestsAsync({ page: 1, pageSize: 10 })),
```
bsekachev: Should be initialized from InvitationsState/RequestsState instead of hard-coding 10 here.

bsekachev: BTW, `pageSize` is missing in default value for InvitationdState and typescript complaints about it

cvat-ui/src/components/cvat-app.tsx
```diff
@@ -456,7 +471,40 @@ class CVATApplication extends React.PureComponent<CVATAppProps & RouteComponentP
                 }
             }
         }
-
+        const bulkError = notifications.errors.selection.bulkOperation;
```
bsekachev: I believe bulk-progress component should be responsible for this logic. Like:

... progress ...
... N items failed, repeat? ... 

What do you think?

cvat-ui/src/components/bulk-progress.tsx
```diff
@@ -0,0 +1,36 @@
+// Copyright (C) CVAT.ai Corporation
+//
+// SPDX-License-Identifier: MIT
+
+import React from 'react';
+import { Progress, Button } from 'antd';
```
bsekachev: Maybe this is not necessary anymore (need to double check), but let's stick to current style of importing each component from `antd` separately.

cvat-ui/src/components/bulk-progress.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { useSelector, useDispatch } from 'react-redux';
+import { selectionActions } from 'actions/selection-actions';
+import { CombinedState } from 'reducers';
+
+export default function BulkProgress(): JSX.Element | null {
+    const dispatch = useDispatch();
+    const fetching = useSelector((state: CombinedState) => state.selection.fetching);
+    const status = useSelector((state: CombinedState) => state.selection.status);
```
bsekachev: Please, combine useSelectors.

cvat-ui/src/components/bulk-progress.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+    return (
+        <div className='cvat-bulk-progress-wrapper'>
+            <Progress type='circle' percent={percent} size={120} />
+            <div className='cvat-bulk-progress-message'>
+                {message}
+            </div>
+            <Button
+                className='cvat-bulk-progress-cancel'
+                danger
```
bsekachev: I do not think it should be `danger` by concept.

cvat-ui/src/reducers/index.ts
```diff
@@ -257,6 +258,18 @@ export interface CloudStoragesState {
     };
 }
 
+export interface SelectionStatus {
```
bsekachev: It seems this is not SelectionStatus, this is BulkActionStatus

cvat-ui/src/reducers/index.ts
```diff
@@ -257,6 +258,18 @@ export interface CloudStoragesState {
...
+    message: string;
+    percent: number;
+}
+
+export interface SelectionState {
+    selected: (number | string)[];
+    fetching: boolean;
+    status: { message: string; percent: number } | null;
```
bsekachev: SelectionStatus (BulkActionStatus) | null ?

cvat-ui/src/reducers/index.ts
```diff
@@ -645,6 +658,17 @@ export interface NotificationsState {
             fetching: null | ErrorState;
             canceling: null | ErrorState;
             deleting: null | ErrorState;
+        };
+        selection: {
+            bulkOperation: {
+                message: string,
+                remainingItemsCount: number,
```
bsekachev:
```suggestion
                message: string;
                remainingItemsCount: number;
```

etc...

cvat-ui/src/reducers/index.ts
```diff
@@ -1000,6 +1024,9 @@ export interface OrganizationState {
     leaving: boolean;
     removingMember: boolean;
     updatingMember: boolean;
+    gettingMembers: boolean;
```
bsekachev:
```suggestion
    fetchingMembers: boolean;
```

cvat-ui/src/reducers/index.ts
```diff
@@ -645,6 +658,17 @@ export interface NotificationsState {
             fetching: null | ErrorState;
             canceling: null | ErrorState;
             deleting: null | ErrorState;
+        };
+        selection: {
+            bulkOperation: {
```
bsekachev: I would propose to split `selection` logic and `bulkOperation` logic on state level. 
E.g. now in this state nothing is responsible for `selection`, it is only about bulk operation.

bsekachev: Selection is only responsible for adding/removing elements from selection IMO. 
Bulk operation is responsible for type/progress/status/fails.

cvat-ui/src/reducers/selection-reducer.ts
```diff
@@ -0,0 +1,52 @@
...
+    switch (action.type) {
+        case SelectionActionsTypes.DESELECT_RESOURCES:
+            return { ...state, selected: state.selected.filter((id) => !action.payload.resourceIds.includes(id)) };
+        case SelectionActionsTypes.SELECT_RESOURCES:
+            if (action.payload.extendSelection) {
+                return { ...state, selected: Array.from(new Set([...state.selected, ...action.payload.resourceIds])) };
+            }
+            return { ...state, selected: [...action.payload.resourceIds] };
+        case SelectionActionsTypes.CLEAR_SELECTED_RESOURCES:
+            return { ...state, selected: [] };
```
bsekachev: Seems like redundant action. You may just call SELECT_RESOURCES with empty array.

cvat-ui/src/reducers/selection-reducer.ts
```diff
@@ -0,0 +1,52 @@
...
+} from '../actions/selection-actions';
+
+const initialState: SelectionState = {
+    selected: [],
+    fetching: false,
+    status: null,
+    cancelled: false,
+};
+
+export default function selectionReducer(
```
bsekachev: As mentioned before, each state (Projects/Tasks ... etc) should store selected items. 
In this case this one would be `bulkOperationReducer`. 

Select/Deselect may be a separated ActionTypes supported by different reducers in common code.

cvat-ui/src/styles.scss
```diff
@@ -210,4 +210,49 @@ html, body {
...
+    background: rgba(255,255,255,70%);
+    z-index: 2000;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    flex-direction: column;
+}
+
+.cvat-bulk-progress-message {
+    margin-top: 24px;
```
bsekachev: Please, use grid-unit-size in styles

klakhov:
- Fixed shift behaviour, now it only appends
- Fixed arrow buttons for dynamic selector and now its hidden right after selection
- Optimized preview fetching in #9692
- "Reset assignee" was shown only in 'bulk mode'(when several elements were selected), changed it now it is always shown
- Added padding in info icon
- Now when request is cancelled we will show 'Cancelled' status 
- Removed selection of cancelled/deleted resources with 'select all'

changelog.d/20250708_134458_klakhov_bulk_actions.md
klakhov: Added more info

cvat-ui/src/utils/platform.ts
klakhov: Moved this to platform-checker, replaced with two `if`

cvat-ui/src/components/cvat-app.tsx
klakhov: Moved to bulk-progress

cvat-ui/src/reducers/index.ts
klakhov: Makes sence, removed.

cvat-ui/src/reducers/selection-reducer.ts
klakhov: I’d prefer to keep this action. The current `Select` action works as **“append some resources”**, not **“fully replace selected resources.”**

This is useful when we want to append items (e.g., via Ctrl or Shift click), so we don’t need to worry about some previously selected ones. It keeps the "set" logic centralized at the reducer level.

PR #9590, opened 2025-07-01
Added uvicorn upstream backup

azhavoro:
### Motivation and context

This PR adds another uvicorn process as an upstream backup in addition to the main processes.
The idea is to have a uvicorn process that will only handle requests when the main process becomes unavailable during restart (mostly after killing due OOM).

### Checklist

### License

### DIFF
--- a/changelog.d/20250702_110059_andrey_uvicorn_upstream_backup.md
+++ b/changelog.d/20250702_110059_andrey_uvicorn_upstream_backup.md
@@ -1,0 +1,5 @@
+### Added
+
+- Uvicorn process as an upstream backup. This should help if the main process
+  is not available during restart (mostly after killing due OOM).
+  (<https://github.com/cvat-ai/cvat/pull/9590>)


supervisord/server.conf
```diff
@@ -20,3 +20,15 @@ autorestart=true
 environment=CVAT_EVENTS_LOCAL_DB_FILENAME="events_%(process_num)03d.db"
 numprocs=%(ENV_NUMPROCS)s
 process_name=%(program_name)s-%(process_num)d
+
+[fcgi-program:uvicorn-backup]
```
SpecLad, line 24: I'm probably missing something, but how is this better than increasing `NUMPROCS` by one? If we're going to have another server taking up RAM, why not have it help with incoming requests too?

azhavoro: Of course, increasing the number of main processes should improve the situation by reducing the probability of simultaneous unavailability of 2 or more services. But this has the negative effect that the same amount of memory will be consumed by two or more processes and restarts will happen more often. The idea of ​​​​the backup is based on the fact that restarts of the main process happen rarely and the backup will work for a very short time. Accordingly, it is expected that it will not be subject to memory leaks and will live as long as the container lives, without consuming significant resources.

PR #9599, opened 2025-07-03
[Documentation] Limit the documentation versions to the 6 ones

paveltovchigrechko:
### Motivation and context

Currently, we build the documentation for every repository tag starting from 1.5.0. This results in large list of the versions which do not differ too much between themselves. The drop-down menu in the website overflows the visible area at this point.
This PR limits the number of versions to 6 (changeable in `build_docs.py`). It also reverses the versions, so in the drop-down menu on the website, the versions go from the latest to the earliest top-down.

So far, the main blog had only one link to old version documetation, and it is changed now. I hope there are no other links to the old versions, so nothing is broken with this PR.

### How has this been tested?

- Run the website as for production (see site/README.md for instructions). Deploy with `python -m http.server -d <...>/public`
- Open the website
- Open the drop-down menu for versions. The menu should contain the six latest versions.
image

### Checklist

### License

### DIFF
--- a/site/README.md
+++ b/site/README.md
@@ -80,13 +80,11 @@ pip -m venv venv
 pip install -r requirements.txt
 ```
 
-The documentation site includes both old and new releases. Because of this,
-you will need tooling for all these releases. Currently, it means you need
-in your environment:
-- `hugo-0.110` - for new docs
-- `hugo-0.83` - for older docs
+The documentation site includes both old and new releases. Since we now only build
+the 6 most recent releases, you only need:
+- `hugo-0.110` - for all documentation builds
 
-Please download these hugo releases (both extended), and make such binaries
+Please download this hugo release (extended), and make the binary
 available in your `PATH` environment variable.
 
 On Linux, you can install it this way:
@@ -96,11 +94,6 @@ wget https://github.com/gohugoio/hugo/releases/download/v0.110.0/hugo_extended_0
 (mkdir hugo_extended_0.110.0_Linux-64bit && tar -xf hugo_extended_0.110.0_Linux-64bit.tar.gz -C hugo_extended_0.110.0_Linux-64bit)
 cd hugo_extended_0.110.0_Linux-64bit
 sudo cp hugo /usr/local/bin/hugo-0.110
-
-wget https://github.com/gohugoio/hugo/releases/download/v0.83.0/hugo_extended_0.83.0_Linux-64bit.tar.gz
-(mkdir hugo_extended_0.83.0_Linux-64bit && tar -xf hugo_extended_0.83.0_Linux-64bit.tar.gz -C hugo_extended_0.83.0_Linux-64bit)
-cd hugo_extended_0.83.0_Linux-64bit
-sudo cp hugo /usr/local/bin/hugo-0.83
 ```
 
 2. Use the commands that generate a static site in the `public/` folder:
--- a/site/README.md
+++ b/site/README.md
@@ -80,8 +80,8 @@ pip -m venv venv
 pip install -r requirements.txt
 ```
 
-The documentation site includes both old and new releases. Since we now only build
-the 6 most recent releases, you only need:
+The documentation site includes the last MAX_VERSIONS_TO_BUILD releases
+(they are defined in `build_docs.py`). To build the website, you only need:
 - `hugo-0.110` - for all documentation builds
 
 Please download this hugo release (extended), and make the binary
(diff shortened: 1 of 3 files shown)


site/build_docs.py
```diff
@@ -19,6 +19,9 @@
 # the initial version for the documentation site
 MINIMUM_VERSION = version.Version("1.5.0")
 
+# Number of most recent tags to build documentation for
+MAX_VERSIONS_TO_BUILD = 6
+
 # apply new hugo version starting from release 2.9.2
 UPDATED_HUGO_FROM = version.Version("2.9.2")
```
SpecLad: The new tag selection logic you implemented obsoletes both the `UPDATED_HUGO_FROM` and `MINIMUM_VERSION` variables, so I suggest you remove them and everything that depends on them.

paveltovchigrechko: Seems reasonable. Deleted now unused variables, and it still works
image

site/build_docs.py
```diff
@@ -16,32 +16,26 @@
...
 
-# apply new hugo version starting from release 2.9.2
-UPDATED_HUGO_FROM = version.Version("2.9.2")
-
-# Start the name with HUGO_ for Hugo default security checks
-VERSION_URL_ENV_VAR = "HUGO_VERSION_REL_URL"
-
-# Hugo binaries for different versions
-hugo110 = "hugo-0.110"  # required for new docs
-hugo83 = "hugo-0.83"  # required for older docs
```
SpecLad, line 30: Please go over other occurrences of `hugo-0.83` in the codebase and remove them as well.

site/build_docs.py
```diff
@@ -16,32 +16,26 @@
...
     for tag in repo.tags:
         tag_version = version.parse(tag.name)
-        if tag_version >= MINIMUM_VERSION and not tag_version.is_prerelease:
-            release_version = (tag_version.major, tag_version.minor)
-            if release_version not in tags or tag_version > version.parse(
-                tags[release_version].name
-            ):
-                tags[release_version] = tag
+        if not tag_version.is_prerelease:
+            valid_tags.append(tag)
```
SpecLad: I believe the logic implemented here is incorrect. What you do is take the last `MAX_VERSIONS_TO_BUILD` releases. What we want is to take the last `MAX_VERSIONS_TO_BUILD` minor versions, and take the latest patch version from each. That's how the current implementation works, except it takes all minor versions after `MINIMUM_VERSION`, rather than a fixed number of versions.

site/build_docs.py
```diff
@@ -140,24 +126,19 @@ def run_hugo(
...
             )
 
 
 def validate_env():
-    for hugo in [hugo83, hugo110]:
-        try:
-            subprocess.run([hugo, "version"], capture_output=True)  # nosec
-        except (subprocess.CalledProcessError, FileNotFoundError) as ex:
-            raise Exception(f"Failed to run '{hugo}', please make sure it exists.") from ex
+    # Only validate hugo110 since we use it for all builds now
```
SpecLad: This comment won't make sense to someone who doesn't know what the code looked like before. IMO, it can be safely removed.

site/build_docs.py
paveltovchigrechko: Deleted in the workflow (`docs.yml`) and README.

site/build_docs.py
paveltovchigrechko: Rewrited the function to get the latest minor versions from the latest MAX_VERSIONS_TO_BUILD releases.

site/build_docs.py
paveltovchigrechko: Deleted the comment

site/README.md
```diff
@@ -80,13 +80,11 @@ pip -m venv venv
 pip install -r requirements.txt
 ```
 
-The documentation site includes both old and new releases. Because of this,
-you will need tooling for all these releases. Currently, it means you need
-in your environment:
-- `hugo-0.110` - for new docs
-- `hugo-0.83` - for older docs
+The documentation site includes both old and new releases. Since we now only build
+the 6 most recent releases, you only need:
```
SpecLad: 1. It doesn't really make sense to say "old and new" anymore, since all included releases are now "new" in the sense that the original paragraph intended.

2. "Now" doesn't mean anything unless you also explain what the situation was before (which I don't think is really useful for someone reading this document). Point is, try to just explain the build requirements at the current point in time, without references to history.

3. Please avoid hardcoding the number 6 here, since we might want to change it someday (and we'll inevitably forget to update this).

paveltovchigrechko: Updated README.

site/README.md
```diff
@@ -128,7 +126,7 @@ git submodule update --remote
 
 Add and then commit the change to project:
 
-```bash
+```bashs
```
SpecLad: This looks like an unintentional change, could you fix it?

paveltovchigrechko: Done.

PR #9607, opened 2025-07-04
UI: add request types to lambda function calls

SpecLad:
### Motivation and context

These are used in the Enterprise version to validate annotation requests to external lambda functions.

Having an explicit type allows us to verify that only fields that are appropriate to the particular AR type are included, as well as disambiguate between AR types if a function supports several of them (like `init_tracking` and `track`).

This is needed to support agent-based tracker functions.

The code for Nuclio-based functions currently does not use the `type` field, but eventually I'd like to merge the validation logic between them and external functions, at which point it will.

### Checklist

### License

### DIFF
--- a/cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/tools-control.tsx
+++ b/cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/tools-control.tsx
@@ -41,7 +41,7 @@ import {
     updateAnnotationsAsync,
     createAnnotationsAsync,
 } from 'actions/annotation-actions';
-import DetectorRunner, { DetectorRequestBody } from 'components/model-runner-modal/detector-runner';
+import DetectorRunner, { AnnotateTaskRequestBody } from 'components/model-runner-modal/detector-runner';
 import LabelSelector from 'components/label-selector/label-selector';
 import CVATTooltip from 'components/common/cvat-tooltip';
 import CVATMarkdown from 'components/common/cvat-markdown';
@@ -782,6 +782,7 @@ export class ToolsControlComponent extends React.PureComponent<Props, State> {
                         });
 
                         const response = await core.lambda.call(jobInstance.taskId, tracker, {
+                            type: 'init_tracking',
                             frame: frame - 1,
                             shapes: trackableObjects.shapes,
                             job: jobInstance.id,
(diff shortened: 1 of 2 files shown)


cvat-ui/src/components/model-runner-modal/detector-runner.tsx
```diff
@@ -38,6 +38,7 @@ type ServerMapping = Record<string, {
 }>;
 
 export interface DetectorRequestBody {
+    type: string,
```
bsekachev:
```suggestion
    type: 'annotate_task' | 'annotate_frame';
```

cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/tools-control.tsx
```diff
@@ -1213,7 +1215,7 @@ export class ToolsControlComponent extends React.PureComponent<Props, State> {
                         const { cleanup, ...restOfBody } = body;
 
                         const result = await core.lambda.call(jobInstance.taskId, model, {
-                            ...restOfBody, frame, job: jobInstance.id,
+                            ...restOfBody, type: 'annotate_frame', frame, job: jobInstance.id,
```
bsekachev, line 1218: 1. `type` should come from `DetectorRunner` as it declared in `DetectorRequestBody` which is argument of `runInference` callback. This `type` may be added as a prop to `DetectorRunner` and then passed to callback for correct typing.
2. Alternatively we may remove `type` from `DetectorRequestBody`, but in this case this is not `DetectorRequestBody`, it is more like `InferenceParameters`. And `DetectorRequestBody` is something like `{ ...InferenceParameters, type: 'val1' | val2 }`


`DetectorBody` but

  `body: DetectorRequestBody`

bsekachev:
Just a reminder to do the same in AI Tracker action in enterprise repo

cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/tools-control.tsx
SpecLad: I'm proposing something simpler. `DetectorRunner` always submits a request body for an `annotate_task` request, so I renamed `DetectorRequestBody` to `AnnotateTaskRequestBody` and made `type` a literal type. This function takes that request and converts it into an `annotate_frame` request, for which no interface exists (but we could define one if need be).

SpecLad:
> Just a reminder to do the same in AI Tracker action in enterprise repo

Yeah, I already did that in the PR that links to this one.

PR #9617, opened 2025-07-08
Improved performance of database queries when receiving tracks

bsekachev:
### Checklist

### License


bsekachev:
@zhiltsov-max May you please also look at this one?

cvat/apps/dataset_manager/task.py
```diff
@@ -592,30 +613,16 @@ def _init_tags_from_db(self):
...
-                        "spec_id": attr["spec_id"],
-                        "value": attr["value"],
-                    }
-                )
-            )
+        )
 
-        db_tags = list(db_tags.values())
         for db_tag in db_tags:
+            db_tag.attributes = labeledimage_attributes[db_tag.id]
```
zhiltsov-max, line 624: Also a race condition, it's better to skip missing keys

Upd: see you're relying on a defaultdict

cvat/apps/dataset_manager/task.py
```diff
@@ -35,6 +35,7 @@
 from cvat.apps.profiler import silk_profile
 
 dlogger = DatasetLogManager()
+ANNOTATIONS_CHUNK_SIZE = 2000
```
zhiltsov-max: Typically it makes sense to make such performance constants easily configurable. Consider putting it into `cvat/apps/engine/default_settings.py` and using `os.getenv` to allow passing via an env variable.

cvat/apps/dataset_manager/task.py
```diff
@@ -735,41 +718,36 @@ def _init_tracks_from_db(self):
...
-                        "attribute__id",
-                    ]
-                },
-                "id",
-            )
 
             # A result table can consist many equal rows for track/shape attributes
             # We need filter unique attributes manually
-            db_track["attributes"] = list(set(db_track["attributes"]))
+            db_track["attributes"] = list(set(labeledtrack_attributes[db_track["id"]]))
```
zhiltsov-max, line 742: Do you know why there can be duplicates?

bsekachev: Not really. Technically easy as this is one to many relation, however on practice IDK
I wanted to ask @SpecLad about it

And why we do not check the same for shapes and tags.

cvat/apps/dataset_manager/task.py
bsekachev: Added

cvat/apps/dataset_manager/task.py
zhiltsov-max: I guess  the answer is that there are no DB constraints to prevent duplicated spec_ids per shape_id. Ok, let's keep this as it is.

zhiltsov-max: If this is the case, maybe we need to add some deduplication means on saving or use attrs with max `id` for each `spec_id`, otherwise the returned attributes can be random. Anyway, it's not a topic for this PR.

PR #9621, opened 2025-07-08
Fix inconsistent `related_name`s and `related_query_name`s in `engine` models

SpecLad:
### Motivation and context

* Set `related_query_name` to singular nouns, which matches the Django default (and also some of the existing model fields).

* Fix inconsistent `related_name`s in `CloudStorage` fields.

### Checklist

### License


cvat/apps/engine/models.py
```diff
@@ -586,9 +589,9 @@ class Task(TimestampedModel, FileSystemRelatedModel):
     name = SafeCharField(max_length=256)
     mode = models.CharField(max_length=32)
     owner = models.ForeignKey(User, null=True, blank=True,
-        on_delete=models.SET_NULL, related_name="owners")
+        on_delete=models.SET_NULL, related_name="tasks", related_query_name="task")
     assignee = models.ForeignKey(User, null=True,  blank=True,
-        on_delete=models.SET_NULL, related_name="assignees")
+        on_delete=models.SET_NULL, related_name="+")
```
Marishka17: FYI, I partially fixed it in the [PR](https://github.com/cvat-ai/cvat/pull/9528/files#diff-3faaaf4d1d5f3ea9a59cb238ed8a3aae9568773c182916a2c14f62bc11999d54).

Why do you keep a backward relation for owned tasks but delete such a relation for assigned ones?
Additionally, shouldn't we define a consistent scheme for User <-> Job, User <-> Task, User <-> Project relations?
(e.g., now the User modal still has a backward relation for assigned jobs with the default name; backward relations for assigned and owned projects don't exist; backward relation for owned tasks exists)

SpecLad: > Why do you keep a backward relation for owned tasks but delete such a relation for assigned ones?

It wasn't clear to me whether a backwards relation for assigned tasks would be useful or what its name should be (I just knew that the current name wasn't good :slightly_smiling_face:), so I figured removing it would be a reasonable option.

> Additionally, shouldn't we define a consistent scheme for User <-> Job, User <-> Task, User <-> Project relations?

Yeah, I agree. I didn't even realize that `user.jobs` is currently the user's _assigned_ jobs. :face_with_spiral_eyes: 

Do you think it's useful to have backwards relations for assigned tasks, jobs, etc.? I see that in your PR you name them `assigned_X`, but you don't seem to actually be using them.

SpecLad: Okay, having thought about your PR some more, I like your solution for the `assignee` fields, so I'm going to pause this PR until you complete yours.

PR #9626, opened 2025-07-09
Removed extra transaction atomic

bsekachev:
### Motivation and context
I do not think we need atomic transaction during project/task/job export as it only reads data from the database.
At the same time, it leads to locks in database.
During export we use cursor to iterate over annotations in the database. While cursor is active, users can't update any Task in the project (e.g. updated date), or Data object (e.g. deleted_frames), etc, as corresponding tuples are locked via ForeignKeys 

image
image
image
 

### Checklist

### License

### DIFF
--- a/cvat/apps/dataset_manager/project.py
+++ b/cvat/apps/dataset_manager/project.py
@@ -39,14 +39,8 @@ def export_project(
     save_images: bool = False,
     temp_dir: str | None = None,
 ):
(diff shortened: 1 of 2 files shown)


bsekachev:
@Marishka17 @SpecLad thoughts?

bsekachev:
It seems the things are a little more difficult, as Job can be selected_for_update.
Need to think more on it.

PR #9629, opened 2025-07-09
Delegate EmailAddress creation to allauth

Marishka17:
### Motivation and context

This PR fixes an issue where a user cannot log in if they were invited to an organization using an email address with uppercase letters, but then manually registered using the same email address in lowercase.

This issue occurs when the CVAT server is configured with required email verification (ACCOUNT_EMAIL_VERIFICATION = "mandatory"). There is a [bug](https://github.com/iMerica/dj-rest-auth/blob/545fd1f4c9ae570216989a8f5a563dd938e038af/dj_rest_auth/serializers.py#L113) in dj_rest_auth when django allauth version is < 0.62 is used. The old django allauth versions do not "normalize" emails during registration, but use register-independent email filtration. That means if an EmailAddress (from django allauth) is created in another register than User.email -> dj_rest_auth raises an exception (email is not verified).

FYI, starting from version 0.62.0, django-allauth changed how it stores and manages email addresses. In the latest version, email addresses are converted to lowercase during registration, while a case-sensitive search is used during login.
Documentation: https://docs.allauth.org/en/latest/account/email.html
Related changes: https://github.com/pennersr/django-allauth/commit/4bb94eb96b51d152faa39d23c69440ec576b2720
### How has this been tested?

manually
### Checklist

### License


cvat/apps/iam/views.py
```diff
@@ -93,12 +93,15 @@ def post(self, request, *args, **kwargs):
...
-                return HttpResponseBadRequest("Unverified email")
+            if has_verified_email(user):
+                raise
+
+            send_email_confirmation(request, user)
+            # we cannot use redirect to ACCOUNT_EMAIL_VERIFICATION_SENT_REDIRECT_URL here
+            # because redirect will make a POST request and we'll get a 404 code
+            # (although in the browser request method will be displayed like GET)
+            return HttpResponseBadRequest("Unverified email")
+        # TODO: check why we need to handle exceptions here
```
SpecLad: FWIW, it would probably help to replace `pass` with a logging statement.

cvat/apps/organizations/serializers.py
```diff
@@ -126,12 +125,7 @@ def create(self, validated_data):
             user_email = membership_data["user"]["email"]
             user = User.objects.create_user(username=user_email, email=user_email)
             user.set_unusable_password()
-            # User.objects.create_user(...) normalizes passed email and user.email can be different from original user_email
-            email = EmailAddress.objects.create(
-                user=user, email=user.email, primary=True, verified=False
-            )
```
SpecLad, line 132: So if I'm understanding this change correctly, previously a dummy user would have an associated `EmailAddress` object with `verified=False`, while now a dummy user will have no associated `EmailAddress` at all.

That seems reasonable to me (and the implementation seems much simpler, which is nice). However, what happens with users for whom an `EmailAddress` has already been created but who have not yet accepted the invitation? Will they not be able to accept anymore? Can we fix this with a migration?

cvat/apps/organizations/serializers.py
```diff
@@ -126,12 +125,7 @@ def create(self, validated_data):
             user_email = membership_data["user"]["email"]
             user = User.objects.create_user(username=user_email, email=user_email)
             user.set_unusable_password()
-            # User.objects.create_user(...) normalizes passed email and user.email can be different from original user_email
```
SpecLad, line 129: Is this comment inaccurate? It seems like the issue you've described shouldn't happen if this is true.

Marishka17: This comment addressed another issue that could occur when an email **domain** contained uppercase letters and was normalized by django

SpecLad: Got it, thanks.

cvat/apps/organizations/serializers.py
Marishka17: > Will they not be able to accept anymore?

As I know, we do not change the data from the `EmailAddress` table when accepting invitations.
It should be done by django-allauth when a user validates an email by clicking on a link from a confirmation email (`GET /api/account-confirm-email/<key>`).

Marishka17: However, I agree that we need to add a migration to fix possible mismatches between `EmailAddress.email` and `User.email` to fix the login issue for already registered users.

SpecLad: OK, but let's suppose a person (who is not yet a user) is invited into an organization. A dummy user with an attached `EmailAddress` gets created for them. Then CVAT is upgraded to the new version. Then the person tries to register. IIUC, `get_dummy_user` will no longer return the previously created dummy user, since it now looks for users with no attached `EmailAddress`. So `RegisterSerializerEx` will create a new user, which is incorrect. Am I wrong?

Marishka17:
This PR also prohibits re-registration using the same email address. Before these changes, it was possible to create an account using some social login and then "update" account properties by following the basic registration process (of course, before validating email).

cvat/apps/iam/migrations/0002_delete_dummy_email_addresses_and_fix_case_related_mismatches.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+def delete_email_addresses_for_dummy_users(apps, schema_editor):
+    EmailAddress = apps.get_model("account", "EmailAddress")
+
+    # delete EmailAddress records for dummy users that were created by invitations
+    EmailAddress.objects.filter(verified=False).annotate(
+        social_accounts_count=Count("user__socialaccount")
+    ).filter(
+        social_accounts_count=0, user__password__startswith="!"  # nosec
```
SpecLad:
```suggestion
        ~Exists("user__socialaccount"), user__password__startswith="!"  # nosec
```

cvat/apps/iam/migrations/0002_delete_dummy_email_addresses_and_fix_case_related_mismatches.py
```diff
@@ -0,0 +1,53 @@
...
+        EmailAddress.objects.filter(verified=True)
+        .annotate(social_accounts_count=Count("user__socialaccount"))
+        .filter(social_accounts_count=0)
+        .select_related("user")
+    )
+    email_addresses_to_update = []
+
+    for email_address in email_addresses:
+        user_email = email_address.user.email
+        if email_address.email.lower() == user_email.lower() and email_address.email != user_email:
```
SpecLad: Seems like this could be part of the initial query (e.g. `.filter(email__iexact=F("user__email")).exclude(email=F("user__email")`).

In fact, you might even be able to do `.update(email=F("user__email"))`...

cvat/apps/iam/migrations/0002_delete_dummy_email_addresses_and_fix_case_related_mismatches.py
```diff
@@ -0,0 +1,53 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("iam", "0001_remove_business_group"),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            delete_email_addresses_for_dummy_users,
+            reverse_code=migrations.RunPython.noop,
```
SpecLad: It seems like there should be a reverse migration here.

cvat/apps/iam/tests/test_rest_api.py
```diff
@@ -130,3 +117,50 @@ def test_register_account_with_email_verification_mandatory(self):
...
+    def test_register_account_with_different_email_case_than_in_invitation(self):
+        """
+        Ensure a user can log in to the account after being invited to an organization
+        and then registering with the same email but in a different case.
+        """
+        org_slug = "testorg"
+        response = self._post_request(
+            "/api/organizations", self.admin, data={"slug": org_slug, "name": "Test organization"}
+        )
+        self.assertTrue(response.status_code == status.HTTP_201_CREATED)
```
SpecLad:
```suggestion
        self.assertEqual(response.status_code, status.HTTP_201_CREATED)
```

etc.

PR #9635, opened 2025-07-11
Fix test TestRQQueueWorking

archibald1418:
### Motivation and context

This test fails often on CI and during pre-merge.

This is likely due to the task queue being cleared out too quickly leading to user occasionally _clogging import queue with his tasks_  during the test. 

An increase of image size used in task creation attempts to increase task processing time to make the first task from the first user stay in the queue for longer and not allow the first user to clog the import queue, as it should be. 

Also. included `CVAT_ALLOW_STATIC_CACHE` to test name using `allure.dynamic.title` and `pytest.FixtureRequest`. This is needed to get different names in allure reports depending on the value used

### How has this been tested?

The test fail is reproducible locally and with just by launching the `test_queues.py` alone. The fix was confirmed by running this script:

```shell
for i in {1..10};
do
	ONE_RUNNING_JOB_IN_QUEUE_PER_USER=true pytest tests/python/rest_api/test_queues.py
	pytest tests/python --stop-services
	echo "Pause between runs..."
	sleep 3
done
```

The failure is not reproducible anymore at `HEAD`

NB: this is a temporary fix for CI, a rewrite will be needed to test the queue logic at unit tet level

### Checklist

### License

### DIFF
--- a/tests/python/rest_api/test_queues.py
+++ b/tests/python/rest_api/test_queues.py
@@ -65,7 +65,7 @@ class TestRQQueueWorking:
         s3_client = s3.make_client(bucket=cs_name)
         dataset_size = 100
 
-        img_content = generate_image_file(size=(4020, 2080)).getvalue()
+        img_content = generate_image_file(size=(1920, 1080)).getvalue()
 
         for i in range(dataset_size):
             filename = f"dataset/image_{i}.jpeg"


tests/python/rest_api/test_queues.py
```diff
@@ -58,7 +65,7 @@ def _create_task(idx: int, username: str) -> int:
         s3_client = s3.make_client(bucket=cs_name)
         dataset_size = 100
 
-        img_content = generate_image_file(size=(1920, 1080)).getvalue()
+        img_content = generate_image_file(size=(4020, 2080)).getvalue()
```
zhiltsov-max: This change doesn't guarantee the expected execution order.

archibald1418: No change would guarantee it, they are handled by a `ThreadExecutor`. It doesn't guarantee order

archibald1418: `asyncio` can do it but we can't use it

archibald1418:
One correlation I found: test fails when services are not stopped after the first run. So I would say this will work much better for now

archibald1418:
Closed in favour of #9657 which can replace the test in question

PR #9642, opened 2025-07-15
Added reposition logic into issues

klakhov:
### Motivation and context

This PR introduces a UX improvement for review mode. Previously, when the issue dialog appeared in the corner of the frame, users had to zoom out to view it. This update adds logic to automatically reposition the issue dialog upon opening, ensuring it remains fully visible.
Before:
issue-before
After:
issue-after

### Checklist

### License


bsekachev:
Probably the same should be implemented for create new issue dialog
image

bsekachev:
Does not seem to work when the image is rotated 

image

cvat-ui/src/components/annotation-page/review/issue-dialog.tsx
```diff
@@ -66,6 +73,38 @@ export default function IssueDialog(props: Props): JSX.Element {
...
+                newLeft = left - (overflow * scale);
+            }
+
+            setAdjustedTop(newTop);
+            setAdjustedLeft(newLeft);
+        } else {
+            setAdjustedTop(top);
+            setAdjustedLeft(left);
+        }
+    }, [top, left, scale, clickCoordinates, canvasRect]);
```
bsekachev:
```suggestion
    }, [ref.current, top, left, scale, clickCoordinates, canvasRect]);
```

cvat-ui/src/components/annotation-page/review/issue-dialog.tsx
```diff
@@ -66,6 +73,38 @@ export default function IssueDialog(props: Props): JSX.Element {
         }
     }, [resolved]);
 
+    useEffect(() => {
+        if (ref.current && clickCoordinates && canvasRect) {
```
bsekachev: let's add a small comment why we need this effect

cvat-ui/src/components/annotation-page/review/issues-aggregator.tsx
```diff
@@ -255,7 +265,12 @@ export default function IssueAggregatorComponent(): JSX.Element | null {
                     resolved={issueResolved}
                     highlight={highlight}
                     blur={blur}
-                    onClick={() => {
+                    onClick={(event) => {
+                        const clickCoords = {
+                            x: event.clientX,
+                            y: event.clientY,
+                        };
+                        setClickCoordinates(clickCoords);
```
bsekachev:
```suggestion
                        setClickCoordinates({
                            x: event.clientX,
                            y: event.clientY,
                        });
```

cvat-ui/src/components/annotation-page/review/issue-dialog.tsx
bsekachev: Should not we move this logic to issue aggregator? (to cover dialogs for existing issues and new issues by the same code)

klakhov: Moved the logic into separate hook

cvat-ui/src/components/annotation-page/review/issues-aggregator.tsx
klakhov: Removed the client click listener in favour of transformFromSVG

klakhov:
Fixed repositioning for rotated images, also added repostion for create new issue

bsekachev:
Canvas positioning is incorrect after creating issue on corners

https://github.com/user-attachments/assets/b350a04f-a446-4ca3-b285-8d7cad29ed7d

changelog.d/20250716_102252_klakhov_improve_issue_positioning.md
```diff
@@ -0,0 +1,4 @@
+### Added
+
+- Reposition of issue dialog if it's outside the visible area
```
bsekachev:
```suggestion
### Fixed

- Issue dialogs appear outside the visible area when the issue is located near the right or bottom edges of the frame
```

cvat-ui/src/components/annotation-page/review/use-dialog-positioning.ts
```diff
@@ -0,0 +1,81 @@
...
+    clientCoordinates,
+    canvasRect,
+}: UseDialogPositioningParams): UseDialogPositioningResult {
+    const [adjustedTop, setAdjustedTop] = useState(top);
+    const [adjustedLeft, setAdjustedLeft] = useState(left);
+
+    useEffect(() => {
+        if (ref.current && clientCoordinates && canvasRect) {
+            const dialogRect = ref.current.getBoundingClientRect();
+            const margin = 5;
```
bsekachev:
```suggestion
            const margin = 8;
```

cvat-ui/src/components/annotation-page/review/use-dialog-positioning.ts
```diff
@@ -0,0 +1,81 @@
...
+    clientCoordinates: [number, number];
+    canvasRect: DOMRect | null;
+}
+
+interface UseDialogPositioningResult {
+    adjustedTop: number;
+    adjustedLeft: number;
+}
+
+export function useDialogPositioning({
```
bsekachev, line 22: IssueAggregator creates new `clientCoordinates` each render and passes it to children.
Children (CreateIssueDialog/IssueDialog) accepts `clientCoordinates` and passes it to this hook. 
Each time it causes effect as `clientCoordinates` is dependency. 

So, the question is. Should it really be a hook or just a function is enough? From my perspective, it may be a simple function without useState/useEffect. That would be simpler and in current implementation we do not have any benefits of using hook

klakhov: It could just be a function, but the main issue is that before adjusting `top` and `left`, we need to know the dimensions of the issue dialog. That means we still need a hook that runs once `dialogRef.current` is not `null`, and then calls the coordinate transformation function. Also we will need a state for coordinates to change after reposition.

As a result, this hook would end up being duplicated in both `CreateIssueDialog` and `IssueDialog`.

klakhov:
> Canvas positioning is incorrect after creating issue on corners
> 
>  2025-07-31.10.50.21.AM.mov


It looks like this issue isn’t directly related to this PR—it also occurs on `develop`.

After investigating, I found that it's caused by the input’s auto-focus. The browser automatically scrolls to the focused element, so we need to manually apply focus after the repositioning.

PR #9657, opened 2025-07-20
Add unit tests for define_dependent_job function

orazvaliev:
### Checklist

### License

### DIFF
--- a/cvat/apps/engine/tests/test_rq.py
+++ b/cvat/apps/engine/tests/test_rq.py
@@ -1,0 +1,130 @@
+from datetime import datetime, timedelta
+import uuid
+from unittest import mock
+import unittest
+
+from rq.job import Job, Dependency
+from rq.registry import DeferredJobRegistry, StartedJobRegistry
+import django_rq
+from django_rq.queues import DjangoRQ
+from django.test import override_settings
+from django.conf import settings
+
+from cvat.apps.engine.rq import define_dependent_job, is_rq_job_owner
+from cvat.apps.engine.tests.utils import clear_rq_jobs
+
+
+DEFAULT_USER_ID = 1
+
+
+def dummy_task():
+    return 2 + 2
+
+
+def _create_test_job(
+    queue: DjangoRQ, job_id: str = None, user_id: int = DEFAULT_USER_ID, depends_on: list[Job] = None) -> Job:
+    job_id = job_id or str(uuid.uuid4())
+    job = queue.enqueue(
+        f=dummy_task,
+        args=(),
+        job_id=job_id,
+        meta={"user_id": user_id},
+        depends_on=depends_on
+    )
+    return job
+
+
+def _set_rq_async_mode(is_async) -> None:
+    for config in settings.RQ_QUEUES.values():
+        config["ASYNC"] = is_async
+
+
+class TestDefineDependentJob(unittest.TestCase):
+    """
+    Tests for automatic dependency resolution between RQ jobs.
+    The logic finds the latest enqueued job (optionally filtered by rq_id)
+    and defines a new job depending on it unless skipped by flags or self-dependency.
+    """
+    @classmethod
+    def setUpClass(cls) -> None:
+        _set_rq_async_mode(True)
+        cls.queue = django_rq.get_queue(settings.CVAT_QUEUES.IMPORT_DATA.value)
+
+    def setUp(self) -> None:
+        self.patcher = mock.patch('cvat.apps.engine.rq.is_rq_job_owner', return_value=True)
+        self.patcher.start()
+        clear_rq_jobs()
+
+    @classmethod
+    def tearDownClass(cls) -> None:
+        _set_rq_async_mode(False)
+
+    def tearDown(self) -> None:
+        self.addCleanup(self.patcher.stop)
+
+    def _define_dependent_job(
+        self, should_be_dependent: bool = True, rq_id: str | None = None) -> Dependency:
+        return define_dependent_job(
+            self.queue,
+            DEFAULT_USER_ID,
+            should_be_dependent=should_be_dependent,
+            rq_id=rq_id)
+
+    def test_define_dependency_on_latest_user_job(self) -> None:
+        """Ensures a new job correctly depends on the latest job if specific rq_id provided."""
+        job = _create_test_job(self.queue)
+        dependency = self._define_dependent_job(rq_id=str(uuid.uuid4()))
+        assert dependency is not None, "Dependent job not found."
+        assert len(dependency.dependencies) == 1
+        dep_job = dependency.dependencies[0]
+        assert dep_job == job
+
+    def test_no_dependency_when_should_be_dependent_is_false(self) -> None:
+        """Skips dependency if the flag should_be_dependent=False is used."""
+        job = _create_test_job(self.queue)
+        dependency = self._define_dependent_job(should_be_dependent=False)
+        assert dependency is None
+
+    def test_no_dependency_with_empty_queue(self):
+        dependency = self._define_dependent_job()
+        assert dependency is None
+
+    def test_no_dependency_with_empty_queue_and_rq_id(self):
+        dependency = self._define_dependent_job(rq_id=str(uuid.uuid4()))
+        assert dependency is None
+
+    def test_skip_self_dependent_job(self):
+        """Avoids creating dependency if a job is dependent on itself."""
+        first_job_id = str(uuid.uuid4())
+        first_job = _create_test_job(self.queue, job_id=first_job_id)
+        second_job = _create_test_job(
+            self.queue,
+            job_id=first_job_id,
+            depends_on=Dependency(jobs=[first_job_id], allow_failure=False))
+        dependency = self._define_dependent_job(rq_id=first_job_id)
+
+        assert dependency is None
+
+    def test_skip_dependency_for_same_job(self):
+        """Ignores dependencies if rq_id matches already enqueued job."""
+        job_id = str(uuid.uuid4())
+        job = _create_test_job(self.queue, job_id=job_id)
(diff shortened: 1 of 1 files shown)


cvat/apps/engine/tests/test_rq.py
```diff
@@ -0,0 +1,130 @@
...
+
+DEFAULT_USER_ID = 1
+
+
+def dummy_task():
+    return 2 + 2
+
+
+def _create_test_job(
+    queue: DjangoRQ, job_id: str = None, user_id: int = DEFAULT_USER_ID, depends_on: list[Job] = None) -> Job:
```
Marishka17:
```suggestion
    queue: DjangoRQ, job_id: str = None, user_id: int = DEFAULT_USER_ID, depends_on: list[Job] | None = None) -> Job:
```
Consider running code formatting

cvat/apps/engine/tests/test_rq.py
```diff
@@ -0,0 +1,130 @@
...
+    def test_select_latest_created_job(self):
+        """Selects the most recent job as dependency if multiple jobs are chained."""
+        total_jobs = 3
+        job_ids = [str(uuid.uuid4()) for _ in range(total_jobs)]
+        latest_job_id = job_ids[-1]
+        for i, job_id in enumerate(job_ids):
+            if i == 0:
+                _create_test_job(self.queue, job_id)
+            else:
+                _create_test_job(self.queue, job_id, depends_on=Dependency(jobs=[job_ids[i-1]]))
```
Marishka17:
```suggestion
            _create_test_job(self.queue, job_id, depends_on=Dependency(jobs=[job_ids[i - 1]], allow_failure=True) if i else None)
```

cvat/apps/engine/tests/test_rq.py
```diff
@@ -0,0 +1,130 @@
...
+    def test_no_dependency_with_empty_queue(self):
+        dependency = self._define_dependent_job()
+        assert dependency is None
+
+    def test_no_dependency_with_empty_queue_and_rq_id(self):
+        dependency = self._define_dependent_job(rq_id=str(uuid.uuid4()))
+        assert dependency is None
+
+    def test_skip_self_dependent_job(self):
+        """Avoids creating dependency if a job is dependent on itself."""
```
Marishka17: IMHO, `test_skip_dependency_for_same_job` should be enough to prevent cases when a job is dependent on itself. I suggest updating this test to check [this code](https://github.com/cvat-ai/cvat/blob/c759c5bdb4d7d06daaaa81374b0b1a2fb853a9a8/cvat/apps/engine/rq.py#L419) (there is a job `X2` that depends on `X1` (`X1` <- `X2`), somehow earlier it was possible to achieve such a situation, e.g., when the `X1` job was cancelled). We need to check that if we try to create/enqueue a new job with the `X1` id -> it will not depend on `X2` (preventing cyclic dependencies).

```python
 """Avoid creating a job dependency if the new job already exists in another job's dependencies."""
 first_job = _create_test_job(self.queue)
  _create_test_job(
      self.queue,
      depends_on=Dependency(jobs=[first_job.id], allow_failure=True),
  )
 first_job.delete()
 dependency = self._define_dependent_job(rq_id=first_job.id)

 assert dependency is None
```

FYI, we always use `allow_failure=True` when defining dependencies.

cvat/apps/engine/tests/test_rq.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+DEFAULT_USER_ID = 1
+
+
+def dummy_task():
+    return 2 + 2
+
+
+def _create_test_job(
```
Marishka17: Probably need to add a comment that all jobs created in these tests will always be queued, and it's expected.

cvat/apps/engine/tests/test_rq.py
```diff
@@ -0,0 +1,137 @@
...
+        assert dependency is None
+
+    def test_no_dependency_with_empty_queue_and_rq_id(self):
+        dependency = self._define_dependent_job(rq_id=str(uuid.uuid4()))
+        assert dependency is None
+
+    def test_skip_cyclic_dependencies(self):
+        """
+        Avoids creating dependency for the X1 if the X1 was as a dependency
+        for the X2 and then cancelled.
```
archibald1418:
```suggestion
        Avoids creating dependency for job X1 in case when
        X1 depended on X2 and then was cancelled
```

cvat/apps/engine/tests/test_rq.py
```diff
@@ -0,0 +1,137 @@
...
+        first_job = _enqueue_test_job(self.queue, job_id=first_job_id)
+        _enqueue_test_job(
+            self.queue,
+            job_id=second_job_id,
+            depends_on=Dependency(jobs=[first_job_id], allow_failure=True),
+        )
+        first_job.cancel()
+        dependency = self._define_dependent_job(rq_id=first_job_id)
+
+        assert dependency is None
```
archibald1418:
```suggestion
        assert dependency is None, "Job deps should not be cyclical"
```

cvat/apps/engine/tests/test_rq.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def test_skip_dependency_for_same_job(self):
+        """Ignores dependencies if rq_id matches already enqueued job."""
+        job_id = str(uuid.uuid4())
+        _enqueue_test_job(self.queue, job_id=job_id)
+        dependency = self._define_dependent_job(rq_id=job_id)
+
+        assert dependency is None
```
archibald1418:
```suggestion
        assert dependency is None, "Job deps should not be self-referential"
```

cvat/apps/engine/tests/test_rq.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def test_select_latest_created_job(self):
+        """Selects the most recent job as dependency if multiple jobs are chained."""
+        total_jobs = 3
```
archibald1418, line 128: How do we know that? Is there a constant that we can use?

cvat/apps/engine/tests/test_rq.py
```diff
@@ -0,0 +1,137 @@
...
+    return 2 + 2
+
+
+def _enqueue_test_job(
+    queue: DjangoRQ,
+    job_id: str | None = None,
+    user_id: int = DEFAULT_USER_ID,
+    depends_on: list[Job] | None = None,
+) -> Job:
+    job_id = job_id or str(uuid.uuid4())
```
Marishka17, line 35: Is there any reason to define such a custom job_id (here and in other places)? Why not just use the ids defined by rq?

cvat/apps/engine/tests/test_rq.py
```diff
@@ -0,0 +1,137 @@
...
+        """Selects the most recent job as dependency if multiple jobs are chained."""
+        total_jobs = 3
+        job_ids = [str(uuid.uuid4()) for _ in range(total_jobs)]
+        latest_job_id = job_ids[-1]
+        for i, job_id in enumerate(job_ids):
+            _enqueue_test_job(
+                self.queue,
+                job_id,
+                depends_on=Dependency(jobs=[job_ids[i - 1]], allow_failure=True) if i else None,
+            )
```
archibald1418, line 136: This should probably be a fixture too. If it was pytest, If `total_jobs` is a parameter you could even use `pytest_cases.parametrize` but this if for the times when we will move from unittest. 

For now, let's add `TODOs` to mark these two spots where we will have fixtures in the future

cvat/apps/engine/tests/test_rq.py
Marishka17: > in case when X1 depended on X2 and then was cancelled

equals X2 <- X1, but we are interested in X1 <- X2

archibald1418:
```suggestion
        Avoids creating dependency for job X1 in case when 
        X2 depended on X1 and then X1 got cancelled
```

cvat/apps/engine/tests/test_rq.py
archibald1418:
```suggestion
        TOTAL_JOBS = 3
```

cvat/apps/engine/tests/test_rq.py
orazvaliev: Some test scenarios require explicit job id setting because of dependency checks. e.g. `test_skip_cyclic_dependencies`.

cvat/apps/engine/tests/test_rq.py
orazvaliev: I'd change lots of things if i used pytest... Let's delay it until we plan a migration from unittest

cvat/apps/engine/tests/test_rq.py
orazvaliev: Uppercase variable name can be misleading if it's in a local scope of a function. Usually constants are defined on a module level. I didn't want to leave 3 as a magic number so i assign a name for it. But it's still just a short lived local scope

cvat/apps/engine/tests/test_rq.py
orazvaliev: Added a normal assert function on top of this check

cvat/apps/engine/tests/test_rq.py
orazvaliev: Renamed a function. Now it's clear that it adds a job to the queue

cvat/apps/engine/tests/test_rq.py
archibald1418: Ok makes sense

cvat/apps/engine/tests/test_rq.py
Marishka17: No, I don't think so, all tests can be updated to use ids of created jobs

FYI, we do not merge PRs without approvals

PR #9671, opened 2025-07-24
Support tags in detection functions

SpecLad:
### Motivation and context

This fills in another functionality gap between native AA functions and Nuclio functions.

Note that there is no server counterpart to this PR, because the server-side agent code already works correctly with function results containing tags.

### How has this been tested?

Unit tests.

### Checklist

### License

### DIFF
--- a/cvat-cli/src/cvat_cli/_internal/agent.py
+++ b/cvat-cli/src/cvat_cli/_internal/agent.py
@@ -176,7 +176,7 @@ def _worker_job_get_function_spec():
 
 def _worker_job_detect(
     context: _DetectionFunctionContextImpl, image: PIL.Image.Image
-) -> list[cvataa.TagOrShape]:
+) -> list[cvataa.DetectionAnnotation]:
     return _current_function.detect(context, image)
 
 
@@ -820,11 +820,11 @@ class _Agent:
 
         for sample_index, sample in enumerate(ds.samples):
             context = self._create_detection_function_context(ar_params, sample.frame_name)
-            tags_and_shapes = self._executor.result(
+            annotations = self._executor.result(
                 self._executor.submit(_worker_job_detect, context, sample.media.load_image())
             )
 
-            tags, shapes = mapper.validate_and_remap(tags_and_shapes, sample.frame_index)
+            tags, shapes = mapper.validate_and_remap(annotations, sample.frame_index)
             all_annotations.tags.extend(tags)
             all_annotations.shapes.extend(shapes)
 
@@ -847,11 +847,11 @@ class _Agent:
 
         context = self._create_detection_function_context(ar_params, sample.frame_name)
 
-        tags_and_shapes = self._executor.result(
+        annotations = self._executor.result(
             self._executor.submit(_worker_job_detect, context, sample.media.load_image())
         )
 
-        tags, shapes = mapper.validate_and_remap(tags_and_shapes, sample.frame_index)
+        tags, shapes = mapper.validate_and_remap(annotations, sample.frame_index)
         return {"annotations": models.PatchedLabeledDataRequest(tags=tags, shapes=shapes)}
 
     def _calculate_result_for_tracking_ar(self, ar_id: str, ar_params) -> dict[str, Any]:
--- a/cvat-cli/src/cvat_cli/_internal/agent.py
+++ b/cvat-cli/src/cvat_cli/_internal/agent.py
@@ -176,7 +176,7 @@ def _worker_job_get_function_spec():
 
 def _worker_job_detect(
     context: _DetectionFunctionContextImpl, image: PIL.Image.Image
-) -> list[models.LabeledShapeRequest]:
+) -> list[cvataa.TagOrShape]:
     return _current_function.detect(context, image)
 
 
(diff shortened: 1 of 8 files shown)


cvat-sdk/cvat_sdk/auto_annotation/driver.py
```diff
@@ -340,15 +345,15 @@ def _remap_attribute(
 
     def _remap_attributes(
         self,
-        shape: Union[models.LabeledShapeRequest, models.SubLabeledShapeRequest],
+        tors: Union[TagOrShape, models.SubLabeledShapeRequest],
```
zhiltsov-max:
```suggestion
        annotation: Union[TagOrShape, models.SubLabeledShapeRequest],
```

tests/python/sdk/test_auto_annotation.py
```diff
@@ -1038,6 +1089,29 @@ def test_attribute_val_unsuitable_for_spec(self):
...
+            super().__init__()
+            self._label_id = label_id
+
+        def forward(self, images: torch.Tensor) -> torch.Tensor:
+            assert isinstance(images, torch.Tensor)
+
+            probs = [0.0] * 1000
+            # Since the function does a softmax operation over the output,
+            # we can't test edge cases by making the final probability exactly 0.75.
+            # With this value the final probability is ~0.77.
```
zhiltsov-max, line 1103: I don't really see a point in trying to hit the exact threshold value here. These are floating point values that can be arbitrary. Just saying that this value produces something between 0.75 and 0.9 would be fine.

cvat-sdk/cvat_sdk/auto_annotation/functions/torchvision_classification.py
```diff
@@ -0,0 +1,31 @@
...
+
+import PIL.Image
+
+import cvat_sdk.auto_annotation as cvataa
+import cvat_sdk.models as models
+
+from ._torchvision import TorchvisionFunction
+
+
+class _TorchvisionDetectionFunction(TorchvisionFunction):
```
zhiltsov-max:
```suggestion
class _TorchvisionClassificationFunction(TorchvisionFunction):
```
?

cvat-cli/src/cvat_cli/_internal/agent.py
```diff
@@ -802,15 +802,16 @@ def _calculate_result_for_annotate_task_ar(self, ar_id: str, ar_params) -> dict[
 
         mapper = self._create_annotation_mapper_for_detection_ar(ar_params, ds.labels)
 
-        all_annotations = models.PatchedLabeledDataRequest(shapes=[])
+        all_annotations = models.PatchedLabeledDataRequest(tags=[], shapes=[])
 
         for sample_index, sample in enumerate(ds.samples):
             context = self._create_detection_function_context(ar_params, sample.frame_name)
-            shapes = self._executor.result(
+            tags_and_shapes = self._executor.result(
```
zhiltsov-max: Probably need to update the return value type in the line 165

cvat-sdk/cvat_sdk/auto_annotation/driver.py
SpecLad: Are you proposing this for everywhere? Or just here?

zhiltsov-max: Ideally to replace tors with something more meaningful everywhere.

zhiltsov-max: I can suggest DetectionAnnotation or DetectorAnnotation, since we call such functions Detectors in CVAT.

SpecLad: Hmm, maybe just `Detection` then?

tests/python/sdk/test_auto_annotation.py
SpecLad: > I don't really see a point in trying to hit the exact threshold value here.

The point would be to test the behavior when the confidence exactly equals the threshold. The tests for the other torchvision functions do this.

> Just saying that this value produces something between 0.75 and 0.9 would be fine.

Sure, but isn't giving the approximate number also fine?

zhiltsov-max: The code is fine, my comment is only about the line

> we can't test edge cases by making the final probability exactly 0.75.

cvat-sdk/cvat_sdk/auto_annotation/driver.py
zhiltsov-max: Tags are not really "detections" in the computer vision sense, so I'd vote against this variant. It might be ok in the current CVAT terminology, but it looks a little bit broken to me at the moment.

SpecLad: Yeah, I agree. I went with `DetectionAnnotation`, which seems to be most consistent with existing code.

cvat-sdk/cvat_sdk/auto_annotation/functions/torchvision_classification.py
SpecLad: Right, thanks. Fixed.

cvat-cli/src/cvat_cli/_internal/agent.py
SpecLad: Updated.

tests/python/sdk/test_auto_annotation.py
SpecLad: Okay. So do you think I should change something?

zhiltsov-max (approved):

tests/python/sdk/test_auto_annotation.py
zhiltsov-max: I thought to remove the part about the edge cases, because I was confused by the wording, but I don't think it's a stopper anyway.

PR #9686, opened 2025-07-30
Added user deletion manager that removes user together with all data

bsekachev:
### Motivation and context
- The manager supports validation (e.g. in community version user should not have organizations with other members, in SaaS user also should not have any active subscriptions)
- Added command: `python manage.py deleteuser <user_id>` 

### Checklist

### License

### DIFF
--- a/cvat/apps/engine/management/commands/deleteuser.py
+++ b/cvat/apps/engine/management/commands/deleteuser.py
@@ -27,6 +27,10 @@ class Command(BaseCommand):
 
         manager = UserDeletionManager(user)
         try:
-            manager.delete_user_with_cleanup(dry_run=dry_run)
+            deleted_resources = manager.delete_user_with_cleanup(dry_run=dry_run)
+            for resource_type in deleted_resources:
+                if deleted_resources[resource_type]:
+                    for resource_id in deleted_resources[resource_type]:
+                        self.stdout.write(f"Deleted {resource_type}: #{resource_id}")
         except ValidationError as e:
             raise CommandError(e.detail)
--- a/cvat/apps/engine/management/commands/deleteuser.py
+++ b/cvat/apps/engine/management/commands/deleteuser.py
@@ -29,8 +29,7 @@ class Command(BaseCommand):
         try:
             deleted_resources = manager.delete_user_with_cleanup(dry_run=dry_run)
             for resource_type in deleted_resources:
-                if deleted_resources[resource_type]:
-                    for resource_id in deleted_resources[resource_type]:
-                        self.stdout.write(f"Deleted {resource_type}: #{resource_id}")
+                for resource_id in deleted_resources[resource_type]:
+                    self.stdout.write(f"Deleted {resource_type}: #{resource_id}")
         except ValidationError as e:
             raise CommandError(e.detail)
--- a/cvat/apps/engine/management/commands/deleteuser.py
+++ b/cvat/apps/engine/management/commands/deleteuser.py
@@ -1,0 +1,33 @@
+# Copyright (C) CVAT.ai Corporation
+#
+# SPDX-License-Identifier: MIT
+
+from django.core.management.base import BaseCommand, CommandError
+from django.utils.encoding import force_str
+from rest_framework.exceptions import ValidationError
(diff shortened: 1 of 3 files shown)


cvat/apps/engine/management/commands/deleteuser.py
```diff
@@ -0,0 +1,35 @@
...
+    def add_arguments(self, parser):
+        parser.add_argument("user_id", type=int, help="ID of the user to delete")
+        parser.add_argument("--dry-run", action="store_true")
+
+    def handle(self, *args, **options):
+        user_id = options["user_id"]
+        dry_run = options["dry_run"]
+
+        try:
+            user = User.objects.get(pk=user_id)
```
zhiltsov-max:
```suggestion
            user = User.objects.select_for_update().get(pk=user_id)
```

\+ transaction.atomic to avoid [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) for validations

cvat/apps/engine/user_deletion.py
```diff
@@ -0,0 +1,122 @@
...
+    def validate(self, user: User) -> None:
+        max_members = (
+            Organization.objects.filter(owner=user)
+            .annotate(members_count=Count("members"))
+            .aggregate(max_members=Coalesce(Max("members_count"), 0))["max_members"]
+        )
+
+        if max_members > 1:
+            raise ValidationError(
+                "Cannot delete a user who is the owner of an organization with multiple members."
```
zhiltsov-max: Consider adding the org id in the error message, it can help if there are several owned orgs.

cvat/apps/engine/user_deletion.py
```diff
@@ -0,0 +1,122 @@
...
+    task: list[int]
+    cloud_storage: list[int]
+
+
+@dataclass
+class UserDeletionManager:
+    user: User
+
+    @transaction.atomic
+    def delete_user_with_cleanup(self, dry_run=True) -> _DeletedResources:
```
zhiltsov-max:
```suggestion
    def delete_user_with_cleanup(self, dry_run: bool = True) -> _DeletedResources:
```

cvat/apps/engine/user_deletion.py
```diff
@@ -0,0 +1,122 @@
...
+            "task": [],
+            "cloud_storage": [],
+        }
+
+        for ValidatorClass in _USER_DELETION_VALIDATORS:
+            ValidatorClass().validate(self.user)
+
+        db_orgs = Organization.objects.filter(owner=self.user).select_for_update()
+
+        def filter_by_owner_and_org(queryset: QuerySet) -> QuerySet:
```
zhiltsov-max:
```suggestion
from typing import TypeVar
...
_T = TypeVar("_T", bound=django.db.models.Model)
...

        def filter_by_owner_and_org(queryset: QuerySet[_T]) -> QuerySet[_T]:
```

Can also use `from cvat.apps.engine.model_utils import _ModelT`

cvat/apps/engine/user_deletion.py
```diff
@@ -0,0 +1,122 @@
...
+            ("project", db_projects),
+            ("task", db_tasks),
+            ("cloud_storage", db_cloud_storages),
+        ):
+            for db_resource in db_resources:
+                deleted_resources[resource_type].append(db_resource.id)
+
+        if not dry_run:
+            db_orgs.delete()
+            db_cloud_storages.delete()
```
zhiltsov-max: Given that this code is going to be used rarely and doesn't have any tests, it might be better to just use the default for loop for object deletion in all these cases.

cvat/apps/engine/management/commands/deleteuser.py
bsekachev: moved user fetching to the deletion function with select_for_update

cvat/apps/engine/user_deletion.py
bsekachev: Updated,
Anyway we will definitely need some tests for that.

zhiltsov-max (approved):

PR #9688, opened 2025-07-31
retry loop for yarn install during e2e tests

archibald1418:
Workaround in case when npm registry returns 500 when downloading packages for `e2e_testing`

### Checklist

### License


archibald1418:
\check

archibald1418:
`/check` launches outdated `.yml` files, so I can only trigger the new pipeline with pushes

.github/workflows/eslint.yml
```diff
@@ -11,8 +11,9 @@ jobs:
 
       - name: Install dependencies
         run: |
-          yarn install --frozen-lockfile
-          (cd tests && yarn install --frozen-lockfile)
+          corepack enable yarn && yarn set version berry
```
klakhov: Do we really need to run `yarn set version berry` every time we install dependencies?
As I understand it, this command is meant to migrate Yarn to a new version. But since we already have `"packageManager": "yarn@4.9.2"` in `package.json`, Corepack should automatically use the specified version.

According to [[this thread](https://github.com/yarnpkg/berry/issues/5823)], running this command will upgrade Yarn to the latest stable version. My concern is that doing this automatically could unintentionally upgrade Yarn and potentially break something.

archibald1418: Sure, I'll add a connected PR in the private repo

archibald1418: You are right, I'll keep only the pinned version

archibald1418: I will have to merge this PR though, then enterprise, then app.cvat.ai

klakhov: Why can’t we create the PRs in advance? I believe we should merge them almost simultaneously — otherwise, bumping submodules to `develop` will cause the CI to fail.

cvat-core/README.md
```diff
@@ -10,7 +10,9 @@ It contains the core logic of the Computer Vision Annotation Tool.
 - Dependencies installation
 
 ```bash
-yarn install --frozen-lockfile
+corepack enable yarn
+yarn set version berry
```
klakhov: I am not sure we need this command here as we already have specified version in package.json

klakhov (commented):
Let’s think more carefully about our Yarn setup and documentation.

Previously, Yarn was installed via npm using:

```
sudo npm install --global yarn
```

This is still how it’s described in our documentation. However, with modern Yarn versions (Berry and later), Yarn is now managed through **Corepack**, and the installation process has changed.

Even though Corepack is bundled with Node.js, the [[official Yarn documentation](https://yarnpkg.com/getting-started/install)](https://yarnpkg.com/getting-started/install) recommends running:

```
npm install -g corepack
```

just in case — and that makes sense, especially because [Corepack will be removed starting from Node 25](https://nodejs.org/dist/latest/docs/api/corepack.html). So we’ll need to **explicitly install it** as part of our development environment setup.

---

For users like me (and likely others) who previously installed Yarn globally via npm, there's an extra migration step needed.

Our current documentation includes:

```
corepack enable yarn
yarn set version berry
yarn install --immutable
```

But `corepack enable yarn` **does not override** the older global Yarn version installed via npm. As a result, when I run `yarn install`, it still uses my old Yarn 1.22.22 version:

> Yarn version mismatch

So after running the current setup commands, I ended up with **two Yarn installations**:

> Multiple Yarn versions

To actually make the `yarn` command use the correct version managed by Corepack, I had to run:

```
npm uninstall -g yarn
hash -r
```

---
So I have two conclusions here:

1. **We need to add a "Migration Guide"** for users who previously installed Yarn via npm. Without this, users are likely to run into confusing issues during setup.

2. **We should update the installation guide** so that all new installations are set up with the Corepack-managed version of Yarn from the beginning.

cvat-core/README.md
archibald1418: Ok, this makes a lot of sense. I would then update the build cmds and add a section to the READMEs on migrating from classic Yarn to modern Yarn on your local machine

.github/workflows/eslint.yml
archibald1418: OK I'll reference them in this PR when they are ready

archibald1418:
PRs in order:

1. this one
1. https://github.com/cvat-ai/cvat_enterprise/pull/395 - add cvat to enterprise, merge to develop
1. https://github.com/cvat-ai/app.cvat.ai/pull/625 - add enterprise to app.cvat.ai, merge to develop

cvat-core/README.md
```diff
@@ -5,14 +5,34 @@
...
+hash -r
+
+# Ensure corepack is installed
+sudo npm install -g corepack
+
+# Enable new yarn
+corepack enable yarn
+yarn --version # should show 4.9.2
+```
+
```
klakhov: This section is now duplicated in several places, and I don’t think that’s necessary.

Instead, let’s move the **Yarn modern migration** instructions to the **Development Environment** docs page, and simply add links to that guide wherever it’s relevant. That way, we avoid duplication and keep everything easier to maintain.

cvat-core/README.md
```diff
@@ -5,14 +5,34 @@
 This CVAT module is a client-side JavaScript library for management of objects, frames, logs, etc.
 It contains the core logic of the Computer Vision Annotation Tool.
 
+### Migration to Yarn Modern
+
+We have recently updated our Yarn version from Classic to Modern.
+If you are still using CVAT with Yarn 1.22 you need to first migrate to Yarn 4.9.2:
```
klakhov: Let’s avoid specifying exact versions. Users might have a different version than 1.22, and we may also update from Yarn 4.9.2 in the future.

It’s better to follow Yarn’s own terminology: refer to **“Yarn Classic (1.x)”** for older versions, and **“Yarn Modern”** for newer ones.

cvat-core/README.md
```diff
@@ -5,14 +5,34 @@
...
+
+```bash
+# Remove old yarn
+sudo npm uninstall -g yarn
+hash -r
```
klakhov:
```suggestion
```
Lets remove this, it was some kind of my local problem with hash. Runing uninstall -> install should be enough

cvat-core/README.md
```diff
@@ -5,14 +5,34 @@
...
+
+# Ensure corepack is installed
+sudo npm install -g corepack
+
+# Enable new yarn
+corepack enable yarn
+yarn --version # should show 4.9.2
```
klakhov:
```suggestion
yarn --version # should show 4.X
```

cvat-core/README.md
klakhov: On the other hand, I don’t think we need to include links to the migration guide throughout the docs. Let’s just keep it in one place—in the *Development Environment* guide—and assume the use of Yarn Modern everywhere else.

site/content/en/docs/contributing/development-environment.md
```diff
@@ -16,10 +16,9 @@ description: 'Installing a development environment for different operating syste
   ```
 
   ```bash
-  # Install Node.js 20 and yarn
+  # Install Node.js 20
   curl -fsSL https://deb.nodesource.com/setup_20.x | sudo bash -
   sudo apt-get install -y nodejs
-  sudo npm install --global yarn
   ```
```
klakhov: Why are we removing yarn installation from here? I suppose we need to put here instructions to activate yarn from corepack

archibald1418: Added it back

Dockerfile.ui
```diff
@@ -15,7 +16,8 @@ COPY cvat-data/package.json /tmp/cvat-data/
 
 # Install common dependencies
 WORKDIR /tmp/
-RUN DISABLE_HUSKY=1 yarn --frozen-lockfile
+RUN npm install -g corepack && corepack enable yarn
```
SpecLad: IIRC, installing corepack via NPM enables it automatically, so there's no need for a `corepack enable` command.

It would also make sense to move this command higher in the Dockerfile, so that we don't reinstall corepack every time a `package.json` file changes.

archibald1418: removed installation and moved to higher layers

klakhov: I suppose we shoud not remove installation, but instead we should remove `corepack enable`

site/content/en/docs/contributing/development-environment.md
```diff
@@ -16,10 +16,10 @@ description: 'Installing a development environment for different operating syste
...
 
   ```bash
-  # Install Node.js 20 and yarn
+  # Install Node.js 20 and enable yarn
   curl -fsSL https://deb.nodesource.com/setup_20.x | sudo bash -
   sudo apt-get install -y nodejs
-  sudo npm install --global yarn
+  corepack enable yarn
```
klakhov: Missing `sudo npm install -g corepack`?

archibald1418: added and removed enable yarn since it's enabled automatically

klakhov:
From my perspective PR looks good. @SpecLad, do you have any more comments regarding the workflows?

SpecLad:
> From my perspective PR looks good. @SpecLad, do you have any more comments regarding the workflows?

No, I'm good.

PR #9689, opened 2025-07-31
Support clickhouse migrations

zhiltsov-max:
### Motivation and context

- Added a way to add multiple migration files for the Clickhouse instance

### Checklist

### License

### DIFF
--- a/components/analytics/clickhouse/000-initial.cmf
+++ b/components/analytics/clickhouse/000-initial.cmf
@@ -1,0 +1,36 @@
+#!/bin/bash
+
+# Copyright (C) CVAT.ai Corporation
+#
+# SPDX-License-Identifier: MIT
+
+set -e
+
+clickhouse-client --query "CREATE DATABASE IF NOT EXISTS ${CLICKHOUSE_DB};"
+
+echo "
+CREATE TABLE IF NOT EXISTS ${CLICKHOUSE_DB}.events
+(
+    \`scope\` String NOT NULL,
+    \`obj_name\` String NULL,
+    \`obj_id\` UInt64 NULL,
+    \`obj_val\` String NULL,
+    \`source\` String NOT NULL,
+    \`timestamp\` DateTime64(3, 'Etc/UTC') NOT NULL,
+    \`count\` UInt16 NULL,
+    \`duration\` UInt32 DEFAULT toUInt32(0),
+    \`project_id\` UInt64 NULL,
+    \`task_id\` UInt64 NULL,
+    \`job_id\` UInt64 NULL,
+    \`user_id\` UInt64 NULL,
+    \`user_name\` String NULL,
+    \`user_email\` String NULL,
+    \`org_id\` UInt64 NULL,
+    \`org_slug\` String NULL,
+    \`payload\` String NULL
+)
+ENGINE = MergeTree
+PARTITION BY toYYYYMM(timestamp)
+ORDER BY (timestamp)
+SETTINGS index_granularity = 8192
+;" | clickhouse-client
--- a/components/analytics/clickhouse/README.md
+++ b/components/analytics/clickhouse/README.md
@@ -1,0 +1,40 @@
+<!--
+ Copyright (C) CVAT.ai Corporation
+
+ SPDX-License-Identifier: MIT
+-->
+
+# Clickhouse DB migrations
+
+This directory contains migrations for the Clickhouse DB used by CVAT.
+
+## Implementation details
+
+The Clickhouse documentation explains options to customize DB loading here:
+https://clickhouse.com/docs/install/docker#how-to-extend-image
+
+Migrations are mounted into the container into the `/docker-entrypoint-initdb.d` directory.
+There, only the main script `init.sh` is an executable shell script. The migration files
+are also shell scripts inside, but they have the `.cmf` extension, which stands
+for "clickhouse migration file". We use a custom extension to avoid automatic execution
+of these files by Clickhouse.
+
+**Please note that Clickhouse only runs these files if the DB is not initialized.**
+
+## Adding a migration
+
+To add a migration, create a new `NNN-<custom name>.cmf` file in this directory.
+
+Recommendations on the migrations:
+- Avoid heavy data migrations as much as possible. In most cases changing only the table
+  columns should be enough and you won't need a data migration.
+- Typically the migration should include "IF NOT EXISTS" at some point to allow idempotency
+  and avoid failing the migration on an existing DB.
+
+## How to run
+
+If you need to run migrations manually, use the following command:
+
+```bash
+docker exec -it cvat_clickhouse bash /docker-entrypoint-initdb.d/init.sh
+```
--- a/components/analytics/clickhouse/init.sh
+++ b/components/analytics/clickhouse/init.sh
@@ -1,32 +1,13 @@
 #!/bin/bash
 
+set -e
+
 CLICKHOUSE_DB="${CLICKHOUSE_DB:-cvat}";
 
-clickhouse-client --query "CREATE DATABASE IF NOT EXISTS ${CLICKHOUSE_DB};"
+# Run scripts in the directory sequentially
+# Alpine/Busybox's find uses posix basic regex syntax
+INIT_SCRIPTS_DIR="$(dirname "$0")"
+find "${INIT_SCRIPTS_DIR}/" -regex ".*/[0-9][0-9][0-9]-[^/]*\\.cmf" -type f -print0 | \
+    sort -z -n -t "-" -k 1 | \
+    xargs -0 -I {} bash -c "echo \"Running migration file {}\" && bash \"{}\""
 
-echo "
-CREATE TABLE IF NOT EXISTS ${CLICKHOUSE_DB}.events
-(
-    \`scope\` String NOT NULL,
-    \`obj_name\` String NULL,
-    \`obj_id\` UInt64 NULL,
-    \`obj_val\` String NULL,
-    \`source\` String NOT NULL,
-    \`timestamp\` DateTime64(3, 'Etc/UTC') NOT NULL,
-    \`count\` UInt16 NULL,
-    \`duration\` UInt32 DEFAULT toUInt32(0),
-    \`project_id\` UInt64 NULL,
-    \`task_id\` UInt64 NULL,
-    \`job_id\` UInt64 NULL,
-    \`user_id\` UInt64 NULL,
-    \`user_name\` String NULL,
-    \`user_email\` String NULL,
-    \`org_id\` UInt64 NULL,
-    \`org_slug\` String NULL,
-    \`payload\` String NULL
-)
-ENGINE = MergeTree
-PARTITION BY toYYYYMM(timestamp)
-ORDER BY (timestamp)
-SETTINGS index_granularity = 8192
-;" | clickhouse-client
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -25,6 +25,10 @@ x-backend-deps: &backend-deps
     condition: service_started
   cvat_db:
     condition: service_started
+  cvat_clickhouse:
+    condition: service_started
+  cvat_clickhouse_init:
+    condition: service_completed_successfully
 
 x-clickhouse-env: &clickhouse-env
   CLICKHOUSE_HOST: clickhouse
@@ -337,8 +341,44 @@ services:
         aliases:
           - clickhouse
     volumes:
-      - ./components/analytics/clickhouse/init.sh:/docker-entrypoint-initdb.d/init.sh:ro
       - cvat_events_db:/var/lib/clickhouse/
+    healthcheck:
+      test: ["CMD", "clickhouse-client", "--query", "SELECT 1"]
+      interval: 3s
+      timeout: 300s
+      retries: 3
+
+  cvat_clickhouse_init:
+    container_name: cvat_clickhouse_init
+    image: clickhouse/clickhouse-server:23.11-alpine
+    restart: no
+    environment:
+      <<: *clickhouse-env
+    depends_on:
+      cvat_clickhouse:
+        condition: service_healthy
+    networks:
+      cvat:
+        aliases:
+          - clickhouse
+    volumes:
+      - ./components/analytics/clickhouse/:/docker-entrypoint-initdb.d/:ro
+      - cvat_events_db:/var/lib/clickhouse/
+    entrypoint:
+      # https://github.com/ClickHouse/ClickHouse/blob/master/programs/client/clickhouse-client.xml
+      - /bin/bash
+      - -c
+      - |
+        mkdir -p /etc/clickhouse-client
+        cat << EOF > /etc/clickhouse-client/config.xml
+        <config>
+          <user>$${CLICKHOUSE_USER}</user>
+          <password>$${CLICKHOUSE_PASSWORD}</password>
+          <host>$${CLICKHOUSE_HOST}</host>
+          <port>9000</port><!-- https://clickhouse.com/docs/guides/sre/network-ports -->
+        </config>
+        EOF
+        bash "/docker-entrypoint-initdb.d/init.sh"
 
   cvat_vector:
     container_name: cvat_vector
--- a/helm-chart/values.yaml
+++ b/helm-chart/values.yaml
@@ -408,7 +408,7 @@ clickhouse:
   shards: 1
   replicaCount: 1
   extraEnvVarsSecret: cvat-analytics-secret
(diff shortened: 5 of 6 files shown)


docker-compose.yml
```diff
@@ -329,16 +333,51 @@ services:
...
+    restart: "no"
+    environment:
+      <<: *clickhouse-env
+    depends_on:
+      cvat_clickhouse:
+        condition: service_healthy
+    networks:
+      cvat:
+        aliases:
+          - clickhouse
```
azhavoro:
```suggestion
      - cvat:
```

It seems that we don't need define alias at all, since the service is not supposed to accept requests.

helm-chart/values.yaml
```diff
@@ -408,7 +408,7 @@ clickhouse:
   shards: 1
   replicaCount: 1
   extraEnvVarsSecret: cvat-analytics-secret
-  initdbScriptsSecret: cvat-clickhouse-init
+  startdbScriptsSecret: cvat-clickhouse-start
```
azhavoro: It seems that a separate job needs to be added for initialization in case of external clichouse use. like this https://github.com/cvat-ai/cvat/blob/develop/helm-chart/templates/cvat_backend/initializer/job.yml

helm-chart/templates/analytics/clickhouse/initializer/job.yml
```diff
@@ -0,0 +1,56 @@
...
+      labels:
+        {{- include "cvat.labels" . | nindent 8 }}
+        app: cvat-app
+        tier: clickhouse
+        component: initializer
+    spec:
+      containers:
+        - name: clickhouse
+          image: {{ .Values.clickhouse.image.registry }}/{{ .Values.clickhouse.image.repository }}:{{ .Values.clickhouse.image.tag }}
+          imagePullPolicy: Never
```
azhavoro:
```suggestion
          imagePullPolicy: IfNotPresent
```

azhavoro: Otherwise, lgtm

zhiltsov-max: I had an idea that if the default clickhouse is used, it will be downloaded by the time this job starts. If a custom instance is used, it's version might require adjusting, so it's better to fail here to highlight the configuration error. What do you think?

zhiltsov-max: OTOH, clickhouse-client versions are likely to be compatible with a wide range of clickhouse versions, so it may be overcautious.

components/analytics/clickhouse/README.md
```diff
@@ -0,0 +1,40 @@
...
+The Clickhouse documentation explains options to customize DB loading here:
+<https://clickhouse.com/docs/install/docker#how-to-extend-image>
+
+Migrations are mounted into the container into the `/docker-entrypoint-initdb.d` directory.
+There, only the main script `init.sh` is an executable shell script. The migration files
+are also shell scripts inside, but they have the `.cmf` extension, which stands
+for "clickhouse migration file". We use a custom extension to avoid automatic execution
+of these files by Clickhouse.
+
+**Please note that Clickhouse only runs these files if the DB is not initialized.**
```
SpecLad: Doesn't this undermine the entire mechanism? The whole purpose of migrations is to change the DB after it's already initialized.

components/analytics/clickhouse/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Adding a migration
+
+To add a migration, create a new `NNN-<custom name>.cmf` file in this directory.
+
+Recommendations on the migrations:
+- Avoid heavy data migrations as much as possible. In most cases changing only the table
+  columns should be enough and you won't need a data migration.
+- Typically the migration should include "IF NOT EXISTS" at some point to allow idempotency
+  and avoid failing the migration on an existing DB.
```
SpecLad: Can you show an example of what you expect a typical migration to look like? E.g. adding a column. It doesn't seem to me that `IF NOT EXISTS` would be a viable solution in most cases.

components/analytics/clickhouse/README.md
zhiltsov-max: Clickhouse only runs init.d/ scripts if the DB is not initialized yet. They only run once to initialize the DB and that's it. This PR changes it so that there's additional unconditional run of the migrations in an extra job, that is migrations will be applied even if the DB was initialized already. All the migrations we have at the moment are idempotent and only change schema, so it's safe to run them several times. It can be extended to the model with recorded migration names in the DB later.

components/analytics/clickhouse/README.md
zhiltsov-max: The example for adding a column with IF NOT EXISTS is in https://github.com/cvat-ai/cvat/pull/9680. Basically, we need to be able to run a migration several times without changing the result and without failing.

components/analytics/clickhouse/000-initial.cmf
```diff
@@ -0,0 +1,36 @@
...
+    \`user_email\` String NULL,
+    \`org_id\` UInt64 NULL,
+    \`org_slug\` String NULL,
+    \`payload\` String NULL
+)
+ENGINE = MergeTree
+PARTITION BY toYYYYMM(timestamp)
+ORDER BY (timestamp)
+SETTINGS index_granularity = 8192
+;" | clickhouse-client
```
SpecLad: I would suggest to maybe do:

```bash
clickhouse-client --database="$CLICKHOUSE_DB" <<'EOF'
CREATE TABLE IF NOT EXISTS events
(
    scope String NOT NULL,
    ...
EOF
```

By moving the database name to the dedicated option, we no longer need to interpolate data into the SQL query, and thus we don't need to escape special characters (although in this case it doesn't seem like the special characters are needed to begin with).

components/analytics/clickhouse/init.sh
```diff
@@ -1,32 +1,13 @@
...
+
 CLICKHOUSE_DB="${CLICKHOUSE_DB:-cvat}";
 
-clickhouse-client --query "CREATE DATABASE IF NOT EXISTS ${CLICKHOUSE_DB};"
+# Run scripts in the directory sequentially
+# Alpine/Busybox's find uses posix basic regex syntax
+INIT_SCRIPTS_DIR="$(dirname "$0")"
+find "${INIT_SCRIPTS_DIR}/" -regex ".*/[0-9][0-9][0-9]-[^/]*\\.cmf" -type f -print0 | \
+    sort -z -n -t "-" -k 1 | \
+    xargs -0 -I {} bash -c "echo \"Running migration file {}\" && bash \"{}\""
```
SpecLad:
```suggestion
for migration in "${INIT_SCRIPTS_DIR}"/[0-9][0-9][0-9]-*.cmf; do
    echo "Running migration file $migration"
    "$migration"
done
```

This works because glob expansion sorts the paths lexicographically.

docker-compose.yml
```diff
@@ -329,16 +333,49 @@ services:
...
+    entrypoint:
+      # https://github.com/ClickHouse/ClickHouse/blob/master/programs/client/clickhouse-client.xml
+      - /bin/bash
+      - -c
+      - |
+        mkdir -p /etc/clickhouse-client
+        cat << EOF > /etc/clickhouse-client/config.xml
+        <config>
+          <user>$${CLICKHOUSE_USER}</user>
+          <password>$${CLICKHOUSE_PASSWORD}</password>
```
SpecLad: This will break if the password contains a `<` or `&` character.

docker-compose.yml
```diff
@@ -329,16 +333,49 @@ services:
...
     volumes:
-      - ./components/analytics/clickhouse/init.sh:/docker-entrypoint-initdb.d/init.sh:ro
       - cvat_events_db:/var/lib/clickhouse/
+    healthcheck:
+      test: ["CMD", "clickhouse-client", "--query", "SELECT 1"]
+      interval: 3s
+      timeout: 300s
+      retries: 3
+
+  cvat_clickhouse_init:
```
SpecLad: It doesn't seem right to add another init container when we already have a place for the initialization logic - the `init` command in the backend entrypoint script. That's where the PostgreSQL and Redis migrations happen, so shouldn't we do the Clickhouse migrations there as well?

You'll need to rewrite the migrations as Python modules (since the backend container only has the Python client library), but I think it's better anyway, since it'll make it easier to implement more complex logic if it becomes necessary in the future.

components/analytics/clickhouse/README.md
SpecLad: Why do we even need to put this script in `/docker-entrypoint-initdb.d`, then? Seems like it'll be run twice for no reason.

components/analytics/clickhouse/README.md
SpecLad: Ah, okay, I didn't realize that `ADD COLUMN` has its own `IF NOT EXISTS` option.

However, I'm concerned that this will pretty much restrict us to migrations that add stuff. If we add e.g. renames or deletions into the mix, even if each migration is idempotent, the entire sequence isn't. E.g., consider:

1. Add column A.
2. Rename column A to B.

If you run this sequence twice, you'll have a duplicate column and the 2nd step will fail.

So IMO, we should just bite the bullet and record migration information in the DB.

components/analytics/clickhouse/README.md
zhiltsov-max: This directory is not mounted into `/docker-entrypoint-initdb.d` anymore, except for tests. Maybe this line in the documentation should be updated somehow.

components/analytics/clickhouse/README.md
zhiltsov-max: I agree that we might need to support more complex migrations in the future. I just don't think this is worth being supported now.

docker-compose.yml
zhiltsov-max: Clickhouse uses a dedicated docker image for executing commands, so it would require changing the existing migrations and/or installing more components in the base image. I don't see why we have to spend another several weeks now on supporting more complex migrations there, given that we don't have plans to create them.

components/analytics/clickhouse/init.sh
zhiltsov-max: Ok. It seems to work only because there are 0's in the beginning of file names, but we'll probably stick to this naming convention.

docker-compose.yml
SpecLad: > installing more components in the base image

What components? I'm talking about using the existing `clickhouse-connect` library, which is already installed.

helm-chart/templates/analytics/clickhouse/initializer/job.yml
SpecLad: > I had an idea that if the default clickhouse is used, it will be downloaded by the time this job starts.

This will not work if this pod is assigned to a different node than the main Clickhouse pod.

components/analytics/clickhouse/README.md
SpecLad: Okay, I can agree with this. However, then the documentation must be more stringent in describing what kind of migrations are allowed.

(Incidentally, if we're not recording applied migrations, then there's no real reason to split them into different files - you could just execute all SQL commands in a single `clickhouse-connect` command.)

docker-compose.yml
zhiltsov-max: Checked such passwords - apparently, the default clickhouse container writes the user and password env variables unescaped into a similar server config xml, which results in a server startup error later. I guess there's no point for us in supporting it at the moment.

components/analytics/clickhouse/README.md
zhiltsov-max: Ok, updated the documentation.

docker-compose.yml
zhiltsov-max: As I said, if we use this library, the code to run the migrations has to be modified.

docker-compose.yml
SpecLad: FWIW, it seems that they fixed it: https://github.com/ClickHouse/ClickHouse/commit/d17b20705a698c5705c9caf0e19a854be95af567

docker-compose.yml
SpecLad: Sure, but like... the whole code is just executing a bunch of SQL queries. That's [pretty straightforward](https://github.com/cvat-ai/cvat/blob/715ced23e3fab1376c9bb00dbcbff83e976858c1/cvat/apps/events/export.py#L67-L74).

You don't even need to implement the migration loading code, since, as I've mentioned in another thread, all queries could just be performed in one module.

components/analytics/clickhouse/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+This directory contains migrations for the Clickhouse DB used by CVAT.
+
+## Implementation details
+
+The directory contains initialization scripts for the Clickhouse instance.
+Only the main script `init.sh` is an executable shell script. The migration files
+are also shell scripts inside, but they have the `.cmf` extension, which stands
+for "clickhouse migration file". We use a custom extension to avoid automatic execution
+of these files by Clickhouse.
```
SpecLad: It seems that this reasoning doesn't hold if we no longer mount this directory into the server container. So we could rename the files to `*.sh`, to make them easier to work with in IDEs.

docker-compose.yml
zhiltsov-max: Updated implementation

azhavoro:
@zhiltsov-max Please update the chart version

components/analytics/clickhouse/init.py
```diff
@@ -0,0 +1,105 @@
...
+        port=parsed_args.port or os.getenv(CLICKHOUSE_PORT_ENV_VAR),
+        username=parsed_args.user or os.getenv(CLICKHOUSE_USER_ENV_VAR),
+        password=parsed_args.password or os.getenv(CLICKHOUSE_PASSWORD_ENV_VAR),
+    ) as client:
+        if parsed_args.clear:
+            client.database = db_name
+
+            print("Clearing the DB")
+            clear_db(client)
+            print("done")
```
SpecLad: Seems like this should be moved to after the `create_db` call - then it'll work regardless of whether the database exists prior to the script execution or not (and the `client.database` assignment can then be removed).

components/analytics/clickhouse/init.py
```diff
@@ -0,0 +1,105 @@
...
+
+CLICKHOUSE_HOST_ENV_VAR = "CLICKHOUSE_HOST"
+CLICKHOUSE_PORT_ENV_VAR = "CLICKHOUSE_PORT"
+CLICKHOUSE_DB_ENV_VAR = "CLICKHOUSE_DB"
+CLICKHOUSE_USER_ENV_VAR = "CLICKHOUSE_USER"  # nosec
+CLICKHOUSE_PASSWORD_ENV_VAR = "CLICKHOUSE_PASSWORD"  # nosec
+
+
+def clear_db(client: clickhouse_connect.driver.client.Client):
+    client.query(f"DROP TABLE IF EXISTS {client.database}.events;")
```
SpecLad:
```suggestion
    client.query("DROP TABLE IF EXISTS events;")
```

`client.database` should be the default database.

components/analytics/clickhouse/init.py
zhiltsov-max: > and the `client.database` assignment can then be removed

The problem is that you can't create a client instance with the database parameter, if the database does not exist. The client tries to connect to the db and fails in the constructor. So, you're kind of forced to assign the database name directly to the client object after the db is created.

Example from [1c519f9](https://github.com/cvat-ai/cvat/pull/9689/commits/1c519f90c443b7230a22353c94b15d32e6da3850) https://github.com/cvat-ai/cvat/actions/runs/17327801721/job/49196616030

SpecLad: > The problem is that you can't create a client instance with the database parameter,

That's not what I mean. Currently, you have two assignments to `client.database`. But if you swap the positions of the `if parsed_args.clear` block and the `create_db(...)` block, then you can remove the first assignment (the second one will have to stay, of course).

components/analytics/clickhouse/init.py
Eldies: May be use `TRUNCATE DATABASE` instead of dropping table by name? No difference for now, but we will not need to update this function if we ever add other tables

components/analytics/clickhouse/init.py
zhiltsov-max: Ok, done.

components/analytics/clickhouse/init.py
zhiltsov-max: Yeah, good idea.

SpecLad (approved):

zhiltsov-max:
@azhavoro, updated.

PR #9691, opened 2025-07-31
Stream shapes on backup

Eldies:
### Motivation and context

On backup, only one job annotations resides in RAM at a time. After it is processed, it is removed from RAM and the next job processing starts. Therefore, project or task backup only consumes as much RAM as the largest job.

But if there is a job with large annotations (several Gb), it can be a problem.

Reading and processing job annotations in a streaming manner in this PR to reduce RAM usage.
Only doing it for shapes because shapes consume a lot more RAM generally than other types of annotations.

### Checklist

### License


cvat/apps/engine/backup.py
```diff
@@ -574,10 +577,21 @@ def serialize_annotations():
...
-                annotations_serializer.is_valid(raise_exception=True)
-                yield self._prepare_annotations(annotations_serializer.data, self._label_mapping)
+                with transaction.atomic():
+                    annotations = dm.task.get_job_data(db_job_id, streaming=True)
+                    annotations_serializer = LabeledDataSerializer(data=dict(annotations, shapes=[]))
+                    annotations_serializer.is_valid(raise_exception=True)
+                    annotation_data = annotations_serializer.data
+
+                    def serialize_shapes():
+                        for shape in annotations["shapes"]():
```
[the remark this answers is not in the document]
Eldies: added assert that check that it is a callable

cvat/apps/engine/backup.py
```diff
@@ -574,10 +577,22 @@ def serialize_annotations():
...
+                            shape_serializer = LabeledShapeSerializer(data=shape)
+                            shape_serializer.is_valid(raise_exception=True)
+                            yield shape_serializer.data
+
+                    annotation_data["shapes"] = serialize_shapes()
```
zhiltsov-max: It looks like you could call the `LabeledDataSerializer` normally if `annotations["shapes"] = generate_shapes()` . In this case it will be a generator that will only be called during iteration:

```python
def gen():
  print("gen() started")
  yield 1

it = gen() # prints nothing
list(it) # prints
```

It's also possible to check that the value is a generator:
```python
import inspect
inspect.isgenerator(gen())
```
instead of `assert callable(annotations["shapes"])`.

I think it would be good to use a generator directly instead of requiring to call it explicitly to simplify accessing annotations in other components (export, quality, consensus). 

We need to check that we're in a transaction if annotation streaming is requested, it can be done in the generator call itself. Otherwise, the lifetime of the server-side cursor will be equal to the lifetime of the iterator variable.

zhiltsov-max: Ideally, it should be wrapped in a context manager like this:
```python
with contextlib.closing(annotations["shapes"]):
    serializer(annotations)
```

Otherwise it's possible to get warnings if the loop is exited early and the iterator object is destroyed after the transaction closed, because the db-side cursor was not closed during the transaction:

```
Exception ignored in: <generator object cursor_iter at 0x7258dd849a10>
  File "venv3.10/lib/python3.10/site-packages/django/db/models/sql/compiler.py", line 2096, in cursor_iter
    cursor.close()
psycopg2.ProgrammingError: named cursor isn't valid anymore
```

Calling it for each field explicitly in the client code is a little bit clumsy, so maybe a context manager could be provided.

Eldies: As far as I understand, `psycopg2.ProgrammingError: named cursor isn't valid anymore` happens, if cursor is created inside of a transaction, and then accessed after the transaction is closed. So, it may happen if partially iterate shapes inside of transaction and then iterate the remaining shapes outside of transaction. 
I think, it is a wrong way to use shapes (because they should be iterated only once) and it is great that it fails. 
Therefore, no closing is needed.

It is not possible to use `LabeledDataSerializer` normally, because it expects a list, not any other iterable.

I now check in `_init_shapes_from_db` that streaming happens inside of transaction and also read the first shape immediately to create a db-side cursor. Without it, it could work like this:
```
with transaction.atomic():
    annotations = dm.task.get_job_data(db_job_id, streaming=True)

for _ in annotations["shapes']: # db-side cursor only initialised here, outside of transaction
    pass
```

zhiltsov-max: > As far as I understand, psycopg2.ProgrammingError: named cursor isn't valid anymore happens, if cursor is created inside of a transaction, and then accessed after the transaction is closed. So, it may happen if partially iterate shapes inside of transaction and then iterate the remaining shapes outside of transaction.
I think, it is a wrong way to use shapes (because they should be iterated only once) and it is great that it fails.
Therefore, no closing is needed.

Yes, but it's not the only case. I played with the code a little bit. The error can be obtained even without explicit interaction with the cursor:

```python
from contextlib import closing
from cvat.apps.engine.models import Project
from django.db import transaction


def open_cursor():
    with transaction.atomic():
        it = Project.objects.iterator(chunk_size=1)
        # with closing(it):
        for _ in it:
            break

open_cursor()

-> named cursor isn't valid anymore because the generator wasn't closed and it was destroyed on function exit
```

I agree that this requires special conditions in the code, so maybe we don't need to do anything special.

> I now check in _init_shapes_from_db that streaming happens inside of transaction and also read the first shape immediately to create a db-side cursor.

Nice

cvat/apps/dataset_manager/task.py
```diff
@@ -629,8 +629,8 @@ def _init_tags_from_db(self):
         serializer = serializers.LabeledImageSerializerFromDB(db_tags, many=True)
         self.ir_data.tags = serializer.data
 
-    def _init_shapes_from_db(self):
-        db_shapes = [
+    def _init_shapes_from_db(self, streaming: bool = False):
```
zhiltsov-max:
```suggestion
    def _init_shapes_from_db(self, *, streaming: bool = False):
```

cvat/apps/engine/backup.py
```diff
@@ -579,10 +582,22 @@ def serialize_annotations():
...
             db_job_ids = (j.id for j in db_jobs)
             for db_job_id in db_job_ids:
-                annotations = dm.task.get_job_data(db_job_id)
-                annotations_serializer = LabeledDataSerializer(data=annotations)
-                annotations_serializer.is_valid(raise_exception=True)
-                yield self._prepare_annotations(annotations_serializer.data, self._label_mapping)
+                with transaction.atomic():
+                    annotations = dm.task.get_job_data(db_job_id, streaming=True)
+                    assert not isinstance(annotations["shapes"], list)
+                    annotations_serializer = LabeledDataSerializer(data=dict(annotations, shapes=[]))
```
zhiltsov-max, line 608:
```suggestion
                    # Django many=True fields can only handle the list type
                    # we're using a generator here, so it's processed separately
                    annotations_serializer = LabeledDataSerializer(data=dict(annotations, shapes=[]))
```

cvat/apps/engine/backup.py
```diff
@@ -706,6 +721,7 @@ def read(zip_object):
 
     def _create_annotations(self, db_job, annotations):
         self._prepare_annotations(annotations, self._labels_mapping)
+        annotations["shapes"] = list(annotations["shapes"])
```
zhiltsov-max, line 745:
```suggestion
        assert not isinstance(annotations["shapes"], list)
        annotations["shapes"] = list(annotations["shapes"])
```

cvat/apps/dataset_manager/task.py
```diff
@@ -1059,9 +1084,9 @@ def data(self):
 
 @silk_profile(name="GET job data")
 @transaction.atomic
-def get_job_data(pk):
+def get_job_data(pk, streaming: bool = False):
```
zhiltsov-max:
```suggestion
def get_job_data(pk, *, streaming: bool = False):
```

cvat/apps/dataset_manager/task.py
```diff
@@ -648,38 +648,63 @@ def _init_shapes_from_db(self):
...
+                    if db_shape.parent not in elements:
+                        elements[db_shape.parent] = []
+                    elements[db_shape.parent].append(db_shape)
+
+            yield from yield_shapes_for_one_frame(shapes, elements)
 
-        for shape_id, shape_elements in elements.items():
-            shapes[shape_id].elements = shape_elements
+        if streaming:
+            assert transaction.get_connection().in_atomic_block
```
zhiltsov-max, line 696: Please add a unit test for this check

zhiltsov-max: https://github.com/cvat-ai/cvat/pull/9694/files#diff-e6ee90bf834947c1146c08132e0c8a5f7f4498bb9b28e92d1d3f60930d8d258cR137-R139

PR #9692, opened 2025-08-01
Optimization: fetch previews one by one

klakhov:
### Motivation and context

Currently we can have quite large pages (100 tasks for example). We start sending preview request when card becomes visible, but scrolling too fast can lead to lots of requests from ui. This pr adds queue to fetching previews to request them one by one

### Checklist

### License


cvat-ui/src/utils/preview-queue.ts
```diff
@@ -0,0 +1,124 @@
...
+        this.isProcessing = true;
+
+        while (this.queue.length > 0) {
+            const request = this.queue.shift()!;
+            this.currentRequest = request;
+
+            await this.executeRequest(request);
+
+            // Small delay between requests to avoid overwhelming the server
+            await new Promise<void>((resolve) => {
```
bsekachev: Not sure we really need it. Just sending requests one by one is enough

cvat-ui/src/utils/preview-queue.ts
```diff
@@ -0,0 +1,124 @@
...
+
+const PREVIEW_REQUEST_DELAY = 100;
+
+class PreviewQueue {
+    private queue: PreviewRequest[] = [];
+    private isProcessing: boolean = false;
+    private currentRequest: PreviewRequest | null = null;
+
+    public addRequest(request: PreviewRequest): void {
+        const isDuplicate = this.queue.some((req) => req.id === request.id) ||
```
bsekachev, line 45: If we receive request for another entity type, it usually means that user changed the page. 
Maybe in this case it would be better to push new entity in the beginning of the queue.

bsekachev (approved):
overall lgtm, please see couple of comments

cvat-ui/src/utils/preview-queue.ts
klakhov: Removed.

cvat-ui/src/utils/preview-queue.ts
klakhov: Currently, if the preview component is unmounted before its request is processed, the request is removed from the queue entirely. This means that previews for new entities start fetching immediately.

Do you think we should avoid aborting previous requests and instead move them to the end of the queue?

bsekachev: Good

PR #9694, opened 2025-08-01
Stream shapes for job export

Eldies:
### Motivation and context

Reducing RAM consumption on job export by not keeping all db annotations in RAM

depends on https://github.com/cvat-ai/cvat/pull/9691

### Checklist

### License

### DIFF
--- a/cvat/apps/engine/utils.py
+++ b/cvat/apps/engine/utils.py
@@ -489,6 +489,7 @@ def defaultdict_to_regular(d):
 @contextmanager
 def transaction_with_repeatable_read():
     with transaction.atomic():
-        connection.cursor().execute("SET TRANSACTION ISOLATION LEVEL REPEATABLE READ;")
-        connection.cursor().execute("SET TRANSACTION READ ONLY;")
+        if connection.vendor != "sqlite":
+            connection.cursor().execute("SET TRANSACTION ISOLATION LEVEL REPEATABLE READ;")
+            connection.cursor().execute("SET TRANSACTION READ ONLY;")
         yield


cvat/apps/dataset_manager/annotation.py
```diff
@@ -182,6 +183,10 @@ def reset(self):
         self.shapes = []
         self.tracks = []
 
+    @property
+    def is_stream(self) -> bool:
+        return not isinstance(self.shapes, list)
```
zhiltsov-max, line 188: It would probably be better to change the type annotation to `Iterable | list` in this case, otherwise I'd prefer using `inspect.isgenerator()`.

cvat/apps/dataset_manager/task.py
```diff
@@ -1133,14 +1134,8 @@ def export_job(
...
 ):
-    # For big tasks dump function may run for a long time and
-    # we dont need to acquire lock after the task has been initialized from DB.
-    # But there is the bug with corrupted dump file in case 2 or
-    # more dump request received at the same time:
-    # https://github.com/cvat-ai/cvat/issues/217
-    with transaction.atomic():
-        job = JobAnnotation(job_id, prefetch_images=True, lock_job_in_db=True)
-        job.init_from_db()
+    job = JobAnnotation(job_id, prefetch_images=True, lock_job_in_db=True)
```
zhiltsov-max: Probably we don't need to lock now, otherwise the job will be unavailable during the export.

cvat/apps/dataset_manager/bindings.py
```diff
@@ -544,9 +559,118 @@ def get_frame(idx):
...
+                task_id=self._db_task.id,
+                construct_frame=partial(construct_frame, frame_idx),
+            )
+
     @property
     def shapes(self):
-        for shape in self._annotation_ir.shapes:
+        shapes = self._annotation_ir.shapes
+        if callable(shapes):
+            shapes = shapes()
```
zhiltsov-max:
```suggestion
```
Probably not relevant anymore

cvat/apps/dataset_manager/bindings.py
```diff
@@ -544,9 +559,118 @@ def get_frame(idx):
 
         return iter(frames.values())
 
+    def group_by_frame_stream(self) -> Generator[CommonData.LazyFrame, None, None]:
```
zhiltsov-max, line 562: The original function already tries to be a stream, do we really need another one?

Eldies: Well, I plan to remove the old function in the next PR, where streaming will be done for tasks. I am not sure the new function works for tasks right now

cvat/apps/dataset_manager/task.py
Eldies: Removed the lock.
I think something like `connection.cursor().execute("SET TRANSACTION ISOLATION LEVEL REPEATABLE READ")` should be done in this transaction to ensure exported data is consistent. But I do not see anything like that anywhere in the code. Are there any reasons to not do it?

cvat/apps/dataset_manager/bindings.py
Eldies: removed it

cvat/apps/dataset_manager/task.py
zhiltsov-max: Yeah, this isolation level can be a potential solution for use cases where we only need to read something from the DB. There aren't many such use cases, but annotation fetching for exporting or quality checks can really be the one, because such logic is, basically, supposed there. Typically, we'd like to avoid using raw SQL, so the reason why it's not used is, probably, that nobody researched SQL/DB provider capabilities for transactions that deep and it's not supported by Django ORM.

cvat/apps/dataset_manager/bindings.py
zhiltsov-max: Ok, please try to make the interface the same (without separation on `Frame` / `LazyFrame`) then. I think it can be beneficial to implement annotation caching just for 1 frame and maybe perform loading on access.

cvat/apps/dataset_manager/task.py
Eldies: I made a decorator/context_manager to run transactions with repeatable read (and read-only since it only intended for export right now). It uses raw sql to set isolation level, and needs to check what db is used (because sqlite (used for tests) does not allow to set isolation level). So, does not look good.

I also tried to make a second entry in settings.DATABASES with pre-set isolation level, but it makes code extremely error-prone, because with that approach `using` should be manually set for all queries, and it is easy to mix something up.

zhiltsov-max: No, I think the current solution is fine. I need to test it a little bit. Implementation exceptions for sqlite aren't something new in the codebase. It's probably simply not needed for sqlite, because transactions are [serializable there](https://www.sqlite.org/isolation.html).

zhiltsov-max (approved):

cvat/apps/dataset_manager/bindings.py
Eldies: Removed LazyFrame in https://github.com/cvat-ai/cvat/pull/9772

PR #9702, opened 2025-08-06
Added support to remove users having organizations with other members

bsekachev:
### Motivation and context
As discussed:
- Deleting a user who is owner of an organizations should:
  - Should trigger deleting the organization and all resources if not more members inside the organization
  - Keep the organization and resources if there are any members inside the organization

- Additionally we will need to detect and remove empty organizations, but this task seems to be easy. 

I updated user deletion validators registration to explicit one which seems to be better. 

### Checklist

### License


bsekachev:
@zhiltsov-max 

Here I removed restrictions related to removing users who owners of orgs with many members.
Pls, check

cvat/apps/engine/user_deletion.py
```diff
@@ -16,21 +15,18 @@
...
     """
     Base class for user deletion validators.
     To add a new validator, inherit from this class and implement the `validate` method.
-    The implemented class will be automatically registered and used during user deletion.
+    Then call classmethod 'register_validator()' to enable the class as validator.
     """
 
+    @classmethod
+    def register_validator(cls):
+        if cls not in _USER_DELETION_VALIDATORS:
```
zhiltsov-max, line 27: Consider changing the type of `_USER_DELETION_VALIDATORS` to `set` if you're trying to repeatedly look for the existing values

bsekachev: I think `list` looks more natural here. 
As this will be always short (e.g. not 1000s of values), I hardly think changing to `set` makes sense from algorithm pov

zhiltsov-max (approved):

PR #9711, opened 2025-08-11
Remove redundant dispatch_uid parameters

SpecLad:
### Motivation and context

The goal of `dispatch_uid` is to prevent the same signal handler from being registered more than once if the registration code is run again. However, the `@receiver` decorators are run during module import, which should only ever happen once. Therefore, it seems that all of the `dispatch_uid` parameters in our code are redundant. Remove them to simplify the code.

Ironically, the signal handlers registered in the IAM `register_signals` function actually _do_ need `dispatch_uid`s, but don't have them. This is because this function is called from `IAMConfig.ready`, which allegedly [1] can be run multiple times. This hasn't been a problem yet, but add some UIDs just in case.

[1]: https://docs.djangoproject.com/en/5.2/topics/signals/#preventing-duplicate-signals

### How has this been tested?

I manually checked that all receivers of `post_delete` are unique at runtime.

### Checklist

### License


SpecLad:
@zhiltsov-max Mind taking a look?

zhiltsov-max:
It's hard to say. Having names for signals looks like a good thing, e.g. if we wanted to override signal handlers from a specific component, but it doesn't seem like we ever really needed it. I'd prefer to automatically generate this parameter if it is not specified, and also check that there is no signal handler with this name registered already.

SpecLad:
> Having names for signals looks like a good thing, e.g. if we wanted to override signal handlers from a specific component, but it doesn't seem like we ever really needed it.

I don't think you can actually override signal handlers with this. IIRC, if a handler with the same `dispatch_uid` already exists, then the new handler just isn't registered.

> I'd prefer to automatically generate this parameter if it is not specified

That's basically what the implementation does already. If `dispatch_uid` is not specified, it uses `id(handler)` as the key.

SpecLad:
@zhiltsov-max So, what do you think?

PR #9725, opened 2025-08-15
Fixing tests for lightweight backup

Eldies:
### Motivation and context

https://github.com/cvat-ai/cvat/pull/9535 does not correctly process lightweight backups for cloud based tasks for case when static cache is allowed. (i.e. treats them like not cloud based tasks).
It is not obvious right now how lightweight backup should behave for that case.
Fixing tests to expect current behaviour.

### Checklist

### License

### DIFF
--- a/tests/python/rest_api/test_tasks.py
+++ b/tests/python/rest_api/test_tasks.py
@@ -26,6 +26,7 @@ from typing import Any, Optional
 
 import numpy as np
 import pytest
+from attr.converters import to_bool
 from cvat_sdk import exceptions
 from cvat_sdk.api_client import models
 from cvat_sdk.api_client.api_client import ApiClient, Endpoint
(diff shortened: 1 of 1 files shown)


tests/python/rest_api/test_tasks.py
zhiltsov-max, line 1: What is different in this case? These are regular tasks with CS, so they should be handled the same way.

Eldies: if static cache is used for the CS based task, then its `storage` field is switched to `local` [here](https://github.com/cvat-ai/cvat/blob/0d197e6b4b3fe2dbc249654242b732ede101f82a/cvat/apps/engine/task.py#L746)

I use `storage` field to determine whether the task is CS based to make lightweight backups, so it does not work with static cache.

CS based task with static cache still has non null `cloud_storage_id` field and can be distinguished from regular task by it.
But I don't think `cloud_storage_id` should be used to determine whether the task is CS based, because it can be `None` if the CS is deleted.

Also, I see no way to distinguish a CS based task with static cache from a regular task, if the CS is deleted.

Probably, `storage` should stay `cloud_storage` even if static cache is used

archibald1418:
Could we try and merge this today until the next CI-nightly run?

tests/python/rest_api/test_tasks.py
```diff
@@ -1315,6 +1315,14 @@ def test_can_export_and_import_backup_task_with_cloud_storage(self, tasks, light
 
         assert filename.is_file()
         assert filename.stat().st_size > 0
+
+        if lightweight_backup and os.getenv("CVAT_ALLOW_STATIC_CACHE") != "true":
+            with zipfile.ZipFile(filename, "r") as zf:
+                files_in_data = [
+                    os.path.split(name)[1] for name in zf.namelist() if name.startswith("data")
+                ]
+                assert files_in_data == ["manifest.jsonl"]
```
zhiltsov-max:
```suggestion
from attr.converters import to_bool
...
        if lightweight_backup:
            with zipfile.ZipFile(filename, "r") as zf:
                files_in_data = {
                    name.split("data/", maxsplit=1)[1]
                    for name in zf.namelist() if name.startswith("data/")
                }

            expected_media = {"manifest.jsonl"}
            if to_bool(os.getenv("CVAT_ALLOW_STATIC_CACHE")):
                # FIXME: remove extra media files
                expected_media.update(cloud_storage_content)
            
            assert files_in_data == expected_media
```
- need to check both cases of lightweight backups
- `os.path.split` won't work on Windows, zip uses only "/" as separator
- we use `to_bool` (or it's equivalent) for boolean env vars

tests/python/rest_api/test_tasks.py
zhiltsov-max: Yeah, changing `storage` to `cloud_storage` seems to be the correct solution. This will need a data migration, though. In this case, for CS-based tasks:
- if not lightweight, check for media inside the upload dir, download what's not found, copy into the backup archive
- if lightweight, don't copy media (even if available locally)

I suggest to make this change in another PR and only fix the CI issue here.

PR #9732, opened 2025-08-18
Exporting with server filename in sdk/cli

zhiltsov-max:
### Motivation and context

Currently, it's only possible to download with a custom filename. If the resource is updated frequently and the exported data is cached locally, using a custom filename can result in an incorrect filename (e.g. outdated). This PR allows to export annotations and backups using the server-generated filename, the same way as it if was downloaded by a web browser.

- Add support for exporting with server filename in sdk/cli

### Checklist

### License

### DIFF
--- a/cvat-sdk/cvat_sdk/core/downloading.py
+++ b/cvat-sdk/cvat_sdk/core/downloading.py
@@ -43,7 +43,7 @@ class Downloader:
         if not stem or len(ext) < 2:
             return None
 
-        if filename.startswith(".") or re.search(r"[^A-Za-z0-9_\-\.]", filename):
+        if filename.startswith(".") or re.search(r"[^A-Za-z0-9_\-\. ]", filename):
             return None
 
         return filename


cvat-cli/src/cvat_cli/_internal/commands_tasks.py
```diff
@@ -309,12 +316,19 @@ def execute(
         status_check_period: int,
         include_images: bool,
     ) -> None:
+        if filename is None:
+            filename = os.getcwd()
+
+        if filename[-1:] in ("/", "\\") and not os.path.exists(filename):
```
SpecLad:
```suggestion
        if filename.endswith((os.sep, os.altsep or os.sep)):
```

Second condition is redundant with `exist_ok=True`.

cvat-cli/src/cvat_cli/_internal/commands_tasks.py
```diff
@@ -276,7 +278,12 @@ class TaskExportDataset:
 
     def configure_parser(self, parser: argparse.ArgumentParser) -> None:
         parser.add_argument("task_id", type=int, help="task ID")
-        parser.add_argument("filename", type=str, help="output file")
+        parser.add_argument(
+            "filename",
+            type=str,
+            nargs="?",
+            help="output file or directory (default: current directory)",
```
SpecLad, line 286:
```suggestion
            default=os.getcwd(),
            help="output file or directory (default: current directory)",
```

SpecLad: Also, adding a default to this option should be reflected in the changelog.

cvat-cli/src/cvat_cli/_internal/commands_tasks.py
```diff
@@ -309,12 +316,19 @@ def execute(
...
+        if filename[-1:] in ("/", "\\") and not os.path.exists(filename):
+            os.makedirs(filename, exist_ok=True)
+
         client.tasks.retrieve(obj_id=task_id).export_dataset(
             format_name=fileformat,
             filename=filename,
             pbar=DeferredTqdmProgressReporter(),
             status_check_period=status_check_period,
             include_images=include_images,
+            location=Location.LOCAL,
```
SpecLad, line 332: This seems like it should be in the changelog.

cvat-sdk/cvat_sdk/core/downloading.py
```diff
@@ -34,14 +34,17 @@ def download_file(
         *,
         timeout: int = 60,
         pbar: Optional[ProgressReporter] = None,
-    ) -> None:
+    ) -> Path:
         """
         Downloads the file from url into a temporary file, then renames it to the requested name.
+        If output_path is a directory, saves the file into the directory with
+        the server-defined name.
```
SpecLad, line 88: You should document the return value.

cvat-sdk/cvat_sdk/core/downloading.py
```diff
@@ -58,6 +61,17 @@ def download_file(
             except ValueError:
                 file_size = None
 
+            if output_path.is_dir():
+                content_disposition = next(
+                    part.strip()
+                    for part in response.headers.get("Content-Disposition", "").split(";")
+                    if part.strip().startswith("filename=")
```
SpecLad: This should be matched case-insensitively to conform to the spec.

cvat-sdk/cvat_sdk/core/downloading.py
```diff
@@ -58,6 +61,17 @@ def download_file(
...
+                )
```
SpecLad: What if there is no filename, or no `Content-Disposition` header at all? I think we should produce an exception with an understandable message in this case.

cvat-sdk/cvat_sdk/core/downloading.py
```diff
@@ -58,6 +61,17 @@ def download_file(
...
+
+                filename = Path(content_disposition.split("=", maxsplit=1)[1].strip('"'))
+                filename = filename.with_stem(normalize_filename(filename.stem))
```
SpecLad: I don't agree with the `normalize_filename` approach. This function promises to save the file with the "server-generated name". So we should do that, rather than mangling the name in arbitrary ways. If the generated name is not a valid filename, then we should throw an exception. Presumably this will never happen with a legitimate server.

cvat-sdk/cvat_sdk/core/downloading.py
```diff
@@ -34,14 +34,17 @@ def download_file(
...
         Downloads the file from url into a temporary file, then renames it to the requested name.
+        If output_path is a directory, saves the file into the directory with
+        the server-defined name.
         """
 
         CHUNK_SIZE = 10 * 2**20
 
-        assert not output_path.exists()
+        if output_path.exists() and not output_path.is_dir():
+            raise FileExistsError(output_path)
```
SpecLad, line 96: Perhaps this check should be done later, after the server-defined name has been received?

cvat-cli/src/cvat_cli/_internal/commands_tasks.py
```diff
@@ -309,12 +316,19 @@ def execute(
         status_check_period: int,
         include_images: bool,
     ) -> None:
+        if filename is None:
+            filename = os.getcwd()
+
+        if filename[-1:] in ("/", "\\") and not os.path.exists(filename):
+            os.makedirs(filename, exist_ok=True)
```
SpecLad, line 324: Perhaps we should move this logic to the SDK function? It seems strange to only support automatic directory creation in the CLI.

cvat-sdk/cvat_sdk/core/downloading.py
zhiltsov-max: `headers` is a special [dict that handles this](https://github.com/urllib3/urllib3/blob/b8f512b225c99c552f461c5a8ed250652c41ee93/src/urllib3/_collections.py#L204).

cvat-sdk/cvat_sdk/core/downloading.py
zhiltsov-max: Ok, this is done on the server anyway.

cvat-cli/src/cvat_cli/_internal/commands_tasks.py
zhiltsov-max: I'm not sure it's a good behavior for an sdk function. Typically we don't do this on this level. For CLI, it reduces the amount of extra typing for the user, so I think it should be done just there.

cvat-sdk/cvat_sdk/core/downloading.py
zhiltsov-max: It this case the tests should fail, but I added error handling.

cvat-sdk/cvat_sdk/core/downloading.py
zhiltsov-max: Agree, added a check for this case.

cvat-sdk/cvat_sdk/core/downloading.py
zhiltsov-max: Ok, added.

cvat-cli/src/cvat_cli/_internal/commands_tasks.py
zhiltsov-max: Added.

cvat-cli/src/cvat_cli/_internal/commands_tasks.py
zhiltsov-max: Added.

cvat-sdk/cvat_sdk/core/downloading.py
SpecLad: I'm not talking about getting the header, I'm talking about the `startswith` check. The string `filename` is supposed to be matched case-insensitively.

cvat-cli/src/cvat_cli/_internal/commands_tasks.py
SpecLad: Alright.

cvat-sdk/cvat_sdk/core/downloading.py
```diff
@@ -27,21 +29,52 @@ class Downloader:
...
+
+        filename = None
+        if content_disposition:
+            filename_parts = content_disposition.split("=", maxsplit=1)
+            if len(filename_parts) == 2:
+                filename = filename_parts[1].strip('"')
+                filename = Path(filename).name
+
+        if not filename:
+            raise Exception(
```
SpecLad:
```suggestion
            raise CvatSdkException(
```

cvat-sdk/cvat_sdk/core/downloading.py
```diff
@@ -27,21 +29,52 @@ class Downloader:
     def __init__(self, client: Client):
         self._client = client
 
+    @staticmethod
+    def _get_server_filename(response: urllib3.HTTPResponse) -> str:
+        content_disposition = next(
+            (
+                filename
+                for part in response.headers.get("Content-Disposition", "").split(";")
+                if (filename := part.strip()) and filename.startswith("filename=")
```
SpecLad: The code is a bit confusing, because here `filename` includes the `filename=` string, but below it doesn't. Maybe you could rename this to `filename_parameter` or something?

cvat-sdk/cvat_sdk/core/downloading.py
```diff
@@ -27,21 +29,52 @@ class Downloader:
...
+            ),
+            None,
+        )
+
+        filename = None
+        if content_disposition:
+            filename_parts = content_disposition.split("=", maxsplit=1)
+            if len(filename_parts) == 2:
```
SpecLad: It seems like this should always be true, since you already verified that the string contains an equals sign in the comprehension.

cvat-sdk/cvat_sdk/core/downloading.py
```diff
@@ -27,21 +29,52 @@ class Downloader:
...
+                filename = filename_parts[1].strip('"')
+                filename = Path(filename).name
+
+        if not filename:
+            raise Exception(
+                "Can't find the output filename in the server response, "
+                "please try to specify the output filename explicitly"
+            )
+
+        return filename
```
SpecLad, line 75: I think we need to at least check that the filename is not one of the [Windows reserved names](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions), like `nul`, `con`, etc. Otherwise, a malicious server could cause unintended effects on the client machine by using one of these names.

cvat-sdk/cvat_sdk/core/downloading.py
zhiltsov-max: Ok, renamed

cvat-sdk/cvat_sdk/core/downloading.py
zhiltsov-max: Updated

cvat-sdk/cvat_sdk/core/downloading.py
zhiltsov-max: Added validation

PR #9737, opened 2025-08-19
Test for bulk actions in UI

archibald1418:
### Motivation and context

Coverage for #9582

### How has this been tested?

Set up the following structure
image

This way we can test bulk actions on all entities while maintaining a minimal setup and suite without redundant requests or zip archives. For this, I added a `smallArchive.zip` which just has the first two pics from `archive.zip`.

Task1 is `defaultTaskSpec`, Task2 is same but with `segmentSize: 1` to init 2 jobs in one `/tasks` request. Both have same `projectID`.

- Click on 'Select all' to highlight all entities on page
- Set assignee for tasks, ensure it is `admin` for both
- Go to Task2, set assignee and status for both jobs, ensure UI reflects that without reload

### Checklist

### License


tests/cypress/support/commands_projects.js
archibald1418, line 1: the command is duplicated in `commands.js` hence the removal

tests/cypress/e2e/features/bulk_actions.js
```diff
@@ -0,0 +1,208 @@
...
+
+            cy.get('.cvat-modal-export-job')
+                .should('exist').and('be.visible')
+                .find('.ant-modal-header')
+                .should('have.text', `Export ${nobjs} jobs as datasets`);
+            cy.get('.cvat-modal-export-job').contains('button', 'OK').click();
+
+            cy.get('.cvat-notification-notice-export-job-start')
+                .should('be.visible');
+            cy.closeNotification('.cvat-notification-notice-export-job-start');
```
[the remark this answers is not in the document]
archibald1418: There's a duplicate in the other module

tests/cypress/e2e/features/bulk_actions.js
```diff
@@ -0,0 +1,208 @@
...
+
+        context('Task page, change jobs', () => {
+            it('Ensure task was assigned to admin', () => {
+                cy.get('[value="admin"]').should('exist');
+            });
+
+            it('Bulk-change assignees', () => {
+                getBulkActionsMenu().within(() => {
+                    cy.contains(`Assignee (${nobjs})`).click();
+                    cy.get('.cvat-user-search-field').type('admin', { delay: 0 }); // all at once
```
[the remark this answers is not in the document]
archibald1418: good point, ended up copyasting duplicated code

tests/cypress/e2e/features/bulk_actions.js
```diff
@@ -0,0 +1,205 @@
...
+            cy.get('.cvat-actions-menu-button').click();
+        });
+        return cy.get('.ant-dropdown');
+    }
+    function assignToAdmin() {
+        cy.contains(`Assignee (${nobjs})`).click();
+        cy.get('.cvat-user-search-field').type('admin', { delay: 0 }); // all at once
+        return cy.get('.cvat-user-search-field').type('{enter}');
+    }
+    before(() => {
```
bsekachev:
```suggestion
    function getBulkActionsMenu() {
        cy.get('.cvat-bulk-wrapper').should('exist').and('be.visible');
        cy.contains('Select all').click();
        cy.get('.cvat-item-selected').first().within(() => {
            cy.get('.cvat-actions-menu-button').click();
        });
        return cy.get('.ant-dropdown');
    }

    function assignToAdmin() {
        cy.contains(`Assignee (${nobjs})`).click();
        cy.get('.cvat-user-search-field').type('admin', { delay: 0 }); // all at once
        return cy.get('.cvat-user-search-field').type('{enter}');
    }

    before(() => {
```

tests/cypress/e2e/features/bulk_actions.js
```diff
@@ -0,0 +1,205 @@
...
+            // default task with 1 job
+            createTaskInProject(1, projectID, { taskName, labelName, serverFiles });
+
+            // default task with 2 jobs
+            createTaskInProject(2, projectID, {
+                taskName, labelName, serverFiles, segmentSize: framesPerJob,
+            }).then(() => {
+                cy.visit(`/projects/${projectID}`);
+            });
+            projects.push(projectID);
```
bsekachev:
```suggestion
            projects.push(projectID);

            // default task with 1 job
            createTaskInProject(1, projectID, { taskName, labelName, serverFiles });

            // default task with 2 jobs
            createTaskInProject(2, projectID, {
                taskName, labelName, serverFiles, segmentSize: framesPerJob,
            }).then(() => {
                cy.visit(`/projects/${projectID}`);
            });
```

tests/cypress/e2e/features/bulk_actions.js
```diff
@@ -0,0 +1,205 @@
...
+                cy.get('.cvat-item-selected')
+                    .should('exist')
+                    .its('length')
+                    .should('eq', nobjs);
+                cy.get('.cvat-resource-selection-count')
+                    .should('be.visible')
+                    .and('have.text', `Selected: ${nobjs}`);
+                cy.get('.cvat-resource-deselect-button')
+                    .should('be.visible')
+                    .and('have.text', 'Deselect').click();
```
bsekachev, line 124: Please, make sure no one is selected after clicking "Deselect"

tests/cypress/e2e/features/bulk_actions.js
```diff
@@ -0,0 +1,205 @@
...
+                cy.get('.cvat-bulk-progress-wrapper').should('be.visible');
+
+                // Navigate to task with 2 jobs
+                cy.get('.cvat-item-open-task-button').first().click();
+            });
+        });
+
+        context('Task page, change jobs', () => {
+            it('Ensure task was assigned to admin', () => {
+                cy.get('[value="admin"]').should('exist');
```
bsekachev: It seems unreliable. Potentially there may be more fields having value=["admin"].
If you want to make sure it is assigned to admin, please check exactly assignee field

tests/cypress/e2e/features/bulk_actions.js
```diff
@@ -0,0 +1,205 @@
...
+                cy.get('.cvat-resource-selection-count')
+                    .should('be.visible')
+                    .and('have.text', `Selected: ${nobjs}`);
+                cy.get('.cvat-resource-deselect-button')
+                    .should('be.visible')
+                    .and('have.text', 'Deselect').click();
+            });
+
+            it('Bulk-change assignees', () => {
+                getBulkActionsMenu().within(() => {
```
bsekachev, line 129: Try to rewrite tests the way they do not depend each on other. This test will not work as expected if previous is not run.

tests/cypress/e2e/features/bulk_actions.js
```diff
@@ -0,0 +1,205 @@
...
+                    assignToAdmin();
+                });
+                cy.get('.cvat-bulk-progress-wrapper').should('be.visible');
+
+                // Navigate to task with 2 jobs
+                cy.get('.cvat-item-open-task-button').first().click();
```
bsekachev: Please, try to use `taskId` to defined a task with 2 jobs. I am not sure, just getting `.first()` is a good option

tests/cypress/e2e/features/bulk_actions.js
```diff
@@ -0,0 +1,205 @@
...
+            });
+        });
+
+        context('Task page, change jobs', () => {
+            it('Ensure task was assigned to admin', () => {
+                cy.get('[value="admin"]').should('exist');
+            });
+
+            it('Bulk-change assignees', () => {
+                getBulkActionsMenu().within(() => {
```
bsekachev, line 153: As `getBulkActionsMenu` not only opens menu, but also changes selected item, I am not sure `getBulkActionsMenu` is the best option for the name. Alternatively, you may split it to two functions

tests/cypress/e2e/features/bulk_actions.js
```diff
@@ -0,0 +1,205 @@
...
+            });
+
+            it('Bulk-change state', () => {
+                getBulkActionsMenu().within(() => {
+                    cy.contains(`State (${nobjs})`).click();
+                    cy.get('.cvat-job-item-state').click();
+
+                    // state is new by default, so pick second option (=in progress)
+                    cy.get('.ant-select-selection-search').type('{downarrow}');
+                    cy.get('.ant-select-selection-search').type('{enter}');
```
bsekachev, line 170: I think better to type state as text. 
{downarrow}, {enter} is not readable on the code. And tests depends on their order in UI, what is again not reliable

tests/cypress/e2e/features/bulk_actions.js
```diff
@@ -0,0 +1,205 @@
...
+            cy.get('.cvat-notification-notice-export-job-start')
+                .should('be.visible');
+            cy.closeNotification('.cvat-notification-notice-export-job-start');
+
+            cy.get(':visible:contains("Export is finished")')
+                .should('have.length', nobjs);
+            // cy.contains only yields the first
+            // https://docs.cypress.io/api/commands/contains#Single-Element
+
+            cy.get('.ant-notification-notice').should('not.exist');
```
bsekachev: You waiting while all the jobs are exported and then you waiting while notifications get closed?
What if second job (in theory) is exported after the first notification is already closed? Will it work?

tests/cypress/support/default-specs.js
```diff
@@ -11,16 +11,21 @@ function defaultTaskSpec({
...
         name: taskName,
-        project_id: null,
+        project_id: projectID || null,
         source_storage: { location: 'local' },
         target_storage: { location: 'local' },
     };
+    if (projectID && taskSpec.labels) {
+        delete taskSpec.labels;
+        // can only have labels in one place
+    }
```
bsekachev: Are you sure it should be fixed this way? It feels that the function received incorrect parameters if labels in both projectID and labels are specified. But I do not think the function itself should implicitly fix that

tests/cypress/e2e/features/bulk_actions.js
archibald1418: Assignee fields are all `.cvat-user-search-field`, couldn't add extra class so the best I could do is
```js
cy.get('.cvat-user-search-field').first().within(() => {
   cy.get('[value="admin"]').should('exist');
});
```

tests/cypress/e2e/features/bulk_actions.js
archibald1418: added a `selectAll()`

tests/cypress/e2e/features/bulk_actions.js
archibald1418: ensured no `.cvat-item-selected` exist

tests/cypress/e2e/features/bulk_actions.js
archibald1418: IIRC there's no such option, I cannot type in this field since it's neither an input field nor a search field, you only have selection dropdown here

archibald1418: https://github.com/user-attachments/assets/de343640-8b7d-4abc-9b52-75cae46c49b6

tests/cypress/e2e/features/bulk_actions.js
archibald1418: There shouldn't be a problem since the assertion on line 176 will wait until all of the notifications are gone, the order does not matter. `cy.get` retries until the assertion works

tests/cypress/e2e/features/bulk_actions.js
archibald1418: moved projects.push higher

tests/cypress/e2e/features/bulk_actions.js
archibald1418: added `before`'s with `visits` and checked isolation via `.only` and `.skip`

tests/cypress/e2e/features/bulk_actions.js
archibald1418: refactored to visiting taskID URLs instead

tests/cypress/support/default-specs.js
archibald1418: Agreed, I moved the fix to the caller instead

tests/cypress/e2e/features/bulk_actions.js
bsekachev: Is there any reason to wait for this?

image

bsekachev:
Locally tests are failed often with the following issue:

image
image

Maybe click is too early. Do not you have such issue?

Three of four failed the same reason

image

tests/cypress/e2e/features/bulk_actions.js
archibald1418: I guess not, removing. Any exceptions should just crash the app

archibald1418: It's better to wait, because after navigation it has closed some elements for me, failing the next test. I see no issue in waiting, they disappear quickly

archibald1418: > It's better to wait, because after navigation it has closed some elements for me, failing the next test. I see no issue in waiting, they disappear quickly

Yeah I ran the tests in the loop and that wasn't right. Rewriting this to close the export notifications

archibald1418:
Not having this, but I had this problem in development. I'll try to find a state to stabilize this better. Btw it sometimes happens when you use functions instead of doing it in the same scope. Might work better if using a cy.command instead

archibald1418:
Not getting failures with `.cvat-item-selected`

bsekachev (approved):

PR #9823, opened 2025-09-12
[Helm] Set the value for fsGroup security context for Kvrocks pod

### LINKED ISSUE #9589: problems with kvrocks permissions?

### Steps to Reproduce

any workarraund? Or what have I missed?

```
I20250701 10:08:34.634157 140382387080448 main.cc:144] kvrocks version 2.12.1 (commit 06e24fa0)
E20250701 10:08:34.648941 140382387080448 main.cc:176] Failed to open: failed to create column families: IO error: While open a file for appending: /var/lib/kvrocks/db/LOG: Permission denied
```

```
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: cvat
  namespace: argocd
spec:
  project: default
  source:
    repoURL: https://github.com/cvat-ai/cvat.git
    targetRevision: v2.40.0
    path: helm-chart
    helm:
      values: |
        analytics:
          enabled: true
        nuclio:
          enabled: true
        ingress:
          enabled: true
          hostname: cvat.mydomain.com
          annotations: 
            kubernetes.io/ingress.class: "nginx"
            cert-manager.io/cluster-issuer: "letsencrypt-prod"
          className: "nginx"
          tls: true
          tlsSecretName: ingress-tls-cvat
        cvat:
          kvrocks:
            defaultStorage:
              enabled: true
          backend:
            defaultStorage:
              enabled: true
              storageClassName: csi-cephfs-sc
              accessModes:
                - ReadWriteMany
              size: 20Gi


  destination:
    server: https://kubernetes.default.svc
    namespace: cvat

  syncPolicy:
    automated:
      prune: true
      selfHeal: true
    syncOptions:
      - ServerSideApply=true
      - CreateNamespace=true
```

### Expected Behavior

_No response_

### Possible Solution

_No response_

### Context

_No response_

### Environment

```Markdown

```

azhavoro:
Resolve #9589

### Checklist

### License

### DIFF
--- a/changelog.d/20250916_170406_andrey_set_fsgroup.md
+++ b/changelog.d/20250916_170406_andrey_set_fsgroup.md
@@ -1,0 +1,4 @@
+### Added
+
+- \[Helm\] Set fsGroup for the Kvrocks pod to gid of the user on whom the process is running.
+  (<https://github.com/cvat-ai/cvat/pull/9823>)
--- a/changelog.d/20250916_170406_andrey_set_fsgroup.md
+++ b/changelog.d/20250916_170406_andrey_set_fsgroup.md
@@ -1,4 +1,4 @@
 ### Added
 
-- \[Helm\] Set fsGroup for the Kvrocks pod to gid of the user on whom the process is running.
+- \[Helm\] Set fsGroup for the Kvrocks pod to GID of the user the process runs as
   (<https://github.com/cvat-ai/cvat/pull/9823>)
--- a/helm-chart/templates/cvat_backend/initializer/job.yml
+++ b/helm-chart/templates/cvat_backend/initializer/job.yml
@@ -75,6 +75,3 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
-      securityContext:
-        fsGroup: 1000
-        fsGroupChangePolicy: "OnRootMismatch"
--- a/helm-chart/templates/cvat_backend/initializer/job.yml
+++ b/helm-chart/templates/cvat_backend/initializer/job.yml
@@ -75,3 +75,6 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      securityContext:
+        fsGroup: 1000
+        fsGroupChangePolicy: "OnRootMismatch"
--- a/helm-chart/templates/cvat_backend/worker_annotation/deployment.yml
+++ b/helm-chart/templates/cvat_backend/worker_annotation/deployment.yml
@@ -127,3 +127,6 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      securityContext:
+        fsGroup: 1000
+        fsGroupChangePolicy: "OnRootMismatch"
--- a/helm-chart/templates/cvat_backend/worker_annotation/deployment.yml
+++ b/helm-chart/templates/cvat_backend/worker_annotation/deployment.yml
@@ -127,6 +127,3 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
-      securityContext:
-        fsGroup: 1000
-        fsGroupChangePolicy: "OnRootMismatch"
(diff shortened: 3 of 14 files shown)


helm-chart/templates/cvat_kvrocks/statefulset.yml
```diff
@@ -106,6 +106,9 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      securityContext:
+        fsGroup: 999
+        fsGroupChangePolicy: "OnRootMismatch"
```
SpecLad, line 111: I think we'll need to change `kubeVersion` to >=1.23 for `fsGroupChangePolicy`.

helm-chart/templates/cvat_backend/initializer/job.yml
```diff
@@ -75,3 +75,6 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      securityContext:
+        fsGroup: 1000
+        fsGroupChangePolicy: "OnRootMismatch"
```
SpecLad: Did you check if this works with subdirectory mounts? It's not going to be much use to us if it doesn't.

helm-chart/templates/cvat_backend/worker_annotation/deployment.yml
```diff
@@ -127,3 +127,6 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      securityContext:
+        fsGroup: 1000
+        fsGroupChangePolicy: "OnRootMismatch"
```
SpecLad: Do we actually need this in every pod? I would assume the init job would be sufficient.

helm-chart/templates/cvat_backend/worker_chunks/deployment.yml
```diff
@@ -96,3 +96,6 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      securityContext:
+        fsGroup: 1000
+        fsGroupChangePolicy: "OnRootMismatch"
```
SpecLad: Do you think we could now disable the permission fix containers by default? Seems like this should accomplish the same goal (but in a better way).

helm-chart/templates/cvat_kvrocks/statefulset.yml
azhavoro: Fixed

helm-chart/templates/cvat_backend/initializer/job.yml
azhavoro: In my experiments, this works at the root volume level, regardless of whether subPath is specified during mounting or not. For example, if a volume has two directories, dir1 and dir2, and I mount only dir1 and fsGroup specified, then dir2 will also change its group even though it is not mounted.

helm-chart/templates/cvat_backend/worker_annotation/deployment.yml
azhavoro: I think that in this case part of the startup code will fail in some scenarios, because we are trying to perform several actions that require RW to the file system (creating directories, key, ...) in settings/base.py.

```
    echo "waiting for migrations to complete..."
    while ! ~/manage.py migrate --check; do
        sleep 10
    done
```

helm-chart/templates/cvat_backend/worker_chunks/deployment.yml
azhavoro: disabled

helm-chart/templates/cvat_backend/worker_annotation/deployment.yml
SpecLad: Sure, but is that a problem? Kubernetes will just keep restarting the pod until it works.

I'm just concerned that multiple pods could trigger multiple recursive chgrp operations, making startup even slower than it needs to be.

changelog.d/20250916_170406_andrey_set_fsgroup.md
```diff
@@ -0,0 +1,4 @@
+### Added
+
+- \[Helm\] Set fsGroup for the Kvrocks pod to gid of the user on whom the process is running.
```
SpecLad:
```suggestion
- \[Helm\] Set fsGroup for the Kvrocks pod to GID of the user the process runs as
```

SpecLad (approved):
LGTM, just some suggestions for the changelog.

SpecLad:
Oh, and update the PR title before merging, please.

Issue #5261, opened 2022-11-09
CVAT server doesn't handle interrupted uploads well


SpecLad:
When I was working on #4839, I checked how the server would handle an upload that was interrupted (e.g. by a network disconnection), and the result was not good. The server tries to read the entire request body into memory at once:

https://github.com/opencv/cvat/blob/e5d01359aa09521c9ca87f0ec77a6dede097211b/cvat/apps/engine/mixins.py#L94

If the upload is interrupted, that raises an exception, and all data that has already been transmitted is discarded. This is obviously suboptimal - the whole point of TUS is to support upload resumption, so if some data has been transmitted successfully, it should be written to the file, and the upload should then be able to resume from that point.

The current behavior is also problematic from another perspective: by reading the entire upload into memory, we might potentially consume an unbounded amount of memory.

I think the right thing to do here is to read the uploaded file in small chunks, and write each chunk to the destination file before handling the next one. That will solve both problems.

SpecLad:
> The current behavior is also problematic from another perspective: by reading the entire upload into memory, we might potentially consume an unbounded amount of memory.

Okay, that wasn't quote accurate: the amount of memory _is_ bounded by the `DATA_UPLOAD_MAX_MEMORY_SIZE` Django setting, which is currently set to 100 MiB. Still, we could avoid consuming even that much by reading the request body in chunks.

SpecLad:
Come to think of it, this also means that we don't accept TUS upload requests with a chunk size greater than 100 MB. Which is not great, but also not terrible, since neither the UI nor the SDK should send chunks bigger than that.

SpecLad:
It looks like this can't be fixed at this moment, because Django insists on buffering the entire request body before calling the view, at least in the ASGI handler: <https://code.djangoproject.com/ticket/33699>.

SpecLad:
> The current behavior is also problematic from another perspective: by reading the entire upload into memory, we might potentially consume an unbounded amount of memory.

This part of the issue was fixed by #10073. I'm going to leave this issue open in case the Django developers change their mind and decide to support streaming request bodies after all.

PR #9824, opened 2025-09-12
Remove the tuspy dependency

SpecLad:
### Motivation and context

The version we're depending on (0.2.5) has a dependency on python-future, which had a CVE filed against it (PythonCharmers/python-future#650). While the CVE itself seems to be bogus (and in fact, tuspy doesn't actually _use_ python-future), it would still be nice to get rid of this dependency to placate various security dependency scanners.

tuspy did remove this dependency in version 1.0.2. However, this version isn't immediately compatible with the various hacks we have applied to marry it to the API client. Instead of adapting them, I chose to throw tuspy out, and reimplement the TUS protocol. I think this is better for a few reasons:

* We won't have to keep adapting the hacks every time we need to update the dependency.

* The SDK can now be installed in an environment together with other packages that have incompatible tuspy requirements.

* The reimplementation turned out to be shorter than the original code.

I also removed the CVAT 2.2 compatibility code that was in `get_offset`, since we're no longer compatible with those versions for other reasons anyway (e.g., they don't have `/api/requests`).

### How has this been tested?

Unit tests.

### Checklist

### License

### DIFF
--- a/cvat-sdk/cvat_sdk/core/uploading.py
+++ b/cvat-sdk/cvat_sdk/core/uploading.py
@@ -4,141 +4,120 @@
 
 from __future__ import annotations
 
+import base64
 import json
+import logging
 import os
 from contextlib import AbstractContextManager
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Optional
 
-import requests
 import urllib3
 
 from cvat_sdk.api_client.api_client import ApiClient, Endpoint
-from cvat_sdk.api_client.exceptions import ApiException
-from cvat_sdk.api_client.rest import RESTClientObject
+from cvat_sdk.core.exceptions import CvatSdkException
 from cvat_sdk.core.helpers import StreamWithProgress, expect_status
 from cvat_sdk.core.progress import NullProgressReporter, ProgressReporter
 
 if TYPE_CHECKING:
     from cvat_sdk.core.client import Client
 
-import tusclient.uploader as tus_uploader
-from tusclient.client import TusClient as _TusClient
-from tusclient.client import Uploader as _TusUploader
-from tusclient.request import TusRequest as _TusRequest
-from tusclient.request import TusUploadFailed as _TusUploadFailed
-
 MAX_REQUEST_SIZE = 100 * 2**20
+MAX_TUS_RETRIES = 3
+TUS_CHUNK_SIZE = 10 * 2**20
 
 
-class _RestClientAdapter:
-    # Provides requests.Session-like interface for REST client
-    # only patch is called in the tus client
+def _upload_with_tus(
+    api_client: ApiClient,
+    create_url: str,
+    metadata: dict[str, str],
+    file_stream: StreamWithProgress,
+    logger: logging.Logger,
+) -> str:
+    common_headers = {**api_client.get_common_headers(), "Tus-Resumable": "1.0.0"}
 
-    def __init__(self, rest_client: RESTClientObject):
-        self.rest_client = rest_client
+    file_stream.seek(0, os.SEEK_END)
+    upload_length = file_stream.tell()
+    assert metadata
 
-    def _request(self, method, url, data=None, json=None, **kwargs):
-        raw = self.rest_client.request(
-            method=method,
-            url=url,
-            headers=kwargs.get("headers"),
-            query_params=kwargs.get("params"),
-            post_params=json,
-            body=data,
-            _parse_response=False,
-            _request_timeout=kwargs.get("timeout"),
-            _check_status=False,
-        )
+    response = api_client.rest_client.POST(
+        create_url,
+        headers={
+            **common_headers,
+            "Upload-Length": str(upload_length),
+            "Upload-Metadata": ",".join(
+                [f"{k} {base64.b64encode(v.encode()).decode()}" for k, v in metadata.items()]
+            ),
+        },
+    )
+    real_filename = response.headers.get("Upload-Filename")
 
-        result = requests.Response()
-        result._content = raw.data
-        result.raw = raw
-        result.headers.update(raw.headers)
-        result.status_code = raw.status
-        result.reason = raw.msg
-        return result
+    if not real_filename:
+        raise CvatSdkException("Server did not send Upload-Filename after upload creation")
 
-    def patch(self, *args, **kwargs):
-        return self._request("PATCH", *args, **kwargs)
+    upload_url = response.headers.get("Location")
+    if upload_url is None:
+        raise CvatSdkException("Server did not send Location after upload creation")
 
+    upload_offset = 0
+    num_errors = 0
 
-class _MyTusUploader(_TusUploader):
-    # Adjusts the library code for CVAT server
-    # Allows to reuse session
+    while True:
+        file_stream.seek(upload_offset)
+        chunk = file_stream.read(TUS_CHUNK_SIZE)
 
-    def __init__(self, *_args, api_client: ApiClient, **_kwargs):
-        self._api_client = api_client
-        super().__init__(*_args, **_kwargs)
-
-    def _do_request(self):
-        self.request = _TusRequest(self)
-        self.request.handle = _RestClientAdapter(self._api_client.rest_client)
         try:
-            self.request.perform()
-            self.verify_upload()
-        except _TusUploadFailed as error:
-            self._retry_or_cry(error)
-
-    @tus_uploader._catch_requests_error
-    def create_url(self):
-        """
-        Return upload url.
-
-        Makes request to tus server to create a new upload url for the required file upload.
-        """
-        headers = self.headers
-        headers["upload-length"] = str(self.file_size)
-        headers["upload-metadata"] = ",".join(self.encode_metadata())
-        resp = self._api_client.rest_client.POST(self.client.url, headers=headers)
-        self.real_filename = resp.headers.get("Upload-Filename")
-        url = resp.headers.get("location")
-        if url is None:
-            msg = "Attempt to retrieve create file url with status {}".format(resp.status_code)
-            raise tus_uploader.TusCommunicationError(msg, resp.status_code, resp.content)
-        return tus_uploader.urljoin(self.client.url, url)
-
-    @tus_uploader._catch_requests_error
-    def get_offset(self):
-        """
-        Return offset from tus server.
-
-        This is different from the instance attribute 'offset' because this makes an
-        http request to the tus server to retrieve the offset.
-        """
-        try:
-            resp = self._api_client.rest_client.HEAD(self.url, headers=self.headers)
-        except ApiException as ex:
-            if ex.status == 405:  # Method Not Allowed
-                # In CVAT up to version 2.2.0, HEAD requests were internally
-                # converted to GET by mod_wsgi, and subsequently rejected by the server.
-                # For compatibility with old servers, we'll handle such rejections by
-                # restarting the upload from the beginning.
-                return 0
-
-            raise tus_uploader.TusCommunicationError(
-                f"Attempt to retrieve offset failed with status {ex.status}",
-                ex.status,
-                ex.body,
-            ) from ex
-
-        offset = resp.headers.get("upload-offset")
-        if offset is None:
-            raise tus_uploader.TusCommunicationError(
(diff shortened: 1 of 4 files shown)


cvat-sdk/cvat_sdk/core/uploading.py
```diff
@@ -4,142 +4,121 @@
...
+                upload_url,
+                headers={
+                    **common_headers,
+                    "Content-Type": "application/offset+octet-stream",
+                    "Upload-Offset": str(upload_offset),
+                },
+                body=chunk,
+                _check_status=False,
+            )
+        except urllib3.exceptions.HTTPError as e:
```
zhiltsov-max, line 100: Was this condition taken from the tuspy implementation? It seems to cover quite a broad class of errors. Shouldn't it be reduced to just 5xx errors?

cvat-sdk/cvat_sdk/core/uploading.py
```diff
@@ -4,142 +4,121 @@
...
-        return result
-
-    def patch(self, *args, **kwargs):
-        return self._request("PATCH", *args, **kwargs)
-
-
-class _MyTusUploader(_TusUploader):
-    # Adjusts the library code for CVAT server
-    # Allows to reuse session
+            if 200 <= response.status < 300:
```
zhiltsov-max: The protocol seems to allow only the 204 status here.

cvat-sdk/cvat_sdk/core/uploading.py
```diff
@@ -4,142 +4,121 @@
...
 
-    def __init__(self, *_args, api_client: ApiClient, **_kwargs):
-        self._api_client = api_client
-        super().__init__(*_args, **_kwargs)
+                # prevent the server from asking us to keep uploading the same chunk
+                if new_upload_offset < upload_offset + len(chunk):
+                    raise CvatSdkException("Server reported unexpected Upload-Offset")
+            else:
+                logger.error("Chunk upload returned status code %d", response.status)
+                num_errors += 1
```
zhiltsov-max: Do we need to support retrying in this case?

cvat-sdk/cvat_sdk/core/uploading.py
```diff
@@ -4,142 +4,121 @@
...
+    file_stream: StreamWithProgress,
+    logger: logging.Logger,
+) -> str:
+    common_headers = {**api_client.get_common_headers(), "Tus-Resumable": "1.0.0"}
+
+    file_stream.seek(0, os.SEEK_END)
+    upload_length = file_stream.tell()
+    assert metadata
+
+    response = api_client.rest_client.POST(
```
zhiltsov-max, line 44: BTW, it seems the server incorrectly misses the `creation` extension in the response headers and also includes the headers supposed for the `OPTIONS` response in every response.

cvat-sdk/cvat_sdk/core/uploading.py
```diff
@@ -4,142 +4,121 @@
...
+        if num_errors >= MAX_TUS_RETRIES:
+            raise CvatSdkException("Too many upload errors")
 
-    @tus_uploader._catch_requests_error
-    def create_url(self):
-        """
-        Return upload url.
+        if new_upload_offset is None:
+            response = api_client.rest_client.HEAD(upload_url, headers=common_headers)
+            new_upload_offset = int(response.headers["Upload-Offset"])
```
zhiltsov-max, line 109: This branch looks like dead code at the moment, because each file upload actually starts in this function and the file id returned by the server is either server-generated or matches the input meta `filename` field. If it matches the `filename`, but the file exists, the server returns 409. I think, this code block should either be removed or it could be moved into the handling of the 409 response of the `POST` request to support continuation of task data uploading, but this use case isn't supported well in SDK at the moment anyway.

cvat-sdk/cvat_sdk/core/uploading.py
SpecLad: Yeah, but I don't really see a reason to be picky about it. If a server returned a success status, then presumably the upload succeeded.

cvat-sdk/cvat_sdk/core/uploading.py
SpecLad: > Was this condition taken from the tuspy implementation?

No, but it happens to be equivalent:

https://github.com/tus/tus-py-client/blob/e786fa850f0d8e0a2bd137dd91b5fef7c7b703a7/tusclient/request.py#L103

https://github.com/tus/tus-py-client/blob/e786fa850f0d8e0a2bd137dd91b5fef7c7b703a7/tusclient/uploader/uploader.py#L91-L92

> Shouldn't it be reduced to just 5xx errors?

Note that this doesn't handle HTTP errors at all (despite the confusingly named exception), since I set `_check_status` to `False`. It's supposed to handle network errors.

cvat-sdk/cvat_sdk/core/uploading.py
SpecLad: Sure, why not? E.g. if the backend server crashes and Traefik returns 502 Bad Gateway, this would be the way we handle it.

cvat-sdk/cvat_sdk/core/uploading.py
SpecLad: I'm not sure I understand what you mean. This block is meant to handle cases when the chunk fails to upload, but the server does not return an `Upload-Offset` header. Why is it dead code?

If the server returns a 409 error (but not `Upload-Offset`), then the code will work as-is, no?

cvat-sdk/cvat_sdk/core/uploading.py
zhiltsov-max: I agree about supporting retries on 5xx responses, but not about retrying on the 4xx ones, except maybe 409, because many of the 4xx errors are not supposed to be recoverable.

cvat-sdk/cvat_sdk/core/uploading.py
zhiltsov-max: Ok, I thought the PATCH request is supposed to be atomic, so we can just continue from the place we were before the request if there's an error, and there are no unsuccessful server responses that would result in a changed uploading offset. But I can also see similar logic in the tuspy implementation, maybe it's not an issue.

> If the server returns a 409 error (but not Upload-Offset), then the code will work as-is, no?

It will work if the PATCH request returns 409. But this status can also be returned [after the POST request](https://github.com/cvat-ai/cvat/blob/0e789ecc254ce677be129fffa79d1730e9590a78/cvat/apps/engine/mixins.py#L145-L149). If the first POST request returns the 409 status, this uploading function will fail with an error. My idea is that it's better to check if the file sizes match and then either fail or resume uploading. This should make the client code for task data uploading simple and straightforward.

cvat-sdk/cvat_sdk/core/uploading.py
```diff
@@ -4,142 +4,121 @@
...
-            self._retry_or_cry(error)
+        if num_errors >= MAX_TUS_RETRIES:
+            raise CvatSdkException("Too many upload errors")
 
-    @tus_uploader._catch_requests_error
-    def create_url(self):
-        """
-        Return upload url.
+        if new_upload_offset is None:
+            response = api_client.rest_client.HEAD(upload_url, headers=common_headers)
```
zhiltsov-max, line 108: Why only the PATCH request has handling for network errors and 5xx errors, but other requests do not?

cvat-sdk/cvat_sdk/core/uploading.py
SpecLad: Yeah, I agree. I'll update this.

cvat-sdk/cvat_sdk/core/uploading.py
SpecLad: > Ok, I thought the PATCH request is supposed to be atomic,

FYI, the [protocol spec](https://tus.io/protocols/resumable-upload#patch) explicitly recommends against this:

> The Server SHOULD always attempt to store as much of the received data as possible.

I've actually created #5261 because CVAT does not do this.

> My idea is that it's better to check if the file sizes match and then either fail or resume uploading.

You mean on the server? It seems reasonable to me, but it's not in scope for this PR.

cvat-sdk/cvat_sdk/core/uploading.py
SpecLad: The HEAD requests are idempotent, so they will automatically be retried by urllib3 according to the global configuration.

The POST requests will not be automatically retried (if I understand the documentation correctly...), but this is consistent with how the SDK handles any other modification request (e.g. `POST /api/tasks`). I believe it's also consistent with how the original library behaved.

OTOH, there is special recovery logic for PATCH requests, because it's undesirable to abort a large upload if a random error occurs in the middle, and there is a well-defined recovery algorithm.

cvat-sdk/cvat_sdk/core/uploading.py
zhiltsov-max: > You mean on the server? It seems reasonable to me, but it's not in scope for this PR.

No, on the client, after the POST request.

SpecLad: How would the client know what the previous size was?

zhiltsov-max: From the [HEAD request description](https://tus.io/protocols/resumable-upload#head):
> If the size of the upload is known, the Server MUST include the Upload-Length header in the response.

CVAT [returns this header in the HEAD request response](https://github.com/cvat-ai/cvat/blob/2cbf51d9e90e08f9c426c74074b221c1d4c440d5/cvat/apps/engine/mixins.py#L197-L204), so it could be used to check if the file size matches.

cvat-sdk/cvat_sdk/core/uploading.py
zhiltsov-max: From the [urllib3 implementation](https://github.com/urllib3/urllib3/blob/6aa09f09189d472bcace6de824c8857cb6e64511/src/urllib3/util/retry.py#L184-L189):

```python
    #: Default methods to be used for ``allowed_methods``
    DEFAULT_ALLOWED_METHODS = frozenset(
        ["HEAD", "GET", "PUT", "DELETE", "OPTIONS", "TRACE"]
    )

    #: Default status codes to be used for ``status_forcelist``
    RETRY_AFTER_STATUS_CODES = frozenset([413, 429, 503])
```

So it does some retries with the default config, but not for POST and PATCH requests and not for 502 responses, which we also can randomly receive from the server. I agree that this problem is not specific only to this PR or this function. Given that POST and PATCH methods can change the server, the handling typically lies on the client side, because it's not clear whether it's correct to repeat the request automatically. It looks like the 502 statuses could be included in the list on the SDK level, but not POST and PATCH requests.

The use case I'm thinking about is `DataUploader.upload_files()` function called during task creation from local data, which makes several TUS uploads. There are also several other POST requests during file uploading that can raise recoverable errors. My suggestion is to allow retries for the POST request in this function or to support resuming (discussed in https://github.com/cvat-ai/cvat/pull/9824#discussion_r2367289863), so that the clients could try to call the uploading function for the same task once again.

cvat-sdk/cvat_sdk/core/uploading.py
SpecLad: Okay, but to make the `HEAD` request, you'd need to know the upload URL, which you can only get if the initial `POST` requests succeeds, no?

zhiltsov-max: Ok, seems to be a server-side issue.

cvat-sdk/cvat_sdk/core/uploading.py
SpecLad: Alright, updated.

cvat-sdk/cvat_sdk/core/uploading.py
SpecLad: Sorry, but I think this goes beyond the scope of the PR. My goal is to replicate the existing functionality without the tuspy library, and the library does not (AFAICT) attempt to recover from errors that occur during creation.

It seems like it could be useful to implement some sort of recovery logic for multi-request uploads, but it would be a separate feature.

zhiltsov-max (approved):

PR #9826, opened 2025-09-14
[WIP] add verbose logs

archibald1418:
### Checklist

### License

### DIFF
--- a/.github/workflows/schedule.yml
+++ b/.github/workflows/schedule.yml
@@ -144,7 +144,7 @@ jobs:
 
       - name: REST API and SDK tests
         run: |
-          pytest tests/python/ --alluredir=tests/python/allure-results
+          pytest -vv tests/python/ --alluredir=tests/python/allure-results
           pytest tests/python/ --stop-services
 
           CVAT_ALLOW_STATIC_CACHE="true" pytest tests/python


zhiltsov-max:
@archibald1418, check in these changes https://github.com/cvat-ai/cvat/pull/9757/files#diff-576b4c6552f3050841188805c0670754c035c80c955340b2718e7ec30701aaee (in `utils/dataset_manifest/create.py`) can be useful for the problem with `types` module import.

utils/dataset_manifest/core.py
```diff
@@ -18,8 +18,8 @@
 import av
 from PIL import Image
 
+from ._types import NamedBytesIO
```
zhiltsov-max: I think it's better to change how the `types.py` file is visible to python instead - it should not be visible as the builtin `types`. Consider changing the `PYTHONPATH` instead, similarly to what I tried in the PR linked.

archibald1418: I'm not sure this is a good practice. What if a new file or its imported modules (like `functools`) need  built-in types instead of `.types`? `sys.path.append` fixes this only for one module and doesn't seem to scale well for the whole package. 

Agree that renaming to _`types.py` is not good, but renaming the file to e.g. `interfaces.py` is better and makes the intent clearer

zhiltsov-max: Modifying path inside the script - isn't. However, this issue is only relevant to direct `create.py` calls (as it's the only callable file here) and only because this component isn't organized as a regular module. So by prepending the parent directory into the pythonpath we just pretend it's a normal module called `dataset_manifest`, as it's supposed to be.

orazvaliev: In case of a name conflict with a built-in name usually an additional `_` is added as a suffix, not a prefix BTW. I agree with Oleg, it's better to change a file name than a way how this file is visible to interpreter. Because this may cause side effects

zhiltsov-max: Which side effects?

archibald1418: Like the one that broke the tests - helpers.py imported functools which imported GenericAlias from types which lead to circular import. Basically, any 3rd party library that uses built-in `types` can be broken like this. Also, it creates inconsistency in module behaviour which can lead to e.g. CI and testing issues. Not to mention debugging nightmares like incorrect attribute lookups in logs

zhiltsov-max: So your idea is to never call a python file like an existing library? I think, it's more reasonable to fix the incorrect import instead.

archibald1418: My idea was to rename just one file instead of having to deal with cryptic errors later on when testing and patching this module or code including this module

utils/dataset_manifest/create.py
```diff
@@ -13,7 +13,12 @@
 
 from tqdm import tqdm
 
-from utils import SortingMethod, detect_related_images, is_image, is_video
+if __name__ == "__main__":
+    # fix types.py import
+    base_dir = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
```
SpecLad, line 18: This logic is already present at the bottom of the file - you should delete the unnecessary copy.

PR #9827, opened 2025-09-15
Optimize task creation from CS without manifest

zhiltsov-max:
### Motivation and context

Related: #9757

When a raw images task is created from a CS without a manifest attached, CVAT downloads image headers to get image resolution. This operation can be quite time-consuming for big tasks, but it can be optimized quite simply.

- Improved performance of CS image header downloading and manifest creation ~2-6x

The default chunk size used in the downloader is 64KB. For most image formats the required information is available in the first 1KB, while 64KB (the previous value) can be the size of the whole file. It is tempting to change it to a lower value, e.g. 1500 (the default Ethernet v2 MTU size) and it works fine, except for jpgs that include an embedded thumbnail (preview) image in the header, which can basically be of any size. <s>Probably, this can be implemented with a more advanced JPEG parser. It doesn't seem reasonable to use the reduced chunk size and download the whole image for such images, as such jpegs seem to be quite common, but maybe it can be implemented as an exception just for the jpg format.</s> Now, multiple header sizes are attempted per file.

AWS connections limits are floating and depend on the data filenames. In the worst case, we can expect about 100 connections per prefix, up to infinite in the best case (random prefixes). It's also possible to get throttled by AWS (e.g. 503 Slow Down), it should be handled by the boto library itself. 
https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance.html
https://stackoverflow.com/questions/37432285/maximum-no-of-connections-that-can-be-held-by-s3
https://repost.aws/knowledge-center/http-5xx-errors-s3
https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html

Details:
test dataset: 26822 .jpg images (~10 GB)
baseline: 550s
with queue: 320 - 350s
with reduced chunk size (1 MTU): 220 - 280s
with improved connection reuse for AWS: up to 82s (64 connections for 16 cores, up from 10 kept alive with the default config)

### How has this been tested?

<details>

Sample script for testing

```python
from time import perf_counter
from tempfile import TemporaryDirectory

from tqdm import tqdm

from cvat.apps.engine import models, cloud_provider
from utils.dataset_manifest.core import ImageManifestManager


cloud_storage = models.CloudStorage.objects.get(id=yourcs)
storage_client = cloud_provider.db_storage_to_storage_instance(cloud_storage)

media = [v["name"] for v in storage_client.list_files(prefix="images/", _use_flat_listing=True)]

header_downloader = cloud_provider.HeaderFirstMediaDownloader.create(
    models.DimensionType.DIM_2D, client=storage_client
)
content_generator = (
    v
    for v in tqdm(
        storage_client.bulk_download_to_memory(media, object_downloader=header_downloader.download),
        total=len(media),
    )
)


with TemporaryDirectory() as tempdir:
    start_time = perf_counter()

    manifest = ImageManifestManager(tempdir, upload_dir=tempdir, create_index=False)
    manifest.link(
        sources=content_generator,
        stop=len(media) - 1,
        DIM_3D=False,
    )
    manifest.create()

    duration = perf_counter() - start_time
    print(
        f"Manifest for {len(media)} files created in",
        duration,
        "seconds",
        f"avg. {duration / (len(media) or 1)}s.",
    )

# run with  
# cat test_cs_downloading.py | python manage.py shell
```

</details>

### Checklist

### License

### DIFF
--- a/cvat/apps/engine/cloud_provider.py
+++ b/cvat/apps/engine/cloud_provider.py
@@ -397,12 +397,13 @@ class HeaderFirstDownloader(ABC):
 
     def get_header_sizes_to_try(self) -> Sequence[int]:
         return (
-            # Standard Ethernet v2 MTU size, should result in just a several packets
-            # for the first header check
-            1500,
-            # Try bigger sizes, but less than the whole file
-            # TODO: Maybe implement exponential increase
-            15000,
+            # The first 1-2Kb are typically enough for most formats with the static header size.
+            # Unfortunately, it's not enough for some popular formats, such as jpeg,
+            # which can optionally include a preview image embedded in the header, so we try
+            # other bigger sizes, but less than the whole file.
+            # For comparison, the standard Ethernet v2 MTU size is 1500 bytes.
+            2048,
+            16384,
             65536,
         )
 


zhiltsov-max:
@SpecLad, @azhavoro, could you review please?

cvat/apps/engine/cloud_provider.py
```diff
@@ -214,9 +216,21 @@ def bulk_download_to_memory(
...
+                    next_job_params = next(input_iter, None)
+                    if next_job_params is None:
+                        input_iter = None
+                        break
+
+                    next_job = executor.submit(func, next_job_params)
+                    queue.put(next_job)
+
+                top_job = queue.get()
+                yield top_job.result()
```
SpecLad, line 238: I don't think I understand the purpose of this code. How is this different from just `yield from executor.map(func, files)`? The executor already has to maintain a queue of its own; it doesn't seem like it would add anything to have another queue on top of that.

cvat/apps/engine/cloud_provider.py
```diff
@@ -354,26 +368,36 @@ def write_access(self):
...
+        message = (
+            f'The first {header_size} bytes were not enough to parse the "{key}" object header. '
+        )
+
+        if full_file:
+            full_object_size = len(full_file.getvalue())
+
+            message += (
+                f'Object size was {full_object_size} bytes. '
+                f'Downloaded percent was '
```
SpecLad:
```suggestion
                f'Downloaded percentage was '
```

cvat/apps/engine/cloud_provider.py
```diff
@@ -354,26 +368,36 @@ def write_access(self):
...
+                f'{round(min(header_size, full_object_size) / full_object_size):%}'
```
SpecLad: I think this is going to print either 0% or 100%, because of the `round` call. You should use a precision specifier in the format specifier instead, e.g. `.0%`.

cvat/apps/engine/cloud_provider.py
```diff
@@ -354,26 +368,36 @@ def write_access(self):
...
+            )
+
+        slogger.glob.warning(message)
 
-        slogger.glob.warning(
-            f'The {self.chunk_size} bytes were not enough to parse the "{key}" object header. '
-            f'Object size was {full_object_size} bytes. '
-            f'Downloaded percent was '
-            f'{round(min(self.chunk_size, full_object_size) / full_object_size):%}'
+    def get_header_size(self) -> Sequence[int]:
```
SpecLad: This name seems misleading, because it implies a single size, whereas the function actually returns multiple sizes. I would suggest `header_sizes_to_try`, or something along those lines.

cvat/apps/engine/cloud_provider.py
```diff
@@ -354,26 +368,36 @@ def write_access(self):
...
+        return (
+            # Standard Ethernet v2 MTU size, should result in just 1 packet
```
SpecLad: This is a questionable conclusion. A lot more stuff goes into an Ethernet frame than just data:

* IP header
* TCP header
* TLS header
* TLS authentication tag (and sometimes padding)
* HTTP headers

cvat/apps/engine/cloud_provider.py
```diff
@@ -388,34 +412,51 @@ def download(self, key: str) -> NamedBytesIO:
...
+            buff.seek(0)
+
+            if self.try_parse_header(buff):
+                buff.seek(0)
+                return buff
 
+            self.log_header_miss(key=key, header_size=header_size)
+
+        buff = self.client.download_fileobj(key)
+        self.log_header_miss(key=key, header_size=header_size, full_file=buff)
```
SpecLad, line 444: Isn't this going to print a duplicate message? You already call `log_header_miss` at the end of each iteration.

cvat/apps/engine/cloud_provider.py
```diff
@@ -388,34 +412,51 @@ def download(self, key: str) -> NamedBytesIO:
...
 
+        if full_file:
+            full_object_size = len(full_file.getvalue())
+
+            message += (
+                f'Object size was {full_object_size} bytes. '
+                f'Image resolution was {Image.open(full_file).size}. '
+                f'Downloaded percent was '
+                f'{round(min(header_size, full_object_size) / full_object_size):.0%}'
+            )
```
SpecLad, line 466: Seems like there's a lot of duplication here with the superclass method. Why not call `super().log_header_miss` and then log another message with just the resolution?

cvat/apps/engine/cloud_provider.py
```diff
@@ -551,7 +592,16 @@ def __init__(self,
...
         session = boto3.Session(**kwargs)
         self._s3 = session.resource("s3", endpoint_url=endpoint_url,
-            config=Config(proxies=PROXIES_FOR_UNTRUSTED_URLS or {}),
+            config=Config(
+                proxies=PROXIES_FOR_UNTRUSTED_URLS or {},
+                max_pool_connections=(
+                    # AWS can throttle the requests if there are too many of them,
+                    # the SDK handles it with the retry policy:
+                    # https://boto3.amazonaws.com/v1/documentation/api/latest/guide/retries.html
+                    # 10 is the default value
```
SpecLad, line 611: What relation does the default number of retries have with the number of concurrent connections?

cvat/apps/engine/cloud_provider.py
```diff
@@ -551,7 +592,16 @@ def __init__(self,
...
+                    max(10, CPU_NUMBER * settings.CLOUD_DATA_DOWNLOADING_MAX_THREADS_NUMBER_PER_CPU)
```
SpecLad, line 612: Your comment suggests that you're trying to limit the number of connections, but this will set `max_pool_connections` to a _minimum_ of 10 - did you mean the reverse?

cvat/apps/engine/cloud_provider.py
zhiltsov-max: The purpose of this is to improve resource utilization in the parallel execution context. 

In this execution model:
1. enqueue batch (size = thread count * 4)
2. process batch
3. repeat
the threads are utilized only 4 times per batch on average. It roughly means that once a thread finished processing its requests, it's going to stay without work until the batch processing is finished.

In the updated model:
1. enqueue unil full
2. process item if available
3. repeat
the threads are always loaded while there are remaining downloads. 

This optimization alone gives about 60% speedup, as can be seen from the comparison table in the PR description.

cvat/apps/engine/cloud_provider.py
zhiltsov-max: With the increased number of requests, there are increased chances of getting throttled by AWS (the link attached). We don't really want to fail task creation in such cases, so we at least have to consider retries. Next, AWS recommends gradually reducing the number of requests and using some advanced handling strategies when requests are throttled. The comment just says that we don't have to implement the retries ourselves, because the boto library already handles it.

cvat/apps/engine/cloud_provider.py
zhiltsov-max: The default value is 10, we need either the default or more. It's possible to get less using our formula. This max() guarantees at least the same performance as prior the PR.

cvat/apps/engine/cloud_provider.py
SpecLad: I can see how the new code is an improvement on the old code. But what I don't understand is, how is the new code better than what I wrote? It just seems like a more complicated way to accomplish the same result.

cvat/apps/engine/cloud_provider.py
SpecLad: Okay; but if `CPU_NUMBER * settings.CLOUD_DATA_DOWNLOADING_MAX_THREADS_NUMBER_PER_CPU` is less than 10, then why do we need to adjust it? There would be no way for the code to use 10 connections anyway.

cvat/apps/engine/cloud_provider.py
zhiltsov-max: The difference with simple `.map()` is that this implementation has a limited maximum number of jobs in the queue. `.map()` simply puts everything into a single queue and then, in our case with file downloading, all the results are just stored in the memory until collected. Here with a custom queue we limit the maximum memory used by all the downloaded files, which are not processed yet. It's true that in an unfortunate case the downloaded files can explode in the memory even with this implementation, if we download something uncompressed and big, but it was like this before this PR as well, probably needs a separate fix if we ever face this.

cvat/apps/engine/cloud_provider.py
zhiltsov-max: Yeah, good point. Do you have a good estimation of what could fit into 1 packet with http? It seems like we need to talk about 2 packets at least (and it's in the best case scenario with no connection and delivery issues), because the first one is almost guaranteed to be occupied by http headers.

cvat/apps/engine/cloud_provider.py
zhiltsov-max: Given that we are using the same global logger from all the downloading threads, these 2 messages can be interleaved with some other messages. I think the best option is to produce identifiable and full messages so far.

cvat/apps/engine/cloud_provider.py
zhiltsov-max: If we need more and can utilize more, we request more. If we don't need more, we use the default, as if we didn't touch this setting. If we use less, then only these connections will be created and kept, they are created lazily.

cvat/apps/engine/cloud_provider.py
SpecLad: IMO, the number of packets is the wrong thing to focus on. You're never going to get it right, because the underlying technologies can change over time and between deployments (e.g. IP version, TCP options, TLS version, HTTP version...). What we should use is the smallest number of bytes that would be sufficient in the majority of cases (e.g. 90%).

cvat/apps/engine/cloud_provider.py
SpecLad: It seems pointless to add extra logic to request more connections than we actually need; but it's not harmful, so I'll leave it to your discretion.

cvat/apps/engine/cloud_provider.py
SpecLad: Okay, I get it now. Could you add a comment to the code to explain this?

cvat/apps/engine/cloud_provider.py
zhiltsov-max: Fixed

cvat/apps/engine/cloud_provider.py
zhiltsov-max: I agree with the idea. Updated

cvat/apps/engine/cloud_provider.py
zhiltsov-max: Added

PR #9829, opened 2025-09-16
API tokens UI

klakhov:
### Motivation and context

Provides a card inside 'security' of user profile with a table of API keys, and a form to create/edit them

Depends on #9680, #9835

The UI part of https://github.com/cvat-ai/cvat/issues/4961

### Checklist

### License

### DIFF
--- a/cvat-ui/src/components/profile-page/security-content/api-token-form.tsx
+++ b/cvat-ui/src/components/profile-page/security-content/api-token-form.tsx
@@ -68,7 +68,7 @@ function ApiTokenForm({
             className='cvat-api-token-form'
             initialValues={initialValues}
         >
-            <Typography.Title level={4}>
+            <Typography.Title level={5}>
                 {isEditing ? 'Edit API Token' : 'Create API Token'}
             </Typography.Title>
             <Form.Item
--- a/cvat-ui/src/components/profile-page/security-content/api-tokens-card.tsx
+++ b/cvat-ui/src/components/profile-page/security-content/api-tokens-card.tsx
@@ -13,6 +13,7 @@ import Tag from 'antd/lib/tag';
 import Dropdown from 'antd/lib/dropdown';
 import Modal from 'antd/lib/modal';
 import Text from 'antd/lib/typography/Text';
+import Title from 'antd/lib/typography/Title';
 import {
     MoreOutlined, PlusOutlined, QuestionCircleOutlined,
 } from '@ant-design/icons';
@@ -220,7 +221,7 @@ function ApiTokensCard(): JSX.Element {
                 title={(
                     <Row className='cvat-security-api-tokens-card-title' justify='space-between'>
                         <Col>
-                            <Text strong>Personal Access Tokens (PATs)</Text>
+                            <Title level={5}>Personal Access Tokens (PATs)</Title>
                             <CVATTooltip
                                 title={(
                                     <Row className='cvat-api-tokens-tooltip-inner'>
@@ -273,7 +274,7 @@ function ApiTokensCard(): JSX.Element {
                     />
                 ) : (
                     <CVATTable
-                        tableTitle='Existing Tokens'
+                        tableTitle={<Title level={5}>Existing Tokens</Title>}
                         className='cvat-api-tokens-table'
                         csvExport={{ filename: 'access_tokens.csv' }}
                         columns={apiTokenColumns}
--- a/cvat-ui/src/components/profile-page/styles.scss
+++ b/cvat-ui/src/components/profile-page/styles.scss
@@ -62,8 +62,11 @@
 }
 
 .cvat-security-api-tokens-card {
+    h5 {
+        margin-top: $grid-unit-size;
+    }
+
     .cvat-security-api-tokens-card-title > div:first-child {
-        font-size: 16px;
         display: flex;
         align-items: center;
 


cvat-ui/src/components/profile-page/security-content/api-tokens-card.tsx
zhiltsov-max, line 1: user_page

1. Shouldn't this table use the same layout and buttons as other tables? What about the csv export button?
2. I think we should include some context help for this feature. Maybe like GitHub does or in a `(?)` element that we use in other tooltips. For example, it could be but after the title.

gh_api_keys_page

3. Consider changing colors for permissions: RW -> yellow/orange, RO - as is or maybe the green from RW
image

4. Consider highlighting the `Never` values in the `Expires` column as well. Maybe with the same yellow as for RW or simply add an underline.

The general idea of highlighting here is to warn the user about security of the current configuration.

zhiltsov-max, line 1: new_token

1. Consider adding a pre-generated name with a sequential suffix by default (e.g. "new token 1", ...)
2. Expiration date should be set to now() + 1 year by default

zhiltsov-max, line 1: token_value

1. Consider changing the color of the confirmation button to yellow or orange to make the user be more cautious about the action
2. Rename the button to "I have __securely__ saved my token"
3. Maybe add a download button next to copying

zhiltsov-max, line 1: image

The size of the buttons feels a little bit off in the UI, I don't know why specifically. It's like they aren't aligned with any other elements in CVAT. The current active item should probably be highlighted when hovering the cursor.

zhiltsov-max, line 1: Consider filling the Revoke button in the dialog.

cvat-core/src/server-response-types.ts
```diff
@@ -392,6 +403,18 @@ export interface SerializedInvitationData {
     organization_info: SerializedOrganization;
 }
 
+export interface SerializedApiTokenData {
```
bsekachev:
```suggestion
export interface SerializedApiToken {
```

cvat-core/src/server-request-types.ts
```diff
@@ -0,0 +1,12 @@
...
+// SPDX-License-Identifier: MIT
+
+import { SerializedApiTokenData, SerializedUser } from './server-response-types';
+import { CamelizedV2 } from './type-utils';
+
+export type APIApiTokenSaveFields = Partial<Pick<SerializedApiTokenData, 'name' | 'expiry_date' | 'read_only'>>;
+export type ApiTokenSaveFields = CamelizedV2<APIApiTokenSaveFields>;
+
+export type APIUserSaveFields = Partial<Pick<SerializedUser, 'first_name' | 'last_name'>>;
+export type UserSaveFields = CamelizedV2<APIUserSaveFields>;
```
bsekachev: What do you think of using Modifiable/Mutable/Updatable fields instead?
SaveFields does not exactly reflect the idea IMO

cvat-core/src/api-implementation.ts
```diff
@@ -189,6 +190,38 @@ export default function implementAPI(cvat: CVATCore): CVATCore {
...
+    });
+
+    implementationMixin(cvat.apiTokens.update, async (id, tokenData) => {
+        const result = await serverProxy.apiTokens.update(id, tokenData);
+        return new ApiToken(result);
+    });
+
+    implementationMixin(cvat.apiTokens.revoke, async (id) => {
+        await serverProxy.apiTokens.revoke(id);
+    });
```
bsekachev: I believe `update` and `revoke` should be implemented as methods of ApiToken class. 
I see you already call `serverProxy::update` from `save` method.

cvat-core/src/api-token.ts
```diff
@@ -0,0 +1,137 @@
...
+export default class ApiToken {
+    #id: number;
+    #name: string;
+    #createdDate: string;
+    #updatedDate: string;
+    #expiryDate: string | null;
+    #lastUsedDate: string | null;
+    #readOnly: boolean;
+    #owner: number;
+    #value?: string;
```
bsekachev: Until initial create, some fields might be `undefined`.  E.g. see how you initiate the object in action creator.

So, some declared types seem not to be valid

cvat-core/src/api-token.ts
```diff
@@ -0,0 +1,137 @@
...
+
+        if (this.#lastUsedDate) {
+            result.last_used_date = this.#lastUsedDate;
+        }
+
+        if (this.#value !== undefined) {
+            result.value = this.#value;
+        }
+
+        return result as SerializedApiTokenData;
```
bsekachev: From implementation I see it returns `Partial<SerializedApiTokenData>`, not `SerializedApiTokenData`. 
Declaration is not correct

cvat-core/src/api-token.ts
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+    get isExpired(): boolean {
+        if (!this.#expiryDate) {
+            return false;
+        }
+        return new Date(this.#expiryDate) < new Date();
+    }
+
+    get isNew(): boolean {
```
bsekachev: unused?

cvat-core/src/api-token.ts
```diff
@@ -0,0 +1,137 @@
...
+
+    get owner(): number {
+        return this.#owner;
+    }
+
+    get value(): string | undefined {
+        return this.#value;
+    }
+
+    get isExpired(): boolean {
```
bsekachev: unused?

cvat-core/src/api.ts
```diff
@@ -178,6 +179,24 @@ function build(): CVATCore {
...
+                return result;
+            },
+            async revoke(id) {
+                const result = await PluginRegistry.apiWrapper(cvat.apiTokens.revoke, id);
+                return result;
+            },
+            async self() {
+                const result = await PluginRegistry.apiWrapper(cvat.apiTokens.self);
+                return result;
+            },
```
bsekachev: Unused and not implemented

cvat-core/src/api.ts
```diff
@@ -178,6 +179,24 @@ function build(): CVATCore {
...
         },
+        apiTokens: {
+            async get(filter = {}) {
+                const result = await PluginRegistry.apiWrapper(cvat.apiTokens.get, filter);
+                return result;
+            },
+            async update(id, tokenData) {
+                const result = await PluginRegistry.apiWrapper(cvat.apiTokens.update, id, tokenData);
+                return result;
+            },
```
bsekachev: Seems unused

cvat-core/src/server-proxy.ts
```diff
@@ -574,6 +576,74 @@ async function authenticated(): Promise<boolean> {
...
+
+    let response = null;
+    try {
+        response = await Axios.get(`${backendAPI}/auth/api_tokens/self`);
+    } catch (errorData) {
+        throw generateError(errorData);
+    }
+
+    return response.data;
+}
```
bsekachev: unused

cvat-ui/src/actions/auth-actions.ts
```diff
@@ -5,8 +5,11 @@
 
 import { ActionUnion, createAction, ThunkAction } from 'utils/redux';
 import { RegisterData } from 'components/register-page/register-form';
-import { getCore, User } from 'cvat-core-wrapper';
+import {
+    getCore, User, ApiToken, ApiTokenSaveFields,
+} from 'cvat-core-wrapper';
 import { ChangePasswordData } from 'reducers';
+import { APIApiTokensFilter, SerializedApiTokenData } from 'cvat-core/src/server-response-types';
```
bsekachev: Types should be imported via cvat-core-wrapper to avoid potential mess if we change code in cvat-core

cvat-ui/src/actions/auth-actions.ts
```diff
@@ -64,6 +79,20 @@ export const authActions = {
     updateUser: () => createAction(AuthActionTypes.UPDATE_USER),
     updateUserSuccess: (user: User) => createAction(AuthActionTypes.UPDATE_USER_SUCCESS, { user }),
     updateUserFailed: (error: unknown) => createAction(AuthActionTypes.UPDATE_USER_FAILED, { error }),
+    getApiTokens: () => createAction(AuthActionTypes.GET_API_TOKENS),
+    getApiTokensSuccess: (tokens: ApiToken[]) => createAction(AuthActionTypes.GET_API_TOKENS_SUCCESS, { tokens }),
+    getApiTokensFailed: (error: any) => createAction(AuthActionTypes.GET_API_TOKENS_FAILED, { error }),
```
bsekachev: Everywhere (here and lines below) should be `error: Error` according to `ErrorState` interface and your code `reason: action.payload.error`

cvat-core/src/server-proxy.ts
```diff
@@ -574,6 +576,74 @@ async function authenticated(): Promise<boolean> {
     return true;
 }
 
+async function getApiTokens(filter: APIApiTokensFilter = {}): Promise<PaginatedResource<SerializedRequest>> {
```
bsekachev, line 561: The filter supports pagination and, nevertheless, always returns all results. 
Do you see any reasons why we cannot paginate it?

cvat-ui/src/actions/auth-actions.ts
```diff
@@ -64,6 +79,20 @@ export const authActions = {
     updateUser: () => createAction(AuthActionTypes.UPDATE_USER),
     updateUserSuccess: (user: User) => createAction(AuthActionTypes.UPDATE_USER_SUCCESS, { user }),
     updateUserFailed: (error: unknown) => createAction(AuthActionTypes.UPDATE_USER_FAILED, { error }),
+    getApiTokens: () => createAction(AuthActionTypes.GET_API_TOKENS),
+    getApiTokensSuccess: (tokens: ApiToken[]) => createAction(AuthActionTypes.GET_API_TOKENS_SUCCESS, { tokens }),
```
bsekachev: Probably should return PaginatedResource, not just Array

cvat-ui/src/actions/auth-actions.ts
```diff
@@ -199,3 +228,76 @@ export const updateUserAsync = (
...
+        const tokens = await cvat.apiTokens.get(filter);
+        dispatch(authActions.getApiTokensSuccess(tokens));
+    } catch (error) {
+        dispatch(authActions.getApiTokensFailed(error));
+    }
+};
+
+export const createApiTokenAsync = (
+    tokenData: ApiTokenSaveFields,
+    onSuccess?: (token: ApiToken) => void,
```
bsekachev: Should it be optional if the only usage of this action created has the argument?

cvat-ui/src/actions/auth-actions.ts
```diff
@@ -199,3 +228,76 @@ export const updateUserAsync = (
...
+            name: tokenData.name,
+            expiry_date: tokenData.expiryDate,
+            read_only: tokenData.readOnly,
+        };
+
+        let token = new cvat.classes.ApiToken(data);
+        token = await token.save();
+
+        dispatch(authActions.createApiTokenSuccess(token));
+        if (onSuccess) onSuccess(token);
```
bsekachev:
```suggestion
        if (onSuccess) {
            onSuccess(token);
        }
```

cvat-ui/src/components/profile-page/security-content/api-tokens-card.tsx
```diff
@@ -0,0 +1,259 @@
...
+
+    const onSubmitTokenForm = async (data: ApiTokenSaveFields): Promise<void> => {
+        if (editingToken) {
+            dispatch(updateApiTokenAsync(editingToken, data, () => {
+                setShowCreateTokenForm(false);
+                setEditingToken(null);
+                dispatch(getApiTokensAsync());
+            }));
+        } else {
+            dispatch(createApiTokenAsync(data, (token) => {
```
bsekachev, line 110: Unhandled promise rejection if `createApiTokenAsync` fails

cvat-ui/src/actions/auth-actions.ts
bsekachev: Additionally, if you re-raise the error in throw, do you think we need this argument at all? We can just do .then() after dispatch

cvat-ui/src/actions/auth-actions.ts
```diff
@@ -199,3 +228,76 @@ export const updateUserAsync = (
...
+
+        return token;
+    } catch (error) {
+        dispatch(authActions.createApiTokenFailed(error));
+        throw error;
+    }
+};
+
+export const updateApiTokenAsync = (
```
bsekachev, line 274: The same comments about onSuccess like in createApiTokenAsync

cvat-ui/src/actions/auth-actions.ts
```diff
@@ -199,3 +228,76 @@ export const updateUserAsync = (
...
+        return token;
+    } catch (error) {
+        dispatch(authActions.updateApiTokenFailed(error));
+        throw error;
+    }
+};
+
+export const revokeApiTokenAsync = (
+    id: number,
+    onSuccess?: () => void,
```
bsekachev: The same comments about onSuccess like in createApiTokenAsync

cvat-ui/src/components/profile-page/security-content/api-tokens-card.tsx
```diff
@@ -0,0 +1,259 @@
...
+                dispatch(revokeApiTokenAsync(token.id, () => {
+                    dispatch(getApiTokensAsync());
+                }));
+            },
+        });
+    }, [dispatch]);
+
+    const onSubmitTokenForm = async (data: ApiTokenSaveFields): Promise<void> => {
+        if (editingToken) {
+            dispatch(updateApiTokenAsync(editingToken, data, () => {
```
bsekachev, line 104: Unhandled promise rejection if `createApiTokenAsync` fails

cvat-ui/src/components/profile-page/security-content/api-tokens-card.tsx
```diff
@@ -0,0 +1,259 @@
...
+
+    const onRevokeToken = useCallback((token: ApiToken): void => {
+        Modal.confirm({
+            title: 'Revoke API Token',
+            content: `Are you sure you want to revoke the token "${token.name}"? This action cannot be undone.`,
+            okText: 'Revoke',
+            okType: 'danger',
+            cancelText: 'Cancel',
+            onOk: () => {
+                dispatch(revokeApiTokenAsync(token.id, () => {
```
bsekachev: Unhandled promise rejection if `revokeApiTokenAsync` fails

bsekachev:
@klakhov 

Please, merge with develop and resolve conflicts

zhiltsov-max:
@klakhov, the server has the updated `owner` field representation in responses - now it's a standard basic user structure instead of just an id.

cvat-ui/src/components/profile-page/security-content/api-tokens-card.tsx
klakhov: 1. Added csv button
2. Added tooltip next to table heading. You can suggest what text is better to put here
3. Changed colors to orange and blue
4. Added underline

klakhov: Done

klakhov: Changed color to 'warning yellow', renamed button. Not sure about download, Ive never seen such feature for just a string

klakhov: Its not something custom, but just default `Menu` component from antd, it was not just used anywhere as navigation element

klakhov: Done

cvat-core/src/server-request-types.ts
klakhov: Updated to Modifiable

cvat-core/src/api-implementation.ts
klakhov: Removed.

cvat-core/src/api-token.ts
klakhov: True, the token can be initialized with name, expiry and a readOnly. Updated

cvat-core/src/api-token.ts
klakhov: Updated implementation and SerializedApiToken type so no conversion is needed

cvat-core/src/api-token.ts
klakhov: Removed

cvat-ui/src/actions/auth-actions.ts
klakhov: The problem with typing error here as `Error` is that we get typescript error in actions in catch section: `Catch clause variable type annotation must be 'any' or 'unknown' if specified`. Should we catch `any` and convert it explicitly to Error?

cvat-ui/src/actions/auth-actions.ts
klakhov: Updated to the same logic we use in other actions (convert to array and pass count as second argument)

cvat-ui/src/actions/auth-actions.ts
klakhov: It was my mistake, I dont think we need to re-throw an Error, removed it.
Made onSuccess required.

cvat-core/src/server-proxy.ts
klakhov: Technically, we could paginate it. But to do that, we’d need to add support for server-side pagination in our `cvat-table` component, which currently expects the full dataset.

The question is: do we really need it? I’d assume the table will have, at most, tens of rows—not hundreds.

cvat-ui/src/components/profile-page/security-content/api-tokens-card.tsx
zhiltsov-max: Regarding the help popup text, I can suggest something like this:

Personal Access Tokens (PATs) is a text values that can be used for authentication instead of the username/email and password pair. They can be used for interaction with CVAT server API using various clients, including custom scripts, CVAT Python SDK, and CVAT CLI. It's possible to configure expiration and allowed actions per token for additional security. Tokens can be created and revoked by the user at any time.

cvat-core/src/server-proxy.ts
bsekachev: So, maybe at least implement pagination on core level now? (to correspond server endpoint schema) and receive all keys in corresponding action in UI module?

cvat-ui/src/actions/auth-actions.ts
bsekachev: In `catch` we should specify `error: unknown` and then convert it to Error with additional check (instaceof `Error`). Or wrap object to Error if not Error.

As this is a common action, maybe we just need utility function for this

cvat-ui/src/actions/auth-actions.ts
bsekachev: onSuccess is still optional here

cvat-ui/src/components/profile-page/profile-content/user-info-card.tsx
```diff
@@ -0,0 +1,120 @@
...
+                layout='vertical'
+                onFinish={onFinish}
+                initialValues={initialValues}
+            >
+                <Row gutter={16}>
+                    <Col span={12}>
+                        <Form.Item
+                            label='First Name'
+                            name='firstName'
+                            rules={[{ required: true, message: 'Please input your first name!' }]}
```
bsekachev: Should have the same validators as on register form

cvat-ui/src/components/profile-page/profile-content/user-info-card.tsx
```diff
@@ -0,0 +1,120 @@
...
+                        >
+                            <Input />
+                        </Form.Item>
+                    </Col>
+                    <Col span={12}>
+                        <Form.Item
+                            label='Last Name'
+                            name='lastName'
+                            rules={[{ required: true, message: 'Please input your last name!' }]}
```
bsekachev: Should have the same validators as on register form

cvat-ui/src/components/profile-page/profile-content/user-info-card.tsx
```diff
@@ -0,0 +1,120 @@
...
+                    label='Email'
+                    name='email'
+                    rules={[
+                        { required: true, message: 'Please input your email!' },
+                        { type: 'email', message: 'Please enter a valid email!' },
+                    ]}
+                >
+                    <Input disabled />
+                </Form.Item>
+                <Form.Item
```
bsekachev, line 95: Should have the same validators as on register form

cvat-ui/src/components/profile-page/security-content/api-token-created-modal.tsx
```diff
@@ -0,0 +1,84 @@
...
+}
+
+function ApiTokenCreatedModal({
+    visible, token, onClose,
+}: Props): JSX.Element {
+    const [copied, setCopied] = useState(false);
+    const { value: tokenValue } = token;
+
+    const handleCopyToClipboard = useCallback(async (): Promise<void> => {
+        toClipboard(tokenValue ?? '');
```
bsekachev: Implementation of `toClipboad` has fallback saying that clipboard API is not available
At the same time this function sets `copied` to true, what displays `Copied!` text. 
Seems confusing

cvat-ui/src/components/profile-page/security-content/api-token-created-modal.tsx
```diff
@@ -0,0 +1,84 @@
...
+        setCopied(true);
+    }, [tokenValue]);
+
+    const handleClose = useCallback((): void => {
+        setCopied(false);
```
bsekachev: `copied` should be reset before opening, not before closing (e.g. visible may be changed outside of this component). This is more about design, rather than current implementation
Maybe to handle this in useEffect

cvat-ui/src/components/profile-page/security-content/api-token-form.tsx
```diff
@@ -0,0 +1,125 @@
...
+    onSubmit: (data: ApiTokenModifiableFields) => void;
+    onCancel: () => void;
+    submitting: boolean;
+    token: ApiToken | null;
+    tokenCount: number;
+}
+
+interface FormData {
+    name: string;
+    expirationDate: moment.Moment | null;
```
bsekachev: moment?

cvat-ui/src/components/profile-page/security-content/api-token-form.tsx
```diff
@@ -0,0 +1,125 @@
...
+                readOnly: values.isReadOnly,
+            });
+        } catch (error) {
+            // Form validation failed
+        }
+    };
+
+    const initialValues = {
+        name: token?.name || (tokenCount === 0 ? 'New token' : `New token ${tokenCount + 1}`),
+        expirationDate: token?.expiryDate ? dayjs(token.expiryDate) : dayjs().add(1, 'year'),
```
bsekachev: does server accepts UTC? dayjs() returns UTC?

cvat-ui/src/components/profile-page/security-content/api-token-form.tsx
```diff
@@ -0,0 +1,125 @@
...
+                    { max: 50, message: 'Token name must not exceed 50 characters' },
+                ]}
+            >
+                <Input placeholder='Enter a descriptive name for this token' allowClear />
+            </Form.Item>
+            <Form.Item
+                className='cvat-api-token-form-expiration-date'
+                label='Expiration Date'
+                name='expirationDate'
+                help='Leave this field empty if you dont want token to expire'
```
bsekachev:
```suggestion
                help='Leave this field empty if you do not want token to expire'
```

cvat-ui/src/components/profile-page/security-content/api-token-form.tsx
zhiltsov-max, line 1: When clicking Edit for an existing token without expiration date, the date is being auto-filled with now + 1 year, like for a new token. It doesn't seem to be correct, I think it should stay as it is for an existing token. Did you see such behavior in other projects?

klakhov: Its not really correct, I dont think editing should have any auto-fills, Ill fix that.

cvat-ui/src/actions/auth-actions.ts
klakhov: Applied. Created `ensureError` util func to do this.

cvat-core/src/server-proxy.ts
klakhov: Added `fetchAllPaginated` on cvat-ui level that calls cvat-core methods with page=500 param till all instances are fetched.

cvat-ui/src/actions/auth-actions.ts
klakhov: Removed optional and excessive checks

cvat-ui/src/components/profile-page/profile-content/user-info-card.tsx
klakhov: Created util file for validators and re-used them on register form and here.

cvat-ui/src/components/profile-page/security-content/api-token-created-modal.tsx
klakhov: Now `toClipboard` returns true if auto copy was successful, then we set up copied.

cvat-ui/src/components/profile-page/security-content/api-token-created-modal.tsx
klakhov: Added useEffect with copied to false when modal is visible.

cvat-ui/src/components/profile-page/security-content/api-token-form.tsx
klakhov: Server uses UTC, dayjs by default works with local time. I suppose we should display date in local time, so added conversion from UTC to local.

cvat-ui/src/components/profile-page/security-content/api-token-form.tsx
klakhov: Fixed, now the field will be empty for such tokens

cvat-ui/src/components/profile-page/security-content/api-token-form.tsx
```diff
@@ -0,0 +1,132 @@
...
+        } catch (error) {
+            // Form validation failed
+        }
+    };
+
+    const getInitialExpirationDate = (): dayjs.Dayjs | null => {
+        if (isEditing) {
+            return token?.expiryDate ? dayjs.utc(token.expiryDate).local() : null;
+        }
+        return dayjs().add(1, 'year');
```
bsekachev, line 55: Both returns UTC time in my case:

console.log(dayjs())
console.log(dayjs.utc())

2025-10-02T05:50:50.410Z
2025-10-02T05:50:50.411Z

The same about .toISOString() method. 

My local time is +3 hours 8:50

However 
console.log(dayjs().isUTC()) // False
console.log(dayjs.utc().isUTC()) // True

Could you please investigate about timezone?

klakhov: What method did you use to display the string timestamp? When I run console.log(dayjs()) and console.log(dayjs().utc()) I see objects in my console (the difference in them is $u flag and hours internal variable first one is with timezone offset second one is without it):
image


According to [dayjs utc plugin docs](https://day.js.org/docs/en/plugin/utc) by default dayjs works in local mode. When we use `.utc()` it marks(just a boolean inside a object) that this object should work in utc mode. We can see that by using `isUTC()`.

When I call `.format()` methods it displays correct info:
```
console.log(dayjs().format());     -> 2025-10-02T12:47:04+03:00
console.log(dayjs.utc().format()); -> 2025-10-02T09:47:04Z
```

As far as I understand `.toISOString()` should always return zero offset UTC. As [docs for default js Date say](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString), so [does dayjs](https://day.js.org/docs/en/display/as-iso-string).
```
console.log(dayjs().toISOString());     -> 2025-10-02T09:47:04Z
console.log(dayjs.utc().toISOString()); -> 2025-10-02T09:47:04Z
```

If we will try to use `.toString()` method we will see the same results for both dates in RFC 7231 format. That happens because dayjs uses `Date` native method `.toUTCString()` [which is always UTC](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toUTCString)
```
console.log(dayjs().toString());     -> Thu, 02 Oct 2025 10:04:50 GMT
console.log(dayjs.utc().toString()); -> Thu, 02 Oct 2025 10:04:50 GMT
```

PR #9855, opened 2025-09-29
Fix `Cannot read properties of undefined (reading 'points')`

klakhov:
### Motivation and context

The `Cannot read properties of undefined (reading 'points')` error occurs in the `initializeWithContour` method when an incorrect `undefined` argument is passed.
This happens in the callback provided by the slice handler to the object selector. The selection callback can be invoked with an empty array, but the slice handler callback always expects at least one element. As a result, the error occurs.

Can be reproduced via different user actions that trigger this chain: `canvasInstance.setup -> UpdateReasons.OBJECTS_UPDATED -> objectSelector.resetSelected`, for example:

1. Enable slice tool
    1. Delete an object via sidebar
    2. Change z layer of an object via sidebar
    3. Enable gamma filter
    4. Add new Z layer
2. Get an error
image
 
### Checklist

### License

### DIFF
--- a/cvat-canvas/src/typescript/sliceHandler.ts
+++ b/cvat-canvas/src/typescript/sliceHandler.ts
@@ -554,8 +554,12 @@ export class SliceHandlerImpl implements SliceHandler {
             }], 'slice');
 
             this.objectSelector.enable(([state]) => {
-                this.objectSelector.disable();
-                initializeWithContour(state);
+                if (state) {
+                    this.objectSelector.disable();
+                    initializeWithContour(state);
+                } else {
+                    this.release();
+                }
             }, { maxCount: 1, shapeType: ['polygon', 'mask'], objectType: ['shape'] });
         } else if (this.enabled && !sliceData.enabled) {
             this.release();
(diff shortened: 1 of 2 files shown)


cvat-canvas/src/typescript/sliceHandler.ts
```diff
@@ -554,8 +554,12 @@ export class SliceHandlerImpl implements SliceHandler {
...
 
             this.objectSelector.enable(([state]) => {
-                this.objectSelector.disable();
-                initializeWithContour(state);
+                if (state) {
+                    this.objectSelector.disable();
+                    initializeWithContour(state);
+                } else {
+                    this.release();
+                }
```
bsekachev: Would not it be more correct just do nothing if `state` is undefined? 

I believe, e.g. changing of z layer or deleting another object after enabling slice might be, in general, a legal operation.

Additionally, if we want to release slice handler, it should be done from notify method as for other staff, to avoid one more implicit way of cancellation

klakhov: Yes, it seems not resetting here would be better option. It doesnt break anything and allow us to delete objects/changing z layer and then continue to select object for slice.

PR #9864, opened 2025-10-02
Link views to permissions rather than the reverse

SpecLad:
### Motivation and context

This solves a number of problems with the current implementation:

* Currently, we iterate over every known permission class every time we check permissions for an API call. This is needlessly inefficient, since there is only one permission class for every endpoint.
* Every permission class has to compare the view name to the expected one. If this check is done incorrectly, the permission class will effectively be ignored, permitting unauthorized access.
* It's impossible to have intermediate classes in the permission inheritance hierarchy.
* If an app fails to call `load_app_permissions`, there's a risk that permission classes won't be loaded. This will also permit unauthorized access.
* It's just inconsistent that for List calls, the view determines the permission class, but for all other calls, the permission class determines the view.

### How has this been tested?

Automatic tests.

### Checklist

### License

### DIFF
--- a/cvat/apps/iam/permissions.py
+++ b/cvat/apps/iam/permissions.py
@@ -312,10 +312,10 @@ class PolicyEnforcer(BasePermission):
                 return True
 
             assert hasattr(
-                view, "opa_permission_class"
-            ), f"View {view} has no 'opa_permission_class' attribute"
+                view, "iam_permission_class"
+            ), f"View {view} has no 'iam_permission_class' attribute"
 
-            perm_class = view.opa_permission_class
+            perm_class = view.iam_permission_class
             iam_context = get_iam_context(request, obj)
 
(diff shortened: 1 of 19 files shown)


cvat/apps/iam/permissions.py
```diff
@@ -362,18 +358,11 @@ def has_object_permission(self, request, view, obj) -> bool:
         )
 
 
-def load_app_permissions(config: AppConfig) -> None:
+def load_app_opa_rules(config: AppConfig) -> None:
```
zhiltsov-max: I don't think that the fact that we use OPA as the permission checking backend should be accentuated. It's an implementation detail and I think it's right. Consider renaming to something without mentioning OPA. The same for `opa_permission_class` - maybe `pe_permission_class` or `iam_permission_class`.

SpecLad: Alright, updated.

PR #9869, opened 2025-10-06
Work around spurious "No module named 'pkg_resources'" errors

SpecLad:
### Motivation and context

Datumaro depends on the pyemd library, which fails to be imported in the CVAT Docker image, because it uses `pkg_resources` without declaring a dependency on setuptools (see also wmayner/pyemd#67).

Normally this isn't a problem, because Datumaro imports it lazily (and the Datumaro component that triggers its actual loading is not used by CVAT). However, the `current_function_name` function calls `inspect.getouterframes`, which indirectly causes every lazy-imported module to be loaded (this behavior might be a Python bug). In particular, this loads pyemd, which raises an exception.

`current_function_name` is usually called while handling another exception, so when this problem occurs, that original exception becomes lost.

Work around this by replacing `inspect.getouterframes` with custom code that does the bare minimum necessary to get the function name.

### How has this been tested?

Manual testing.

### Checklist

### License

### DIFF
--- a/changelog.d/20251006_193945_roman_no_getouterframes.md
+++ b/changelog.d/20251006_193945_roman_no_getouterframes.md
@@ -1,0 +1,5 @@
+### Fixed
+
+- Worked around an issue that may cause an error message to be replaced
+  with "No module named 'pkg_resources'"
+  (<https://github.com/cvat-ai/cvat/pull/9869>)
--- a/cvat/apps/dataset_manager/util.py
+++ b/cvat/apps/dataset_manager/util.py
@@ -27,7 +27,15 @@ from pottery import Redlock
 
 
 def current_function_name(depth=1):
-    return inspect.getouterframes(inspect.currentframe())[depth].function
+    frame = inspect.currentframe()
+    if frame is None:
+        return "[unknown]"
+
+    for _ in range(depth):
+        frame = frame.f_back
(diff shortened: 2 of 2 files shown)


SpecLad:
I reported the Python issue as python/cpython#139669.

zhiltsov-max:
It seems like package dependencies should be fixed on the Datumaro side as well, unless pyemd is going to merge their PR soon. The Datumaro dependencies have to be updated anyway to be compatible with python 3.12 [deprecations](https://docs.python.org/3/whatsnew/3.12.html), if the support is [claimed](https://github.com/cvat-ai/datumaro/blob/26d5f3508e80e40d81034b8c504e4790bdc74b65/setup.py#L83). However, I just checked the upstream - and it has [just removed](https://github.com/open-edge-platform/datumaro/releases/tag/v1.12.0) the component that used pyemd, so maybe we should just remove it as well, given that we're not dependent on it.

SpecLad:
> However, I just checked the upstream - and it has [just removed](https://github.com/open-edge-platform/datumaro/releases/tag/v1.12.0) the component that used pyemd, so maybe we should just remove it as well, given that we're not dependent on it.

Yeah, I think we should. The pyemd project seems to be dead, so I'd be glad to get rid of it.

PR #9880, opened 2025-10-09
SDK: Make the default retry logic more robust

SpecLad:
### Motivation and context

1. Add some typical transient server error status codes to the `status_forcelist` setting to enable retries for them. This will make the SDK more reliable when dealing with a flaky server. urllib3 will only retry requests with idempotent methods, so this seems reasonably safe.

   By default, urllib3 will already retry on 503, but only if the `Retry-After` header is set; including it in `status_forcelist` will make the retries unconditional.

   The 500 code is explicitly not included, as it typically indicates an unhandled exception on the server. Retrying the request in this case is likely to just trigger the exception again.

2. Add delay settings to prevent the SDK hammering the server with retries. Note that urllib3 will use the `Retry-After` header instead if it's available.

### Checklist

### License

### DIFF
--- a/changelog.d/20251009_135408_roman_sdk_retry.md
+++ b/changelog.d/20251009_135408_roman_sdk_retry.md
@@ -1,0 +1,4 @@
+### Changed
+
+- \[SDK\] Enabled retrying on some server error statuses by default 
+  (<https://github.com/cvat-ai/cvat/pull/9880>)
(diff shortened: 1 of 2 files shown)


SpecLad:
> LGTM, consider adding a simple test for this

I don't really see the point of testing this TBH; all the logic is inside urllib3, and they already have tests.

PR #9888, opened 2025-10-10
Limit of jobs per task

skar404:
### Motivation and context

This change introduces a maximum limit on the number of jobs in a single task. The restriction is enforced on the server side and is applicable to all new tasks. This is implemented primarily as a system stability measure to avoid tasks that cannot be processed by the platform in a reasonable amount of time with standard resources.

The limit doesn't affect Ground Truth jobs at the moment. It's only possible to have 1 GT job in a task and restricting this along with other job types could negatively affect the user experience for no actual system stability gains.

- Added a new server setting that specifies the maximum number of jobs per task. It can be set via the `CVAT_MAX_JOBS_PER_TASK` environment variable.

### How has this been tested?

Unit tests.

### Checklist

### License

### DIFF
--- a/cvat/apps/engine/tests/test_rest_api.py
+++ b/cvat/apps/engine/tests/test_rest_api.py
@@ -7914,7 +7914,7 @@ class TaskJobLimitAPITestCase(ApiTestBase):
             response = self.client.get(f"/api/requests/{rq_id}")
             return response
 
-    def _create_gt_task(self, task_id: int):
+    def _create_gt_job(self, task_id: int):
         data = {
             "type": "ground_truth",
             "task_id": task_id,
@@ -7954,7 +7954,7 @@ class TaskJobLimitAPITestCase(ApiTestBase):
         self.assertEqual(job_count, 0)
 
     @override_settings(MAX_JOBS_PER_TASK=10)
-    def test_create_task_gt_job_with_extra_limit(self):
+    def test_gt_jobs_are_not_affected_by_job_limit(self):
         response = self._create_task(
             segment_size=10,
             img_size=100,
@@ -7963,7 +7963,7 @@ class TaskJobLimitAPITestCase(ApiTestBase):
         self.assertEqual(response.data["status"], RequestStatus.FINISHED)
 
         task_id = Task.objects.latest("id").id
-        response = self._create_gt_task(task_id)
+        response = self._create_gt_job(task_id)
         self.assertEqual(response.status_code, status.HTTP_201_CREATED)
 
         job_count = Job.objects.filter(segment__task_id=task_id).count()


cvat/apps/engine/task.py
```diff
@@ -174,6 +174,10 @@ def _create_segments_and_jobs(
     db_task.segment_size = segment_size
     db_task.overlap = overlap
 
+    segments_size = db_task.data.size
+    if segments_size > settings.MAX_JOBS_PER_TASK:
```
zhiltsov-max: This check is incorrect. It has to check the number of segments, not their sizes and not task size.

changelog.d/20251014_150201_denis_limit_of_jobs_per_task.md
```diff
@@ -0,0 +1,4 @@
+### Fixed
+
+- Implemented job limit per task
+  (<https://github.com/cvat-ai/cvat/pull/9888>)
```
zhiltsov-max:
```suggestion
### Added

- \[Server\] A configuration option to set maximum job limit per task
  (<https://github.com/cvat-ai/cvat/pull/9888>)
```

cvat/apps/engine/task.py
```diff
@@ -69,6 +69,7 @@ class SegmentsParams(NamedTuple):
     segments: Iterator[SegmentParams]
     segment_size: int
     overlap: int
+    job_size: Optional[int] = None
```
zhiltsov-max: Why not `segments_count`?

cvat/apps/engine/task.py
```diff
@@ -148,17 +150,20 @@ def _segments():
                 else 5 if db_task.mode == 'interpolation' else 0,
             segment_size // 2,
         )
+        segments_range = range(0, data_size - overlap, segment_size - overlap)
+        job_size = len(list(segments_range))
```
```suggestion
        job_size = len(segments_range)
```
There's no need to convert `range` into a `list`.

cvat/apps/engine/tests/test_job_limit.py
zhiltsov-max, line 1: Typically we don't write tests at this low level (I mean functions and classes), unless absolutely necessary. Please try to add at least 1 test that tries to create a task via an endpoint call. You can find examples in `cvat/apps/engine/tests/test_rest_api.py`.

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -7867,3 +7868,155 @@ def test_can_not_change_to_not_available_cloud_storage(self):
...
+
+        image_data = {
+            **image_files,
+            "image_quality": 75,
+        }
+
+        response = self._create_task(data, image_data)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.data["status"], RequestStatus.FAILED)
+
```
zhiltsov-max, line 7951: I think it makes sense to also check the error message. There are quite a few reasons why task creation could fail, we need to make sure the reason is what we expect.

skar404: @zhiltsov-max ok, added error message checking

cvat/apps/engine/task.py
```diff
@@ -69,6 +69,7 @@ class SegmentsParams(NamedTuple):
     segments: Iterator[SegmentParams]
     segment_size: int
     overlap: int
+    segments_count: Optional[int] = None
```
zhiltsov-max: Why it is `Optional`? If it can be `None`, the code below is incorrect at the moment. So far, I don't see reasons to make allow the `None` value here.

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -25,7 +25,7 @@
 from pathlib import Path
 from pprint import pformat
 from time import sleep
-from typing import BinaryIO
+from typing import BinaryIO, Optional
```
zhiltsov-max:
```suggestion
from typing import BinaryIO
```

Since python 3.10, it's allowed to write `x | None`. We're using this syntax on the server side. Actually, after the 3.9 EOL in 2 weeks, it can safely be used in all other components as well.

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -77,6 +77,7 @@
 from utils.dataset_manifest import ImageManifestManager, VideoManifestManager
 from utils.dataset_manifest.utils import PcdReader, find_related_images
 
+from ...redis_handler.serializers import RequestStatus
```
zhiltsov-max: Please use absolute imports when importing from other apps.

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -7867,3 +7868,119 @@ def test_can_not_change_to_not_available_cloud_storage(self):
...
+    def setUp(self):
+        super().setUp()
+        self.error_message = "Too many jobs would be created for the task"
+
+    @classmethod
+    def setUpTestData(cls):
+        create_db_users(cls)
+
+    def _create_task(
+        self, segment_size: int, img_size: int, consensus_replicas: Optional[int] = None
```
zhiltsov-max: `img_size` is quite a confusing name in this context. I suggest `images_count` as an alternative.

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -7867,3 +7868,119 @@ def test_can_not_change_to_not_available_cloud_storage(self):
...
+
+            tid = response.data["id"]
+            response = self.client.post(f"/api/tasks/{tid}/data", data=image_data)
+
+            rq_id = response.data["rq_id"]
+            response = self.client.get(f"/api/requests/{rq_id}")
+            return response
+
+    @override_settings(MAX_JOBS_PER_TASK=5)
+    def test_create_task_within_job_limit(self):
```
zhiltsov-max, line 7930: This test currently checks the same case as `test_create_task_exactly_at_job_limit`.

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -7867,3 +7868,119 @@ def test_can_not_change_to_not_available_cloud_storage(self):
                 response.status_code,
                 response.content,
             )
+
+
+class TaskJobLimitAPITestCase(ApiTestBase):
```
zhiltsov-max, line 7873: Please add a test that GT jobs are not counted

skar404: created test for GT jobs

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -7867,3 +7868,132 @@ def test_can_not_change_to_not_available_cloud_storage(self):
...
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.data["status"], RequestStatus.FAILED)
+        self.assertIn(self.error_message, response.data["message"])
+
+        task_id = Task.objects.latest("id").id
+        job_count = Job.objects.filter(segment__task_id=task_id).count()
+        self.assertEqual(job_count, 0)
+
+    @override_settings(MAX_JOBS_PER_TASK=10)
+    def test_create_task_gt_job_with_extra_limit(self):
```
zhiltsov-max:
```suggestion
    def test_gt_jobs_are_not_affected_by_job_limit(self):
```

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -7867,3 +7868,132 @@ def test_can_not_change_to_not_available_cloud_storage(self):
...
+                return response
+
+            tid = response.data["id"]
+            response = self.client.post(f"/api/tasks/{tid}/data", data=image_data)
+
+            rq_id = response.data["rq_id"]
+            response = self.client.get(f"/api/requests/{rq_id}")
+            return response
+
+    def _create_gt_task(self, task_id: int):
```
zhiltsov-max:
```suggestion
    def _create_gt_job(self, task_id: int):
```
There are only GT jobs in CVAT

zhiltsov-max (approved):

PR #9905, opened 2025-10-14
Ordering jobs for task annotations

Eldies:
### Motivation and context

Ordering jobs to make track merging work correctly.

Fixes https://github.com/cvat-ai/cvat/issues/9860

### Checklist

### License

### DIFF
--- a/cvat/apps/dataset_manager/tests/test_annotation.py
+++ b/cvat/apps/dataset_manager/tests/test_annotation.py
@@ -699,9 +699,9 @@ class TestTaskAnnotation(TestCase):
             "segment_size": 11,
         }
         db_task = models.Task.objects.create(data=db_data, **data)
-        
-        # We assume that normally segments and annotation jobs 
-        # are created in the ascending order for start_frame, 
+
+        # We assume that normally segments and annotation jobs
+        # are created in the ascending order for start_frame,
         # so their ids correspond to this order. The DB, however,
         # can return them in an arbitrary order, if not specified explicitly.
         # This test tries to reproduce this by specifying job ids.


zhiltsov-max:
Could you add a test for the regression?

Eldies:
> Could you add a test for the regression?

I don't think it is possible. If the jobs are not ordered explicitly, if their order is undefined, for any possible test there is a chance it randomly will not fail in case of regression.

~~I added an assert to check for the correct order in the merge function.~~
I tried to add an assert to merge function to ensure correct order of jobs, but then it fails on scenarios other then export

Eldies:
Added a test which tests that JobAnnotations are initialised in ascending order of job ids

cvat/apps/dataset_manager/tests/test_annotation.py
```diff
@@ -679,6 +681,60 @@ def test_slice_track_does_not_duplicate_outside_frame_on_the_end(self):
...
+
+        db_data = models.Data.objects.create(size=200, image_quality=50)
+
+        data = {
+            "name": "my task",
+            "owner": user,
+            "overlap": 0,
+            "segment_size": 100,
+        }
+        db_task = models.Task.objects.create(data=db_data, **data)
```
zhiltsov-max:
```suggestion
        db_data = models.Data.objects.create(size=31, stop_frame=30, image_quality=50)

        data = {
            "name": "my task",
            "owner": user,
            "overlap": 1,
            "segment_size": 11,
        }
        db_task = models.Task.objects.create(data=db_data, **data)
        
        # We assume that normally segments and annotation jobs 
        # are created in the ascending order for start_frame, 
        # so their ids correspond to this order. The DB, however,
        # can return them in an arbitrary order, if not specified explicitly.
        # This test tries to reproduce this by specifying job ids.
        # https://github.com/cvat-ai/cvat/issues/9860
```

- The values specified are not correct for these jobs (stop_frame is included in the segment)
- Specify the preconditions why this is expected to be working and why this does actually test the problem being fixed

PR #9922, opened 2025-10-20
Create a squashed migration for all engine migrations from release v2.0.0

SpecLad:
### Motivation and context

This will permit us to later drop those migrations, thus easing maintenance burden at the cost of no longer supporting direct upgrades from pre-v2 versions. This patch doesn't drop anything by itself, though.

I'm using v2 as the cutoff, since most of the really complicated migrations happened before then, and also it's a pretty old release at this point.

### Checklist

### License

### DIFF
--- a/cvat/apps/engine/migrations/0001_squashed_0051_auto_20220220_1824.py
+++ b/cvat/apps/engine/migrations/0001_squashed_0051_auto_20220220_1824.py
@@ -193,7 +193,7 @@ class Migration(migrations.Migration):
                         on_delete=django.db.models.deletion.CASCADE, to="engine.label"
                     ),
                 ),
-                ("default_value", models.CharField(default="", max_length=128)),
+                ("default_value", models.CharField(max_length=128)),
                 (
                     "input_type",
                     models.CharField(
@@ -207,9 +207,9 @@ class Migration(migrations.Migration):
                         max_length=16,
                     ),
                 ),
-                ("mutable", models.BooleanField(default=True)),
-                ("name", models.CharField(default="test", max_length=64)),
-                ("values", models.CharField(default="", max_length=4096)),
+                ("mutable", models.BooleanField()),
+                ("name", models.CharField(max_length=64)),
+                ("values", models.CharField(max_length=4096)),
             ],
             options={
                 "default_permissions": (),


cvat/apps/engine/migrations/0001_squashed_0051_auto_20220220_1824.py
```diff
@@ -0,0 +1,1144 @@
...
+            name="size",
+        ),
+        migrations.RemoveField(
+            model_name="task",
+            name="start_frame",
+        ),
+        migrations.RemoveField(
+            model_name="task",
+            name="stop_frame",
+        ),
```
zhiltsov-max, line 660: It seems there are several task fields that are first added and then removed, and they can probably be safely removed. Was this migration created with `--no-optimize`?

cvat/apps/engine/migrations/0001_squashed_0051_auto_20220220_1824.py
```diff
@@ -0,0 +1,1144 @@
+# Generated by Django 4.2.25 on 2025-10-20 15:38
+
+import django.db.models.deletion
+from django.conf import settings
+from django.db import migrations, models
+
+import cvat.apps.engine.models
```
zhiltsov-max, line 7: I thought one of the reasons was to try to avoid dependencies on the current symbols in the code, but it looks some of them are still here. Could you check if it's possible to avoid them too? Specifically, non-enum defaults. Maybe [this](https://docs.djangoproject.com/en/5.2/topics/migrations/#custom-serializers) could help.

cvat/apps/engine/migrations/0001_squashed_0051_auto_20220220_1824.py
SpecLad: No, and it actually did optimize quite a bit:

```
Optimizing...
  Optimized from 326 operations to 43 operations.
```

I guess it just couldn't manage to eliminate these particular operations.

If you really want, I can try to optimize this manually, but it doesn't seem that important to me. Removing columns from an empty table isn't going to take much time.

zhiltsov-max: Ok. I'd prefer using the autogenerated variant for safety.

cvat/apps/engine/migrations/0001_squashed_0051_auto_20220220_1824.py
SpecLad: I think the proper long-term solution for this is to use `TextChoices` instead of enums. Django has a built-in serializer for it.

However, it seems that we can just retroactively edit the migrations to replace the enum values with string literals. There don't seem to be any negative side effects from this. As such, there's no urgency to fix this problem.

PR #9925, opened 2025-10-21
publish helm

yaizkazani:
- [x] I submit _my code changes_ under the same [MIT License]

### DIFF
--- a/helm-chart/Chart.yaml
+++ b/helm-chart/Chart.yaml
@@ -16,7 +16,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 2.48.1
+version: 2.48.2
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to


.github/workflows/publish-artifacts.yml
```diff
@@ -53,3 +53,31 @@ jobs:
...
+      - name: Set up Helm
+        uses: azure/setup-helm@1a275c3b69536ee54be43f2070a358922e12c8d4
+        with:
+          version: v3.14.4
+       # v3
+      - name: Log in to Docker Hub
+        uses: docker/login-action@343f7c4344506bcbf9b4de18042ae17996df046d
+        with:
+          username: ${{ secrets.DOCKERHUB_RW_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_RW_TOKEN }}
```
SpecLad: What's the difference between these credentials and the ones without `RW`?

.github/workflows/publish-artifacts.yml
```diff
@@ -53,3 +53,31 @@ jobs:
...
+    runs-on: ubuntu-latest
+    needs: docker-images
+    steps:
+      - uses: actions/checkout@v4
+       # v4.3.1
+      - name: Set up Helm
+        uses: azure/setup-helm@1a275c3b69536ee54be43f2070a358922e12c8d4
+        with:
+          version: v3.14.4
+       # v3
```
SpecLad: v3 is a floating tag; please add the specific version.

.github/workflows/publish-artifacts.yml
```diff
@@ -53,3 +53,31 @@ jobs:
...
+      - name: Log in to Docker Hub
+        uses: docker/login-action@343f7c4344506bcbf9b4de18042ae17996df046d
+        with:
+          username: ${{ secrets.DOCKERHUB_RW_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_RW_TOKEN }}
+
+      - name: Package Helm chart
+        working-directory: helm-chart
+        run: |
+          helm dependency update && helm package . --version ${{ github.event.release.tag_name }} && mv cvat-*.tgz ../
```
SpecLad: The first command can be replaced with the `--dependency-update` option of `helm package`, and the last command can be replaced with the `--destination` option.

.github/workflows/publish-artifacts.yml
```diff
@@ -53,3 +53,31 @@ jobs:
...
+          echo "CHART_FILE=$(ls ../cvat-*.tgz | xargs basename)" >> $GITHUB_ENV
+
+      - name: Push Helm chart to Docker Hub
+        run: |
+          helm push ${{ env.CHART_FILE }} oci://registry-1.docker.io/cvat
```
SpecLad: The environment variable seems unnecessary; why not just use `cvat-*.tgz`?

changelog.d/20251021_112837_petr.iosipov_DEV_710.md
```diff
@@ -0,0 +1,4 @@
+### Added
+
+- Push of public helm chart to Dockerhub
```
SpecLad: I would suggest something more user-facing, e.g. `Helm charts are now available on Docker Hub, at [...]`.

.github/workflows/publish-artifacts.yml
yaizkazani: I think I've tried to use "without RW" and it has failed so i've created a new one

.github/workflows/publish-artifacts.yml
yaizkazani: v3 is a human readable, and version is locked with SHA

.github/workflows/publish-artifacts.yml
yaizkazani: creds fixed

.github/workflows/publish-artifacts.yml
```diff
@@ -53,3 +53,30 @@ jobs:
 
       - name: Publish to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
+
+
+  helm:
+    runs-on: ubuntu-latest
+    needs: docker-images
```
SpecLad, line 60: Given that this job runs sequentially after the `docker-images` job, and it pushes to the same repository, why not merge the two jobs?

.github/workflows/publish-artifacts.yml
```diff
@@ -53,3 +53,30 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+       # v4.3.1
+      - name: Set up Helm
+        uses: azure/setup-helm@1a275c3b69536ee54be43f2070a358922e12c8d4
+        with:
+          version: v3.14.4
```
SpecLad: This version is more than a year old. Why not use the latest?

.github/workflows/publish-artifacts.yml
```diff
@@ -53,3 +53,30 @@ jobs:
...
+      - name: Log in to Docker Hub
+        uses: docker/login-action@5e57cd118135c172c3672efd75eb46360885c0ef
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Package Helm chart
+        working-directory: helm-chart
+        run: |
+          helm package . --version ${{ github.event.release.tag_name }} --dependency-update --destination ../
```
SpecLad: I don't think we should be overriding the version here; it wouldn't make much sense if a user who installs the chart directly from the Git repository would get a different version from a user who installs it from the Docker repo.

IMO, it would be reasonable to align the Helm version in the repository with the CVAT version; this could be done easily by adding a rule to `dev/update_version.py`. I'd like to hear @azhavoro's opinion on this, though.

.github/workflows/publish-artifacts.yml
```diff
@@ -53,3 +53,30 @@ jobs:
...
+
+      - name: Push Helm chart to Docker Hub
+        run: |
+          helm push cvat-*.tgz oci://registry-1.docker.io/cvat
```
SpecLad:
```suggestion
          helm push cvat-*.tgz oci://registry-1.docker.io/${{ secrets.DOCKERHUB_WORKSPACE }}
```

.github/workflows/publish-artifacts.yml
yaizkazani: This is for CI readability. Previous job pushes docker images, this one pushed Helm chart

.github/workflows/publish-artifacts.yml
yaizkazani: v4.3.1 [Latest](https://github.com/Azure/setup-helm/releases/latest)
@[github-actions](https://github.com/apps/github-actions) github-actions released this Aug 19

.github/workflows/publish-artifacts.yml
yaizkazani: I agree, this could be done, although this:

1. Doesn't change anything in terms of chart usability
2. We'd better have our chart published with same tag as cvat release

To me it looks like just a separate task to update update_version.py

.github/workflows/publish-artifacts.yml
SpecLad: I'm talking about the version of Helm itself.

yaizkazani: PS if you are talking about Helm version - does not matter. The only condition we care about is not to go over 3.17.2 because, as I've told you already, Helm > 3.17.2 will fail to package our chart because they've introduced limits on chart size that we do not fit into
https://github.com/helm/helm/issues/30738

SpecLad: Something isn't right here.

1. I have Helm v3.18.6 on my machine, and it can package the chart just fine, using the exact same command.
2. According to the linked issue, the limits are 5 MiB for any single file, and 100 MiB for the chart as a whole. The chart I get is nowhere near these limits: the chart is under 4 MiB, and the biggest file in it is around 600 KiB.

So why does it fail for you? Is it packaging something it shouldn't?

yaizkazani: Absolutely agree :) 

Seems they have fixed it in latest versions so we'll go with v3.19.0

dev/update_version.py
```diff
@@ -167,6 +167,11 @@ def apply(self, new_version: Version, *, verify_only: bool) -> bool:
         re.compile(r'^  "version": "[\d.]+",$', re.M),
         lambda v, m: f'  "version": "{v.major}.{v.minor}.{v.patch}",',
     ),
+    ReplacementRule(
+        "helm-chart/Chart.yaml",
```
SpecLad: Please move this rule next to the one for `helm-chart/values.yaml`.

yaizkazani: yup, done

SpecLad:
I think you have to update the version number, but otherwise LGTM.

PR #9928, opened 2025-10-22
Clean up cloud provider-related identifiers in the backend code

SpecLad:
### Motivation and context

Currently, the naming pattern is inconsistent and sometimes doesn't match the official name of the external service. This patch renames such identifiers, starting from the following official names:

* Amazon S3 (ref: <https://aws.amazon.com/s3/>).

* Azure Blob Storage (ref: <https://azure.microsoft.com/en-us/products/storage/blobs>).

* Google Cloud Storage (ref: <https://cloud.google.com/storage?hl=en>). I guess technically they just call it "Cloud Storage", but it would be ambiguous without Google.

Specifically, I renamed the members of `CloudProviderChoice`, and the names of the `_CloudStorage` subclasses.

I didn't change the _values_ of `CloudProviderChoice`, because this would be a breaking API change, and I don't think that's worth doing.

In addition, I:

* Removed the `GOOGLE_DRIVE` member, which has no implementation.
* Used `TextChoices` to give each member a human-readable description.

### Checklist

### License

### DIFF
--- a/cvat/apps/engine/cloud_provider.py
+++ b/cvat/apps/engine/cloud_provider.py
@@ -557,7 +557,7 @@ def get_cloud_storage_instance(
             prefix=specific_attributes.get("prefix"),
         )
     elif cloud_provider == CloudProviderChoice.AZURE_BLOB_STORAGE:
-        instance = AbsCloudStorage(
+        instance = AzureBlobCloudStorage(
             resource,
             account_name=credentials.account_name,
             sas_token=credentials.session_token,
@@ -808,7 +808,7 @@ class S3CloudStorage(_CloudStorage):
         return allowed_actions
 
 
-class AbsCloudStorage(_CloudStorage):
+class AzureBlobCloudStorage(_CloudStorage):
     MAX_CONCURRENCY = 3
 
     class Effect:


cvat/apps/engine/cloud_provider.py
```diff
@@ -808,7 +808,7 @@ def supported_actions(self):
         return allowed_actions
 
 
-class AzureBlobContainer(_CloudStorage):
+class AbsCloudStorage(_CloudStorage):
```
zhiltsov-max: I'm not sure Abs can be considered the established abbreviation for this service. Have you seen it anywhere? I'd prefer calling it `Azure...`.

SpecLad: > I'm not sure Abs can be considered the established abbreviation for this service. Have you seen it anywhere?

It's admittedly not hugely prominent, but [examples exist](https://www.google.com/search?q=%22azure%20blob%20storage%22%20%22abs%22%20site%3Amicrosoft.com&client=ubuntu-sn&channel=fs&start=0).

> I'd prefer calling it `Azure...`.

IMO, just "Azure" is not specific enough. Would "AzureBlobCloudStorage" work for you?

zhiltsov-max: Yes, I'd prefer `AzureBlobCloudStorage` instead. 

I also think it would be right to rename all `*CloudStorage` classes to something like `*Client` eventually, because it better reflects the purpose of these classes and allows to avoid stuff like `storage = db_storage_to_storage_instance(db_storage)` in several places in the code.

SpecLad: > Yes, I'd prefer `AzureBlobCloudStorage` instead.

Okay, I did it.

> I also think it would be right to rename all `*CloudStorage` classes to something like `*Client` eventually, because it better reflects the purpose of these classes and allows to avoid stuff like `storage = db_storage_to_storage_instance(db_storage)` in several places in the code.

Yeah, I agree. I even thought about doing that in this PR, but I figured I should do one thing at a time, or I'd never finish. :slightly_smiling_face:

PR #9930, opened 2025-10-23
Remove migrations that were squashed in #9922

SpecLad:
### Motivation and context

This eliminates a lot of essentially untestable code, at the cost of making it impossible to upgrade CVAT directly from a version before v2.0.0.

The [Django docs](https://docs.djangoproject.com/en/5.2/topics/migrations/#migration-squashing) recommend waiting until everyone has applied the squashed migration before removing the old ones. That would mean that we couldn't actually remove the old migrations until a few years after #9922. However, my testing shows that Django works fine even if the old migrations are removed at the same time as the squashed one is added, as long as all of the old migrations had already been applied. Therefore, I think it's reasonably safe to delete the old migrations in the same release that adds the squashed migration.

### How has this been tested?

* I tried to migrate from a state where the v2.0.0 migrations were partially applied - this produced an error, as expected.
* I tried to migrate from a state where the v2.0.0 migrations were fully applied - this worked correctly; the squashed migration was recorded in the database, but it wasn't actually run.

### Checklist

### License

### DIFF
--- a/cvat/apps/engine/media_extractors.py
+++ b/cvat/apps/engine/media_extractors.py
@@ -619,7 +619,6 @@ class VideoReader(IMediaReader):
         self,
         *,
         frame_filter: Union[bool, Iterable[int]] = True,
-        video_stream: Optional[av.video.stream.VideoStream] = None,
     ) -> Iterator[tuple[av.VideoFrame, str, int]]:
         """
         If provided, frame_filter must be an ordered sequence in the ascending order.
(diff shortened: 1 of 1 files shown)


zhiltsov-max:
The Django docs recommend doing this after a release, I agree with it. Let's merge this PR after at least 1 release with the new migration.

SpecLad:
What difference do you think it'll make?

zhiltsov-max:
It's just following the standard deprecation and removal process. Removing something publically-facing without a notification is typically a bad idea, even if we estimate the potential impact as low.

SpecLad:
Alright, let's do that.

SpecLad:
Well, the release is out. Also, I removed some more unused code.

PR #9950, opened 2025-10-28
Return only own tokens in `api/auth/access_tokens`

zhiltsov-max:
### Motivation and context

Admins could see access tokens of other users in the `api/auth/access_tokens` responses. This was not a security problem, but it created inconveniences in using the functionality. This PR makes the token visibility rules uniform for all the user categories. The list of all tokens is still available in the admin panel, if needed.

- Only own access tokens will be visible in the `api/auth/access_tokens` responses for all users (including admins)
- Removed the `owner` filter, as there's no point in using it anymore

### Checklist

### License

### DIFF
--- a/changelog.d/20251028_140529_mzhiltsov_list_only_own_tokens.md
+++ b/changelog.d/20251028_140529_mzhiltsov_list_only_own_tokens.md
@@ -5,7 +5,7 @@
 
 ### Removed
 
-- \[Server API\] Only own access tokens will be returned in the `GET api/auth/access_tokens`
+- \[Server API\] Only own access tokens will be returned in the `GET /api/auth/access_tokens`
   responses for everyone, including admins
   (<https://github.com/cvat-ai/cvat/pull/9950>)
 - \[Server API\] The `owner` filters are removed from the `GET /api/auth/access_tokens` endpoint


SpecLad:
It seems like this creates an inconsistency in permissions: admins can't list other users' tokens, but they can still interact with them by identifying them directly. IMO, we should either not make this change (and solve the problem on the UI side), or fully disable admins' access to other users' tokens.

zhiltsov-max:
Maybe, but it's consistent with other endpoints with own-only resources, like requests, so if change it, change everywhere. There's no real reason to limit this for admins, only the list was the problem. It could be some kind of default filter, but there are no known use cases for this yet.

changelog.d/20251028_140529_mzhiltsov_list_only_own_tokens.md
```diff
@@ -0,0 +1,12 @@
+### Changed
+
+- Admins will no longer see access tokens of other users on the token management page
+  (<https://github.com/cvat-ai/cvat/pull/9950>)
+
+### Removed
+
+- \[Server API\] Only own access tokens will be returned in the `GET api/auth/access_tokens`
```
SpecLad:
```suggestion
- \[Server API\] Only own access tokens will be returned in the `GET /api/auth/access_tokens`
```

PR #9962, opened 2025-10-31
`create_thread`: remove the `is_dataset_import` parameter

SpecLad:
### Motivation and context

It is used 3 times, and each time appears unnecessary:

1. `data['remote_files']` implies `not is_dataset_import`, because dataset imports can't have remote files.

2. `db_data.storage == models.StorageChoice.SHARE` implies `not is_dataset_import`, because dataset imports always use `StorageChoice.LOCAL`.

3. `data['sorting_method'] == models.SortingMethod.PREDEFINED` implies `not is_dataset_import`, because dataset imports always use `LEXICOGRAPHICAL`.

### Checklist

### License

### DIFF
--- a/cvat/apps/engine/task.py
+++ b/cvat/apps/engine/task.py
@@ -603,7 +603,6 @@ def create_thread(
     data: dict[str, Any],
     *,
     is_backup_restore: bool = False,
-    is_dataset_import: bool = False,
 ) -> None:
     if isinstance(db_task, int):
         db_task = models.Task.objects.select_for_update().get(pk=db_task)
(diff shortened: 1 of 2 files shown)


cvat/apps/engine/task.py
```diff
@@ -886,7 +885,7 @@ def update_status(msg: str) -> None:
         if extractor is not None:
             raise ValidationError('Combined data types are not supported')
 
-        if (is_dataset_import or is_backup_restore) and media_type == 'image' and db_data.storage == models.StorageChoice.SHARE:
+        if is_backup_restore and media_type == 'image' and db_data.storage == models.StorageChoice.SHARE:
```
zhiltsov-max, line 888: If I'm not mistaken, this code block under the condition is dead code entirely. It's not possible to control the storage location for a task created from a backup, and the location is always LOCAL. I'm not totally sure if this has to be called without the storage type check. Could you add a comment to revisit this fragment later or maybe just remove it?

SpecLad: I think I would rather do that in #9972. Without that patch it should be possible for this block to be executed, since share task backups will actually have `"storage": "share"` in their `task.json` file, which will be passed directly to `create_thread`.

zhiltsov-max: Ok

zhiltsov-max (approved):

PR #9984, opened 2025-11-06
engine: accept contextual images from file share in path validator

kaustubh2332:


zhiltsov-max:
Hi, thank you for the PR. Please add at least 1 test to check if the changes work correctly and there are no regressions in the future. Please also keep in mind that there's another PR that tries to address this issue: https://github.com/cvat-ai/cvat/pull/9985. I think, it will be better to focus each PR on just 1 problem.

cvat/apps/engine/cache.py
```diff
@@ -677,20 +677,45 @@ def read_raw_context_images(
         decode: bool = True,
     ) -> Generator[tuple[int, tuple[PIL.Image.Image | str, str, str]], None, None]:
         data_upload_dir = db_data.get_upload_dirname()
+        # Get share root directory
+        share_root = None
+        if hasattr(db_data, 'get_share_root'):
+             share_root = db_data.get_share_root()
```
zhiltsov-max, line 683: What is this method? There is no `get_share_root()` method in the whole repository.

cvat/apps/engine/cache.py
zhiltsov-max, line 1: Please check the comments in the other PR. There are good ideas in both implementations, let's make the combined version.

PR #9985, opened 2025-11-06
Fix issue #9978: Support related images from local file share

### LINKED ISSUE #9978: Issues with related images in tasks from local file share

### Steps to Reproduce

Can't display related images in a task from share:

1. Create a task with related images from a locally attached file share (you can use these files - [with_related_images.zip](https://github.com/user-attachments/files/23390390/with_related_images.zip), unpack the archive)
2. Try to open a job for annotation - the related images will fail to be displayed 

Related images are not included in the task backup:

[test_task_from_share_with_ri_backup.zip](https://github.com/user-attachments/files/23390403/test_task_from_share_with_ri_backup.zip)

Task creation from the backup fails (being fixed in https://github.com/cvat-ai/cvat/pull/9972 and https://github.com/cvat-ai/cvat/commit/485860a82f3c9bde7372cc0f0fa8add628999b38).
And if the fix is applied, the task can be imported, but the related images will be missing in the task.

### Expected Behavior

The task can be opened with the related images displayed in the UI.
The related images are present in the backup.
The backup can be imported and the related images are visible in the new task.

### Possible Solution

- Fix image path validation to support share paths in https://github.com/cvat-ai/cvat/blob/d2cc133d29c0e935f69ddec5a723781ae9081bc1/cvat/apps/engine/cache.py#L679-L693
- Copy related images into the backup (possible implementation in https://github.com/cvat-ai/cvat/commit/485860a82f3c9bde7372cc0f0fa8add628999b38).

### Context

_No response_

### Environment

```Markdown

```

harshasiddartha:
Fixes #9978

## Changes
- Fixed path validation in `_validate_ri_path` to support SHARE storage paths by checking both `raw_data_dir` (SHARE_ROOT) and `data_upload_dir`
- Added related images to backup for SHARE storage tasks

## Testing
- Related images from local file share should now display correctly in the UI
- Related images are now included in task backups
- Backups can be imported with related images visible

### DIFF
--- a/cvat/apps/engine/cache.py
+++ b/cvat/apps/engine/cache.py
@@ -677,17 +677,22 @@ class MediaCache:
         decode: bool = True,
     ) -> Generator[tuple[int, tuple[PIL.Image.Image | str, str, str]], None, None]:
         data_upload_dir = db_data.get_upload_dirname()
+        raw_data_dir = db_data.get_raw_data_dirname()
 
         def _validate_ri_path(path: str) -> str:
             if os.path.isabs(path):
-                if not path.startswith(data_upload_dir + os.sep):
+                if path.startswith(raw_data_dir + os.sep):
+                    path = os.path.relpath(path, raw_data_dir)
+                elif path.startswith(data_upload_dir + os.sep):
+                    path = os.path.relpath(path, raw_data_dir)
+                else:
                     raise Exception("Invalid related image path")
-
-                path = os.path.relpath(path, data_upload_dir)
             else:
-                if not os.path.normpath(os.path.join(data_upload_dir, path)).startswith(
-                    data_upload_dir + os.sep
-                ):
+                normalized_upload_path = os.path.normpath(os.path.join(data_upload_dir, path))
+                normalized_raw_path = os.path.normpath(os.path.join(raw_data_dir, path))
+                
+                if not (normalized_raw_path.startswith(raw_data_dir + os.sep) or
+                        normalized_upload_path.startswith(data_upload_dir + os.sep)):
                     raise Exception("Invalid related image path")
 
             return path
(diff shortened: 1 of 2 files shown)


zhiltsov-max:
Hi. Thank you for the PR, please add at least 1 test for each problem fixed.

zhiltsov-max:
Please also keep in mind that there's another PR that tries to address this issue: https://github.com/cvat-ai/cvat/pull/9984. I think, it will be better to focus each PR on just 1 problem.

cvat/apps/engine/cache.py
```diff
@@ -677,17 +677,22 @@ def read_raw_context_images(
         decode: bool = True,
     ) -> Generator[tuple[int, tuple[PIL.Image.Image | str, str, str]], None, None]:
         data_upload_dir = db_data.get_upload_dirname()
+        raw_data_dir = db_data.get_raw_data_dirname()
 
         def _validate_ri_path(path: str) -> str:
             if os.path.isabs(path):
```
zhiltsov-max, line 683: I also think it would be better to normalize any input paths, not only relative ones. Could you add this conversion as well?

cvat/apps/engine/cache.py
```diff
@@ -677,17 +677,22 @@ def read_raw_context_images(
...
-                if not path.startswith(data_upload_dir + os.sep):
+                if path.startswith(raw_data_dir + os.sep):
+                    path = os.path.relpath(path, raw_data_dir)
+                elif path.startswith(data_upload_dir + os.sep):
+                    path = os.path.relpath(path, raw_data_dir)
```
zhiltsov-max, line 687: I'm not sure we have a use case for this branch at the moment. If the task is local or from cloud, the data is supposed to be in the upload dir. The only other case is the file share, in which case the files are supposed to be there. Both these cases should be covered just by using `get_raw_data_dirname` instead of `get_upload_dirname`. I suggest to keep only 1 branch and remove the `data_upload_dir` variable.

PR #9990, opened 2025-11-07
set annotation of nginx to 800MB upload file maximum to avoid 413

tweigel-dev:
### Motivation and context

This change increases the allowed client request body size for the NGINX ingress to **800 MB** by adding the annotation:

```yaml
nginx.ingress.kubernetes.io/proxy-body-size: 800m
```

Previously, large file uploads (e.g., video datasets) could trigger a **`413 Request Entity Too Large`** error when creating tasks in CVAT.  
Raising this limit ensures smoother uploads and avoids task creation failures for larger datasets.  

No functional or behavioral changes to CVAT itself — only configuration improvement for more robust deployments.

https://github.com/cvat-ai/cvat/issues/9923
---

### How has this been tested?

- Applied the updated annotation in a CVAT deployment on Kubernetes (Helm-based).  
- Verified successful creation of tasks with large file uploads (>100 MB).  
- Confirmed no regressions in task creation or ingress routing.  
- Environment:  
  - **CVAT:** latest `develop` branch  

---

### Checklist
- [ ] I have updated the documentation accordingly (deployment/Helm configuration)  
- [x] I have tested the change in a real deployment environment  
- [x] I have linked related issues (if applicable)

---

### License

### DIFF
--- a/helm-chart/values.yaml
+++ b/helm-chart/values.yaml
@@ -501,7 +501,9 @@ ingress:
   ## annotations:
   ##   kubernetes.io/ingress.class: nginx
   ##
-  annotations: {}
+  annotations:
+    nginx.ingress.kubernetes.io/proxy-body-size: 800m # choose a higher value based on your data size
+  
   ## @param ingress.className IngressClass that will be used to implement the Ingress (Kubernetes 1.18+)
   ## This is supported in Kubernetes 1.18+ and required if you have more than one IngressClass marked as the default for your cluster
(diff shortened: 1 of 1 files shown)


azhavoro:
This should be done in a user-specific override file, not in the default values file.

tweigel-dev:
@azhavoro  
with the default settings normal sized images do not pass the nginx, thus increasing it does make sense to adapt the default value for a image labeling platform

PR #9994, opened 2025-11-10
relaxing manifest creation

Eldies:
### Motivation and context

After update of ffmpeg, for some videos, when seeking to a keyframe, seek lands on a previous keyframe.

With our current way of creating manifest, for such videos the manifest is empty. 
It is not optimal: it is a lot better to be able to seek to 1000th keyframe of some video and land on 999th keyframe and decode to 1000th, than to decode all the way from video beginning.

Relaxing manifest creation for such cases.

### Checklist

### License


utils/dataset_manifest/core.py
```diff
@@ -70,12 +70,34 @@ def __len__(self):
...
+    def validate_key_frame(
+        self,
+        container: av.container.InputContainer,
+        video_stream: av.video.stream.VideoStream,
+        key_frame: dict,
+        prev_seek_pts: Optional[int],
+    ) -> tuple[bool, Optional[int]]:
+        """
+        Returns a tuple
+        The first element indicates whether it is possible to seek to the key_frame using its pts (possibly with some decoding after seek).
```
SpecLad: Seems like the first element is redundant - the caller can just check whether the pts is None.

utils/dataset_manifest/core.py
```diff
@@ -111,15 +134,13 @@ def __iter__(self) -> Iterator[Union[int, tuple[int, int, str]]]:
...
-                        checking_container.seek(
-                            offset=key_frame_data["pts"],
-                            stream=checking_v_stream,
-                        )
-                        is_valid_key_frame = self.validate_key_frame(
+                        is_valid_key_frame, seek_pts = self.validate_key_frame(
                             checking_container,
                             checking_v_stream,
                             key_frame_data,
+                            prev_seek_pts,
```
SpecLad, line 145: FWIW, I think you could use `seek_pts` here and eliminate the `prev_seek_pts` variable.

utils/dataset_manifest/core.py
```diff
@@ -70,12 +70,34 @@ def __len__(self):
...
+
+        seek_pts = None
         for packet in container.demux(video_stream):
             for frame in packet.decode():
+                if seek_pts is None:
+                    seek_pts = frame.pts
+                    # if seek landed on the same frame as previous seek, it is redundant
+                    if prev_seek_pts == seek_pts:
+                        return False, None
+                if frame.pts < key_frame["pts"]:
```
SpecLad: Should we perhaps add a limit on how many frames we're willing to iterate over before we give up and raise an `InvalidVideoError`? It seems to me that if we allow unlimited iteration, then there could be pathological inputs that make CVAT do much more work than expected (e.g. decoding the entire video just to get some frames from the middle).

utils/dataset_manifest/core.py
Eldies: only returning pts now

utils/dataset_manifest/core.py
Eldies: There is already an error "The number of keyframes is not enough for smooth iteration over the video" for case with too few good keyframes, and this PR does not really change anything in this regard.

Limited the iteration in `VideoManifestValidator`, because there is a risk of a lot of extra useless work

utils/dataset_manifest/core.py
Eldies: eliminated

Eldies: I just understood that there was a mistake, and prev_seek_pts variable is needed.
E.g. imagine several keyframe which lands on the same pts:
1) seek to first, lands on first - seek_pts is not None, saving it.
2) seek to second, lands on first - the same pts, so seek_pts is None, skipping it
3) seek to third, lands on first - previous seek_pts is None here, so the code does not treat it as repeating pts, so saving it again - undesired behaviour.

Fixed it by only saving seek_pts to prev_seek_pts if it is not None

utils/dataset_manifest/core.py
SpecLad: > There is already an error "The number of keyframes is not enough for smooth iteration over the video" for case with too few good keyframes, and this PR does not really change anything in this regard.

I don't think that's quite the same thing; that check verifies that there are enough keyframes, whereas I'm talking about changing which keyframes are considered good.

> Limited the iteration in `VideoManifestValidator`, because there is a risk of a lot of extra useless work

Yeah, that's good, but I do think we should add this limit here as well. Otherwise wouldn't it be possible to generate a manifest that would not pass validation?

SpecLad:
Please add a changelog entry.

utils/dataset_manifest/core.py
Eldies: limited iteration here

Eldies:
> Please add a changelog entry.

done

PR #9972, opened 2025-11-04
Fix backups for tasks using a mounted file share

SpecLad:
Fixes #4989
Fixes #6149
Fixes #7965
Fixes #8297

### Motivation and context

There are three problems here.

First, CVAT refuses to back up such tasks if they use static chunks. There doesn't seem to be much purpose for this; the backup process doesn't even use chunks. With the check removed, the backups work fine.

Second, even when backing up succeeds, restoring fails with "No such file or directory: <root>/share/manifest.jsonl".

IMO, the problem here is not really in the restore logic, but in the backup logic. Share task backups are "heavyweight", in that they include a copy of the media files. But the `storage` field in `task.json` is still set to "share". As a result, CVAT tries to import it as a share task, and that logic is broken.

I think we should handle such backups consistently with heavyweight cloud storage backups, and set `storage` to "local". The restore logic will then work perfectly well. In the future, if we happen to implement lightweight backups for share tasks, we can use `storage: "share"` for those.

Third, if a task has images and a custom frame range (start frame/stop frame/step), then importing the backup _also_ doesn't work for reasons described in #10004. We can fix this using the `_write_filtered_media_manifest` function introduced in that PR. 

This PR also removes logic from `create_thread` that tries to fix up backups that have `storage="share"`. After this patch, such backups no longer need fixing up (and before the patch, they couldn't be imported anyway).

### How has this been tested?

Unit tests.

### Checklist

- [x] I submit my changes into the `develop` branch
- [x] I have created a changelog fragment 
- ~~[ ] I have updated the documentation accordingly~~
- [x] I have added tests to cover my changes
- [x] I have linked related issues (see [GitHub docs](
  https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword))

### License

- [x] I submit _my code changes_ under the same [MIT License](
  https://github.com/cvat-ai/cvat/blob/develop/LICENSE) that covers the project.
  Feel free to contact the maintainers if that's a concern.

PR #10004, opened 2025-11-14
Fix heavyweight backups of tasks that have images and a non-default frame range

SpecLad:
### Motivation and context

Let's say we have a cloud storage-based task that was created with 6 input images, and the following settings: start frame 1, stop frame 4, frame step 2. A heavyweight backup of such a task will then contain only frames 1 and 3 in the `data` directory; however, the manifest will still contain all 6 frames. If you try to restore such a backup, CVAT will fail, because it checks that the files in the `data` directory correspond 1:1 to the manifest.

This could potentially be fixed in the restore code; however, it seems to me that the backups created in this case are incorrect, as they have manifests referencing nonexistent files. As such, I think it's more appropriate to fix this in the backup code.

The fix is to filter the manifest during backup, leaving only entries corresponding to frames that actually get backed up. We also need to reset the frame range to the default, so that it matches the filtered manifest.

Note that the same bug also affects backups of tasks based on the mounted share. It should be reasonably easy to fix (just use `_write_filtered_media_manifest` in the `StorageChoice.SHARE` branch), but I cannot test such a fix, because share backups are currently broken entirely. So I will defer this fix until #9972.

This bug does _not_ affect backups of tasks based on local files, because those currently include _all_ frames, even those that fall outside that frame range. I think that behavior is probably wrong, but it doesn't require immediate fixing.

### How has this been tested?

Unit tests. Also manually.

### Checklist

### License

### DIFF
--- a/tests/python/rest_api/test_cloud_storages.py
+++ b/tests/python/rest_api/test_cloud_storages.py
@@ -693,7 +693,8 @@ class TestGetCloudStorageContent:
             if not next_token:
                 break
 
-        assert expected_content == current_content
+        # Each page is currently sorted individually, so we have to resort after combining.
+        assert expected_content == sorted(current_content, key=lambda el: el["type"].value)
 
     @pytest.mark.parametrize("cloud_storage_id", [2])
     def test_can_get_storage_content_with_manually_created_dirs(
--- a/tests/python/rest_api/test_task_data.py
+++ b/tests/python/rest_api/test_task_data.py
@@ -853,7 +853,7 @@ class TestPostTaskData:
             ("abc_manifest.jsonl", "[a-c]*.jpeg", False, 2, ""),
             ("abc_manifest.jsonl", "[d]*.jpeg", False, 1, ""),
             ("abc_manifest.jsonl", "[e-z]*.jpeg", False, 0, "No media data found"),
-            (None, "*", True, 0, "Combined media types are not supported"),
+            (None, "*", True, 0, "Only one video, archive, pdf, zip"),
             (None, "test/*", True, 3, ""),
             (None, "test/sub*1.jpeg", True, 1, ""),
             (None, "*image*.jpeg", True, 3, ""),


cvat/apps/engine/backup.py
```diff
@@ -436,6 +437,46 @@ def _write_annotation_guide(self, zip_object: ZipFile, target_dir: str) -> None:
             target_dir=target_dir,
         )
 
+    def _write_filtered_media_manifest(self, zip_object: ZipFile, target_dir: str) -> None:
```
zhiltsov-max, line 439: It looks like these frame filtering parameters (start, stop, and step) only make sense and should only apply to video tasks. This might be a big change for a bug fix, but could you check if we can move this into task creation instead? I'd probably prefer having lightweight and regular backups with the same data configurations.

SpecLad: > It looks like these frame filtering parameters (start, stop, and step) only make sense and should only apply to video tasks.

I don't think I agree. For example, when a user creates a task, they can select a directory full of images and then set these parameters to only use a subset of them.

Also, regardless of whether creating tasks like this _should_ be supported, de facto it already works. So we should be able to back up and restore such tasks.

zhiltsov-max: > I don't think I agree. For example, when a user creates a task, they can select a directory full of images and then set these parameters to only use a subset of them.

My point was only that we probably should stop including all the unused images into task manifests for non-video tasks. Basically, do the same manifest filtering that was added in this PR, but during the task creation after we apply all the frame filters (these parameters, wildcards etc.).

SpecLad: I think I tentatively agree that that would be a good thing. However:

1. It would be a pretty significant change, and I can't predict in advance if it'll cause other problems.
2. I think that we should apply the changes in this patch anyway, because any changes in task creation will not apply to tasks that already exist, and I think it's desirable to support restoring their backups too.

PR #10012, opened 2025-11-17
Support plugins for job menu and player

klakhov:
### Motivation and context

This PR adds plugin support for job annotation view menu and player. 
Also splitted logic for fetching and decoding contextImages into two separate methods

### Checklist

- ~~[ ] I have created a changelog fragment ~~

### License


cvat-ui/src/reducers/index.ts
```diff
@@ -416,6 +424,9 @@ export interface PluginsState {
         jobActions: {
             items: PluginComponent[];
         };
+        annotationActions: {
```
bsekachev: I would propose this schema:

annotationPage: {
    menuActions: {
        ...
    }
}

jobActions, taskActions, projectActions actually are used on different pages and this is why they are on top level

cvat-ui/src/actions/annotation-actions.ts
```diff
@@ -863,6 +863,11 @@ export function closeJob(): ThunkAction {
         const { instance: canvasInstance } = state.annotation.canvas;
         const { jobInstance, groundTruthInstance } = receiveAnnotationsParameters();
 
+        const beforeCloseCallbacks = state.plugins.callbacks.annotationActions.beforeJobClose;
+        for await (const callback of beforeCloseCallbacks) {
+            await callback(jobInstance);
+        }
+
```
bsekachev: I believe you may override Job.close method via `cvat-core` mechanism

cvat-ui/src/components/annotation-page/top-bar/annotation-menu.tsx
```diff
@@ -137,15 +142,15 @@ function AnnotationMenuComponent(): JSX.Element {
                                             min={0}
                                             max={stopFrame}
                                             onChange={(value) => {
-                                                removeFrom = value;
+                                                removeFrom = value ?? undefined;
```
bsekachev, line 145: Why do we need these two changes? 
If value is not specified, it is already `undefined` or `null`.

klakhov: We have ts error here, value from on change can be null, and the method removeAnnotationsAsyncAction and jobInstance.annotations.clear support only number | undefined

bsekachev: Should not we fix types in `jobInstance.annotations.clear` instead?
BTW `removeAnnotationsAsync` always expects numbers according to specified types, so, the types are inconsistent anyway

klakhov: `jobInstance.annotations.clear` accept `number | undefined` for those params and have separate logic for handling such case (clearing all annotations). So I suppose the correct way should be changing types here and in `removeAnnotationsAsync`.

cvat-core/src/frames.ts
```diff
@@ -752,7 +752,11 @@ async function refreshJobCacheIfOutdated(jobID: number): Promise<void> {
     }
 }
 
-export async function getContextImage(jobID: number, frame: number): Promise<Record<string, ImageBitmap>> {
+export async function getContextImage(
+    jobID: number,
+    frame: number,
+    getImageContext: (frame: number) => Promise<ArrayBuffer>,
+): Promise<Record<string, ImageBitmap>> {
```
bsekachev, line 759: Actually I see a small issue in this function, not related to the pull request

We have this check

```
if (!(jobID in frameDataCache)) {
```

However before we have 

```
const frameData = frameDataCache[jobID];

...
const meta = await frameData.getMeta();
const { jobStartFrame } = frameData;
```

So, if jobID not in frameDataCache, the code fails even before the check.

cvat-core/src/session-implementation.ts
```diff
@@ -286,12 +286,21 @@ export function implementJob(Job: typeof JobClass): typeof JobClass {
         },
     });
 
+    Object.defineProperty(Job.prototype.frames.contextImageData, 'implementation', {
```
bsekachev, line 289: The same has to be defined for a `Task` (just add stub method throwing `not implemented` error as in case with `Task.contextImage`)

cvat-ui/src/components/plugins-entrypoint.tsx
```diff
@@ -19,6 +19,8 @@ const core = getCore();
 export type PluginActionCreators = {
     getModelsSuccess: typeof modelsActions['getModelsSuccess'],
     changeFrameAsync: typeof changeFrameAsync,
+    confirmCanvasReadyAsync: typeof confirmCanvasReadyAsync,
```
bsekachev: As far as I can see, not used anymore in plugins

cvat-ui/webpack.config.js
```diff
@@ -108,6 +109,7 @@ module.exports = (env) => {
                             plugins: [
                                 '@babel/plugin-proposal-class-properties',
                                 '@babel/plugin-proposal-optional-chaining',
+                                '@babel/plugin-transform-private-methods',
```
bsekachev, line 112: Could you please clarify why we need this plugin?

cvat-ui/src/components/annotation-page/top-bar/player-navigation.tsx
```diff
@@ -204,6 +210,12 @@ function PlayerNavigation(props: Props): JSX.Element {
                                 })}
                             </svg>
                         )}
+                        {playerSliderPlugins
+                            .sort((a, b) => a.weight - b.weight)
+                            .map(({ component: Component }, index) => {
+                                const ComponentToRender = Component as React.ComponentType<any>;
+                                return <ComponentToRender key={index} {...props} />;
```
bsekachev: It should be `targetProps: props`

cvat-ui/webpack.config.js
klakhov: This plugin allow us to define private class methods with `#` notation, like this: `#doSomething(param: number): void`, without it only private fields are allowed: `this.#field`.

bsekachev (approved):

PR #10019, opened 2025-11-19
Allow to open browser context menu on right click in resource cards

klakhov:
### Motivation and context

Refactored code so when we click elements like links default browser context menu is opened, not card context menu 

### Checklist

### License

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -362,6 +362,7 @@ jobs:
           DJANGO_SU_NAME: 'admin'
           DJANGO_SU_EMAIL: 'admin@localhost.company'
           DJANGO_SU_PASSWORD: '12qwaszx'
+
         run: |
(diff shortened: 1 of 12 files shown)


klakhov:
I updated PR to apply new approach for all cards.

I dont think we can use `onClickCapture` to cover ctrl+click opening in new tab as it is capturing right clicks, not left clicks.

ctrl+left click works fine even now, but only for `Link` components. So its applicable now only for job item id link and task item open link. I think we need to add links on job card and project card to cover this case. If I remember correctly before bulk actions we used preview as a link, but now it would be confusing so its better to use id or name for link element.

bsekachev (approved):
Generally it does not solve initial problem as such elements like `a` with links even do not exists on target cards/elements (except for TaskItemComponent). 

However they may be added as a different pull requests, this one lgtm

PR #10021, opened 2025-11-20
added kvrocks SS PVC annotations + more rules for helmignore

yaizkazani:
We need to be able to configure volumes with annotations through CSI 

### License

### DIFF
--- a/helm-chart/values.yaml
+++ b/helm-chart/values.yaml
@@ -324,6 +324,7 @@ cvat:
 #     accessModes:
 #       - ReadWriteOnce
       size: 100Gi
+      annotations: {}
 
 postgresql:
   #See https://github.com/bitnami/charts/blob/master/bitnami/postgresql/ for more info


changelog.d/20251120_113157_petr.iosipov_DEV_762.md
```diff
@@ -0,0 +1,4 @@
+### Added
+
+- PVC configuration via annotations
```
azhavoro:
```suggestion
- \[Helm\] Kvrocks PVC configuration via annotations
```

helm-chart/templates/cvat_kvrocks/statefulset.yml
```diff
@@ -118,6 +118,10 @@ spec:
         {{- include "cvat.labels" . | nindent 8 }}
         app: cvat-app
         tier: kvroocks
+      {{- with .Values.cvat.kvrocks.defaultStorage.annotations }}
```
azhavoro, line 121: Please add an empty value for `cvat.kvrocks.defaultStorage.annotations` to the values file, otherwise it will be impossible to guess that annotations can be specified without looking at the template file.

yaizkazani: True! Good idea indeed

PR #10022, opened 2025-11-20
Fix quality conflicts display in consensus replicas

zhiltsov-max:
### Motivation and context

- Quality conflicts can now be displayed for consensus replicas in the review mode
- Refactored quality settings access in the server code
- Added a migration for old tasks without quality settings to initialize default settings

### Checklist

### License

### DIFF
--- a/cvat/apps/quality_control/signals.py
+++ b/cvat/apps/quality_control/signals.py
@@ -5,7 +5,7 @@
 from django.db.models.signals import post_save
 from django.dispatch import receiver
 
-from cvat.apps.engine.models import Job, JobType, Project, Task
+from cvat.apps.engine.models import Project, Task
 from cvat.apps.quality_control.models import QualitySettings
 
 


cvat/apps/quality_control/quality_reports.py
```diff
@@ -2648,6 +2656,22 @@ def _check_project_quality(cls, *, project_id: int) -> int:
         return ProjectQualityCalculator().compute_report(project=project_id).id
 
 
+class QualitySettingsManager:
+    def get_project_settings(self, project: Project) -> models.QualitySettings:
+        return project.quality_settings
```
SpecLad, line 2661: Is this function really needed? Seems like it just adds an extra layer of indirection.

cvat/apps/quality_control/quality_reports.py
```diff
@@ -2648,6 +2656,22 @@ def _check_project_quality(cls, *, project_id: int) -> int:
...
+
+    def get_task_settings(self, task: Task, *, inherit: bool = True) -> models.QualitySettings:
+        try:
+            quality_settings = task.quality_settings
+        except models.QualitySettings.DoesNotExist:
+            quality_settings, _ = models.QualitySettings.objects.get_or_create(task_id=task.id)
```
SpecLad: The `get_or_create` seems unnecessary here - if the quality object already existed, we would not be in this block. So `create` should be enough.

cvat/apps/quality_control/quality_reports.py
```diff
@@ -2648,6 +2656,22 @@ def _check_project_quality(cls, *, project_id: int) -> int:
...
+
+        if inherit and quality_settings.inherit and task.project:
```
SpecLad, line 2666: Can we avoid creating an unnecessary settings object in the DB when inheriting task settings? Right now it seems that `get_task_settings` will always create a QS object for a task, even if it's not really used afterwards.

cvat/apps/quality_control/quality_reports.py
```diff
@@ -2648,6 +2656,22 @@ def _check_project_quality(cls, *, project_id: int) -> int:
         return ProjectQualityCalculator().compute_report(project=project_id).id
 
 
+class QualitySettingsManager:
```
SpecLad, line 2659: IMO, it would be clearer to merge this into `models.QualitySettings`, and make the methods static. Something like `QualitySettings.get_for_task(...)`.

cvat/apps/quality_control/quality_reports.py
zhiltsov-max: Just calling create would be subject for race conditions, so I just made an optimistic version of what we had earlier. However, I decided to simply add a migration instead, there should not be tasks without settings at all.

cvat/apps/quality_control/quality_reports.py
zhiltsov-max: We create settings for all new tasks in a signal. In the case of inheritance we could remove the task settings, but in this case we wouldn't be able to remember the individual settings used before switching to the inherited ones.

cvat/apps/quality_control/quality_reports.py
zhiltsov-max: Technically, such a method should go into `QualitySettings.objects` in Django. I decided to add a separate manager on the component logic side because of the extra logic with inheritance so far. This might change soon, when we understand how to generalize settings.

cvat/apps/quality_control/quality_reports.py
zhiltsov-max: The idea was to encapsulate access to the effective settings in a single class. Honestly, I'm not totally satisfied with the current implementation, because clients still have to manage the `inherited` parameter in  `ComparisonParameters.from_settings`, but now it is at least a required parameter.

cvat/apps/quality_control/quality_reports.py
SpecLad: I see you've added a migration to create a QS object for every task, so this thread is no longer relevant.

cvat/apps/quality_control/quality_reports.py
SpecLad: I don't understand why you'd want to encapsulate one attribute access, but I'll defer to you on this.

cvat/apps/quality_control/quality_reports.py
SpecLad: > Technically, such a method should go into `QualitySettings.objects` in Django.

That would work for me too.

> I decided to add a separate manager on the component logic side because of the extra logic with inheritance so far. 

Can you at least call it something other than a manager? That term already has a meaning in Django.

(TBH, I don't know why it needs to be a class at all. It could just be a couple free functions.)

zhiltsov-max: > That would work for me too.

I don't really like it because the Django manager classes are primarily focused on implementing DB or table-level operations, while this class is supposed to provide a facade for business logic in quality-related applications.

There are several other manager classes in this file already, I don't see significant issues with this naming scheme so far. I could call it *Repository or *Provider as an alternative.

SpecLad: Alright, let's keep it.

SpecLad (approved):

Issue #9225, opened 2025-03-18
Return to the first page after visiting invalid page


zhiltsov-max:
### Actions before raising this issue

- [x] I searched the existing issues and did not find anything similar.
- [x] I read/searched [the docs](https://docs.cvat.ai/docs/)

### Is your feature request related to a problem? Please describe.

When visiting a non-existent page for some resource in UI, you're shown a error notification

![Image](https://github.com/user-attachments/assets/af067487-6056-4145-b439-5f746ec9ba08)

It would be nice to redirect the user automatically to the first page or to provide a button to do so. There are other buttons on the page (Tasks, Jobs etc.), but if there was some filter enabled, visiting these pages will clear the filter.

Godzilaa:
please assign me this

zhiltsov-max:
Hi, you can start working on this without any assignment. Just send us a PR that fixes the problem.

AbhinavKRN:
@zhiltsov-max hey, i was searching for some starter issues, can you assign me this one to start the contributions?

zhiltsov-max:
@AbhinavKRN, hi, you can start working on this without any assignment. Just send us a PR that fixes the problem.

dhruthi-ty:
heyy @zhiltsov-max I came across this issue where visiting a non-existent resource page shows an error notification. I think it would be a good improvement to redirect users automatically to the first page or provide a “Go to first page” button, especially since other buttons like Tasks or Jobs currently clear filters.

A possible approach could be:
Detect when a user lands on a non-existent page for a resource.
Automatically redirect to the first page or show a clear button allowing users to go there.
Ensure any active filters are preserved or appropriately reset when redirecting.
Test across multiple resource types to confirm consistent behavior.

I’d be happy to collaborate on implementing this enhancement. could you please assign it to me?

prathamesh-patil-5090:
Was this issue fixed? Or I can start working on this too?

zhiltsov-max:
@prathamesh-patil-5090, the issue was not fixed yet, it's available for contributions.

prathamesh-patil-5090:
@zhiltsov-max I think I might haved solve the issue

How the issue was solved

The problem occurred because the ProjectNotFound, TaskNotFound, and JobNotFound pages did not provide any navigation option, forcing the user to manually change the URL or navigate back using the browser UI.

To address this, I updated the CVAT React components for all “Not Found” screens to include a “Return to Previous Page” button.

[...]

PR #10028, opened 2025-11-23
Fix: Issue #9225 Solved

prathamesh-patil-5090:
**Problem:**
When users navigate to non-existent resources (invalid task ID, project ID, job ID, or cloud storage ID), they see a Not Found error page but have no easy way to navigate back. They must either:
- Manually edit the URL
- Use the browser's back button (which may not always be intuitive)
- Navigate using the main menu

**Solution:**
This PR adds a "Return to Previous Page" button to Not Found pages with smart fallback logic:
- If browser history exists → navigates back using `history.goBack()` (preserves query parameters and navigation context)
- If browser history is empty (e.g., user arrived via direct link) → redirects to the appropriate resource list page:
  - Task Not Found → `/tasks`
  - Project Not Found → `/projects`
  - Job Not Found → `/jobs`
  - Cloud Storage Not Found → `/cloudstorages`

**Changes implemented:**
- Added "Return to Previous Page" button to `TaskNotFoundComponent`, `ProjectNotFoundComponent`, and `CloudStorageNotFoundComponent`
- Button placed in `Result.extra` prop (centered by Ant Design)
- Used React Router v5's `useHistory()` hook with smart fallback:
  ```tsx
  const handleReturn = () => {
      if (history.length > 2) {
          history.goBack();
      } else {
          history.push('/tasks'); // or appropriate list page
      }
  };
  ```

**File affected:**
- `cvat-ui/src/components/common/not-found.tsx`

### How has this been tested?

Manual verification steps performed locally:
1. Started the frontend dev server in the project root:
```bash
cd cvat-ui
yarn install
yarn start
```
2. Tested navigation scenarios for each Not Found page:

   **Scenario A: Browser history exists (normal navigation)**
   - Navigate to valid resource (e.g., `/tasks`)
   - Navigate to invalid resource (e.g., `/tasks/99999`)
   - Click "Return to Previous Page" button
   - ✅ Verified: Returns to `/tasks` preserving any query parameters

   **Scenario B: Direct URL access (no browser history)**
   - Open browser in new tab/incognito
   - Navigate directly to invalid resource (e.g., `http://localhost:3000/tasks/99999`)
   - Click "Return to Previous Page" button
   - ✅ Verified: Redirects to `/tasks` (appropriate fallback)

   **Tested for all Not Found pages:**
   - `/tasks/invalid-id` → fallback to `/tasks`
   - `/projects/invalid-id` → fallback to `/projects`
   - `/jobs/invalid-id` → fallback to `/jobs`
   - `/cloudstorages/invalid-id` → fallback to `/cloudstorages`

3. Confirmed:
   - Button is visually centered below error message
   - No console errors or warnings
   - Query parameters preserved when using `history.goBack()`
   - Smooth navigation in both scenarios
4. Verified TypeScript and lint checks in UI folder:
```bash
cd cvat-ui
yarn run type-check
yarn run lint
```
Note: All checks passed locally for this change.

5. Proofs:
Screenshot 2025-11-21 162622

Screenshot 2025-11-21 162638

### Checklist

### Changelog

**Added:**
- "Return to Previous Page" button on Task, Project, Job, and Cloud Storage Not Found pages with smart fallback navigation

### DIFF
--- a/cvat/apps/dataset_manager/annotation.py
+++ b/cvat/apps/dataset_manager/annotation.py
@@ -8,7 +8,6 @@ import math
 from collections.abc import Callable, Container, Generator, Iterable, Iterator, Sequence
 from copy import copy, deepcopy
 from itertools import chain
-from typing import Optional
 
 import numpy as np
 from scipy.optimize import linear_sum_assignment
@@ -750,8 +749,8 @@ class TrackManager(ObjectManager):
         end_frame: int,
         dimension: DimensionType | str,
         *,
-        included_frames: Optional[Sequence[int]] = None,
-        deleted_frames: Optional[Sequence[int]] = None,
+        included_frames: Sequence[int] | None = None,
+        deleted_frames: Sequence[int] | None = None,
         include_outside: bool = False,
     ):
         # If a task or job contains deleted frames that contain track keyframes,
--- a/cvat/apps/dataset_manager/bindings.py
+++ b/cvat/apps/dataset_manager/bindings.py
@@ -14,7 +14,7 @@ from functools import partial, reduce
 from operator import add
 from pathlib import Path
 from types import SimpleNamespace
-from typing import TYPE_CHECKING, Any, Literal, NamedTuple, Optional, Union
+from typing import TYPE_CHECKING, Any, Literal, NamedTuple
 
 import attr
 import datumaro as dm
@@ -80,7 +80,7 @@ class InstanceLabelData:
             'sublabels',
         ])
 
-    def __init__(self, instance: Union[Task, Project]) -> None:
+    def __init__(self, instance: Task | Project) -> None:
         instance = instance.project if isinstance(instance, Task) and instance.project_id is not None else instance
 
         db_labels = self.add_prefetch_info(instance.label_set.all())
@@ -300,7 +300,7 @@ class CommonData(InstanceLabelData):
         host: str = '',
         create_callback=None,
         use_server_track_ids: bool = False,
-        included_frames: Optional[Sequence[int]] = None
+        included_frames: Sequence[int] | None = None
     ) -> None:
         self._dimension = annotation_ir.dimension
         self._annotation_ir = annotation_ir
@@ -313,7 +313,7 @@ class CommonData(InstanceLabelData):
         self._db_data: models.Data = db_task.data
         self._use_server_track_ids = use_server_track_ids
         self._required_frames = included_frames
-        self._initialized_included_frames: Optional[set[int]] = None
+        self._initialized_included_frames: set[int] | None = None
         self._db_subset = db_task.subset
 
         super().__init__(db_task)
@@ -790,8 +790,8 @@ class CommonData(InstanceLabelData):
         return osp.splitext(path)[0]
 
     def match_frame(self,
-        path: str, root_hint: Optional[str] = None, *, path_has_ext: bool = True
-    ) -> Optional[int]:
+        path: str, root_hint: str | None = None, *, path_has_ext: bool = True
+    ) -> int | None:
         if path_has_ext:
             path = self._get_filename(path)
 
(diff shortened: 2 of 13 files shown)


klakhov:
Hi @prathamesh-patil-5090,

Could you please update the PR description to clearly reflect the actual purpose of the changes?
Right now it looks AI-generated and describes things that do not exist in our codebase — for example:

* We don’t have a return/back button on those pages that needs centering.
* We can’t use `useNavigate` because the project is still on React Router v5, so this is not technically possible.

I’m not against adding a return/back button — it can be useful — but please note that browser history may be empty. In such cases the button must redirect to the appropriate main page depending on the resource type (tasks → tasks list, projects → projects list, etc.).

Also, please add a changelog entry describing the feature.

prathamesh-patil-5090:
Hey @klakhov  Sorry for the inconvience of the PR description.  In short I tried to solve the problem of Issue: https://github.com/cvat-ai/cvat/issues/9225 . Also Thanks for picking out the edge case of browser history being empty, that case has now been handled, too. I basically used useHistory to navigate if incase user goes like tasks/2?page=2 -> tasks/3232(which may not exist) to return him/her again to tasks/2?page=2. If he directly goes to tasks/3232 if it doesn't exist the button will redirect it to /tasks/

CHANGELOG.md
```diff
@@ -16,6 +16,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+<a id='changelog-2.50.0'></a>
+## \[2.50.0\] - 2025-11-26
+
+### Added
+
+- "Return to Previous Page" button on Task, Project, Job, and Cloud Storage Not Found pages
+  with smart fallback navigation: navigates back in history if available, otherwise redirects
+  to the appropriate resource list page
+  (<https://github.com/cvat-ai/cvat/pull/10028>)
+
```
klakhov, line 95: I suggest to use text:
```
### Added

- "Return to Previous Page" button on Task, Project, Job, and Cloud Storage Not Found pages
```

Also, do not modify this file directly. Our project uses `scriv` to handle changelogs to avoid merge conflicts.
To use it use the command from the note of CHANGELOG.md below:
```
 Developers: this project uses scriv (<https://scriv.readthedocs.io/en/stable/index.html>)
  to maintain the changelog. To add an entry, create a fragment:

    $ scriv create --edit

  Fragments will be merged into this file whenever a release is made.
```

cvat-ui/src/components/common/not-found.tsx
```diff
@@ -4,39 +4,84 @@
...
+import { useHistory } from 'react-router-dom';
+export const JobNotFoundComponent = React.memo((): JSX.Element => {
+    const history = useHistory();
+    const handleReturn = () => {
+        if (history.length > 2) {
+            history.goBack();
+        } else {
+            history.push('/jobs');
+        }
+    };
```
klakhov: The code of this function is duplicated, need to extract and re-use the logic.

klakhov: Also, why do we use `>2`? Shouldnt it be '`>1`?

CHANGELOG.md
prathamesh-patil-5090: Noted, thanks

cvat-ui/src/components/common/not-found.tsx
prathamesh-patil-5090: Yup i have extracted and reused the logic now

I used the '>1' in the code first-hand but then I came across the unusual behaviour of browser history, that they count homepage in history so if user navigates to http://127.0.0.1:3000/tasks/1365 which does not exist directly from the homepage and if the condition is '>1' rather than '>2', it will redirect the user to homepage of browser after clicking the button that I have added rather than taking him to  http://127.0.0.1:3000/tasks. So to counter it, I used '>2' which solves this edge case

PR #10041, opened 2025-11-26
RAM optimization for sparse tracks on export

Eldies:
### Motivation and context

When tracks are exported (in most formats), interpolation happens and interpolated shapes are kept in RAM.
E.g. if a track is 100 frames long and has 10 keyframes, only 10 shapes are read from db and kept in RAM before the interpolation.
Then, after interpolation, there are 100 shapes which are kept in RAM during export - 10x more ram is used.

This PR makes interpolation function into a generator, to not keep all those non-key shapes in RAM.

### Checklist

### License

### DIFF
--- a/changelog.d/20251210_133520_dmitrii.lavrukhin_ram_optimization_for_sparse_tracks.md
+++ b/changelog.d/20251210_133520_dmitrii.lavrukhin_ram_optimization_for_sparse_tracks.md
@@ -1,0 +1,4 @@
+### Changed
+
+- Reduced RAM usage on track export
+  (<https://github.com/cvat-ai/cvat/pull/10041>)


cvat/apps/dataset_manager/annotation.py
```diff
@@ -1068,84 +1078,81 @@ def interpolate(shape0, shape1):
...
+                        if shape["frame"] < end_frame:
+                            yield shape
+                        else:
+                            break
 
-                # Update the last added shape
-                shape["keyframe"] = True
-                prev_shape = shape
+                    # Update the last added shape
+                    shape["keyframe"] = True
```
zhiltsov-max, line 1122: The last shape was supposed to be returned modified here. The updated version first returns the shape, then changes it.

Eldies: If I understand correctly, the original version does not work like that:
`interpolated` always contains at least one shape with frame >= `end_frame`, it is explicitly appended there by `interpolated.append(shape)`
which means that the cycle above always quits with break.
Which means, the shape modified here is not in the list yet

zhiltsov-max: Ok, but the problem is that the last returned shape is being modified after it's returned in the new version. It doesn't look like a good solution to me.

Eldies: the last returned shape and the modified shape are not the same shape
i.e., if end_frame = 5, and interpolated contains frames [3,4,5,6,7], then the last returned shape is frame 4 and the modified is frame 5

zhiltsov-max: Agree, my bad

zhiltsov-max:
Consider adding a changelog message.

Eldies:
> Consider adding a changelog message.

added

PR #10073, opened 2025-12-04
Improving TUS implementation

Grigorii777:
#### **Problem:** 
The TUS implementation in `cvat/apps/engine/tus.py` was using `request.body` to access uploaded data. Django's `request.body` property attempts to read the entire request body at once, which:
- Consumes large amounts of memory for big files

Bug:
- Returns a 500 Internal Server Error status for HEAD requests after a file upload is completed via TUS

#### **Solution:** Changed to streaming approach using `shutil.copyfileobj(chunk.request, file)`:
- Reads data in 64KB chunks instead of loading entire body
- Significantly reduces memory usage for large uploads

### Checklist

### License

### DIFF
--- a/changelog.d/20251201_131452_grigorio714_handle_interrupted_uploads.md
+++ b/changelog.d/20251201_131452_grigorio714_handle_interrupted_uploads.md
@@ -1,4 +1,4 @@
 ### Fixed
 
-- Added reading HTTP request data in chunks instead of loading entire body into memory
+- Added reading HTTP request data in chunks instead of loading entire body into memory for TUS PATCH requests
   (<https://github.com/cvat-ai/cvat/issues/5261>)


cvat/apps/engine/tus.py
```diff
@@ -203,11 +206,25 @@ def init_file(self):
             file.seek(self.file_size - 1)
             file.write(b"\0")
 
+    def __write_chunk_data(self, chunk: TusChunk, file: Any) -> int:
```
SpecLad: It seems that this whole function could be replaced with `shutil.copyfileobj`. It doesn't return the number of bytes copied, but you can just use `file.tell()` to get the same information.

cvat/apps/engine/tus.py
```diff
@@ -203,11 +206,25 @@ def init_file(self):
...
+
     def write_chunk(self, chunk: TusChunk):
+        bytes_written = 0
         with open(self.file_path, "r+b") as file:
             file.seek(chunk.offset)
-            file.write(chunk.content)
-        self.meta_file.meta.offset += chunk.size
+            bytes_written = self.__write_chunk_data(chunk, file)
+        # Update offset with actual bytes written (handles interrupted uploads)
+        self.meta_file.meta.offset += bytes_written
```
SpecLad: You need to make sure that the meta file is still updated if an `UnreadablePostError` is raised during the writing process. This should happen if the connection is terminated (in fact, it's kind of strange that it doesn't in your test).

Grigorii777: Yes, I already noticed that when I was doing it through the socket.

SpecLad:
Please redo the title, description and changelog taking the reduced scope into account. In particular, this doesn't close #5261 anymore.

cvat/apps/engine/tus.py
```diff
@@ -35,10 +36,13 @@ class TusTooLargeFileError(Exception):
 
 
 class TusChunk:
+    # Buffer size for reading request body in chunks (64KB)
+    BUFFER_SIZE: int = 64 * 1024
```
SpecLad: Is there any reason to override the default buffer size of `copyfileobj`?

cvat/apps/engine/tus.py
```diff
@@ -206,9 +210,18 @@ def init_file(self):
...
         with open(self.file_path, "r+b") as file:
             file.seek(chunk.offset)
-            file.write(chunk.content)
-        self.meta_file.meta.offset += chunk.size
+            start_position = file.tell()
+            try:
+                # It handles partial reads automatically (e.g., when connection drops)
+                shutil.copyfileobj(chunk.request, file, length=TusChunk.BUFFER_SIZE)
+            finally:
+                # Always update offset, even if connection was interrupted
```
SpecLad: This isn't going to work as stated. The offset isn't actually saved until the `self.meta_file.dump()` call, which is outside the `finally` block, so it will not be triggered if an exception occurs. (And if it was inside the `finally` block, that would _also_ be incorrect, as we would then update the meta file before the main file was closed.)

Given that handling connection interruption turned out to be unviable anyway (due to [Django bug 33699](https://code.djangoproject.com/ticket/33699)), I think it would be best to just remove the `try` block here.

cvat/apps/engine/tus.py
```diff
@@ -206,9 +210,18 @@ def init_file(self):
...
+                self.meta_file.meta.offset = file.tell()
         self.meta_file.dump()
+        slogger.glob.info(
```
SpecLad: We should not be logging every uploaded chunk. It would be far too verbose.

cvat/apps/engine/mixins.py
```diff
@@ -194,6 +193,8 @@ def append_tus_chunk(self, request: ExtendedRequest, file_id: str):
         except TusFileForbiddenError:
             return self._tus_response(status=status.HTTP_403_FORBIDDEN)
 
+        tus_file.meta_file.init_from_file()
```
SpecLad, line 196: This is supposed to fix a bug, right? Can you add a test for it?

tests/python/rest_api/test_tus_uploads.py
```diff
@@ -189,3 +189,28 @@ def test_cannot_upload_chunk_exceeding_file_size(self, fxt_task):
...
+
+        location = self._start_upload(task.id, file_size, "test_image.jpg")
+        current_offset = self._get_upload_offset(location)
+        assert current_offset == 0
+        response = self._upload_chunk(location, 0, image_data)
+
+        assert response.status == HTTPStatus.NO_CONTENT
+        assert int(response.headers.get("Upload-Offset")) == file_size
+
+        # After completing the download, the HEAD request should return 404.
```
SpecLad, line 209: Seems like this would be the perfect check to stick at the end of `test_upload_offset_is_updated_incrementally`. It doesn't need its own test.

tests/python/rest_api/test_tus_uploads.py
```diff
@@ -189,3 +189,28 @@ def test_cannot_upload_chunk_exceeding_file_size(self, fxt_task):
...
+        head_response = self._call_tus_endpoint(
+            "HEAD",
+            location,
+            headers={"Tus-Resumable": "1.0.0"},
+            check_status=False,
+        )
+        assert head_response.status == HTTPStatus.NOT_FOUND  # was 500
```
SpecLad: Was when? This comment doesn't make sense in this context.

changelog.d/20251201_131452_grigorio714_handle_interrupted_uploads.md
```diff
@@ -0,0 +1,4 @@
+### Fixed
+
+- Added reading request data in chunks instead of loading entire body into memory
```
SpecLad: You should clarify what requests you're talking about.

changelog.d/20251201_131452_grigorio714_handle_interrupted_uploads.md
```diff
@@ -0,0 +1,4 @@
+### Fixed
```
SpecLad, line 1: You should also mention that bugfix you added.

SpecLad:
You didn't update the PR title. Also, the description is dubious:

> 1. Consumes large amounts of memory for big files

A list with a single item.

> Changed to streaming approach using request.read(BUFFER_SIZE):

`request.read` doesn't actually occur in the PR.

tests/python/rest_api/test_tus_uploads.py
Grigorii777: Not quite, as this test is added specifically to test the transfer of tus_file.meta_file.init_from_file()
It was a bug and we are checking that it has disappeared
I think it's best to do it atomically

SpecLad: Okay, let's do that.

changelog.d/20251224_155301_grigorio714_handle_interrupted_uploads.md
```diff
@@ -0,0 +1,4 @@
+### Fixed
+
+- Returns a 404 for HEAD requests after a file upload is completed via TUS, instead of a 500 Internal Server Error
```
SpecLad:
```suggestion
- Return a 404 for HEAD requests after a file upload is completed via TUS, instead of a 500 Internal Server Error
```

changelog.d/20251201_131452_grigorio714_handle_interrupted_uploads.md
SpecLad: You still haven't clarified this. We should make clear that this is about TUS `PATCH` requests.

SpecLad:
> Django's request.body property attempts to read the entire request body at once, which:
> *  Consumes large amounts of memory for big files
> *  Returns a 500 Internal Server Error status for HEAD requests after a file upload is completed via TUS

`request.body` does not return a 500 status.

Issue #9956, opened 2025-10-29
An option to filter frames without hiding annotations in UI


zhiltsov-max:
### Actions before raising this issue

- [x] I searched the existing issues and did not find anything similar.
- [x] I read/searched [the docs](https://docs.cvat.ai/docs/)

### Is your feature request related to a problem? Please describe.

Currently, a frame filter in the annotation view in the UI allows to filter frames and annotations based on some annotation properties. This filters both frames and annotations on the frames. Sometimes it's useful to only filter the frames, but see all the annotations on the frame, e.g. during review. Currently, you'd need to constantly turn the filter on (configuring it each time or using the last saved one) and off by clicking on the UI elements.

### Describe the solution you'd like

An option to filter just the frames based on the filter condition.

### Describe alternatives you've considered

A possible alternative is to add a keybinding to quickly turn on and off the frame filter or to show all annotations on the current frame.

huangcha03:
hi, can I claim this issue and start working on it?

zhiltsov-max:
@huangcha03, hi, sure.

prathamesh-patil-5090:
<img width="1859" height="972" alt="Image" src="https://github.com/user-attachments/assets/b5558e96-eebe-4691-93b7-61b5f1530a91" />

I  believe this is the frame filter but can you give me little bit more detail of the suggestion ?

zhiltsov-max:
@prathamesh-patil-5090, yes, it is the frame filter mentioned in the issue description. Basically, the problem is that you can't use the filter to just filter the frames based on annotations for navigation - only with filtering the annotations. The potential solution is to provide a way to apply the filter only for navigation purposes without changing the displayed frame contents, or to provide a keybind to quickly turn the filter on and off.

prathamesh-patil-5090:
Hey check this  - https://github.com/cvat-ai/cvat/pull/10074

PR #10074, opened 2025-12-04
Add "Filter frames only" option to annotation filters - Issue(#9956)

prathamesh-patil-5090:
# Add "Filter frames only" option to annotation filters - Issue(#9956)

## Motivation and context
- Problem: Current annotation filters both (a) choose which frames are considered for navigation and (b) which annotations are displayed on the current frame. When reviewing, users often want to jump only to frames that match a condition but still see all annotations on those frames.
- Goal: Let filters continue to drive frame navigation while optionally showing all annotations on the landed frame.

## What we implemented (short)
- State & action: add `annotation.annotations.filterFramesOnly` and action `SWITCH_FILTER_FRAMES_ONLY` (`switchFilterFramesOnly()`).
- Behaviour change: `fetchAnnotations()` uses `displayFilters = filterFramesOnly ? [] : filters` so display is unfiltered when the flag is ON, navigation still uses the original filters.
- UI: add a checkbox "Filter frames only" to `FiltersModalComponent` (disabled when no filters), plus small footer styling.
- Files touched (high level):
  - `cvat-ui/src/actions/annotation-actions.ts`
  - `cvat-ui/src/reducers/annotation-reducer.ts`
  - `cvat-ui/src/components/annotation-page/top-bar/filters-modal.tsx`
  - `cvat-ui/src/components/annotation-page/styles.scss`
  - `changelog.d/20251204_075011_122prathamesh6002_9956.md`

## How has this been tested (short)
- Manual test on local dev: created a task with a test image and labels (Square/Circle/Rectangle), added annotations, applied a label filter, then toggled the checkbox:
  - Checkbox OFF: only annotations matching filter are shown.
  - Checkbox ON: navigation still jumps to frames matching filter, but all annotations on each landed frame are visible.

## Checklist
- [x] I have created a changelog fragment (`changelog.d/20251204_075011_122prathamesh6002_9956.md`)

## License
- I submit my code changes under the same MIT License used by this project.

### DIFF
--- a/cvat-ui/src/components/annotation-page/styles.scss
+++ b/cvat-ui/src/components/annotation-page/styles.scss
@@ -570,6 +570,16 @@
             float: left;
             margin-top: $grid-unit-size;
         }
+
+        .cvat-filters-modal-navigation-checkbox {
+            padding: $grid-unit-size * 2 $grid-unit-size * 3;
+            margin-top: $grid-unit-size * 2;
+            border-top: 1px solid $border-color-1;
+
+            .ant-checkbox-wrapper {
+                font-size: 14px;
+            }
+        }
     }
 }
 
--- a/cvat-ui/src/components/annotation-page/top-bar/filters-modal.tsx
+++ b/cvat-ui/src/components/annotation-page/top-bar/filters-modal.tsx
@@ -6,7 +6,14 @@
 import React, { useEffect, useState } from 'react';
 import { useDispatch, useSelector, shallowEqual } from 'react-redux';
 import {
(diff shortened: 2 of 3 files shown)


prathamesh-patil-5090:
@zhiltsov-max @klakhov Hey guys, don't miss this one, issue - https://github.com/cvat-ai/cvat/issues/9956

zhiltsov-max:
1. The "Filter frames only" checkbox can't be toggled after the filter is added. The element is enabled, but doesn't react to clicks.
2. The checkbox is only available after the Submit button is clicked. I think, it would be better to allow this if there're any filter terms configured instead, because otherwise the user has to click Submit and then open the filter configuration again to use this feature.
3. Consider adding a keybinding for enabling the feature or the filter, it would make the workflow more convenient to the users.

PR #10077, opened 2025-12-05
Fix nginx config for react app

### LINKED ISSUE #10076: After updating ONNX Runtime Web, SAM cannot be used and displays a MIME type error.

### Steps to Reproduce

Failed to load module script: Expected a JavaScript-or-Wasm module script but the server responded with a MIME type of "application/octet-stream". Strict MIME type checking is enforced for module scripts per HTML spec

### Expected Behavior

_No response_

### Possible Solution

_No response_

### Context

_No response_

### Environment

```Markdown

```

bsekachev:
### Motivation and context
`nginx` server returns `.mjs` modules from `onnxruntime-web` as `application/octet-stream` what is unacceptable for the runtime. Instead, specify that such modules must have `application/javascript`. 

Resolved #10076

### Checklist

### License

### DIFF
--- a/cvat-ui/react_nginx.conf
+++ b/cvat-ui/react_nginx.conf
@@ -36,6 +36,7 @@ server {
         add_header Cross-Origin-Embedder-Policy "require-corp";
         add_header Service-Worker-Allowed "/" always;
 
+        include /etc/nginx/mime.types;
         types {
             # mjs modules must have application/javascript, not default application/octet-stream
             application/javascript mjs;


cvat-ui/react_nginx.conf
```diff
@@ -36,6 +36,11 @@ server {
         add_header Cross-Origin-Embedder-Policy "require-corp";
         add_header Service-Worker-Allowed "/" always;
 
+        types {
+            # mjs modules must have application/javascript, not default application/octet-stream
+            application/javascript mjs;
+        }
+
```
azhavoro, line 44: It seems that should be defined for *.mjs files only, something like this
```suggestion
        # mjs modules must have application/javascript, not default application/octet-stream
        location ~* \.mjs$ {
            types {
                application/javascript mjs;
            }
        }
```

bsekachev: The line `application/javascript mjs;` itself does not define that this is only for mjs extension in `/assets/` directory?

azhavoro: It is not entirely clear from the documentation, but most likely the `types` directive completely redefines the MIME types from `conf/mime.types`, rather than simply adding some values. Otherwise, the example from the documentation would not work.
image

bsekachev: It seems we may extend it instead of redefining https://stackoverflow.com/questions/16789494/extending-default-nginx-mime-types-file

azhavoro (approved):

PR #10079, opened 2025-12-05
Fix task creation from cloud storage with static chunks and custom range

SpecLad:
### Motivation and context

Currently this doesn't work, because:

1. `create_thread` calls `_download_data_from_cloud_storage` to download only those frames that match the frame range;

2. It tries to create a manifest, passing paths to _all_ frames.

3. `ImageManifestManager` tries to open a frame that hasn't been downloaded, and crashes.

Fix this by removing the frame filtering when downloading the frames. This is obviously not optimal performance-wise, but it's better than not working at all.

To be fair, this also slows down the case where a cloud storage manifest is used. IMO, it is acceptable to take a performance hit in this case to reduce complexity.

### How has this been tested?

I tested this manually; however, #10004 includes a test that exercises this case.

### Checklist

### License

### DIFF
--- a/cvat/apps/engine/task.py
+++ b/cvat/apps/engine/task.py
@@ -1133,6 +1133,10 @@ def create_thread(
 
             manifest = ImageManifestManager(db_data.get_manifest_path())
             if not manifest.exists:
+                # TODO: Try to avoid adding manifest entries for images that are not in
+                # extractor.frame_range. In addition to less processing here, it would also allow
+                # us to avoid downloading such images from cloud storage (when using static chunks),
+                # or copying them from the attached share (when using copy_data).
                 manifest.link(
                     sources=extractor.absolute_source_paths,
                     meta={


cvat/apps/engine/task.py
```diff
@@ -778,24 +778,11 @@ def update_status(msg: str) -> None:
...
-                    stop_frame = min(stop_frame, data['stop_frame'])
-
-                step = db_data.get_frame_step()
-                if start_frame or step != 1 or stop_frame != len(filtered_data) - 1:
-                    media_to_download = filtered_data[start_frame : stop_frame + 1: step]
-
-            _download_data_from_cloud_storage(db_data.cloud_storage, media_to_download, upload_dir)
-            del media_to_download
-            del filtered_data
+            _download_data_from_cloud_storage(
```
zhiltsov-max, line 781:
```suggestion
            # TODO: try to avoid storing records for unused files in the manifest
            _download_data_from_cloud_storage(
```

SpecLad: I'll put this comment next to where we actually create the manifest. This has implications for more than cloud storage (e.g. share storage with `copy_data`, and probably local storage too).

PR #10099, opened 2025-12-10
Fix duplicate image export issue 8076

huangcha03:
Closes https://github.com/cvat-ai/cvat/pull/9929
Closes https://github.com/cvat-ai/cvat/issues/8076

Motivation and context
The issue addressed is https://github.com/cvat-ai/cvat/issues/8076, which describes a problem where image names could conflict during dataset export when multiple images share the same name across tasks or jobs.
This caused overwriting or inconsistent naming during export.

This PR modifies the mangle_image_name function to handle duplicate image names more gracefully:

Returns the original filename for the first occurrence.
Otherwise, modifies the filename to now include "-task_{task_id}".
If the name is still taken, a numeric suffix (_1, _2, etc.) is appended.
This ensures consistent and conflict-free exports, especially in multi-task environments.
Also attempts to add unit testing for the changes.

How has this been tested?
Checklist
 I submit my changes into the develop branch
I have added tests to cover my changes
 I have linked related issues (https://github.com/cvat-ai/cvat/issues/8076)
License
 I submit my code changes under the same [MIT License](https://github.com/cvat-ai/cvat/blob/develop/LICENSE) that covers the project.


tests/python/rest_api/test_tasks.py
```diff
@@ -731,6 +731,78 @@ def interpolate(frame):
 @pytest.mark.usefixtures("restore_redis_ondisk_after_class")
 class TestGetTaskDataset:
 
+     # Add this fixture within your test class structure, perhaps in TestGetTaskDataset
+    @fixture(scope="class")
```
zhiltsov-max, line 735: The actual scope will depend on the calling test and `restore_*` marks defined there, so this scope is not guaranteed. If you intend to reuse this fixture in several tests, you'll need to move it into a dedicated class. I suggest simply removing this line.

tests/python/rest_api/test_tasks.py
```diff
@@ -731,6 +731,78 @@ def interpolate(frame):
...
+    def fxt_project_with_name_collision(
```
zhiltsov-max, line 736: Consider adding the 3rd task to test the logic proposed in the PR description and also annotations (e.g. 1 bbox with a source task name in an attribute) to make sure the annotations do not get mixed up.

tests/python/rest_api/test_tasks.py
```diff
@@ -989,6 +1061,42 @@ def test_datumaro_export_without_annotations_includes_image_info(
                     if "size" in related_image:
                         assert tuple(related_image["size"]) > (0, 0)
 
+    @pytest.mark.usefixtures("restore_db_per_function")
+    @pytest.mark.parametrize("project_id, expected_filenames", [fixture_ref(fxt_project_with_name_collision)])
+    def test_project_export_with_image_name_collision(
```
zhiltsov-max, line 1066: As this function tests project logic, it should be moved to `test_projects.py`.

tests/python/rest_api/test_tasks.py
```diff
@@ -989,6 +1061,42 @@ def test_datumaro_export_without_annotations_includes_image_info(
...
+                if name.endswith(".jpg"):
+                    # The image is expected to be in a subdirectory like 'images/default/...'
+                    # We need only the filename part for comparison.
+                    exported_image_names.add(osp.basename(name))
+
+        # The exported image names must exactly match the original name and the mangled name.
+        # Note: The expected_filenames are built as full filename with extension.
+        expected_image_names_set = set(expected_filenames)
+
+        assert exported_image_names == expected_image_names_set, \
```
zhiltsov-max, line 1097: I'd expect this test to check annotations as well.

cvat/apps/dataset_manager/bindings.py
zhiltsov-max, line 1: It seems there quite a few extra changes, please sync with the recent `develop` branch version and clean up extra changes.

cvat/apps/dataset_manager/bindings.py
```diff
@@ -1907,25 +2000,29 @@ def _clean_display_message(self) -> str:
             message = "Dataset must contain a file:" + message
         return re.sub(r' +', " ", message)
 
-def mangle_image_name(name: str, subset: str, names: defaultdict[tuple[str, str], int]) -> str:
-    name, ext = name.rsplit(osp.extsep, maxsplit=1)
+def mangle_image_name(name: str, subset: str, names: defaultdict[tuple[str, str], int], task_id: Optional[int] = None) -> str:
```
zhiltsov-max, line 2003:
```suggestion
def mangle_image_name(name: str, subset: str, names: defaultdict[tuple[str, str], int], task_id: int | None = None) -> str:
```

We're not using this old syntax anymore.

cvat/apps/dataset_manager/bindings.py
```diff
@@ -1907,25 +2000,29 @@ def _clean_display_message(self) -> str:
...
+        name_base = f"{name_base}-task_{task_id}"
+
+    names[(subset, name_base)] += 1
+    i = names[subset, name_base]
+    while i < sys.maxsize:
+        mangled_name = f"{name_base}_{i}"
+
+        if not names[(subset, mangled_name)]:
+            names[(subset, mangled_name)] = 1
+            return osp.extsep.join([mangled_name, ext])
```
zhiltsov-max, line 2022: If I understand correctly, this implementation will only produce names like `basename_task-X_Y`, despite what's said in the PR description. I think it would be better to avoid adding extra numerical suffix (`_Y`) if the name wasn't occupied.

PR #10102, opened 2025-12-11
TUS chunk validation

Grigorii777:
### Motivation and context
**Problem:**
The TUS upload implementation incorrectly validated chunk boundaries by checking only the **start offset** (`chunk.offset`) instead of the **end offset** (`chunk.offset + chunk.size`). This allowed clients to upload chunks that exceed the declared file size, potentially causing:
- Buffer overflows
- Data corruption
- Files larger than expected
- Violation of TUS protocol specification

**Example:**
- File size: 10,000 bytes
- Current offset: 9,000 bytes
- Chunk size: 2,000 bytes
- Old code: checks `9,000 > 10,000` -> passes (incorrect)
- New code: checks `11,000 > 10,000` -> fails with 413 (correct)

**Solution:**
1. Added `end_offset` property to `TusChunk` class that calculates `offset + size`
2. Changed validation in `mixins.py` from `chunk.offset > file_size` to `chunk.end_offset > file_size`
3. Added comprehensive test `test_cannot_upload_chunk_exceeding_file_size` to verify the fix

This ensures the server correctly rejects chunks whose end position would exceed the declared file size, preventing data corruption and enforcing TUS protocol compliance.

### How has this been tested?

**New test added:**
`test_cannot_upload_chunk_exceeding_file_size` - Verifies that:
1. First chunk (1000 bytes) uploads successfully
2. Second chunk with `offset=1000` and `size=(file_size - 500)` is rejected
3. Server returns 413 (Request Entity Too Large)
4. This ensures `end_offset = 1000 + (file_size - 500) > file_size` is properly validated

### Checklist

- [ ] ~~I have created a changelog fragment~~ 

### License

### DIFF
--- a/tests/python/rest_api/test_tus_uploads.py
+++ b/tests/python/rest_api/test_tus_uploads.py
@@ -188,6 +188,4 @@ class TestTUSUpload:
         oversized_chunk = b"x" * (file_size - 500)  # This will go beyond file_size
         response = self._upload_chunk(location, 1000, oversized_chunk, check_status=False)
 
-        assert (
-            response.status == HTTPStatus.REQUEST_ENTITY_TOO_LARGE
-        ), f"Expected 413 when chunk end exceeds file size, got {response.status}"
+        assert response.status == HTTPStatus.REQUEST_ENTITY_TOO_LARGE


cvat/apps/engine/mixins.py
SpecLad, line 1: Please add a changelog entry.

tests/python/rest_api/test_tus_uploads.py
```diff
@@ -0,0 +1,254 @@
...
+@pytest.mark.usefixtures("restore_db_per_function")
+@pytest.mark.usefixtures("restore_cvat_data_per_function")
+@pytest.mark.usefixtures("restore_redis_ondisk_per_function")
+@pytest.mark.usefixtures("restore_redis_inmem_per_function")
+class TestTUSUpload:
+    """Integration tests for TUS resumable upload protocol"""
+
+    _USERNAME = "admin1"
+
+    def _create_task_via_api(self, task_spec):
```
SpecLad:
```suggestion
    def _create_task(self, task_spec):
```

Everything is via the API.

tests/python/rest_api/test_tus_uploads.py
```diff
@@ -0,0 +1,254 @@
...
+        with make_api_client(self._USERNAME) as api_client:
+            (task, response) = api_client.tasks_api.create(task_spec)
+            assert response.status == HTTPStatus.CREATED
+            return task
+
+    def _start_upload(self, task_id, file_size, filename):
+        """Start TUS upload and return file_id"""
+        metadata = f"filename {base64.b64encode(filename.encode()).decode()}"
+
+        response = requests.post(
```
SpecLad: Please try to make these queries using the API client. You can use `.call_api` for making custom requests when the autogenerated functions are not sufficient.

tests/python/rest_api/test_tus_uploads.py
```diff
@@ -0,0 +1,254 @@
...
+                "Tus-Resumable": "1.0.0",
+            },
+            auth=(self._USERNAME, USER_PASS),
+        )
+
+        assert response.status_code == HTTPStatus.CREATED
+        location = response.headers.get("Location")
+        assert location is not None
+        file_id = location.split("/")[-1]
+        return file_id
```
SpecLad: Don't examine the returned URL. Just take it and use as-is in subsequent requests.

tests/python/rest_api/test_tus_uploads.py
```diff
@@ -0,0 +1,254 @@
...
+        return int(offset)
+
+    def test_can_upload_file_via_tus_in_single_chunk(self):
+        """Test uploading a complete file in one chunk"""
+        task = self._create_task_via_api(
+            {
+                "name": "test TUS single chunk upload",
+                "labels": [{"name": "car"}],
+            }
+        )
```
SpecLad: Perhaps you could move this into a fixture? It seems like most of these tests basically create the same task. I would also suggest to check if there's an existing fixture that'll work.

tests/python/rest_api/test_tus_uploads.py
```diff
@@ -0,0 +1,254 @@
...
+
+        file_id = self._start_upload(task.id, file_size, "test_image.jpg")
+
+        # Upload 40% of file
+        chunk1 = image_data[:first_chunk_size]
+        response = self._upload_chunk(task.id, file_id, 0, chunk1)
+        assert response.status_code == HTTPStatus.NO_CONTENT
+        assert int(response.headers.get("Upload-Offset")) == first_chunk_size
+
+        # Simulate interruption - check current offset
```
SpecLad: Where's the simulation? This test seems to basically be the same as the previous test.

tests/python/rest_api/test_tus_uploads.py
```diff
@@ -0,0 +1,254 @@
...
+        image_data = image_file.getvalue()
+        file_size = len(image_data)
+
+        file_id = self._start_upload(task.id, file_size, "test_image.jpg")
+        response = self._upload_chunk(task.id, file_id, 0, image_data)
+
+        assert response.status_code == HTTPStatus.NO_CONTENT
+        assert int(response.headers.get("Upload-Offset")) == file_size
+
+    def test_can_upload_file_via_tus_in_multiple_chunks(self):
```
SpecLad: This test seems completely redundant. `test_upload_offset_is_updated_incrementally` tests the same stuff and more.

tests/python/rest_api/test_tus_uploads.py
```diff
@@ -0,0 +1,254 @@
...
+        assert response.status_code == HTTPStatus.NO_CONTENT
+
+        # Get offset from response (file is not complete yet)
+        current_offset = int(response.headers.get("Upload-Offset"))
+        assert current_offset == 1000
+
+        # Try wrong offset (should be 1000, we send 500)
+        chunk2 = image_data[1000:2000]
+        response = self._upload_chunk(task.id, file_id, 500, chunk2)
+        assert response.status_code in (HTTPStatus.CONFLICT, HTTPStatus.INTERNAL_SERVER_ERROR)
```
SpecLad: This should never be an `INTERNAL_SERVER_ERROR`. What's going on here?

tests/python/rest_api/test_tus_uploads.py
```diff
@@ -0,0 +1,254 @@
...
+    def test_upload_offset_is_updated_incrementally(self):
+        """Test offset updates correctly after each small chunk"""
+        task = self._create_task_via_api(
+            {
+                "name": "test TUS incremental offset",
+                "labels": [{"name": "car"}],
+            }
+        )
+
+        image_file = generate_image_file("test_image.jpg", size=(150, 150))
```
SpecLad, line 112: Could you verify that the image is actually longer than `chunk_size`? Otherwise, this test is useless.

tests/python/rest_api/test_tus_uploads.py
```diff
@@ -0,0 +1,254 @@
...
+        """Test that upload fails if offset doesn't match"""
+        task = self._create_task_via_api(
+            {
+                "name": "test TUS wrong offset",
+                "labels": [{"name": "car"}],
+            }
+        )
+
+        # Generate larger image to ensure multiple chunks needed
+        image_file = generate_image_file("test_image.jpg", size=(200, 200))
```
SpecLad: Given that you never complete the upload in this test, there's no reason to generate valid data. You can just generate a string of zeros, which is faster and guarantees that it has the correct length.

tests/python/rest_api/test_tus_uploads.py
```diff
@@ -0,0 +1,254 @@
...
+    def test_cannot_upload_chunk_exceeding_file_size(self):
+        """Test that server rejects chunks whose end exceeds file size"""
+        task = self._create_task_via_api(
+            {
+                "name": "test TUS chunk exceeds file size",
+                "labels": [{"name": "car"}],
+            }
+        )
+
+        image_file = generate_image_file("test_image.jpg", size=(200, 200))
```
SpecLad: You also don't need a valid image for this test.

tests/python/rest_api/test_tus_uploads.py
```diff
@@ -0,0 +1,254 @@
...
+        assert int(response.headers.get("Upload-Offset")) == 1000
+
+        # Try to upload chunk that would exceed file size
+        # offset=1000, chunk size would make end_offset > file_size
+        oversized_chunk = b"x" * (file_size - 500)  # This will go beyond file_size
+        response = self._upload_chunk(task.id, file_id, 1000, oversized_chunk)
+
+        assert (
+            response.status_code == HTTPStatus.REQUEST_ENTITY_TOO_LARGE
+        ), f"Expected 413 when chunk end exceeds file size, got {response.status_code}"
```
SpecLad: This custom message is unnecessary. The default message pytest generates will contain the same information.

tests/python/rest_api/test_tus_uploads.py
```diff
@@ -0,0 +1,254 @@
...
+        # Upload first chunk
+        chunk1 = image_data[:1000]
+        response = self._upload_chunk(task.id, file_id, 0, chunk1)
+        assert response.status_code == HTTPStatus.NO_CONTENT
+
+        # Get offset from response (file is not complete yet)
+        current_offset = int(response.headers.get("Upload-Offset"))
+        assert current_offset == 1000
+
+        # Try wrong offset (should be 1000, we send 500)
```
SpecLad: I think for completeness it would make sense to try both something below the expected value and above it.

tests/python/rest_api/test_tus_uploads.py
```diff
@@ -0,0 +1,199 @@
...
+            )
+            assert response.status == HTTPStatus.CREATED
+            return task
+
+    def _call_tus_endpoint(self, method, path, headers=None, body=None, check_status=True):
+        """Call a low-level TUS endpoint via ApiClient"""
+        headers = headers or {}
+
+        with make_api_client(self._USERNAME) as api_client:
+            if path.startswith("http"):
```
SpecLad: This `if` doesn't make sense - if the condition is false, we just... don't do anything?

tests/python/rest_api/test_tus_uploads.py
```diff
@@ -0,0 +1,199 @@
...
+                if isinstance(body, io.BytesIO):
```
SpecLad: It doesn't seem like this is ever true.

changelog.d/20251215_170405_grigorio714_chunk_start_to_chunk_end.md
```diff
@@ -0,0 +1,4 @@
+### Fixed
+
+- Changed chunk validation from checking chunk.offset to chunk.end_offset for tus_file.file_size
```
SpecLad: This doesn't mean anything to a user. Please reframe this in terms that a user can understand.

tests/python/rest_api/test_tus_uploads.py
```diff
@@ -0,0 +1,199 @@
...
+    """Integration tests for TUS resumable upload protocol"""
+
+    _USERNAME = "admin1"
+
+    @pytest.fixture
+    def fxt_task(self):
+        """Fixture to create a task for TUS upload tests"""
+        with make_api_client(self._USERNAME) as api_client:
+            (task, response) = api_client.tasks_api.create(
+                {"name": "test TUS upload", "labels": [{"name": "car"}]}
```
SpecLad:
```suggestion
                {"name": "test TUS upload"}
```

AFAICS, labels are irrelevant for these tests.

tests/python/rest_api/test_tus_uploads.py
Grigorii777: Yes, I used it when I referred to call_api in one of the tests.

tests/python/rest_api/test_tus_uploads.py
Grigorii777: Yes, this is also left over from the old test with call_api.

PR #10106, opened 2025-12-11
Add openable file protocols and make manifest classes use them

SpecLad:
### Motivation and context

Introduce protocols, `Openable` and `NamedOpenable`, to be used instead of real paths in code that deals with media data. This is intended to help transparently support files in cloud storage in the future, although it just happens to make existing code simpler too.

The protocols contain just enough functionality to be useful: the ability to open in read-only binary mode, and (in the `Named` variant) the ability to get a path-like string for purposes of comparison and manifest creation. The protocol definitions are tailored so that `pathlib.Path` objects are `NamedOpenable` without any adaptation (I hope we can transition the codebase to use pathlib eventually, making the `Path` casts go away), and `zipfile.Path` objects are `Openable`.

Do a little bit of cleanup too: adding type annotations, removing unnecessary kwargs, etc.

### How has this been tested?

Mostly automated tests, although I also checked some cases manually:

* Creating a task from cloud storage with automatic manifest generation.
* Creating a task from a `.bin` point cloud.

### Checklist

### License


utils/dataset_manifest/core.py
```diff
@@ -21,22 +20,30 @@
...
+    sort,
+)
 
 # how many frames to check after seeking to validate key frame
 SEEK_MISMATCH_UPPER_BOUND = 200
 
 
 class VideoStreamReader:
-    def __init__(self, source_path, chunk_size, force):
+    def __init__(self, source_path: NamedOpenable, *, chunk_size: int, force: bool) -> None:
```
zhiltsov-max, line 39:
```suggestion
    def __init__(self, source_path: Openable, *, chunk_size: int, force: bool) -> None:
```
Probably, there's no point in requiring fspath support here, except maybe for better error messages.

utils/dataset_manifest/core.py
```diff
@@ -234,30 +243,26 @@ def __iter__(self) -> Iterator[int | tuple[int, int, str]]:
 class DatasetImagesReader:
     def __init__(
         self,
-        sources: list[str | NamedBytesIO] | Iterable[str | NamedBytesIO],
+        sources: Iterable[NamedOpenable],
```
zhiltsov-max, line 246: Can we restore support for `str` inputs and treat them like `Openable` or `NamedOpenable`, depending on the conext? This would simplify the client code and obsolete some of the changes made there. The sorting code below would remain the same, because `os.fspath` supports `str` and `bytes` input.
https://docs.python.org/3/library/os.html#os.fspath

In other places, like `link()` calls, as well. It looks like there should be a type like `OpenableLike = Openable | os.PathLike[str]` that would be cast to `Path()` if needed.

utils/dataset_manifest/core.py
```diff
@@ -621,10 +617,8 @@ def __init__(self, manifest_path, create_index=True):
         setattr(self._manifest, "TYPE", "video")
         self.BASE_INFORMATION["properties"] = 3
 
-    def link(self, media_file, upload_dir=None, chunk_size=36, force=False, **kwargs):
-        self._reader = VideoStreamReader(
-            os.path.join(upload_dir, media_file) if upload_dir else media_file, chunk_size, force
-        )
+    def link(self, media_file: NamedOpenable, *, chunk_size=36, force=False) -> None:
```
zhiltsov-max, line 620: `media_file` can probably be just `Openable` instead.

utils/dataset_manifest/create.py
```diff
@@ -106,7 +107,7 @@ def main():
             assert len(sources), "A images was not found"
             manifest = ImageManifestManager(manifest_path=manifest_directory)
             manifest.link(
-                sources=sources,
+                sources=list(map(Path, sources)),
```
zhiltsov-max, line 110: This extra client-side casting can potentially be removed, if the function supports `OpenableLike`, the one below as well.

utils/dataset_manifest/core.py
SpecLad: `NamedOpenable` is needed indirectly, because `VideoManifestManager._write_base_information` calls `os.path.basename(self._reader.source_path)`.

utils/dataset_manifest/core.py
SpecLad: I would rather solve this problem from the other direction and update the clients to use `Path` objects throughout. Part of my motivation here is to reduce the amount of branching, and I don't think I can support `str` without adding the branching back.

zhiltsov-max: Do you mean branching on the client side or the function implementation side? Unless there are considerable performance gains in doing otherwise, I'd prioritize having simpler client-side code.

SpecLad: > Do you mean branching on the client side or the function implementation side?

I mean branching in general, but the savings happen to come mostly from the implementation side. :slightly_smiling_face: I did remove one `if` in the tests, but that was just a lucky coincidence.

> Unless there are considerable performance gains in doing otherwise, I'd prioritize having simpler client-side code.

My point is that client and server code can _both_ be simple, if we update everything to use `Path` consistently. I just can't do it everywhere at once, because then the PR will be enormous. :slightly_smiling_face:

I've already done some of that update work here: #10128. You can see there that the `Path` casts tend to disappear.

utils/dataset_manifest/core.py
SpecLad: No, it has to be a `NamedOpenable` because it's passed to `VideoStreamReader` (and that needs a `NamedOpenable` for reasons I explained in another thread).

utils/dataset_manifest/core.py
zhiltsov-max: Ok, let's merge after the next release

PR #10125, opened 2025-12-16
Modernize SDK packaging

SpecLad:
### Motivation and context

Disable the autogeneration of packaging-related files by OpenAPI Generator. We're basically either suppressing and overriding all of them (except for `requirements/api_client.txt`), so we're not getting any value from having them generated.

Instead, add our own packaging configuration. This allows us to use modern packaging standards:

* Switch to declarative project metadata (the `project` table in `pyproject.toml`).

* Remove setuptools from runtime dependencies.

* Remove wheel from build dependencies (it is no longer needed, as the relevant functionality has been merged into setuptools).

* Use PEP-639-style licensing metadata (note that this entails removing the `License :: OSI Approved :: MIT License` classifier).

Also, set better values for the summary, author and keywords fields.

Note that setting `generateSourceCodeOnly` to `true` causes OpenAPI Generator to output docs into `cvat_sdk/api_client` for some reason - hence the changes in `generate.sh`.

### How has this been tested?

I have compared the sdists and wheels of the SDK before and after the patch and verified that only the expected changes are present. I also verified that the generated docs are identical.

### Checklist

### License

### DIFF
--- a/cvat-sdk/pyproject.toml
+++ b/cvat-sdk/pyproject.toml
@@ -21,9 +21,11 @@ dependencies = [
     "packaging >= 21.3",
     "Pillow >= 10.3.0",
     "platformdirs >= 2.1.0",
-    "python_dateutil >= 2.5.3",
     "tqdm >= 4.64.0",
     "typing_extensions >= 4.2.0",
+
+    # these dependencies are used by auto-generated code
+    "python_dateutil >= 2.5.3",
     "urllib3 >= 1.25.3",
 ]
 


cvat-sdk/pyproject.toml
```diff
@@ -1,7 +1,50 @@
...
+    "Programming Language :: Python :: 3",
+    "Operating System :: OS Independent",
+]
+requires-python = ">=3.10"
+dependencies = [
+    "attrs >= 21.4.0",
+    "packaging >= 21.3",
+    "Pillow >= 10.3.0",
+    "platformdirs >= 2.1.0",
+    "python_dateutil >= 2.5.3",
```
zhiltsov-max: I'm ok with the change, but we need a way to check that the low-level API dependencies are included in the list. It's not critical now, as we don't update very often, but anyway. Maybe include copying requirements into `pyproject.toml` as a postprocessing step in `generate.sh`?

SpecLad: I don't think this is a problem worth solving. The list of dependencies used by the generated code has not changed until OpenAPI Generator v7.0.0, in which the entire generator we're using was discontinued. So if we upgrade the generator to a v6.x version, this won't need updating, whereas if we upgrade to v7.x, it'll pretty much require a rewrite of the SDK; updating the dependencies will be the least of our problems.

But just in case, I have visually separated the requirements used by the generated code from the rest.

zhiltsov-max (approved):

PR #10131, opened 2025-12-17
Fix bulk actions progress bar test

archibald1418:
The progress bar appears and disappears very quickly when bulk deleting, occasionally leading to cypress still looking for the progress bar. This adds an intercept to the first request in the batch so that cypress doesn't miss it

### Checklist

### License

### DIFF
--- a/tests/cypress/e2e/features/bulk_actions.js
+++ b/tests/cypress/e2e/features/bulk_actions.js
@@ -206,6 +206,8 @@ context('Bulk actions in UI', () => {
         });
 
         it('Delete all tasks, ensure deletion', () => {
+            cy.intercept('DELETE', '/api/tasks/**').as('deleteTask');
+
             getBulkActionsMenu().within(() => {
                 cy.contains(`Delete (${numberOfObjects})`).click();
             });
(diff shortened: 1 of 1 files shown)


tests/cypress/e2e/features/bulk_actions.js
```diff
@@ -206,18 +206,24 @@ context('Bulk actions in UI', () => {
...
                     cy.contains(`Delete ${numberOfObjects} selected tasks`);
-                    cy.contains('Delete selected')
-                        .should('be.visible')
-                        .click();
                 });
-            cy.get('.cvat-bulk-progress-wrapper').should('be.visible');
+            cy.contains('Delete selected')
+                .should('be.visible')
+                .click();
+            cy.wait('@deleteTask').then(() => {
```
klakhov, line 222: I believe we should display the progress wrapper first, and only then send the request. That’s how it’s supposed to work, at least conceptually. Does it actually work the other way around right now?

archibald1418: The test used to work like that before, but it didn't work and lead to flake

PR #10134, opened 2025-12-18
Fix manifest and frame order mismatches in task creations

### LINKED ISSUE #10104: Task creation fails if a manifest is present but frame order does not match it

### Steps to Reproduce

```jsonl
{"version":"1.1"}
{"type":"images"}
{"name":"image_1",...}
{"name":"image_0",...}
```
4. click "create a new task"
6. click "submit"

### Expected Behavior

Successfully created task.

### Possible Solution

Option 1: ignore the manifest

### Context

_No response_

### Environment

```Markdown

```

RavinduWeerakoon:
### Motivation and context

closes #10104 

### Checklist

### License

Fixes

### DIFF
--- a/cvat/apps/engine/task.py
+++ b/cvat/apps/engine/task.py
@@ -1135,16 +1135,16 @@ def create_thread(
                 manifest.init_index()
 
             # Align manifest order with the selected sorting method for image-based tasks.
-            try:
-                if data['sorting_method'] != models.SortingMethod.PREDEFINED:
-                    desired_order = [
-                        os.path.relpath(extractor.get_path(frame_id), upload_dir)
-                        for frame_id in extractor.frame_range
-                    ]
+            if data['sorting_method'] != models.SortingMethod.PREDEFINED:
+                desired_order = [
+                    os.path.relpath(extractor.get_path(frame_id), upload_dir)
+                    for frame_id in extractor.frame_range
+                ]
+
+                try:
                     manifest.reorder(desired_order)
-            except Exception as e:
-                # If reordering fails for some reason, proceed without altering
-                raise ValidationError("Manifest reordering failed") from e
+                except Exception as e:
+                    raise ValidationError("Manifest reordering failed") from e
 
             for frame_id in extractor.frame_range:
                 image_path = extractor.get_path(frame_id)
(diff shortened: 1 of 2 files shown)


cvat/apps/engine/task.py
zhiltsov-max, line 1: Please add a test for this change. Such tests typically go into the `tests/python/rest_api` directory.

cvat/apps/engine/task.py
```diff
@@ -1140,6 +1135,18 @@ def update_status(msg: str) -> None:
...
+            try:
+                if data['sorting_method'] != models.SortingMethod.PREDEFINED:
+                    desired_order = [
+                        os.path.relpath(extractor.get_path(frame_id), upload_dir)
+                        for frame_id in extractor.frame_range
+                    ]
+                    manifest.reorder(desired_order)
+            except Exception:
+                # If reordering fails for some reason, proceed without altering
+                raise ValidationError("Manifest reordering failed")
```
zhiltsov-max: The comment contradicts the code below.

```suggestion
            except Exception as e:
                # If reordering fails for some reason, proceed without altering
                raise ValidationError("Manifest reordering failed") from e
```

```suggestion
            if data['sorting_method'] != models.SortingMethod.PREDEFINED:
                desired_order = [
                    os.path.relpath(extractor.get_path(frame_id), upload_dir)
                    for frame_id in extractor.frame_range
                ]
                try:
                    manifest.reorder(desired_order)
                except Exception:
                    # If reordering fails for some reason, proceed without altering
                    raise ValidationError("Manifest reordering failed")
```
Please check if the exception class can be narrowed. It you want to catch a broad exception class, minimize the scope that can throw an exception.

RavinduWeerakoon:
Thanks, I added a test case. Not sure whether it's the right class though

RavinduWeerakoon:
Hi @zhiltsov-max, any updates on this

tests/python/rest_api/test_tasks.py
```diff
@@ -449,6 +450,51 @@ def test_can_create_without_labels(self, admin_user):
...
+            "image_quality": 75,
+            "client_files": client_files,
+            "sorting_method": "natural",
+        }
+
+        task_id, _ = create_task(admin_user, task_spec, data_spec)
+
+        with make_api_client(admin_user) as api_client:
+            (task, _) = api_client.tasks_api.retrieve(task_id)
+            assert task.size == 3
```
zhiltsov-max, line 493: Probably, it makes sense to also check the resulting file order here. It can be done via the `tasks/<id>/data/meta` endpoint.

zhiltsov-max:
Please update the existing tests so that the tests are passing.

PR #10152, opened 2025-12-23
Fix chapters for videos without manifest

zhiltsov-max:
### Motivation and context

It's possible that some tasks with videos have no manifest. There are several reasons for that, including manual playback speed manipulation attepts. CVAT will generate no manifest for such videos, but the chapters functionality will try to read the chapters from the manifest regardless. It's technically possible to ignore bad keyframes, but it can lead to invalid chapters.

- Fixed task access for tasks with videos with bad keyframes

### How has this been tested?

Unit tests and manual tests

### Checklist

### License

### DIFF
--- a/tests/python/rest_api/_test_base.py
+++ b/tests/python/rest_api/_test_base.py
@@ -3,7 +3,6 @@ import math
 import os
 from collections.abc import Generator, Mapping, Sequence
 from contextlib import closing
-from fractions import Fraction
 from functools import partial
 from typing import IO
 
@@ -536,10 +535,8 @@ class TestTasksBase:
 
             chapters = [
                 {
-                    "id": 0,
                     "start": 0,
-                    "end": 100,
-                    "time_base": Fraction(1, 1000),
+                    "stop": 5,
                     "metadata": {"title": "Intro"},
                 }
             ]
--- a/tests/python/rest_api/test_task_data.py
+++ b/tests/python/rest_api/test_task_data.py
@@ -1782,7 +1782,8 @@ class TestTaskData(TestTasksBase):
     def test_can_get_task_meta(self, task_spec: ITaskSpec, task_id: int):
 
         with make_api_client(self._USERNAME) as api_client:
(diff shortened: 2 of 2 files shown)


tests/python/rest_api/_test_base.py
```diff
@@ -522,8 +525,25 @@ def _uploaded_video_task_fxt_base(
             range(start_frame or 0, (stop_frame or frame_count - 1) + 1, step or 1)
         )
 
-        video_file = generate_video_file(frame_count)
-        video_data = video_file.getvalue()
+        assert (not video_file) ^ (
+            chapters is not None
+        ), "Chapters must be specified with a custom video file"
```
SpecLad, line 530: Does this condition really need to be here? It seems like these parameters could be processed independently:

```
if chapters is None:
    chapters = [...]

if video_file is None:
    video_file = ...
```

tests/python/rest_api/_test_base.py
```diff
@@ -510,6 +511,8 @@ def _uploaded_video_task_fxt_base(
         start_frame: int | None = None,
         stop_frame: int | None = None,
         step: int | None = None,
+        video_file: IO[bytes] | None = None,
```
SpecLad, line 514: Perhaps instead of this it would be better to have a parameter like `video_kwargs: dict | None`? `fxt_uploaded_video_task_without_manifest` would then look like:

```
        yield from self._uploaded_video_task_fxt_base(
            request=request, video_kwargs={"invalid_keyframes": True}, chapters=[]
        )
```

And the logic in this function would become simpler.

tests/python/rest_api/_test_base.py
zhiltsov-max: They could be, but it's not how they are supposed to be used now.

tests/python/rest_api/_test_base.py
zhiltsov-max: Yes, I started from this, but I found that if there are more different cases, it's more clear to pass already created video instead of just the parameters. One of the new cases was removed later (the one that tests partially invalid PTS values), because there's no difference in the handling at the moment.

PR #10153, opened 2025-12-23
Fix related image chunks in tasks from share

zhiltsov-max:
### Motivation and context

Related https://github.com/cvat-ai/cvat/issues/9978
Obsoletes https://github.com/cvat-ai/cvat/pull/9984, https://github.com/cvat-ai/cvat/pull/9985

Backups are broken too, they will be fixed in another PR, as there are some deep issues discussed in https://github.com/cvat-ai/cvat/pull/10004#discussion_r2599348149

- Fixed related image chunks in image tasks from share

### Checklist

### License

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -501,8 +501,8 @@ jobs:
         # We don't have external services in Helm tests, so we ignore corresponding cases
         # They are still tested without Helm
         run: |
-          tar -C tests/mounted_file_share/ -c . | \
-            kubectl exec $(kubectl get pods -l component=server -o jsonpath='{.items[0].metadata.name}') -- tar -C /home/django/share -x
+          kubectl cp tests/mounted_file_share $(kubectl get pods -l component=server -o jsonpath='{.items[0].metadata.name}'):/home/django/share/
+          kubectl exec $(kubectl get pods -l component=server -o jsonpath='{.items[0].metadata.name}') -- bash -c "mv /home/django/share/mounted_file_share/* /home/django/share/ && rm -r /home/django/share/mounted_file_share"
           pytest --timeout 30 --platform=kube -m "not with_external_services" tests/python --log-cli-level DEBUG
 
       - name: Creating a log file from "cvat" container logs


.github/workflows/main.yml
```diff
@@ -501,8 +501,8 @@ jobs:
         # We don't have external services in Helm tests, so we ignore corresponding cases
         # They are still tested without Helm
         run: |
-          kubectl cp tests/mounted_file_share/images $(kubectl get pods -l component=server -o jsonpath='{.items[0].metadata.name}'):/home/django/share
-          kubectl cp tests/mounted_file_share/videos $(kubectl get pods -l component=server -o jsonpath='{.items[0].metadata.name}'):/home/django/share
+          kubectl cp tests/mounted_file_share $(kubectl get pods -l component=server -o jsonpath='{.items[0].metadata.name}'):/home/django/share/
+          kubectl exec $(kubectl get pods -l component=server -o jsonpath='{.items[0].metadata.name}') -- bash -c "mv /home/django/share/mounted_file_share/* /home/django/share/ && rm -r /home/django/share/mounted_file_share"
```
SpecLad, line 505: Couldn't this be simplified to 

```
kubectl cp tests/mounted_file_share/* $(kubectl get pods -l component=server -o jsonpath='{.items[0].metadata.name}'):/home/django/share/
```

?

tests/python/rest_api/_test_base.py
```diff
@@ -21,6 +22,14 @@
 from shared.utils.config import make_api_client
 from shared.utils.helpers import generate_image_files, generate_video_file
 
+SHARE_DIR = Path(os.path.dirname(__file__), "../../mounted_file_share").resolve()
```
SpecLad: Seems like it would be useful to have this available globally, in `tests/python/shared/utils/config.py`.

SpecLad:
```suggestion
SHARE_DIR = Path(__file__).parents[2] / "mounted_file_share"
```

It just seems weird to combine `pathlib` and `os.path` functions in the same line. 🤷‍♂️

tests/python/rest_api/_test_base.py
```diff
@@ -96,7 +105,11 @@ def get_frame(i: int) -> bytes:
 
             def get_related_files(i: int) -> Mapping[str, bytes]:
                 frame_ri = related_files.get(i)
-                common_prefix = os.path.commonpath(os.path.dirname(f.name) for f in frame_ri)
+
+                common_prefix = ""
```
SpecLad, line 109: FWIW, I think this assignment could be removed.

.github/workflows/main.yml
zhiltsov-max: The official example 2 says to use `kubectl exec` and the reasons are apparently discussed [here](https://devops.stackexchange.com/questions/6669/copy-multiple-files-using-wildcard-from-kubernetes-container). I didn't try to dig this quesiton deeper.

tests/python/rest_api/_test_base.py
zhiltsov-max: We need the condition, because `commonpath()` fails for an empty sequence in the argument. If the line is just removed, it doesn't seem good to me to reference an undefined variable in the return statement.

.github/workflows/main.yml
SpecLad: Both the example 2 and the question you link deal with downloading files _from_ a pod. Here we're uploading files _to_ a pod, which is much easier, since the shell will expand the wildcard.

tests/python/rest_api/_test_base.py
SpecLad: Yeah, okay.

.github/workflows/main.yml
SpecLad: Ah, okay, I see that `kubectl cp` doesn't actually support more than one source. :slightly_frowning_face:

How about this then?

```
tar -C tests/mounted_file_share/ -c . | \
    kubectl exec $(kubectl get pods -l component=server -o jsonpath='{.items[0].metadata.name}') -- tar -C /home/django/share -x
```

zhiltsov-max: I can try this, but it doesn't really look significantly better than the previous implementation.

SpecLad:
Hmm, this should probably not be marked as fixing #9978, since that includes the backup issue, which is not fixed here.

PR #10170, opened 2026-01-07
Feat about AI Tools (autoClassifier) for annotation

Yakuho:
### Motivation and context

During the daily annotation process, I have noticed that as the number of Project tags increases, although using SAM for assisted annotation can speed up the polygon drawing process, when selecting the corresponding tags, the current fastest method is to search by entering keywords due to the large number of tags. As a deep learning developer, I believe this issue can be completely delegated to a classification model to perform the task of automatic categorization. Therefore, I have implemented this approach and have experienced relatively good results in practical use.

Why does the classification model work? The reason is that the classification task is simple enough, and the accuracy of classification can easily exceed 90% (if the classification tasks are not overly similar, as in most practical deep learning classification tasks, the categories are generally distinguishable, leading to high accuracy). What does this mean? It means that out of 100 annotated objects, approximately 90 may not require tag adjustments, and only about 10 tags might need manual keyword input or selection from a dropdown menu to correct the labels.

How to train your own classification model? At the beginning of the annotation task, of course, a classification model cannot be used because there is no effective data available for training. However, once the data accumulates to a certain extent (not too much—perhaps 50 to 100 images per category), it becomes possible to train an automatic category classification model for this classification task. At this stage, relatively good results can be achieved. As your annotation task progresses, each time the data increases, updating your automatic classification model through training will enhance its categorization capability gradually. Achieving an accuracy rate of over 95% will become effortless. The specific training method involves cropping each object after annotating targets in CVAT, and arranging the label format according to the training format required for the model you intend to apply. For instance, training can be conducted using the dataset format for classification tasks in [Ultralytics](https://docs.ultralytics.com/datasets/classify/) (if needed, I can help users in the issue section with questions regarding training methods for Ultralytics or other models, or you can consult ChatGPT for training guidance).

showing: 

https://github.com/user-attachments/assets/4e39dcad-dcc4-4d33-971b-6ef782fe9dce

### How has this been tested?

I am debugging `cvat-ui` controls to verify their functionality based on this document: `https://docs.cvat.ai/docs/contributing/development-environment/`
use, check and build with
```bash
yarn run start:cvat-ui
yarn run build:cvat-ui
```
at last, after multiple people used it continuously for several days, there were no any issues.

### Checklist

- [ ] ~~I have created a changelog fragment ~~
- [ ] ~~I have linked related issues (see [GitHub docs](

### License


bsekachev:
Hello,

This is a good feature, but the current PR seems more like a prototype. 
We are not ready to accept and support it. I believe such major features have to be well designed and discussed first.

PR #10172, opened 2026-01-08
Improved consensus UX

klakhov:
### Motivation and context

This PR implements a simplified consensus scoring system for merged annotations. Instead of filtering annotations based on quorum thresholds, all annotations are now merged and assigned a consensus score (0.0 to 1.0) that represents the level of agreement among annotators.

Also PR contains a bunch of improvements to review mode allowing to unlock and edit objects, navigate in between objects using shortcuts

### Checklist

### License

### DIFF
--- a/cvat/apps/engine/serializers.py
+++ b/cvat/apps/engine/serializers.py
@@ -1431,7 +1431,6 @@ class JobValidationLayoutWriteSerializer(serializers.Serializer):
                         chunk_real_frame, quality=quality
                     ).data,
                     os.path.basename(db_frame_path),
-                    chunk_frame,
                 )
 
         with closing(_iterate_chunk_frames()) as frame_iter:
@@ -3928,6 +3927,11 @@ class AnnotationGuideReadSerializer(WriteOnceMixin, serializers.ModelSerializer)
 class AnnotationGuideWriteSerializer(WriteOnceMixin, serializers.ModelSerializer):
     project_id = serializers.IntegerField(required=False, allow_null=True)
     task_id = serializers.IntegerField(required=False, allow_null=True)
+
+    def validate_markdown(self, markdown: str) -> str:
+        if len(models.AnnotationGuide.get_asset_ids_from_markdown(markdown)) > settings.ASSET_MAX_COUNT_PER_GUIDE:
+            raise serializers.ValidationError("Maximum number of assets per guide reached")
+        return markdown
 
     @transaction.atomic
     def create(self, validated_data):
--- a/changelog.d/20260113_174651_sekachev.bs_updated_assets_count.md
+++ b/changelog.d/20260113_174651_sekachev.bs_updated_assets_count.md
@@ -1,0 +1,9 @@
+### Fixed
+
+- Missing validation for the maximum number of annotation guide assets
+  (<https://github.com/cvat-ai/cvat/pull/10185>)
+
+### Changed
+
+- Increased the maximum number of assets per guide from 30 to 150
+  (<https://github.com/cvat-ai/cvat/pull/10185>)
--- a/changelog.d/20260127_101859_sekachev.bs_fix_3d_camera_pos.md
+++ b/changelog.d/20260127_101859_sekachev.bs_fix_3d_camera_pos.md
@@ -1,0 +1,4 @@
+### Fixed
+
+- Incorrect camera position on 3d scene when position of point clouds
+  shifted from center of coordinates (<https://github.com/cvat-ai/cvat/pull/10221>)
--- a/components/analytics/grafana/dashboards/monitoring.json
+++ b/components/analytics/grafana/dashboards/monitoring.json
@@ -953,9 +953,119 @@
       ],
       "title": "Exceptions table",
       "type": "table"
+    },
+    {
+      "datasource": {
+        "type": "grafana-clickhouse-datasource",
+        "uid": "PDEE91DDB90597936"
+      },
+      "description": "MegaBytes per second grouped by queue",
+      "fieldConfig": {
+        "defaults": {
+          "color": {
+            "mode": "palette-classic"
+          },
+          "custom": {
+            "axisCenteredZero": false,
+            "axisColorMode": "text",
+            "axisLabel": "",
+            "axisPlacement": "auto",
+            "barAlignment": 0,
+            "drawStyle": "line",
+            "fillOpacity": 50,
+            "gradientMode": "none",
+            "hideFrom": {
+              "legend": false,
+              "tooltip": false,
+              "viz": false
+            },
+            "insertNulls": 5000,
+            "lineInterpolation": "smooth",
+            "lineStyle": {
+              "fill": "solid"
+            },
+            "lineWidth": 1,
+            "pointSize": 5,
+            "scaleDistribution": {
+              "type": "linear"
+            },
+            "showPoints": "auto",
+            "spanNulls": false,
+            "stacking": {
+              "group": "A",
+              "mode": "normal"
+            },
+            "thresholdsStyle": {
+              "mode": "off"
+            }
+          },
+          "mappings": [],
+          "thresholds": {
+            "mode": "absolute",
+            "steps": [
+              {
+                "color": "green",
+                "value": null
+              },
+              {
+                "color": "red",
+                "value": 80
+              }
+            ]
+          },
+          "unit": "MBs"
+        },
+        "overrides": []
+      },
+      "gridPos": {
+        "h": 16,
+        "w": 24,
+        "x": 0,
+        "y": 40
+      },
+      "id": 13,
+      "options": {
+        "legend": {
+          "calcs": [],
+          "displayMode": "list",
+          "placement": "bottom",
+          "showLegend": true
+        },
+        "tooltip": {
+          "mode": "single",
+          "sort": "none"
+        }
+      },
+      "targets": [
+        {
+          "datasource": {
+            "type": "grafana-clickhouse-datasource",
+            "uid": "PDEE91DDB90597936"
+          },
+          "editorType": "sql",
+          "format": 0,
+          "meta": {
+            "builderOptions": {
+              "columns": [],
+              "database": "",
+              "limit": 1000,
+              "mode": "list",
+              "queryType": "table",
+              "table": ""
+            }
+          },
+          "pluginVersion": "4.8.2",
+          "queryType": "timeseries",
+          "rawSql": "SELECT\r\n    $__timeInterval(timestamp) AS time,\r\n    JSONExtractString(payload, 'queue') AS queue,\r\n    sum(JSONExtractUInt(JSONExtractRaw(payload, 'cache_item'), 'size')) \r\n        / $__interval_s \r\n        / 1048576 AS Throughput\r\nFROM events\r\nWHERE\r\n    $__timeFilter(timestamp)\r\n    AND scope = 'create:cache_item'\r\nGROUP BY\r\n    time,\r\n    queue\r\nORDER BY\r\n    time ASC",
+          "refId": "A"
+        }
+      ],
+      "title": "Cache Throughput by Queue",
+      "transformations": [],
+      "type": "timeseries"
     }
   ],
-  "refresh": "",
+  "refresh": false,
   "schemaVersion": 38,
   "style": "dark",
   "tags": [],
@@ -1028,6 +1138,10 @@
           "value": [
             "$__all"
           ]
+        },
+        "datasource": {
+          "type": "grafana-clickhouse-datasource",
+          "uid": "PDEE91DDB90597936"
         },
         "definition": "SELECT DISTINCT user_name\nFROM events\nWHERE $__timeFilter(timestamp)",
         "hide": 0,
@@ -1198,6 +1312,6 @@
   "timezone": "",
   "title": "Monitoring",
   "uid": "WvDvWK04k",
-  "version": 3,
+  "version": 4,
   "weekStart": ""
 }
--- a/cvat-canvas3d/src/typescript/canvas3dView.ts
+++ b/cvat-canvas3d/src/typescript/canvas3dView.ts
@@ -137,7 +137,6 @@ export class Canvas3dViewImpl implements Canvas3dView, Listener {
         };
         scan: any;
         rotation: any;
-        frameCoordinates: any;
         detected: any;
         initialMouseVector: any;
     };
@@ -207,11 +206,6 @@ export class Canvas3dViewImpl implements Canvas3dView, Listener {
         this.isCtrlDown = false;
         this.action = {
             scan: null,
-            frameCoordinates: {
-                x: 0,
-                y: 0,
-                z: 0,
-            },
             detected: false,
             initialMouseVector: new THREE.Vector2(),
             translation: {
@@ -674,10 +668,10 @@ export class Canvas3dViewImpl implements Canvas3dView, Listener {
         up.copy(new THREE.Vector3().crossVectors(right, forward).normalize());
 
         // 4) Project corners onto right/up to get world extents independent of current distance
-        let minU = +Infinity; let
-            maxU = -Infinity;
-        let minV = +Infinity; let
-            maxV = -Infinity;
+        let minU = +Infinity;
+        let maxU = -Infinity;
+        let minV = +Infinity;
+        let maxV = -Infinity;
         for (const c of worldCorners) {
             const u = c.dot(right);
             const v = c.dot(up);
@@ -715,7 +709,7 @@ export class Canvas3dViewImpl implements Canvas3dView, Listener {
     }
 
     private fitCanvas(animation: boolean): void {
-        const { x, y, z } = this.action.frameCoordinates;
+        const [x, y, z] = this.cameraSettings.perspective.lookAt;
         this.positionAllViews(x, y, z, animation);
         this.updateCameraFrustumPlane();
     }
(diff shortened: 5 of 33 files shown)


cvat-canvas/src/typescript/canvasView.ts
zhiltsov-max, line 1: I can't find that the score is actually being displayed anywhere in the review mode, do we have it somewhere? Without it, it may be harder to pick the threshold for filtering and actually remember to do some filtering at all.

zhiltsov-max, line 1: Tracked shapes seem to miss the score attribute, so it's not possible to see them after filtering by Votes or Score. Actually, they should not be produced by the algorithm, but it's a separate issue.

cvat/apps/consensus/intersect_merge.py
zhiltsov-max, line 1: Could you update skeleton merger to also overwrite the `source` attribute to `consensus` for keypoints? Apparently, it's missing now.

cvat/apps/consensus/intersect_merge.py
```diff
@@ -452,14 +437,11 @@ def merge_cluster_shape(self, cluster: Sequence[dm.Annotation]) -> tuple[dm.Anno
...
 
-        # when the merged annotation is rejected due to quorum constraint
-        if label is None:
-            return None
-
         shape, shape_score = self.merge_cluster_shape(cluster)
         shape.z_order = max(cluster, key=lambda a: a.z_order).z_order
         shape.label = label
-        shape.attributes["score"] = label_score * shape_score
+        shape.attributes["consensus_score"] = label_score * shape_score
```
zhiltsov-max: Probably, this line should be updated. The `consensus_score` attribute is used only here.

tests/python/shared/assets/annotations.json
zhiltsov-max, line 1: Please remove the changes where only the annotation order is changed.

cvat/apps/consensus/intersect_merge.py
```diff
@@ -400,25 +393,17 @@ def merge_clusters(self, clusters):
...
 
-    def find_cluster_label(self, cluster: Sequence[dm.Annotation]) -> tuple[int | None, float]:
+    def find_cluster_label(self, cluster: Sequence[dm.Annotation]) -> tuple[int, float]:
         votes = {}
         for s in cluster:
             label = self._context.get_src_label_name(s, s.label)
-            state = votes.setdefault(label, [0, 0])
-            state[0] += s.attributes.get("score", 1.0)
-            state[1] += 1
+            votes[label] = votes.get(label, 0) + s.attributes.get("score", 1.0)
```
zhiltsov-max:
```suggestion
            votes[label] = votes.get(label, 0) + 1
```
I think, it's better to just use 1 instead, to avoid potential obscure problems with user-imported score values.

cvat/apps/dataset_manager/bindings.py
zhiltsov-max, line 1: Please update `CVAT_INTERNAL_ATTRIBUTES`.

cvat/apps/consensus/intersect_merge.py
zhiltsov-max, line 1:
```python
IntersectMerge(conf={..., ignored_attributes={"track_id", "keyframe"}})

from datumaro.util import filter_dict

class IntersectMerge:
    ...

    def _merge_clusters(self, t, clusters):
        # A workaround for the incorrect implementation of attribute filtering in Datumaro
        # TODO: fix in Datumaro
        merged_annotations = super()._merge_clusters(t, clusters)

        for ann in merged_annotations:
            ann.attributes = filter_dict(ann.attributes, exclude_keys=self.conf.ignored_attributes)

        return merged_annotations
```

I found that there are no special actions about tracks in the merging algorithm, but they should not be supported. Here is the workaround, if you want to include it in this PR and avoid being blocked.

cvat-canvas/src/typescript/canvasView.ts
klakhov: To display Score/Votes we need to enable it in settings (disabled by default, probably need to enable by default)
image
image

zhiltsov-max: Can we just show it by default in consensus tasks? It looks quite natural to show it there. Maybe, not in tasks, but if the annotation source is consensus.

cvat/apps/consensus/intersect_merge.py
klakhov: Thanks. Applied the workaround.

cvat-canvas/src/typescript/canvasView.ts
klakhov: Not relevant, I suppose, if we use workaround with ignore `track_id`

cvat/apps/consensus/intersect_merge.py
klakhov: Added

cvat/apps/consensus/intersect_merge.py
klakhov: Removed this attr

cvat/apps/dataset_manager/bindings.py
klakhov: Added `score` and `source`

tests/python/shared/assets/annotations.json
klakhov: Do we really need to revert this change? In the places where the order was modified, the ID ordering was probably incorrect (a larger ID appeared before a smaller one).

zhiltsov-max: The problem is that the values are returned in an arbitrary order, which depends on the python interpreter, so we typically clean up this manually. I agree that it's not very convenient, but we haven't found a better solution yet. But if you're sure the items were out of order, I'd prefer to keep the change.

cvat-canvas/src/typescript/canvasView.ts
bsekachev: Agree, such information must be always shown in consensus task and never otherwise

changelog.d/20260114_132218_klakhov_consensus_ux.md
```diff
@@ -0,0 +1,19 @@
+### Changed
+
+- Consensus merge function now preserves all shapes with their scores, regardless of quorum threshold
+  (<https://github.com/cvat-ai/cvat/pull/10172>)
+- In review mode, all annotations are locked by default; Users can unlock and edit individual annotations as needed
+  (<https://github.com/cvat-ai/cvat/pull/10172>)
+- Next/Previous object navigation (`Shift+Tab`/`Shift`) now works in Standard, Review, Attribute Annotation workspaces
+  (<https://github.com/cvat-ai/cvat/pull/10172>)
```
bsekachev: Let's put this change to Added section

e.g. A user now may navigate between different shapes with shortcuts (`Tab`/`Shift+Tab` by default) in **Standard** and **Review** modes

changelog.d/20260114_132218_klakhov_consensus_ux.md
```diff
@@ -0,0 +1,19 @@
...
+
+### Added
+
+- Score visualization in UI with a virtual "Votes" attribute calculated as `score × replica_jobs`
+  (<https://github.com/cvat-ai/cvat/pull/10172>)
+- Double-clicking an object in the sidebar now centers it on the canvas and expands its details
+  (<https://github.com/cvat-ai/cvat/pull/10172>)
+
+### Removed
+- Consensus `quorum` setting
```
bsekachev: It hardly says something to those regular users who reads this changelog. May you please rephrase/delete?

changelog.d/20260114_132218_klakhov_consensus_ux.md
```diff
@@ -0,0 +1,19 @@
+### Changed
+
+- Consensus merge function now preserves all shapes with their scores, regardless of quorum threshold
+  (<https://github.com/cvat-ai/cvat/pull/10172>)
+- In review mode, all annotations are locked by default; Users can unlock and edit individual annotations as needed
+  (<https://github.com/cvat-ai/cvat/pull/10172>)
```
bsekachev: Also would move to Added. The core idea should be highlighted here is that review mode now supports editing objects, however the objects are locked by default

cvat-core/src/server-response-types.ts
```diff
@@ -423,8 +422,9 @@ export interface SerializedShape {
     group: number;
     frame: number;
     source: Source;
+    score?: number;
     attributes: { spec_id: number; value: string }[];
-    elements: Omit<SerializedShape, 'elements'>[];
+    elements: Omit<SerializedShape, 'elements' | 'score'>[];
```
bsekachev: Why individual skeleton points may not have consensus score?

@klakhov @zhiltsov-max

bsekachev (commented):
We either need tooltip here, or extra spans clarifying what these values mean. 
Also, I would add top margin 8px. 

image

cvat-canvas/src/typescript/canvasView.ts
```diff
@@ -3138,10 +3138,14 @@ export class CanvasViewImpl implements CanvasView, Listener {
         const withSource = content.includes('source');
         const withDescriptions = content.includes('descriptions');
         const withDimensions = content.includes('dimensions');
+
         const textFontSize = this.configuration.textFontSize || 12;
         const {
-            label, clientID, attributes, source, descriptions,
+            label, clientID, attributes, source, descriptions, score, votes,
         } = state;
+        const isConsensus = source === 'consensus';
```
bsekachev, line 3146: Let's display always for all objects with type `Consensus`

cvat-core/src/annotations-filter.ts
```diff
@@ -166,6 +170,8 @@ export default class AnnotationsFilter {
                     shape: shape.type,
                     occluded: shape.occluded,
                     objectID: shape.clientID ?? null,
+                    score: null,
```
bsekachev: Why hardcoded null here?
At least `shape` may have `score` as I see

cvat-core/src/server-response-types.ts
zhiltsov-max: With the current merging algorithm, one of the input skeletons is chosen, so there's no consensus scoring between individual points. However, it works not great and I'd like to change it at some point later. I'd vote to add the field to be uniform. but it won't be used right now.

cvat-canvas/src/typescript/canvasView.ts
zhiltsov-max: Need to think how to handle this after golden task changes, we might lose consensus job type information.

cvat-ui/src/actions/annotation-actions.ts
```diff
@@ -739,6 +754,18 @@ export function changeFrameAsync(
...
             } = await fetchAnnotations(toFrame);
+
+            // Auto-lock objects in review mode (except ones user explicitly unlocked)
+            if (state.annotation.workspace === Workspace.REVIEW) {
+                // Clear user unlocks when changing frame since we're showing different objects
+                userUnlockedInReviewMode.clear();
+                states.forEach((objectState: ObjectState) => {
+                    if (!objectState.lock && !objectState.isGroundTruth) {
+                        objectState.lock = true;
+                    }
```
bsekachev: Would not all this `locking` logic be simple if we use Proxy? 

Review? Return Proxy with `locked` by default. 
Unlocked? Okay, remember it somewhere (e.g. in private object of the same module).
Switched mode to standard? Drop proxy and use default logic

cvat-ui/src/components/annotation-page/review-workspace/review-workspace.tsx
```diff
@@ -13,15 +13,19 @@ import ObjectSideBarComponent from 'components/annotation-page/standard-workspac
...
 
 export default function ReviewWorkspaceComponent(): JSX.Element {
     return (
         <Layout hasSider className='cvat-review-workspace'>
             <ControlsSideBarContainer />
             <CanvasLayout />
-            <ObjectSideBarComponent objectsList={<ObjectsListContainer readonly />} />
-            <CanvasContextMenuContainer readonly />
+            <BrushTools />
+            <ObjectSideBarComponent objectsList={<ObjectsListContainer />} />
```
bsekachev, line 25: Do we use `readonly` prop somewhere else?
As I see, we do not. So, probably it may be removed.

cvat-ui/src/components/annotation-page/standard-workspace/objects-side-bar/object-item-details.tsx
```diff
@@ -49,16 +55,49 @@ function attrAreTheSame(prevProps: Props, nextProps: Props): boolean {
...
+function ItemAttributesComponent(props: Props): JSX.Element | null {
     const {
         collapsed, attributes, values, readonly, changeAttribute, collapse,
-        sizeParams, changeSize,
+        sizeParams, changeSize, source, score, votes, textContent,
     } = props;
 
+    const isConsensus = source === Source.CONSENSUS;
+    const withScore = isConsensus && textContent.includes('score');
+    const withVotes = isConsensus && textContent.includes('votes');
```
bsekachev: Content of sidebar items should not depend on settings. Settings only affect what we show on canvas.
In sidebar all details are always visible

cvat-canvas/src/typescript/canvasView.ts
klakhov: Well, the simpliest approach is to always enable it by default

cvat-ui/src/components/annotation-page/review-workspace/review-workspace.tsx
klakhov: Removed

cvat-ui/src/actions/annotation-actions.ts
klakhov: Implemented approach with proxy

cvat-core/src/server-response-types.ts
klakhov: Added score to elements (not shown by UI right now)

cvat-canvas/src/typescript/canvasView.ts
klakhov: Now all objects with source type of 'consensus' will show those values. We do not change source if modified.

cvat-ui/src/containers/annotation-page/standard-workspace/objects-side-bar/objects-list.tsx
zhiltsov-max, line 1: I'm facing the following bug:

https://github.com/user-attachments/assets/46ac8716-7cae-4ee8-a4b9-d057c2535d64

when hovering the cursor over the votes number. Apparently, when there's no scroll bar for objects.

cvat/apps/engine/serializers.py
```diff
@@ -3262,6 +3262,7 @@ def bad_num_points_unless(condition: bool) -> None:
 
 class SubLabeledShapeSerializer(ShapeSerializer, AnnotationSerializer):
     attributes = AttributeValSerializer(many=True, default=[])
+    score = serializers.FloatField(min_value=0, max_value=1, default=1)
```
```suggestion
    score = serializers.FloatField(min_value=0, max_value=1, default=1, read_only=True)
```

cvat-canvas/src/typescript/canvasModel.ts
```diff
@@ -955,7 +955,7 @@ export class CanvasModelImpl extends MasterImpl implements CanvasModel {
 
         if (typeof configuration.textContent === 'string') {
             const splitted = configuration.textContent.split(',').filter((entry: string) => !!entry);
-            if (splitted.every((entry: string) => ['id', 'label', 'attributes', 'source', 'descriptions', 'dimensions'].includes(entry))) {
+            if (splitted.every((entry: string) => ['id', 'label', 'attributes', 'source', 'descriptions', 'dimensions', 'score', 'votes'].includes(entry))) {
```
bsekachev: 'score', 'votes' are unused now

cvat-ui/src/actions/annotation-actions.ts
```diff
@@ -310,17 +310,49 @@ async function fetchAnnotations(predefinedFrame?: number): Promise<{
...
+                if (userUnlockedInReviewMode.has(target.clientID as number)) {
+                    return Reflect.get(target, prop);
+                }
+                return target.isGroundTruth ? Reflect.get(target, prop) : true;
+            }
+            return Reflect.get(target, prop);
+        },
+        set(target, prop, value) {
+            if (prop === 'lock') {
+                if (value === false && !target.isGroundTruth) {
```
bsekachev:
```suggestion
                if (!value && !target.isGroundTruth) {
```

cvat-ui/src/actions/annotation-actions.ts
```diff
@@ -310,17 +310,49 @@ async function fetchAnnotations(predefinedFrame?: number): Promise<{
...
+                    userUnlockedInReviewMode.add(target.clientID as number);
+                }
+                if (value === true) {
```
bsekachev: else if (value)

cvat-ui/src/actions/annotation-actions.ts
```diff
@@ -1171,6 +1212,22 @@ export function updateAnnotationsAsync(statesToUpdate: any[]): ThunkAction {
...
+        const { workspace: currentWorkspace } = state.annotation;
+
+        if (currentWorkspace === Workspace.REVIEW && workspace !== Workspace.REVIEW) {
+            userUnlockedInReviewMode.clear();
+        }
+
+        dispatch(changeWorkspace(workspace));
+
+        // Re-fetch annotations to apply or remove the Proxy wrapper
+        dispatch(fetchAnnotationsAsync());
```
bsekachev: We need it only when previous or next workspace is `REVIEW`, right?

changelog.d/20260114_132218_klakhov_consensus_ux.md
```diff
@@ -0,0 +1,20 @@
+### Changed
+
+- Consensus merge function now preserves all shapes with their scores, regardless of quorum threshold
+  (<https://github.com/cvat-ai/cvat/pull/10172>)
+
+### Added
+
+- Score visualization in UI with a virtual "Votes" attribute calculated as `score × replica_jobs`
```
zhiltsov-max, line 8: This formula can stop working if we allow job creation after merging, will need to understand how to handle it.

cvat-ui/src/actions/annotation-actions.ts
klakhov: Yes, added condition

changelog.d/20260114_132218_klakhov_consensus_ux.md
klakhov: Yes, that may be an issue. But for now, as we discussed, we’ll explicitly mention in the documentation that this value is UI-only and approximate.

PR #10183, opened 2026-01-13
[DEV-786] updated opa to latest version

yaizkazani:

### License

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -321,7 +321,7 @@ services:
     command:
       - run
       - --server
-      - --addr=0.0.0.0:8181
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -321,6 +321,7 @@ services:
     command:
       - run
       - --server
+      - --addr=0.0.0.0:8181
```
SpecLad:
```suggestion
      - --addr=:8181
```

It's shorter and avoids relying on IPv4.

yaizkazani: I've updated this to reflect notes in opa docs

Issue #9284, opened 2025-03-28
Is it possible to filter by tag, but not hide the rest of labels?


gstrat88:
Hi!
If i filter using the label i am using for tagging, the rest objects are getting filtered as per the documentation says.
Is it possible to use the filter to only filter the frames containing the tags, and not the objects?
Thanks!

klakhov:
Hi @gstrat88, you can try to setup a filter with label and type=tag:

![Image](https://github.com/user-attachments/assets/c6342747-456d-4ea0-9b99-3d1aff06e0f7)


And then use `right arrow`/`left arrow` to navigate to frames containing the tags:

![Image](https://github.com/user-attachments/assets/8f45d06d-a378-4896-ab30-c637ddcd8b70)

gstrat88:
nice, thanks a lot! I would try to implement a last filter toggle shortcut then to emulate this option, do you happen to know if already exists?

bsekachev:
Good idea, it is not implemented
There is a feature request #6623

davy-blavette:
Among 1,000 images, I'd like to look for those that aren't marked. If the filters don't work for that, I don't see how to find those images? Not possible?

itsakiki:
Agree a filter for Frames would be helpful, or at least improve the Filters to allow logic with Tags

zhiltsov-max:
Related: https://github.com/cvat-ai/cvat/issues/9956

PR #10200, opened 2026-01-19
Fix(filters): enable 'Filter frames only' toggle when filters configured

### LINKED ISSUE #9956: An option to filter frames without hiding annotations in UI

prathamesh-patil-5090:
# Add "Filter frames only" option to annotation filters - Issue(#9956)

## Motivation and context
- Problem: Current annotation filters both (a) choose which frames are considered for navigation and (b) which annotations are displayed on the current frame. When reviewing, users often want to jump only to frames that match a condition but still see all annotations on those frames.
- Goal: Let filters continue to drive frame navigation while optionally showing all annotations on the landed frame.

## What we implemented (short)
- State & action: add `annotation.annotations.filterFramesOnly` and action `SWITCH_FILTER_FRAMES_ONLY` (`switchFilterFramesOnly()`).
- Behaviour change: `fetchAnnotations()` uses `displayFilters = filterFramesOnly ? [] : filters` so display is unfiltered when the flag is ON, navigation still uses the original filters.
- UI: add a checkbox "Filter frames only" to `FiltersModalComponent` (disabled when no filters), plus small footer styling.
- Files touched (high level):
  - `cvat-ui/src/actions/annotation-actions.ts`
  - `cvat-ui/src/reducers/annotation-reducer.ts`
  - `cvat-ui/src/components/annotation-page/top-bar/filters-modal.tsx`
  - `cvat-ui/src/components/annotation-page/styles.scss`
  - `changelog.d/20251204_075011_122prathamesh6002_9956.md`

## How has this been tested (short)
- Manual test on local dev: created a task with a test image and labels (Square/Circle/Rectangle), added annotations, applied a label filter, then toggled the checkbox:
  - Checkbox OFF: only annotations matching filter are shown.
  - Checkbox ON: navigation still jumps to frames matching filter, but all annotations on each landed frame are visible.

## Checklist
- [x] I have created a changelog fragment (`changelog.d/20251204_075011_122prathamesh6002_9956.md`)

## License
- I submit my code changes under the same MIT License used by this project.

Closes #9956


prathamesh-patil-5090:
@zhiltsov-max check this out 

All the fixes are already in place. The implementation is complete with:

✅ Checkbox with proper [onChange](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) handler
✅ Visible when [isModalConfirmable()](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) returns true (when filter terms are configured)
✅ Keyboard shortcut Shift+F with handler
✅ Styles for the checkbox

zhiltsov-max:
1. I can see the checkbox working now, but it doesn't seem to do what it says. When enabled, it simply works normally. When disabled, the filter is ignored for frame navigation, but the annotations stay filtered. Have you tested the changes? As an option, we could add 2 selectors - filter frames and filter annotations.
2. I think, the previous version, where the checkbox was always displayed, was better. In the current version the dialog window gets resized several times during filter configuration, and it's not very obvious that there will be such a button.

prathamesh-patil-5090:
@zhiltsov-max 
can you check now?
I guess I have fixed both the issues!

[Screencast from 2026-02-15 16-38-09.webm](https://github.com/user-attachments/assets/99c7ef72-5f42-4fd3-aa63-b7bf760d5463)

zhiltsov-max:
I can see the 1st issue is still there - filtering is applied only to annotations when the checkbox is unset, which contradicts the checkbox description and also changes the existing behavior. I still think it's better to make 2 checkboxes instead: 
- [ ] filter frames
- [ ] filter annotations

By default, both should be toggled on (which reproduces the existing behavior).

prathamesh-patil-5090:
@zhiltsov-max 

Changes made:

1. Added [filterAnnotations: boolean](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) state (default true) alongside existing [navigationType](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)
2. Added [SET_FILTER_ANNOTATIONS](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) action + reducer case
3. fetchAnnotations() now passes [filters](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) to the SDK only when [filterAnnotations](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is true, otherwise passes []
4. Replaced the single "Filter frames only" checkbox with two independent checkboxes: Filter frames and Filter annotations — both default ON to preserve existing behavior
5. Checkboxes are always visible in the modal (disabled when no filter is configured) to prevent modal resize
6. Checkbox states are preserved when re-submitting a modified filter; only reset to both-ON defaults on first apply or clear


Also here is the video of testing:

[Screencast from 2026-03-08 18-35-27.webm](https://github.com/user-attachments/assets/75a21729-0442-4fb2-9433-85a5f5c921bc)

bsekachev:
Hello, thanks for the contribution, but the concept is not fully correct in CVAT now and this patch does not really makes sence.

Default navigation - next, previous frame
Filter frames only (should be called "Non empty frames") - search for the nearest frame having annotations (also considers current filter if applied)
Empty frames only - opposite of the previous one.

Also, you can always navigate with the second mode using left, right arrows.

HJJ256:
Hello @bsekachev @zhiltsov-max 

I think there may be some utility in this PR since there have been multiple open issues regarding this feature (#9956 #9284)

If I am tagging an object detection dataset and I have frames which are classified using CVAT tags as tag_a and tag_b. I want my annotation team to complete all frames which have tag_a first and then complete all frames which have tag_b. So that I can start training with those annotations. Also, I do not want to build 2 new pipelines to separate tag_a and tag_b into 2 separate cvat datasets and then another pipeline to stitch them back together once both annotation are completed. Since it does not always make sense to split a dataset into pieces and I might have more than 2 tags.

In the current state, the team would not be able to filter the frames by tag_a without making all the other annotations disappear for each frame. Even if they do navigate frames by turning the filter on, they would have to toggle it off to tag and then on again to navigate.

With this PR, the team can turn off filter annotations checkbox and navigate through the frames containing tag_a without toggling the filter on/off

PR #10203, opened 2026-01-20
Refactor bulk assign for reuse in other tests

archibald1418:
### Motivation and context
Create a new `cypress/support/utils.cy.js` file with cy-related sub-steps for re-use in cypress commands.
Move bulk-actions utility functions into shared `cypress/support`
Add a cypress command to bulk-assign all jobs in the task (analogously to `cy.assignJobToUser` but using bulk assignment (#9582)  which is much faster and much more convenient for consensus tests)

It is useful to keep the cy.queries inside js functions where possible because cypress's stack trace won't show cypress commands as separate stack frames. But the functions will be visible in the stack trace, thus preserving the intended modularity in test cases and making debugging easier.

### How has this been tested?
`e2e/features/bulk_actions.js` doesn't crash

### Checklist

### License

### DIFF
--- a/tests/cypress/support/utils.cy.js
+++ b/tests/cypress/support/utils.cy.js
@@ -8,8 +8,8 @@ export function selectAll() {
     cy.get('.cvat-bulk-wrapper').should('exist').and('be.visible');
     cy.contains('Select all').click();
 }
+
 export function getBulkActionsMenu() {
-    selectAll();
     cy.get('.cvat-item-selected').first().within(() => {
         cy.get('.cvat-actions-menu-button').click();
     });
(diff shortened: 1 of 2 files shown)


tests/cypress/support/utils.cy.js
```diff
@@ -0,0 +1,29 @@
...
+//
+// SPDX-License-Identifier: MIT
+
+/// <reference types="cypress" />
+
+export function selectAll() {
+    cy.get('.cvat-bulk-wrapper').should('exist').and('be.visible');
+    cy.contains('Select all').click();
+}
+export function getBulkActionsMenu() {
```
klakhov, line 12:
```suggestion
}

export function getBulkActionsMenu() {
```

tests/cypress/support/utils.cy.js
```diff
@@ -0,0 +1,29 @@
...
+    selectAll();
```
klakhov: Can we split the logic a bit and remove the `selectAll` call from here? It feels odd that `getBulkActionsMenu` not only builds the menu but also performs selection as a side effect.

I think `selectAll` should be called explicitly in the client code *before* invoking `getBulkActionsMenu`, so the function does exactly what its name suggests.

archibald1418: applied

PR #10215, opened 2026-01-23
Fix a typing problem in VideoReader

SpecLad:
### Motivation and context

`iterate_frames` is supposed to return tuples with `str` as the second element. However, what it _actually_ returns is `self._source_path`, which has the type `str | BytesIO`.

As far as I can tell, this doesn't cause any runtime problems. The `BytesIO` case only occurs when `VideoReader` is used to read a chunk, and in all code paths where this happens, the second tuple element is ignored.

Moreover, even when `self._source_path` is a string, it's not used for much. The only place I found where the second tuple element is used is when `ZipCompressedChunkWriter` uses it to add context to exceptions. But this context isn't really needed when the source is a video, and the message used is clearly intended for individual images rather than videos.

So I think it's appropriate to always return `None` instead of the video path. `ZipCompressedChunkWriter` can already handle `None` values.

### Checklist

### License

### DIFF
--- a/cvat/apps/engine/media_extractors.py
+++ b/cvat/apps/engine/media_extractors.py
@@ -258,7 +258,7 @@ class IMediaReader(ABC):
         self._dimension = dimension
 
     @abstractmethod
-    def __iter__(self):
+    def __iter__(self) -> Iterator[tuple[Any, str | None, Any]]:
         pass
 
     @abstractmethod
@@ -614,7 +614,7 @@ class VideoReader(IMediaReader):
         self,
         *,
         frame_filter: bool | Iterable[int] = True,
-    ) -> Iterator[tuple[av.VideoFrame, str, int]]:
+    ) -> Iterator[tuple[av.VideoFrame, None, int]]:
         """
         If provided, frame_filter must be an ordered sequence in the ascending order.
         'True' means using the frames configured in the reader object.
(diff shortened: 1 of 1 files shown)


cvat/apps/engine/media_extractors.py
```diff
@@ -258,7 +258,7 @@ def __init__(
         self._dimension = dimension
 
     @abstractmethod
-    def __iter__(self):
+    def __iter__(self) -> Iterator[tuple[Any, str | None, Any]]:
```
zhiltsov-max, line 261: I'd actually prefer having some type with more explanations about the arguments, because every time I work with these tuples, I spent some time to investigate what they mean. Consider adding a dedicated type with a docstring. Also, given that video and image tuples are different, it might be better to split the declaration for 2 types, otherwise the code using the second member can probably be considered incorrect with the updated annotations.

SpecLad: > I'd actually prefer having some type with more explanations about the arguments, because every time I work with these tuples, I spent some time to investigate what they mean.

Yeah, me too. :face_with_spiral_eyes: 

> Consider adding a dedicated type with a docstring.

I actually have another patch in the works which simplifies this type; do you mind if I postpone this suggestion until that patch?

> Also, given that video and image tuples are different, it might be better to split the declaration for 2 types, otherwise the code using the second member can probably be considered incorrect with the updated annotations.

What do you mean? The base `__iter__` method has to have some type as the output, so how could I split it?

zhiltsov-max: > I actually have another patch in the works which simplifies this type; do you mind if I postpone this suggestion until that patch?

Ok.

> What do you mean? The base __iter__ method has to have some type as the output, so how could I split it?

For example, like this:

```
MediaFrameTuple = VideoFrameTuple | ImageFrameTuple
   
  ...
  def __iter__(self) -> MediaFrameTuple: ...
```

SpecLad: Okay, I'll give it a shot in the followup PR.

PR #10217, opened 2026-01-26
Flattening job list on a task page

Eldies:
### Motivation and context

We are going to increase the possible number of replicas, and current way of displaying replicas is not convenient if there are hundreds of them.

So,
- flatten list of jobs
- on both jobs page and task page:
  - every parent job has a tag "parent"
  - every replica job has a tag "replica"
  - every parent job has an action "go to replicas" which switches filters on a page to show it's replicas
  - every replica job has an action "go to parent" which switches filters on a page to show it's parent
- task page be default does not show replicas
- added "replicas_count" field to api on job retrieving 

depends on:
https://github.com/cvat-ai/cvat/pull/10257

### Checklist

### License


changelog.d/20260126_133853_dmitrii.lavrukhin_flatten_job_list.md
```diff
@@ -0,0 +1,4 @@
+### Changed
+
+- Flattening job list on a task page
```
klakhov: Need to say something more convininient to user. Something like:
Changed the display of parent and replica jobs to a flat list.
For convenience, jobs are now marked with “parent” and “replica” tags.
By default, replica jobs are hidden using a filter, but they can be shown if needed.

cvat-ui/src/components/common/cvat-tag.tsx
```diff
@@ -22,6 +24,10 @@ function CVATTag(props: TagProps): JSX.Element | null {
             return <Tag className='cvat-tag-ground-truth' color='#ED9C00'>Ground truth</Tag>;
         case TagType.CONSENSUS:
             return <Tag className='cvat-tag-consensus' color='#1890FF'>Consensus</Tag>;
+        case TagType.PARENT:
+            return <Tag className='cvat-tag-parent' color='#1890FF'>Parent</Tag>;
+        case TagType.REPLICA:
+            return <Tag className='cvat-tag-replica' color='#58C0FF'>Replica</Tag>;
```
klakhov:
```suggestion
            return <Tag className='cvat-tag-replica' color=#69c0ff'>Replica</Tag>;
```

Lets use color from antd color pallet

cvat-ui/src/components/jobs-page/actions-menu-items.tsx
```diff
@@ -134,6 +138,26 @@ export default function JobActionsItems(
         }, 100]);
     }
 
+    if (onGoToParent || onGoToReplicas) {
+        menuItems.push([{ type: 'divider' }, 199]);
```
klakhov: I dont think we need a divider here, divider usually indicates dangerous actions. We already have go to the task on top.

cvat-ui/src/components/jobs-page/actions-menu-items.tsx
```diff
@@ -134,6 +138,26 @@ export default function JobActionsItems(
...
+            disabled: isDisabled('go-to-parent'),
+        }, 200]);
+    }
+    if (onGoToReplicas) {
+        menuItems.push([{
+            key: 'go-to-replicas',
+            onClick: onGoToReplicas,
+            label: withCount('Go to replicas', 'go-to-replicas'),
+            disabled: isDisabled('go-to-replicas'),
+        }, 201]);
```
klakhov:
```suggestion
        }, 210]);
```

The step should be larger for plugins

cvat-ui/src/components/jobs-page/actions-menu.tsx
```diff
@@ -126,6 +126,18 @@ function JobActionsComponent(
...
+        if (jobInstance.parentJobId && onApplyFilter) {
+            onApplyFilter(`{"and":[{"==":[{"var":"id"},${jobInstance.parentJobId}]}]}`);
+        }
+    }, [jobInstance.parentJobId, onApplyFilter]);
+
+    const onGoToReplicas = useCallback(() => {
+        if (onApplyFilter) {
+            onApplyFilter(`{"and":[{"==":[{"var":"parent_job_id"},${jobInstance.id}]}]}`);
+        }
+    }, [jobInstance.id, onApplyFilter]);
```
klakhov: I belive we can make bulk actions work with those buttons instead of just disabling it. We can try create combined filter to show all parents/replicas if several jobs are selected

cvat-ui/src/components/jobs-page/jobs-filter-configuration.ts
```diff
@@ -109,6 +109,13 @@ export const config: Partial<Config> = {
                 ],
             },
         },
+        parent_job_id: {
+            label: 'parentID',
```
klakhov:
```suggestion
            label: 'Parent ID',
```

cvat-core/src/session.ts
```diff
@@ -534,7 +534,7 @@ export class Job extends Session {
         source_storage: Storage,
         target_storage: Storage,
         parent_job_id: number | null;
-        consensus_replicas: number;
+        has_replicas: number;
```
klakhov: I see a potential problem with newly added consensus UX patch. We calculate consensus votes based on amount of job replicas by `score x replicas`, but here we are losing this information.

cvat-ui/src/components/resource-sorting-filtering/hooks.ts
klakhov, line 1: What’s the benefit of introducing a new hook here?
Now we have `useResourceFilteringQuery` and `useResourceQuery`, which essentially do the same thing, just with different parameters. Its better to just update `useResourceQuery`

cvat-ui/src/components/task-page/job-list.tsx
```diff
@@ -61,6 +58,7 @@ function filterJobs(jobs: Job[], query: JobsQuery): Job[] {
             updatedDate: job.updatedDate,
             type: job.type,
             id: job.id,
+            parent_job_id: job.parentJobId,
```
klakhov, line 61:
```suggestion
            parentJobId: job.parentJobId,
```
Lets use camel case here

tests/cypress/e2e/features2/consensus.js
```diff
@@ -55,30 +55,43 @@ context('Basic manipulations with consensus job replicas', () => {
...
         });
 
-        it('Check new consensus task has correct tags and drop-down with replicas', () => {
+        it('Check new consensus task has correct tags and replicas', () => {
             cy.goToTaskList();
             cy.openTask(taskName);
             cy.get('.cvat-task-details-wrapper').should('be.visible');
             cy.get('.ant-notification-notice-error').should('not.exist');
-            // Check tags
+            // Check task tag
```
klakhov: We dont need much comments here, I think the one comment with note that by default replicas are hidden is enough

cvat/apps/engine/serializers.py
```diff
@@ -751,13 +751,23 @@ def to_representation(self, data):
                 ).values_list("id", "issue__count")
             )
 
+            # Fetching it here removes 1 extra join for all jobs in the COUNT(*) request,
+            # limiting in only for the page
```
```suggestion
            # limiting it only for the page
```

cvat/apps/engine/serializers.py
```diff
@@ -751,13 +751,23 @@ def to_representation(self, data):
...
+            children_counts = dict(
+                models.Job.objects.with_child_jobs_counts().filter(
+                    id__in=set(j.id for j in page)
```
zhiltsov-max: Could you extract this `set` into a variable? It gets collected twice in a row.

cvat/apps/events/handlers.py
```diff
@@ -275,7 +275,7 @@ def _get_object_name(instance):
     (Organization, OrganizationReadSerializer),
     (Project, ProjectReadSerializer),
     (Task, TaskReadSerializer),
-    (Job, JobReadSerializer),
+    (Job, JobReadForEventSerializer),
```
zhiltsov-max: Why was this change needed in this PR? Can it be extracted into a separate one?

Eldies: extracted it to a separate PR: https://github.com/cvat-ai/cvat/pull/10257

changelog.d/20260126_133853_dmitrii.lavrukhin_flatten_job_list.md
Eldies: copy pasted your suggestion in here

cvat-ui/src/components/task-page/job-list.tsx
Eldies: On jobs-page, the filtering is done on the backend side, which needs this name in snake case. 
I do not want onGoToReplicas action to depend on the page it is used on, therefore parent_job_id should be in snake case on both jobs-page and here. 
I do not see more convenient way to have new actions on both pages

cvat/apps/engine/serializers.py
Eldies: done

tests/cypress/e2e/features2/consensus.js
Eldies: left only one comment

cvat-ui/src/components/jobs-page/actions-menu-items.tsx
Eldies: removed it

cvat-ui/src/components/jobs-page/actions-menu.tsx
Eldies: implemented bulk actions with new buttons

cvat-core/src/session.ts
Eldies: dropped has_replicas field, added replicas_count field

cvat-ui/src/components/resource-sorting-filtering/hooks.ts
Eldies: updated `useResourceQuery`

cvat-core/src/server-response-types.ts
```diff
@@ -163,7 +163,7 @@ export interface SerializedJob {
     source_storage: SerializedStorage | null;
     target_storage: SerializedStorage | null;
     parent_job_id: number | null;
-    consensus_replicas: number;
+    replicasCount: number;
```
klakhov:
```suggestion
    replicas_count: number;
```
its snake case from server I suppose

cvat-ui/src/components/jobs-page/actions-menu-items.tsx
```diff
@@ -41,10 +43,15 @@ export default function JobActionsItems(
...
         selectedIds = [],
+        onGoToParent,
+        onGoToReplicas,
     } = menuItemsData;
 
     const isBulkMode = selectedIds.length > 1;
-    const bulkAllowedKeys = ['edit_assignee', 'edit_state', 'edit_stage', 'export_job', 'delete'];
+    const bulkAllowedKeys = [
+        'edit_assignee', 'edit_state', 'edit_stage', 'export_job', 'delete',
+        'go-to-parent', 'go-to-replicas',
```
klakhov:
```suggestion
        'go_to_parent', 'go_to_replicas',
```

cvat-ui/src/components/jobs-page/actions-menu-items.tsx
```diff
@@ -134,6 +141,23 @@ export default function JobActionsItems(
...
+        }, 200]);
+    }
+    if (onGoToReplicas) {
+        menuItems.push([{
+            key: 'go-to-replicas',
+            onClick: onGoToReplicas,
+            label: withCount('Go to replicas', 'go-to-replicas'),
+            disabled: isDisabled('go-to-replicas'),
+        }, 210]);
+    }
```
klakhov: I would suggest to move this under 'Go to the task' button to keep the menu sections aligned

cvat-ui/src/components/jobs-page/actions-menu.tsx
```diff
@@ -126,6 +126,30 @@ function JobActionsComponent(
         });
     }, [jobInstance, allJobs, selectedIds, dispatch]);
 
+    const onGoToParent = useCallback(() => {
```
klakhov, line 133: I see a couple of issues here:
1. After I used 'Go to replicas' the old parent jobs are still selected, need to clear selection after applying filter. We can use `selectionActions.clearSelectedResources()`
Image

2. When I select both parent and replica I can see `Go to parent(2)` or `Go to replica(2)`. But I should see two buttons `Go to parent` `Go to replica`

Image

You can refer to [members menu](https://github.com/cvat-ai/cvat/blob/develop/cvat-ui/src/components/organization-page/actions-menu.tsx) or/and [requests menu](https://github.com/cvat-ai/cvat/blob/develop/cvat-ui/src/components/requests-page/actions-menu.tsx) for implementation reference.

cvat-ui/src/components/task-page/job-list.tsx
klakhov: Ok. I see

cvat-ui/src/components/jobs-page/actions-menu-items.tsx
Eldies: done

cvat-ui/src/components/jobs-page/actions-menu.tsx
Eldies: fixed

cvat-ui/src/components/jobs-page/actions-menu-items.tsx
```diff
@@ -40,13 +43,32 @@ export default function JobActionsItems(
...
+        onGoToParent,
+        onGoToReplicas,
     } = menuItemsData;
 
-    const isBulkMode = selectedIds.length > 1;
-    const bulkAllowedKeys = ['edit_assignee', 'edit_state', 'edit_stage', 'export_job', 'delete'];
+    const isBulkMode = jobsToAct.length > 1;
+    const bulkAllowedKeys = [
+        'edit_assignee', 'edit_state', 'edit_stage', 'export_job', 'delete',
+        'go_to_parent', 'go_to_replicas',
```
klakhov: Can you please create `enum MenuKeys` like in organization actions menu to avoid hardcoding snake case values

cvat-ui/src/components/jobs-page/actions-menu-items.tsx
```diff
@@ -40,13 +43,32 @@ export default function JobActionsItems(
...
+        edit_assignee: jobsToAct,
+        edit_state: jobsToAct,
+        edit_stage: jobsToAct,
+        export_job: jobsToAct,
+        delete: jobsToAct.filter((j) => j.type === JobType.GROUND_TRUTH),
+        go_to_parent: jobsToActWithParents,
+        go_to_replicas: jobsToActWithReplicas,
+    };
+
+    const withCount = LabelWithCountHOF(jobsToAct, bulkAllowedKeys, actionsApplicable);
```
klakhov: Something is wrong with types here, first param should be just an array of ids, not class instances.

cvat-ui/src/components/jobs-page/actions-menu.tsx
```diff
@@ -122,17 +126,38 @@ function JobActionsComponent(
...
+            const parentIds = [...new Set(
+                jobsToAct.map((j) => j?.parentJobId).filter((id) => id != null),
+            )];
+            const logic = JSON.stringify({
+                or: parentIds.map((id) => ({ '==': [{ var: 'id' }, id] })),
+            });
+            onApplyFilter(logic);
+            dispatch(selectionActions.clearSelectedResources());
+        }
+    }, [jobsToAct, dispatch]);
```
klakhov: onApplyFIlter should be in the list of the deps, in other functions too

cvat-ui/src/components/jobs-page/actions-menu-items.tsx
```diff
@@ -56,12 +78,29 @@ export default function JobActionsItems(
         disabled: isDisabled('task'),
     }, 0]);
 
+    if (jobsToActWithParents.length) {
```
klakhov:
```suggestion
    if (jobsToActWithParents.length && onGoToParent) {
```

cvat-ui/src/components/jobs-page/actions-menu-items.tsx
```diff
@@ -56,12 +78,29 @@ export default function JobActionsItems(
...
+        menuItems.push([{
+            key: 'go_to_parent',
+            onClick: onGoToParent,
+            label: withCount('Go to parent', 'go_to_parent'),
+            disabled: isDisabled('go_to_parent'),
+        }, 10]);
+    }
+    if (jobsToActWithReplicas.length) {
```
klakhov:
```suggestion
    if (jobsToActWithReplicas.length && onGoToReplicas) {
```

cvat-ui/src/components/jobs-page/actions-menu-items.tsx
Eldies: done

cvat-ui/src/components/jobs-page/actions-menu-items.tsx
Eldies: fixed

cvat-ui/src/components/jobs-page/actions-menu.tsx
Eldies: fixed

cvat-ui/src/components/jobs-page/actions-menu-items.tsx
```diff
@@ -40,98 +59,135 @@ export default function JobActionsItems(
...
+        MenuKeys.DELETE, MenuKeys.GO_TO_PARENT, MenuKeys.GO_TO_REPLICAS,
+    ];
     const isDisabled = (key: string): boolean => isBulkMode && !bulkAllowedKeys.includes(key);
-    const withCount = LabelWithCountHOF(selectedIds, bulkAllowedKeys);
+
+    const jobsToActWithParents = jobsToAct.filter((j) => j.parentJobId != null);
+    const jobsToActWithReplicas = jobsToAct.filter((j) => j.replicasCount > 0);
+
+    const actionsApplicable = {
+        edit_assignee: jobsToAct,
```
klakhov:
```suggestion
        [MenuKeys.EDIT_ASSIGNEE]: jobsToAct,
```
and so on

klakhov (approved):

cvat/apps/engine/models.py
zhiltsov-max, line 1: Shouldn't the `consensus_replicas` in the task model and serializer be updated?

cvat/apps/events/handlers.py
```diff
@@ -298,7 +298,7 @@ def get_serializer(instance):
 SERIALIZER_CLEAN_UP_FIELDS = [
     (ProjectReadSerializer, ["tasks", "labels"]),
     (TaskReadSerializer, ["jobs", "labels"]),
-    (JobReadSerializer, ["labels", "issues"]),
+    (JobReadSerializer, ["labels", "issues", "replicas_count", "consensus_replicas"]),
```
zhiltsov-max, line 301: The removal of the field should probably be mentioned in the changelog

cvat-ui/src/components/jobs-page/job-card.tsx
zhiltsov-max, line 1: Going to the replica view doesn't add a new entry in the browser history, so going back returns the user to the tasks list or some other page.

zhiltsov-max, line 1: Are you sure about the colors chosen for the Parent/Replica tags? To me, it feels like Replica uses a "visited link" color, relatively to the Parent tag color.

Image

Image

zhiltsov-max, line 1: Doing this on the task page:

1. Clear filters
2. Quick filters -> Not a replica
3. Go to replicas

breaks the last button behavior - it enables the replica display and immediately returns back. 

The user has no other [relatively obvious] way to return to the "original" page look after clicking the "Go to replicas" button, except than using the quick filter.

zhiltsov-max, line 1: Consensus Replicas on the task creation page should probably be renamed now, if this is not related to consensus directly anymore. Maybe, the consensus use case can be mentioned in the pop up help for the field.

The same is also about the "Consensus" task tag and the "Merge consensus job(s)" buttons - they should probably be updated somehow, as now they look quite confusing. I can suggest removing the "Consensus" tag and renaming the buttons to "Merge annotations from replicas" (the merging dialog will have to be updated to mention consensus, as won't be obvious as well).

Image

zhiltsov-max, line 1: On the task page, when the filter is disabled, currently the jobs are ordered by id and this, actually, also works to group the replicas with their parents. This is an occasional implementation detail and it won't work in the future, when new replicas can be created after the task creation. Think if we can support such grouping in this case.

klakhov: Those are from antd [primary colors pallete](https://ant.design/docs/spec/colors). Do you have better suggestions?

klakhov: Do you mean applying filter? I believe it just works this way on all pages and not really relevent to this patch

zhiltsov-max: I only say that the implementation of the "Go to replicas" button is really inconvenient now. The fact that this is implemented via filters and they behave this way everywhere is the implementation detail, though I'd vote to review this logic.

zhiltsov-max: Globally, I can see 2 options: use the same colors for the same UI elements, or use more distinctive colors. The first one is more scalable, but less visually recognizable, so I'd choose changing the color to a different one. Also, if both colors are from the primary color palette in ant, maybe they should not be used to identify different object types? I think that maybe a different hue will work better in this case, like green or cyan, e.g. `#00c7ab` or `#00969c`. Did you use some algorithm to pick the orange for GT jobs? It doesn't seem to be an ant default color.

Eldies: fixed the bug

Eldies: I am going to do it in the next PR. In this PR, replicas are still only related to consensus

Eldies: replica tag is now cyan

cvat-ui/src/components/jobs-page/actions-menu-items.tsx
Eldies: done

cvat/apps/engine/models.py
Eldies: I plan to do it in the next PR

cvat/apps/events/handlers.py
Eldies: added a changelog entry about deprecating consensus_replicas

cvat-ui/src/components/jobs-page/job-card.tsx
zhiltsov-max: I like the new color

Eldies: now, new entry added to browser history on every filter change, back/forward buttons are working.
I've done it for task page and for jobs page. 

If we need it for other pages, I think it should be done in separate PR

cvat-ui/src/components/jobs-page/jobs-page.tsx
klakhov, line 1: As we now changed how filters behave, it should be changed in every list, not just jobs.

cvat-ui/src/components/jobs-page/jobs-page.tsx
```diff
@@ -57,28 +59,46 @@ function JobsPageComponent(): JSX.Element {
     }, []);
 
     useEffect(() => {
+        if (isMounted && !_.isEqual(query, updatedQuery)) {
+            dispatch(getJobsAsync({ ...updatedQuery }));
+        }
+    }, [updatedQuery, query, isMounted]);
```
klakhov, line 65: Im not sure If I understand how this works, when we press go back, what params are changed so this use effect runs?

Eldies: updatedQuery is changed

cvat-ui/src/components/jobs-page/jobs-page.tsx
klakhov: As discussed created an issue about it: https://github.com/cvat-ai/cvat/issues/10292

changelog.d/20260126_133853_dmitrii.lavrukhin_flatten_job_list.md
zhiltsov-max, line 1: Please update the changelog with a note about filter changes.

Eldies: done

PR #10224, opened 2026-01-27
Update IMediaReader subclasses to use `Path`s/`Openable`s

SpecLad:
### Motivation and context

The main purpose of this change is to update `VideoReader` and `VideoReaderWithManifest` to work with arbitrary openables rather than just strings and pre-opened files. This should make it easy to implement support for directly reading files located in cloud storage.

Doing this means we need to update client code to supply Openables (mostly `pathlib.Path`s), which causes a chain reaction of updating other code. So most of this patch is actually updating other code to use `Path` object and methods. 

### Checklist

### License

### DIFF
--- a/cvat/apps/engine/frame_provider.py
+++ b/cvat/apps/engine/frame_provider.py
@@ -34,7 +34,6 @@ from cvat.apps.engine.media_extractors import (
 )
 from cvat.apps.engine.mime_types import mimetypes
 from cvat.apps.engine.utils import take_by
-from utils.dataset_manifest.utils import MemOpenable
 
 _T = TypeVar("_T")
 
@@ -462,9 +461,7 @@ class SegmentFrameProvider(IFrameProvider):
         # disable threading to avoid unpredictable server
         # resource consumption during reading in endpoints
         # can be enabled for other clients
-        models.DataChoice.VIDEO: lambda source: VideoReader(
-            [MemOpenable(source.getvalue())], allow_threading=False
-        ),
+        models.DataChoice.VIDEO: lambda source: VideoReader([source], allow_threading=False),
     }
 
     def __init__(self, db_segment: models.Segment) -> None:
--- a/cvat/apps/engine/media_extractors.py
+++ b/cvat/apps/engine/media_extractors.py
@@ -44,7 +44,7 @@ from cvat.apps.engine.mime_types import mimetypes
 from utils.dataset_manifest import ImageManifestManager, VideoManifestManager
 from utils.dataset_manifest.errors import InvalidPcdError
 from utils.dataset_manifest.utils import MediaDimension as _MediaDimension
-from utils.dataset_manifest.utils import Openable, PcdReader, detect_media_dimension
+from utils.dataset_manifest.utils import MemOpenable, Openable, PcdReader, detect_media_dimension
 
 ORIENTATION_EXIF_TAG = 274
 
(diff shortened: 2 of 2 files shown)


cvat/apps/engine/frame_provider.py
```diff
@@ -461,7 +462,9 @@ class SegmentFrameProvider(IFrameProvider):
         # disable threading to avoid unpredictable server
         # resource consumption during reading in endpoints
         # can be enabled for other clients
-        models.DataChoice.VIDEO: lambda source: VideoReader([source], allow_threading=False),
+        models.DataChoice.VIDEO: lambda source: VideoReader(
+            [MemOpenable(source.getvalue())], allow_threading=False
```
zhiltsov-max: Do you plan to change this call soon? If not, I'd rather keep the `BytesIO` input support in `VideoReader` for clearer client code.

SpecLad: IMO, directly supporting `BytesIO` in `VideoReader` is problematic, because it means that (for example), calling `iterate_frames` multiple times will result in an error, since the stream will be exhausted. I would rather change `read_chunk` to return a `bytes` or `MemOpenable` object, which would simplify the code here.

(As a side note, I think we should prefer `bytes` to `BytesIO` when it comes to passing data around in general. With `bytes`, we don't have to keep track of the file pointer, which means there's less potential for error.) 

That said, I'm probably not going to do this _soon_, since it'll probably cause a cascade of other changes.

zhiltsov-max: How about wrapping the input `BytesIO` object into `MemOpenable`, just on the `VideoReader` side? It can be done the same way as here. There are even no extra memory copying happens on this conversion.

zhiltsov-max: We could try to move towards using https://docs.python.org/3/reference/datamodel.html#python-buffer-protocol.

SpecLad: > How about wrapping the input `BytesIO` object into `MemOpenable`, just on the `VideoReader` side? It can be done the same way as here. There are even no extra memory copying happens on this conversion.

Yeah, I guess I can do that.

> We could try to move towards using https://docs.python.org/3/reference/datamodel.html#python-buffer-protocol.

I don't think the buffer protocol can help us here, since regular files are not buffers.

SpecLad: > Yeah, I guess I can do that.

Alright, I did it.

cvat/apps/engine/media_extractors.py
```diff
@@ -382,19 +382,19 @@ def frame_range(self):
...
-        image_paths = []
-        for source in source_paths:
-            for root, _, files in os.walk(source):
-                paths = [os.path.join(root, f) for f in files]
-                paths = filter(lambda x: get_mime(x) == "image", paths)
-                image_paths.extend(paths)
+        image_paths = [
+            path
+            for source in source_paths
+            for path in source.rglob("*")
```
zhiltsov-max, line 395: This PR seems to add a regression for task creation from images on a share. In the task.py:840 (this PR's numbers):

```python
    update_status("Media files are being extracted...")
    ...
                MEDIA_TYPES['directory']['extractor'](
                    source_paths=[os.path.join(upload_dir, f) for f in media['directory']],
                ).absolute_source_paths
```

`source_paths` contains strings, and they have no `rglob` method, so task creation fails. Could you check it please? I think it makes sense to check the remaining os.path.join / relpath uses and maybe add a test for this use case.

SpecLad: Yeah, sure. It's weird that this case isn't already covered, though. 🤔

SpecLad: Oh, I see. It's directories on a share specifically.

SpecLad: Should be fixed by #10281.

PR #10227, opened 2026-01-27
Refactor url building on server

zhiltsov-max:
### Motivation and context

In the production configuration, Traefik [adds](https://doc.traefik.io/traefik/reference/routing-configuration/http/middlewares/headers/) `X-Forwarded-*` headers, including `X-Forwarded-Host`. The only cases when this header is missing is the local development configuration, where some services (including Traefik) can be missing, and if `USE_X_FORWARDED_HOST` configuration option is disabled, which is a custom change to CVAT. If a TUS uploading via SDK/CLI is performed in such a scenario and the PAT authentication is used, [the Origin header](https://datatracker.ietf.org/doc/rfc6454/), which is optional, is not added to the request, because it's supposed for browsers and as a security measure against XSS and CSRF. Using Origin was supposed to fix the issue that there is no schema in the resulting location, but the host and schema are always added by the `build_absolute_uri()` function. 

The 500 error can only be obtained in non-production configurations, so no tests added.

- Fixed possible server 500 error in TUS uploads, when there's no proxy and no Origin header in the request (e.g. dev environment with direct server calls from CLI/SDK)
- Refactored url building calls on the server

### Checklist

### License

### DIFF
--- a/cvat/apps/engine/mixins.py
+++ b/cvat/apps/engine/mixins.py
@@ -152,7 +152,7 @@ class UploadMixin:
             metadata=metadata, upload_dir=upload_dir, user_id=request.user.id
         )
 
-        location = request.build_absolute_uri(request.path_info)
+        location = request.build_absolute_uri(request.path)
 
         # FUTURE-TODO: migrate to common TMP cache where files
         # are deleted automatically by a periodic background job


cvat/apps/engine/mixins.py
```diff
@@ -152,9 +152,7 @@ def init_tus_upload(self, request: ExtendedRequest):
             metadata=metadata, upload_dir=upload_dir, user_id=request.user.id
         )
 
-        location = request.build_absolute_uri()
-        if 'HTTP_X_FORWARDED_HOST' not in request.META:
-            location = request.META.get('HTTP_ORIGIN') + request.META.get('PATH_INFO')
+        location = request.build_absolute_uri(request.path_info)
```
SpecLad: Isn't the argument unnecessary here? IIUC, `build_absolute_uri` with no argument is supposed to produce the full request URI.

zhiltsov-max: There's a slight difference between `request.path` (used by default in `build_absolute_uri()`) and [`request.path_info`](https://docs.djangoproject.com/en/5.2/ref/request-response/#django.http.HttpRequest.path_info). It's unlikely to be a problem in foreseeable future (unless we change uvicorn to something), but it's better to use `path_info` to never think about this.

SpecLad: But are you sure that `path_info` is the correct attribute to use here? Looking at the implementation of `build_absolute_uri`, it seems to append the passed value to the domain directly (without adding the script prefix), which means `path` should be the correct value to use.

(Also, it seems unlikely that Django developers would use a default value that doesn't work correctly in some configurations.)

zhiltsov-max: > But are you sure that path_info is the correct attribute to use here?

Not really. I also don't know why it was used previously in the removed code. Probably, using `path` is as fine. We just need to clean up query params, if any.

PR #10239, opened 2026-01-30
fix(helm): Deploy server/frontend using RollingUpdate

jackylamhk:
### Motivation and context

Use [`RollingUpdate`](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#rolling-update-deployment) to minimize downtime during deployments; unless there's a specific reason to ensure all pods are terminated before creating new ones, this change will allow the CVAT UI (and API) to remain online while an update is being deployed.

### Checklist

### License

### DIFF
--- a/changelog.d/20260130_152930_jackylamhk_helm_deployment_strategy.md
+++ b/changelog.d/20260130_152930_jackylamhk_helm_deployment_strategy.md
@@ -1,4 +1,6 @@
 ### Changed
 
-- Deploy server updates using RollingUpdate strategy in Helm chart
+- Deploy server/frontend updates using RollingUpdate strategy in Helm chart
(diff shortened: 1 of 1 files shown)


azhavoro:
Hello,

Thank you for your contribution!
Unfortunately, we cannot accept the PR due to an unresolved database migration issue.

Details:

1. You roll out a change that contains database migrations. 
2. A job is started that applies these migrations. 
3. None of the new pods can start because they check that the migrations have been applied and wait for this to happen.
4. The migrations are applied, and after that, the old pods may become inoperable because the database schema may change. We do not guarantee backward compatibility during migrations.
5. As a result, you will formally get the behavior from the RollingUpdate strategy, but CVAT will not work correctly with this strategy.

PR #10247, opened 2026-02-02
Hotfix: correct consensus score is missing

klakhov:
### Motivation and context

It seems adding `readonly` param to score field in SubLabeledShapeSerializer leads to an error as in merging process we use the same serializer to write into parent job. As this field is readonly, the serializer omits the value

Added basic tests for the updated consensus

### Checklist

### License


cvat/apps/dataset_manager/annotation.py
```diff
@@ -68,7 +68,14 @@ def serialize(self):
...
         serializer = LabeledDataSerializer(data=self.data)
         if serializer.is_valid(raise_exception=True):
-            return serializer.data
+            data = serializer.data
+
+            # Explicitly preserve scores from shapes
+            for i, shape in enumerate(self.shapes):
+                if 'score' in shape:
+                    data['shapes'][i]['score'] = shape['score']
+
```
klakhov: Maybe not the best solution to extend serializer data, but I've seen we do something similar in `backup.py`

zhiltsov-max: I think it will be better to remove `read_only` from the field in the serializer, because this serialization logic is used in many places at the moment, namely dataset import and backups, where it also seems to be broken. 

There's an option is to make a new API-specific serializer with this field set to read only. However, I suppose setting it to read only was just a wrong idea, and there are no problems with making this field editable via the API as well, since it can be used for models or other user-defined workflows. Need to make sure that the UI behaves correctly if the consensus results are edited manually.

tests/python/rest_api/test_consensus.py
```diff
@@ -698,68 +686,133 @@ def test_quorum_is_applied(self, admin_user, jobs, labels, consensus_settings, t
...
+                ),
             )
 
             self.merge(job_id=parent_job["id"], user=admin_user)
 
             merged_annotations = json.loads(
                 api_client.jobs_api.retrieve_annotations(parent_job["id"])[1].data
             )
+
+            assert len(merged_annotations["shapes"]) >= 1
```
zhiltsov-max:
```suggestion
            assert len(merged_annotations["shapes"]) == 2
```
?

cvat/apps/dataset_manager/annotation.py
klakhov: Ok. I will remove `read_only` from that field. UI now does not update score if shape edited manually. 
Do you think we should update it to `1.0`?

zhiltsov-max: > Do you think we should update it to 1.0?

No, just as you did. Just make sure that it doesn't have unexpected side-effects in other scenarios and modes.

klakhov: Well, for now we dont support scenarios with score editing, so we will keep it as read-only field on UI side.

PR #10257, opened 2026-02-04
Ignoring job issues when handling update

Eldies:
### Motivation and context

JobReadSerializer has a field `issues`. The field is only filled when jobs are requested in API. 
When update handler for jobs is triggered, the field is empty for old job instance.
Due to that, if the job has X issues, every update to the job, even if it does not involve issues, creates a server event which incorrectly says that number of issues changed from 0 to X.

Ignore the `issues` field in a job serialiser when handling update

Children summary fields contained invalid data and were not supposed to be included in the events. They also triggered incorrect "update" events because the old object typically didn't contain the summary data, while the updated object had such data.

- Children summary fields will now be excluded from the server events, where applicable
- Children summary field changes will not result in creation of "update" events anymore

### How has this been tested?

Unit tests

### Checklist

### License

### DIFF
--- a/changelog.d/20260211_152614_dmitrii.lavrukhin_serializer_for_job_events.md
+++ b/changelog.d/20260211_152614_dmitrii.lavrukhin_serializer_for_job_events.md
@@ -1,4 +1,5 @@
-### Changed
+### Fixed
 
-- Do not create events for some fields on update/create
+- Creation of extra "create" and "update" events on children summary field
+  updates and inclusion of invalid data in the events for these fields
   (<https://github.com/cvat-ai/cvat/pull/10257>)


cvat/apps/engine/serializers.py
```diff
@@ -819,6 +818,19 @@ def to_representation(self, instance):
 
         return data
 
+
+class JobReadSerializer(JobReadForEventSerializer):
```
zhiltsov-max: It seems currently it's designed to be implemented along the lines of `get_serializer_without_url()`, `get_instance_diff()`, and `_cleanup_fields()` in `handlers.py`. I suggest to move the implementation there. If you want to refactor it somehow, it should probably be yet another PR after this one.

Eldies: Now ignoring the field in `handlers.py`.
I added an additional function, analogous to `get_serializer_without_url` because I do not like to ignore fields in `get_instance_diff` because we do not know which model is processed there. Also, it is unintuitive that `get_instance_diff ` ignores some fields

cvat/apps/events/handlers.py
```diff
@@ -345,6 +342,15 @@ def handle_create(scope, instance, **kwargs):
     )
 
 
+def get_serializer_cleaned_up_for_update(instance):
```
zhiltsov-max: I think this function has to be used in `handle_create()` as well.

cvat/apps/events/handlers.py
```diff
@@ -345,6 +342,15 @@ def handle_create(scope, instance, **kwargs):
...
+    serializer = get_serializer_without_url(instance=instance)
+    if serializer:
+        serializer.fields.pop("labels", None)
+        if isinstance(instance, Job):
+            serializer.fields.pop("issues", None)
```
zhiltsov-max: Could you add the other `*Summary` fields for comments, jobs, and tasks as well? I could reproduce the problem only for task jobs and job issues, but all these fields probably should not be included.

cvat/apps/events/handlers.py
zhiltsov-max, line 1: Please add a regression test for this change and update the changelog.

Eldies: done

cvat/apps/events/handlers.py
Eldies: using it there now. Also, merged it with `get_serializer_without_url`

cvat/apps/events/handlers.py
Eldies: added all the summary fields

changelog.d/20260211_152614_dmitrii.lavrukhin_serializer_for_job_events.md
```diff
@@ -0,0 +1,4 @@
+### Changed
```
zhiltsov-max:
```suggestion
### Fixed
```

changelog.d/20260211_152614_dmitrii.lavrukhin_serializer_for_job_events.md
```diff
@@ -0,0 +1,4 @@
...
+
+- Do not create events for some fields on update/create
+  (<https://github.com/cvat-ai/cvat/pull/10257>)
```
zhiltsov-max:
```suggestion
- Creation of extra "create" and "update" events on children summary field
  updates and inclusion of invalid data in the events for these fields
  (<https://github.com/cvat-ai/cvat/pull/10257>)
```

PR #10266, opened 2026-02-09
Fix the location URL returned by `POST /api/tasks/<id>/data`

### LINKED ISSUE #10115: API should normalize URLs or document trailing slash requirement for TUS uploads

### Description
When initiating a TUS upload via POST request to `/api/tasks/{task_id}/data` (without trailing slash), the server returns a `Location` header that omits the data segment in the path. This causes subsequent PATCH/HEAD requests to fail with 404 errors because the URL points to an incorrect endpoint.

### Impact
- Client code that doesn't ensure trailing slashes will receive invalid URLs
- Subsequent TUS operations (PATCH for chunks, HEAD for offset verification) fail with 404
- Inconsistent behavior compared to CVAT SDK, which always adds trailing slash

### Steps to Reproduce
See the test case in test below:

```python
# This works correctly
response = client.post(
    f"/api/tasks/{task_id}/data/",  # with slash
    headers={"Upload-Length": "1000", "Upload-Metadata": metadata, "Tus-Resumable": "1.0.0"}
)
# Location: /api/tasks/{task_id}/data/1_abc123... ✓

# This returns wrong location
response = client.post(
    f"/api/tasks/{task_id}/data",  # no slash
    headers={"Upload-Length": "1000", "Upload-Metadata": metadata, "Tus-Resumable": "1.0.0"}
)
# Location: /api/tasks/{task_id}/1_abc123... ✗ (missing /data/)
```

I wrote a test for playback.
```python
#

import base64
import io
from http import HTTPStatus

import pytest

from shared.utils.config import BASE_URL, make_api_client
from shared.utils.helpers import generate_image_file


@pytest.mark.usefixtures("restore_db_per_function")
@pytest.mark.usefixtures("restore_cvat_data_per_function")
@pytest.mark.usefixtures("restore_redis_ondisk_per_function")
@pytest.mark.usefixtures("restore_redis_inmem_per_function")
class TestTUSUpload:
    """Integration tests for TUS resumable upload protocol"""

    _USERNAME = "admin1"

    def _create_task(self, task_spec):
        """Helper to create a task"""
        with make_api_client(self._USERNAME) as api_client:
            (task, response) = api_client.tasks_api.create(task_spec)
            assert response.status == HTTPStatus.CREATED
            return task

    def _call_tus_endpoint(self, method, path, headers=None, body=None, check_status=True):
        """Call a low-level TUS endpoint via ApiClient"""
        headers = headers or {}

        with make_api_client(self._USERNAME) as api_client:
            if path.startswith("http"):
                if isinstance(body, io.BytesIO):
                    body = body.getvalue()
                api_client.update_params_for_auth(headers=headers, queries=[], method=method)
                response = api_client.request(
                    method,
                    path,
                    headers=headers,
                    body=body,
                    _parse_response=False,
                    _check_status=check_status,
                )

        return response

    def test_start_upload(self):
        """Start TUS upload and return location URL"""
        task = self._create_task(
            {
                "name": "test TUS incremental offset",
                "labels": [{"name": "car"}],
            }
        )
        task_id = task.id
        image_file = generate_image_file("test_image.jpg", size=(100, 100))
        metadata = f"filename {base64.b64encode('test_image.jpg'.encode()).decode()}"

        response = self._call_tus_endpoint(
            "POST",
            BASE_URL + f"/api/tasks/{task_id}/data/",
            headers={
                "Upload-Length": 1000,
                "Upload-Metadata": metadata,
                "Tus-Resumable": "1.0.0",
            },
        )
        assert response.status == HTTPStatus.CREATED
        location = response.headers.get("Location")
        assert location is not None
        assert location.startswith(BASE_URL + f"/api/tasks/{task_id}/data/")

    def test_start_upload_wrong(self):
        """Start TUS upload and return location URL"""
        task = self._create_task(
            {
                "name": "test TUS incremental offset",
                "labels": [{"name": "car"}],
            }
        )
        task_id = task.id
        image_file = generate_image_file("test_image.jpg", size=(100, 100))
        metadata = f"filename {base64.b64encode('test_image.jpg'.encode()).decode()}"


        response = self._call_tus_endpoint(
            "POST",
            BASE_URL + f"/api/tasks/{task_id}/data", # No slash
            headers={
                "Upload-Length": 1000,
                "Upload-Metadata": metadata,
                "Tus-Resumable": "1.0.0",
            },
        )

        assert response.status == HTTPStatus.CREATED
        location = response.headers.get("Location") #  /api/tasks/{task_id}/{file_name/id}
        # Expecting /api/tasks/{task_id}/data/ in the location URL
        assert location is not None
        assert location.startswith(BASE_URL + f"/api/tasks/{task_id}")
        assert not location.startswith(BASE_URL + f"/api/tasks/{task_id}/data/") # Expected behaviour, that the test should fail here

```

### Expected Behavior
Regardless of whether the initial POST request is made to:
- `/api/tasks/{task_id}/data/` (with trailing slash), or
- `/api/tasks/{task_id}/data` (without trailing slash)

The `Location` header should consistently return a URL in the format:
```
http://localhost:8080/api/tasks/{task_id}/data/{file_id}
```

### Actual Behavior
- **With trailing slash** (`/api/tasks/{task_id}/data/`): Returns correct location
  ```
  Location: http://localhost:8080/api/tasks/{task_id}/data/1_abc123...
  ```

- **Without trailing slash** (`/api/tasks/{task_id}/data`): Returns incorrect location (missing data)
  ```
  Location: http://localhost:8080/api/tasks/{task_id}/1_abc123...
  ```

### Suggested Fix
The `init_tus_upload` method in `UploadMixin` should normalize the request path or ensure the `Location` header always includes the correct data segment, independent of whether the client provides a trailing slash.
- The CVAT SDK (uploading.py) works around this by always adding a trailing slash: `url.rstrip("/") + "/"`

### Additional Context
- The CVAT SDK (uploading.py) works around this by always adding a trailing slash: `url.rstrip("/") + "/"`
- This issue affects API clients that don't strictly follow this convention
- Related to TUS protocol implementation in Django REST framework views

This is standard behavior:
- URLs without trailing slashes treat the last segment as a "file" that gets replaced
- URLs with trailing slashes treat the path as a "directory" and append to it
It would be good to have a mention of this in the documentation or a safety fix.

### Environment

```Markdown
### Environment
- CVAT version: [current development branch]
- Branch: `develop`
- Component: mixins.py (`UploadMixin.init_tus_upload`)
- Start from 
'Location': urljoin(location, tus_file.file_id.as_str),
```

SpecLad:
### Motivation and context

... and other similar endpoints.

Instead of manually building the URL, use reversing. This works correctly whether the original URL has a trailing slash or not.

Fixes #10115.

### How has this been tested?

Unit test.

### Checklist

### License

### DIFF
--- a/cvat/apps/engine/mixins.py
+++ b/cvat/apps/engine/mixins.py
@@ -90,6 +90,12 @@ class UploadMixin:
         return response
 
     def upload_data(self, request: ExtendedRequest, *, append_url_name: str) -> Response:
+        """
+        Implements the "creation" TUS endpoint.
+        append_url_name must be the URL name of an action in the same viewset that implements
+        the "append chunk" endpoint. This action must accept the same path parameters (if any)
+        as the creation action, plus a "file_id" parameter.
+        """
         tus_request = request.headers.get('Upload-Length', None) is not None or request.method == 'OPTIONS'
         bulk_file_upload = request.headers.get('Upload-Multiple', None) is not None
         start_upload = request.headers.get('Upload-Start', None) is not None


cvat/apps/engine/mixins.py
```diff
@@ -90,7 +89,7 @@ def _tus_response(self, status, data=None, extra_headers=None):
                 response.__setitem__(key, value)
         return response
 
-    def upload_data(self, request: ExtendedRequest):
+    def upload_data(self, request: ExtendedRequest, append_url_name: str) -> Response:
```
zhiltsov-max:
```suggestion
    def upload_data(self, request: ExtendedRequest, *, append_url_name: str) -> Response:
```

Consider making kw-only, the parameter doesn't seem to be natural for the operation.

cvat/apps/engine/mixins.py
```diff
@@ -42,7 +41,7 @@
 class UploadedFileError(ValueError):
     pass
 
-class UploadMixin:
+class UploadMixin(viewsets.ViewSet):
```
zhiltsov-max: Consider using a type annotation for `self` instead of explicitly inheriting and possibly introducing an additional `ViewSet` base in the type hierarchy of the view. It can lead to issues if there is another mixin inheriting from `ViewSet` or the view itself does so.

```python
class UploadMixin:
    def foo(self: "_UploadMixinSelf"):
        pass

if TYPE_CHECKING:
    class _UploadMixinSelf(UploadMixin, viewsets.ViewSet):
        pass
```

cvat/apps/engine/mixins.py
```diff
@@ -175,7 +172,9 @@ def init_tus_upload(self, request: ExtendedRequest):
         return self._tus_response(
             status=status.HTTP_201_CREATED,
             extra_headers={
-                'Location': urljoin(location, tus_file.file_id.as_str),
+                "Location": self.reverse_action(
+                    append_url_name, kwargs={**self.kwargs, "file_id": tus_file.file_id.as_str}
```
zhiltsov-max, line 182: There's no obvious relationship between the current endpoint kwargs and the append endpoint ones. I think we can keep this implementation for now, but in the future it might be better to allow to specify them explicitly in the function args. Consider adding a documentation line about this in the function description to highlight this calling agreement.

cvat/apps/engine/mixins.py
SpecLad: Hmm... I did make a mistake here by using `ViewSet` as the base class (I thought it was already part of the hierarchy), but do you think there would still be a problem if I used `ViewSetMixin` instead? What kind if issues do you have in mind?

zhiltsov-max: Inheriting from `ViewSetMixin` should be fine, unless the parent class methods are overridden in the mixin. But in this case we are really speaking just about typing for `self`. If you check DRF mixins, they don't inherit from anything, and I believe it's done this way for a reason. Ideally, I'd like to avoid investigating MRO between inherited mixins in a view.

cvat/apps/engine/mixins.py
SpecLad: It's basically a weaker version of the requirement that the append endpoint URL must begin with the create endpoint URL.

> Consider adding a documentation line about this in the function description to highlight this calling agreement.

Alright, added.

cvat/apps/engine/mixins.py
SpecLad: Alright, done.

cvat/apps/engine/mixins.py
SpecLad: Okay, I just reverted this part. It's not really material to the bugfix anyway.

PR #10269, opened 2026-02-09
Make `RelatedFile.path` a relative path

SpecLad:
### Motivation and context

Absolute paths don't fit very well here:

* Files can be stored in cloud storage, in which case the absolute path stored in the database doesn't actually exist.

* They make it unnecessarily hard to move storage to a different base directory.

* The code becomes simpler with relative paths.

### How has this been tested?

I manually verified that the migration works correctly in both directions.

### Checklist

### License

### DIFF
--- a/cvat/apps/engine/serializers.py
+++ b/cvat/apps/engine/serializers.py
@@ -3760,13 +3760,6 @@ class CloudStorageContentSerializer(serializers.Serializer):
         help_text="This token is used to continue listing files in the bucket.")
     content = FileInfoSerializer(many=True)
 
-class RelatedFileSerializer(serializers.ModelSerializer):
-
-    class Meta:
-        model = models.RelatedFile
-        fields = '__all__'
-        read_only_fields = ('path',)
-
 
 def _update_related_storages(
     instance: models.Project | models.Task,


cvat/apps/engine/migrations/0097_alter_relatedfile_path.py
```diff
@@ -0,0 +1,43 @@
...
+
+def make_path_relative(apps, schema_editor):
+    RelatedFile = apps.get_model("engine", "RelatedFile")
+
+    # Before we chop off the prefix, let's make sure it's actually there.
+    if files_without_prefix := RelatedFile.objects.exclude(path__startswith=path_prefix)[:10]:
+        raise RuntimeError(
+            "Found RelatedFile objects with unexpected paths. Example IDs: "
+            + " ".join(f.id for f in files_without_prefix)
+        )
```
zhiltsov-max, line 20: Consider adding sorting to make the error messages reproducible.

SpecLad: This queryset is already sorted, because `RelatedFile` has `Meta.ordering` set.

SpecLad:
> Consider removing `RelatedFileSerializer` as well. Apparently, it's not used anywhere.

Yeah, alright. Removed.

cvat/apps/engine/task.py
zhiltsov-max, line 1492: Apparently, the migration misses that if the task was created from images on a share, it could contain the share path in the `upload_dir`. Could you update the migration to handle this?

zhiltsov-max: A possible implementation:

```
def make_path_relative(apps, schema_editor):
    RelatedFile = apps.get_model("engine", "RelatedFile")

    RelatedFile.objects.filter(path__startswith=path_prefix).update(
        path=Substr("path", Length(path_prefix) + 1)
    )

    share_prefix = models.Value("/home/django/share/")
    RelatedFile.objects.filter(path__startswith=share_prefix).update(
        path=Substr("path", Length(share_prefix) + 1)
    )

    # Check if we missed some paths
    if files_without_prefix := RelatedFile.objects.filter(path__startswith="/")[:10]:
        raise RuntimeError(
            "Found RelatedFile objects with unexpected paths. Example IDs: "
            + " ".join(str(f.id) for f in files_without_prefix)
        )
```

or the same with a precheck.

SpecLad: I think I should be able to do this with a `Case(When(data__storage="share", then=...), default=...)` in `path_prefix`. That way, the whole thing can still be done with a single query.

zhiltsov-max: Yes, another option

zhiltsov-max: There is a small problem that in development environments, the instance can contain both tasks created via a bare deployment and a docker deployment - if so, the upload_dir paths are different, e.g. `/home/django` for docker and `/home/<username>/cvat/` for bare metal ones.

SpecLad: > There is a small problem that in development environments, the instance can contain both tasks created via a bare deployment and a docker deployment 

I don't really care about this case. If you have mixed paths like that, there are going to be more issues than the migration.

zhiltsov-max: I don't say we need to really support this, but it will be better to not truncate the path using the `default` case without looking at the contents.

SpecLad: See #10279.

(Turns out Django won't let you use joins in UPDATE queries, so I had to make due with a subquery.)

PR #4150, opened 2022-01-12
Update Documentation about export, import and backup features

TOsmanov:
Related #3852 
Related #3790

### Motivation and context

Preview: 
https://tosmanov.github.io/cvat/docs/getting_started/
https://tosmanov.github.io/cvat/docs/manual/advanced/backup/
https://tosmanov.github.io/cvat/docs/manual/advanced/export-import-datasets/
https://tosmanov.github.io/cvat/docs/manual/advanced/projects/
https://tosmanov.github.io/cvat/docs/manual/basics/tasks-page/

This PR contain few updates:
- updated `getting started` section
- `export-import.md` renamed to `backup.md` and updated
- `downloadin-annotation.md` renamed to `export-import-datasets.md` and updated
- update `projects` section
- update `task-page` section
- delete image029.jpg (сontaining example of xml file), since there are examples in the `XML annotation format` section
- fix the titles of the modal windows (`import-dataset-modal.tsx` file)

### Checklist

- [x] I submit my changes into the `develop` branch
- [ ] I have added a description of my changes into [CHANGELOG](https://github.com/opencv/cvat/blob/develop/CHANGELOG.md) file
- [x] I have updated the [documentation](
  https://github.com/opencv/cvat/blob/develop/README.md#documentation) accordingly
- [ ] I have added tests to cover my changes
- [x] I have linked related issues ([read github docs](
  https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword))
- [ ] I have increased versions of npm packages if it is necessary ([cvat-canvas](https://github.com/opencv/cvat/tree/develop/cvat-canvas#versioning),
[cvat-core](https://github.com/opencv/cvat/tree/develop/cvat-core#versioning), [cvat-data](https://github.com/opencv/cvat/tree/develop/cvat-data#versioning) and [cvat-ui](https://github.com/opencv/cvat/tree/develop/cvat-ui#versioning))

### License

- [x] I submit _my code changes_ under the same [MIT License](
  https://github.com/opencv/cvat/blob/develop/LICENSE) that covers the project.
  Feel free to contact the maintainers if that's a concern.
- [x] I have updated the license header for each file (see an example below)

```python
# Copyright (C) 2021 Intel Corporation
#
# SPDX-License-Identifier: MIT
```

PR #10272, opened 2026-02-12
[DEV-795] Docker compose for SAM2 agent deployment

yaizkazani:
1. Scripts to connect to CVAT API + run cvat-cli + run agent
2. Dockerfile
3. Docker compose file + .env

- [x] I have updated the documentation accordingly (will do later after Helm)

### License


ai-models/agents_deployment/sam2/function_registration.sh
```diff
@@ -0,0 +1,128 @@
...
+USERNAME=$(curl "$CVAT_BASE_URL"/api/users/self -s -H "Authorization: Bearer $CVAT_ACCESS_TOKEN" | jq -r '.username')
+if [ -z "$USERNAME" ] || [ "$USERNAME" = "null" ]; then
+    echo "Error: Unable to retrieve username from CVAT API. Check your access token."
+    exit 1
+else
+    echo "Authenticated as user: $USERNAME"
+fi
+
+# Try to find existing function with the same name and owner
+if request_function_id=$(curl ""$CVAT_BASE_URL"/api/functions?filter=%7B%0A%20%20%22and%22%3A%20%5B%0A%7B%22%3D%3D%22%3A%20%5B%7B%22var%22%3A%20%22name%22%7D%2C%20%22"$FUNCTION_NAME"%22%5D%7D%2C%0A%7B%22%3D%3D%22%3A%5B%7B%22var%22%3A%20%22owner%22%7D%2C%22"$USERNAME"%22%5D%7D%0A%20%20%5D%0A%7D" -s -H "Authorization: Bearer "$CVAT_ACCESS_TOKEN"" | jq -r '.results[0].id') && [ "$request_function_id" != "null" ]; then
```
breznak: hello, i was trying to test and replicate your PR on OSS self-hosted CVAT. (I wanted self-hosted SAM, and when I found this PR I wanted to test SAM2.) 

I have rebuild bot cvat and sam2/ from this PR branch. All seems to work but i run into problem, that the endpoint `/api/functions` is not available, only the (former?) /api/lambda/functions (as used by nuclio - SAM1). 

Am I missing something? Any plans to release the API as OSS? 
Thanks for your contribution, cheers,

ai-models/agents_deployment/sam2/docker-compose.yaml
```diff
@@ -0,0 +1,42 @@
...
+    restart: "on-failure"
+    environment:
+      - CVAT_BASE_URL=${CVAT_BASE_URL}
+      - CVAT_ACCESS_TOKEN=${CVAT_ACCESS_TOKEN}
+      - MODEL_ID=${MODEL_ID}
+
+  cvat-agents:
+    image: ${IMAGE_URL}
+    volumes:
+      - shared-data:/shared
```
azhavoro:
```suggestion
      - shared-data:/shared:ro
```

ai-models/agents_deployment/sam2/docker-compose.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+  #NB! MANUAL RUN!
+  # docker compose run --rm cvat-function-deregister
+  # or
+  # docker compose --profile manual up cvat-function-deregister
+  cvat-function-deregister:
+    image: ${IMAGE_URL}
+    entrypoint: ["/app/function_deregistration.sh"]
+    volumes:
+      - shared-data:/shared
```
azhavoro:
```suggestion
      - shared-data:/shared:ro
```

ai-models/agents_deployment/sam2/Dockerfile
azhavoro, line 1: Please specify a non-root user; this may be a critical requirement.

ai-models/agents_deployment/sam2/function_registration.sh
azhavoro: Hello, and thank you for your interest in CVAT. Yes, this feature is only available on app.cvat.ai and with an enterprise subscription.

ai-models/agents_deployment/sam2/entrypoint.sh
```diff
@@ -0,0 +1,41 @@
...
+#FUNCTION_FILE_PATH="../../tracker/sam2/func.py"
+## Docker container run
+FUNCTION_FILE_PATH="func.py"
+
+echo -e "Running SAM2 function agent for FUNCTION_ID: $FUNCTION_ID with MODEL_ID: $MODEL_ID\n..."
+## Local run from cvat/ai-models/agents_deployment/sam2
+#cvat-cli --server-host "$CVAT_BASE_URL" function run-agent "$FUNCTION_ID" --function-file="$FUNCTION_FILE_PATH" -p model_id=str:"$MODEL_ID"
+
+#Docker container run
+cvat-cli --server-host "$CVAT_BASE_URL" function run-agent "$FUNCTION_ID" --function-file="$FUNCTION_FILE_PATH" -p model_id=str:"$MODEL_ID"
```
azhavoro:
```suggestion
exec cvat-cli --server-host "$CVAT_BASE_URL" function run-agent "$FUNCTION_ID" --function-file="$FUNCTION_FILE_PATH" -p model_id=str:"$MODEL_ID"
```

ai-models/agents_deployment/sam2/docker-compose.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+  cvat-agents:
+    image: ${IMAGE_URL}
+    volumes:
+      - shared-data:/shared
+    depends_on:
+      cvat-function-register:
+        condition: service_completed_successfully
+    deploy:
+      replicas: ${AGENTS_COUNT}
```
azhavoro:
```suggestion
      replicas: ${AGENTS_COUNT:-1}
```

ai-models/agents_deployment/sam2/function_registration.sh
```diff
@@ -0,0 +1,128 @@
...
+    mkdir /shared
+fi
+
+# Hardcoded for SAM2
+
+FUNCTION_NAME="SAM2"
+## Local run from cvat/ai-models/agents_deployment/sam2
+#FUNCTION_FILE_PATH="../../tracker/sam2/func.py"
+## Docker container run
+FUNCTION_FILE_PATH="func.py"
```
azhavoro, line 13: Shouldn't this be moved to .env?

ai-models/agents_deployment/sam2/function_registration.sh
azhavoro:
```suggestion
if request_function_id=$(curl --get --data-urlencode "filter={\"and\":[{\"==\":[{\"var\":\"name\"},\"${FUNCTION_NAME}\"]},{\"==\":[{\"var\":\"owner\"},\"$USERNAME\"]}]}" "$CVAT_BASE_URL"/api/functions -s -H "Authorization: Bearer "$CVAT_ACCESS_TOKEN"" | jq -r '.results[0].id') && [ "$request_function_id" != "null" ]; then
```
It should work, but of course, it needs to be checked.

ai-models/agents_deployment/sam2/function_registration.sh
```diff
@@ -0,0 +1,128 @@
...
+if [ -z "$USERNAME" ] || [ "$USERNAME" = "null" ]; then
+    echo "Error: Unable to retrieve username from CVAT API. Check your access token."
+    exit 1
+else
+    echo "Authenticated as user: $USERNAME"
+fi
+
+# Try to find existing function with the same name and owner
+if request_function_id=$(curl ""$CVAT_BASE_URL"/api/functions?filter=%7B%0A%20%20%22and%22%3A%20%5B%0A%7B%22%3D%3D%22%3A%20%5B%7B%22var%22%3A%20%22name%22%7D%2C%20%22"$FUNCTION_NAME"%22%5D%7D%2C%0A%7B%22%3D%3D%22%3A%5B%7B%22var%22%3A%20%22owner%22%7D%2C%22"$USERNAME"%22%5D%7D%0A%20%20%5D%0A%7D" -s -H "Authorization: Bearer "$CVAT_ACCESS_TOKEN"" | jq -r '.results[0].id') && [ "$request_function_id" != "null" ]; then
+    export FUNCTION_ID="$request_function_id"
```
azhavoro: Is it really necessary to export `FUNCTION_ID`?

ai-models/agents_deployment/sam2/function_registration.sh
```diff
@@ -0,0 +1,128 @@
...
+EOF
+)
+
+# Create ConfigMap
+response=$(curl -s -w "\n%{http_code}" --cacert ${CACERT} \
+     --header "Authorization: Bearer ${TOKEN}" \
+     --header "Content-Type: application/json" \
+     -X POST \
+     -d "${CONFIGMAP_DATA}" \
+     "${APISERVER}"/api/v1/namespaces/"${NAMESPACE}"/configmaps)
```
azhavoro:
```suggestion
response=$(curl -s -w "\n%{http_code}" --cacert ${CACERT} \
     --header "Authorization: Bearer ${TOKEN}" \
     --json "${CONFIGMAP_DATA}" \
     "${APISERVER}"/api/v1/namespaces/"${NAMESPACE}"/configmaps)
```

ai-models/agents_deployment/sam2/function_registration.sh
yaizkazani: Good day, thanks for your question!

AFAIK, this functionality is available only for SaaS and Enterprise versions of CVAT. I'm 95% sure, though. 

Regarding "Any plans to release the API as OSS" - hard to say, I think not at the moment.

@azhavoro can you confirm?

ai-models/agents_deployment/sam2/docker-compose.yaml
```diff
@@ -0,0 +1,42 @@
...
+  cvat-function-register:
+    image: ${IMAGE_URL}
+    entrypoint: ["/app/function_registration.sh"]
+    volumes:
+      - shared-data:/shared
+    restart: "on-failure"
+    environment:
+      - CVAT_BASE_URL=${CVAT_BASE_URL}
+      - CVAT_ACCESS_TOKEN=${CVAT_ACCESS_TOKEN}
+      - MODEL_ID=${MODEL_ID}
```
azhavoro: I suggest moving away from a specific parameter for sam2 and defining it through a general `--function-parameter` (`-p`) as `MODEL_ID=int:12345`.

```
 cvat-cli function create-native --help
usage: cvat-cli function create-native [-h] (--function-module MODULE | --function-file PATH) [--function-parameter NAME=TYPE:VALUE] name

Create a CVAT function that can be powered by an agent running the given local function.

positional arguments:
  name                  a human-readable name for the function

options:
  -h, --help            show this help message and exit
  --function-module MODULE
                        qualified name of a module to use as the function
  --function-file PATH  path to a Python source file to use as the function
  --function-parameter NAME=TYPE:VALUE, -p NAME=TYPE:VALUE
                        parameter for the function
```

ai-models/agents_deployment/sam2/function_registration.sh
azhavoro: It is also necessary to ensure that this works correctly if FUCNTION_NAME contains spaces (the user may specify a name with spaces, even if the backend does not accept this)

ai-models/agents_deployment/sam2/entrypoint.sh
```diff
@@ -0,0 +1,41 @@
...
+
+# This script registers the SAM2 model function in the CVAT.
+
+if [ -z "$CVAT_ACCESS_TOKEN" ]; then
+    echo "Error: CVAT_ACCESS_TOKEN environment variable must be set."
+    exit 1
+fi
+
+if [ -z "$CVAT_BASE_URL" ]; then
+    echo "Error: CVAT_BASE_URL environment variable is missing, using https://app.cvat.ai as default."
```
azhavoro:
```suggestion
    echo "Warning: CVAT_BASE_URL environment variable is missing, using https://app.cvat.ai as default."
```

ai-models/agents_deployment/sam2/entrypoint.sh
```diff
@@ -0,0 +1,41 @@
...
+    export CVAT_BASE_URL="https://app.cvat.ai"
+fi
+
+if [ -z "$MODEL_ID" ]; then
+    echo "Error: MODEL_ID environment variable not found. Default is facebook/sam2.1-hiera-tiny"
```
azhavoro: I'm not sure there should be specific SAM2 logic for checking arguments here. Not all functions require additional parameters.

ai-models/agents_deployment/sam2/entrypoint.sh
```diff
@@ -0,0 +1,41 @@
...
+# This script registers the SAM2 model function in the CVAT.
+
+if [ -z "$CVAT_ACCESS_TOKEN" ]; then
+    echo "Error: CVAT_ACCESS_TOKEN environment variable must be set."
+    exit 1
+fi
+
+if [ -z "$CVAT_BASE_URL" ]; then
+    echo "Error: CVAT_BASE_URL environment variable is missing, using https://app.cvat.ai as default."
+    export CVAT_BASE_URL="https://app.cvat.ai"
```
azhavoro:
```suggestion
    CVAT_BASE_URL="https://app.cvat.ai"
```
It seems that export is not needed here.

ai-models/agents_deployment/sam2/entrypoint.sh
```diff
@@ -0,0 +1,41 @@
...
+
+if [ -z "$MODEL_ID" ]; then
+    echo "Error: MODEL_ID environment variable not found. Default is facebook/sam2.1-hiera-tiny"
+    export MODEL_ID="facebook/sam2.1-hiera-tiny"
+fi
+
+
+if [ -f /shared/FUNCTION_ID ]; then
+    echo "FUNCTION_ID file found. Reading FUNCTION_ID from /shared/FUNCTION_ID"
+    export FUNCTION_ID="$(cat /shared/FUNCTION_ID)"
```
azhavoro:
```suggestion
    FUNCTION_ID="$(cat /shared/FUNCTION_ID)"
```

ai-models/agents_deployment/sam2/entrypoint.sh
yaizkazani: This is not a universal solution yet, just for SAM2. We expect this code to be used to deploy SAM2 only

ai-models/agents_deployment/sam2/entrypoint.sh
yaizkazani: Good point btw

ai-models/agents_deployment/sam2/function_registration.sh
yaizkazani: The function path is configured in the Dockerfile, so I don't think it should be flexible here. In general, in universal script for private models - yes, but this one is SAM2 only.

ai-models/agents_deployment/sam2/function_registration.sh
yaizkazani: No, not strictly necessary.

ai-models/agents_deployment/sam2/entrypoint.sh
yaizkazani: agree

ai-models/agents_deployment/sam2/function_registration.sh
yaizkazani: Good point!

ai-models/agents_deployment/sam2/docker-compose.yaml
yaizkazani: This will be done in universal script for sure, as I also think that not all models would require this parameter. 

So "yes, but not in this script"

ai-models/agents_deployment/sam2/function_registration.sh
yaizkazani: Btw it works, good point :)

ai-models/agents_deployment/sam2/Dockerfile
yaizkazani: Agreed

ai-models/agents_deployment/sam2/function_registration.sh
breznak: edit - moving my comment to "Comments", not under code review.

breznak:
thank you for explanations. My goal will be to get SAM2/3 running in the OSS version - and I'd like to do it "the right way".

  ```
┌─────────────────┬──────────────────────────────────────┬─────────────────────────────────────┐
  │                 │  Native Functions (/api/functions)   │   Nuclio (/api/lambda/functions)    │
  ├─────────────────┼──────────────────────────────────────┼─────────────────────────────────────┤
  │ Server-side API │ Cloud-only, not in OSS               │ Available in OSS                    │
  ├─────────────────┼──────────────────────────────────────┼─────────────────────────────────────┤
  │ Model runtime   │ Docker containers managed by CVAT    │ Nuclio serverless functions         │
  ├─────────────────┼──────────────────────────────────────┼─────────────────────────────────────┤
  │ Deployment      │ cvat-cli function create             │ nuctl deploy                        │
  ├─────────────────┼──────────────────────────────────────┼─────────────────────────────────────┤
  │ SAM2 support    │ PR #10272 (unusable for self-hosted) │ Not available out-of-box, SAM1 only │
  └─────────────────┴──────────────────────────────────────┴─────────────────────────────────────┘
```
it is understandable SAM2 is now differentiable "paid only feature", but I'd like to know your plans/roadmap regarding the API. is /api/lambda/functions = nuclio still the standard way? Or are you planning to switch to cvat-cli functions = dockerized communication?

From POV of easy integrations (and also architectionally for CVAT) I would strongly prefer if both paid/oss versions used one codepath - cvat-cli functions -- does it align with your plans?
(if not, I will know i should continue with nuclio, but the "docker compose" way of integrating new tools would provide more benefit,integrations to CVAT).
again, thank you for your work! cheers,

referencing our discussion under https://github.com/cvat-ai/cvat/pull/10272#discussion_r2801858666 

cc @yaizkazani @azhavoro

File: ai-models/agents_deployment/sam2/.env [hunk 14]
```diff
@@ -0,0 +1,6 @@
+# Docker compose
+CVAT_BASE_URL=your_cvat_url # or https://app.cvat.ai
+CVAT_ACCESS_TOKEN=your_cvat_access_token_here
+MODEL_ID=facebook/sam2-hiera-tiny
+IMAGE_URL= your_cvat_image_here #443536374637.dkr.ecr.eu-west-1.amazonaws.com/agents/sam2:1.0.7
```
azhavoro:
```suggestion
IMAGE_URL= your_cvat_image_here
```

azhavoro:
I attempted to run this locally and have encountered several issues.

1. After identifying a user without root privileges, registering a new function does not work due to lack of w permissions to the `/shared` directory:
    ```
    ➜ docker compose up
    [+] up 5/5
     ✔ Network sam2_default                    Created                                                                                                                                                                                                                                  0.0s
     ✔ Volume sam2_shared-data                 Created                                                                                                                                                                                                                                  0.0s
     ✔ Container sam2-cvat-function-register-1 Created                                                                                                                                                                                                                                  0.0s
     ✔ Container sam2-cvat-agents-2            Created                                                                                                                                                                                                                                  0.0s
     ✔ Container sam2-cvat-agents-1            Created                                                                                                                                                                                                                                  0.0s
    Attaching to cvat-agents-1, cvat-agents-2, cvat-function-register-1
    Container sam2-cvat-function-register-1 Waiting
    cvat-function-register-1  | Authenticated as user: azhavoro
    cvat-function-register-1  | Function with name SAM2 not found. Proceeding to create a new one.
    cvat-function-register-1  | Please have some patience, function creation might take some time...
    cvat-function-register-1  | [2026-02-17 14:26:48] WARNING: Server version '2.56.1' is not compatible with SDK version '2.57.0'. Some SDK functions may not work properly with this server. You can continue using this SDK, or you can try to update with 'pip install cvat-sdk'.
    cvat-function-register-1  | [2026-02-17 14:26:57] INFO: Created function #4150: SAM2
    cvat-function-register-1  | INFO:cvat_cli.__main__:Created function #4150: SAM2
    cvat-function-register-1  | Successfully created SAM2 function
    cvat-function-register-1  | This is not a Kubernetes deployment. Skipping creation of ConfigMap.
    cvat-function-register-1  | Function registration complete!
    cvat-function-register-1  | /app/function_registration.sh: line 51: /shared/FUNCTION_ID: Permission denied
    Container sam2-cvat-function-register-1 Exited
    cvat-function-register-1 exited with code 0
    cvat-agents-1             | FUNCTION_ID environment variable not found. In compose it should be available in /shared/FUNCTION_ID file.
    cvat-agents-1             | If this is Helm - something is wrong with function registration
    cvat-agents-1             | If this is local run - please ensure FUNCTION_ID environment variable is set or /shared/FUNCTION_ID file is created with the function id of the function you want to run.
    cvat-agents-1             | Exiting...
    cvat-agents-1 exited with code 1
    cvat-agents-2             | FUNCTION_ID environment variable not found. In compose it should be available in /shared/FUNCTION_ID file.
    cvat-agents-2             | If this is Helm - something is wrong with function registration
    cvat-agents-2             | If this is local run - please ensure FUNCTION_ID environment variable is set or /shared/FUNCTION_ID file is created with the function id of the function you want to run.
    cvat-agents-2             | Exiting...
    cvat-agents-2 exited with code 1
    ```
2. This only worked with agent replica=1; otherwise, I got errors as shown below. I'm not sure if this is related to the deployment method, but in any case, it doesn't work out of the box:
    ```
    Attaching to cvat-agents-1, cvat-agents-2, cvat-agents-3, cvat-function-register-1
    Container sam2-cvat-function-register-1 Waiting
    cvat-function-register-1  | Authenticated as user: azhavoro
    cvat-function-register-1  | Found existing function with ID: 4150, new function won't be created.
    cvat-function-register-1  | total 12
    cvat-function-register-1  | drwxr-xr-x 2 agent agent 4096 Feb 17 14:40 .
    cvat-function-register-1  | drwxr-xr-x 1 root  root  4096 Feb 17 14:40 ..
    cvat-function-register-1  | -rw-r--r-- 1 agent agent    5 Feb 17 14:57 FUNCTION_ID
    cvat-function-register-1  | This is not a Kubernetes deployment. Skipping creation of ConfigMap.
    cvat-function-register-1  | Function registration complete!
    cvat-function-register-1 exited with code 0
    
    cvat-agents-2             | FUNCTION_ID file found. Reading FUNCTION_ID from /shared/FUNCTION_ID
    cvat-agents-2             | Running SAM2 function agent for FUNCTION_ID: 4150 with MODEL_ID: facebook/sam2-hiera-tiny
    cvat-agents-2             | ...
    cvat-agents-3             | FUNCTION_ID file found. Reading FUNCTION_ID from /shared/FUNCTION_ID
    cvat-agents-3             | Running SAM2 function agent for FUNCTION_ID: 4150 with MODEL_ID: facebook/sam2-hiera-tiny
    cvat-agents-3             | ...
    cvat-agents-1             | FUNCTION_ID file found. Reading FUNCTION_ID from /shared/FUNCTION_ID
    cvat-agents-1             | Running SAM2 function agent for FUNCTION_ID: 4150 with MODEL_ID: facebook/sam2-hiera-tiny
    cvat-agents-1             | ...
    cvat-agents-2             | [2026-02-17 15:00:31] WARNING: Server version '2.56.1' is not compatible with SDK version '2.57.0'. Some SDK functions may not work properly with this server. You can continue using this SDK, or you can try to update with 'pip install cvat-sdk'.
    cvat-agents-2             | [2026-02-17 15:00:31] INFO: Will store cache at /tmp/tmpxuy5orwi/cache
    cvat-agents-3             | [2026-02-17 15:00:31] WARNING: Server version '2.56.1' is not compatible with SDK version '2.57.0'. Some SDK functions may not work properly with this server. You can continue using this SDK, or you can try to update with 'pip install cvat-sdk'.
    cvat-agents-3             | [2026-02-17 15:00:31] INFO: Will store cache at /tmp/tmpe2t6hnhq/cache
    cvat-agents-1             | [2026-02-17 15:00:32] WARNING: Server version '2.56.1' is not compatible with SDK version '2.57.0'. Some SDK functions may not work properly with this server. You can continue using this SDK, or you can try to update with 'pip install cvat-sdk'.
    cvat-agents-1             | [2026-02-17 15:00:32] INFO: Will store cache at /tmp/tmp67v5jpz0/cache
    cvat-agents-1             | [2026-02-17 15:00:35] INFO: Agent starting with ID '5bee3e145174b31ec26e7582ecafe302'
    cvat-agents-1             | [2026-02-17 15:00:35] INFO: Attempting to watch the function's queue...
    cvat-agents-1             | [2026-02-17 15:00:35] INFO: Trying to acquire an annotation request of category 'interactive'...
    cvat-agents-1             | [2026-02-17 15:00:36] INFO: Connected to the queue event stream
    cvat-agents-1             | [2026-02-17 15:00:36] INFO: New queue event stream reconnection delay is 10.000000s
    cvat-agents-1             | [2026-02-17 15:00:36] INFO: Trying to acquire an annotation request of category 'batch'...
    cvat-agents-3             | [2026-02-17 15:00:42] INFO: Agent starting with ID '52b25c1f3a44749f8d6830f5d9dc0cb5'
    cvat-agents-3             | [2026-02-17 15:00:42] INFO: Attempting to watch the function's queue...
    cvat-agents-3             | [2026-02-17 15:00:42] INFO: Trying to acquire an annotation request of category 'interactive'...
    cvat-agents-3             | [2026-02-17 15:00:42] INFO: Connected to the queue event stream
    cvat-agents-3             | [2026-02-17 15:00:42] INFO: New queue event stream reconnection delay is 10.000000s
    cvat-agents-2             | [2026-02-17 15:00:42] INFO: Agent starting with ID '8b73c0875d898b67a818fbe8da80eca2'
    cvat-agents-2             | [2026-02-17 15:00:42] INFO: Attempting to watch the function's queue...
    cvat-agents-2             | [2026-02-17 15:00:42] INFO: Trying to acquire an annotation request of category 'interactive'...
    cvat-agents-2             | [2026-02-17 15:00:42] INFO: Connected to the queue event stream
    cvat-agents-2             | [2026-02-17 15:00:42] INFO: New queue event stream reconnection delay is 10.000000s
    cvat-agents-3             | [2026-02-17 15:00:42] INFO: Trying to acquire an annotation request of category 'batch'...
    cvat-agents-2             | [2026-02-17 15:00:43] INFO: Trying to acquire an annotation request of category 'batch'...
    cvat-agents-1             | [2026-02-17 15:01:13] INFO: Received notification about a new request of category 'interactive'
    cvat-agents-1             | [2026-02-17 15:01:13] INFO: Trying to acquire an annotation request of category 'interactive'...
    cvat-agents-3             | [2026-02-17 15:01:13] INFO: Received notification about a new request of category 'interactive'
    cvat-agents-3             | [2026-02-17 15:01:13] INFO: Trying to acquire an annotation request of category 'interactive'...
    cvat-agents-2             | [2026-02-17 15:01:13] INFO: Received notification about a new request of category 'interactive'
    cvat-agents-2             | [2026-02-17 15:01:13] INFO: Trying to acquire an annotation request of category 'interactive'...
    cvat-agents-1             | [2026-02-17 15:01:13] INFO: Got assigned annotation request '726b16d7-1435-435f-87c8-67e80da91165' of type 'init_tracking' (task=1883394 frame=0)
    cvat-agents-1             | [2026-02-17 15:01:13] INFO: Fetching task 1883394...
    cvat-agents-1             | [2026-02-17 15:01:13] INFO: Task 1883394 is not yet cached or the cache is corrupted
    cvat-agents-1             | [2026-02-17 15:01:13] INFO: Fetching labels...
    cvat-agents-1             | [2026-02-17 15:01:13] INFO: Downloading data metadata...
    cvat-agents-1             | [2026-02-17 15:01:14] INFO: Downloaded data metadata
    cvat-agents-1             | /app/func.py:106: UserWarning: cannot import name '_C' from 'sam2' (/usr/local/lib/python3.10/site-packages/sam2/__init__.py)
    cvat-agents-1             |
    cvat-agents-1             | Skipping the post-processing step due to the error above. You can still use SAM 2 and it's OK to ignore the error above, although some post-processing functionality may be limited (which doesn't affect the results in most cases; see https://github.com/facebookresearch/sam2/blob/main/INSTALL.md).
    cvat-agents-1             |   "pred_masks": fill_holes_in_mask_scores(
    cvat-agents-1             | [2026-02-17 15:01:15] INFO: Submitting result for AR '726b16d7-1435-435f-87c8-67e80da91165'...
    cvat-agents-1             | [2026-02-17 15:01:15] INFO: AR '726b16d7-1435-435f-87c8-67e80da91165' completed
    cvat-agents-1             | [2026-02-17 15:01:15] INFO: Trying to acquire an annotation request of category 'interactive'...
    cvat-agents-1             | [2026-02-17 15:01:16] INFO: Received notification about a new request of category 'interactive'
    cvat-agents-1             | [2026-02-17 15:01:16] INFO: Trying to acquire an annotation request of category 'interactive'...
    cvat-agents-3             | [2026-02-17 15:01:16] INFO: Received notification about a new request of category 'interactive'
    cvat-agents-3             | [2026-02-17 15:01:16] INFO: Trying to acquire an annotation request of category 'interactive'...
    cvat-agents-2             | [2026-02-17 15:01:16] INFO: Received notification about a new request of category 'interactive'
    cvat-agents-2             | [2026-02-17 15:01:16] INFO: Trying to acquire an annotation request of category 'interactive'...
    cvat-agents-1             | [2026-02-17 15:01:16] INFO: Got assigned annotation request 'adcb3cde-3acf-42af-8a57-2bc3f17d47b4' of type 'track' (task=1883394 frame=1)
    cvat-agents-1             | [2026-02-17 15:01:16] INFO: Fetching task 1883394...
    cvat-agents-1             | [2026-02-17 15:01:16] INFO: Fetching labels...
    cvat-agents-1             | [2026-02-17 15:01:16] INFO: Loaded data metadata from cache
    cvat-agents-1             | [2026-02-17 15:01:18] INFO: Submitting result for AR 'adcb3cde-3acf-42af-8a57-2bc3f17d47b4'...
    cvat-agents-1             | [2026-02-17 15:01:18] INFO: AR 'adcb3cde-3acf-42af-8a57-2bc3f17d47b4' completed
    cvat-agents-1             | [2026-02-17 15:01:18] INFO: Trying to acquire an annotation request of category 'interactive'...
    cvat-agents-1             | [2026-02-17 15:01:19] INFO: Received notification about a new request of category 'interactive'
    cvat-agents-1             | [2026-02-17 15:01:19] INFO: Trying to acquire an annotation request of category 'interactive'...
    
    
    cvat-agents-3             | [2026-02-17 15:01:19] INFO: Received notification about a new request of category 'interactive'
    cvat-agents-2             | [2026-02-17 15:01:19] INFO: Received notification about a new request of category 'interactive'
    
    
    cvat-agents-2             | [2026-02-17 15:01:19] INFO: Trying to acquire an annotation request of category 'interactive'...
    
    
    cvat-agents-3             | [2026-02-17 15:01:19] INFO: Trying to acquire an annotation request of category 'interactive'...
    cvat-agents-3             | [2026-02-17 15:01:19] INFO: Got assigned annotation request '74a386f4-7521-47df-9498-92f1c61ded83' of type 'track' (task=1883394 frame=2)
    cvat-agents-3             | [2026-02-17 15:01:19] INFO: Fetching task 1883394...
    cvat-agents-3             | [2026-02-17 15:01:19] INFO: Task 1883394 is not yet cached or the cache is corrupted
    cvat-agents-3             | [2026-02-17 15:01:19] INFO: Fetching labels...
    cvat-agents-3             | [2026-02-17 15:01:19] INFO: Downloading data metadata...
    cvat-agents-3             | [2026-02-17 15:01:19] INFO: Downloaded data metadata
    cvat-agents-3             | [2026-02-17 15:01:21] ERROR: Failed to process AR '74a386f4-7521-47df-9498-92f1c61ded83'
    cvat-agents-3             | concurrent.futures.process._RemoteTraceback:
    cvat-agents-3             | """
    cvat-agents-3             | Traceback (most recent call last):
    cvat-agents-3             |   File "/usr/local/lib/python3.10/concurrent/futures/process.py", line 246, in _process_worker
    cvat-agents-3             |     r = call_item.fn(*call_item.args, **call_item.kwargs)
    cvat-agents-3             |   File "/usr/local/lib/python3.10/site-packages/cvat_cli/_internal/agent.py", line 249, in _worker_job_track
    cvat-agents-3             |     return list(map(track, states))
    cvat-agents-3             |   File "/usr/local/lib/python3.10/site-packages/cvat_cli/_internal/agent.py", line 232, in track
    cvat-agents-3             |     inner_state, original_shape_type = _tracking_states.retrieve(
    cvat-agents-3             |   File "/usr/local/lib/python3.10/site-packages/cvat_cli/_internal/agent.py", line 151, in retrieve
    cvat-agents-3             |     raise _BadArError(f"Tracking state {state_id!r} not found - possibly expired")
    cvat-agents-3             | cvat_cli._internal.agent._BadArError: Tracking state 'ViX0KUQrHhIatgc5O8BYGpsV0BTpy8U4lQy5V3fHnx4' not found - possibly expired
    cvat-agents-3             | """
    cvat-agents-3             |
    cvat-agents-3             | The above exception was the direct cause of the following exception:
    cvat-agents-3             |
    cvat-agents-3             | Traceback (most recent call last):
    cvat-agents-3             |   File "/usr/local/lib/python3.10/site-packages/cvat_cli/_internal/agent.py", line 705, in _process_ar
    cvat-agents-3             |     result = self._calculate_result_for_ar(ar_id, ar_params)
    cvat-agents-3             |   File "/usr/local/lib/python3.10/site-packages/cvat_cli/_internal/agent.py", line 908, in _calculate_result_for_tracking_ar
    cvat-agents-3             |     return self._calculate_result_for_track_ar(ar_id, ar_params)
    cvat-agents-3             |   File "/usr/local/lib/python3.10/site-packages/cvat_cli/_internal/agent.py", line 937, in _calculate_result_for_track_ar
    cvat-agents-3             |     shapes = self._executor.result(
    cvat-agents-3             |   File "/usr/local/lib/python3.10/site-packages/cvat_cli/_internal/agent.py", line 103, in result
    cvat-agents-3             |     return future.result()
    cvat-agents-3             |   File "/usr/local/lib/python3.10/concurrent/futures/_base.py", line 458, in result
    cvat-agents-3             |     return self.__get_result()
    cvat-agents-3             |   File "/usr/local/lib/python3.10/concurrent/futures/_base.py", line 403, in __get_result
    cvat-agents-3             |     raise self._exception
    cvat-agents-3             | cvat_cli._internal.agent._BadArError: Tracking state 'ViX0KUQrHhIatgc5O8BYGpsV0BTpy8U4lQy5V3fHnx4' not found - possibly expired
    cvat-agents-3             | [2026-02-17 15:01:21] INFO: AR '74a386f4-7521-47df-9498-92f1c61ded83' failed
    cvat-agents-3             | [2026-02-17 15:01:21] INFO: Trying to acquire an annotation request of category 'interactive'...
    ```
3. It is not possible to register a function in an organization, only in a personal workspace. I believe it would be better to support this case as well.

ai-models/agents_deployment/sam2/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+WORKDIR /app
+
+COPY ai-models/tracker/sam2/requirements.txt .
+RUN pip install --no-cache-dir cvat-cli -r requirements.txt
+
+COPY ai-models/tracker/sam2/ .
+COPY ai-models/agents_deployment/sam2/function_registration.sh .
+COPY ai-models/agents_deployment/sam2/function_deregistration.sh .
+COPY ai-models/agents_deployment/sam2/entrypoint.sh /entrypoint.sh
```
azhavoro: Why did you choose use relative paths from the repo root here? This requires using the repo root as the build context, which isn't very intuitive from my point...

yaizkazani: Because i want to add image build to CI so we will always have updated image

ai-models/agents_deployment/sam2/docker-compose.yaml
SpecLad: I should note that even for SAM2 additional parameters would likely be required; in particular, users will likely want to set the device the model runs on.

ai-models/agents_deployment/sam2/docker-compose.yaml
```diff
@@ -0,0 +1,42 @@
...
+      - shared-data:/shared:ro
+    depends_on:
+      cvat-function-register:
+        condition: service_completed_successfully
+    deploy:
+      replicas: ${AGENTS_COUNT:-1}
+    environment:
+      - CVAT_BASE_URL=${CVAT_BASE_URL}
+      - CVAT_ACCESS_TOKEN=${CVAT_ACCESS_TOKEN}
+      - MODEL_ID=${MODEL_ID}
```
SpecLad:
```suggestion
      CVAT_BASE_URL:
      CVAT_ACCESS_TOKEN:
      MODEL_ID:
```

ai-models/agents_deployment/sam2/docker-compose.yaml
```diff
@@ -0,0 +1,42 @@
...
+    entrypoint: ["/app/function_registration.sh"]
+    volumes:
+      - shared-data:/shared
+    restart: "on-failure"
+    environment:
+      - CVAT_BASE_URL=${CVAT_BASE_URL}
+      - CVAT_ACCESS_TOKEN=${CVAT_ACCESS_TOKEN}
+      - MODEL_ID=${MODEL_ID}
+
+  cvat-agents:
```
SpecLad:
```suggestion
  cvat-agent:
```

Compose will create containers with names like "cvat-agent-1", "cvat-agent-2", etc., so plural doesn't really make sense here.

ai-models/agents_deployment/sam2/docker-compose.yaml
```diff
@@ -0,0 +1,42 @@
...
+      replicas: ${AGENTS_COUNT:-1}
+    environment:
+      - CVAT_BASE_URL=${CVAT_BASE_URL}
+      - CVAT_ACCESS_TOKEN=${CVAT_ACCESS_TOKEN}
+      - MODEL_ID=${MODEL_ID}
+
+  #NB! MANUAL RUN!
+  # docker compose run --rm cvat-function-deregister
+  # or
+  # docker compose --profile manual up cvat-function-deregister
```
SpecLad: IMO, it doesn't make much sense to provide an alternative command which is longer and requires more information (and also leaves a stopped container behind, if I understand correctly).

ai-models/agents_deployment/sam2/Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM python:3.10-slim
```
SpecLad: 3.10 is the oldest version of Python that is still in support, and it's due to enter EOL this year. We should use the newest possible version instead, so that it can remain in support for longer.

ai-models/agents_deployment/sam2/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+
+COPY ai-models/tracker/sam2/requirements.txt .
+RUN pip install --no-cache-dir cvat-cli -r requirements.txt
+
+COPY ai-models/tracker/sam2/ .
+COPY ai-models/agents_deployment/sam2/function_registration.sh .
+COPY ai-models/agents_deployment/sam2/function_deregistration.sh .
+COPY ai-models/agents_deployment/sam2/entrypoint.sh /entrypoint.sh
+
+RUN chmod +x /entrypoint.sh function_registration.sh function_deregistration.sh
```
SpecLad: This is odd, shouldn't these files already be executable?

ai-models/agents_deployment/sam2/entrypoint.sh
```diff
@@ -0,0 +1,36 @@
...
+if [ -z "$CVAT_ACCESS_TOKEN" ]; then
+    echo "Error: CVAT_ACCESS_TOKEN environment variable must be set."
+    exit 1
+fi
+
+if [ -z "$CVAT_BASE_URL" ]; then
+    echo "Warning: CVAT_BASE_URL environment variable is missing, using https://app.cvat.ai as default."
+    CVAT_BASE_URL="https://app.cvat.ai"
+fi
+
```
SpecLad: It seems that this logic is repeated in all scripts, could you factor it out?

ai-models/agents_deployment/sam2/entrypoint.sh
```diff
@@ -0,0 +1,36 @@
...
+if [ -z "$MODEL_ID" ]; then
+    echo "Error: MODEL_ID environment variable not found. Default is facebook/sam2.1-hiera-tiny"
```
SpecLad: I'm not sure a reasonable default exists for this. Maybe we should just require this variable to be set.

ai-models/agents_deployment/sam2/entrypoint.sh
```diff
@@ -0,0 +1,36 @@
...
+
+
+if [ -f /shared/FUNCTION_ID ]; then
+    echo "FUNCTION_ID file found. Reading FUNCTION_ID from /shared/FUNCTION_ID"
+    FUNCTION_ID="$(cat /shared/FUNCTION_ID)"
+fi
+
+
+if [ -z "$FUNCTION_ID" ]; then
+    echo -e "FUNCTION_ID environment variable not found. In compose it should be available in /shared/FUNCTION_ID file.\nIf this is Helm - something is wrong with function registration\nIf this is local run - please ensure FUNCTION_ID environment variable is set or /shared/FUNCTION_ID file is created with the function id of the function you want to run.\nExiting..."
```
SpecLad: Could you use a separate echo for every line? This is hard to read.

ai-models/agents_deployment/sam2/entrypoint.sh
```diff
@@ -0,0 +1,36 @@
...
+    exit 1
+fi
+
+FUNCTION_FILE_PATH="func.py"
+
+echo -e "Running SAM2 function agent for FUNCTION_ID: $FUNCTION_ID with MODEL_ID: $MODEL_ID\n..."
```
SpecLad: There's a linebreak before the ellipsis, is this a typo?

ai-models/agents_deployment/sam2/function_deregistration.sh
```diff
@@ -0,0 +1,34 @@
...
+    exit 1
+fi
+
+if result=$(cvat-cli --server-host "$CVAT_BASE_URL" function delete "$FUNCTION_ID"); then
+  echo "Function with ID: $FUNCTION_ID has been removed successfully."
+else
+  echo "cvat-cli function delete failed."
+  echo "$result"
+  exit 1
+fi
```
SpecLad: All this seems unnecessary. The `cvat-cli function delete` command already logs a message explaining whether it was successful or not. You can just `exec` it.

ai-models/agents_deployment/sam2/function_registration.sh
```diff
@@ -0,0 +1,122 @@
...
+
+if [ -z "$MODEL_ID" ]; then
+    echo "Warning: MODEL_ID environment variable not found. Default is facebook/sam2.1-hiera-tiny"
+    MODEL_ID="facebook/sam2.1-hiera-tiny"
+fi
+
+# We use it for compose but we have same image for Helm.
+if [ ! -d /shared ]; then
+    mkdir /shared
+fi
```
SpecLad:
```suggestion
mkdir -p /shared
```

Although will this work when executing as a non-root user? You might want to create this directory in the dockerfile instead.

ai-models/agents_deployment/sam2/function_registration.sh
```diff
@@ -0,0 +1,122 @@
...
+
+# Hardcoded for SAM2
+
+FUNCTION_NAME="SAM2"
+FUNCTION_FILE_PATH="func.py"
+
+
+# Get the username associated with the access token
+USERNAME=$(curl "$CVAT_BASE_URL"/api/users/self -s -H "Authorization: Bearer $CVAT_ACCESS_TOKEN" | jq -r '.username')
```
SpecLad:
```suggestion
USERNAME=$(curl "$CVAT_BASE_URL"/api/users/self -s --oauth2-bearer "$CVAT_ACCESS_TOKEN" | jq -r '.username')
```

ai-models/agents_deployment/sam2/function_registration.sh
```diff
@@ -0,0 +1,122 @@
...
+if [ -z "$USERNAME" ] || [ "$USERNAME" = "null" ]; then
```
SpecLad: `null` is technically a valid username. What is this check supposed to guard against?

yaizkazani: image

ai-models/agents_deployment/sam2/docker-compose.yaml
yaizkazani: Valid point, however:
https://docs.docker.com/compose/how-tos/environment-variables/set-environment-variables/

"You can also take advantage of [interpolation](https://docs.docker.com/compose/how-tos/environment-variables/variable-interpolation/#interpolation-syntax). In the following example, the result is similar to the one above but Compose gives you a warning if the DEBUG variable is not set in the shell environment or in an .env file in the project directory."

But since the scripts actually throw warnings if env vars are missing this is a good point.

ai-models/agents_deployment/sam2/docker-compose.yaml
yaizkazani: Why not? :)

ai-models/agents_deployment/sam2/docker-compose.yaml
yaizkazani: Agreed

ai-models/agents_deployment/sam2/Dockerfile
yaizkazani: Ok, this is a good point indeed. Lets try python:3.12-slim 

I will rebuild the image and test compose

ai-models/agents_deployment/sam2/Dockerfile
yaizkazani: Not by default. During image build they inherit the +x permission if it is set on the host where image is built, however to have chmod is a good practice imo for reliability.

ai-models/agents_deployment/sam2/entrypoint.sh
yaizkazani: Yes, I can factor it out, but i think it won't make things much better in terms of visibility. I respect functional programming paradigm, but not in this case :) 

I think it could be better just to write some comments around these blocks.

ai-models/agents_deployment/sam2/entrypoint.sh
yaizkazani: Sorry, but i disagree. My point is:

1. We should have some default value for all the images that we build for functions that we have in ai-models/ 
2. Scripts should notify user that the model id is missing and that some default is used

I think the solution should run with minimal configuration and user should be aware when defaults are used so he can tune the agent as he likes.

ai-models/agents_deployment/sam2/entrypoint.sh
yaizkazani: Of course

ai-models/agents_deployment/sam2/entrypoint.sh
yaizkazani: Yep

ai-models/agents_deployment/sam2/function_registration.sh
yaizkazani: yes, this was removed, I just haven't pushed the latest version. Indeed it is created in dockerfile + chown for non-root user

ai-models/agents_deployment/sam2/function_deregistration.sh
yaizkazani: I agree - this is a good point, btw.

ai-models/agents_deployment/sam2/function_registration.sh
yaizkazani: Ok, agreed

ai-models/agents_deployment/sam2/function_registration.sh
yaizkazani: Well, ok, I will remove that, but my idea was "if API returns null - something wrong has happened", but maybe we should not allow such username?!

ai-models/agents_deployment/sam2/Dockerfile
SpecLad: Why not 3.14?

ai-models/agents_deployment/sam2/Dockerfile
SpecLad: Okay, I suppose this will help if you're building on Windows. I'd suggest to use `COPY --chmod=+x`, though, to avoid creating another layer.

ai-models/agents_deployment/sam2/Dockerfile
yaizkazani: Agent is starting on 3.14

yaizkazani: SAM2 tasks are running fine

ai-models/agents_deployment/sam2/entrypoint.sh
SpecLad: It's not functional programming, it's just normal programming. :man_shrugging: 

I don't really know what you mean by "visibility", but factoring this logic out would prevent divergent behavior between scripts.

ai-models/agents_deployment/sam2/function_registration.sh
SpecLad: It's just a normal username. There's people whose actual name is Null, you know. :slightly_smiling_face:

ai-models/agents_deployment/sam2/docker-compose.yaml
```diff
@@ -0,0 +1,48 @@
...
+    entrypoint: ["/app/function_registration.sh"]
+    volumes:
+      - shared-data:/shared
+    restart: "on-failure"
+    environment:
+      CVAT_BASE_URL:
+      CVAT_ACCESS_TOKEN:
+      MODEL_ID:
+      ORG_SLUG:
+      SERVER_PORT:
```
SpecLad: This variable isn't really necessary; you can already override the port by appending :XXX to `CVAT_BASE_URL`.

ai-models/agents_deployment/sam2/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+
+WORKDIR /app
+
+COPY ai-models/tracker/sam2/requirements.txt .
+RUN pip install --no-cache-dir cvat-cli -r requirements.txt
+
+COPY ai-models/tracker/sam2/ .
+COPY ai-models/agents_deployment/sam2/function_registration.sh .
+COPY ai-models/agents_deployment/sam2/function_deregistration.sh .
+COPY ai-models/agents_deployment/sam2/entrypoint.sh /entrypoint.sh
```
SpecLad: Why is this script copied to `/`, but the other two to `/app`? Seems like a pointless inconsistency.

ai-models/agents_deployment/sam2/function_deregistration.sh
```diff
@@ -0,0 +1,43 @@
...
+else
+    echo "ORG_SLUG environment variable found. Function will be registered for the organization with slug: $ORG_SLUG"
+    ORG_SLUG_ARGS=(--organization "$ORG_SLUG")
+fi
+
+# Compose way to share FUNCTION_ID between containers
+if [ -f /shared/FUNCTION_ID ]; then
+    FUNCTION_ID="$(cat /shared/FUNCTION_ID)"
+else
+    echo -e "Warning: FUNCTION_ID file not found at /shared/FUNCTION_ID. If this is docker-compose run, something went wrong with function registration.\nOtherwise, if this is local run, please ensure FUNCTION_ID environment variable is set or /shared/FUNCTION_ID file is created with the function id of the function you want to remove."
```
SpecLad: Will this warning be printed in a Kubernetes deployment?

ai-models/agents_deployment/sam2/Dockerfile
yaizkazani: It will require docker 23+ if google doesnt lie + these files has almost 0 weight, but why not? Ok

ai-models/agents_deployment/sam2/function_registration.sh
```diff
@@ -0,0 +1,134 @@
...
+fi
+
+# Hardcoded for SAM2
+
+FUNCTION_NAME="SAM2"
+FUNCTION_FILE_PATH="func.py"
+
+
+# Get the username associated with the access token
+USERNAME=$(curl "$CVAT_BASE_URL"/api/users/self -s --oauth2-bearer "$CVAT_ACCESS_TOKEN" | jq -r '.username')
```
SpecLad: I would suggest using the `--fail`/`-f` curl option here and everywhere else, so that error responses are not passed to jq.

ai-models/agents_deployment/sam2/function_registration.sh
```diff
@@ -0,0 +1,134 @@
...
+if [ -z "$USERNAME" ]; then
+    echo "Error: Unable to retrieve username from CVAT API. Check your access token."
+    exit 1
+else
+    echo "Authenticated as user: $USERNAME"
+fi
+
+# Try to find existing function with the same name and owner
+if request_function_id=$(curl --get --data-urlencode "filter={\"and\":[{\"==\":[{\"var\":\"name\"},\"${FUNCTION_NAME}\"]},{\"==\":[{\"var\":\"owner\"},\"$USERNAME\"]}]}" "$CVAT_BASE_URL"/api/functions -s -H "Authorization: Bearer "$CVAT_ACCESS_TOKEN"" | jq -r '.results[0].id') && [ "$request_function_id" != "null" ]; then
```
SpecLad:
```suggestion
if request_function_id=$(curl --get --data-urlencode "filter={\"and\":[{\"==\":[{\"var\":\"name\"},\"${FUNCTION_NAME}\"]},{\"==\":[{\"var\":\"owner\"},\"$USERNAME\"]}]}" "$CVAT_BASE_URL"/api/functions -s --oauth2-bearer "$CVAT_ACCESS_TOKEN" | jq -r '.results[0].id') && [ "$request_function_id" != "null" ]; then
```

ai-models/agents_deployment/sam2/function_registration.sh
```diff
@@ -0,0 +1,134 @@
...
+    FUNCTION_ID="$request_function_id"
```
SpecLad: It seems that both `request_function_id` and `function_id` are redundant - you can just use `FUNCTION_ID` instead.

ai-models/agents_deployment/sam2/function_registration.sh
```diff
@@ -0,0 +1,134 @@
...
+      echo "$function_id" > /shared/FUNCTION_ID
+    else
+      echo "cvat-cli function create-native failed."
+      exit 1
+    fi
+fi
+
+
+# Create configmap with FUNCTION_ID
+if [ -z "$K8S_DEPLOY" ]; then
```
SpecLad:
```suggestion
if [ -z "$KUBERNETES_SERVICE_HOST" ]; then
```

There's no need to invent a custom variable when there's a builtin one.

ai-models/agents_deployment/sam2/function_registration.sh
```diff
@@ -0,0 +1,134 @@
...
+    echo "This is not a Kubernetes deployment. Skipping creation of ConfigMap."
+    echo "Function registration complete!"
+    exit 0
+fi
+
+## https://kubernetes.io/docs/tasks/run-application/access-api-from-pod/
```
SpecLad: Can we just install kubectl into the container and use it? It seems like basically everything below this point can be replaced with `kubectl create configmap "${CONFIGMAP_NAME}" --from-literal=FUNCTION_ID="${FUNCTION_ID}"`.

ai-models/agents_deployment/sam2/Dockerfile
yaizkazani: done

ai-models/agents_deployment/sam2/function_registration.sh
yaizkazani: Done

ai-models/agents_deployment/sam2/docker-compose.yaml
yaizkazani: cvat-function-register-1  | usage: cvat-cli [-h] [--version] [--insecure] [--auth USER[:PASS]]
cvat-function-register-1  |                 [--server-host SERVER_HOST] [--server-port SERVER_PORT]
cvat-function-register-1  |                 [--organization SLUG] [--debug]
cvat-function-register-1  |                 {function,project,task,create,ls,delete,frames,dump,upload,export,import,auto-annotate}

This is what CLI returns when it get malformed input. So..

But i will remove it - ok.

ai-models/agents_deployment/sam2/Dockerfile
yaizkazani: Why not? This is not an issue to me.

ai-models/agents_deployment/sam2/docker-compose.yaml
SpecLad: What does this have to do with my comment?

ai-models/agents_deployment/sam2/function_deregistration.sh
yaizkazani: Goooood point indeed. Fixed

ai-models/agents_deployment/sam2/function_registration.sh
yaizkazani: :) 

I was thinking about this honestly. In general it doesnt seem to be strictly necessary, but since you also point to this i will add this.

ai-models/agents_deployment/sam2/function_registration.sh
yaizkazani: Yes, I agree. Ok, ill change this

ai-models/agents_deployment/sam2/function_registration.sh
yaizkazani: Ok, why not. I'll test that now

ai-models/agents_deployment/sam2/function_registration.sh
yaizkazani: That's a good point btw, but again this is not Helm yet :) 

Changed that

ai-models/agents_deployment/sam2/docker-compose.yaml
yaizkazani: Am I missing something but this is not mentioned in cvat-cli arguments reference?

SpecLad: Ah, I guess not. But it does work.

ai-models/agents_deployment/sam2/entrypoint.sh
yaizkazani: I have added this.

ai-models/agents_deployment/sam2/docker-compose.yaml
yaizkazani: Yes, of course, but i have added this var looking at the example output from the cli.

yaizkazani: anyway i have removed that so ill close this

ai-models/agents_deployment/sam2/function_registration.sh
yaizkazani: It works :)

ai-models/agents_deployment/sam2/function_deregistration.sh
```diff
@@ -0,0 +1,28 @@
...
+fi
+
+
+if [ -z "$FUNCTION_ID" ]; then
+    echo -e "Error: FUNCTION_ID environment variable must be set to remove function from CVAT.\nPlease consider manual removal using cvat-cli --server-host $CVAT_BASE_URL function delete FUNCTION_ID"
+    exit 1
+fi
+
+if [ -n "$KUBERNETES_SERVICE_HOST" ]; then
+    kubectl delete configmap "$CONFIGMAP_NAME" --namespace "$NAMESPACE"
```
SpecLad, line 25:
```suggestion
    kubectl delete configmap "$CONFIGMAP_NAME"
```

This variable doesn't seem to be defined anywhere (and it shouldn't be necessary anyway).

ai-models/agents_deployment/sam2/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    curl \
+    git \
+    jq \
+    && rm -rf /var/lib/apt/lists/*
+
+ARG TARGETARCH
+RUN curl -LO "https://dl.k8s.io/release/v1.34.0/bin/linux/${TARGETARCH}/kubectl" \
+    && chmod +x kubectl && mv kubectl /usr/local/bin/
```
SpecLad:
```suggestion
ADD --chmod=+x https://dl.k8s.io/release/v1.34.0/bin/linux/${TARGETARCH}/kubectl /usr/local/bin/
```

ai-models/agents_deployment/sam2/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+
+RUN if [ "$TARGETARCH" = "amd64" ]; then \
```
SpecLad: This condition seems unnecessary; the alternate index has ARM64 wheels as well.

ai-models/agents_deployment/sam2/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+        pip install --no-cache-dir torch torchvision --index-url https://download.pytorch.org/whl/cpu; \
```
SpecLad: It would be better to add `--extra-index-url=https://download.pytorch.org/whl/cpu` to the main `pip install` command. Otherwise, the second command might reinstall some packages installed by the first, if the required versions are different.

ai-models/agents_deployment/sam2/check_env.sh
```diff
@@ -0,0 +1,50 @@
...
+        echo "Using MODEL_ID: $MODEL_ID"
+    fi
+}
+
+resolve_cuda() {
+    if [ "$USE_CUDA" = "true" ]; then
+        echo "Using CUDA! Please ensure that you are using proper image with CUDA support"
+        USE_CUDA_ARGS=(-p device=str:cuda)
+    else
+        echo "Warning: USE_CUDA environment variable not found. Model will run on CPU."
```
SpecLad: I don't think this qualifies as a warning, since running on the CPU is not a problem. It's an informational message.

ai-models/agents_deployment/sam2/function_deregistration.sh
yaizkazani: It is defined in Helm. We'd better have it just in case 

image

ai-models/agents_deployment/sam2/check_env.sh
yaizkazani: I agree.

yaizkazani: Done

ai-models/agents_deployment/sam2/Dockerfile
yaizkazani: Thanks !

ai-models/agents_deployment/sam2/Dockerfile
yaizkazani: Thank for helping with this :)

PR #10277, opened 2026-02-13
Fix empty related chunk generation for tasks in cloud storage

SpecLad:
### Motivation and context

When `frame_media` is empty (i.e. there are no related images for a frame), the `truncate_common_filename_prefix` logic crashes, because `os.path.commonpath` requires at least one path as input.

This only happens when the task is in cloud storage, because the non-cloud branch constructs the `media` list in such a way that frames with no related images are simply omitted. This works fine, so factor out that logic and use it in the cloud case as well.

### How has this been tested?

REST API tests.

### Checklist

### License

### DIFF
--- a/cvat/apps/engine/cache.py
+++ b/cvat/apps/engine/cache.py
@@ -696,8 +696,6 @@ class MediaCache:
             except ValueError as ex:
                 raise Exception("Invalid related image path") from ex
 
-        manifest_path = db_data.get_manifest_path()
-
         with ExitStack() as es:
             ThroughModel = models.RelatedFile.images.through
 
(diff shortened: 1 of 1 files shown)


cvat/apps/engine/cache.py
```diff
@@ -699,68 +699,53 @@ def _validate_ri_path(path: str) -> PurePath:
         manifest_path = db_data.get_manifest_path()
```
zhiltsov-max: This variable can be removed now

SpecLad: Oh yeah, good point. Removed.

PR #10282, opened 2026-02-16
Support using cloud storages as backing storage for local tasks

SpecLad:
### Motivation and context

This allows storing media data of "local" tasks in cloud storage, transparently to the user. The CVAT administrator must explicitly move tasks to backing cloud storage (or back) using the new management commands.

Currently, only tasks with images using cached chunks are supported.

### How has this been tested?

I manually checked this works with all 3 CS providers + unit tests for S3.

### Checklist

### License

### DIFF
--- a/tests/python/rest_api/test_tasks.py
+++ b/tests/python/rest_api/test_tasks.py
@@ -1459,6 +1459,32 @@ class TestTaskBackups:
 
         self._test_can_restore_task_from_backup(task["id"])
 
+    @pytest.mark.with_external_services
+    def test_can_export_and_import_backup_with_backing_cs(self, request, cloud_storages):
+        cloud_storage_id = next(cs["id"] for cs in cloud_storages if cs["resource"] == "backingcs")
+
+        with make_sdk_client(self.user) as client:
+            task = client.tasks.create_from_data(
+                models.TaskWriteRequest(name="Canvas3D"),
+                [SHARE_DIR / "test_canvas3d.zip"],
+                data_params={"use_cache": True},
+            )
+
+            container_exec_cvat(
+                request, ["./manage.py", "movetasktobackingcs", str(task.id), str(cloud_storage_id)]
+            )
+
+            backup_path = self.tmp_dir / "backup.zip"
+            task.download_backup(backup_path)
+
+            with zipfile.ZipFile(backup_path) as zip_file:
+                names = zip_file.namelist()
+
+                assert any(name.endswith(".pcd") for name in names)
+                assert any(name.endswith(".png") for name in names)
+
+            self._test_can_restore_task_from_backup(task.id, backup_file=backup_path)
+
     def _test_can_restore_task_from_backup(
         self,
         task_id: int,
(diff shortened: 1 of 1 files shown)


cvat/apps/engine/cloud_provider.py
```diff
@@ -241,6 +241,12 @@ def bulk_download_to_memory(
                 top_job = queue.get()
                 yield top_job.result()
 
+    def _in_parallel(self, fn: Callable[[T], object], args: Sequence[T]) -> None:
+        threads_number = get_max_threads_number(len(args))
+
+        with ThreadPoolExecutor(max_workers=threads_number) as executor:
+            list(executor.map(fn, args))
```
zhiltsov-max, line 248: Previously we used a custom queue to limit the maximum number of uncollected results stored in memory. Why it was removed? Though I agree that the 2 remaining uses have no output value in the called funciton.

cvat/apps/engine/management/commands/movetasktobackingcs.py
zhiltsov-max, line 1: Consider adding a script for bulk task transfer, because moving them one by one can take a lot of time for a big number of tasks. I think it can be added in a later PR, if you're going to run this one in an experimental mode first.

cvat/apps/engine/cloud_provider.py
SpecLad: The custom queue is only used in `bulk_download_to_memory`, which I didn't touch. I only used `_in_parallel` with functions returning `None`, since in this case the extra storage should be inconsequential.

cvat/apps/engine/models.py
```diff
@@ -540,6 +544,96 @@ def update_validation_layout(
     def validation_mode(self) -> ValidationMode | None:
         return getattr(getattr(self, 'validation_layout', None), 'mode', None)
 
+    def get_all_media_rel_paths(self) -> list[PurePath]:
+        if video := getattr(self, "video", None):
+            return [PurePath(video.path)]
+        else:
+            return (
+                [PurePath(image.path) for image in self.images.all()]
```
zhiltsov-max: Need to skip `is_placeholder` ones here or deduplicate, they represent honeypots. Honeypots reuse the same file multiple times. Having them included resulted in a `movetasktobackingcs` failure in my case.

cvat/apps/engine/signals.py
```diff
@@ -126,12 +126,30 @@ def __delete_job_handler(instance, **kwargs):
...
         functools.partial(shutil.rmtree, instance.get_data_dirname(), ignore_errors=True)
     )
 
+    if instance.local_storage_backing_cs:
+        storage_instance = instance.get_cloud_storage_instance()
+        assert storage_instance
+
+        transaction.on_commit(
+            functools.partial(
+                storage_instance.bulk_delete,
```
zhiltsov-max, line 147: Calling it in the concurrent mode from the main server process can create quite a few connections to the storage, especially if there are multiple users. Also, calling it here means we have to wait till the remote server completes the deletion. I can see there is no retry logic, so if the operation fails here for some reason, the files will stay on the cloud storage. I think that maybe this call should be deferred and handled by the cleanup queue? This way, tasks will be deleted as fast they are now, and the other issues can be covered as well. Please include a performance check with 50k files in the manual tests, it's our implicit target size for performance concerns.

cvat/apps/engine/management/commands/movetaskfrombackingcs.py
```diff
@@ -0,0 +1,24 @@
...
+        task = Task.objects.get(id=task_id)
+        data = task.require_data()
+
+        if not data.supports_backing_cs():
+            raise CommandError(f"Task #{task_id} does not support backing cloud storage")
+
+        if not data.local_storage_backing_cs_id:
+            raise CommandError(f"Task #{task_id} has no backing cloud storage")
+
+        data.move_from_backing_cs()
```
zhiltsov-max, line 34: I got these numbers for a local minio and 50k frames:

```python
time python manage.py movetaskfrombackingcs 2738

real    3m46.175s
user    3m39.291s
sys     0m30.190s
```

cvat/apps/engine/management/commands/movetasktobackingcs.py
SpecLad: I'm not sure bulk transfer would be all that much more efficient than sequential. What time are we going to save? I guess it would avoid repeating the Django startup overhead, but is there anything else you have in mind?

zhiltsov-max: No, not really that much. It can help if there are many small tasks (with less than 1000 files, if I understand the batching done). I assume there's going to be a managing shell script anyway.

cvat/apps/engine/models.py
SpecLad: Thanks. I updated the code to filter on `is_placeholder=False`.

tests/python/rest_api/test_tasks.py
```diff
@@ -1601,6 +1601,52 @@ def _test_can_restore_task_from_backup(
...
+@pytest.mark.usefixtures("restore_cvat_data_per_function")
+@pytest.mark.usefixtures("restore_redis_inmem_per_function")
+class TestTaskBackupsWithBackingCs:
+    @pytest.fixture(autouse=True)
+    def setup(self, tmp_path: Path, admin_user: str) -> None:
+        self.tmp_dir = tmp_path
+        self.user = admin_user
+
+    def test_can_export_and_import_backup(self, request, cloud_storages):
+        # task_id = next(task["id"] for task in tasks if task["dimension"] == "3d")
```
zhiltsov-max: Please delete or uncomment this line

tests/python/rest_api/test_tasks.py
```diff
@@ -1601,6 +1601,52 @@ def _test_can_restore_task_from_backup(
...
+@pytest.mark.usefixtures("restore_db_per_function")
+@pytest.mark.usefixtures("restore_cvat_data_per_function")
+@pytest.mark.usefixtures("restore_redis_inmem_per_function")
+class TestTaskBackupsWithBackingCs:
+    @pytest.fixture(autouse=True)
+    def setup(self, tmp_path: Path, admin_user: str) -> None:
+        self.tmp_dir = tmp_path
+        self.user = admin_user
+
+    def test_can_export_and_import_backup(self, request, cloud_storages):
```
zhiltsov-max: Why not move this test into the `TestTaskBackups` class?

SpecLad: Yeah, I guess I can. IIRC I separated it because I initially thought `TestTaskBackups` didn't restore the DB per-function; but actually it does, so I might as well merge it back. Done.

tests/python/rest_api/test_tasks.py
SpecLad: Thanks, I missed it. Removed.

PR #10286, opened 2026-02-17
3D canvas memory optimization

klakhov:
### Motivation and context

Users working with 3D point cloud annotation tasks experience browser tab crashes after approximately 30 minutes of active use.
It seems we have memory leak that is caused by WebGL resources (BufferGeometry, Materials) not being properly disposed.

1. **WebGL Binding State Cache**: When `clearScene()` removed objects from the scene, it only cleared JavaScript references without calling Three.js `dispose()` methods
2. **Point Cloud Geometry**: Each frame creates 4 point cloud clones (one per view), but old point clouds were never disposed
3. **VAO (Vertex Array Object)**: WebGL context maintained a binding state cache

Info about dispose: [Here](https://discourse.threejs.org/t/three-js-dispose-things-so-hard/46664) and [Here](https://threejs.org/manual/?q=disp#en/how-to-dispose-of-objects)
### How has this been tested?

To reproduce I took a task with ~800mb of data(100 frames). I played frames from 0 to 100 repeatedly. Memory growth can be seen on heap snapshots (first block before fix, second block after fix):
memory-fix

### Checklist

### License

### DIFF
--- a/changelog.d/20260223_105115_sekachev.bs_fix_3d_memory.md
+++ b/changelog.d/20260223_105115_sekachev.bs_fix_3d_memory.md
@@ -1,0 +1,4 @@
+### Fixed
+
+- Memory leaks on 3D canvas because of unreleased GL resources
+  (<https://github.com/cvat-ai/cvat/pull/10268>)
--- a/cvat-canvas3d/src/typescript/cuboid.ts
+++ b/cvat-canvas3d/src/typescript/cuboid.ts
@@ -7,7 +7,7 @@ import * as THREE from 'three';
 import { OrientationVisibility, ViewType } from './canvas3dModel';
 import constants from './consts';
 import controlPointTexture from './controlPointTexture';
-import { disposeObjectResources } from './utils';
+import { disposeObject3D, disposeObjectResources } from './utils';
 
 export interface Indexable {
     [key: string]: any;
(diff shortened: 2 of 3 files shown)


cvat-canvas3d/src/typescript/canvas3dView.ts
```diff
@@ -1762,7 +1780,9 @@ export class Canvas3dViewImpl implements Canvas3dView, Listener {
 
         this.cameraSettings.perspective.position = position;
         this.sceneBBox = new THREE.Box3().setFromObject(points);
-        this.views.perspective.scene.add(points.clone());
+        const perspectiveCloud = points.clone();
```
bsekachev, line 1778: We cloned `points` three times, but it seems the original `points` objects was not disposed properly.

cvat-canvas3d/src/typescript/canvas3dView.ts
```diff
@@ -127,6 +128,9 @@ export class Canvas3dViewImpl implements Canvas3dView, Listener {
         data: DrawnObjectData;
         cuboid: CuboidModel;
     }>;
+    private pointClouds: {
+        [key in ViewType]?: THREE.Points;
+    };
```
bsekachev: Do we really need to store these point clouds in advance? Would not it be better to modify `clearScene` method.

For the `scene` we may recursively go over all children and dispose objects

bsekachev (commented):
Perhaps one more place to dispose objects is `deleteObjects` method which can be called.

And also some elements are added/removed when activating/deactivating the object (e.g. rotation/resize helpers)

cvat-canvas3d/src/typescript/canvas3dView.ts
klakhov: Removed.

cvat-canvas3d/src/typescript/canvas3dView.ts
klakhov: Well in this case `perspectiveCloud.geometry === points.geometry ` is true so it seems its shared, not cloned

klakhov:
Also added `.dispose` method for cuboids

PR #10288, opened 2026-02-17
Stop recreating the database after the REST API testsuite

SpecLad:
### Motivation and context

Frankly, I don't know why we're even doing this. It seems to only have downsides:

* It slows down testsuite execution, which is especially annoying because it happens _before_ detailed exception messages are printed.

* It erases evidence that could be used to investigate why a test failed.

The DB gets reloaded every time the testsuite starts anyway, so a dirty database should not affect any subsequent executions.

### Checklist

### License

### DIFF
--- a/tests/python/shared/fixtures/init.py
+++ b/tests/python/shared/fixtures/init.py
@@ -499,7 +499,9 @@ def local_start(
 
     wait_for_services(waiting_time)
 
-    docker_exec_cvat("python manage.py loaddata /tmp/data.json")
+    docker_exec_cvat(
+        ["sh", "-c", "./manage.py flush --no-input && ./manage.py loaddata /tmp/data.json"]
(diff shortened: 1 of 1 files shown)


SpecLad:
@zhiltsov-max What do you think?

zhiltsov-max:
> The DB gets reloaded every time the testsuite starts anyway

A little bit more precise, the startup routine populates the existing DB and it overrides only the objects ids mentioned in `data.json`. However, `--stop-services` does `docker compose down -v`, which also removes the volumes. As an alternative solution I can suggest adding the DB cleanup into the startup routine to support repeated invocations of tests without restarting the containers.

SpecLad:
Hmm, good point. I'll try to add a `manage.py flush` before the `loaddata` command.

PR #10291, opened 2026-02-18
Added a new annotation filter for rotated shapes

filichabout:
### Motivation and context

Added an internal filter for annotations related to shapes that have been changed in rotation direction.

- ~~I have updated the documentation accordingly~~
- ~~I have added tests to cover my changes~~
- ~~I have linked related issues~~

### License

### DIFF
--- a/cvat-core/src/annotations-filter.ts
+++ b/cvat-core/src/annotations-filter.ts
@@ -73,7 +73,7 @@ type ConvertedAttributes = Record<string, string | number | boolean>;
 interface ConvertedObjectData {
     width: number | null;
     height: number | null;
-    rotate: number | null;
+    rotation: number | null;
     attr: Record<string, ConvertedAttributes>;
     label: string;
     serverID: number;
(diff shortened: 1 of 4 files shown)


cvat-core/src/annotations-filter.ts
```diff
@@ -111,6 +112,7 @@ export default class AnnotationsFilter {
             return {
                 width,
                 height,
+                rotate: state.rotation ?? null,
```
bsekachev: ObjectState::rotation already returns `number | null`. So `??` seems extra here

changelog.d/20260218_131006_filich98_DEV_801_filters_upd.md
```diff
@@ -0,0 +1,5 @@
+### Added
+
+- Added an internal filter for annotations related to shapes that have been changed
+  in rotation direction.
```
bsekachev:
```suggestion
- Added a new annotation filter to find rotated bounding boxes and ellipses.
```

PR #10306, opened 2026-02-21
fix(serverless): remove Nuclio dashboard port binding

vittorio-prodomo:
## Summary

The default `docker-compose.serverless.yml` published the Nuclio dashboard port (`8070`) on the host. Because the Nuclio dashboard does not require authentication, any machine that can reach the host could access it and deploy or manage serverless functions.

Since CVAT communicates with Nuclio over the internal Docker network, the published port is not needed for normal operation. This PR removes the port binding entirely.

## Motivation

I discovered this vulnerability the hard way: an attacker exploited the exposed, unauthenticated Nuclio dashboard on my server to deploy **xmrig** cryptocurrency-mining containers (cryptojacking). Investigating the unexpected resource consumption led me to the open port and the root cause described above.

## Changes

- `components/serverless/docker-compose.serverless.yml`: removed the `ports` section from the `nuclio` service.
- `site/.../installation_automatic_annotation.md`: added an **"Accessing the Nuclio Dashboard (Advanced)"** section explaining how to re-enable the port via a compose override file, with a security warning against binding to `0.0.0.0`. Updated the troubleshooting section to use `nuctl get functions` as the primary status check instead of the dashboard.
- `site/.../serverless-tutorial.md`: updated `docker ps` output examples to reflect the removed port binding.

## Test plan

- [x] Verify CVAT can still invoke serverless functions without the published port (container-to-container communication is unaffected).
- [x] Confirm the Nuclio dashboard is **not** reachable from the host on port 8070 by default.
- [ ] Verify the documented override file (`docker-compose.serverless.override.yml`) correctly re-enables dashboard access at `localhost:8070`.

### DIFF
--- a/components/serverless/docker-compose.serverless.yml
+++ b/components/serverless/docker-compose.serverless.yml
@@ -13,8 +13,6 @@ services:
       no_proxy: ${no_proxy:-}
       NUCLIO_CHECK_FUNCTION_CONTAINERS_HEALTHINESS: 'true'
       NUCLIO_DASHBOARD_DEFAULT_FUNCTION_MOUNT_MODE: 'volume'
-    ports:
-      - '8070:8070'
     logging:
       driver: "json-file"
       options:
--- a/site/content/en/docs/administration/community/advanced/installation_automatic_annotation.md
+++ b/site/content/en/docs/administration/community/advanced/installation_automatic_annotation.md
@@ -87,10 +87,51 @@ If you did, make sure all containers are stopped by `docker compose down`.
   Important requirement: you should have the latest versions of Docker Desktop, Nvidia drivers for WSL,
   and the latest updates from the Windows Insider Preview Dev channel.
 
+**Accessing the Nuclio Dashboard (Advanced):**
+
+By default, the Nuclio dashboard port is **not published** to the host.
+CVAT communicates with Nuclio over the internal Docker network,
+so the dashboard is not required for normal operation.
+
+If you need direct access to the Nuclio dashboard
+(e.g. for debugging, manual function management, or when running Nuclio on a separate machine),
+create a `docker-compose.serverless.override.yml` file with the following contents:
+
+```yaml
+services:
+  nuclio:
+    ports:
+      - '127.0.0.1:8070:8070'
+```
+
+Then include it when starting CVAT:
+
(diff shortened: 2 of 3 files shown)


azhavoro:
@vittorio-prodomo thank you for the contribution.

In general, LGTM. However, for CVAT, there is no need to open this port at all. From my perspective, it is better to completely remove the port binding for the Nuclio dashboard.

The only case I’m aware of where this port must be open is when Nuclio is deployed on a different machine. I believe this is a more advanced setup, and users with that configuration should have enough experience to modify the compose config themselves.

That said, it would be great to mention this in the documentation. Could you please add a short note about this scenario to the docs as well?

Thank you!

vittorio-prodomo:
@azhavoro thanks for the feedback, and you're right... since CVAT talks to Nuclio over the internal Docker network, there's no reason to publish the port at all.

I've updated the PR to:

1. **Remove the `ports` section entirely** from `docker-compose.serverless.yml` instead of just binding to localhost.
2. **Added documentation** in `installation_automatic_annotation.md` with:
   - An "Accessing the Nuclio Dashboard (Advanced)" section explaining how to re-enable the port via a compose override file for users who need it (debugging, separate-machine setups, etc.).
   - A security warning against binding to `0.0.0.0`.
   - Updated the troubleshooting section to recommend `nuctl get functions` as the primary way to check function status.
3. **Updated `serverless-tutorial.md`** to reflect the removed port in `docker ps` output examples.

I also rebased onto the latest `develop`.

vittorio-prodomo:
@azhavoro do you agree with the Copilot-flagged concerns? From what I can tell, your suggestion was valid because the "nuctl --platform local" commands don't need the WebUI at all. Can you confirm the proposed changes are indeed correct and we can dismiss Copilot's comments?

azhavoro:
> @azhavoro do you agree with the Copilot-flagged concerns? From what I can tell, your suggestion was valid because the "nuctl --platform local" commands don't need the WebUI at all. Can you confirm the proposed changes are indeed correct and we can dismiss Copilot's comments?

I'm pretty sure nuctl uses docker directly to create its resources, including functions. I'll check.

azhavoro (approved):

PR #10316, opened 2026-02-25
[DEV-807] added org slug to curl that looks for existing function + better format for filter

yaizkazani:
This is needed to better search for existing functions in org workspace and private workspace. Also added some better formatting for json logic filter.

I've tested this as a new image in compose. 

- [ ] I have updated the documentation accordingly (I will write complete docs a bit later)

- [x] I submit _my code changes_ under the same [MIT License](https://github.com/cvat-ai/cvat/blob/develop/LICENSE)

### DIFF
--- a/cvat/requirements/base.txt
+++ b/cvat/requirements/base.txt
@@ -189,7 +189,7 @@ networkx==3.4.2
     # via datumaro
 nibabel==5.3.2
     # via datumaro
-nltk==3.9.1
+nltk==3.9.3
     # via datumaro
 oauthlib==3.3.1
     # via requests-oauthlib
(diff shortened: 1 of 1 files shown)


changelog.d/20260225_143512_petr.iosipov_DEV_807.md
```diff
@@ -0,0 +1,4 @@
+### Added
+
+- Added org_slug to curl function search
```
SpecLad: The buggy version hasn't been released yet, so I don't think a changelog entry is necessary here.

ai-models/agents_deployment/sam2/check_env.sh
```diff
@@ -25,6 +25,16 @@ resolve_org_slug() {
     fi
 }
 
+resolve_org_slug_curl() {
```
SpecLad: Why not merge this into `resolve_org_slug`? It tests the same condition.

PR #10319, opened 2026-02-26
[DEV-802] Helm chart to deploy agents

yaizkazani:
- [ ] I have updated the documentation accordingly (will do later as separate task)

### License


ai-models/agents_deployment/helm/templates/helpers.tpl
```diff
@@ -0,0 +1,81 @@
...
+Common environment variables
+*/}}
+{{- define "agent.commonEnv" -}}
+- name: CVAT_BASE_URL
+  value: {{ .Values.agent.cvat_base_url | default "https://app.cvat.ai" | quote }}
+{{- if .Values.agent.cvat_access_token }}
+- name: CVAT_ACCESS_TOKEN
+  value: {{ .Values.agent.cvat_access_token | quote }}
+{{- end }}
+{{- if not .Values.agent.cvat_access_token }}
```
azhavoro:
```suggestion
{{- else }}

```

ai-models/agents_deployment/helm/templates/helpers.tpl
```diff
@@ -0,0 +1,81 @@
...
+
+{{- define "agent.image" -}}
+image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+imagePullPolicy: {{ .Values.image.pullPolicy }}
+{{- end }}
+
+{{- define "agent.podSecurityContext" -}}
+runAsNonRoot: true
+runAsUser: 1000
+runAsGroup: 1000
```
SpecLad, line 79: Instead of this, could you change the `USER` instruction in the Dockerfile to `USER 1000:1000`? This would still be compatible with `runAsNonRoot`, but with less redundant configuration.

ai-models/agents_deployment/helm/templates/job.yaml
```diff
@@ -0,0 +1,39 @@
...
+      restartPolicy: Never
+      containers:
+        - name: function-registration
+          securityContext:
+            allowPrivilegeEscalation: false
+          {{- include "agent.image" . | nindent 10 }}
+          command: ["/bin/sh", "-c"]
+          args:
+            - |
+              ./function_registration.sh
```
SpecLad: This looks quite odd, why not `command: ["./function_registration.sh"]`?

ai-models/agents_deployment/helm/templates/deployment.yaml
```diff
@@ -0,0 +1,38 @@
...
+      {{- include "agent.podSecurityContext" . | nindent 8 }}
+      containers:
+        - name: {{ .Chart.Name }}
+          securityContext:
+            allowPrivilegeEscalation: false
+          {{- include "agent.image" . | nindent 10 }}
+          #NB! configMap is created by job.yaml
+          envFrom:
+            - configMapRef:
+                optional: true
```
SpecLad, line 27: Why is it optional? The agent can't start without the function ID.

yaizkazani: The pod itself is able to process missing env var and fail. I don't like much using k8s mechanisms to make pod stuck in backoff not being able to create a container. 

This actually leads to another error of release being stuck in uninstallation process. I need to change scripts a little bit. The problem now is that when you uninstall chart and configMap does not exist pre-delete hook job fails and you cannot uninstall the release (only with --no-hooks).

ai-models/agents_deployment/helm/templates/job.yaml
yaizkazani: Actually, this is literally the same, i can change :)

ai-models/agents_deployment/helm/templates/helpers.tpl
yaizkazani: In docker we have this:

RUN groupadd --gid 1000 agent && \
    useradd --uid 1000 --gid 1000 --create-home agent
    
So we use 1000:1000 already, thing is I'd like to keep this visible for user in case they are not happy to dig through dockerfile to look what user is used. I really dont mind having 2 lines of code to save users some time

ai-models/agents_deployment/helm/values.yaml
```diff
@@ -0,0 +1,49 @@
...
+  # TODO need to find out optimal recommended resources.
+  resources:
+    limits:
+      cpu: 1
+#      memory: 1.5Gi
+    requests:
+      cpu: 200m
+      memory: 256Mi
+  envVars: []
+#    - name: EXAMPLE_ENV_VAR
```
azhavoro, line 32: With this approach, it is impossible to set env variables from CM or Secret.

ai-models/agents_deployment/helm/templates/job.yaml
yaizkazani: done

ai-models/agents_deployment/helm/templates/NOTES.txt
```diff
@@ -0,0 +1,18 @@
...
+
+The function registration job runs as a pre-install/pre-upgrade hook.
+Check its status:
+  kubectl get jobs -n {{ .Release.Namespace }} -l app.kubernetes.io/instance={{ .Release.Name }}
+
+View agent logs:
+  kubectl logs -n {{ .Release.Namespace }} -l app.kubernetes.io/name={{ .Chart.Name }},app.kubernetes.io/component!=job -f
+
+{{- if .Values.agent.cvat_access_token }}
+WARNING: You are using a hardcoded CVAT access token in plain text. This token WILL be visible in the Helm release history and in pod definition.
```
SpecLad, line 15: Why are we even supporting this option then, if it's so insecure?

ai-models/agents_deployment/helm/templates/helpers.tpl
yaizkazani: Yeah, good find :)

ai-models/agents_deployment/helm/templates/NOTES.txt
yaizkazani: I'm not sure if we should force users to create a secret... I think we can let them choose in this case.

ai-models/agents_deployment/helm/values.yaml
yaizkazani: It is possible to use 

  secret_env:
    - name: CVAT_ACCESS_TOKEN
      secretName: cvat-agent
      secretKey: token
  
  
 This list will be converted to Pod's env vars with this block in helpers.tpl
 
 {{- range .Values.agent.secret_env }}
- name: {{ .name }}
  valueFrom:
    secretKeyRef:
      name: {{ .secretName }}
      key: {{ .secretKey }}
{{- end }}

ai-models/agents_deployment/helm/templates/deployment.yaml
yaizkazani: I've changed the deregistration function's logic so that it does not fail if FUNCTION_ID is missing, resulting in a graceful uninstallation of the chart. I think this is the lesser evil, as if a new function was not registered, it's OK not to panic because of a missing env var.

ai-models/agents_deployment/helm/templates/helpers.tpl
yaizkazani: Added this

ai-models/agents_deployment/helm/templates/NOTES.txt
yaizkazani: + we have warning not only in NOTES but in values as well.

SpecLad: > I'm not sure if we should force users to create a secret...

Why not? Is there any situation where that's not the preferable solution?

yaizkazani: In theory, yes; in practice, users might like to be able to avoid creating a secret. Especially since this secret is not a part of a chart. So right now it works out of the box, if we force them to use a secret, they have to create it first.

ai-models/agents_deployment/helm/values.yaml
yaizkazani: We have agreed that this is not an issue. :)

ai-models/agents_deployment/helm/templates/NOTES.txt
yaizkazani: I think we have finally reached an agreement on this :)

PR #10325, opened 2026-03-03
[DEV-810] Docker + compose for YOLO

yaizkazani:
This is the implementation of an agent running for YOLO functions in CVAT

I have tested this with YOLO26 (yolo26s.pt)

- [x] I have updated the documentation accordingly (i will do that later together with Helm)

### License


yaizkazani:
This should not be merged until I will test changes to cvat-cli command arguments generation logic.

PR #10328, opened 2026-03-03
Added some flexibility to chart to support YOLO

yaizkazani:
Just to support YOLO env vars + removed SAM2 hardcoded mentions as chart is supposed to deploy both models now


yaizkazani:
https://github.com/cvat-ai/cvat/pull/10332

PR #10332, opened 2026-03-05
[DEV-811] Make helm chart to support maps of values for agents configuration + fixed SAM2 image/compose to support this

yaizkazani:
- [x] I have updated the documentation accordingly (separate task)

### DIFF
--- a/changelog.d/20260305_151444_sekachev.bs_updated_interactors.md
+++ b/changelog.d/20260305_151444_sekachev.bs_updated_interactors.md
@@ -1,0 +1,29 @@
+### Added
+
+- Interactors can now return multiple shapes instead of a single one
+  (<https://github.com/cvat-ai/cvat/pull/10329>)
+
+- Interactors can now return a confidence attribute. If present, interaction
+  results can be filtered in the UI (<https://github.com/cvat-ai/cvat/pull/10329>)
+
+- Added UI controls to remove point and rectangle prompts when using Interactors or OpenCV tools
+  (<https://github.com/cvat-ai/cvat/pull/10329>)
+
+- Server may add text prompts to interactors as `text_prompts` field in `POST: /api/lambda/functions/<id>`
+  (<https://github.com/cvat-ai/cvat/pull/10329>)
+
+### Changed
+
+- Unified the expected interactor interface to align with the `/annotations` API and AI detector outputs
+  (<https://github.com/cvat-ai/cvat/pull/10329>)
+
+- Updated the IoG serverless function to support the new interactor interface
+  (<https://github.com/cvat-ai/cvat/pull/10329>)
+
+### Fixed
+
+- Reduced memory usage by replacing JavaScript arrays with typed arrays
+  (<https://github.com/cvat-ai/cvat/pull/10329>)
+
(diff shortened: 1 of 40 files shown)


ai-models/agents_deployment/helm/values.yaml
```diff
@@ -1,26 +1,54 @@
...
+        value: ""
+    sam2:
+      model_id:
+        type: str
+        value: "facebook/sam2.1-hiera-tiny"
+
+  # Please pick a preset for your agent's model configuration. The preset will automatically fill in the necessary parameters for the chosen model type. If you choose a preset, you can still override specific parameters in the `modelParams` section below.
+  # Keys with empty 'value' will be ignored.
+  preset: sam2
+  modelParams: { }
```
azhavoro:
```suggestion
  modelParamsOverride: { }
```

According to the description, it seems that it would be better to explicitly use **override** here.

ai-models/agents_deployment/sam2/check_env.sh
```diff
@@ -27,12 +27,18 @@ resolve_org_slug() {
...
-        echo "Warning: MODEL_ID environment variable not found. Default is facebook/sam2.1-hiera-tiny"
-        MODEL_ID="facebook/sam2.1-hiera-tiny"
+
+resolve_model_params() {
+    if [ -z "$MODEL_CONFIG_PARAMS" ]; then
+        echo "Warning: MODEL_CONFIG_PARAMS environment variable not found. Default model will be used facebook/sam2.1-hiera-tiny"
+        MODEL_CONFIG_PARAMS="-p model_id=str:facebook/sam2.1-hiera-tiny"
+    elif result=$(echo "$MODEL_CONFIG_PARAMS" | grep -oP 'model_id=str:\K[^ ]+'); then
+        echo "Extracted MODEL_ID from MODEL_CONFIG_PARAMS: $result"
+        export MODEL_ID="$result"
```
azhavoro, line 37: It seems that the `MODEL_ID` env var is no longer necessary.

ai-models/agents_deployment/helm/values.yaml
```diff
@@ -1,26 +1,54 @@
...
+        type: str
+        value: ""
+    sam2:
+      model_id:
+        type: str
+        value: "facebook/sam2.1-hiera-tiny"
+
+  # Please pick a preset for your agent's model configuration. The preset will automatically fill in the necessary parameters for the chosen model type. If you choose a preset, you can still override specific parameters in the `modelParams` section below.
+  # Keys with empty 'value' will be ignored.
+  preset: sam2
```
azhavoro, line 43: Can I choose not to select a preset at all?

ai-models/agents_deployment/sam2/check_env.sh
yaizkazani: image
In the entrypoint 🤣

ai-models/agents_deployment/helm/values.yaml
yaizkazani: Yes, i have added empty "custom" preset. If you select it you can just use modelParamsOverride

PR #10312, opened 2026-02-24
Export resource list as `csv`

klakhov:
### Motivation and context

The feature allows to export current list of Projects/Tasks/Jobs as .csv file via button in filters UI:
image

### Checklist

- [x] I submit my changes into the `develop` branch
- [x] I have created a changelog fragment 
- ~~[ ] I have updated the documentation accordingly~~
- [x] I have added tests to cover my changes
- [x] I have linked related issues (see [GitHub docs](
  https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword))

### License

- [x] I submit _my code changes_ under the same [MIT License](
  https://github.com/cvat-ai/cvat/blob/develop/LICENSE) that covers the project.
  Feel free to contact the maintainers if that's a concern.

PR #10345, opened 2026-03-09
Fix: always escape sting in .csv

klakhov:
### Motivation and context

In the previous CSV patch (#10312), the CSV export logic was refactored. One subtle change was introduced: only strings containing special characters (quotes, commas, etc.) were escaped.

According to the **RFC 4180 CSV standard**:

* Fields containing commas, quotes, or line breaks **must** be quoted.
* Other fields **may** also be quoted.

However, this behavior is not consistent with our previous implementation, where all string values were quoted.

This patch restores the previous behavior by ensuring that all string fields are quoted, keeping the export logic consistent with what we had before.

### Checklist

### License

### DIFF
--- a/changelog.d/20260309_164944_klakhov_hotfix_csv.md
+++ b/changelog.d/20260309_164944_klakhov_hotfix_csv.md
@@ -1,0 +1,4 @@
+### Fixed
+
+- Missing escaping for string fields in `.csv` export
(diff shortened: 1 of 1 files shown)


zhiltsov-max:
Is there a reason why we are not using any library to work with CSV?

PR #10358, opened 2026-03-11
Add dynamic nuclio version to cvat docs

archibald1418:
### Motivation and context

https://docs.cvat.ai/docs/administration/community/advanced/installation_automatic_annotation/ 
says that it is important that `nuclio` version has to be the same as in `docker-compose.serverless.yml`, but, ironically, contains a link for the older version: `1.15.9` in`.yaml` vs `1.13.0` in the docs.

- Added a Hugo shortcode`verison-link.html` and a script that parses the `compose*.yaml` file for the nuclio version and places it into a `versions.yaml` to be read by Hugo. 
- Added a preview step for docs

### How has this been tested?

The docs get built and the text contains the correct link

### Checklist

### License

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -64,7 +64,7 @@ jobs:
         uses: actions/upload-artifact@v4
         with:
           name: docs-site
-          path: ./public
+          path: ./public/docs
           retention-days: 1
 
       - name: Deploy


.github/workflows/docs.yml
```diff
@@ -45,6 +45,10 @@ jobs:
         run: |
           npm ci
 
+      - name: Sync versions in docs
```
azhavoro, line 48: I don't like the idea of running additional scripts, since those are steps that are easy to forget.
Besides, I'm not sure this approach will work correctly for all CVAT versions we're documenting; most likely, the same nuclio version will be rendered.

site/layouts/shortcodes/version-link.html
```diff
@@ -0,0 +1,3 @@
+{{- $name := .Get "name" -}}
+{{- $tool := index .Site.Data.versions $name -}}
+<a href="{{ $tool.url }}/{{ $tool.version }}">{{ .Inner }}</a>
```
azhavoro, line 3:
```suggestion
{{ $yaml := resources.Get "cvat_components/serverless/docker-compose.serverless.yml" }}
{{ with $yaml }}
    {{ $yaml_data := unmarshal .Content }}
    {{ $nuclio_version_full := index (strings.Split $yaml_data.services.nuclio.image ":" ) 1 }}
    {{ $nuclio_version := index (strings.Split $nuclio_version_full "-" ) 0 }}
    <a href="https://github.com/nuclio/nuclio/releases/tag/{{ $nuclio_version }}">{{ $.Inner }}</a>
{{ end }}
```

I would suggest something like this, which would avoid the need for external scripts
In addition, you will need to modify the `config.toml` file in this case
```yaml
[module]
  [[module.mounts]]
    files = ["docker-compose.serverless.yml"]
    source = "../components/serverless/"
    target = "assets/cvat_components/serverless"
```

PR #10387, opened 2026-03-17
Refactor Rego filters WRT organizations

SpecLad:
### Motivation and context

This changes how filters are defined in Rego rules in two major ways:

* Instead of reverse polish notation, filters are now combined by using List-like prefix notation: `[foo, bar, "|", baz, "|"]` is now `["|", foo, bar, baz]`. Subjectively, I find this more readable, especially in complex expressions, because it's easier to see which operands correspond to which operator. It also makes it easier to combine an arbitrary amount of filters using Rego (it's just `array.concat([op], filters)`), which enables the second change:

* Instead of manually writing out filters that check that a resource belongs to an organization, add a generic function, `add_organization_filter`, that automatically adds the necessary filters given the list of organization ID lookups. This eliminates a lot of duplicate logic.

### Checklist

### License

### DIFF
--- a/cvat/apps/events/rules/events.rego
+++ b/cvat/apps/events/rules/events.rego
@@ -55,8 +55,6 @@ allow if {
 }
 
 
-filter := utils.add_organization_filter(base_filter, ["org_id"])
-
 base_filter := {} if {
     utils.is_admin
 } else := qobject if {
@@ -72,3 +70,5 @@ base_filter := {} if {
     organizations.has_perm(organizations.WORKER)
     qobject := {"user_id": input.auth.user.id}
 }
+
+filter := utils.add_organization_filter(base_filter, ["org_id"])
--- a/cvat/apps/consensus/rules/consensus_settings.rego
+++ b/cvat/apps/consensus/rules/consensus_settings.rego
@@ -57,11 +57,6 @@ allow if {
     organizations.is_member
 }
 
-filter := utils.add_organization_filter(base_filter, [
-    "task__organization",
-    "task__project__organization",
-])
-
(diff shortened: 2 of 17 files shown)


cvat/apps/events/rules/events.rego
```diff
@@ -55,24 +55,20 @@ allow if {
 }
 
 
-filter := [] if {
-    utils.is_admin
-    utils.is_sandbox
-} else := qobject if {
+filter := utils.add_organization_filter(base_filter, ["org_id"])
```
zhiltsov-max: Even though these declarations seem to work, I'd suggest moving `filter` after `base_filter` for clarity.

cvat/apps/iam/rules/utils.rego
```diff
@@ -101,3 +101,16 @@ is_sandbox if {
...
 }
+
+add_organization_filter(base_filter, organization_fields) := base_filter if is_sandbox
+
+add_organization_filter(base_filter, organization_fields) := qobject if {
+    is_organization
+    qobject := ["&", base_filter, make_organization_filter(organization_fields)]
+}
+
+make_organization_filter(organization_fields) := qobject if {
```
zhiltsov-max, line 112: `organization_fields` can even be forwarded from the `view.iam_organization_field` in the future.

SpecLad: My plan is actually to remove `iam_organization_field`, making the Rego files the single source of truth.

cvat/apps/events/rules/events.rego
SpecLad: Alright, done.

PR #10388, opened 2026-03-18
[DEV-815] Added compose for transformers + added Helm feature to support function name

yaizkazani:
- [x] I have updated the documentation accordingly (separate task)


ai-models/agents_deployment/helm/templates/helpers.tpl
```diff
@@ -94,3 +94,16 @@ Configure command arguments for cvat-cli
...
+Configure function name for agent
+*/}}
+{{- define "agent.functionNameEnv" -}}
+{{- if .Values.agent.function_name }}
+- name: FUNCTION_NAME
+  value: {{ .Values.agent.function_name | quote }}
+{{- else }}
+- name: FUNCTION_NAME
+  value: "MyAgentFunction"
+{{- end }}
```
azhavoro:
```suggestion
- name: FUNCTION_NAME
{{- if .Values.agent.function_name }}
  value: {{ .Values.agent.function_name | quote }}
{{- else }}
  value: "MyAgentFunction"
{{- end }}
```

azhavoro:
Generally fine, but please wait until I've tested the PR locally

ai-models/agents_deployment/helm/templates/helpers.tpl
yaizkazani: Indeed this is better :)

PR #10389, opened 2026-03-18
Fix #10331: persist cloud storage link after task data import

### LINKED ISSUE #10331: No direct link between Task/Project and Cloud Storage after data import

### Steps to Reproduce

1. Create Test Project
2.  Create Test Task and attach files from connected cloud storage

3. Check Project/Task After create and annotation
# Project response shows:
project_info, data = self.client.api_client.projects_api.retrieve(project_id)
{'assignee': None,
 'assignee_updated_date': None,
 'bug_tracker': '',
 'created_date': datetime.datetime(2026, 2, 20, 11, 42, 1, 195628, tzinfo=tzutc()),
 'dimension': '2d',
 'guide_id': None,
 'id': 608,
 'labels': {'url': 'https://cvat/api/labels?project_id=608'},
 'name': 'test_project',
 'organization': None,
 'owner': {'first_name': '',
           'id': 93,
           'last_name': '',
           'url': 'https://cvat/api/users/93',
           'username': 'v'},
 'source_storage': {'cloud_storage_id': None, 'id': 4321, 'location': 'local'},
 'status': 'annotation',
 'target_storage': {'cloud_storage_id': None, 'id': 4322, 'location': 'local'},
 'task_subsets': [],
 'tasks': {'count': 1,
           'url': 'https://cvat/api/tasks?project_id=608'},
 'updated_date': datetime.datetime(2026, 3, 4, 15, 0, 19, 245976, tzinfo=tzutc()),
 'url': 'https://cvat/api/projects/608'}
 
# Task response shows:
task = client.tasks.retrieve(task.id)
{'assignee': {'first_name': '',
              'id': 53,
              'last_name': '',
              'url': 'https://cvat/api/users/53',
              'username': ''},
 'assignee_updated_date': datetime.datetime(2026, 2, 27, 6, 35, 35, 232249, tzinfo=tzutc()),
 'bug_tracker': '',
 'created_date': datetime.datetime(2026, 2, 26, 9, 44, 18, 189227, tzinfo=tzutc()),
 'data': 5110,
 'data_chunk_size': 36,
 'data_compressed_chunk_type': 'imageset',
 'data_original_chunk_type': 'imageset',
 'dimension': '2d',
 'guide_id': None,
 'id': 5155,
 'image_quality': 70,
 'jobs': {'completed': 0,
          'count': 1,
          'url': 'https://cvat/api/jobs?task_id=5155',
          'validation': 0},
 'labels': {'url': 'https://cvat/api/labels?task_id=5155'},
 'mode': 'annotation',
 'name': 'test task',
 'organization': None,
 'overlap': 0,
 'owner': {'first_name': '',
           'id': 93,
           'last_name': '',
           'url': 'https://cvat/api/users/93',
           'username': 'v.'},
 'project_id': 608,
 'segment_size': 398,
 'size': 398,
 'source_storage': {'cloud_storage_id': None, 'id': 4329, 'location': 'local'},
 'status': 'annotation',
 'subset': '',
 'target_storage': {'cloud_storage_id': None, 'id': 4330, 'location': 'local'},
 'updated_date': datetime.datetime(2026, 2, 27, 18, 47, 44, 936206, tzinfo=tzutc()),
 'url': 'https://cvat/api/tasks/5155',
 'validation_mode': None}

# Cloud response shows:
cloud_storage_info, _ = self.client.api_client.cloudstorages_api.retrieve(254)
{'created_date': datetime.datetime(2026, 2, 24, 13, 19, 52, 147147, tzinfo=tzutc()),
 'credentials_type': 'KEY_SECRET_KEY_PAIR',
 'description': '',
 'display_name': 'test_cloud',
 'id': 254,
 'manifests': [],
 'organization': None,
 'owner': {'first_name': '',
           'id': 498,
           'last_name': '',
           'url': 'https://cvat/api/users/498',
           'username': 'n.'},
 'provider_type': 'AWS_S3_BUCKET',
 'resource': 'test_cloud',
 'specific_attributes': 'endpoint_url=https%3A%2F%2Fstorage.yandexcloud.net',
 'updated_date': datetime.datetime(2026, 2, 24, 13, 19, 52, 211632, tzinfo=tzutc())}, 

### Expected Behavior

When importing data from a cloud storage to a task, there is no persistent link between the task/project and the source cloud storage. After successful import, the source_storage field in both project and task objects shows location: 'local' with an internal storage ID that doesn't reference the original cloud storage.

The task and project should maintain a reference to the original cloud storage after import. Expected source_storage field:
'source_storage': {
    'cloud_storage_id': 254,  # Original cloud storage ID
    'location': 'cloud_storage'
}

### Possible Solution

_No response_

### Context

Image

After import, source_storage always shows location: 'local' with internal storage IDs (4321, 4322, 4329, 4330 and so on) that have no relation to the original cloud storage ID (254) - despite confirming that the data was explicitly imported from the cloud storage during task creation. 
This creates several issues:

No traceability - Cannot determine which cloud storage contains the original data

Multiple storage entries - Each task/project creates new internal storage entries with different IDs

Lost connection - No way to programmatically find related cloud storage for a task

### Environment

```Markdown
Name: cvat_sdk
Version: 2.24.0
```

DuarteCruz6:
When creating a task (and project task) from cloud storage, CVAT did not persist the originating cloud storage in `source_storage`, leaving it as local/empty and breaking traceability and defaults.

Persist the cloud storage link by setting Task/Project `source_storage` to a reusable cloud-backed Storage record when cloud data is used, and reuse Storage rows to avoid ID churn. Add REST API regression coverage.

### Motivation and context
Fixes #10331 

Currently, when creating a task (and project task) and importing data from an attached Cloud Storage, CVAT successfully fetches the data but fails to persist the originating cloud storage in the `source_storage` field of the Task/Project. Instead, it drops the reference and defaults back to local with a null cloud storage ID. This breaks data traceability and prevents expected default behaviors for cloud-backed pipelines.

### Changes described in detail
This PR introduces a fix in the engine's data import pipeline (cvat/apps/engine/views.py, cvat/apps/engine/task.py, cvat/apps/engine/models.py, and cvat/apps/engine/serializers.py).

- Storage Record Reusability (models.py & serializers.py): Added a get_or_create_cloud_storage_link classmethod to reuse existing cloud-backed Storage database records. The serializer now intercepts storage creation; if the payload specifies cloud storage, it uses this link to prevent database ID churn and bloat instead of creating new instances.

- Cloud Storage Persistence (views.py & task.py): When data is attached using a cloud_storage_id, the backend now explicitly checks the Task and associated Project storage configurations. If the `source_storage` is still set to the default local location, it updates it to point to the originating cloud storage link.

- Garbage Collection: Implemented logic to safely delete the old local storage records if they become completely orphaned (i.e., not used by any other Tasks or Projects as a source or target).

#### Manual Testing:
- Deployed a local CVAT instance with an attached Cloud Storage.
- Created a new Project and a new Task via the web UI.
- Imported image data directly from the Cloud Storage tab.
- Verified via GET /api/tasks/{id} and GET /api/projects/{id} that source_storage correctly reflects location: "cloud_storage" and the correct cloud_storage_id, rather than reverting to local.

#### Automated Testing:
- Added REST API regression coverage (tests/python/rest_api/test_task_data.py) to verify that creating a task in a project using cloud storage files correctly mutates and persists the source_storage for both the task and the parent project.
- Ran the existing test suite (pytest) to ensure standard local file uploads and explicit storage assignments remain unaffected.

### Checklist
- [X] I have created a changelog fragment
- [ ] ~~I have updated the documentation accordingly~~ (Not applicable: Backend bug fix, no public API schema changes)
- [X] I have added tests to cover my changes
- [X] I have linked related issues

### License

- [X] I submit _my code changes_ under the same [MIT License](https://github.com/cvat-ai/cvat/blob/develop/LICENSE) that covers the project.

### DIFF
--- a/changelog.d/20260318_162225_duarte.cruz_issue10331.md
+++ b/changelog.d/20260318_162225_duarte.cruz_issue10331.md
@@ -1,0 +1,2 @@
+### Fixed
+- Dropped cloud storage links reverting to local storage when importing data into tasks and projects (#10331)
--- a/cvat/apps/engine/models.py
+++ b/cvat/apps/engine/models.py
@@ -354,6 +354,13 @@ class Storage(models.Model):
     class Meta:
         default_permissions = ()
 
+    @classmethod
+    def get_or_create_cloud_storage_link(cls, cloud_storage: CloudStorage) -> "Storage":
+        storage, _ = cls.objects.get_or_create(
+            location=Location.CLOUD_STORAGE,
+            cloud_storage=cloud_storage,
+        )
+        return storage
 
 class ValidationMode(str, Enum):
     GT = "gt"
--- a/cvat/apps/engine/serializers.py
+++ b/cvat/apps/engine/serializers.py
@@ -3960,9 +3960,14 @@ def _configure_related_storages(validated_data: dict[str, Any]) -> dict[str, mod
                 not models.CloudStorage.objects.filter(id=cloud_storage_id).exists()
             ):
                 raise serializers.ValidationError(f'The specified cloud storage {cloud_storage_id} does not exist.')
-            storage_instance = models.Storage(**storage_conf)
-            storage_instance.save()
-            storages[i] = storage_instance
+            if storage_conf.get("location") == models.Location.CLOUD_STORAGE and cloud_storage_id:
+                storages[i] = models.Storage.get_or_create_cloud_storage_link(
+                    models.CloudStorage.objects.get(id=cloud_storage_id)
+                )
+            else:
+                storage_instance = models.Storage(**storage_conf)
+                storage_instance.save()
+                storages[i] = storage_instance
     return storages
 
 class AssetReadSerializer(WriteOnceMixin, serializers.ModelSerializer):
--- a/cvat/apps/engine/task.py
+++ b/cvat/apps/engine/task.py
@@ -1155,6 +1155,48 @@ def create_thread(
     )
 
     db_data = db_task.require_data()
+
+    if db_data.cloud_storage:
+        cloud_link_storage = models.Storage.get_or_create_cloud_storage_link(db_data.cloud_storage)
+
+        def _is_default_local_storage(storage: models.Storage | None) -> bool:
+            return storage is None or (
+                storage.location == models.Location.LOCAL and storage.cloud_storage_id is None
+            )
+
+        old_task_source_storage = db_task.source_storage
+        if _is_default_local_storage(old_task_source_storage):
+            db_task.source_storage = cloud_link_storage
+            db_task.save(update_fields=["source_storage"])
+
+            if (
+                old_task_source_storage
+                and old_task_source_storage.location == models.Location.LOCAL
+                and old_task_source_storage.cloud_storage_id is None
+                and not models.Task.objects.filter(source_storage=old_task_source_storage).exists()
+                and not models.Project.objects.filter(source_storage=old_task_source_storage).exists()
+                and not models.Task.objects.filter(target_storage=old_task_source_storage).exists()
+                and not models.Project.objects.filter(target_storage=old_task_source_storage).exists()
+            ):
+                old_task_source_storage.delete()
+
+        if db_project := db_task.project:
+            old_project_source_storage = db_project.source_storage
+            if _is_default_local_storage(old_project_source_storage):
+                db_project.source_storage = cloud_link_storage
+                db_project.save(update_fields=["source_storage"])
+
+                if (
+                    old_project_source_storage
+                    and old_project_source_storage.location == models.Location.LOCAL
+                    and old_project_source_storage.cloud_storage_id is None
+                    and not models.Task.objects.filter(source_storage=old_project_source_storage).exists()
+                    and not models.Project.objects.filter(source_storage=old_project_source_storage).exists()
+                    and not models.Task.objects.filter(target_storage=old_project_source_storage).exists()
+                    and not models.Project.objects.filter(target_storage=old_project_source_storage).exists()
+                ):
+                    old_project_source_storage.delete()
+
     upload_dir = (
         db_data.get_upload_dirname()
         if db_data.storage != models.StorageChoice.SHARE
--- a/cvat/apps/engine/views.py
+++ b/cvat/apps/engine/views.py
@@ -1116,6 +1116,47 @@ class TaskViewSet(viewsets.GenericViewSet, mixins.ListModelMixin,
                 if db_data.cloud_storage:
                     self._object.data.storage = StorageChoice.CLOUD_STORAGE
                     self._object.data.save(update_fields=['storage'])
+                    cloud_link_storage = models.Storage.get_or_create_cloud_storage_link(
+                        db_data.cloud_storage
+                    )
+
+                    def _is_default_local_storage(storage: models.Storage | None) -> bool:
+                        return storage is None or (
+                            storage.location == models.Location.LOCAL and storage.cloud_storage_id is None
+                        )
+
+                    old_task_source_storage = self._object.source_storage
+                    if _is_default_local_storage(old_task_source_storage):
+                        self._object.source_storage = cloud_link_storage
+                        self._object.save(update_fields=["source_storage"])
+
+                        if (
+                            old_task_source_storage
+                            and old_task_source_storage.location == models.Location.LOCAL
+                            and old_task_source_storage.cloud_storage_id is None
+                            and not Task.objects.filter(source_storage=old_task_source_storage).exists()
+                            and not Project.objects.filter(source_storage=old_task_source_storage).exists()
+                            and not Task.objects.filter(target_storage=old_task_source_storage).exists()
+                            and not Project.objects.filter(target_storage=old_task_source_storage).exists()
+                        ):
+                            old_task_source_storage.delete()
+
+                    if db_project := self._object.project:
+                        old_project_source_storage = db_project.source_storage
+                        if _is_default_local_storage(old_project_source_storage):
+                            db_project.source_storage = cloud_link_storage
+                            db_project.save(update_fields=["source_storage"])
+
+                            if (
+                                old_project_source_storage
+                                and old_project_source_storage.location == models.Location.LOCAL
+                                and old_project_source_storage.cloud_storage_id is None
+                                and not Task.objects.filter(source_storage=old_project_source_storage).exists()
+                                and not Project.objects.filter(source_storage=old_project_source_storage).exists()
+                                and not Task.objects.filter(target_storage=old_project_source_storage).exists()
+                                and not Project.objects.filter(target_storage=old_project_source_storage).exists()
+                            ):
+                                old_project_source_storage.delete()
                 if 'stop_frame' not in serializer.validated_data:
                     # if the value of stop_frame is 0, then inside the function we cannot know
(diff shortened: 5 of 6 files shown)


DuarteCruz6:
Hi @zhiltsov-max, @SpecLad, @nmanovic, just pinging on this for a review when you have a chance! Let me know if you need me to make any adjustments or if there's anything else I can do to help move this forward.

prabaljainn:
Please have a look at this and do the needful, the issue is right

zhiltsov-max:
Source and target storages are supposed for annotations. The attached data storage is available in the `data_cloud_storage_id` response field of the `/api/tasks/<id>` endpoint. Does it work for you?

DuarteCruz6:
Thanks for the clarification, @zhiltsov-max! I checked ```GET /api/tasks/{id}``` and confirmed that ```data_cloud_storage_id``` does indeed provide the attached data origin.

That said, ```source_storage``` and ```target_storage``` are highly confusing in practice. Both the original issue reporter and I (along with another person on this PR) interpreted them as the raw data origin, not annotation I/O defaults. The current naming makes that misinterpretation very easy.

To prevent this confusion for future users and API consumers, I suggest two improvements:
1. Rename or clearly re-label ```source_storage``` and ```target_storage``` to reflect that they are strictly annotation import/export defaults.
2. Expose original data storage more directly: Keep ```task.data_cloud_storage_id``` as the source of truth. Consider adding a read-only field on jobs (e.g., ```task_data_cloud_storage_id```) so clients don’t need to make an extra fetch to the task endpoint just to find out where their job's media lives.

If you agree this is the best path forward, I can just repurpose this PR to revert my previous backend changes and focus strictly on the labeling, and adding that explicit read-only field. Let me know what you think!

zhiltsov-max:
I agree that the source/target storage feature lacks proper documentation now and could be explained better to the users.  So, I'd recommend adding more information about this in the project documentation and improving the API documentation instead to avoid confusion. 
- For the project documentation, it's mentioned in the [Advanced options](https://docs.cvat.ai/docs/workspace/tasks-page/#advanced-configuration) section. I think, it could be discussed in more detail on a separate page, maybe together with cloud storages in general. 
- For the API documentation, the request and response fields have no explanations at all. This could also be improved.

DuarteCruz6:
That sounds great, @zhiltsov-max. I can definitely pivot this PR to focus on the project and API documentation to clear up the ```source_storage``` / ```target_storage``` confusion.

Before I do, I just wanted to circle back to the idea of exposing ```data_cloud_storage_id``` directly on the Job response (read-only). Currently, if a client is iterating through Jobs and needs to know the data origin, they have to make a fetch to the Tasks endpoint. Would you be open to including that field alongside the doc updates, or would you prefer I leave the API payloads strictly alone for now?

zhiltsov-max:
Could you add more information on the use cases for this field in a job response? Why it is more useful than requesting a task? This information is related to the task.

DuarteCruz6:
You make a totally fair point, it is absolutely task-level data. My reasoning for exposing it (read-only) on the Job response is to prevent the N+1 API query problem for downstream automation pipelines that interact specifically with Jobs.

Here are the two main CVAT-specific use cases where having this on the Job response prevents significant API overhead:

**1. CVAT Webhook Automation (`update:job` events)**
Many teams configure CVAT Webhooks to fire when a job's state changes to completed. The JSON payload sent by the webhook is the serialized Job object. If the external receiving service (like an MLOps pipeline) needs to automatically export the annotations and route them back to the specific cloud storage where the raw data originated, it currently cannot do so with just the webhook payload. The service is forced to make a synchronous REST API call back to `GET /api/tasks/{task_id}` just to retrieve that single `data_cloud_storage_id` integer before it can proceed.

**2. Polling via the CVAT Python SDK / REST API**
When writing scripts to bulk-process finished work, it's standard practice to query `GET /api/jobs?state=completed` (or use `api_client.jobs_api.list()` in the SDK). If the goal is to sync the output of those jobs back to their respective origin cloud storages, the script must iterate through the returned jobs and execute `api_client.tasks_api.retrieve(job.task_id)` for every unique task represented in the list. This creates a N+1 query bottleneck.

I noticed that the Job API response already includes fields like `project_id` and `labels`. While those are also technically parent-level data, they are included on the Job to make the payloads more self-contained for clients. `task_data_cloud_storage_id` would serve the exact same convenience purpose for data routing.

However, if you feel this crosses the line into cluttering the Job serializer or violates strict normalization boundaries, I completely respect that architectural decision! I am perfectly happy to drop it and just focus this PR entirely on the documentation updates. Let me know what you think!

zhiltsov-max:
Thank you for providing information. Please find the comments below.

> The service is forced to make a synchronous REST API call back to GET /api/tasks/{task_id} just to retrieve that single data_cloud_storage_id integer before it can proceed.

I see you point. However, how inclusion of the data cloud storage id would improve the situation? I assume, even if the field was included, the MLOps service would still need to query CVAT to get information about the storage - it's bucket name or endpoint address - via the cloud storage API. It is also possible that there is some DB with the storage ids on the MLOps side already, but in this case why not store the storage info for task ids as well?

Based on the description, I think that the first use case can actually be a good fit for the target storage feature. Basically, if you configure such a storage for the task, you can export annotations to the cloud storage directly from CVAT. You can export at the project, task, and job level independently.

For the second use case, I can also see the issue with added task requests for each job. But there's still the same question of how the inclusion of the storage id would help with the problem. If the point is to perform a bulk operation with jobs or tasks, I think you should check if you can fetch the objects using a filter query instead. for example, if you need to fetch tasks for a number of jobs, you could use the filter like this (using CVAT Python SDK):

```python
from cvat_sdk import make_client
from cvat_sdk.core.helpers import get_paginated_collection

with make_client("https://app.cvat.ai", ...) as client:
    job_list_kwargs = {...}
    jobs = get_paginated_collection(
        client.jobs.api.list_endpoint,
        **job_list_kwargs,
        return_json=True,
        page_size=500,
    )

    task_list_kwargs = {
        "filter": '{"and":[{"<=":[%d,{"var":"task_id"},%d]}]}' % (
            min(j["task_id"] for j in jobs),
            max(j["task_id"] for j in jobs),
        )
    }
    tasks = get_paginated_collection(
        client.tasks.api.list_endpoint,
        **task_list_kwargs,
        return_json=True,
        page_size=100,
    )
```

There are many available filters - e.g. the project id, project name, task name, organization etc., depending on the endpoint.

DuarteCruz6:
You make excellent points across the board, @zhiltsov-max. Thank you for the code snippet, batch filtering definitely cleanly solves the N+1 issue for SDK polling!

Just to clarify my original thought on the webhook: while the downstream service still needs to query the `/cloudstorages` endpoint for credentials, having the ID directly on the Job payload does cut that specific chain from two synchronous API calls down to one.

However, your larger point is totally correct, meaning that if users are properly leveraging `target_storage` to handle those cloud exports natively, that custom script overhead goes away entirely anyway! That highlights exactly why this documentation update is so necessary.

I am completely happy to drop the API payload change and focus strictly on the docs. If you agree, I can just keep this PR open, revert my backend changes, and repurpose it entirely for the documentation and OpenAPI schema updates. Let me know what you think!

DuarteCruz6:
Hi @zhiltsov-max, just gently bumping this! If you are good with the plan above to repurpose this PR for the docs/OpenAPI updates, let me know and I will get those commits pushed.

PR #10400, opened 2026-03-20
Move organization filtering entirely to Rego policy files

SpecLad:
### Motivation and context

There are currently two places in the codebase that do very similar things:

1. `add_organization_filter` in Rego rules.
2. `OrganizationFilterBackend`.

Both of them apply filters to list requests based on the current organization context, except that (1) does it only when access is performed in the context of an organization, whereas (2) also handles the case when access is performed in the personal workspace, and this is specified explicitly using one of the organization parameters (`org` in the query string, etc.)

There is obviously some duplication of logic here, and this patch resolves that by making the Rego rules solely responsible. This permits us to be more flexible when it comes to determining which resources should be visible in an organization context (although for most resources the default `add_organization_filter` logic should be sufficient).

### Checklist

### License

### DIFF
--- a/changelog.d/20260403_171717_roman_rm_organization_filter.md
+++ b/changelog.d/20260403_171717_roman_rm_organization_filter.md
@@ -1,0 +1,6 @@
+### Changed
+
+- CVAT now verifies that the filters defined in Rego policy files
+  call `add_organization_filter` when the corresponding object belongs
+  to an organization
+  (<https://github.com/cvat-ai/cvat/pull/10400>)


cvat/apps/iam/filters.py
```diff
@@ -34,75 +31,14 @@
...
-            return queryset.filter(query).distinct()
-
+        # This filter only exists to add the organization parameters to the schemas
+        # of all endpoints that supports them. The actual filtering logic must be implemented
+        # in the Rego policy files for each endpoint using the add_organization_filter function.
         return queryset
 
     def get_schema_operation_parameters(self, view):
-        if not view.iam_organization_field or view.detail:
+        if not getattr(view, "iam_supports_organization_params", False) or view.detail:
```
zhiltsov-max: I think this field should be considered mandatory in the viewset, because it directly affects the security. So that the viewsets that do not support organizations specified this explicitly.

cvat/apps/iam/filters.py
```diff
@@ -34,75 +31,14 @@
...
-
-        if visibility:
-            org_id = visibility.pop("organization")
-            query = self._construct_filter_query(view.iam_organization_field, org_id)
-
-            return queryset.filter(query).distinct()
-
+        # This filter only exists to add the organization parameters to the schemas
+        # of all endpoints that supports them. The actual filtering logic must be implemented
+        # in the Rego policy files for each endpoint using the add_organization_filter function.
```
zhiltsov-max: I can see the issue that if a rego file misses the organization filtering logic, the code doesn't know about this, so doesn't complain, and we can only find this either in rego tests or integration tests. I think there should be some validation logic for the rego rules of a viewset. For example, there could be a `manage.py` command or a new test checking that if a viewset says it supports organization filtering, the rego file contains the organization filter variable (possibly, empty, so that the filter could be modified).

cvat/apps/iam/filters.py
SpecLad: Okay, done.

cvat/apps/iam/filters.py
SpecLad: Yeah, it's a good point. I have added some validation to ensure that the code crashes if an organization filter isn't applied; take a look.

cvat/apps/iam/permissions.py
```diff
@@ -259,14 +276,22 @@ def parse_filter(expr):
                     return ~Q(pk__in=[])
                 case {}:
                     return Q(**expr)
+                case "org_filter_proof":
+                    add_org_filter_proof = True
+                    return ~Q(pk__in=[])
```
zhiltsov-max, line 281: Weirdly, using `Q(Value(True))` here and in the case above results in a duplicated `True` in the SQL. For `~Q(pk__in=[])`, I can see Django automatically produces just 1 `True` in the SQL.

zhiltsov-max (approved):
LGTM. Maybe add a changelog migration note about the org filter proof.

PR #10403, opened 2026-03-23
Fix #10297: dispatch canvas.zoom event on double-click annotation zoom

### LINKED ISSUE #10297: Zooming-in onto an annotation increases issue text size

### Steps to Reproduce

1. Create a task
2. Go to review
3. Create an issue
4. Navigate to an annotation nearby, so that zooming in would include the issue short description
5. Zoom in onto the annotation using a double click

The issue text gets scaled up. Normal scaling with the mouse scroll doesn't do this.

Fit to size:
Image

Zoomed:
Image

### Expected Behavior

Scaling behavior is consistent between different ways to scale.

### Possible Solution

_No response_

### Context

_No response_

### Environment

```Markdown

```

DanielBorgs:
### Motivation and context

Fixes #10297

When zooming into an annotation via double-click, issue labels shown in review mode would incorrectly scale up with the canvas transform. This happened because the onFocusRegion method in canvasView.ts was not dispatching the canvas.zoom custom event after zooming, unlike the scroll wheel zoom handler which already dispatched it correctly.

The canvas.zoom event is used by the UI to recalculate and counteract the CSS transform applied to issue labels, keeping their font size constant regardless of zoom level. Since onFocusRegion never fired this event, the labels would scale up visually when using double-click zoom.

The fix dispatches canvas.zoom at the end of onFocusRegion, matching the existing behavior of the scroll wheel zoom handler.

#### Manual testing:
Created a task, opened a job in review mode, created an issue on an annotation, and verified that double-click zooming no longer scales up the issue label text.

#### Automated testing:
Added a Cypress e2e test in tests/cypress/e2e/issues_prs2/issue_10297_zooming_in_onto_annotation_increases_issue_text_size.js that patches EventTarget.prototype.dispatchEvent and asserts that a canvas.zoom event is fired when double-clicking an annotation shape.

### Checklist

~~[ ] I have updated the documentation accordingly~~

### License

### DIFF
--- a/tests/cypress/e2e/issues_prs2/issue_10297_zooming_in_onto_annotation_increases_issue_text_size.js
+++ b/tests/cypress/e2e/issues_prs2/issue_10297_zooming_in_onto_annotation_increases_issue_text_size.js
@@ -7,10 +7,13 @@
 context('Zooming onto an annotation must not scale issue text size', () => {
     const issueId = '10297';
     const labelName = `Issue ${issueId}`;
+    const issueDescription = `Issue for bug ${issueId}`;
+    const issueLabelSelector = '.cvat-hidden-issue-label';
+    const shapeSelector = '#cvat_canvas_shape_1';
+    const issueRegionSelector = '.cvat_canvas_issue_region';
 
     let taskID = null;
     let jobID = null;
-    let origDispatchEvent = null;
 
     before(() => {
         cy.visit('/auth/login');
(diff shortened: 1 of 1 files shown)


DanielBorgs:
Hi @bsekachev, @nmanovic, just pinging on this for a review when you have a chance! Let me know if you need any adjustments.

Also, the `rest_api_testing` job failed due to a transient Docker Hub TLS timeout when pulling the `cvat_opa` image, unrelated to these changes. I believe a rerun should be enough to clear it.

PR #10442, opened 2026-04-02
[DEV-831] update helm chart for GPU support + topology features + extraVolumes(mounts)

yaizkazani:


ai-models/agents_deployment/helm/values.yaml
```diff
@@ -63,9 +63,11 @@ agent:
   resources:
     limits:
       cpu: 1
-#      memory: 1.5Gi
+      gpu: 1   # Only used if agent.use_cuda is true
```
azhavoro, line 66: In my opinion, that comment isn't clear.
Does this mean that if `use_cuda=false`, the value will be ignored, or is the user expected to change both values consistently?

yaizkazani: If use_cuda = true, then this value will be considered; otherwise, it will not. Do you have any ideas for making this clearer?

azhavoro (approved):

PR #10446, opened 2026-04-02
Optimize task and job viewsets

zhiltsov-max:
### Motivation and context

Improved DB queries in Task and Job endpoints. The main focus was to remove extra requests from the non-CRUD endpoints in these endpoint groups (e.g. `/metadata`, `/annotations`). The CRUD endpoints were also slightly optimized.

### Checklist

### License


zhiltsov-max:
@archibald1418, you can check perf tests on this branch.

archibald1418:
perf testing:
image

cvat/apps/engine/serializers.py
```diff
@@ -2483,6 +2490,15 @@ def to_representation(self, data):
...
             }
 
+            page_storage_ids = set(
+                v
+                for task in page
+                for v in (task.source_storage_id, task.target_storage_id)
+            )
+            storages = {
+                s.id: s for s in models.Storage.objects.filter(id__in=page_storage_ids)
+            }
```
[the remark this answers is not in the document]
zhiltsov-max: Django doesn't send requests for empty `in` filters.

cvat/apps/engine/views.py
```diff
@@ -1654,18 +1673,11 @@ class JobViewSet(viewsets.GenericViewSet, mixins.ListModelMixin, mixins.CreateMo
...
-            'segment__task',
-            'segment__task__project',
-        )
-        .prefetch_related(
-            'segment__task__data',
-            'segment__task__annotation_guide',
-            'segment__task__project__annotation_guide',
-        )
+    queryset = Job.objects.select_related(
+        # prefetch data for permission checks
```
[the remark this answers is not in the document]
zhiltsov-max: The word "prefetch" can be used in the more broad sense than just the `prefetch_related` function call in Django.

cvat/apps/engine/views.py
```diff
@@ -891,8 +884,16 @@ def get_queryset(self):
             queryset = Task.objects.select_related('data')
         elif self.action == 'validation_layout':
             queryset = Task.objects.select_related('data', 'data__validation_layout')
-        else:
-            queryset = queryset.with_job_summary()
+        elif self.action not in ('metadata', 'annotations'):
```
SpecLad, line 884: Should this branch be merged with the `preview` and `validation_layout` branches, perhaps? It seems like those branches are currently suboptimal, because they discard the `assignee` and `owner` selects from `queryset`.

cvat/apps/engine/views.py
```diff
@@ -1460,13 +1461,31 @@ def _get_rq_response(queue, job_id):
...
-        self.get_object() #force to call check_object_permissions
-        db_task = models.Task.objects.prefetch_related(
-            'segment_set',
-            Prefetch('data', queryset=models.Data.objects.select_related('video').prefetch_related(
-                Prefetch('images', queryset=models.Image.objects.prefetch_related('related_files').order_by('frame'))
-            ))
-        ).get(pk=pk)
+        db_task = self.get_object() #force to call check_object_permissions
+
+        def prefetch():
```
SpecLad, line 1465: Any reason to not put all this prefetching stuff into `get_queryset`? Seems like it would be more consistent.

cvat/apps/engine/serializers.py
```diff
@@ -2498,6 +2514,22 @@ def to_representation(self, data):
                 for k in job_summary_fields:
                     setattr(task, k, task_job_summary[k])
 
+                if task.source_storage_id in storages:
```
SpecLad, line 2539: Maybe `task.source_storage_id is not None` would be a better condition? `storages` should always contain all non-null storage IDs, by construction.

cvat/apps/engine/views.py
zhiltsov-max: I'd actually prefer having prefetch optimizations in the specific endpoints. `get_queryset` works good for CRUD mixins, because they are not supposed for overriding in every class, and for some shared logic between different endpoints. But putting complex prefetches from every request into 1 function will only result in more difficult code, I think.

cvat/apps/engine/views.py
zhiltsov-max: I didn't check these 2 endpoints, but it looks like a good idea.

cvat/apps/engine/serializers.py
zhiltsov-max: It's possible that some objects were locked or removed after the ids were fetched, so they might be missing. We're not using a repeatable read transaction here.

cvat/apps/engine/views.py
zhiltsov-max: I found that as we removed the inefficient pattern like `task.assignee.id` for building the OPA request, more fields can be removed from the prefetches.

SpecLad: Nice!

cvat/apps/engine/views.py
SpecLad: Alright.

cvat/apps/engine/serializers.py
SpecLad: I guess that's true, although it seems like the code would crash either way - either from accessing a missing dictionary key or from accessing a non-existent related object later. We can keep the condition as-is, although it would be nice to have a comment to explain it.

SpecLad (approved):

PR #10449, opened 2026-04-03
Polygon/polyline simplification

klakhov:
### Motivation and context

- Polygon/polyline simplification in object actions (and shortcut)
- Polygon/polyline simplification via annotation actions
- Polygon/polyline auto simplification after object is created

- Refactor opencv code (moved to cvat-core)

simplify

### Checklist

### License


cvat-ui/src/components/annotation-page/annotations-actions/annotations-actions-modal.tsx
```diff
@@ -242,14 +245,26 @@ const componentStorage = createStore(reducer, {
...
-        defaultValue, type, values, onChange,
+        defaultValue, type, values, onChange, parameterName, tooltip,
     } = props;
     const store = getCVATStore();
 
     const job = store.getState().annotation.job.instance as Job;
-    const computedDefaultValue = typeof defaultValue === 'function' ? defaultValue({ instance: job }) : defaultValue;
+    let computedDefaultValue = typeof defaultValue === 'function' ? defaultValue({ instance: job }) : defaultValue;
+
+    if (parameterName === 'Threshold') {
```
klakhov: Maybe not the best approach, should we configure this in cvat-core config and set it up via some callback from ui?

cvat-core/src/object-utils.ts
```diff
@@ -385,3 +385,104 @@ export function propagateShapes<T extends SerializedShape | ObjectState>(
...
+            // Linear interpolation from (1, 0.25) to (8, 3)
+            return (2.75 * approxPolyMaxDistance - 1) / 7;
+        }
+        // Exponential: 4 for 9, 8 for 10, 16 for 11, 32 for 12, 64 for 13
+        return 2 ** (approxPolyMaxDistance - 7);
+    }
+    return 0;
+}
+
+export function approxPolyDP(cv: any): (points: number[], threshold: number, closed?: boolean) => number[] {
```
bsekachev: Exposed to cvat-core-wrapper, but unused further

cvat-core/src/object-utils.ts
```diff
@@ -385,3 +385,104 @@ export function propagateShapes<T extends SerializedShape | ObjectState>(
...
+    return (points: number[], threshold: number, closed = true): number[] => {
+        if (points.length < 3 || points.length % 2) {
+            return points;
+        }
```
bsekachev: The implementation repeated now 3 times. Need to reduce it somehow. 

- I like the idea with making OpenCV interface builder in core somewhere in a separated file. 
- It accepts OpenCV implementation as argument and builds interface around it
- Finally we can import this file to both main thread and workers

What do you think?

cvat-core/src/object-utils.ts
bsekachev: It is better to use as common interface as possible

For example if in function you use only `length` property, it is better to use ArrayLike<number> instead of number[].

Thus you may pass any implementations satisfying this interface (e.g Array, TypedArray, etc.)

bsekachev: Please look also at other types of new functions

cvat-core/src/object-utils.ts
```diff
@@ -385,3 +385,104 @@ export function propagateShapes<T extends SerializedShape | ObjectState>(
...
+            while (simplified.length < minValues && attempts < maxAttempts && adjustedThreshold > 0.01) {
+                simplified = approxPolyDPFn(points, adjustedThreshold, closed);
+                adjustedThreshold *= 0.5;
+                attempts++;
+            }
+
+            if (simplified.length < minValues) {
+                return {
+                    points,
+                    warning: `Could not simplify to at least ${minPoints} points even with reduced threshold`,
```
bsekachev: I would suggest just return initial points instead of overcomplicating the interface

`threshold * (0.5 ^ 5)` is literally 0, so, it results to two points only if initial polygon is two points. But in this case approxPolyDP already returns initial polygon.

cvat-core/src/annotations-actions/poly-simplify.ts
```diff
@@ -0,0 +1,112 @@
...
+        if (!this.#cv) {
+            throw new Error('OpenCV is not loaded. The action modal should have initialized it.');
+        }
+
+        onProgress('Simplifying polygons', 0);
+
+        const simplifyFn = simplifyPoly(this.#cv);
+        const totalShapes = input.collection.shapes.length;
+
+        const simplifiedShapes = input.collection.shapes.map((shape, index) => {
```
bsekachev: Do not you think that such logic should be implemented in a WebWorker? Interface may be frozen if there are many thousands of objects. 
Additionally we would get rid of custom code in `annotations-actions.tsx` related to passing OpenCV as an argument

bsekachev: The function to upload OpenCV in a worker is very simple and may be duplicated or put to worker-utils module for example in cvat-core. 

```
    public initialize(): Promise<void> {
        return new Promise((resolve, reject) => {
            (self as any).Module = {
                onRuntimeInitialized: () => {
                    this.initialized = true;
                    resolve();
                },
            };

            try {
                (self as any).importScripts(config.OPENCV_PATH);
            } catch (error: unknown) {
                reject(error);
            }
        });
    }
```

cvat-core/src/annotations-actions/poly-simplify.ts
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    public get parameters(): ActionParameters | null {
+        return {
+            OpenCV: {
+                type: ActionParameterType.OPENCV_DEPENDENCY,
+                values: [],
+                defaultValue: '',
+            },
+            Threshold: {
```
bsekachev: As I remember we discussed it should not be the same slider when running as action for many objects. It is more logic to pass maximum `distance` itself to the action (it is closer by nature to IoU metric)

cvat-core/src/object-utils.ts
klakhov: I suppose we already have an interface builder - `opencv-wrapper` in `cvat-ui`. Do you mean moving it to `cvat-core` and then using it across `cvat-ui` tools, actions, and web workers?

bsekachev: Yep

bsekachev: Just interface in opencv-wrapper is not exactly builder. It may upload OpenCV as I remember. 

I propose to left uploading logic on UI level and put just wrapper accepting `cv` as argument and returning the implementation.

klakhov: Ok, I see

nmanovic:
@codex review

cvat-core/src/object-utils.ts
klakhov: Removed

cvat-core/src/object-utils.ts
klakhov: Moved opencv files to core

cvat-core/src/object-utils.ts
klakhov: Removed. Now returning initial points

cvat-core/src/object-utils.ts
klakhov: Removed this function

cvat-core/src/annotations-actions/poly-simplify.ts
klakhov: Created worker

cvat-core/src/annotations-actions/poly-simplify.ts
klakhov: I checked the meeting notes and didn’t find anything about a maximum distance parameter. We discussed that an IoU parameter would be useful, but it would be too difficult to implement, so we decided to use a threshold-based approach instead.

Testing the threshold on a single polygon first, and then applying it to the group to see how it performs, should be enough. At least, that’s what the annotation team said.

nmanovic:
@codex review

cvat-core/src/annotations-actions/poly-simplify.ts
bsekachev: I did not propose to implement IoU. I proposed using `distance`. 
Current indicator "more or less" points is good when you see the result immediately. It is UX friendly. 
But it does not work when approximating many polygons at the same time. It does not mean anything for resulting polygons. 

"more or less" points is even not a metric. This is more interface around `distance` metric.

cvat-core/src/opencv/math-utils.ts
```diff
@@ -0,0 +1,26 @@
+// Copyright (C) 2020-2022 Intel Corporation
```
bsekachev, line 1: Why these are in opencv directory? From first look they are not related to OpenCV in general

cvat-core/src/opencv/opencv-interface.ts
```diff
@@ -0,0 +1,225 @@
+// Copyright (C) 2020-2022 Intel Corporation
```
bsekachev, line 1:
```suggestion
```

cvat-core/src/index.ts
```diff
@@ -15,6 +15,7 @@ import * as enums from './enums';
 import config from './config';
 import { mask2Rle, rle2Mask } from './rle-utils';
 import { propagateShapes, validateAttributeValue } from './object-utils';
+import { createOpenCVInterface, simplifyPoly } from './opencv/opencv-interface';
```
bsekachev, line 18: Module '"./opencv/opencv-interface"' has no exported member 'simplifyPoly'.ts

cvat-core/src/opencv/tracker-mil.ts
```diff
@@ -3,12 +3,22 @@
 //
 // SPDX-License-Identifier: MIT
 
-import { clamp } from 'utils/math';
-import { TrackerModel, TrackingResult } from './opencv-interfaces';
+import { clamp } from './math-utils';
 
-export type TrackerMIL = TrackerModel;
+export interface TrackingResult {
```
bsekachev, line 8: Unused outside

cvat-core/src/opencv/opencv-interface.ts
```diff
@@ -0,0 +1,225 @@
...
+    type IntelligentScissorsInterface,
+} from './intelligent-scissors';
+import HistogramEqualizationImplementation, {
+    type HistogramEqualizationInterface,
+} from './histogram-equalization';
+import TrackerMILImplementation, {
+    type TrackerMILInterface,
+} from './tracker-mil';
+
+export function thresholdFromAccuracy(accuracy: number): number {
```
bsekachev, line 16: Let's put the function to OpenCV interface itself, e.g. cv.utils.thresholdFromAccuracy. Thus we do not need to expose it in a separate path.

cvat-core/src/opencv/histogram-equalization.ts
```diff
@@ -3,21 +3,22 @@
...
+import { BaseImageFilter, type ImageProcessing } from './image-processing';
 
-export interface HistogramEqualization extends ImageProcessing {
+export interface HistogramEqualizationInterface extends ImageProcessing {
     processImage: (src: ImageData, frameNumber: number) => ImageData;
 }
 
-export default class HistogramEqualizationImplementation extends BaseImageFilter {
+export default class HistogramEqualizationImplementation
+    extends BaseImageFilter implements HistogramEqualizationInterface {
```
bsekachev, line 13: Is there a reason to use both extends plus implements (if BaseClass is also just interface with zero implementation)

cvat-core/src/annotations-actions/poly-simplify.ts
```diff
@@ -0,0 +1,149 @@
...
+
+export class PolySimplify extends BaseShapesAction {
+    #accuracy = 0;
+    #worker: Worker | null = null;
+    #opencvPath: string = '';
+
+    public async init(_instance: any, parameters: Record<string, any>): Promise<void> {
+        this.#accuracy = parameters.Threshold as number;
+        // Get the OPENCV_PATH from window.cvat.config if available
+        this.#opencvPath = (window as any)?.cvat?.config?.OPENCV_PATH || '/assets/opencv_4.8.0.js';
```
bsekachev, line 21: 1. `OPENCV_PATH` is not declared in interface (cvat/cvat-core/src/index.ts)
2. It is never set somewhere 
3. Why have you added it? 
4. It is still exported and used from cvat_enterprise/cvat/cvat-ui/src/config.tsx

cvat-core/src/annotations-actions/actions-worker.ts
```diff
@@ -0,0 +1,126 @@
...
+if ((self as any).importScripts) {
+    const manager = new ActionsWorkerManager();
+
+    onmessage = (event: MessageEvent<WorkerRequest>) => {
+        const { command } = event.data;
+
+        if (command === WorkerAction.INITIALIZE) {
+            const { opencvPath } = event.data;
+            if (opencvPath) {
+                (self as any).OPENCV_PATH = opencvPath;
```
bsekachev, line 86: Just pass it as an argument to manager.initialize() -> initializeOpenCVInWorker().
It seems not necessary to pollute globals

cvat-core/src/annotations-actions/actions-worker.ts
```diff
@@ -0,0 +1,126 @@
...
+                    closed,
+                );
+
+                return {
+                    ...shape,
+                    points: simplifiedPoints,
+                };
+            } catch (error: unknown) {
+                // If simplification fails, return original shape
+                console.error(`Failed to simplify shape ${shape.clientID}:`, error);
```
bsekachev, line 66: It would be useful to provide also frame number

cvat-core/src/annotations-actions/actions-worker.ts
```diff
@@ -0,0 +1,126 @@
...
+    public simplifyShapes(shapes: SimplifyShape[], accuracy: number): SimplifyShape[] {
+        if (!this.initialized || !this.cvInterface) {
+            throw new Error('Worker not initialized');
+        }
+
+        return shapes.map((shape) => {
+            try {
+                if (!shape.points || shape.points.length < 6) {
+                    return shape;
+                }
```
bsekachev, line 51: `simplifyPolygon` does the same check

cvat-core/src/annotations-actions/poly-simplify.ts
bsekachev: And now we have duplicated hardcoded literal `/assets/opencv_4.8.0.js`

bsekachev:
Please, run .ts check

image

cvat-ui/src/cvat-core-wrapper.ts
```diff
@@ -48,6 +48,11 @@ import { BaseShapesAction } from 'cvat-core/src/annotations-actions/base-shapes-
 import { BaseCollectionAction } from 'cvat-core/src/annotations-actions/base-collection-action';
 import { ActionParameterType, BaseAction } from 'cvat-core/src/annotations-actions/base-action';
 import { Request, RequestOperation } from 'cvat-core/src/request';
+import {
+    createOpenCVInterface, thresholdFromAccuracy, MatType,
+    type OpenCVInterface,
+} from 'cvat-core/src/opencv/opencv-interface';
```
bsekachev, line 54: You already export `createOpenCVInterface` via general `cvat` interface. 
OpenCVInterface is easily inferred like this

type OpenCVInterface = ReturnType<typeof _cvat.opencv.createOpenCVInterface>

bsekachev: MatType may be enum of openCVInterface (and this would be probably more correct)

cvat-ui/src/actions/annotation-actions.ts
```diff
@@ -1244,19 +1257,25 @@ export function changeWorkspaceAsync(workspace: Workspace): ThunkAction {
     };
 }
 
-export function createAnnotationsAsync(statesToCreate: any[]): ThunkAction {
-    return async (dispatch: ThunkDispatch): Promise<void> => {
+export function createAnnotationsAsync(
+    statesToCreate: any[],
```
bsekachev, line 1261:
```suggestion
    statesToCreate: ObjectState[],
```

cvat-ui/src/components/annotation-page/canvas/views/canvas2d/canvas-wrapper.tsx
```diff
@@ -723,10 +728,15 @@ class CanvasWrapperComponent extends React.PureComponent<Props> {
         } else {
             jobInstance.logger.log(EventScope.pasteObject, { count: 1, duration });
         }
-
         const objectState = new cvat.classes.ObjectState(state);
-        onCreateAnnotations([objectState]);
-        onUpdateEditedObject(null);
+        const states = await onCreateAnnotations([objectState]);
```
bsekachev, line 732: It would be better to add extra argument to `onCreateAnnotations`, e.g. (source = 'draw')
And if `source === 'draw'`, trigger `onSwitchSimplifyVisibility` from `onCreateAnnotations`.

Thus you do not need to have `onSwitchSimplifyVisibility` here and you do not need to update interface of `onCreateAnnotations` and `fetchAnnpotationsAsync`, making it more difficult as I wrote below

cvat-ui/src/actions/annotation-actions.ts
```diff
@@ -353,8 +354,9 @@ async function fetchAnnotations(predefinedFrame?: number): Promise<{
     };
 }
 
-export function fetchAnnotationsAsync(): ThunkAction {
-    return async (dispatch: ThunkDispatch, getState: () => CombinedState): Promise<void> => {
+export function fetchAnnotationsAsync(): (
+    dispatch: ThunkDispatch, getState: () => CombinedState) => Promise<ObjectState[] | null> {
```
bsekachev, line 358: The interface is more difficult now. The function is expected to return a value, but if exception happens, it silents and the function returns null instead. It is become difficult to work with this function.

cvat-core/tsconfig.json
```diff
@@ -10,5 +10,5 @@
       "baseUrl": "src",
       "resolveJsonModule": true,
     },
-    "include": ["src/*.ts"]
+    "include": ["src/*.ts", "src/**/*.ts"]
```
bsekachev, line 13: Perhaps the second pattern includes the first one

cvat-core/src/annotations-actions/poly-simplify.ts
klakhov: Added distance input number

cvat-core/src/opencv/math-utils.ts
klakhov: They are used in `opencv-interface`, so we need to import them in the worker as well. If we move them to `object-utils/common`, those files bring their own dependencies, which causes errors when they are loaded in the worker.

So it seems that the utility functions file for OpenCV needs to stay fairly clean and free of extra dependencies.

cvat-core/src/opencv/opencv-interface.ts
klakhov: Moved.

cvat-core/src/opencv/histogram-equalization.ts
klakhov: Removed extra interface

cvat-core/src/annotations-actions/poly-simplify.ts
klakhov: Removed hardcoding, added import via config

cvat-ui/src/cvat-core-wrapper.ts
klakhov: Added enums to the interface, removed this export.

cvat-ui/src/components/annotation-page/canvas/views/canvas2d/canvas-wrapper.tsx
klakhov: Added source (a bit more specific - DRAW_SIMPLIFIED_POLY as info about simplifyPoly comes from draw event itself) and a logic in onCreateAnnotations that starts simplification. Reverted both fetchAnnpotationsAsync onCreateAnnotations back

bsekachev:
@codex review

bsekachev:
Overall, LGTM, but several comments:
1. I enabled TypeScript checks for all .ts code, please look at failed linter
2. Perhaps these two settings (Number of points and Simplify) should not be used together  (when number of points specified, simplify should not be applied, probably input shall be disabled)
image

3. I believe during simplifyng a single object (from menu or after drawing), the object should be activated (ortherwise I can't really see effect of dragging the slider, especially when I accidently activate another object)

klakhov:
@codex review

klakhov:
@codex review

bsekachev:
https://github.com/user-attachments/assets/45467171-9c3c-45ad-90af-f4503200fedc

Still lose active object while dragging cursor from object to the slider

klakhov:
Applied the fix. Now its working for simplification that is applied from object menu.

bsekachev (approved):

archibald1418:
@codex what does `distance` measure?

PR #10460, opened 2026-04-08
Enforce 8-256 character limits for new passwords

nmanovic:
## Summary
- enforce an explicit `8..256` password length policy for newly set passwords across registration, password change, and password reset confirm flows
- keep the OpenAPI request component names stable while owning the serializer limits in CVAT
- keep the current registration UX where the initial submit on an empty form reveals all required fields, instead of disabling submit before any validation feedback is shown
- keep unit-test coverage for the password policy and add the edge case that an already-existing password longer than 256 can still be used as `old_password`

## Testing
- `docker exec cvat_server python manage.py test --keepdb cvat.apps.iam.tests.test_rest_api.UserRegisterAPITestCase -v 2`
- `python -m black --check cvat/apps/iam/tests/test_rest_api.py tests/python/rest_api/test_auth.py`
- `isort --check --resolve-all-configs tests/python/rest_api/test_auth.py`


cvat/schema.yml
```diff
@@ -351,7 +351,7 @@ paths:
         content:
           application/json:
             schema:
-              $ref: '#/components/schemas/PasswordChangeRequest'
+              $ref: '#/components/schemas/PasswordChangeSerializerExRequest'
```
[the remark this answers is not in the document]
nmanovic: Fixed in `59d2468bf`.

I kept the runtime CVAT-owned serializers, but restored the public OpenAPI component names to `PasswordChangeRequest` and `PasswordResetConfirmRequest` via schema annotations, then regenerated `cvat/schema.yml` and the SDK.

That brings the generated client surface back to the stable request model names, so downstream uses like `tests/python/rest_api/test_access_tokens.py` no longer break.

Re-verified locally with:
- `COMPOSE_COMPATIBILITY=1 pytest tests/python/rest_api/test_access_tokens.py -vv`
- `COMPOSE_COMPATIBILITY=1 pytest tests/python/rest_api/test_auth.py -vv`
- `docker exec test_cvat_server_1 python manage.py test --keepdb cvat.apps.iam.tests.test_rest_api.UserRegisterAPITestCase -v 2`

cvat/apps/iam/tests/test_rest_api.py
zhiltsov-max, line 1: I can see these unit tests mostly duplicate the integration tests, so maybe it makes sense to keep only one version? 
Also, there's no test to check that the old password can be longer than the limit, even though there's code specifically handling this situation. Actually, it is the only test that should reside in the unit tests according to our testing policy, as such a password cannot be created after the PR.

cvat/apps/iam/password_validation.py
```diff
@@ -0,0 +1,27 @@
+# Copyright (C) CVAT.ai Corporation
+#
+# SPDX-License-Identifier: MIT
+
+from django.core.exceptions import ValidationError
+
+DEFAULT_MIN_PASSWORD_LENGTH = 8
+DEFAULT_MAX_PASSWORD_LENGTH = 256
```
zhiltsov-max, line 8: Typically, we store such configuration constants either in `settings/*.py` or in `apps/*/default_settings.py`. I think it makes sense for these parameters as well.

cvat/settings/base.py
```diff
@@ -400,13 +405,17 @@ class REDIS_INMEM_DATABASES(IntEnum):
     },
     {
         "NAME": "django.contrib.auth.password_validation.MinimumLengthValidator",
+        "OPTIONS": {"min_length": DEFAULT_MIN_PASSWORD_LENGTH},
```
zhiltsov-max, line 408: It looks like the max length variable should either be forwarded to the max length validator below or be configurable via this file.

cvat/apps/iam/tests/test_rest_api.py
SpecLad: I would vote for keeping unit tests over integration, because such tests are easier to run and debug, and they are faster as well.

cvat-ui/src/components/register-page/register-form.tsx
```diff
@@ -243,16 +243,18 @@ function RegisterFormComponent(props: Props): JSX.Element {
...
+                <Form.Item shouldUpdate>
+                    {(): JSX.Element => (
+                        <Button
+                            type='primary'
+                            htmlType='submit'
+                            className='cvat-credentials-action-button'
+                            loading={fetching}
+                            disabled={fetching || form.getFieldsError().some(({ errors }) => errors.length > 0)}
+                        >
+                            Create account
```
bsekachev, line 255: If the idea was to disable Submit button until all UI validators are ok, the button also should be disabled by default when the form itself is empty or partially filled out

cvat/apps/iam/tests/test_rest_api.py
bsekachev: > I would vote for keeping unit tests over integration

``+ ``
It is too heavy to cover such minor things by E2E tests.

cvat/apps/iam/password_validation.py
SpecLad: FWIW, I don't think maximum length needs to be a setting. I see no practical reason why anyone would want to allow passwords longer than 256 characters, and there is no reason to reduce this limit either.

But _also_... given that the limits are hardcoded in the UI, it seems pointless to make them adjustable on the backend.

nmanovic: I agree with Roman. I don't want to see them in settings.

cvat/settings/base.py
nmanovic: We have no choice here, because we need to configure an external validator. With cvat.apps.iam.password_validation.MaximumLengthPasswordValidator we don't need that actually. But somebody can redefine base.py if it is required.

cvat/apps/iam/tests/test_rest_api.py
nmanovic: I simplified the PR in that direction.

I removed the extra integration and E2E coverage added by this PR and kept the backend unit tests only. I also added the missing unit test for the special case we explicitly support here: an already-existing password longer than 256 can still be used as `old_password` when changing to a valid new password.

cvat-ui/src/components/register-page/register-form.tsx
nmanovic: I’d prefer to keep the current behavior.

If the button is disabled from the initial empty state, the form gives no immediate feedback about what exactly is missing. With the current flow, the user can click `Create account` once and instantly see all required fields highlighted in red, which makes the requirements explicit.

So from a UX perspective I think this is more informative:
- initial state: button is enabled
- first submit on an empty or invalid form: validation errors are shown
- after that: the button stays disabled until the errors are fixed

This way the form both explains what is required and prevents repeated invalid submits.

cvat/apps/iam/password_validation.py
nmanovic: I kept the limits as internal backend policy values rather than moving them into settings.

At this point both UI and backend intentionally hardcode the same `8..256` limits, so making them configurable would add indirection without a practical use case for this feature.

cvat/settings/base.py
nmanovic: I kept the current shape here.

`MinimumLengthValidator` still needs its explicit option in `base.py` because it is an external validator, but the max limit is enforced by CVAT-owned serializer fields and `MaximumLengthPasswordValidator`, so there is no separate settings knob for it in this PR.

cvat/apps/iam/tests/test_rest_api.py
nmanovic: Pushed in `32f8b86e8`.

The PR now keeps only the backend unit coverage for this policy change:
- removed the added REST integration tests from `tests/python/rest_api/test_auth.py`
- removed the Cypress spec
- kept the unit tests in `cvat/apps/iam/tests/test_rest_api.py`
- added the missing unit case for a long existing `old_password`

cvat-ui/src/components/register-page/register-form.tsx
bsekachev: > If the button is disabled from the initial empty state, the form gives no immediate feedback about what exactly is missing.

If you try you will see that it is not true, it gives immediate feedback about validators while inputting and the default state of the button does not play role here

bsekachev: In the current implementation the button initially enabled and get disabled after first character in password field, what looks weird

cvat/apps/iam/serializers.py
```diff
@@ -25,6 +34,20 @@ class RegisterSerializerEx(RegisterSerializer):
     email = serializers.EmailField(
         required=allauth_settings.EMAIL_REQUIRED, max_length=allauth_settings.EMAIL_MAX_LENGTH
     )
+    password1 = serializers.CharField(
+        write_only=True,
+        style={"input_type": "password"},
+        trim_whitespace=False,
```
SpecLad, line 42: Why false? IMO, a user is much more likely to accidentally put spaces around their password than to actually want them there.

cvat/apps/iam/serializers.py
```diff
@@ -25,6 +34,20 @@ class RegisterSerializerEx(RegisterSerializer):
     email = serializers.EmailField(
         required=allauth_settings.EMAIL_REQUIRED, max_length=allauth_settings.EMAIL_MAX_LENGTH
     )
+    password1 = serializers.CharField(
```
SpecLad, line 39: A comment here about why the fields are being overridden would be useful.

cvat/apps/iam/serializers.py
```diff
@@ -107,6 +130,50 @@ def get_email_options(self):
...
+        style={"input_type": "password"},
+        trim_whitespace=False,
+        min_length=DEFAULT_MIN_PASSWORD_LENGTH,
+        max_length=DEFAULT_MAX_PASSWORD_LENGTH,
+    )
+
+
+@extend_schema_serializer(component_name="PasswordChange")
+class PasswordChangeSerializerEx(PasswordChangeSerializer):
+    # Keep old password uncapped here so existing users are not blocked by serializer limits.
```
SpecLad, line 156: I'm not sure that's a good idea:

1. By leaving the length uncapped, we're still letting users submit requests with unbounded body length, stopping which is presumably the whole point of introducing a cap in the first place.
2. 256 characters is a huge limit; I think nobody or almost nobody using the tool legitimately would have a password longer than this. And even if they do, they can still change it by resetting.

cvat/settings/base.py
```diff
@@ -171,6 +172,10 @@ def generate_secret_key():
     "REGISTER_SERIALIZER": "cvat.apps.iam.serializers.RegisterSerializerEx",
     "LOGIN_SERIALIZER": "cvat.apps.iam.serializers.LoginSerializerEx",
     "PASSWORD_RESET_SERIALIZER": "cvat.apps.iam.serializers.PasswordResetSerializerEx",
+    # Define password-setting serializers explicitly so CVAT controls length limits
+    # instead of inheriting hardcoded third-party defaults.
```
SpecLad, line 176: This comment duplicates information that is already present in the comments inside the serializers themselves.

tests/python/rest_api/test_auth.py
```diff
@@ -309,12 +309,7 @@ def test_can_report_weak_password(self, admin_user: str):
                 _check_status=False,
             )
             assert response.status == HTTPStatus.BAD_REQUEST
-            assert json.loads(response.data) == {
-                "new_password2": [
-                    "This password is too short. It must contain at least 8 characters.",
```
SpecLad, line 314: This seems like a regression. We used to report multiple problems at once, but now we only report one?

cvat/apps/iam/tests/test_rest_api.py
```diff
@@ -208,3 +210,142 @@ def test_basic_authentication_respects_email_confirmation(self):
         db_email_address.verified = True
         db_email_address.save()
         self._authenticate_with_basic_auth(should_be_authorized=True)
+
+    @override_settings(ACCOUNT_EMAIL_VERIFICATION="none")
+    def test_register_rejects_oversized_passwords(self):
```
SpecLad, line 215: Frankly, I'm not sure if this is even worth testing. The entire implementation of this check is a `max_length` parameter of the corresponding field - surely we don't need to check that `max_length` works correctly?

SpecLad:
There is some logic duplication here: the minimum and maximum checks are both performed by the serializers, and also by password validators. IMO, the purposes of these checks are different, and so each should be performed by the appropriate tool:

* The minimum check is to reject weak passwords, so it should be performed by a password validator (and in fact, it already is in `develop`, so no changes are needed).
* The maximum check is to prevent excessive processing, so it should be performed as early as possible, i.e. in the serializers.

cvat/apps/iam/serializers.py
nmanovic: I don't think it is the right approach. Probably we can forbidden to add spaces in the beginning or ending of a password, but trim them is a wrong strategy for sure. I agree with AI here:  `Passwords should be treated as opaque strings of bytes, and spaces (including leading/trailing ones) are valid, high-entropy characters that increase security. Silently altering a user’s password by stripping spaces can lead to confusion, as the user may think their password is "password " while your system stores it as "password`

cvat/settings/base.py
nmanovic: It explains why the code exists and what is the reason to do that at all. If the comment was something like "Define password-setting serializers" I would agree. But the rest of the sentence explain an important information and reason behind the code exists.

cvat/apps/iam/serializers.py
nmanovic: I think that both approaches are good enough due to your second comment. In any case, we fix the issue not because it is very expensive to handle such passwords, but because we don't want to get new complaints from "security experts" that it is a problem. Also, the current behavior is better from a compatibility point of view even if the probability to meet the problem is very low.

cvat/apps/iam/tests/test_rest_api.py
nmanovic: We have some contracts for the public RESTful API, and we are testing them here. What is your proposal?

tests/python/rest_api/test_auth.py
nmanovic: We have tests which check short passwords. The test checks other behaviours like too common and unreliable passwords. I don't think it is a regression. We check a weak password.

nmanovic:
Replying to https://github.com/cvat-ai/cvat/pull/10460#issuecomment-4225449948

I agree there is duplication here, and it is intentional.

We want the serializer fields to carry the explicit `8..256` contract so the OpenAPI schema and generated SDK reflect the actual accepted range for these API endpoints.

At the same time, we also want the password validation rules to remain part of Django’s password validation pipeline, because passwords may be validated outside these serializers as well.

So in this case I’d prefer to keep the overlap explicitly:
- serializer constraints define the API-level contract and provide early rejection
- password validators keep the backend password policy centralized

cvat-ui/src/components/register-page/register-form.tsx
nmanovic: OK. Agree. Let's revert the part. Probably blocking of the button was a wrong idea.

PR #10468, opened 2026-04-12
Fix 9676

### LINKED ISSUE #9676: 3d annotation box dimensions are swapped

### Steps to Reproduce

Normal convention for Length x Width xHeight are: 
Length -> along X,
Width  -> along Y
Height -> z dimension.

In current version, I see the following convention: 
Width -> along X
Height -> along Y
Length -> along Z

Image

### Expected Behavior

_No response_

### Possible Solution

_No response_

### Context

_No response_

### Environment

```Markdown

```

prathamesh-patil-5090:
## Description

Fixes #9676 

This PR addresses the issue where the dimensions of 3D annotation cuboids were displayed with incorrect labels in the Object Details sidebar. 

According to standard conventions, the dimensions are now mapped as follows:
- **Length** corresponds to the X dimension (`points[6]`)
- **Width** corresponds to the Y dimension (`points[7]`)
- **Height** corresponds to the Z dimension (`points[8]`)

### Changes made
- Updated `SizeType` enum and `SizeParams` interface in `cvat-ui/src/components/annotation-page/standard-workspace/objects-side-bar/object-item-details.tsx` to match the new layout order (`Length`, `Width`, `Height`).
- Adjusted the parameter passing and `changeSize` payload in `cvat-ui/src/containers/annotation-page/standard-workspace/objects-side-bar/object-item-details.tsx` to correctly map `points[6]`, `points[7]`, and `points[8]` to the updated fields.
- Created a changelog fragment via `scriv`.

### Testing
- Tested the CVAT UI development server.
- Created a 3D Task and drew a Cuboid annotation.
- Verified that the "DETAILS" panel displays the dimensions in the correct "Length", "Width", and "Height" fields, and changes accurately reflect the corresponding X, Y, and Z plane values.

Closes #9676


klakhov:
I’m not sure about this change. The main concern is consistency with 2D, where `x` represents width and `y` represents height. It might be acceptable to interpret them differently here, but I think this needs a bit more thought.

prathamesh-patil-5090:
> I’m not sure about this change. The main concern is consistency with 2D, where `x` represents width and `y` represents height. It might be acceptable to interpret them differently here, but I think this needs a bit more thought.

Hi @klakhov ,

You make a very good point. In the context of computer vision and image processing (2D), the established convention is indeed that **X represents width** and **Y represents height**.

If we were to change the 3D dimensions to Length (X), Width (Y), and Height (Z), it would cause an inconsistency between how 2D and 3D annotations are interpreted within the same tool. Maintaining `Width -> X`, `Height -> Y`, and `Length/Depth -> Z` aligns the 3D annotations natively with the 2D plane of the UI and imaging space. 

Given this consistency concern, I agree that it makes sense to keep the current implementation where X = width, Y = height, and Z = length. 

Should we close this PR and the corresponding issue, or is there an alternative approach (e.g., labeling them as Width(X), Height(Y), Length(Z) explicitly to avoid confusion) that you would prefer?

klakhov:
Lets close this for now.

PR #10480, opened 2026-04-15
Add `pytest --keep-data` to keep test data volumes across sessions

archibald1418:
### Motivation and context

At times, it is useful to use rest api tests for troubleshooting, i.e. reproducing issues manually or automatically. For this activity, data persistence can be useful. This PR adds `pytest --keep-data` option to conditionally restore docker volumes and postgres data

### Checklist

### License

### DIFF
--- a/tests/python/shared/fixtures/init.py
+++ b/tests/python/shared/fixtures/init.py
@@ -429,15 +429,19 @@ def session_start(
 
     if session.config.getoption("--collect-only"):
         if any((stop, start, rebuild, cleanup, dumpdb)):
-            raise Exception("""--collect-only is not compatible with any of the other options:
-                --stop-services --start-services --rebuild --cleanup --dumpdb""")
+            raise Exception(
+                """--collect-only is not compatible with any of the other options:
+                --stop-services --start-services --rebuild --cleanup --dumpdb"""
+            )
         return  # don't need to start the services to collect tests
 
     platform = session.config.getoption("--platform")
 
     if platform == "kube" and any((stop, start, rebuild, cleanup, dumpdb)):
-        raise Exception("""--platform=kube is not compatible with any of the other options
-            --stop-services --start-services --rebuild --cleanup --dumpdb""")
+        raise Exception(
+            """--platform=kube is not compatible with any of the other options
+            --stop-services --start-services --rebuild --cleanup --dumpdb"""
+        )
 
     if platform == "local":
         local_start(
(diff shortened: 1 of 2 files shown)


tests/python/shared/fixtures/init.py
SpecLad, line 442: What about the Kubernetes branch?

SpecLad:
I don't have a principled objection to this, but I don't quite understand the use case. Could you explain in more detail how you're expecting to use this?

archibald1418:
> Could you explain in more detail how you're expecting to use this?

For instance, yesterday I needed to quickly generate data to reproduce issue with annotations, using pytest setup for this is the most convenient way, since we already have tests scenarios that we can modify + it's the quickest and the most flexible way to setup some particular data for manual checks and debuggung (e.g. how such and such page looks with such and such tasks and which backend code executes). Cypress tests don't give you such control

SpecLad:
I still don't understand how `--keep-data` fits into this. Could you describe the workflow you have in mind, step by step?

archibald1418:
- Set up some special task with annotations using UI. Remember its id.
- Run test with this task id.
- go to ui to see changes. Change something (i.e. add some annotations)
- run test again with different data to see new changes

With current setup, every time the data gets reset, I need to setup admin again and the task again, manually. Basically, using test fixtures, clients etc. in a script to send requests. Why reinvent the wheel and write extraneuous sdk scripts when we already have test cases that implement this

This will help me to check customers' data-related issues  without too much tedious manual work

SpecLad:
Okay, makes sense.

tests/python/shared/fixtures/init.py
archibald1418: added

tests/python/shared/fixtures/init.py
```diff
@@ -506,17 +523,18 @@ def kube_start(cvat_db_dir):
 
     wait_for_services()
 
-    kube_exec_cvat(
-        ["sh", "-c", "./manage.py flush --no-input && ./manage.py loaddata /tmp/data.json"]
-    )
+    if not keep_data:
```
SpecLad: Seems that `kube_restore_data_volumes` should also be conditioned on this.

SpecLad: Probably the `kube_cp` calls too?

archibald1418: Inverted the condition

SpecLad (approved):

PR #10481, opened 2026-04-15
Uvicorn: use --uds instead of --fd 0 to enable multi-worker support

azhavoro:
Supervisord socket delegation via `--fd` is incompatible with Uvicorn's `--workers` argument (forces TCP fallback). Moved socket creation to Uvicorn internal manager via `--uds` to allow multiple workers.

### Checklist

### License


cvat/settings/base.py
```diff
@@ -454,7 +454,7 @@ class REDIS_INMEM_DATABASES(IntEnum):
 EVENTS_LOCAL_DB_ROOT.mkdir(parents=True, exist_ok=True)
 EVENTS_LOCAL_DB_FILE = Path(
     EVENTS_LOCAL_DB_ROOT,
-    os.getenv("CVAT_EVENTS_LOCAL_DB_FILENAME", "events.db"),
+    f"events_{os.getpid()}.db",
```
SpecLad, line 457: Won't this result in a loss of buffered events if the server gets killed and restarted?

azhavoro:
@SpecLad @yaizkazani Do you have any other comments?

cvat/settings/base.py
```diff
@@ -452,9 +454,49 @@ class REDIS_INMEM_DATABASES(IntEnum):
 
 EVENTS_LOCAL_DB_ROOT = BASE_DIR / "events"
 EVENTS_LOCAL_DB_ROOT.mkdir(parents=True, exist_ok=True)
+_events_local_db_slot_lock_fd = None
```
SpecLad: Why is this a global variable? It's not used outside the function.

supervisord/server.conf
```diff
@@ -9,14 +9,14 @@ startretries=5
...
     python3 -m uvicorn
-        --fd 0
+        --uds /tmp/uvicorn.sock
         --forwarded-allow-ips='*'
+        --workers %(ENV_NUMPROCS)s
         cvat.asgi:application
 autorestart=true
-environment=CVAT_EVENTS_LOCAL_DB_FILENAME="events_%(process_num)03d.db"
-numprocs=%(ENV_NUMPROCS)s
+environment=CVAT_EVENTS_LOCAL_DB_FILENAME="events_{worker:03d}.db"
```
SpecLad: Since this is now a constant value, it doesn't seem necessary to pass it anymore. CVAT can just hardcode this string.

cvat/settings/base.py
```diff
@@ -452,9 +454,49 @@ class REDIS_INMEM_DATABASES(IntEnum):
...
+
+
+def _resolve_events_local_db_filename(template: str) -> str:
+    global _events_local_db_slot_lock_fd
+
+    pid = os.getpid()
+    worker = pid
```
SpecLad: This default is inconsistent with the worker numbering used in the other branch. Seems like it should be 0.

cvat/settings/base.py
```diff
@@ -452,9 +454,49 @@ class REDIS_INMEM_DATABASES(IntEnum):
...
+    if "{worker" in template:
+        try:
+            workers = max(1, int(os.getenv("NUMPROCS", "1")))
+        except ValueError:
+            workers = 1
+
+        slot_locks_root = EVENTS_LOCAL_DB_ROOT / ".slots"
+        slot_locks_root.mkdir(parents=True, exist_ok=True)
+
+        for slot in range(workers):
```
SpecLad: Seems like this `for` should have an `else` branch, so that we can abort if we somehow reach the end of this loop.

cvat/settings/base.py
```diff
@@ -452,9 +454,49 @@ class REDIS_INMEM_DATABASES(IntEnum):
...
+                os.close(fd)
+                continue
+
+            _events_local_db_slot_lock_fd = fd
+            worker = slot
+            break
+
+    try:
+        return template.format(worker=worker)
+    except (KeyError, ValueError, IndexError):
```
SpecLad: I don't think it's a good idea to just ignore these errors - if something unexpected happens, multiple workers could end up using the same filename and stepping on each other's toes.

cvat/settings/base.py
```diff
@@ -452,9 +454,49 @@ class REDIS_INMEM_DATABASES(IntEnum):
...
+    pid = os.getpid()
+    worker = pid
+
+    if "{worker" in template:
+        try:
+            workers = max(1, int(os.getenv("NUMPROCS", "1")))
+        except ValueError:
+            workers = 1
+
+        slot_locks_root = EVENTS_LOCAL_DB_ROOT / ".slots"
```
SpecLad: IMO, this directory shouldn't be hidden. I don't think anyone would look into `EVENTS_LOCAL_DB_ROOT` unless they were debugging, and in this case it's better to show everything.

(Arguably, we don't need to have a subdirectory for this at all, but that's subjective.)

cvat/settings/base.py
```diff
@@ -452,9 +454,48 @@ class REDIS_INMEM_DATABASES(IntEnum):
 
 EVENTS_LOCAL_DB_ROOT = BASE_DIR / "events"
 EVENTS_LOCAL_DB_ROOT.mkdir(parents=True, exist_ok=True)
+
+
+def _resolve_events_local_db_filename() -> str:
+    worker = 0
```
SpecLad, line 460: This assignment seems redundant, since it will always be overridden by `worker = slot` in the loop.

cvat/settings/base.py
```diff
@@ -452,9 +454,48 @@ class REDIS_INMEM_DATABASES(IntEnum):
...
+        lock_file = slot_locks_root / f"worker_{slot:03d}.lock"
+        fd = os.open(lock_file, os.O_RDWR | os.O_CREAT, stat.S_IRUSR | stat.S_IWUSR)
+        try:
+            fcntl.flock(fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
+        except BlockingIOError:
+            os.close(fd)
+            continue
+
+        # Keep descriptor reachable for the process lifetime to hold the lock.
+        _resolve_events_local_db_filename._slot_lock_fd = fd
```
SpecLad, line 480: I don't think this would actually do anything. `fd` is just an integer, so nothing is going to happen if it's garbage collected. Although it would still be useful to have a comment stating that we are deliberately keeping the descriptor open.

cvat/settings/base.py
```diff
@@ -452,9 +454,48 @@ class REDIS_INMEM_DATABASES(IntEnum):
...
+        workers = max(1, int(os.getenv("NUMPROCS", "1")))
+    except ValueError:
+        workers = 1
+
+    slot_locks_root = EVENTS_LOCAL_DB_ROOT / "slots"
+    slot_locks_root.mkdir(parents=True, exist_ok=True)
+
+    for slot in range(workers):
+        lock_file = slot_locks_root / f"worker_{slot:03d}.lock"
+        fd = os.open(lock_file, os.O_RDWR | os.O_CREAT, stat.S_IRUSR | stat.S_IWUSR)
```
SpecLad, line 472:
```suggestion
        fd = os.open(lock_file, os.O_RDWR | os.O_CREAT | os.O_CLOEXEC, stat.S_IRUSR | stat.S_IWUSR)
```

cvat/settings/base.py
```diff
@@ -452,9 +454,48 @@ class REDIS_INMEM_DATABASES(IntEnum):
...
+        break
+    else:
+        raise ImproperlyConfigured(
+            f"Failed to allocate events DB worker slot: all {workers} slots are busy"
+        )
+
+    return f"events_{worker:03d}.db"
+
+
+if to_bool(os.getenv("CVAT_EVENTS_LOCAL_DB_PER_WORKER", False)):
```
SpecLad, line 491: Do we actually need to have this setting? What'll happen if we always behave as if it's true?

PR #10485, opened 2026-04-16
Enable typescript checks in our pipeline (cvat-data)

bsekachev:
### Motivation and context
1. Added typescript check for `cvat-data`
2. Reduced code duplication by composing common steps into a separated action

### Checklist

### License

### DIFF
--- a/.github/actions/setup-yarn/action.yml
+++ b/.github/actions/setup-yarn/action.yml
@@ -1,0 +1,62 @@
+name: Setup Yarn
+description: Set up Node.js, enable Yarn via Corepack, cache dependencies, and install packages
+
+inputs:
+  node-version:
+    description: Node.js version to install
+    required: false
+    default: 22.x
+  node-modules-path:
+    description: Paths to cache for node_modules
+    required: false
+    default: node_modules
+  install-command:
+    description: Command used to install dependencies
+    required: false
+    default: yarn install --immutable
+
+runs:
+  using: composite
+  steps:
+    - uses: actions/setup-node@v4
+      with:
+        node-version: ${{ inputs.node-version }}
+
+    - name: Enable corepack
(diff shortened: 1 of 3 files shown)


SpecLad:
I don't think the prettifier made an improvement here. It made some really odd choices:

* ```
  'on': pull_request
  ```

* ```
          run: >
            curl -sLo "$HADOLINT"
            "https://github.com/hadolint/hadolint/releases/download/v$HADOLINT_VER/hadolint-Linux-x86_64"
  ```

  It's hard to understand how this even works unless you are a YAML specialist. :slightly_smiling_face: 

* Deleting blank lines between steps.

IMO, you should revert the style changes.

bsekachev:
> 'on': pull_request

It seems it is a standard thing. 

In YAML 1.1, on is interpreted as a boolean true value, along with yes and y. To treat "on" as a literal string rather than a boolean, it must be quoted (e.g., 'on' or "on"). While YAML 1.2 removed this behavior to focus only on true/false, many parsers still use 1.1 conventions.

bsekachev:
> Deleting blank lines between steps

In some places, there are blank lines between steps.
In some places, there are not. 

We need a convention, not important `with` or `without` (for me personally), but it should be consistent. 


BTW in order to not look at such things, we probably need a linter for yaml files.

SpecLad:
> In YAML 1.1, on is interpreted as a boolean true value, along with yes and y. To treat "on" as a literal string rather than a boolean, it must be quoted (e.g., 'on' or "on"). While YAML 1.2 removed this behavior to focus only on true/false, many parsers still use 1.1 conventions.

I mean... the only parser that matters here is the one GitHub uses to parse the workflow files, and that's never had a problem with "on" being unquoted.

PR #10362, opened 2026-03-12
Join for polygons

klakhov:
### Motivation and context

Extended join tool to support polygons
polygon-join

### Checklist

- [x] I submit my changes into the `develop` branch
- [x] I have created a changelog fragment 
- [ ] I have updated the documentation accordingly
- [ ] I have added tests to cover my changes
- [x] I have linked related issues (see [GitHub docs](
  https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword))

### License

- [x] I submit _my code changes_ under the same [MIT License](
  https://github.com/cvat-ai/cvat/blob/develop/LICENSE) that covers the project.
  Feel free to contact the maintainers if that's a concern.

PR #10490, opened 2026-04-16
Add snap-to-point case where first point is snapped

archibald1418:
### Motivation and context

Depends on #10509

Found a bug in snap-to-point #10421 whereby a snapped polygon's first point is not snapping when it is on an another shape.
The following points snap correctly.
This bug is also crucial for 'Join polygons' #10362  feature, since in order for polygons to join correctly, they must have shared sides, with no gaps in between, which is typically achieved with the help of snap-to-point feature (if you are adding a polygon on top of something else)

https://github.com/user-attachments/assets/fd69fe55-0b64-4255-846a-f8d7bf0165ea

### How has this been tested?

Added a test scenario:
- draw a rectangle
- turn on snap-to-point
- draw a polygon, starting from the rectangle's first point. Click in the region of the two point (previous test proved that delta region of 1.3*r is correct), put third point away, resulting in a triangle. Thus, the both shapes have a common side, the triangle should visually sit perfectly on the rectangle
- check that the polygon's and the rectangle's first point are identical in global coordinates (=snapping is correct)

### Checklist

### License


archibald1418:
@codex review

klakhov (approved):

PR #10494, opened 2026-04-18
Tests: stabilize local Python and Cypress baselines

nmanovic:
## Summary
This is the first PR in a planned series that splits the large `tests-infra-kube-stable-v2` branch into smaller, reviewable changes.

This PR is intentionally narrow. It only fixes baseline test instability that blocks reliable validation of the later runtime and CI refactors.

## What changed
- relaxed the REST video-frame comparison tolerance slightly to handle small local decode variance
- replaced brittle clipboard shortcut simulation in the raw-labels E2E flow with the editor paste path while keeping the UI responsible for stripping server-assigned IDs
- removed two `intercept(...)` -> `wait(...)` races in task-opening Cypress flows
- stabilized bulk-actions E2E checks by waiting on durable request/results state instead of transient notifications/progress wrappers
- hardened Cypress video cleanup so missing video files do not turn passing specs into failed runs

## Files
- `tests/python/rest_api/_test_base.py`
- `tests/cypress/e2e/actions_tasks/case_117_paste_labels_from_another_task.js`
- `tests/cypress/e2e/actions_tasks/task_rectangles_only.js`
- `tests/cypress/e2e/features/skeletons_pipeline.js`
- `tests/cypress/e2e/features/bulk_actions.js`
- `tests/cypress/plugins/index.js`

## Why this PR exists
The split PR series needs a stable baseline first. Without these fixes, later PRs in the series keep failing for unrelated local/Cypress flakiness, which makes review and validation noisy.

## PR series
This is **PR 1** in the split series.
Planned follow-ups will cover backend fixes, Helm/OPA fixes, dependency compatibility, pytest runtime foundation, parallel runtime, kube runtime, and CI/docs.

PR #10495, opened 2026-04-18
Backend: harden chunk job errors and S3 status timeouts

nmanovic:
## Summary
This is **PR 2** in the split series for the large test-infra branch.

Base PR:
- #10494 `Tests: stabilize local Python and Cypress baselines`

This PR is intentionally small and backend-only. It builds on top of PR 1 because PR 1 contains the baseline test fixes needed to validate the later series reliably.

## What changed
- harden chunk job failure reconstruction in `cvat/apps/engine/cache.py`
  - if rebuilding the original exception from stored type/args fails, raise a stable `ChunkCreationError` instead of masking the real failure with a secondary constructor error
- make S3 cloud-storage status probes fail fast in `cvat/apps/engine/cloud_provider.py`
  - use a dedicated low-timeout client for status checks
  - degrade only transport-level reachability failures (`ConnectTimeoutError`, `EndpointConnectionError`, `ReadTimeoutError`)
  - keep `ClientError` handling separate so existing `403/404` semantics are preserved

## Why this PR exists separately
These changes are independent from the pytest runtime / kube / CI refactor. They are backend robustness fixes and can be reviewed on their own.

## Validation
Local checks completed on this branch:
- `py_compile`
- `black --check --diff`
- `isort --check --diff --resolve-all-configs`
- `pylint`
- `typos` on the changelog fragment
- `remark` on the changelog fragment

## Changelog
- added a dedicated PR fragment for the backend robustness changes

### DIFF
--- a/changelog.d/20260418_160500_nmanovic_backend_timeout_and_chunk_failures.md
+++ b/changelog.d/20260418_160500_nmanovic_backend_timeout_and_chunk_failures.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- Backend chunk job failures and S3 cloud-storage health checks are now more robust by preserving meaningful chunk creation errors and by making S3 status probes fail fast on unreachable or misconfigured endpoints.
--- a/changelog.d/20260418_160500_nmanovic_backend_timeout_and_chunk_failures.md
+++ b/changelog.d/20260418_160500_nmanovic_backend_timeout_and_chunk_failures.md
@@ -1,0 +1,7 @@
+### Fixed
+
+- Backend chunk job failures now preserve more meaningful exception details
+  (<https://github.com/cvat-ai/cvat/pull/10495>)
+- S3 cloud-storage status probes now fail fast on unreachable or misconfigured
+  endpoints
+  (<https://github.com/cvat-ai/cvat/pull/10495>)
--- a/cvat/apps/engine/cache.py
+++ b/cvat/apps/engine/cache.py
@@ -77,13 +77,13 @@ class ChunkCreationError(Exception):
 def _build_chunk_job_failure_exception(
     rq_job: rq.job.Job, job_meta: RQMetaWithFailureInfo
 ) -> Exception:
-    exc_type = job_meta.exc_type or Exception
+    exc_type = job_meta.exc_type or ChunkCreationError
     exc_args = job_meta.exc_args or ("Cannot create chunk",)
 
     try:
         return exc_type(*exc_args)
     except TypeError:
-        exc_name = getattr(exc_type, "__name__", repr(exc_type))
+        exc_name = exc_type.__name__
         details = job_meta.formatted_exception or repr(exc_args)
         return ChunkCreationError(
             f"Chunk job {rq_job.id} failed with {exc_name}: {details.strip()}"
--- a/cvat/apps/engine/cache.py
+++ b/cvat/apps/engine/cache.py
@@ -70,6 +70,26 @@ class CacheTooLargeDataError(Exception):
     pass
 
 
+class ChunkCreationError(Exception):
+    pass
+
+
+def _build_chunk_job_failure_exception(
+    rq_job: rq.job.Job, job_meta: RQMetaWithFailureInfo
+) -> Exception:
+    exc_type = job_meta.exc_type or Exception
+    exc_args = job_meta.exc_args or ("Cannot create chunk",)
+
+    try:
+        return exc_type(*exc_args)
+    except TypeError:
+        exc_name = getattr(exc_type, "__name__", repr(exc_type))
+        details = job_meta.formatted_exception or repr(exc_args)
+        return ChunkCreationError(
+            f"Chunk job {rq_job.id} failed with {exc_name}: {details.strip()}"
+        )
+
+
 def enqueue_create_chunk_job(
     queue: rq.Queue,
     rq_job_id: str,
@@ -110,9 +130,7 @@ def wait_for_rq_job(rq_job: rq.job.Job):
         elif job_status in ("failed",):
             rq_job.get_meta()  # refresh from Redis
             job_meta = RQMetaWithFailureInfo.for_job(rq_job)
-            exc_type = job_meta.exc_type or Exception
-            exc_args = job_meta.exc_args or ("Cannot create chunk",)
-            raise exc_type(*exc_args)
+            raise _build_chunk_job_failure_exception(rq_job, job_meta)
 
         time.sleep(settings.CVAT_CHUNK_CREATE_CHECK_INTERVAL)
         retries -= 1
--- a/cvat/apps/engine/cloud_provider.py
+++ b/cvat/apps/engine/cloud_provider.py
@@ -619,12 +619,6 @@ class S3CloudStorage(AbstractCloudStorage):
         # misconfigured. Keep a dedicated low-timeout client for head_bucket, while
         # the regular resource/client retain their standard retry behavior for normal
         # storage operations.
-        status_config = Config(
-            proxies=PROXIES_FOR_UNTRUSTED_URLS or {},
-            connect_timeout=2,
-            read_timeout=5,
-            retries={"total_max_attempts": 1, "mode": "standard"},
-        )
         self._s3 = session.resource(
             "s3",
             endpoint_url=endpoint_url,
@@ -639,7 +633,16 @@ class S3CloudStorage(AbstractCloudStorage):
                 ),
             ),
         )
-        self._status_client = session.client("s3", endpoint_url=endpoint_url, config=status_config)
+        self._status_client = session.client(
+            "s3",
+            endpoint_url=endpoint_url,
+            config=Config(
+                proxies=PROXIES_FOR_UNTRUSTED_URLS or {},
+                connect_timeout=2,
+                read_timeout=5,
+                retries={"total_max_attempts": 1, "mode": "standard"},
+            ),
+        )
 
         # anonymous access
         if not any([access_key_id, secret_key, session_token]):
--- a/cvat/apps/engine/cloud_provider.py
+++ b/cvat/apps/engine/cloud_provider.py
@@ -24,7 +24,12 @@ from azure.storage.blob import BlobServiceClient, ContainerClient
 from azure.storage.blob._list_blobs_helper import BlobPrefix
 from boto3.s3.transfer import TransferConfig
 from botocore.client import Config
-from botocore.exceptions import ClientError, EndpointConnectionError
+from botocore.exceptions import (
+    ClientError,
+    ConnectTimeoutError,
+    EndpointConnectionError,
+    ReadTimeoutError,
+)
 from botocore.handlers import disable_signing
 from django.conf import settings
 from google.api_core.exceptions import RetryError
@@ -609,6 +614,21 @@ class S3CloudStorage(AbstractCloudStorage):
                 kwargs[key] = arg_v
 
         session = boto3.Session(**kwargs)
+        # Status checks are part of the control plane, not the data-transfer path, so
+        # they must fail fast when the endpoint is unreachable or misconfigured. Some
+        # network setups return a transport timeout instead of an immediate connection
+        # refusal for a wrong host/port, which can make a simple bucket health probe
+        # hang long enough to block API requests and propagate failures to unrelated
+        # operations. Use a dedicated low-timeout client for head_bucket/head_object,
+        # while the regular resource/client keep the default pool and retry behavior
+        # for real uploads/downloads. Tests with intentionally broken endpoints confirm
+        # that this distinction is required in practice.
+        status_config = Config(
+            proxies=PROXIES_FOR_UNTRUSTED_URLS or {},
+            connect_timeout=2,
+            read_timeout=5,
+            retries={"max_attempts": 1, "mode": "standard"},
+        )
         self._s3 = session.resource(
             "s3",
             endpoint_url=endpoint_url,
@@ -623,10 +643,12 @@ class S3CloudStorage(AbstractCloudStorage):
                 ),
             ),
         )
+        self._status_client = session.client("s3", endpoint_url=endpoint_url, config=status_config)
 
         # anonymous access
         if not any([access_key_id, secret_key, session_token]):
             self._s3.meta.client.meta.events.register("choose-signer.s3.*", disable_signing)
+            self._status_client.meta.events.register("choose-signer.s3.*", disable_signing)
 
         self._client = self._s3.meta.client
         self._bucket = self._s3.Bucket(bucket)
@@ -641,9 +663,14 @@ class S3CloudStorage(AbstractCloudStorage):
         return self._bucket.name
 
     def _head(self):
-        return self._client.head_bucket(Bucket=self.name)
+        # Bucket status belongs to the fast-fail health path. Do not route it through
+        # the main transfer client, otherwise a bad endpoint can delay normal API work.
+        return self._status_client.head_bucket(Bucket=self.name)
 
     def _head_file(self, key: str, /):
+        # Metadata reads are used by normal storage logic as well, not just endpoint
+        # health checks. Keep them on the regular client so they retain normal retry
+        # behavior on slower S3-compatible backends.
         return self._client.head_object(Bucket=self.name, Key=key)
 
     def get_status(self):
@@ -658,7 +685,7 @@ class S3CloudStorage(AbstractCloudStorage):
                 return Status.FORBIDDEN
             else:
                 return Status.NOT_FOUND
-        except EndpointConnectionError:
+        except (ConnectTimeoutError, EndpointConnectionError, ReadTimeoutError):
             slogger.glob.warning(
                 f"CloudStorage S3 {self._client.meta.endpoint_url}, {self.name} not available",
                 exc_info=True,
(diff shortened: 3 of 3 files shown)


nmanovic:
Updated this PR after rebasing it onto the merged PR0/develop state.

Review fix applied in `cvat/apps/engine/cloud_provider.py`:
- changed the status client retry config from `{"max_attempts": 1, "mode": "standard"}` to `{"total_max_attempts": 1, "mode": "standard"}`
- this makes the status probe a true single-attempt request instead of allowing an implicit retry

The PR diff is now back to the intended scope only:
- `cvat/apps/engine/cache.py`
- `cvat/apps/engine/cloud_provider.py`
- `changelog.d/20260418_160500_nmanovic_backend_timeout_and_chunk_failures.md`

cvat/apps/engine/cache.py
```diff
@@ -70,6 +70,26 @@ class CacheTooLargeDataError(Exception):
...
+
+def _build_chunk_job_failure_exception(
+    rq_job: rq.job.Job, job_meta: RQMetaWithFailureInfo
+) -> Exception:
+    exc_type = job_meta.exc_type or Exception
+    exc_args = job_meta.exc_args or ("Cannot create chunk",)
+
+    try:
+        return exc_type(*exc_args)
+    except TypeError:
```
SpecLad, line 85: How can this happen?

cvat/apps/engine/cache.py
```diff
@@ -70,6 +70,26 @@ class CacheTooLargeDataError(Exception):
...
+        exc_name = getattr(exc_type, "__name__", repr(exc_type))
```
SpecLad: The `getattr` call seems unnecessary; every type should have a `__name__` attribute.

cvat/apps/engine/cloud_provider.py
```diff
@@ -609,6 +614,21 @@ def __init__(
                 kwargs[key] = arg_v
 
         session = boto3.Session(**kwargs)
+        # Status checks are part of the control plane, not the data-transfer path, so
+        # they must fail fast when the endpoint is unreachable or misconfigured. Some
+        # network setups return a transport timeout instead of an immediate connection
+        # refusal for a wrong host/port, which can make a simple bucket health probe
+        # hang long enough to block API requests and propagate failures to unrelated
+        # operations. Use a dedicated low-timeout client for head_bucket/head_object,
```
SpecLad: Looks like this description is wrong: in the code, `status_client` is only used for `head_bucket`.

cvat/apps/engine/cloud_provider.py
```diff
@@ -675,6 +706,16 @@ def get_file_status(self, key: str, /):
...
+        # Only transport-level reachability timeouts belong to the degraded health
+        # path. Keep ClientError handling separate to preserve 403/404 semantics,
+        # and let unexpected SDK/runtime failures propagate instead of masking them
+        # as a storage availability problem.
+        except (ConnectTimeoutError, EndpointConnectionError, ReadTimeoutError):
+            slogger.glob.warning(
+                f"CloudStorage S3 {self._client.meta.endpoint_url}, {self.name}/{key} not available",
+                exc_info=True,
+            )
+            return Status.NOT_FOUND
```
SpecLad: This entire block is almost 1:1 to the one in `get_status`. Can you factor the common code (and comments) out?

cvat/apps/engine/cloud_provider.py
```diff
@@ -675,6 +706,16 @@ def get_file_status(self, key: str, /):
                 return Status.FORBIDDEN
             else:
                 return Status.NOT_FOUND
+        # Only transport-level reachability timeouts belong to the degraded health
+        # path. Keep ClientError handling separate to preserve 403/404 semantics,
+        # and let unexpected SDK/runtime failures propagate instead of masking them
+        # as a storage availability problem.
+        except (ConnectTimeoutError, EndpointConnectionError, ReadTimeoutError):
+            slogger.glob.warning(
+                f"CloudStorage S3 {self._client.meta.endpoint_url}, {self.name}/{key} not available",
```
SpecLad, line 708: I don't think it makes sense to put the key name in this message, because none of these exceptions are key-specific.

cvat/apps/engine/cloud_provider.py
```diff
@@ -658,7 +685,11 @@ def get_status(self):
                 return Status.FORBIDDEN
             else:
                 return Status.NOT_FOUND
-        except EndpointConnectionError:
+        # Only transport-level reachability timeouts belong to the degraded health
+        # path. Keep ClientError handling separate to preserve 403/404 semantics,
+        # and let unexpected SDK/runtime failures propagate instead of masking them
+        # as a storage availability problem.
```
SpecLad: * `EndpointConnectionError` is not a timeout, but it's handled here anyway.
* What's the "degraded health path"?
* What's the point of the second sentence here? We already see that the `ClientError` handling is separate, and nobody would expect us to _not_ propagate unexpected failures.

cvat/apps/engine/cloud_provider.py
```diff
@@ -641,9 +663,14 @@ def name(self):
...
     def _head(self):
-        return self._client.head_bucket(Bucket=self.name)
+        # Bucket status belongs to the fast-fail health path. Do not route it through
+        # the main transfer client, otherwise a bad endpoint can delay normal API work.
+        return self._status_client.head_bucket(Bucket=self.name)
 
     def _head_file(self, key: str, /):
+        # Metadata reads are used by normal storage logic as well, not just endpoint
+        # health checks. Keep them on the regular client so they retain normal retry
+        # behavior on slower S3-compatible backends.
```
SpecLad: What are "endpoint health checks"?

cvat/apps/engine/cloud_provider.py
```diff
@@ -641,9 +663,14 @@ def name(self):
         return self._bucket.name
 
     def _head(self):
-        return self._client.head_bucket(Bucket=self.name)
+        # Bucket status belongs to the fast-fail health path. Do not route it through
+        # the main transfer client, otherwise a bad endpoint can delay normal API work.
```
SpecLad: It seems rather vague. What is "normal API work"? Why isn't any other S3 call a problem for this work?

changelog.d/20260418_160500_nmanovic_backend_timeout_and_chunk_failures.md
```diff
@@ -0,0 +1,6 @@
+### Fixed
+
+- Backend chunk job failures and S3 cloud-storage health checks are now more robust
+  by preserving meaningful chunk creation errors and by making S3 status probes fail
+  fast on unreachable or misconfigured endpoints
```
SpecLad: These two changes seem to have nothing to do with each other; it would be more prudent to put them in separate bullet points.

cvat/apps/engine/cache.py
```diff
@@ -70,6 +70,26 @@ class CacheTooLargeDataError(Exception):
...
 
 
+class ChunkCreationError(Exception):
+    pass
+
+
+def _build_chunk_job_failure_exception(
+    rq_job: rq.job.Job, job_meta: RQMetaWithFailureInfo
+) -> Exception:
+    exc_type = job_meta.exc_type or Exception
```
nmanovic: Probably it is better to have here `job_meta.exc_type or ChunkCreationError`

cvat/apps/engine/cloud_provider.py
```diff
@@ -623,10 +643,12 @@ def __init__(
                 ),
             ),
         )
+        self._status_client = session.client("s3", endpoint_url=endpoint_url, config=status_config)
```
SpecLad: It would be more readable if the config was inlined in this constructor call. Right now, they're separated by an unrelated assignment.

cvat/apps/engine/cloud_provider.py
SpecLad: * "hang long enough to block API requests" - What does that mean? Any amount of hanging would block API requests.
* "and propagate failures to unrelated operations" - How? What unrelated operations?

cvat/apps/engine/cloud_provider.py
```diff
@@ -609,6 +614,21 @@ def __init__(
...
+        # hang long enough to block API requests and propagate failures to unrelated
+        # operations. Use a dedicated low-timeout client for head_bucket/head_object,
+        # while the regular resource/client keep the default pool and retry behavior
+        # for real uploads/downloads. Tests with intentionally broken endpoints confirm
+        # that this distinction is required in practice.
+        status_config = Config(
+            proxies=PROXIES_FOR_UNTRUSTED_URLS or {},
+            connect_timeout=2,
+            read_timeout=5,
+            retries={"max_attempts": 1, "mode": "standard"},
```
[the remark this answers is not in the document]
nmanovic: Fixed. I changed the status client to use `total_max_attempts=1`, so the status probe is now a true single-attempt request.

cvat/apps/engine/cache.py
nmanovic: This is defensive. We reconstruct the exception from persisted RQ metadata, and not every exception type is guaranteed to round-trip cleanly from `exc_type + exc_args`. The fallback avoids masking the original chunk failure with a secondary `TypeError`.

cvat/apps/engine/cache.py
nmanovic: Agreed. In this path `exc_type` is always an exception class now, so `__name__` is enough. Updated.

cvat/apps/engine/cloud_provider.py
nmanovic: I’d keep these two paths separate here. They differ in the operation they call and the log context they emit, and a helper would mostly turn this into a callback/context abstraction without making it clearer.

cvat/apps/engine/cloud_provider.py
nmanovic: Agreed. The dedicated client is only used for `head_bucket` here. I narrowed the wording.

cvat/apps/engine/cloud_provider.py
nmanovic: The failure itself is transport-level, but I’d still like to keep which file-status check triggered it for debugging context. I only trimmed the surrounding comments in this pass.

cvat/apps/engine/cloud_provider.py
nmanovic: Agreed. I simplified that comment. The intent is just to handle transport-level reachability failures separately from `ClientError`-based 403/404 responses.

cvat/apps/engine/cloud_provider.py
nmanovic: Fair point. I meant bucket status probes here. Reworded.

cvat/apps/engine/cloud_provider.py
nmanovic: Agreed, that wording was too vague. I rewrote it to make the distinction explicit: bucket status probes use the fast-fail client, while regular storage operations keep the standard retry behavior.

changelog.d/20260418_160500_nmanovic_backend_timeout_and_chunk_failures.md
nmanovic: Agreed. I split the changelog into separate bullets.

cvat/apps/engine/cloud_provider.py
nmanovic: Agreed. That wording was too loose. I shortened the comment so it stays concrete and only talks about the bucket status probe path.

cvat/apps/engine/cloud_provider.py
nmanovic: I’d keep `status_config` separate. The dedicated fast-fail config is the main point of the change, so naming it explicitly makes the distinction from the regular client clearer.

cvat/apps/engine/cache.py
zhiltsov-max: If you want to protect from this, I think it's better to capture any exception instead of `TypeError` and raise the new error using the `from e` syntax. If the exception class or its constructor arguments changed after the exception was saved, the error can also be `ValueError` or something else. The returned message probably should not use internal identifiers, such as class names, if it's supposed to be reported to the users.

cvat/apps/engine/cloud_provider.py
```diff
@@ -675,6 +698,14 @@ def get_file_status(self, key: str, /):
...
             else:
                 return Status.NOT_FOUND
+        # Handle transport-level reachability failures separately from ClientError-
+        # based 403/404 responses.
+        except (ConnectTimeoutError, EndpointConnectionError, ReadTimeoutError):
+            slogger.glob.warning(
+                f"CloudStorage S3 {self._client.meta.endpoint_url}, {self.name}/{key} not available",
+                exc_info=True,
+            )
+            return Status.NOT_FOUND
```
zhiltsov-max, line 711: I'm not sure it's correct to return `NOT_FOUND` in this situation. Probably, it's better to return a new error code, otherwise the calling functions will return misleading error messages. It seems like the calling functions themselves should be updated as well - they typically check for the storage availability first, then try to reach files. But it looks like a typical case of the [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) issue.

nmanovic: I agree this is not an ideal distinction. But introducing a new status and updating all callers is a broader change. This PR keeps the existing status contract and only makes the S3 probe fail fast.

cvat/apps/engine/cache.py
nmanovic: Other exceptions are theoretically possible, but exception constructors are usually simple, and the realistic failure mode here is a signature mismatch. I’d prefer to keep the narrower TypeError catch and treat all other exceptions as a mistake in the code.

cvat/apps/engine/cloud_provider.py
zhiltsov-max: Speaking about the file downloading for chunks specifically, I can see that the error handling happens like this:

```python
client.download_files():
    client.download_file() -> (assume Exception)

    # validate_bucket_status()
    client.get_status() -> (assume ok)

    # validate_file_status()
    client.get_status() -> (assume ok)
    client.get_file_status() -> (assume another Exception)
```

which by itself looks like a significant performance issue with a lot of extra actions. I'd say this code should be redesigned, but as a temporary lightweight fix it seems to be fine.

cvat/apps/engine/cache.py
SpecLad: > This is defensive. We reconstruct the exception from persisted RQ metadata, and not every exception type is guaranteed to round-trip cleanly from `exc_type + exc_args`.

Is there any example of this actually happening?

cvat/apps/engine/cloud_provider.py
SpecLad: You already have `_status_client` as an explicit name; I don't think having `status_config` in addition to that adds any information.

cvat/apps/engine/cache.py
nmanovic: I found a concrete reproducible example.

If a cloud storage has a malformed dataset manifest, the cloud storage preview path fails in the `chunks` worker with `json.decoder.JSONDecodeError`:

```text
GET /api/cloudstorages/<id>/preview
-> MediaCache.get_or_set_cloud_preview()
-> _prepare_cloud_preview()
-> ImageManifestManager(...).set_index()
-> manifest[0]
-> json.loads(...)
-> JSONDecodeError
```

In RQ metadata this exception is stored as:

```python
exc_type = json.decoder.JSONDecodeError
exc_args = ('Expecting value: line 2 column 1 (char 9)',)
```

Reconstructing it with the old code:

```python
exc_type(*exc_args)
```

raises a secondary error:

```text
TypeError: JSONDecodeError.__init__() missing 2 required positional arguments: 'doc' and 'pos'
```

I reproduced it with MinIO by uploading a bad `manifest.jsonl`, registering it as a cloud storage manifest, and requesting `/api/cloudstorages/<id>/preview`.

This is only one confirmed path. There may be other exception classes from framework/library code with the same property: their `.args` are not sufficient to call the constructor again. Since this boundary reconstructs exceptions from persisted RQ metadata, I think we should keep the guard to avoid masking the original worker failure with a secondary reconstruction error.

cvat/apps/engine/cloud_provider.py
nmanovic: Fair point. I used the variable to name the fast-fail status config, but `_status_client` and the surrounding comment already provide enough context here. Since the variable is single-use and was separated from its constructor call, I inlined the `Config(...)` into `session.client(...)`.

cvat/apps/engine/cache.py
zhiltsov-max: In this case, why we handle only `TypeError`? We don't have control over exception types in dependencies, and they can raise other error types.

nmanovic: @zhiltsov-max , I already answered the question above. I have no another answer. I don't think it is necessary.

zhiltsov-max: Yes, but I don't think the previous discussion included exceptions from foreign code. They can't be treated as errors in our code. Basically, I'd suggest to just include `ValueError` in the capture list as well to cover most scenarios.

nmanovic: @zhiltsov-max , which scenario with ValueError do you mean?

zhiltsov-max: I don't have a specific scenario of reproduction in mind, but we can potentially get such an error if there is a check for the value range in the exception constructor. We can also get `KeyError`, `AttributeError`, and `IndexError` if the passed value is used as a key in some dict or list with error messages.

nmanovic: @zhiltsov-max , if you don't have a specific example, I would prefer to close the thread.

PR #10496, opened 2026-04-18
Helm: support release-scoped OPA and secret wiring

nmanovic:
## Summary
This PR makes the Helm chart safer for running multiple CVAT releases in one Kubernetes namespace by release-scoping OPA, Redis, and Kvrocks wiring.

## What changed
- Backend pods now receive `CVAT_OPA_URL` pointing to the release-scoped OPA service.
- The OPA service now always renders as `<release>-opa-service`.
- OPA readiness/liveness probe path is configurable with `cvat.opa.probePath`; the default remains `/health?bundles`.
- Redis and Kvrocks default secret names now include the Helm release name.
- Frontend ingress service lookup now uses the root release context.
- Removed obsolete Helm docs warning about the fixed `opa` service name.

## Validation
- `git diff --check origin/develop...HEAD`
- Helm dependency update and `helm template` with CI-style values
- Helm package with dependency update

### DIFF
--- a/changelog.d/20260418_170500_nmanovic_helm_opa_release_scope.md
+++ b/changelog.d/20260418_170500_nmanovic_helm_opa_release_scope.md
@@ -1,3 +1,5 @@
 ### Fixed
 
-- Helm deployments now support release-scoped OPA and secret wiring by allowing backend pods to target a configurable OPA host, by making OPA probe paths configurable, and by rendering Redis/Kvrocks secret names per release.
+- Helm deployments now support multiple releases in one namespace by rendering
+  OPA, Redis, and Kvrocks wiring with release-scoped service and secret names
+  (<https://github.com/cvat-ai/cvat/pull/10496>)
--- a/changelog.d/20260418_170500_nmanovic_helm_opa_release_scope.md
+++ b/changelog.d/20260418_170500_nmanovic_helm_opa_release_scope.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- Helm deployments now support release-scoped OPA and secret wiring by allowing backend pods to target a configurable OPA host, by making OPA probe paths configurable, and by rendering Redis/Kvrocks secret names per release.
--- a/cvat/settings/base.py
+++ b/cvat/settings/base.py
@@ -233,7 +233,7 @@ IAM_DEFAULT_ROLE = "user"
 IAM_ADMIN_ROLE = "admin"
 # Index in the list below corresponds to the priority (0 has highest priority)
 IAM_ROLES = [IAM_ADMIN_ROLE, "user", "worker"]
-IAM_OPA_HOST = "http://opa:8181"
+IAM_OPA_HOST = os.getenv("CVAT_OPA_HOST", "http://opa:8181")
 IAM_OPA_DATA_URL = f"{IAM_OPA_HOST}/v1/data"
 LOGIN_URL = "rest_login"
 LOGIN_REDIRECT_URL = "/"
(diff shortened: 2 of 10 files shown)


helm-chart/values.yaml
```diff
@@ -260,6 +260,11 @@ cvat:
     #     name: tmp
     #     subPath: test
     composeCompatibleServiceName: true # Sets service name to opa in order to be compatible with Docker Compose. Necessary because changing IAM_OPA_DATA_URL via environment variables in current images. Hinders multiple deployment due to duplicate name
```
SpecLad: It seems like we should be able to just delete this option. I don't think there's any benefit in setting it to true if the chart works when it's set to false.

changelog.d/20260418_170500_nmanovic_helm_opa_release_scope.md
```diff
@@ -0,0 +1,3 @@
+### Fixed
+
+- Helm deployments now support release-scoped OPA and secret wiring by allowing backend pods to target a configurable OPA host, by making OPA probe paths configurable, and by rendering Redis/Kvrocks secret names per release.
```
SpecLad: > Helm deployments now support release-scoped OPA and secret wiring

What does this mean for the user? It's not clear at all from the note.

It's under the "Fixed" category - what was the problem that was fixed?

> and by rendering Redis/Kvrocks secret names per release.

This was already possible, since the secret names were customizable.

More generally:

* You need to add a link to the PR, as per the template.
* Please limit the line length to something reasonable, like 80-100 characters.

cvat/settings/base.py
```diff
@@ -233,7 +233,7 @@ def generate_secret_key():
 IAM_ADMIN_ROLE = "admin"
 # Index in the list below corresponds to the priority (0 has highest priority)
 IAM_ROLES = [IAM_ADMIN_ROLE, "user", "worker"]
-IAM_OPA_HOST = "http://opa:8181"
+IAM_OPA_HOST = os.getenv("CVAT_OPA_HOST", "http://opa:8181")
```
SpecLad: IMO, a better name for this would be `CVAT_OPA_URL`, since it's not just a hostname.

helm-chart/values.yaml
```diff
@@ -260,6 +260,11 @@ cvat:
...
+    # Keep the default probe tied to bundle availability in normal deployments.
+    # This makes OPA health reflect whether it has successfully loaded policy data
+    # from the CVAT backend. Test environments may override this to plain `/health`
+    # to avoid false-negative restarts while backend services are still warming up.
```
SpecLad: If plain `/health` is good enough for test environments, why not use it in prod too? It seems reasonable not to punish OPA for failing to load the bundle from the backend if the backend is dead.

helm-chart/values.yaml
yaizkazani: I agree, makes little sense.

nmanovic: Agreed. I removed `cvat.opa.composeCompatibleServiceName`; the chart now always renders the OPA service as `<release>-opa-service`, and backend pods receive `CVAT_OPA_URL` pointing to that release-scoped service.

changelog.d/20260418_170500_nmanovic_helm_opa_release_scope.md
nmanovic: Agreed. I rewrote the changelog entry to describe the fixed multi-release Helm behavior, wrapped the text, and added the PR link.

cvat/settings/base.py
nmanovic: Agreed. I renamed the public env var to `CVAT_OPA_URL` and also renamed the internal base setting to `IAM_OPA_URL`. I kept `IAM_OPA_DATA_URL` unchanged because it is already the derived OPA data endpoint and is used broadly by permission classes.

helm-chart/values.yaml
nmanovic: I kept `/health?bundles` as the default because it preserves the existing Helm behavior and is the stronger readiness signal for CVAT: OPA should not receive authorization checks until the CVAT policy bundle is loaded. Plain `/health` remains available through `cvat.opa.probePath` for deployments that intentionally want process-health-only probes.

PR #10510, opened 2026-04-21
Audio client

filichabout:
commit for present only


cvat-ui/src/components/annotation-page/audio-workspace/audio-regions-list.tsx
zhiltsov-max, line 1: When zoomed-in on an interval and trying to move a boundary of the interval, there's no automatic waveform scroll. I think it would be convenient to allow such scroll, because otherwise you either need to use shift+mouse wheel or drop the boundary and click the scrollbar.

zhiltsov-max, line 1: When trying to create an interval by pressing C, the label selection pop-up always requires to select the label and click "Create". It is different from what we have in the image editing modes with the N key, which immediately starts the annotation with the last selected label. I think, it should be the same as in image editing, because this label selection is not very convenient at the moment.

zhiltsov-max, line 1: The list of folded/expanded annotation properties is somehow bound to the annotation label and even remembered. I can't say it's very intuitive behavior. Probably, it's better to always use the same folding regardless of the annotation label.

https://github.com/user-attachments/assets/66dfafe6-ff67-477d-bf93-5c0662de3fc5

zhiltsov-max, line 1: As we discussed earlier, the header in the annotation editing window should probably be redesigned to show 
`<annotation id> <label>      <start> <end> (<duration>)`

I think, the trash can button should also be removed.

Image

zhiltsov-max, line 1: Minimap behaves differently than the scrollbar, when trying to move the mouse while holding the LMB. They should work the same way.

https://github.com/user-attachments/assets/821820b5-0ec3-4c8a-b348-0eb7a6ca4a25

zhiltsov-max, line 1: Per-instance colors work only for some annotations. Here, the first 3 annotations were loaded, and the next 3 were created on this page, if it matters.

https://github.com/user-attachments/assets/e3dc0309-c850-4143-a3e4-98dd2993feca

zhiltsov-max, line 1: Image

Some issue with the source field, I don't know what it is and what it's supposed to display,

zhiltsov-max, line 1: Annotation filtering does not seem to work.

zhiltsov-max, line 1: I'm editing the red annotation and make a single LMB click in the green area, the selected annotation changes, and the view is centered on the green one. 
- There's still the issue with playback head placement without touching the annotation, if there's one. 
- Sudden viewport movement after just 1 click hurts productivity, at the very least.

As a solution, I think the viewport should not be centered on an annotation after just by 1 click - it should be a double click instead, as in the image editor. The playback head could be repositioned by some modifier, e.g. ctrl.

https://github.com/user-attachments/assets/50df5348-2e99-47de-9303-6c39949d6c7b

zhiltsov-max, line 1: When trying to leave the annotation editing page, if always fires the confirmation dialog for unsaved changes, even if the page was just opened and no editing took place.

zhiltsov-max, line 1: Sometimes the sound played mismatches the position. I'm not totally sure how to reproduce it reliably. For example, in this case:

Image

If I put a playback head before the interval, it plays one sound, but if I draw an interval in the same place and try to play it, a nearby fragment plays. The noticed deviation is < 2 seconds.

zhiltsov-max, line 1: When trying to resize using the shift+mouse scroll from a zoomed-in state, zooming-out moves the viewport to the start first instead of smooth rescaling.

https://github.com/user-attachments/assets/2b5eaa40-8b21-4aa9-a007-f508ae97c8ce

cvat-ui/src/components/create-task-page/quality-configuration-form.tsx
zhiltsov-max, line 1: Please remove the quantity % field for audio GT, it's not needed.

docker-compose.yml
```diff
@@ -13,6 +13,7 @@ x-backend-env: &backend-env
   CVAT_REDIS_ONDISK_PORT: 6666
   CVAT_LOG_IMPORT_ERRORS: 'true'
   CVAT_ALLOW_STATIC_CACHE: '${CVAT_ALLOW_STATIC_CACHE:-no}'
+  CVAT_DEFAULT_BACKING_CS_ID:
```
bsekachev, line 16: Please make sure all changes are presented in pull request related to this pull request

cvat-core/src/annotations-actions/base-collection-action.ts
```diff
@@ -18,7 +18,7 @@ import { BaseAction, prepareActionParameters, validateClientIDs } from './base-a
 export interface CollectionActionInput {
     onProgress(message: string, percent: number): void;
     cancelled(): boolean;
-    collection: Pick<SerializedCollection, 'shapes' | 'tags' | 'tracks'>;
+    collection: Pick<SerializedCollection, 'shapes' | 'tags' | 'tracks' | 'intervals'>;
```
bsekachev: I believe current annotation actions does not handle intervals at all.

cvat-core/src/annotations-actions/base-collection-action.ts
```diff
@@ -148,7 +149,7 @@ export async function call(
             }
 
             return acc;
-        }, { shapes: [], tags: [], tracks: [] });
+        }, { shapes: [], tags: [], tracks: [], intervals: [] });
```
bsekachev: `intervals` unused

bsekachev (commented):
Please update the pull request. 
Usually we do not review any PRs where CI is failed, they are considered as not ready. I will, nevertheless, continue code review, however without working CI it cannot be merged.

cvat-core/src/annotations-actions/base-shapes-action.ts
```diff
@@ -143,8 +144,8 @@ export async function run(
         }
 
         await instance.annotations.commit(
-            { shapes: totalUpdates.created.shapes, tags: [], tracks: [] },
-            { shapes: totalUpdates.deleted.shapes, tags: [], tracks: [] },
+            { shapes: totalUpdates.created.shapes, tags: [], tracks: [], intervals: [] },
```
bsekachev: one small idea is to make keys in `commit` optional and provide empty default values internally. Thus, external code will be simpler

cvat-core/src/annotations-collection.ts
```diff
@@ -151,28 +155,40 @@ export default class Collection {
             }
         }
 
+        for (const interval of (data.intervals || [])) {
```
bsekachev: Why do we need these defaults `|| []`?

cvat-core/src/annotations-collection.ts
```diff
@@ -243,8 +261,11 @@ export default class Collection {
         const { tracks } = this;
         const shapes = this.shapes[frame] ?? [];
         const tags = this.tags[frame] ?? [];
+        const intervals = this.intervals.filter(
+            (interval) => frame >= interval.start && (interval.stop === null || frame <= interval.stop),
```
bsekachev: What ``interval.stop ===  null`` means and when is it used?

cvat-core/src/annotations-collection.ts
```diff
@@ -881,6 +907,10 @@ export default class Collection {
                 this.tags[frame] = [];
             }
 
+            this.intervals = this.intervals.filter(
+                (interval) => interval.start > to || (interval.stop !== null && interval.stop < from),
+            );
```
bsekachev, line 1062: When shall we remove interval? 
When is it fully in deleted interval or partially? What do you think and why?

cvat-core/src/annotations-filter.ts
```diff
@@ -74,6 +74,7 @@ interface ConvertedObjectData {
     width: number | null;
     height: number | null;
     rotation: number | null;
+    duration?: number | null;
```
bsekachev: unused?

cvat-core/src/annotations-filter.ts
```diff
@@ -227,6 +243,21 @@ export default class AnnotationsFilter {
                     votes: null,
                 };
             }),
+            intervals: (collection.intervals ?? []).map((interval): ConvertedIntervalData => {
+                const label = labelByID[interval.label_id];
+
+                return {
+                    duration: interval.stop != null ? interval.stop - interval.start : null,
```
bsekachev: Perhaps it should return type: ObjectType.INTERVAL. 
Otherwise in runtime it will be more difficult to recognize the object

cvat-core/src/annotations-filter.ts
```diff
@@ -245,12 +276,15 @@ export default class AnnotationsFilter {
...
         labelsSpec: Label[],
         filters: object[],
-    ): { shapes: number[]; tags: number[]; tracks: number[]; } {
+    ): { shapes: number[]; tags: number[]; tracks: number[]; intervals: number[]; } {
         if (!filters.length) {
             return {
                 shapes: collection.shapes.map((shape) => shape.clientID),
                 tags: collection.tags.map((tag) => tag.clientID),
                 tracks: collection.tracks.map((track) => track.clientID),
+                intervals: (collection.intervals ?? []).map((interval) => interval.id).filter(
```
bsekachev: Why do we need ``?? []``?

cvat-core/src/annotations-filter.ts
```diff
@@ -245,12 +276,15 @@ export default class AnnotationsFilter {
...
+                    (id): id is number => id != null,
```
bsekachev: Why have you added id filtering here?

cvat-core/src/annotations-filter.ts
```diff
@@ -262,6 +296,10 @@ export default class AnnotationsFilter {
                 .filter((_, index) => jsonLogic.apply(filters[0], converted.tags[index])),
             tracks: converted.tracks.map((shape) => shape.objectID)
                 .filter((_, index) => jsonLogic.apply(filters[0], converted.tracks[index])),
+            intervals: converted.intervals
+                .filter((_, index) => jsonLogic.apply(filters[0], converted.intervals[index]))
+                .map((interval) => interval.serverID)
+                .filter((id): id is number => id != null),
```
bsekachev: Why have you added id filtering here?

cvat-core/src/annotations-objects.ts
```diff
@@ -1560,6 +1560,118 @@ export class Tag extends Annotation {
...
+            result.id = this.serverID;
+        }
+
+        return result;
+    }
+
+    public get(frame: number): Omit<Required<SerializedData>,
+    'elements' | 'occluded' | 'outside' | 'rotation' | 'zOrder' |
+    'points' | 'hidden' | 'pinned' | 'keyframe' | 'shapeType' |
+    'parentID' | 'descriptions' | 'keyframes'
```
bsekachev: I believe we just need a dedicated type SerializedInterval

cvat-core/src/annotations-objects.ts
```diff
@@ -1560,6 +1560,118 @@ export class Tag extends Annotation {
     }
 }
 
+export class Interval extends Annotation {
```
bsekachev, line 1573: I am not sure that `frame` in this class makes sense. If we replace it by `0` everywhere, will it change anything? 

Also, not sure Annotation is a good base class exactly because of `frame`. Now it may be ok, but later we probably need to refactor.

cvat-core/src/annotations-saver.ts
```diff
@@ -285,13 +297,17 @@ export default class AnnotationsSaver {
...
+                const frameKey = 'frame' in objectToSave ? objectToSave.frame : ('start' in objectToSave ? objectToSave.start : undefined);
 
                 // optimization to avoid stringifying each object in collection
-                const potentialObjects = collection.filter(
-                    (object) => object.frame === frame &&
-                    object.label_id === labelID &&
-                    !existingIDs.includes(object.id), // exclude objects that client knows
+                const potentialObjects = (collection as CollectionObject[]).filter(
+                    (object) => {
+                        const objFrame = 'frame' in object ? object.frame : ('start' in object ? object.start : undefined);
```
bsekachev: How is `undefined` case possible?

cvat-core/src/frames.ts
```diff
@@ -129,6 +129,8 @@ export class FramesMetaData {
 
     constructor(initialData: Omit<SerializedFramesMetaData, 'deleted_frames'> & { deleted_frames: Record<number, boolean> }) {
         const data: typeof initialData = {
+            dimension: undefined,
+            media_type: undefined,
```
bsekachev: How these fields are used in the class?

cvat-core/src/quality-conflict.ts
```diff
@@ -9,6 +9,11 @@ export enum QualityConflictType {
     EXTRA = 'extra_annotation',
     MISMATCHING = 'mismatching_label',
     MISSING = 'missing_annotation',
+    LOW_OVERLAP = 'low_overlap',
+    MISMATCHING_DIRECTION = 'mismatching_direction',
+    MISMATCHING_ATTRIBUTES = 'mismatching_attributes',
+    MISMATCHING_GROUPS = 'mismatching_groups',
+    COVERED_ANNOTATION = 'covered_annotation',
```
bsekachev: These conflict types are relevant for audio quality?  @zhiltsov-max

zhiltsov-max: Only `MISMATCHING_ATTRIBUTES`. Maybe `LOW_OVERLAP` can also be adapted, but it's not supposed yet.

cvat-ui/src/components/create-task-page/create-task-page.tsx
zhiltsov-max, line 1: Please add handling for audio media selection in the old task creation dialog. I think there should either be automatic switching to the new dialog, a notification with such suggestion, or an error. The old task creation dialog cannot handle audio task creation.

bsekachev:
The pull request does not run audio workspace tests because they are missing in CI matrix: https://github.com/cvat-ai/cvat/blob/6892f5ffdd615ff32917eacce472504f905124c5/.github/workflows/main.yml#L309 

full.yml also shall be updated

cvat-core/src/audio/audio-proxy.ts
```diff
@@ -0,0 +1,23 @@
+// Copyright (C) CVAT.ai Corporation
```
bsekachev: What is the purpose of this wrapper?

cvat-core/src/audio/audio-response-types.ts
```diff
@@ -0,0 +1,5 @@
+// Copyright (C) CVAT.ai Corporation
```
bsekachev: Why do you need this empty file?

cvat-core/src/quality-conflict.ts
bsekachev: @filichabout may you please explain the change?

cvat-core/src/audio/audio-proxy.ts
```diff
@@ -0,0 +1,23 @@
...
+//
+// SPDX-License-Identifier: MIT
+
+import { ChunkQuality } from 'cvat-data';
+import serverProxy from '../server-proxy';
+
+interface AudioChunkResponse {
+    data: ArrayBuffer;
+    contentOffset: number;
+}
```
bsekachev: Interface duplicated, you have the same in server proxy

bsekachev (commented):
Consider reducing number of added custom icons
Many of them may be replaced by default antd icons: https://ant.design/components/icon

cvat-ui/src/utils/hex-color.ts
```diff
@@ -0,0 +1,31 @@
+const FALLBACK_HEX = 'ffffff';
+
+export function normalizeHex(hex: string): string {
```
bsekachev: unused outside

cvat-ui/src/utils/hex-color.ts
```diff
@@ -0,0 +1,31 @@
...
+    if (clean.length === 3) {
+        clean = clean.split('').map((ch) => ch + ch).join('');
+    }
+    if (!/^([0-9a-f]{6})$/i.test(clean)) {
+        return FALLBACK_HEX;
+    }
+    return clean;
+}
+
+export function hexToRgb(hex: string): { r: number; g: number; b: number } {
```
bsekachev: unused outside

cvat-ui/src/utils/hex-color.ts
```diff
@@ -0,0 +1,31 @@
+const FALLBACK_HEX = 'ffffff';
```
bsekachev, line 5: license header missing

cvat-ui/src/utils/format-audio-time.ts
```diff
@@ -0,0 +1,7 @@
+export function formatTimeShort(seconds: number): string {
```
bsekachev, line 5: missing license header

cvat-ui/src/utils/feature-flags.ts
```diff
@@ -0,0 +1,20 @@
...
+    };
+}
+
+export function useAudioAnnotationsEnabled(): boolean {
+    return useSelector((state: CombinedState) => {
+        const server = state.about.server as ServerWithFeatures | null;
+        const flag = server?.features?.audio_annotations;
+        return flag !== false;
+    });
+}
```
bsekachev: Can you explain the code? I can't really understand why we need it.

bsekachev:
@codex review

cvat-ui/src/actions/annotation-actions.ts
```diff
@@ -194,6 +224,356 @@ export function setHoveredChapter(id: number | null): AnyAction {
...
+
+export function loadAudioAnnotationsAsync(): ThunkAction {
+    return async (dispatch: ThunkDispatch, getState): Promise<void> => {
+        const { job: { instance: jobInstance } } = getState().annotation;
+        if (!jobInstance) {
+            dispatch({
+                type: AnnotationActionTypes.LOAD_AUDIO_ANNOTATIONS_SUCCESS,
+                payload: { regions: [], version: 0 },
+            });
+            return;
```
bsekachev: If job instance not in state, dispatch LOAD_AUDIO_ANNOTATIONS_SUCCESS? What does it means?

cvat-ui/src/actions/annotation-actions.ts
```diff
@@ -194,6 +224,356 @@ export function setHoveredChapter(id: number | null): AnyAction {
...
+                    serverId: interval.id,
+                    source: String(interval.source),
+                    group: interval.group,
+                    color: pickInstanceColor(acc),
+                    zOrder: index,
+                    locked: false,
+                    hidden: false,
+                });
+                return acc;
+            }, []);
```
bsekachev, line 417: Why you added intervals to `cvat-core` if here you work with just serialized server data?

cvat-ui/src/actions/annotation-actions.ts
```diff
@@ -194,6 +224,356 @@ export function setHoveredChapter(id: number | null): AnyAction {
...
+            }));
+
+            const response = await serverProxy.annotations.updateAnnotations(
+                'job',
+                jobInstance.id,
+                {
+                    shapes: [], tracks: [], tags: [], intervals, version,
+                },
+                'put',
+            );
```
bsekachev: Incremental saving not supported?

cvat-ui/src/components/create-task-page/audio-advanced-configuration-form.tsx
```diff
@@ -0,0 +1,215 @@
+// Copyright (C) 2020-2022 Intel Corporation
```
bsekachev: I do not think Intel headers must be in this pull requet

bsekachev:
Create task has invalid hint: "You can upload an archive with images, a video, or multiple images"
Advanced configuration has start/stop frames - frames are not applicable for audio.

Mutable hint probably can be generalized for both CV and audio annotation
image

These changes are not required, but write them somewhere to your TODO list

bsekachev:
Statistics shows me unsaved changes flag (why?) and strange source

image

bsekachev:
When press Esc during recording interval - it adds interval. But the operation has to be cancelled

bsekachev:
I believe loop interval playback is quite small
image

bsekachev:
Sidebar objects must be scrolled into view when clicked on wave

tests/cypress/e2e/setup/setup_audio.js
```diff
@@ -0,0 +1,40 @@
...
+    cy.headlessLogin();
+    cy.intercept('POST', '/api/tasks**').as('createAudioTaskRequest');
+    cy.headlessCreateTask(taskSpec, dataSpec, extras).then(({ jobIDs: jobIds }) => {
+        cy.wait('@createAudioTaskRequest');
+        const jobId = jobIds[0];
+        cy.request({
+            method: 'GET',
+            url: `/api/jobs/${jobId}/data?quality=compressed&type=chunk&index=0`,
+            auth: { username: Cypress.env('user'), password: Cypress.env('password') },
+            timeout: 120000,
```
archibald1418: It could take that long?

tests/cypress/support/commands_audio.js
```diff
@@ -0,0 +1,109 @@
+// Copyright (C) CVAT.ai Corporation
+//
+// SPDX-License-Identifier: MIT
+
+/// <reference types="cypress" />
+
+const WAVEFORM_TIMEOUT = 90000;
```
archibald1418: same question
we need a threshold technical requirement to say an audio test failed, it seems that 90 seconds is a bit excessive for a test. Could it take that long?

tests/cypress/support/commands_audio.js
```diff
@@ -0,0 +1,109 @@
...
+    cy.get('.cvat-audio-extend-region-popover-content').contains('button', 'Extend').click();
+});
+
+Cypress.Commands.add('audioExtendViaHotkey', () => {
+    cy.get('body').type('{shift}E');
+});
+
+Cypress.Commands.add('audioOpenSlider', (controlClass) => {
+    cy.get(`.${controlClass}`).click();
+    cy.get('.cvat-audio-slider-popover-overlay:visible', { timeout: 5000 }).should('exist');
```
archibald1418: nit but a more conventional chai way would be:
```suggestion
            cy.get('.cvat-audio-slider-popover-overlay', { timeout: 5000 }).should('exist').and('be.visible');
```

Also it should assert visibility as well, in your version it only asserts existence

tests/cypress/support/commands_audio.js
```diff
@@ -0,0 +1,109 @@
...
+});
+
+Cypress.Commands.add('audioSliderSetValue', (controlClass, arrowDirection, steps) => {
+    cy.audioOpenSlider(controlClass);
+    cy.get('.cvat-audio-slider-popover-overlay:visible .ant-slider-handle').focus();
```
archibald1418:
```suggestion
    cy.get('.cvat-audio-slider-popover-overlay .ant-slider-handle').should('be.visible').focus();
```

`should` is better as it shows more explicitly in the logs and makes Cypress to actually wait for it to be visible. It also doesn't seem that you have an intention to filter through not visible elements. 

If you do need to filter here, use `cy.filter(':visible')` (but it doesn't seem necessary)

tests/cypress/support/commands_audio.js
```diff
@@ -0,0 +1,109 @@
...
+    for (let i = 0; i < steps; i += 1) {
+        cy.get('.cvat-audio-slider-popover-overlay:visible .ant-slider-handle').type(arrowDirection);
+    }
+    cy.get('.cvat-audio-canvas-wrapper').click('topLeft', { force: true });
+});
+
+Cypress.Commands.add('audioSaveAnnotations', () => {
+    cy.intercept('PATCH', '/api/jobs/**/annotations**').as('audioSaveRequest');
+    cy.get('.cvat-annotation-header-save-button').click();
+    cy.get('.cvat-annotation-header-save-button', { timeout: 30000 }).should('contain.text', 'Save');
```
archibald1418: default `25000` is not enough?

tests/cypress_audio.config.js
```diff
@@ -0,0 +1,21 @@
...
+// SPDX-License-Identifier: MIT
+
+const { defineConfig } = require('cypress');
+const baseConfig = require('./cypress.base.config');
+
+module.exports = defineConfig({
+    ...baseConfig,
+    e2e: {
+        ...baseConfig.e2e,
+        testIsolation: true,
```
archibald1418, line 12: Ditching state between tests is very good but that's not all, we should also strive to make each test run independently (with `.only`), that's when they are truly isolated.

bsekachev:
@codex review frontend only

cvat-core/src/audio/audio-data.ts
```diff
@@ -0,0 +1,106 @@
+import { ChunkQuality } from 'cvat-data';
```
bsekachev, line 5: missing license header

cvat-core/src/annotations-collection.ts
```diff
@@ -232,6 +269,8 @@ export default class Collection {
                 }, [])
                 .filter((tag) => !tag.removed)
                 .map((tag) => tag.toJSON()),
+            intervals: this.intervals.filter((interval) => !interval.removed)
+                .map((interval) => interval.toJSON()),
```
bsekachev: It seems getAllIntervals can be reused here

cvat-core/src/annotations-collection.ts
```diff
@@ -1009,6 +1060,10 @@ export default class Collection {
                 continue;
             }
 
+            if (object instanceof Interval) {
+                continue;
```
bsekachev, line 1191: could you please add a comment why we ignore interval in statistics method?

cvat-core/src/annotations-filter.ts
```diff
@@ -372,6 +388,22 @@ export default class AnnotationsFilter {
                     elements,
                 };
             }),
+            intervals: collection.intervals.map((interval): ConvertedIntervalData => {
+                const label = labelByID[interval.label_id];
+
+                return {
+                    duration: interval.stop != null ? interval.stop - interval.start : null,
```
bsekachev: Could you please clarify when interval.stop can be null?

cvat-core/src/session.ts
```diff
@@ -89,6 +89,22 @@ function buildDuplicatedAPI(prototype) {
...
 
+                async intervals() {
+                    const result = await PluginRegistry.apiWrapper.call(
+                        this,
+                        prototype.annotations.intervals,
+                    );
+                    return result;
+                },
+
+                async version() {
```
bsekachev: Could you please explain how you use version in cvat-ui code? Do we need this api method?

cvat-core/src/session-implementation.ts
```diff
@@ -365,6 +366,22 @@ export function implementJob(Job: typeof JobClass): typeof JobClass {
         },
     });
 
+    Object.defineProperty(Job.prototype.annotations.intervals, 'implementation', {
```
bsekachev, line 369: - The same method shall be implemented for Task class as well as we already have for other annotations
- Currently I do not mind of having a separated method to receive intervals from cvat-core, but in this case we probably do not want to return intervals from `AnnotationCollection.get`

cvat-core/src/session-implementation.ts
```diff
@@ -365,6 +366,22 @@ export function implementJob(Job: typeof JobClass): typeof JobClass {
...
+        value: async function intervalsImplementation(
+            this: JobClass,
+        ): ReturnType<typeof JobClass.prototype.annotations.intervals> {
+            return getAllIntervals(this);
+        },
+    });
+
+    Object.defineProperty(Job.prototype.annotations.version, 'implementation', {
```
bsekachev: Currently I have no idea how you use it in external code and not sure we need this method

cvat-ui/src/utils/opencv-wrapper/annotations-actions/tracker-mil.ts
```diff
@@ -55,8 +55,12 @@ export default class OpenCVTrackerMIL extends BaseCollectionAction {
...
     ): ReturnType<BaseCollectionAction['run']> {
         const noChanges = {
-            created: { shapes: [], tags: [], tracks: [] },
-            deleted: { shapes: [], tags: [], tracks: [] },
+            created: {
+                shapes: [], tags: [], tracks: [], intervals: [],
+            },
+            deleted: {
+                shapes: [], tags: [], tracks: [], intervals: [],
+            },
```
bsekachev: These are fed into `instance.annotations.commit`
As `instance.annotations.commit` now accepts partial body, it can probably be 

```
created: {},
deleted: {},
```

But check it in advance

bsekachev: Or just drop these changes. 
At the end of the method you do not return any intervals

cvat-ui/package.json
```diff
@@ -6,6 +6,8 @@
   "scripts": {
     "build": "webpack --config ./webpack.config.js",
     "start": "webpack serve --env API_URL=http://localhost:7000 --config ./webpack.config.js --mode=development",
+    "type-check": "tsc --noEmit",
+    "type-check:watch": "yarn run type-check --watch",
```
bsekachev: Have you added this?

cvat-ui/src/containers/annotation-page/standard-workspace/controls-side-bar/draw-shape-popover.tsx
```diff
@@ -38,6 +38,7 @@ interface StateToProps {
     shapeType: ShapeType;
     labels: any[];
     jobInstance: any;
+    activeSimplifyPoly?: boolean;
```
bsekachev: Is this change related to the pull request?

cvat-ui/src/actions/annotation-actions.ts
```diff
@@ -178,6 +183,31 @@ export enum AnnotationActionTypes {
     UPDATE_BRUSH_TOOLS_CONFIG = 'UPDATE_BRUSH_TOOLS_CONFIG',
     HIGHLIGHT_CONFLICT = 'HIGHLIGHT_CONFCLICT',
     HOVERED_CHAPTER = 'HOVERED_CHAPTER',
+    SWITCH_AUDIO_PLAY = 'SWITCH_AUDIO_PLAY',
```
bsekachev, line 173: Do not you think if we have a separated audio reducer, we need separated audio actions for better structure?

tests/cypress/support/commands_audio.js
```diff
@@ -0,0 +1,149 @@
...
+        cy.headlessCreateTask(taskSpec, dataSpec, extras).then(({ jobIds }) => {
+            cy.wait('@createAudioTaskRequest');
+            const jobId = jobIds[0];
+            cy.request({
+                method: 'GET',
+                url: `/api/jobs/${jobId}/data?quality=compressed&type=chunk&index=0`,
+                auth: { username: Cypress.env('user'), password: Cypress.env('password') },
+                timeout: 60000,
+                failOnStatusCode: false,
+            });
```
archibald1418: Why is this request needed?  What do we expect from it? Why can it fail? I think its response should be  validated in some way. Also, task creation can take up to 1 minute?

cvat-core/src/annotations-actions/base-collection-action.ts
filichabout: Correct, actions shouldn't be doing this. The branch was added for symmetry with shapes/tracks/tags but was never actually wired up — nothing reads its output, and it keyed on serverID while the siblings use clientID. Effectively dead from the start. Removed.

cvat-core/src/annotations-actions/base-shapes-action.ts
filichabout: Done

cvat-core/src/annotations-collection.ts
filichabout: It's an open-ended interval — no end, visible on every frame from start onwards. The backend stores it that way: the field can be null, so we keep number | null on the frontend as well. Wouldn't change this.

cvat-core/src/annotations-collection.ts
filichabout: The behavior is consistent with other types — a shape is removed entirely if its frame is in range, a track is removed entirely unless delTrackKeyframesOnly is set. Intervals follow the same pattern. Partial clipping is fundamentally different semantics: it would require generating new clientIDs, handling undo/redo, and would diverge from shapes/tracks behavior. Keeping it as is.

cvat-core/src/annotations-filter.ts
filichabout: Correct, removed it along with the entire dead interval branch.

cvat-core/src/annotations-filter.ts
filichabout: dead branch removed

cvat-core/src/annotations-filter.ts
filichabout: This was a filter for not-yet-saved intervals, but the entire branch was never read anywhere. Removed.

cvat-core/src/annotations-filter.ts
filichabout: This was a filter for not-yet-saved intervals, but the entire branch was never read anywhere. Removed.

cvat-core/src/annotations-objects.ts
filichabout: The dedicated type already exists — SerializedInterval, used in the constructor and toJSON. The Omit<…> you saw is SerializedIntervalView, the return type of get(), and it's intentionally an Omit, same as Shape/Tag/Track. Nothing needs to change. Readability could be improved by extracting a shared base view type, but that touches CV paths and doesn't belong in this PR.

cvat-core/src/annotations-objects.ts
filichabout: Yes, frame is vestigial for Interval — the constructor passes a fake frame: data.start just to satisfy the base class. It's effectively never used: get/save/clear address intervals via start/stop, and serialization doesn't emit frame at all. Splitting Annotation into a frame-less base and a framed subclass would be the right fix, but that's a change to the root of the entire CV hierarchy (Shape/Track/Tag/Skeleton*/Mask*) with a large blast radius — pulling it into an audio PR goes against the isolation principle. Planning to address this in the second phase, once the code lands in dev and we start intentionally merging audio and core — that's the right place for it. Until then frame: data.start is a conscious compromise, nothing is functionally broken.

cvat-core/src/annotations-saver.ts
filichabout: It's effectively unreachable — shapes/tags/tracks always have frame, intervals always have start. It's just a compiler fallback for the union type. Happy to replace it with a thrown error to make the invariant explicit if you prefer.

cvat-core/src/frames.ts
filichabout: dimension is an existing 2D/3D field, media_type is used at the task level to distinguish audio jobs. The undefined is just the constructor's default placeholder, overwritten from initialData on initialization.

cvat-core/src/quality-conflict.ts
filichabout: @zhiltsov-max is right — only MISMATCHING_ATTRIBUTES is relevant for audio. The other five values were accidentally pulled in from a backend sync and aren't used anywhere, reverted them. The #frame: number | null change stays — it's needed.

cvat-core/src/annotations-collection.ts
filichabout: The guard is actually needed and I've kept it (`data.intervals ?? []`). While the GET-annotations response does always include `intervals` (the `LabeledDataSerializer` contract), `Collection.import()` is a general boundary — CV flows (consensus merge, ground-truth import, task setup) call `annotations.import()` with `{ shapes, tags, tracks }` and no `intervals` key. Dropping the default crashed those with `intervals is not iterable` (caught in e2e), so the `?? []` stays.

cvat-ui/src/components/create-task-page/create-task-page.tsx
filichabout: @zhiltsov-max, @bsekachev — I'd like to align on the concept first. The audio dialog was deliberately split out so the core task creation flow has no knowledge of audio at all. Any of the three options (redirect / notification / error) reintroduces that coupling one way or another. Right now an audio file dropped into the old dialog already fails — it hits a generic image/video validation error. The question is: do we add an automatic redirect to the audio dialog — or keep core clean and improve discoverability of the audio dialog with a clearer unsupported message instead.

cvat-ui/src/actions/annotation-actions.ts
```diff
@@ -194,6 +224,362 @@ export function setHoveredChapter(id: number | null): AnyAction {
...
+
+            dispatch({
+                type: AnnotationActionTypes.SAVE_AUDIO_ANNOTATIONS_SUCCESS,
+                payload: { version: response.version ?? version },
+            });
+        } catch (error) {
+            dispatch({
+                type: AnnotationActionTypes.SAVE_AUDIO_ANNOTATIONS_FAILED,
+                payload: { error },
+            });
```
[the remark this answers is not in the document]
filichabout: Good catch. Wired SAVE_AUDIO_ANNOTATIONS_FAILED into the notifications reducer next to SAVE_ANNOTATIONS_FAILED — a failed audio save now shows the standard error banner instead of stopping silently.

cvat-ui/src/components/annotation-page/audio-workspace/audio-regions-list.tsx
```diff
@@ -0,0 +1,366 @@
...
+    const onHideAll = useCallback(() => {
+        onSetRegions(patchVisible({ hidden: true }));
+    }, [patchVisible, onSetRegions]);
+    const onShowAll = useCallback(() => {
+        onSetRegions(patchVisible({ hidden: false }));
+    }, [patchVisible, onSetRegions]);
+
+    const onDeleteRegion = useCallback((regionId: string) => {
+        onSetRegions(regions.filter((r) => r.id !== regionId));
+        if (regionId === activeRegionId) {
```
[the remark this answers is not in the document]
filichabout: Fixed. onDeleteRegion now bails on a locked region, and the Remove item in the region menu is disabled when the region is locked. The menu has no Shift modifier, so a locked region simply can't be deleted there — consistent with the keyboard flow (AUDIO_DELETE_REGION, where Shift force-deletes locked regions).

cvat-ui/src/reducers/audio-reducer.ts
```diff
@@ -0,0 +1,423 @@
...
+                ...state,
+                history: {
+                    undo: undoStack,
+                    redo: [...state.history.redo, entry].slice(-MAX_AUDIO_HISTORY),
+                },
+                player: {
+                    ...state.player,
+                    regions: applyReverseDiffs(state.player.regions, entry.diffs),
+                    activeRegionId: entry.activeRegionIdBefore,
+                    hasUnsavedChanges: true,
```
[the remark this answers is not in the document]
filichabout: Fixed. Added a savedRegions baseline (set on load and save-success); undo/redo now sets hasUnsavedChanges = !regionListsEqual(regions, savedRegions) instead of forcing true, so undoing back to the saved state clears the flag. A simple undo-stack-depth marker wasn't viable here — the history is capped by MAX_AUDIO_HISTORY and save doesn't reset it, hence the snapshot comparison.

cvat-core/src/audio/audio-data.ts
filichabout: Added

cvat-core/src/annotations-collection.ts
filichabout: Indeed — export() now calls this.getAllIntervals().

cvat-core/src/annotations-collection.ts
filichabout: statistics() builds the per-label shape/track/tag counts table, and intervals don't belong in any of those buckets — so they're skipped (this also removes the Unexpected object type crash that existed in the earlier version).

cvat-core/src/annotations-filter.ts
filichabout: That line is already gone — the entire interval branch of filterSerializedCollection was removed as dead code, so the question is moot. For the record: stop === null is an open-ended interval, mirroring the backend LabeledInterval.stop = PositiveIntegerField(null=True).

cvat-core/src/session.ts
filichabout: It's used in loadAudioAnnotationsAsync to seed state.audio.player.version, which the current PUT-based saveAudioAnnotationsAsync sends back as the version field. Load-bearing for now. It becomes unnecessary once the audio save path moves to incremental save through cvat-core's AnnotationsSaver — I'll remove it then.

cvat-core/src/session-implementation.ts
filichabout: Done both. Added Task.prototype.annotations.intervals mirroring the Job implementation, and removed intervals from Collection.get() so the per-frame object list no longer includes them. Intervals are now fetched only via the dedicated intervals() API; CV workspace consumers no longer receive audio intervals through get().

cvat-core/src/session-implementation.ts
filichabout: It's used in loadAudioAnnotationsAsync to seed state.audio.player.version, which the current PUT-based saveAudioAnnotationsAsync sends back as the version field. Load-bearing for now. It becomes unnecessary once the audio save path moves to incremental save through cvat-core's AnnotationsSaver — I'll remove it then.

cvat-ui/src/utils/opencv-wrapper/annotations-actions/tracker-mil.ts
filichabout: Dropped — reverted the file to develop. commit accepts a partial body now and the method returns no intervals, so intervals: [] and the formatting churn weren't needed.

cvat-ui/package.json
filichabout: Yes, added type-check / type-check:watch for my own convenience while working — they're unrelated to this PR, removed.

cvat-ui/src/containers/annotation-page/standard-workspace/controls-side-bar/draw-shape-popover.tsx
filichabout: No — that's an unrelated CV polygon-drawing prop that slipped in. Reverted the file to develop.

cvat-ui/src/actions/annotation-actions.ts
filichabout: Agreed, it would be cleaner. I'd like to extract a dedicated audio-actions.ts as part of the audio↔cvat-core follow-up rather than here — it's a sizable mechanical move across the reducers and containers, and this PR is already large.

tests/cypress/support/commands_audio.js
filichabout: It pre-warms the first audio chunk on the server so the waveform can decode when the job is opened — without it the first assertWaveformReady can time out while the chunk is prepared on demand. Agreed on validation: dropped failOnStatusCode: false and added .its('status').should('be.within', 200, 299), so a broken chunk endpoint now fails the test instead of being swallowed. The 60000 is a ceiling for first-time chunk prep on a cold cache, not a fixed wait — task creation itself is awaited separately via the @createAudioTaskRequest intercept.

cvat-ui/src/components/annotation-page/audio-workspace/audio-regions-list.tsx
bsekachev: @filichabout any reaction for Maxim comments?

cvat-core/src/annotations-collection.ts
bsekachev: Truly speaking I can only see `.import(` is called without `intervals` is in `tests/cypress/support/commands.js`
I believe if this is just one place, we need to fix it in this place

cvat-core/src/annotations-collection.ts
bsekachev: I see. 
@zhiltsov-max may you clarify why it was implemented this way? (`null` value to keep inverval `stop` instead of fixed timestamp meaning the end of audio)

cvat-core/src/annotations-objects.ts
bsekachev: Please add // AUDIO TODO comment

cvat-core/src/annotations-objects.ts
bsekachev: Please add // AUDIO TODO comment

cvat-core/src/annotations-saver.ts
bsekachev: Let's replace with error throwing

cvat-core/src/audio/audio-proxy.ts
bsekachev: @filichabout any reaction?

cvat-core/src/audio/audio-response-types.ts
bsekachev: @filichabout any reaction?

cvat-core/src/audio/audio-proxy.ts
bsekachev: @filichabout any reaction?

cvat-core/src/annotations-collection.ts
```diff
@@ -153,28 +157,53 @@ export default class Collection {
             }
         }
 
+        for (const interval of data.intervals ?? []) {
+            const clientID = this.injection.nextClientID();
+            const color = colors[clientID % colors.length];
+            const intervalModel = new Interval(interval, clientID, color, this.injection);
+            this.intervals.push(intervalModel);
+            this.objects[clientID] = intervalModel;
```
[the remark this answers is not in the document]
bsekachev: @filichabout codex comments are also worthy of your attention, especially classified as P1

tests/cypress/support/commands_audio.js
archibald1418: `.its('status').should('be.within', 200, 299)`
I think the assertion needs to be narrower, do we have a strict requirement: "which status code should this return on success in this particular operation? "

cvat-core/src/annotations-collection.ts
bsekachev: > and intervals don't belong in any of those buckets 

That is true
They do not belong to shape/track/tag. But on annotation view you still have statistics and currently you calculate it on component level. This is conceptually wrong approach. 

For example UI knows only about currently filtered objects. Statistics is always built across all objects. 
Thus, it should be calculated on core level.

bsekachev: So, you need to propose abstraction or minimally create a separated method audioStatistics in `cvat-core`

cvat-ui/src/actions/annotation-actions.ts
bsekachev: Please add // AUDIO TODO then

cvat-core/src/session-implementation.ts
bsekachev: As far as I know rest api just ignores this field (another speaking there are not any logic handling it).
Could you please clarify how you use it in client?

cvat-core/src/session.ts
bsekachev: > loadAudioAnnotationsAsync to seed state.audio.player.version

for what? If you remove this field, will something be changed?

bsekachev: > It becomes unnecessary once the audio save path moves to incremental save through cvat-core's AnnotationsSaver — I'll remove it then.

Also, annotation saver already have code to handle intervals.

tests/cypress/e2e/setup/setup_audio.js
filichabout: The chunk-fetch logic moved into ensureAudioTask (commands_audio.js) — setup_audio.js now just calls it. The fetch timeout there is reduced to 60s, plenty of headroom.

tests/cypress/support/commands_audio.js
filichabout: Reduced to 30 seconds.

tests/cypress/support/commands_audio.js
filichabout: Switched to .should('exist').and('be.visible').

tests/cypress/support/commands_audio.js
filichabout: Switched to .should('be.visible').focus().

tests/cypress/support/commands_audio.js
filichabout: Dropped the explicit timeout, the default is enough.

tests/cypress_audio.config.js
filichabout: Fixed. Also cleaned up the two remaining .ant-select-dropdown:visible selectors — replaced with cy.get('.ant-select-dropdown').filter(':visible'). Used .filter(':visible') rather than .should('be.visible') because antd keeps multiple dropdowns mounted in the DOM simultaneously (hidden ones included), and should('be.visible') fails on that set.

cvat-core/src/audio/audio-proxy.ts
filichabout: deleted

cvat-core/src/audio/audio-response-types.ts
filichabout: deleted

cvat-core/src/audio/audio-proxy.ts
filichabout: deleted

cvat-core/src/annotations-collection.ts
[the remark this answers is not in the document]
filichabout: Fixed — statistics() now skips Interval instances instead of throwing, so audio jobs with intervals return stats without crashing.

cvat-ui/src/reducers/audio-reducer.ts
```diff
@@ -0,0 +1,420 @@
...
+    switch (action.type) {
+        case BoundariesActionTypes.RESET_AFTER_ERROR:
+        case AnnotationActionTypes.GET_JOB_SUCCESS: {
+            const { job } = action.payload;
+            return {
+                ...state,
+                player: {
+                    ...state.player,
+                    activeLabelId: job.labels.length ? job.labels[0].id : null,
+                },
```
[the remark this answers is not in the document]
filichabout: Fixed — the audio reducer now reinitializes to defaultState on GET_JOB_SUCCESS (and RESET_AFTER_ERROR), so regions/version/history no longer carry over from the previous job — no cross-job leakage

cvat-ui/src/actions/annotation-actions.ts
filichabout: Good point — dispatching a "success" with an empty payload there was meaningless. loadAudioAnnotationsAsync now just early-returns when there's no job instance in state

cvat-ui/src/components/annotation-page/audio-workspace/audio-regions-list.tsx
filichabout: I already raised this issue before and we came to the conclusion that implementing it is unreasonably complex, as this is a flaw in the library itself. I tried to change this behavior through workarounds and direct intervention into the library. That approach already looks broken on its own, and on top of that the library still intercepts the behavior over my handler. In the end I'll have to spend N amount of time creating a workaround.

tests/cypress/support/commands_audio.js
filichabout: Removed failOnStatusCode: false and narrowed the assertion to .its('status').should('eq', 200) — a successful GET for a chunk returns exactly 200 (206/416 only apply with a Range header, and this request doesn't send one). 60000 is a ceiling for the first chunk warm-up on a cold cache, not a fixed wait (task creation is awaited separately via the @createAudioTaskRequest intercept).

cvat-core/src/annotations-saver.ts
filichabout: Done — getFrameKey now throws a DataError instead of returning undefined if the object has neither frame nor start. The invariant is now explicit.

cvat-core/src/annotations-collection.ts
filichabout: Went through all the call sites: the only one not passing intervals was the cypress helper headlessCreateObjects in tests/cypress/support/commands.js. Fixed it there (added intervals: []) and removed the ?? [] default from Collection.import — the boundary stays honest, and the Omit<SerializedCollection,'version'> type already requires intervals.

cvat-core/src/session.ts
filichabout: The field is inert, removed it. The server ignores it on PUT (there's even a # TODO: remove right on it — LabeledDataSerializer.version in cvat/apps/engine/serializers.py), and the client's state.audio.player.version was never read for any logic — it only echoed it back. Removed the version() API method, getAnnotationsVersion, and the state.audio.player.version tracking (load/save/reducer/type). The PUT body still keeps version: 0 solely to satisfy the shared SerializedCollection type that updateAnnotations requires — this field is core-level (AnnotationsSaver passes it back and forth on every CV save but also doesn't read it for logic), so I didn't touch the shared type from the audio PR.

cvat-core/src/session-implementation.ts
filichabout: Directly answering "how do you use it in client": not at all for logic. The client only stored the value in state.audio.player.version and then echoed it back in the PUT body — not a single read for any decision/comparison/check. So you're right, the field is inert on both sides: the server ignores it on PUT (your own # TODO: remove on LabeledDataSerializer.version in cvat/apps/engine/serializers.py), and the client builds nothing on top of it.
So removed it entirely: the version() method, getAnnotationsVersion, and all state.audio.player.version tracking (load/save/reducer/type). The PUT body keeps version: 0 only to satisfy the shared SerializedCollection type that updateAnnotations requires (this field is core-level — AnnotationsSaver also just passes it back and forth without reading it). Once audio saving migrates to AnnotationsSaver, version: 0 will disappear as well.

filichabout:
@codex Are there any comments that have no reply?

PR #10511, opened 2026-04-21
[DEV-849] Added '--visibility' CLI argument support for global agents to Docker Compose and Helm

yaizkazani:
- [x] I submit my changes to the `develop` branch


yaizkazani:
@SpecLad @azhavoro, could you guys please take a look? 🙏

ai-models/agents_deployment/helm/values.yaml
```diff
@@ -22,6 +22,9 @@ agent:
   # If you choose to use a secret, make sure to create it in your cluster and reference it in the `secret_env` section below.
   # Token takes precedence over the secret, so if both are provided, the value here will be used.
   cvat_access_token: ""
+  # Visibility of the function in CVAT. Can be "public" or "private". If not specified, it will default to "private".
+  # Private functions are visible to you and your organization (if specified), while public functions are visible to all users in CVAT instance.
+  visibility: "private"
```
SpecLad: It should probably be called `function_visibility` for consistency with `function_name`.

ai-models/agents_deployment/sam2/check_env.sh
```diff
@@ -60,8 +60,26 @@ resolve_function_name() {
...
+    if [ -z "$VISIBILITY" ]; then
+        echo "Warning: VISIBILITY environment variable not found. Default is private"
+        VISIBILITY_ARGS=(--visibility private)
+        echo "This function will be visible to you and ORG members if ORG_SLUG is set, but not to other users of this CVAT instance."
+    else
+        if [ "$VISIBILITY" = "public" ]; then
+            echo "This function will be visible to all users of this CVAT instance."
+            VISIBILITY_ARGS=(--visibility "$VISIBILITY")
+        else
+            echo "VISIBILITY must be private or public. Defaulting to private visibility."
```
SpecLad: This seems like unnecessary magic. Why not just pass the visibility to the function as-is? If we happen to add a new visibility setting in the future, there will be one fewer location to update.

yaizkazani: Mmmm, well, yes and no. I'd strongly prefer our code to be as foolproof as possible, so why not?

ai-models/agents_deployment/helm/values.yaml
yaizkazani: I agree, i'll change this

yaizkazani: done

ai-models/agents_deployment/sam2/check_env.sh
yaizkazani: We have agreed with @SpecLad that this is acceptable in this case.

PR #10521, opened 2026-04-24
Validate annotation source

zhiltsov-max:
### Motivation and context

Requires a UI fix for https://github.com/cvat-ai/cvat/issues/8874 - annotations in GT jobs are incorrectly saved with the `Ground truth` source.

- Added validation for the `source` field in uploaded annotations

Ideally, there could be a DB migration for the invalid DB records. This PR avoids adding one, because such a migration would be quite heavy. Instead, the old invalid annotations are allowed. New annotations are validated only, and updated annotations are fixed automatically.

### Checklist

### License


cvat/apps/engine/migrations/0099_fix_invalid_annotation_source.py
```diff
@@ -0,0 +1,19 @@
+from django.db import migrations
+
+
+def fix_invalid_source(apps, schema_editor):
+    for model_name in ("LabeledImage", "LabeledShape", "LabeledTrack"):
+        model = apps.get_model("engine", model_name)
+
+        model.objects.filter(source__iexact="Ground truth").update(source="manual")
```
bsekachev: Fixed it manually on our prod, but:

Even a simple filter + count took 340 seconds to execute. 
So, I am not sure about this migration. 

I see two questions:

1. How to fix it for other customers in order to they do not get errors when re-save incorrect objects (e.g. with updated coordinates).
2. In theory we (or our customers) may have more different values, stored in `source`. This migration does not fix them.

So, maybe we should fallback this value to "manual" when type is not valid and saved object already has `id` defined?

zhiltsov-max: Yeah, if we don't want to use a migration, we'll have to convert before writing.

cvat/apps/engine/serializers.py
```diff
@@ -3927,6 +3931,30 @@ def validate_id(self):
...
+                # Workaround for the DB records that could have been introduced by the UI before
+                # https://github.com/cvat-ai/cvat/issues/8874 was fixed.
+                # We allow the DB to store the old invalid annotations and return them from
+                # the server API, but disallow saving new ones. If the annotations were used
+                # for updating the existing annotations, we silently fix the input annotations.
+                # This is done this way to avoid heavy DB migrations.
+                source = str(models.SourceType.MANUAL)
+            else:
+                raise serializers.ValidationError(
+                    "must be one of {}, got '{}'".format(
```
SpecLad: You should include the field name in the message (probably by using `{"source": ...}`).

zhiltsov-max: Thanks, added.

SpecLad (approved):

PR #10522, opened 2026-04-24
Disallow none for annotation group and source

zhiltsov-max:
### Motivation and context

The `null` values are allowed for the `group` and `source` annotation fields, but never used:
- for `group`, 0 is used as the "no group" value. The UI sends 0, and null can only be used manually via the API
- for `source`, there is a fixed set of allowed values, and neither `null`, nor `""` are allowed or used anywhere

Related: #10521 - enforces `source` field values

This PR does not include a DB migration, because such a migration is expected to be quite long. However, a proper DB migration is desirable and should be done together with other changes to the annotation tables. 

### Checklist

### License


SpecLad:
The idea is good, but do you have any estimation for how long this migration will take with our production DB? It's a little concerning.

SpecLad:
Also, you should add a changelog entry.

SpecLad:
I tried to approximate the time the migration would take, and the results are not encouraging. :face_with_diagonal_mouth: 

```
> select count(*) from engine_labeledshape where "group" is null;
  count   
----------
 <redacted>
(1 row)

Time: 309442.248 ms (05:09.442)
```

This is just a single query on a single table, which doesn't even do the update, and it still takes over 5 minutes. The other two tables are much smaller, granted, but we're still looking at >10 minutes downtime (and I suspect that flipping fields to `NOT NULL` will add at least 5 minutes more). So I'm not sure this is viable. :-(

Maybe we should just fix the serializers for now and prevent invalid values from being added to the DB?

zhiltsov-max:
Agree, let's discuss with the rest of the team. I also think this specific change can be reclassified as optional for audio, it's a general fix.

zhiltsov-max:
Removed the DB changes from the PR.

cvat/apps/engine/serializers.py
```diff
@@ -3286,8 +3286,11 @@ class AnnotationSerializer(serializers.Serializer):
     id = serializers.IntegerField(default=None, allow_null=True)
     frame = serializers.IntegerField(min_value=0)
     label_id = serializers.IntegerField(min_value=0)
-    group = serializers.IntegerField(min_value=0, allow_null=True, default=None)
-    source = serializers.CharField(default='manual')
+    group = serializers.IntegerField(
+        min_value=0, default=0,
+        allow_null=True # backward compatibility; TODO: disallow on the DB level
```
SpecLad, line 3291: FWIW, the TODO comment should really be in the model class rather than here.

SpecLad (approved):

cvat/apps/engine/serializers.py
zhiltsov-max: Moved

PR #10523, opened 2026-04-24
Fix chunk creation for tasks with .bin files in cloud storage

SpecLad:
### Motivation and context

I inadvertently broke this in #10106 by changing the order of file operations in `PcdReader.convert_bin_to_pcd`. The current version of `read_raw_images` saves the `.bin` file with a `.pcd` extension and then tries to convert it to the PCD format. `convert_bin_to_pcd` opens the output file first; but in this case the input and output paths are the same, so this truncates the input file to 0 bytes before `convert_bin_to_pcd` can read it.

We could probably reorder the opens again to fix this, but IMO, this scheme with saving the `.bin` file with a `.pcd` extension is rather confusing and it's better to fix this instead. Save the file with its original extension and let `convert_bin_to_pcd` save the output to a new file instead.

In addition, fix a bug in `HeaderFirstDownloader` that prevents a task from being created if a `.bin` file is small. This is needed for the test to work. The problem is that `_HeaderFirstPcdDownloader.try_parse_header` always returns False, so if the entire file fits within `headers_to_try[0]` bytes, the loop will then try to download a range that doesn't exist. Fix it by detecting that there is no more data left to download and exiting early.

### How has this been tested?

Manual testing + I added an automatic test.

### Checklist

### License


SpecLad:
@zhiltsov-max Could you review this?

tests/python/rest_api/_test_base.py
```diff
@@ -394,6 +395,44 @@ def fxt_uploaded_images_task_with_gt_and_segments_and_consensus(
...
+                    pytest.mark.timeout(60),
+                    pytest.mark.xfail(
+                        to_bool(os.getenv("CVAT_ALLOW_STATIC_CACHE", False)),
+                        reason="Creating a task from a cloud .bin file with static cache doesn't work at the moment",
+                    ),
+                ],
+            )
+        ],
+    )
+    def fxt_cloud_bin_task(
```
zhiltsov-max:
```suggestion
    def fxt_cloud_bin_pointcloud_task(
```
`.bin` is almost just a memory dump that is open for the interpretation. I think it will be better to also note what this bin is.

zhiltsov-max: There are several fixture lists in the end of the class definition. I don't know if they were skipped intentionally, but maybe some of them should be updated with this new fixture.

tests/mounted_file_share/bin_pointcloud/000002.bin
zhiltsov-max, line 1: I suggest moving it to some directory. Is it from 000002.pcd?

tests/python/rest_api/_test_base.py
SpecLad: Okay, I renamed it. I also clarified the changelog entry.

SpecLad: Okay, updated.

tests/mounted_file_share/bin_pointcloud/000002.bin
SpecLad: > I suggest moving it to some directory.

Okay, moved.

> Is it from 000002.pcd?

I took it from `cvat/apps/engine/tests/assets/test_velodyne_points.zip`.

zhiltsov-max (approved):

PR #10526, opened 2026-04-24
Add server enum for task modes

zhiltsov-max:
### Checklist

### License


cvat/apps/engine/models.py
```diff
@@ -820,14 +820,20 @@ def with_job_summary(self):
...
+
 class Task(TimestampedModel, AssignableModel, FileSystemRelatedModel):
     objects = TaskQuerySet.as_manager()
 
     project = models.ForeignKey(Project, on_delete=models.CASCADE,
         null=True, blank=True, related_name="tasks",
         related_query_name="task")
     name = SafeCharField(max_length=256)
-    mode = models.CharField(max_length=32)
+    mode = models.CharField(max_length=32, choices=TaskMode.choices, default="", blank=True)
```
SpecLad, line 835: Why is blank allowed?

zhiltsov-max: For tasks without data. The mode (and some other values, later) should be determined once the task data is processed.

zhiltsov-max: There will be several similar changes soon, would you prefer `""` or `None` for such initial enum values? For me, `None` looks a little bit better, but CVAT already uses `""` in some cases.

SpecLad: Let's just stick with `""` if it's already used.

cvat/apps/engine/migrations/0100_alter_task_mode.py
```diff
@@ -0,0 +1,47 @@
+# Generated by Django 4.2.26 on 2026-04-24 17:06
+
+from django.db import migrations, models
+
+
+def validate_task_mode(apps, schema_editor):
```
SpecLad: FWIW, do we really need to do this? AFAIK, the mode was never settable by the user, so it shouldn't be possible to have any other values in the DB. Also, I don't think Django actually enforces `choices` values, so this will not prevent illegal values going forward.

cvat/apps/engine/serializers.py
```diff
@@ -2511,6 +2514,7 @@ class TaskReadSerializer(serializers.ModelSerializer):
     guide_id = serializers.IntegerField(source='annotation_guide.id', required=False, allow_null=True)
     organization_id = serializers.IntegerField(required=False, read_only=True, allow_null=True)
     dimension = serializers.CharField(allow_blank=True, required=False)
+    mode = serializers.CharField(allow_blank=True, required=False, read_only=True)
```
SpecLad, line 2523: `mode` is already in the `Meta.fields` list, so shouldn't this field be autogenerated correctly without being explicitly specified?

cvat/apps/engine/migrations/0100_alter_task_mode.py
zhiltsov-max: I can't say this validation is mandatory for anything, it just checks that the DB state is valid. I didn't find any other values in our DB as well.
Yes, the DB doesn't check choices values.

cvat/apps/engine/serializers.py
zhiltsov-max: This declaration is needed for the SDK generator. Currently, for choices it will generate an enum field, but this enum doesn't have the == str operator and requires explicit `.value` use. We should switch the server API to choicefield once SDK allows seamless transition from string to enums.

SpecLad: Okay, makes sense. Could you add a comment to explain this, though?

cvat/apps/engine/migrations/0100_alter_task_mode.py
SpecLad: > I can't say this validation is mandatory for anything, it just checks that the DB state is valid.

Yeah, my point is that it doesn't seem especially useful to check that the DB state is valid now, given that we're not preventing it from becoming invalid in the future.

It's not super important, though, so I'm fine with keeping it.

zhiltsov-max: Removed

cvat/apps/engine/serializers.py
zhiltsov-max: Added

SpecLad (approved):

PR #10538, opened 2026-04-27
Introduce task media type

zhiltsov-max:
### Motivation and context

- Added `media_type` field for tasks. It represents the media type of the frames in the task.
- Added `dimension` `media_type`, `mode` filters for jobs
- Added `media_type` filter for tasks
- Fixed the default value for `dimension`. It changes the returned value for the tasks without data, so this change is potentially breaking for some clients. The previous returned value was invalid, and long-standing tasks without data and depending business pipelines are quite rare, so it's unlikely to affect many users.

Now, the `media_type` and `mode` pair is recommended for determining task or job content in clients.

### How has this been tested?

Unit tests.

### Checklist

### License


cvat/apps/engine/migrations/0101_add_task_media_type.py
```diff
@@ -0,0 +1,54 @@
...
+        dimension="2d"
+    )
+
+
+def infer_task_media_type(apps, schema_editor):
+    Task = apps.get_model("engine", "Task")
+
+    Task.objects.filter(media_type="", dimension="3d").update(media_type="point_cloud")
+    Task.objects.filter(media_type="", mode="interpolation").update(media_type="video")
+    Task.objects.filter(media_type="", mode="annotation").update(media_type="image")
```
zhiltsov-max: It's possible that some tasks have `mode = ""`, but size > 0 and a video or images attached. These cases are very rare and it's unclear how these tasks were created. But in these cases this solution won't work. I can add another migration to fix such cases and also handle `media_type`.

cvat/apps/engine/models.py
```diff
@@ -825,28 +825,39 @@ class TaskMode(TextChoices):
     INTERPOLATION = "interpolation"
 
 
+class MediaType(TextChoices):
+    IMAGE = "image"
+    VIDEO = "video"
+    POINT_CLOUD = "point_cloud"
```
SpecLad, line 833: As I understand it, this enum is supposed to be a merged version of `dimension` and `mode`, but IMO, they should not be merged, because they represent distinct pieces of information about the task:

1. `dimension` represents the nature of a task's frame; it affects what gets returned from the `/api/tasks/<id>/data` endpoint and consequently what type of canvas will be used for the task. It affects what gets passed to a lambda function when it's called on a single frame (okay, I don't think we actually support calling functions with point clouds, but there's not really anything to stop us from adding this support).
2. `mode` indicates whether the task media is in a single file containing all frames or multiple files containing one frame each. AFAIK, this is currently used to select applicable export formats and not much else.

IMO, this enum should be a replacement for `dimension` only (in other words, `video` should not be a member). For most clients it doesn't matter whether the task was created from images or from a video, so we should make it easy for them and use the same value for both.

zhiltsov-max: It's not really supposed to be a merged version, it just happens to be so. As I said in the PR description, both dimension and mode can be inferred from the media type. It's also more extensible than both `dimension` and `mode` - e.g. there can be other 1d or 2d media types, that are not images or audio (potential examples - audio spectrogram annotation (2d), text annotation (1d)). Dimension and mode can still work at least for vision data, so there's no reason to remove them now. Historically, there were videos, then `mode` was added first to distinguish between videos and images, and then `dimension` was added to distinguish between images and point clouds. 

Now, this new field can be used to distinguish between audio and vision, but it can also be used to distinguish between the other 3 types. Then, to interpret values in the other fields correctly and to display them in the UI appropriately.

In summary, I think `video` should also be included to allow the UI switching from using `dimension` and `mode`, and maybe displaying something more video-specific eventually.

SpecLad: I understand that `dimension` is limited in extensibility, so it makes sense to replace it. I just don't think that the replacement should distinguish between images and videos, because that's a difference between storage formats, not between semantics of the data. Most clients should not care about it.

Consider a thought experiment: let's say a new point cloud format comes out which is the equivalent of a video for point clouds - i.e. it contains multiple frames, each of which is a point cloud. If we add support for this format, would we have to add a new media type? If we do that, then code that checks that `media_type == 'point_cloud'` will not work for tasks created from such videos, even if the chunks are still ZIP files of individual point clouds.

Another thought: it's not inconceivable that CVAT could add support for video classification datasets, where each "frame" is a short video clip. Wouldn't the "video" media type be better reserved for this use case?

zhiltsov-max: I think both options are interchangeable. Basically, in one case we have a single field with media_type="content_type+mode" values, and in the second case we have 2 separate fields (content_type, mode). I think, it's better to choose 2 separate fields, as seems to be a little bit more clear and extensible. I don't really like that we still need to use 2 fields (or potentially more, in the future), because it results in more complex logic, but it potentially helps with code reuse. The mode is still required, because in different modes there are likely to be different applicable chunk types and content interpretations.

Ok, let's remove `video` then. Hopefully, `mode` members will be renamed eventually to something more self-explanatory.

cvat/apps/engine/task.py
```diff
@@ -1129,6 +1129,67 @@ def _filter_cloud_storage_files(
...
+            detected_media_type = models.MediaType.IMAGE
+        elif detected_dimension == models.DimensionType.DIM_3D:
+            detected_media_type = models.MediaType.POINT_CLOUD
+        else:
+            assert False
+
+    return detected_media_type, detected_dimension
+
+
+def _validate_project_dimension(
```
SpecLad, line 1188: Should we perhaps validate the media type now rather than the dimension? It would work better if we add audio support.

cvat/apps/engine/views.py
```diff
@@ -861,14 +861,16 @@ class TaskViewSet(viewsets.GenericViewSet, mixins.ListModelMixin,
         'project_name', 'name', 'owner', 'status', 'assignee',
         'subset', 'mode', 'dimension', 'tracker_link', 'validation_mode'
     )
-    filter_fields = list(search_fields) + ['id', 'project_id', 'updated_date']
+    filter_fields = list(search_fields) + [
+        'id', 'project_id', 'updated_date', 'status', 'media_type'
```
SpecLad: `status` is redundant, since it's already in `search_fields`. But also, why not add `media_type` to `search_fields` as well?

cvat/apps/engine/task.py
zhiltsov-max: It is updated in the corresponding PR https://github.com/cvat-ai/cvat/pull/10551.

cvat/apps/engine/views.py
zhiltsov-max: Having these fields in search fields is not good in general, it's supposed for full-text search. This is fixed in https://github.com/cvat-ai/cvat/pull/10569, but the PR is not finished yet.

SpecLad: Okay, but let's at least remove the duplicate `status` until that PR.

cvat/apps/engine/views.py
```diff
@@ -861,14 +861,16 @@ class TaskViewSet(viewsets.GenericViewSet, mixins.ListModelMixin,
...
+    ]
     filter_description = textwrap.dedent("""
 
         There are few examples for complex filtering tasks:\n
             - Get all tasks from 1,2,3 projects - { "and" : [{ "in" : [{ "var" : "project_id" }, [1, 2, 3]]}]}\n
             - Get all completed tasks from 1 project - { "and": [{ "==": [{ "var" : "status" }, "completed"]}, { "==" : [{ "var" : "project_id"}, 1]}]}\n
     """)
-    simple_filters = list(search_fields) + ['project_id']
+    simple_filters = list(set(filter_fields) - {'id', 'updated_date'})
```
SpecLad, line 873: IMO, we should do the reverse: `filter_fields = list(simple_filters) + [...]`. That way, every field name only needs to be listed in one of `simple_filters` or `filter_fields`. It seems like a reasonable assumption that every simple filter should also be usable as a general filter.

zhiltsov-max: I agree, but it should be refactored in other places as well. I'd prefer to do this in another PR, possibly in https://github.com/cvat-ai/cvat/pull/10569.

cvat/apps/engine/views.py
zhiltsov-max: Ok, updated

cvat/apps/engine/views.py
SpecLad: Alright.

SpecLad (approved):

PR #10542, opened 2026-04-28
Refactor task media creation

zhiltsov-max:
### Motivation and context

Depends on #10538 

No logic changes. Factored out code for task media creation.

- Added some type annotations
- Other small improvements
- Extracted media creation into separate functions

### Checklist

### License


cvat/apps/engine/task.py
```diff
@@ -1216,6 +1206,255 @@ def _configure_chunk_types(db_task: models.Task, data: dict[str, Any]) -> None:
...
+        path=video_path.relative_to(upload_dir).as_posix(),
+        width=video_frame_size[0],
+        height=video_frame_size[1],
+    )
+
+    return video, video_length, manifest
+
+
+def _create_video_task_media_descriptors(
+    db_task: models.Task,
```
SpecLad, line 1307: Seems like this function should accept `db_data` instead, since it only uses the task to call `require_data`.

cvat/apps/engine/task.py
```diff
@@ -1216,6 +1206,255 @@ def _configure_chunk_types(db_task: models.Task, data: dict[str, Any]) -> None:
...
+    *,
+    extractor: IMediaReader,
+    media: dict[str, Any],
+    upload_dir: Path,
+    manifest_file: str | None,
+    update_status: Callable[[str], None],
+) -> tuple[models.Video, VideoManifestManager]:
+    db_data = db_task.require_data()
+
+    video, video_length, manifest = _collect_video_dataset_descriptors(
```
SpecLad, line 1318: IMO, it would be more readable to merge this function with `_collect_video_dataset_descriptors`, since almost the entire body is just this one function call.

cvat/apps/engine/task.py
```diff
@@ -1216,6 +1206,255 @@ def _configure_chunk_types(db_task: models.Task, data: dict[str, Any]) -> None:
...
+    db_task: models.Task,
+    *,
+    is_backup_restore: bool,
+    validation_params: dict[str, Any],
+    upload_dir: Path,
+    is_data_in_cloud: bool,
+    extractor: IMediaReader,
+    related_images: dict[str, Sequence[dict[str, Any]]],
+    job_file_mapping: JobFileMapping | None,
+) -> tuple[list[models.Image], ImageManifestManager, JobFileMapping | None]:
```
SpecLad, line 1408: It looks like this function is only called once, and the middle value is ignored, so why return it?

cvat/apps/engine/task.py
```diff
@@ -1781,7 +1876,13 @@ def update_status(msg: str) -> None:
             db_data.stop_frame, db_data.start_frame + (db_data.size - 1) * db_data.get_frame_step()
         )
 
-    slogger.glob.info("Found frames {} for Data #{}".format(db_data.size, db_data.id))
+    slogger.glob.info(
+        "Saved media for Data #{}: media type {}, {} frames".format(
```
SpecLad:
```suggestion
        "Saved media for Data #{}: media type '{}', {} frames".format(
```

cvat/apps/engine/task.py
zhiltsov-max: I think we need to slowly factor out some common context to be passed into these building pipeline functions, `db_task` + `data` dict (i.e. the creation parameters) seem to be like a reasonable choice.

cvat/apps/engine/task.py
zhiltsov-max: It's returned because it's modified in the function and can be useful after the function call, even though it's not used now. I'm not totally sure these creation functions have to be final sinks for the manifests.

cvat/apps/engine/task.py
zhiltsov-max: I think I'd prefer having the same structure of the functions in the different branches, unless they found deficient. This should help with further decomposition and simplification of the code. I also think `_collect_video_dataset_descriptors` currently is focused on what it's doing and it's also clear what it's not supposed to do.

SpecLad (approved):

PR #10543, opened 2026-04-28
Split annotation types into more specific ones

zhiltsov-max:
### Motivation and context

Split annotation types into more detailed subclasses for easier combination.

### Checklist

### License


cvat/apps/engine/models.py
```diff
@@ -1416,7 +1415,14 @@ class Meta:
         abstract = True
         default_permissions = ()
 
-class Shape(models.Model):
+class FrameAnnotationMixin(models.Model):
+    frame = models.PositiveIntegerField()
+
+    class Meta:
+        abstract = True
+        default_permissions = ()
```
SpecLad: Is this actually needed? I thought it was just a holdover from Django's permission system (which we don't use).

cvat/apps/engine/models.py
```diff
@@ -1428,33 +1434,40 @@ class Meta:
...
-class LabeledTrack(Annotation):
+class LabeledTrack(Annotation, FrameAnnotationMixin):
     parent = models.ForeignKey('self', on_delete=models.DO_NOTHING, null=True, related_name='elements')
 
 class LabeledTrackAttributeVal(AttributeVal):
     track = models.ForeignKey(LabeledTrack, on_delete=models.DO_NOTHING,
         related_name='attributes', related_query_name='attribute')
 
-class TrackedShape(Shape):
+class TrackedShape(FrameAnnotationMixin, ShapeAnnotationMixin):
```
SpecLad, line 1463: It's a little weird to have this include annotation mixins when it is not itself an annotation... Maybe "Annotation" should be removed from the base class names or replaced with something else?

cvat/apps/engine/serializers.py
```diff
@@ -3296,9 +3296,17 @@ def to_internal_value(self, data):
         data['value'] = str(data['value'])
         return super().to_internal_value(data)
 
+class AttributedAnnotationSerializer(serializers.Serializer):
+    attributes = AttributeValSerializer(many=True, default=[])
+
+class ScoredAnnotationSerializer(serializers.Serializer):
+    score = serializers.FloatField(min_value=0, max_value=1, default=1)
+
+class ImageAnnotationSerializer(serializers.Serializer):
```
SpecLad:
```suggestion
class FrameAnnotationSerializer(serializers.Serializer):
```

This would be consistent with the model mixin.

cvat/apps/engine/serializers.py
```diff
@@ -3296,9 +3296,17 @@ def to_internal_value(self, data):
         data['value'] = str(data['value'])
         return super().to_internal_value(data)
 
+class AttributedAnnotationSerializer(serializers.Serializer):
```
SpecLad, line 3299: This is just a nitpick, but IMO these mini-serializers should be moved after `AnnotationSerializer` for consistency with `models.py`.

cvat/apps/engine/models.py
zhiltsov-max: These mixins are supposed for annotation classes, and it's reflected in the class names. This class is an annotation, even though it's not inherited from `Annotation`. I think there could be another annotation base class, but it would only contain `id` or be empty, so it doesn't look like a good addition so far.

cvat/apps/engine/models.py
zhiltsov-max: I removed it. but I think it should be removed everywhere, if we're not using it.

cvat/apps/engine/serializers.py
zhiltsov-max: Renamed

cvat/apps/engine/serializers.py
zhiltsov-max: I think it's a little bit overkill. The files are modified independently, I don't think it makes sense to maintain the order of type definitions without strong reasons for this.

SpecLad (approved):

PR #10544, opened 2026-04-28
Add cache read events and update Grafana cache throughput panel

azhavoro:
### Checklist

### License


cvat/apps/engine/cache.py
```diff
@@ -301,17 +308,25 @@ def _bulk_delete_cache_items(self, keys: Sequence[str]):
         slogger.glob.info(f"Removed the cache keys {format_list(keys)}")
 
     def _get_cache_item(self, key: str) -> _CacheItem | None:
+        rq_queue = getattr(rq.get_current_job(), "origin", None)
```
zhiltsov-max: Please extract a constant for `"origin"`, it's used at least in 2 places and probably has a special meaning too.

cvat/apps/engine/cache.py
```diff
@@ -235,9 +234,17 @@ def _create_and_set_cache_item(
             key,
         ):
             cached_item = cache.get(key)
-            if cached_item is not None and timestamp <= cached_item[3]:
-                item = cached_item
-            else:
+            if cached_item is not None:
+                cache_item_read_signal.send(
```
zhiltsov-max, line 261: Do we need to a signal for any get call?

azhavoro: The idea is to track of every read operation from the cache, regardless of whether the data is valid or not. As far as I can see, the only scenario not currently covered is corrupted data, but this case is difficult to implement with the current approach, and I think it’s very rare to support that kind of statistics data

zhiltsov-max (approved):

PR #10545, opened 2026-04-28
Track client IP addresses in event logs

bsekachev:
### Motivation and context

This PR adds client IP address tracking to CVAT event logs.

The IP address is resolved with DRF proxy-aware request identification, so deployments behind Traefik and nginx can keep the correct client address by configuring `CVAT_NUM_PROXIES`. The default is `2`, matching the standard `traefik -> nginx -> uvicorn -> Django` deployment path.

Client IP addresses are stored in event log payloads for internal auditing, but are removed from exported event CSV files.

### Summary of changes

- Added `remote_addr` to server event request metadata.
- Added `remote_addr` to client event payload request metadata.
- Added `CVAT_NUM_PROXIES` support for DRF `NUM_PROXIES`.
- Set the default proxy count to `2` in settings, Docker Compose, and Helm values.
- Excluded `remote_addr` from exported event CSV payloads.
- Added a changelog fragment.

### How has this been tested?

- Ran `../../.env/bin/python -m py_compile cvat/apps/events/views.py cvat/apps/events/export.py cvat/settings/base.py`.
- Ran `npx remark changelog.d/20260512_134508_sekachev.bs_refactored_events.md`.
- Ran `./dev/check_changelog_fragments.py`.

### Checklist

### License


bsekachev:
@codex review

bsekachev:
@SpecLad review pls

cvat/apps/events/views.py
```diff
@@ -101,11 +102,18 @@ def create(self, request):
         serializer.is_valid(raise_exception=True)
 
         handle_client_events_push(request, serializer.validated_data)
+        remote_addr = get_remote_addr(request)
         for event in serializer.validated_data["events"]:
             if event["scope"] == USER_ACTIVITY_SCOPE:
                 # do not record these events, we only need them for correct working time computation
                 continue
 
+            if remote_addr:
```
bsekachev: if remote_addr := get_remote_addr(request)

bsekachev:
@SpecLad updated implementation

cvat/apps/events/views.py
```diff
@@ -106,6 +107,12 @@ def create(self, request):
                 # do not record these events, we only need them for correct working time computation
                 continue
 
+            if remote_addr := get_remote_addr(request):
+                event = {
+                    **event,
+                    "payload": add_remote_addr_to_payload(event.get("payload"), remote_addr),
```
SpecLad: We should create a new column for this in Clickhouse. Now that we have limited Clickhouse migration support, we don't need to jam everything into `payload` anymore.

bsekachev: added, but one thing is that for docker deployment, vector pod must be explicitly restarted
after adding "remote_addr" to its config the container not restarted because config is mounted
I do not think we have the same issue with kube because deploy always happens from scratch

bsekachev:
@SpecLad

It looks like using BaseThrottle.get_ident() for remote_addr also changes anon throttling behavior.

Previously, with NUM_PROXIES = None, DRF used the whole X-Forwarded-For chain as the throttle identity. With CVAT_NUM_PROXIES set, DRF picks one address from the chain instead. In Helm REST tests this makes many login requests use the same anon throttle key, so they can hit the rate limit.

cvat/settings/base.py
```diff
@@ -162,6 +178,7 @@ def generate_secret_key():
     "DEFAULT_THROTTLE_RATES": {
         "anon": "100/minute",
     },
+    "NUM_PROXIES": parse_num_proxies(os.getenv("CVAT_NUM_PROXIES", "2")),
```
SpecLad: The default here should probably be 0, since in this code we shouldn't make assumptions about how CVAT is deployed.

helm-chart/test.values.yaml
```diff
@@ -1,5 +1,6 @@
 cvat:
   backend:
+    numProxies: 2
```
SpecLad: This seems redundant. `test.values.yaml` is applied on top of `values.yaml`, which already contains this setting.

docker-compose.yml
```diff
@@ -14,6 +14,7 @@ x-backend-env: &backend-env
   CVAT_LOG_IMPORT_ERRORS: 'true'
   CVAT_ALLOW_STATIC_CACHE: '${CVAT_ALLOW_STATIC_CACHE:-no}'
   CVAT_DEFAULT_BACKING_CS_ID:
+  CVAT_NUM_PROXIES: '${CVAT_NUM_PROXIES:-2}'
```
SpecLad: I'm a little concerned that we're effectively hardcoding implementation details (the internal nginx server) into the configuration. If the internal architecture changes, this could invalidate any custom values that the users have configured.

Perhaps we should default this to 1 instead, and then increment the value in the entrypoint script before passing it to the server?

helm-chart/templates/_helpers.tpl
```diff
@@ -119,6 +119,11 @@ The name of the service account to use for backend pods
       name: "{{ tpl (.Values.postgresql.secret.name) . }}"
       key: password
 
+{{- if and (ne .Values.cvat.backend.numProxies nil) (ne (toString .Values.cvat.backend.numProxies) "") }}
```
SpecLad: IMO, this is an unnecessary condition. `numProxies` already has a default, there is no reason for someone to set it to nil or empty string.

SpecLad:
> Previously, with NUM_PROXIES = None, DRF used the whole X-Forwarded-For chain as the throttle identity. With CVAT_NUM_PROXIES set, DRF picks one address from the chain instead. In Helm REST tests this makes many login requests use the same anon throttle key, so they can hit the rate limit.

Interesting, but why would changing `NUM_PROXIES` affect this? Wouldn't the full chain be the same every time too?

cvat/settings/base.py
bsekachev: Agree with default value in the code

docker-compose.yml
bsekachev: Entrypoint script? 
What do you think if we just increase the value on django settings level when defined?

bsekachev:
> Wouldn't the full chain be the same every time too?

I thought the same. If we have one traefik, one nginx, it should work so. But maybe only if we add logging we know for sure.

docker-compose.yml
SpecLad: > Entrypoint script?

`backend_entrypoint.sh`

> What do you think if we just increase the value on django settings level when defined?

I don't think that's appropriate, because Django itself doesn't know whether there is a proxy in front of it.

backend_entrypoint.sh
```diff
@@ -19,6 +19,18 @@ wait_for_clickhouse() {
     wait-for-it "${CLICKHOUSE_HOST}:${CLICKHOUSE_PORT:-8123}" -t 0
 }
 
+account_for_internal_proxy() {
+    if [ -n "${CVAT_NUM_PROXIES:-}" ]; then
```
SpecLad: You should instead default it to 0 (e.g. `: ${CVAT_NUM_PROXIES:=0}`). Otherwise, rate limiting will not work correctly if the backend container is launched independently (e.g. for debugging).

cvat/settings/testing_rest.py
```diff
@@ -12,6 +12,8 @@
     "django.contrib.auth.hashers.MD5PasswordHasher",
 ]
 
+REST_FRAMEWORK["DEFAULT_THROTTLE_RATES"]["anon"] = "300/minute"
```
bsekachev, line 15:
```suggestion
REST_FRAMEWORK["DEFAULT_THROTTLE_RATES"]["anon"] = "100/minute"
```

bsekachev:
@SpecLad 

So, for the most of requests

x_forwarded_for='10.244.0.1, 10.244.0.3'

But for `/server/api/about` 
x_forwarded_for='10.244.0.1'

I have no idea why (Andrey told it is a probe).

SpecLad:
> Andrey told it is a probe

Yeah, that makes sense. It shouldn't be a problem, since DRF will take the leftmost address if there are fewer than `NUM_PROXIES` addresses in the list.

SpecLad (approved):

PR #10546, opened 2026-04-28
e2e test polygon simplify

archibald1418:
### Motivation and context

Coverage for #10449
Depends on #10568 

Adds Cypress coverage for polygon simplification behavior. The test draws a jagged polygon using the UI Shift-hover workflow, makes aligned copies, applies Simplify with low and high accuracy settings, and verifies the resulting geometry.
The assertions check that aggressive simplification reduces polygon detail and lowers IoU, while high-accuracy simplification preserves the shape much better. The test also verifies that the original polygon remains unchanged.

#### Test: 'Simplify' removes points, decreases areas. Higher accuracy restores shape
Creates 60-point polygon + copy. Runs 3 simplifications (accuracy 0/2/13). Verifies lower accuracy → fewer points/smaller area. Higher accuracy → restores original. Original unchanged.

#### Context (skipped): Auto-simplify
Tests auto-simplify toggle during polyline/polygon drawing. Skipped - Redux state bug causes incorrect toggle between tests.

### Context: 'Simplify' can be invoked as annotations action

Setup: 1 original + 3 copies positioned around canvas.

#### Test: Call 'simplify' on each object
Simplifies 3 copies with distances 20/40/64. Verifies area progression: dist20 < dist40 < dist64. Original unchanged.

### Test: Call 'simplify' on all shapes as a bulk action
Bulk simplifies all 3 copies (distance=64). Verifies all result areas equal. Original unchanged.

### Checklist

### License


archibald1418:
@codex review

tests/cypress/e2e/actions_objects/simplify_polygons_feature.js
```diff
@@ -0,0 +1,245 @@
+// Copyright (C) CVAT.ai Corporation
+//
+// SPDX-License-Identifier: LicenseRef-CVAT-AI-Commercial
```
klakhov:
```suggestion
// Copyright (C) CVAT.ai Corporation
//
// SPDX-License-Identifier: MIT
```

tests/cypress/e2e/actions_objects/simplify_polygons_feature.js
```diff
@@ -0,0 +1,245 @@
...
+        setSimplifyAccuracy(accuracy);
+        cy.get('.cvat-approx-poly-threshold-wrapper .ant-btn-primary').click({ force: true });
+        cy.get('.cvat-approx-poly-threshold-wrapper').should('not.exist');
+    }
+
+    before(() => {
+        cy.prepareUserSession();
+        cy.openTaskJob(taskName);
+        cy.createPolygon(createDetailedPolygon, null, 'shiftHover');
+        // TODO: add different color labels to aid visual confirmation
```
klakhov: What does it mean? we have only one polygon

klakhov (commented):
1. I would vote to remove IoU computations. The feature(and the underlying alorithm) only promises reduced amout of points, in corner cases shape can be significantly changed or unchanged. If you want a very lightweight geometry sanity check, keep shoelaceArea and use it only for relative assertions like: aggressive simplify changes area more than high-accuracy simplify

2. The test title says IoU “grows” as more detail is preserved, but the body only checks two endpoints, we need at least one point inbetween

3. Need to check that the feature works from shortcut

4. Need to check that simplify bulk action works

5. Need to check auto-simplify start after draw:
Image

6.  Need at least one test for polyline

tests/cypress/e2e/actions_objects/simplify_polygons_feature.js
archibald1418: it's for visual comparison with the reference polygon, to see how area changes

archibald1418: removed this, not needed

tests/cypress/e2e/actions_objects/simplify_polygons_feature.js
klakhov: Still not applied

tests/package.json
```diff
@@ -2,7 +2,8 @@
   "scripts": {
     "cypress:run:chrome": "cypress run --browser chrome",
     "cypress:run:firefox": "cypress run --browser firefox --config-file nightly_cypress.config.js",
-    "cypress:run:chrome:canvas3d": "cypress run --headed --browser chrome --config-file cypress_canvas3d.config.js"
+    "cypress:run:chrome:canvas3d": "cypress run --headed --browser chrome --config-file cypress_canvas3d.config.js",
+    "cypress:watch": "nodemon -e js --exec cypress run --headless --browser chrome --spec"
```
klakhov: I’m not sure about this change. Do we really need live Cypress runs on every change?

From my perspective, we should run tests explicitly once we’ve finished making changes. This is not like a Webpack or Vite dev server, where live reload is expected. An occasional save could just result in a failed test run.

Anyway, this seems a bit outside the scope of polygon simplify tests. If you want to proceed with this change, let’s create a separate PR and discuss it there.

klakhov (commented):
Overall test looks fine to me

tests/package.json
archibald1418: fixed license, removed package.json changes

archibald1418: removed nodemon stuff

klakhov (approved):

PR #10547, opened 2026-04-29
Fix manifest frame order mismatch for non-predefined sorting (#10104)

### LINKED ISSUE #10104: Task creation fails if a manifest is present but frame order does not match it

drb101005:
-Summary

Fixes incorrect index-based mapping between dataset frames and manifest entries when sorting is not set to PREDEFINED.

-Changes

* Introduced filename-based lookup for manifest entries in non-PREDEFINED mode
* Added path normalization to ensure consistent matching across platforms
* Added validation for mismatched dataset and manifest contents (missing/extra files, duplicates)
* Preserved existing PREDEFINED behavior and ordering semantics

- Why this approach

Unlike PR #10134, this solution:

* does not reorder manifest data
* does not modify global behavior
* only fixes the incorrect mapping assumption locally

- Testing

* Verified mismatched manifest order works correctly
* Verified missing/extra files raise appropriate validation errors
* Existing PREDEFINED and video workflows remain unchanged

Closes #10104

### DIFF
--- a/ai-models/agents_deployment/sam2/.env
+++ b/ai-models/agents_deployment/sam2/.env
@@ -6,6 +6,8 @@ FUNCTION_NAME=your_function_name
 # Required to authenticate to CVAT API. You can get it in CVAT UI: https://app.cvat.ai/profile#security
 CVAT_ACCESS_TOKEN=your_cvat_token
 # Please specify your image url with tag. For example: myrepo/sam2:latest
+# We provide CPU and GPU images in our repository (https://hub.docker.com/repository/docker/cvat/sam2_agent/tags)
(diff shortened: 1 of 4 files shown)


rob3c:
The 3 files in this PR have nothing to do with what you've described. They suspiciously involve adding remote python debugger support among other unrelated things.

PR #10554, opened 2026-04-30
As/fix create data 500 server files

### LINKED ISSUE #9647: Error 500 when calling ApiClient(...).tasks_api.create_data(...)

### Steps to Reproduce

The following code yields an Error 500 without further explanation

```python
task_id = 1508668  # int | A unique integer value identifying this task.
upload_finish = True  # bool | Finishes data upload. Can be combined with Upload-Start header to create task data with one request (optional)
upload_multiple = True  # bool | Indicates that data with this request are single or multiple files that should be attached to a task (optional)
upload_start = True  # bool | Initializes data upload. Optionally, can include upload metadata in the request body. (optional)
data_request = models.DataRequest(
    image_quality=70,
    start_frame=0,
    server_files=["chemical_spill.v1i.yolov8/train/images/IMG_1141_jpeg.rf.5884a5e0dc8f46c451564dc2b6f3c20e.jpg"],
    cloud_storage_id=cloud_storage_id,
    storage_method=models.StorageMethod("cache"),
    storage=models.StorageType("cloud_storage"),
    sorting_method=models.SortingMethod("lexicographical"),
    job_file_mapping=[
        ["chemical_spill.v1i.yolov8/train/images/IMG_1141_jpeg.rf.5884a5e0dc8f46c451564dc2b6f3c20e.jpg"]
    ],
) # DataRequest |  (optional)

try:
    data, _ = api_client.tasks_api.create_data(
        task_id,
        upload_finish=upload_finish,
        upload_multiple=upload_multiple,
        upload_start=upload_start,
        data_request=data_request,
    )
    pprint(data)
    request_id = data['rq_id']
    print(request_id)
except exceptions.ApiException as e:
    print("Exception when calling TasksApi.create_data(): %s\n" % e)
```

when changing the path in `server_files` it works once and fails on repeted calls with the same path

requests_api.retrieve(request_id) yields: 
```
...
 'message': 'ValueError: Unsupported URL scheme: . Only http and https are '
            'supported\n',
...
```
Regardless of the unexpected error 500 this is not expected, becuase the SDK reference clearly states thet URLs are exprected for **remote_files** but not for **server_files** (the expected path format for server files would be a valuable addition to the documentation imho):

Image

Screenshot of failed call:
Image

### Expected Behavior

Successful execution or more specific error message

### Possible Solution

_No response_

### Context

_No response_

### Environment

```Markdown
- CVAT online
- latest python sdk
- databricks notebook as runtime for the api client
```

Rwwwrl:
### Title
[DEV-857] Fix #9647: validate and surface remote_files download errors

### Motivation and context

`POST /api/tasks/{id}/data` accepted any URL string in `remote_files` and any failure during the per-URL download in the import worker reached the client as a 500 with a raw traceback in `request.message`. Two fixes:

- `DataSerializer.validate_remote_files` rejects non-`http(s)` URLs synchronously, so unsupported schemes (`ftp://`, `file://`, …) are 400 at request time instead of failing the background job.
- `_download_data` now raises a typed `_FailedToDownloadFileError` for every failure mode (HTTP non-200, `requests.exceptions.RequestException`, filename collision). `create_thread` catches it at the call site and re-raises as DRF `ValidationError`, so the requests API surfaces a single readable line via `formatted_exception`. A connect/read timeout was added to `session.get` so a hung URL no longer pins an `import` worker for the queue's 4h job timeout.

Tests cover the four media-source surfaces of `create_thread`:
- `remote_files` failure path (case 1).
- Cloud storage with `use_cache=False` and a missing key (case A) — surfaces DRF `NotFound`.
- Cloud storage with `use_cache=True` and a manifest that references a key missing in the bucket (case C) — task creation is expected to succeed; the manifest is deliberately trusted at create time. This locks in the team-lead-approved trade-off.
- Cloud storage with `use_cache=True` and no manifest (case D) — `HeaderFirstMediaDownloader`'s per-file partial GET surfaces DRF `NotFound`.

Case C does not validate manifest keys against the bucket up front. A stale manifest still surfaces only when the affected frame is requested (chunk endpoint returns 404 via the existing `@validate_file_status` decorator). Adding eager validation was discussed and intentionally deferred.

Also bundled (separate commit, test infra): a `--no-services` pytest flag for running the suite against a stack the developer manages outside the harness — used by the host-based dev loop in `docker-compose.sosov.pytest.yaml`.

Fixes #9647.

### Checklist

### License

### DIFF
--- a/cvat/apps/engine/serializers.py
+++ b/cvat/apps/engine/serializers.py
@@ -21,6 +21,7 @@ from inspect import isclass
 from pathlib import Path
 from tempfile import NamedTemporaryFile
 from typing import Any, cast
+from urllib.parse import urlparse
 
 import django_rq
 from django.conf import settings
@@ -2361,6 +2362,19 @@ class DataSerializer(serializers.ModelSerializer):
 
                 existing_files.add(filename)
 
+        return value
+
+    def validate_remote_files(self, value: list[dict[str, str]]) -> list[dict[str, str]]:
+        errors: list[str] = []
+        for entry in value:
+            url = entry["file"]
+            parsed = urlparse(url)
+            if parsed.scheme not in ("http", "https") or not parsed.netloc:
+                errors.append(
+                    f"{url!r}: remote_files entries must be http(s) URLs."
+                )
+        if errors:
+            raise serializers.ValidationError(errors)
         return value
 
     # pylint: disable=no-self-use
--- a/cvat/settings/base.py
+++ b/cvat/settings/base.py
@@ -233,7 +233,7 @@ IAM_DEFAULT_ROLE = "user"
 IAM_ADMIN_ROLE = "admin"
 # Index in the list below corresponds to the priority (0 has highest priority)
 IAM_ROLES = [IAM_ADMIN_ROLE, "user", "worker"]
-IAM_OPA_HOST = "http://opa:8181"
+IAM_OPA_HOST = os.getenv("IAM_OPA_HOST", "http://opa:8181")
 IAM_OPA_DATA_URL = f"{IAM_OPA_HOST}/v1/data"
 LOGIN_URL = "rest_login"
 LOGIN_REDIRECT_URL = "/"
@@ -756,7 +756,7 @@ ASSET_SUPPORTED_TYPES = ("image/jpeg", "image/png", "image/webp", "image/gif", "
 ASSET_MAX_IMAGE_SIZE = 1920
 ASSET_MAX_COUNT_PER_GUIDE = 150
 
-SMOKESCREEN_ENABLED = True
+SMOKESCREEN_ENABLED = to_bool(os.getenv("SMOKESCREEN_ENABLED", True))
 
 # By default, email backend is django.core.mail.backends.smtp.EmailBackend
 # But it won't work without additional configuration, so we set it to None
--- a/cvat/settings/base.py
+++ b/cvat/settings/base.py
@@ -233,7 +233,7 @@ IAM_DEFAULT_ROLE = "user"
 IAM_ADMIN_ROLE = "admin"
 # Index in the list below corresponds to the priority (0 has highest priority)
 IAM_ROLES = [IAM_ADMIN_ROLE, "user", "worker"]
-IAM_OPA_HOST = os.getenv("IAM_OPA_HOST", "http://opa:8181")
+IAM_OPA_HOST = "http://opa:8181"
 IAM_OPA_DATA_URL = f"{IAM_OPA_HOST}/v1/data"
 LOGIN_URL = "rest_login"
 LOGIN_REDIRECT_URL = "/"
--- a/tests/python/rest_api/test_remote_url.py
+++ b/tests/python/rest_api/test_remote_url.py
@@ -70,7 +70,7 @@ class TestCreateFromRemote:
 
         self._test_can_create(user, self.task_id, remote_resources)
 
-    def test_remote_files_when_downloading_them_fails(self, find_users):
+    def test_cannot_create_task_when_remote_file_download_fails(self, find_users):
         user = find_users(privilege="admin")[0]["username"]
         url = "http://invalid.invalid/x.png"
 
--- a/tests/python/rest_api/test_remote_url.py
+++ b/tests/python/rest_api/test_remote_url.py
@@ -58,24 +58,18 @@ class TestCreateFromRemote:
         assert response_json["status"] == "failed"
         return response_json
 
-    def test_cannot_create(self, find_users):
+    def test_can_report_failed_remote_file_download(self, find_users):
         user = find_users(privilege="admin")[0]["username"]
         remote_resources = ["http://localhost/favicon.ico"]
 
-        self._test_cannot_create(user, self.task_id, remote_resources)
+        response_json = self._test_cannot_create(user, self.task_id, remote_resources)
+
+        message = response_json["message"]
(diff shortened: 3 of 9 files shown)


cvat/settings/base.py
```diff
@@ -233,7 +233,7 @@ def generate_secret_key():
 IAM_ADMIN_ROLE = "admin"
 # Index in the list below corresponds to the priority (0 has highest priority)
 IAM_ROLES = [IAM_ADMIN_ROLE, "user", "worker"]
-IAM_OPA_HOST = "http://opa:8181"
+IAM_OPA_HOST = os.getenv("IAM_OPA_HOST", "http://opa:8181")
```
zhiltsov-max: Please remove this change in favor of https://github.com/cvat-ai/cvat/pull/10496

tests/python/rest_api/test_task_data.py
```diff
@@ -544,6 +544,97 @@ def test_create_task_with_cloud_storage_files(
         kwargs = {"org_id": org_id} if org_id else {}
         create_task(self._USERNAME, task_spec, data_spec, **kwargs)
 
+    @pytest.mark.with_external_services
+    @pytest.mark.timeout(60)
+    def test_cannot_create_task_with_cloud_storage_without_cache_when_server_file_is_missing(
```
zhiltsov-max, line 554: I think this test called a little bit incorrectly, because it fails with and without cache for different reasons. If you want to check that it only happens without cache, you need to add check that `CVAT_ALLOW_STATIC_CACHE` is specified in the environment. I'd suggest to parametrize on this parameter or remove it from the test.

If you decide to remove the parameter, the test can also be removed, because it duplicates `test_cannot_create_task_with_cloud_storage_without_manifest_when_server_file_is_missing`.

tests/python/rest_api/test_task_data.py
```diff
@@ -544,6 +544,97 @@ def test_create_task_with_cloud_storage_files(
...
+        present_image = generate_image_file(filename=present_name)
+        present_image.seek(0)
+        s3_client.create_file(filename=f"{prefix}/{present_name}", data=present_image.getvalue())
+        request.addfinalizer(partial(s3_client.remove_file, filename=f"{prefix}/{present_name}"))
+
+        manifest_body = (
+            '{"version":"1.0"}\n'
+            '{"type":"images"}\n'
+            '{"name":"01_present","extension":".png","width":100,"height":50,'
+            '"checksum":"00000000000000000000000000000000"}\n'
```
zhiltsov-max: I think you can remove checksums here

tests/python/rest_api/test_remote_url.py
```diff
@@ -68,3 +69,13 @@ def test_can_create(self, find_users):
         remote_resources = ["https://docs.cvat.ai/favicons/favicon-32x32.png"]
 
         self._test_can_create(user, self.task_id, remote_resources)
+
+    def test_remote_files_when_downloading_them_fails(self, find_users):
```
zhiltsov-max: Consider renaming to something like `test_can_report_failed_remote_file_download`

cvat/apps/engine/serializers.py
zhiltsov-max, line 1: Please add a changelog note, here is the command for this https://github.com/cvat-ai/cvat/blob/ef4bb4eb647ace264291c859871fbb090a092b07/CHANGELOG.md?plain=1#L12

tests/python/rest_api/test_task_data.py
Rwwwrl: Aforementioned use cases serve different purposes, but you are right that CVAT_ALLOW_STATIC_CACHE (otherwise storage is flipped to CACHE regardless "use_cache": False) should be set up to True value, I missed that.

test_cannot_create_task_with_cloud_storage_without_cache_when_server_file_is_missing is focusing on https://github.com/cvat-ai/cvat/blob/ef4bb4eb647ace264291c859871fbb090a092b07/cvat/apps/engine/task.py#L1230 use case

have updated the test with @pytest.mark.skipif, so it runs only in case CVAT_ALLOW_STATIC_CACHE is True (that we actually have equal to True in https://github.com/cvat-ai/cvat/blob/ef4bb4eb647ace264291c859871fbb090a092b07/.github/workflows/main.yml#L149)

tests/python/rest_api/test_task_data.py
Rwwwrl: removed

tests/python/rest_api/test_remote_url.py
Rwwwrl: renamed

cvat/apps/engine/serializers.py
Rwwwrl: added

tests/python/rest_api/test_remote_url.py
```diff
@@ -68,3 +69,13 @@ def test_can_create(self, find_users):
         remote_resources = ["https://docs.cvat.ai/favicons/favicon-32x32.png"]
 
         self._test_can_create(user, self.task_id, remote_resources)
+
+    def test_cannot_create_task_when_remote_file_download_fails(self, find_users):
```
zhiltsov-max: I think it will be better to just update the `test_cannot_create` test above instead. It already does the same, so we just need to add the error checks.

Rwwwrl: renamed the original test

zhiltsov-max (approved):

Issue #10503, opened 2026-04-20
`saveAnnotationsAsync` can create duplicate shapes when the save pipeline throws after the server PATCH succeeds


dtafler:
### Actions before raising this issue

- [x] I searched the existing issues and did not find anything similar.
- [x] I read/searched [the docs](https://docs.cvat.ai/docs/)

## Summary

If `jobInstance.annotations.save()` rejects after the server has already persisted shapes, the client ends up in a state where the server holds the new shapes with new IDs but the client still has those same shapes marked as unsent. The next save re-ships them, producing coordinate identical duplicates on the server. We have observed this repeatedly in production.

## The bug

```typescript
// cvat-ui/src/actions/annotation-actions.ts:1011
export function saveAnnotationsAsync(): ThunkAction {
    return async (dispatch: ThunkDispatch): Promise<void> {
        const { jobInstance } = receiveAnnotationsParameters();
        dispatch({ type: AnnotationActionTypes.SAVE_ANNOTATIONS, payload: {} });

        try {
            const saveJobEvent = await jobInstance.logger.log(EventScope.saveJob, {}, true);

            await jobInstance.frames.save();
            await jobInstance.annotations.save();    // if this rejects AFTER the server PATCH succeeds, we jump to catch
            await saveJobEvent.close();              // skipped, save:job event is never sent to the server
            dispatch(saveLogsAsync());               // skipped
            // ...
            dispatch(fetchAnnotationsAsync());       // skipped, NO state reconciliation
        } catch (error) {
            dispatch({ type: AnnotationActionTypes.SAVE_ANNOTATIONS_FAILED, payload: { error } });
            throw error;
        }
    };
}
```

`fetchAnnotationsAsync()` is the only call that pulls the server assigned IDs back into local state, so skipping it on the error path leaves the client thinking its shapes are still unsaved. The next invocation of `saveAnnotationsAsync` creates them again.

## Evidence from a production job

Extracted from `/api/events/export` for one real job. An annotator drew 282 new shapes (171 polygons and 111 keypoints), double clicked Save, and the autosave timer 100 seconds later re-submitted the whole batch as duplicates.

```text
timestamp             scope          duration  notes
10:29:49.763 UTC      click:element  -         Save button
10:29:49.819 UTC      click:element  -         Save button (+56ms, double click)
10:29:53.714 UTC      create:shapes  -         request.id=fed955ce…, 171 polygons
                                               server shape IDs 4692125..4692295
[...]

PR #10555, opened 2026-04-30
Fix 10503

prathamesh-patil-5090:
## Fix duplicate shapes in saveAnnotationsAsync pipeline

**Issue:** When `jobInstance.annotations.save()` rejects after the server persists shapes, the client ends up out of sync with the server. The shapes are marked as unsent on the client while the server has already persisted them with new IDs. The next save re-ships the shapes, creating coordinate-identical duplicates.

**Root Cause:** The `fetchAnnotationsAsync()` reconciliation call was skipped when the save pipeline failed, preventing the client from retrieving the server-assigned shape IDs.

**Solution:** Wrapped the save operations (`frames.save()` and `annotations.save()`) in an inner try-finally block to ensure reconciliation steps (`saveJobEvent.close()`, `saveLogsAsync()`, and `fetchAnnotationsAsync()`) always execute, even when save fails. This ensures the client state is always synchronized with the server, preventing duplicate submissions.

**Related Issue:** #10503


bsekachev:
AI generated pull request does not seem to do what it describes. Even other files are responsible for annotations saving.

PR #10569, opened 2026-05-06
Move some fields from search to filters

zhiltsov-max:
### Summary

- Moved most enum-like fields from `search_fields` to `simple_filters`, and `filter_fields` in viewsets
- Added field sorting for schema

### Motivation and context

Enum-like fields (e.g. `status`, `state`, `role`) were declared in `search_fields` on several list endpoints. DRF's `SearchFilter` performs case-insensitive substring matching across `search_fields`, so the global `?search=` parameter would also match against these enum values, which is rarely what users want. They are now exposed only via `simple_filters` (exact match), e.g. `?status=completed`, `?state=in_progress`, `?role=maintainer`.

In some cases, free-text fields are intentionally excluded from simple filters, because exact matching is not very useful for them. By this logic, `project_name` and `task_name` in task and job list endpoints should be removed as well,  but they are kept to avoid breaking changes without a significant reason, as they're available for quite a long time already.

The filters are now sorted for schema generation, which helps to avoid spurious schema diffs. 

#### Per-endpoint changes

| Endpoint | Removed from `search_fields` | Added to `simple_filters` | Added to `filter_fields` |
|---|---|---|---|
| `GET /api/projects` | `status` | — | — |
| `GET /api/tasks` | `status`, `mode`, `dimension`, `validation_mode` | — | — |
| `GET /api/jobs` | `state`, `stage` | — | — |
| `GET /api/memberships` | `role` | — | — |
| `GET /api/webhooks` | `type` | — | — |
| `GET /api/cloudstorages` | `provider_type`, `credentials_type` | — | — |
| `GET /api/access_tokens` | — | `read_only` | — |
| `GET /api/invitations` | — | `user_id`, `accepted` | `id` |
| `GET /api/requests` | — | `org_id` | `org`, `org_id` |

For fields removed from `search_fields`: they remain exact-match filterable as `simple_filters` (e.g. `?status=completed`).
For fields newly in `simple_filters`: they were previously available only via the `filter` parameter and are now also exact-match filterable.
For fields newly in `filter_fields`: they were previously not available for filtering.

The following endpoints were touched only for the construction-style refactor (no field-level behavior change):
`GET /api/issues`, `GET /api/comments`, `GET /api/labels`, `GET /api/users`, `GET /api/organizations`, `GET /api/consensus_settings`, `GET /api/quality/conflicts`, `GET /api/quality/reports`, `GET /api/quality/settings`.

### API response changes

The following list endpoints now return additional fields:

- `GET /api/requests` — `operation.org_id` (integer, nullable). Supports `?org_id=` filtering.
- `GET /api/invitations` — `accepted` (boolean). Mirrors the existing `accepted` filters.

### How has this been tested?

Added simple-filter test coverage for fields newly exposed in this PR:

- `tests/python/rest_api/test_access_tokens.py`: `read_only`.
- `tests/python/rest_api/test_invitations.py`: `user_id`, `accepted`.
- `tests/python/rest_api/test_requests.py`: `org`, `org_id`.

Filled in pre-existing simple-filter coverage gaps:

- `tests/python/rest_api/test_jobs.py`: `dimension`, `media_type`, `mode`, `task_name`, `project_name`.
- `tests/python/rest_api/test_tasks.py`: `media_type`, `project_name`.

`TestRequestsListFilters` was refactored to class-scoped fixtures so the heavy setup (3 projects + 3 tasks + ~13 RQ requests + the new org-context project/task/exports) runs once for the whole class instead of per parametrize case. Wall-clock for `pytest -k TestRequestsListFilters` on the same hardware:

| | Before | After |
|---|---|---|
| Tests collected | 9 | 10 |
| Total time | 78.4 s | 13.1 s |
| Per-case `call` time | ~5.2 s | ~0.05 s |
| Per-case `setup` time | ~4.3 s every case | ~12.1 s once, then negligible |

≈6× faster despite covering one more case.

The corrupted-meta test was also extracted out of the class to a module-level function with `restore_db_per_function` / `restore_redis_inmem_per_function` decorators since it has no class dependencies.

### Checklist

### License
- [ ] I submit _my code changes_ under the same [MIT License](https://github.com/cvat-ai/cvat/blob/develop/LICENSE) that covers the project.


tests/python/rest_api/test_requests.py
zhiltsov-max, line 1: @nmanovic, some tests were optimized in this PR, maybe it's relevant to your PR with optimizations.

cvat/apps/engine/views.py
```diff
@@ -323,9 +323,9 @@ class ProjectViewSet(viewsets.GenericViewSet, mixins.ListModelMixin,
         'annotation_guide', 'source_storage', 'target_storage',
     )
 
-    search_fields = ('name', 'owner', 'assignee', 'status')
-    filter_fields = list(search_fields) + ['id', 'updated_date']
-    simple_filters = list(search_fields)
+    search_fields = sorted(('name', 'owner', 'assignee', 'status'))
```
SpecLad: Shouldn't `status` be removed from here?

cvat/apps/access_tokens/views.py
```diff
@@ -55,15 +55,11 @@ class AccessTokensViewSet(
...
-        "id",
-        "created_date",
-        "updated_date",
-        "expiry_date",
-        "last_used_date",
-        "read_only",
-    ]
-    simple_filters = list(search_fields)
+    simple_filters = sorted(set(search_fields) | {"read_only"})
+    filter_fields = sorted(
```
SpecLad: `simple_fields` and `filter_fields` are implemented by our own code, right? So why not sort them in the common code instead of in each viewset separately?

cvat/apps/consensus/views.py
```diff
@@ -128,8 +128,8 @@ class ConsensusSettingsViewSet(
     iam_permission_class = ConsensusSettingPermission
 
     search_fields = []
-    filter_fields = ["id", "task_id"]
     simple_filters = ["task_id"]
+    filter_fields = sorted(set(simple_filters) | {"id"})
```
SpecLad: IMO, using `set` here could hide mistakes if the same field is mentioned in multiple lists. It would be more robust to use a list or tuple (e.g. `(*simple_filters, "id")`) and check for duplicates in the common code.

cvat/apps/organizations/views.py
```diff
@@ -259,15 +259,16 @@ class InvitationViewSet(
...
+    simple_filters = sorted(set(search_fields) | {"user_id", "accepted"})
+    filter_fields = sorted(set(simple_filters) | {"id"})
     ordering_fields = list(simple_filters) + ["created_date"]
     ordering = "-created_date"
     lookup_fields = {
         "owner": "owner__username",
         "user_id": "membership__user__id",
         "accepted": "membership__is_active",
     }
+    _related = ("owner", "membership__user", "membership__organization")
```
SpecLad: I think you left this in by accident.

cvat/apps/redis_handler/views.py
```diff
@@ -33,6 +33,9 @@
 slogger = ServerLogManager(__name__)
 
 
+_builtin_list = list  # the list symbol is redefined below
```
SpecLad: I think you can use `__builtins__.list` instead. But also... wouldn't `from __future__ import annotations` resolve whatever problem you had in the first place?

cvat/apps/access_tokens/views.py
zhiltsov-max: Yes, they are both ours. I also thought about moving them onto the schema generation side, but I decided against this because it would not be cached (adding schema caching should be a separate topic anyway, but there's no caching now), so will be computed each time a schema is requested. Current implementation computes all this just once. It misses sorting in the quality apis and requires non-public code to be updated for compliance, so it's a potential argument for sorting in at the schema generation step,

cvat/apps/consensus/views.py
zhiltsov-max: Ignoring this kind of errors was actually intentional. I don't think it makes sense to worry about the duplicates, they should just be deduplicated.

cvat/apps/redis_handler/views.py
zhiltsov-max: Builtins would work, and the import would not. The problem being solved is that there's the class method called `list()`, representing the list endpoint.

cvat/apps/access_tokens/views.py
SpecLad: > so will be computed each time a schema is requested

True, but does that actually significantly impact the schema generation time?

cvat/apps/redis_handler/views.py
SpecLad: I see what you mean. I'd still suggest adding the future import, though, since then the type annotations can stay untouched.

SpecLad: Also, I guess you could just write `filter_fields = simple_filters`.

zhiltsov-max: I'm not sure about sharing the same object between the 2 fields, it's potentially unsafe.

cvat/apps/engine/views.py
zhiltsov-max: Agree, done

cvat/apps/organizations/views.py
zhiltsov-max: Removed, thanks

cvat/apps/redis_handler/views.py
zhiltsov-max: Updated

cvat/apps/access_tokens/views.py
zhiltsov-max: I worried about the `/api/schema` endpoint, which is called by the UI in some cases. It works about 1 second, and sorting seems to have some impact (about 10ms, apparently), but it's about several percents. So I think the sorting can be safely moved to schema generation side, allowing for simpler code in the views. Need to add schema caching later.

SpecLad (approved):

PR #10573, opened 2026-05-08
Append annotations instead of replacing when uploading

Grigorii777:
### Motivation and context

Annotation uploads for tasks and jobs always replaced the existing annotations. This makes it hard to add annotations from an imported file on top of already existing data without doing an explicit export/merge/import flow outside CVAT.

This PR adds an `import_mode` query parameter for task and job annotation uploads:

- `replace` keeps the existing behavior and remains the default.
- `append` imports annotations without deleting the current annotations first.

The option is validated by the background import path, passed through to task/job import callbacks, and documented in the OpenAPI schema. Project imports are left unchanged.

### How has this been tested?

Added REST API coverage for task and job annotation uploads with both import modes:

- default upload replaces existing annotations;
- `import_mode=append` preserves existing annotations and adds imported annotations.
- for jobs and tasks

Local checks run:

```bash
python3 -m py_compile cvat/tests/python/rest_api/test_tasks.py
git -C cvat diff --check -- tests/python/rest_api/test_tasks.py
```

### Checklist

### License


Grigorii777:
@codex review

Grigorii777:
@codex review

Grigorii777:
@codex review

Grigorii777:
I’ve generated the code for the UI – a new radio button during the import of annotations; I’ve tested it locally and everything works.
But the implementation needs to be reviewed, as the front-end developer hasn’t been writing code very often.

Grigorii777:
@codex review

bsekachev:
@codex review

Grigorii777:
@codex review

nmanovic:
@codex review the PR

Grigorii777:
@codex review

Grigorii777:
@codex review

cvat/apps/dataset_manager/task.py
```diff
@@ -833,13 +845,22 @@ def export(
...
+        **options,
+    ):
         job_data = JobData(
             annotation_ir=AnnotationIR(self.db_job.segment.task.dimension),
             db_job=self.db_job,
             create_callback=self.create,
         )
-        self.delete()
+        import_mode = AnnotationImportMode(import_mode)
+        if import_mode == AnnotationImportMode.REPLACE:
```
zhiltsov-max, line 862: Please add a guard for unknown values, we typically use `assert False`

cvat/apps/dataset_manager/task.py
```diff
@@ -1207,26 +1237,48 @@ def export_task(
...
 
 
 @transaction.atomic
-def import_job_annotations(src_file, job_id, format_name, conv_mask_to_poly):
+def import_job_annotations(
+    src_file,
+    job_id,
+    format_name,
+    conv_mask_to_poly,
+    import_mode=AnnotationImportMode.REPLACE,
```
zhiltsov-max:
```suggestion
    *,
    import_mode: AnnotationImportMode = AnnotationImportMode.REPLACE,
```

tests/python/rest_api/test_tasks.py
```diff
@@ -3189,6 +3170,186 @@ def _delete_annotations(self, task_id):
...
+        assert annotation_file
+        assert self._annotations_count(original_annotations) > 0
+        return annotation_file, original_annotations
+
+    @pytest.mark.timeout(60)
+    @pytest.mark.parametrize(
+        ("import_mode", "should_append"),
+        [
+            pytest.param(None, False, id="replace-default"),
+            pytest.param("append", True, id="append"),
```
zhiltsov-max, line 3306: Probably, there should also be a test for explicit "replace"

tests/python/rest_api/test_tasks.py
```diff
@@ -3189,6 +3170,186 @@ def _delete_annotations(self, task_id):
...
+
+    def _has_sentinel(self, annotations: dict[str, Any]) -> bool:
+        return any(tag["group"] == self._SENTINEL_GROUP for tag in annotations["tags"])
+
+    def _annotations_count(self, annotations: dict[str, Any]) -> int:
+        return sum(
+            len(annotations[annotation_type]) for annotation_type in ("tags", "shapes", "tracks")
+        )
+
+    def _import_annotations_file(
```
zhiltsov-max, line 3203: Please use functions from `tests/python/rest_api/utils.py`. We have functions for importing and exporting there already.

changelog.d/20260508_150658_grigorio714_append_annotations_upload.md
```diff
@@ -0,0 +1,4 @@
+### Added
+
+- Added append mode for task and job annotation uploads.
```
zhiltsov-max:
```suggestion
- An option to import annotations without removing the existing ones
  in task and job annotation uploads.
```

tests/python/rest_api/test_tasks.py
zhiltsov-max, line 1: Could you support this new import option in the high-level SDK as well?

zhiltsov-max:
Closes https://github.com/cvat-ai/cvat/issues/4747

cvat-ui/src/actions/import-actions.ts
zhiltsov-max, line 1: Why a radio button is used? I think we don't have them anywhere in CVAT. Probably, only in annotation attributes. I think, it looks a little bit odd in the import dialog.

Grigorii777: Well, we have to choose one of the import options
And there are only two of them.
Generally speaking, if we want both options to be visible on the screen, we use a radio button.
The only alternative is a single-select checkbox, but that would confuse the user.
And then there’s the drop-down list, but that requires two clicks instead of one.

zhiltsov-max: I think, it would be better to use a checkbox so far, because there are only 2 options now and it's quite clear what a checkbox would mean. Or change the radio button visuals to be aligned with the existing elements.

Grigorii777: Do you mean just a single checkbox for the replace mode?

zhiltsov-max: It's one of the options, yes.

Grigorii777: For context
image

Grigorii777: I think that import mode is an important setting that can significantly alter the annotations, so it’s worth showing both options and which one is selected
So I’ll do it the way we do it in our system: like "Create task → Advanced configuration → Sorting method"

zhiltsov-max: Ok, it works for me too.

tests/python/sdk/test_annotation_import.py
```diff
@@ -0,0 +1,100 @@
...
+@pytest.mark.parametrize("proxy_cls", [Task, Job])
+def test_import_annotations_passes_import_mode_to_uploader(
+    monkeypatch: pytest.MonkeyPatch, tmp_path: Path, proxy_cls
+):
+    annotation_file = tmp_path / "annotations.zip"
+    annotation_file.write_bytes(b"test")
+    endpoint = SimpleNamespace(path="/api/annotations/")
+    captured = {}
+
+    def upload_file_and_wait(self, endpoint_arg, filename_arg, format_name_arg, **kwargs):
```
zhiltsov-max: We typically have tests that check the results on the server. But if you want to limit it just by the request check - it should probably check the actual request being sent, not an intermediate class in the high-level SDK implementation. So my suggestion would be to join the 2 tests in this file.

cvat-ui/src/actions/import-actions.ts
Grigorii777: image

tests/python/sdk/test_annotation_import.py
zhiltsov-max: The updated version is certainly correct, but it seems to be quite overcomplicated. You could test all of this in ~15 lines of code:

```python
    @parametrize("overwrite", [True, False])
    def test_can_overwrite_annotations(self, fxt_new_task: Task, overwrite: bool, tmp_path: Path):
        # create the initial annotations with set_annotations()

        original_annotations = fxt_new_task.get_jobs()[0].get_annotations()
        assert len(original_annotations.shapes) == 1

        exported_dataset = fxt_new_task.export_dataset(
            format_name="CVAT for images 1.1", filename=temp_dir, include_images=False,
        )

        fxt_new_task.import_annotations(
            format_name="CVAT for images 1.1", filename=exported_dataset, overwrite=overwrite,
        )

        imported_annotations = fxt_new_task.get_jobs()[0].get_annotations()
        assert len(imported_annotations.shapes) == 1 + int(not overwrite)
```

cvat-ui/src/components/import-dataset/import-dataset-modal.tsx
```diff
@@ -481,6 +504,11 @@ function ImportDatasetModal(props: StateToProps): JSX.Element {
     };
 
     const confirmUpload = (): void => {
+        if (uploadParams.importMode === 'append') {
+            onUpload();
+            return;
```
bsekachev: I believe we need to show a confirmation anyway. Just provide another message.
Even if it does not remove existing annotations, it is a long process which cannot be currently aborted. Additionally, further it will be difficult to easily remove appended annotations.
So, we still need confirmation

cvat-ui/src/actions/import-actions.ts
```diff
@@ -79,6 +79,7 @@ export const importDatasetAsync = (
     sourceStorage: Storage,
     file: File | string,
     convMaskToPoly: boolean,
+    importMode: 'replace' | 'append' = 'replace',
```
bsekachev: The only usage is in 'import-dataset-modal.tsx'.
So, just remove default value for this parameter

bsekachev (commented):
Could you please attach a screenshort of upload dialog?

Grigorii777:
> Could you please attach a screenshort of upload dialog?
It was mentioned in the discussion above

image

cvat-ui/src/components/import-dataset/import-dataset-modal.tsx
```diff
@@ -504,21 +504,21 @@ function ImportDatasetModal(props: StateToProps): JSX.Element {
...
+                    'This process cannot be aborted after it starts. Continue?' :
+                `You are going to upload new annotations to ${instanceType}. Continue?`,
             className: `cvat-modal-content-load-${instanceType.split(' ')[0]}-annotation`,
             onOk: () => {
                 onUpload();
             },
             okButtonProps: {
                 type: 'primary',
-                danger: true,
+                danger: !isAppend,
```
bsekachev: Just keep it danger always

cvat-ui/src/components/import-dataset/import-dataset-modal.tsx
```diff
@@ -504,21 +504,21 @@ function ImportDatasetModal(props: StateToProps): JSX.Element {
...
+        const isAppend = uploadParams.importMode === 'append';
 
         confirm({
-            title: 'Current annotation will be lost',
-            content: `You are going to upload new annotations to ${instanceType}. Continue?`,
+            title: isAppend ? 'Append annotations?' : 'Current annotation will be lost',
+            content: isAppend ?
+                `You are going to append new annotations to ${instanceType}. ` +
+                    'This process cannot be aborted after it starts. Continue?' :
+                `You are going to upload new annotations to ${instanceType}. Continue?`,
```
bsekachev: "You are going to upload new annotations" is applicable for both append and replace mode

As an idea

Image

bsekachev (approved):

tests/python/sdk/test_annotation_import.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+@pytest.mark.parametrize(
+    ("import_mode", "expected_shape_count"),
+    [
+        ("replace", 1),
+        ("append", 2),
+    ],
+)
+def test_import_annotations_respects_import_mode(
```
zhiltsov-max, line 20: Looks mostly good to me, except that the check for job import call is missing now.

zhiltsov-max (approved):

PR #10575, opened 2026-05-08
Prevent source files from being modified by the Django user

SpecLad:
### Motivation and context

Currently, all source files are located in `/home/django`, and owned by `django`. This means that if there's any vulnerability that lets an attacker overwrite files in the server, they can replace source files with their own code, and potentially get that code executed. That's pretty bad, so I want to harden against that.

Make all source files owned by root, and move them to `/opt/cvat`. Add a `manage.py` symlink in `/home/django` for backwards compatibility. It happens that if a script is a symlink, Python does not add the symlink's directory to `sys.path`, which is great for us, since that lets us avoid a writable directory on there.

Still, even though `/home/django/manage.py` is owned by root, an attacker could potentially be able to delete it and replace it with their own malicious file. To be a bit more safe, replace `~/manage.py` calls in backend scripts with `django-admin`.

To make sure CVAT can still find the data directory, add a new environment variable, `CVAT_BASE_DIR` and set it in the Docker image.

This also fixes a minor bug: we no longer override the `HOME` environment variable in the `Dockerfile`, so now it's automatically set by `Docker` depending on the current user.

### Checklist

### License

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -2,10 +2,10 @@
 branch = true
 
 source =
-    cvat/apps/
+    cvat.apps
+    utils.dataset_manifest
     cvat-sdk/
     cvat-cli/
-    utils/dataset_manifest
 
 omit =
     cvat/settings/*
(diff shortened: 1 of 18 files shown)


azhavoro (approved):
LGTM, great job!

PR #10578, opened 2026-05-11
[DEV-866] Webhooks: add broker-level retry on delivery

Rwwwrl:
### Title
[DEV-866] Webhooks: add broker-level retry on delivery

### Motivation and context

Failed webhook deliveries were dropped after the first attempt — one 5xx or transport error against a subscriber lost the event.

This adds broker-level retry on `send_webhook` via `rq.Retry` with a Junction-style backoff: 5s → 5min → 30min → 3h → 24h × 4 (8 retries, ~4-day window). 5xx, connection errors, and timeouts retry; 4xx is terminal.

Folded-in prerequisites:
- `send_webhook` now takes `webhook_id` and re-fetches the row on every attempt, so deactivated webhooks are skipped between retries.
- `ping` goes through a new `_perform_webhook_request` helper directly so subscriber 5xxs don't surface as API 500s.
- `--with-scheduler` is enabled on `cvat_worker_webhooks` (`docker-compose.yml` + Helm `worker_webhooks` Deployment). Without it, rq's built-in scheduler doesn't run and delayed retries silently sit in `ScheduledJobRegistry` forever. `backend_entrypoint.sh` was extended to forward extra `--*` args past the queue list into the inner `rqworker` invocation.

### Checklist

### License

### DIFF
--- a/cvat/apps/webhooks/exceptions.py
+++ b/cvat/apps/webhooks/exceptions.py
@@ -3,5 +3,5 @@
 # SPDX-License-Identifier: MIT
 
 
-class WebhookDeliveryRetryable(Exception):
+class WebhookDeliveryError(Exception):
     pass
--- a/cvat/apps/webhooks/exceptions.py
+++ b/cvat/apps/webhooks/exceptions.py
@@ -1,0 +1,7 @@
+# Copyright (C) CVAT.ai Corporation
+#
+# SPDX-License-Identifier: MIT
+
+
+class WebhookDeliveryRetryable(Exception):
+    pass
--- a/cvat/apps/webhooks/utils.py
+++ b/cvat/apps/webhooks/utils.py
@@ -55,7 +55,7 @@ def perform_webhook_request(webhook: Webhook, payload: dict) -> tuple[int, str]:
 
         response = ""
         if response_body is not None and len(response_body) < _RESPONSE_SIZE_LIMIT + 1:
-            response = response_body.decode("utf-8")
+            response = response_body.decode("utf-8", errors="replace")
 
         return status_code, response
     except requests.ConnectionError as ex:
--- a/cvat/apps/webhooks/utils.py
+++ b/cvat/apps/webhooks/utils.py
@@ -2,8 +2,21 @@
 #
 # SPDX-License-Identifier: MIT
 
+import hashlib
+import hmac
+import json
+from http import HTTPStatus
+
+import requests
+
 from cvat.apps.engine.serializers import BasicUserSerializer
 from cvat.apps.events.handlers import get_request, get_user
+from cvat.utils.http import PROXIES_FOR_UNTRUSTED_URLS, make_requests_session
+
+from .models import Webhook
+
+_WEBHOOK_TIMEOUT = 10
+_RESPONSE_SIZE_LIMIT = 1 * 1024 * 1024  # 1 MB
 
 
 def get_sender(instance) -> dict:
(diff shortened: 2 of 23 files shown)


cvat/apps/webhooks/exceptions.py
```diff
@@ -0,0 +1,7 @@
+# Copyright (C) CVAT.ai Corporation
+#
+# SPDX-License-Identifier: MIT
+
+
+class WebhookDeliveryRetryable(Exception):
```
zhiltsov-max: Consider renaming to `FailedDelivery` or `DeliveryError` to make it clear that this is an error.

cvat/apps/webhooks/signals.py
```diff
@@ -79,15 +100,24 @@ def send_webhook(webhook, payload, redelivery=False):
...
     return delivery
 
 
 def add_to_queue(webhook, payload, redelivery=False):
     queue = django_rq.get_queue(settings.CVAT_QUEUES.WEBHOOKS.value)
-    queue.enqueue_call(func=send_webhook, args=(webhook, payload, redelivery))
+    queue.enqueue_call(
+        func=send_webhook,
+        args=(webhook.id, payload, redelivery),
+        retry=Retry(max=len(_SEND_WEBHOOK_TASK_RETRIES), interval=_SEND_WEBHOOK_TASK_RETRIES),
```
zhiltsov-max: This solution works, but it retries the job after any failures. It doesn't seem intended, could you check if it's possible to retry only when the appropriate exception is raised? Maybe via a custom failure handler?

cvat/apps/webhooks/signals.py
```diff
@@ -68,9 +72,26 @@ def send_webhook(webhook, payload, redelivery=False):
...
     if response_body is not None and len(response_body) < RESPONSE_SIZE_LIMIT + 1:
-        response = response_body.decode("utf-8")
+        response_text = response_body.decode("utf-8")
+    return status_code, response_text
+
+
+def send_webhook(
+    webhook_id: int, payload: dict, redelivery: bool = False
+) -> WebhookDelivery | None:
+    # NOTE @sosov: pre-DEV-866 jobs pickled a Webhook instance as args[0].
```
zhiltsov-max: Consider adding a redis migration, they are handled by the redis_handler app.

Rwwwrl: Added. Had not known about that

cvat/apps/webhooks/signals.py
Rwwwrl: > but it retries the job after any failures

Yes, that is actually intended. We should retry request in case of any exception EXCEPT when consumer server return any non-retryable status, like 400, for example

Do not see any problems, there. Could you describe your concerns?

zhiltsov-max: I actually thought about errors like `IndexError` or `AttributeError` that can potentially be raised by the code because of our errors. But it looks like this kind of errors should be found by tests and is not expected to happen in production. And even if it does, the retry logic will not hurt significantly. Could you add at least 1 test for the new retry logic? I think, it could be a unit test that mocks the redelivery intervals to more testable values.

cvat/apps/webhooks/signals.py
```diff
@@ -31,14 +32,17 @@
 from cvat.utils.http import PROXIES_FOR_UNTRUSTED_URLS, make_requests_session
 
 from .event_type import EventTypeChoice, event_name
+from .exceptions import WebhookDeliveryError
 from .models import Webhook, WebhookDelivery, WebhookTypeChoice
 
 WEBHOOK_TIMEOUT = 10
 RESPONSE_SIZE_LIMIT = 1 * 1024 * 1024  # 1 MB
 
+_SEND_WEBHOOK_TASK_RETRIES = [5, 300, 1_800, 10_800, 86_400, 86_400, 86_400, 86_400]
```
zhiltsov-max: I think it makes sense to allow configuration for this variable. Consider integrating it with django settings. An example for this can be found in `cvat/apps/engine/default_settings.py` and `cvat/apps/engine/apps.py`.

cvat/apps/webhooks/signals.py
Rwwwrl: Added tests

cvat/apps/webhooks/signals.py
Rwwwrl: Moved to webhooks/default_settings.py

tests/python/rest_api/test_webhooks_sender.py
```diff
@@ -774,3 +793,58 @@ def test_not_project_staff_cannot_redeliver(self, projects, users):
...
+        project_id = list(projects)[0]["id"]
+
+        webhook_id = self._create_webhook(target_url(), project_id)
+        self._trigger_update(project_id, "trigger_success_no_retry")
+
+        deliveries, _ = get_deliveries(webhook_id, expected_count=1)
+        assert deliveries["results"][0]["status_code"] == HTTPStatus.OK
+
+        # NOTE @sosov: wait past settings.SEND_WEBHOOK_TASK_RETRIES[0] to confirm no retry fires
+        sleep(10)
```
zhiltsov-max: Using sleeps for thread synchronization is very unreliable in tests, especially with multiple processes. Using sleep here slows down the test for no actual reason, and can result in random failures. That's why I suggested doing a unit test with mocks instead. Consider replacing it with a unit test that checks the retry is scheduled after a failed delivery.

Rwwwrl: updated: see TestSendWebhook and TestAddToQueue

zhiltsov-max:
@codex review

cvat/apps/webhooks/utils.py
```diff
@@ -12,3 +25,40 @@ def get_sender(instance) -> dict:
...
+                timeout=_WEBHOOK_TIMEOUT,
+                stream=True,
+                proxies=PROXIES_FOR_UNTRUSTED_URLS,
+            )
+            status_code = response.status_code
+            response_body = response.raw.read(_RESPONSE_SIZE_LIMIT + 1, decode_content=True)
+
+        response = ""
+        if response_body is not None and len(response_body) < _RESPONSE_SIZE_LIMIT + 1:
+            response = response_body.decode("utf-8")
```
zhiltsov-max:
```suggestion
            response = response_body.decode("utf-8", errors="ignore")
```

I think codex was right about the issue, consider updating. There's no point in retrying a delivered webhook.

Rwwwrl: Updated, but made it using "replace" over "ignore", otherwise it would lead to information loss

zhiltsov-max (approved):

PR #10585, opened 2026-05-12
Ass/9613 webhook export backup

### LINKED ISSUE #9613: Need a webhook for requests

Hello ! I am currently using cvat for AI learning. When i export the project at YOLO format, it can take a lot of time due to the amount of data ( 25GB ). In my backend, I have no way to know that the export is done and ready to download, I am obliged to do some polling on the req_id.

Is it possible to add the "resquest" to the webhook workflow ? The request linked to a project so it should not be difficult to send a webhook whenever the export is ready.
Thanks !

_No response_

### Additional context

_No response_

Rwwwrl:
### Title
Fix #9613: webhook events for dataset export and backup completion

### Motivation and context

Dataset export and backup are long-running RQ jobs with no model and no Django signal — to learn when their export/backup finishes, subscribers have to poll `/api/requests`. Add two webhook events, `create:export` and `create:backup`, that fire once per terminal outcome (`completed` / `failed`) carrying the request metadata. The payload mirrors `/api/requests` fields (`target`, `target_id`, `format` / `lightweight`, `rq_id`, `status`, `message`) so subscribers can correlate webhook deliveries to request rows. `LockNotAvailableError` continues to retry the RQ job without firing — only terminal outcomes fire.

Refactor along the way: `cvat/apps/webhooks/signals.py` is split into `dispatch.py` / `services.py` / `tasks.py` / `utils.py`, leaving the module with only the Django signal receivers. `select_webhooks(instance, ...)` becomes `select_webhooks(*, event, organization_id, project_id, ...)`; callers resolve the two ids themselves via `cvat.apps.events.handlers`, so the function no longer takes a Django ORM instance.

Fixes #9613.

### Checklist

### License


cvat/apps/dataset_manager/util.py
```diff
@@ -20,11 +20,51 @@
...
 from datumaro.util.os_util import make_file_name
 from django.conf import settings
 from pottery import Redlock
 
+from cvat.apps.engine.models import Job, Project, Task
+
+from .enums import ExportStatus
+
+
+def queue_export_created_event(
```
zhiltsov-max: This function should probably be either in `dataset_manager/signals.py` or `events/handlers.py`, along with others. Maybe, we should start extracting them into `appname/event_handlers.py` - also an option.

cvat/apps/dataset_manager/views.py
```diff
@@ -175,6 +179,11 @@ def export(
         ):
             if osp_exists(output_path):
                 extend_export_file_lifetime(output_path)
+                queue_export_created_event(
```
zhiltsov-max: Consider creating a signal instead.

cvat/apps/dataset_manager/views.py
```diff
@@ -216,10 +223,23 @@ def export(
             )
         )
         raise
-    except Exception:
+    except Exception as exc:
+        queue_export_created_event(
+            target=db_instance,
+            dst_format=dst_format,
+            status=ExportStatus.FAILED,
+            message=str(exc),
```
zhiltsov-max: The exception type should probably be filtered somehow to return only user-readable messages. I'd probably start with `datumaro.DatasetExportError` or reuse `parse_exception_message()` from failed rq job handling.

cvat/apps/dataset_manager/views.py
Rwwwrl: Rewrote, thank you

cvat/apps/dataset_manager/views.py
Rwwwrl: Added formatting + parse_exception_message, now it wroks exactly like during collection response for 
https://app.cvat.ai/api/requests/<rq_id> request

cvat/apps/dataset_manager/util.py
Rwwwrl: Updated:
1. put signals definition in appname/signals.py
2. put signals handlers in webhook/event_handlers.py

cvat/apps/engine/enums.py
```diff
@@ -0,0 +1,10 @@
+# Copyright (C) CVAT.ai Corporation
+#
+# SPDX-License-Identifier: MIT
+
+from django.db.models import TextChoices
+
+
+class BackupStatus(TextChoices):
```
zhiltsov-max: We store most enums either in models.py or next to the related interface. In this case - it can probably be added in `backup_signals.py`

cvat/apps/webhooks/event_handlers.py
```diff
@@ -0,0 +1,63 @@
...
+from cvat.apps.events.handlers import organization_id as resolve_organization_id
+from cvat.apps.events.handlers import project_id as resolve_project_id
+
+from .dispatch import batch_add_to_queue
+from .event_type import event_name
+from .services import select_webhooks
+
+
+@receiver(export_finished)
+def handle_export_finished(sender, *, target, dst_format, status, message="", **kwargs):
```
zhiltsov-max: Please add type annotations for new functions, where possible.

cvat/apps/engine/enums.py
Rwwwrl: moved to signals.py

cvat/apps/webhooks/event_handlers.py
Rwwwrl: added

cvat/apps/redis_handler/tests/test_redis_migration_naming.py
```diff
@@ -0,0 +1,48 @@
...
+            for app_label, path in _iter_redis_migration_files()
+            if not _FILENAME_PATTERN.match(path.name)
+        ]
+        self.assertFalse(
+            bad,
+            "Redis migration filenames must match NNN_snake_case.py "
+            f"(zero-padded 3-digit prefix). Offending files: {bad}",
+        )
+
+    def test_numbers_unique_per_app(self):
```
zhiltsov-max: I can see the duplicates are checked in the test now. Shouldn't this check be performed at the migration attempt instead?

Rwwwrl: I might be misstaken, but I did not find any checks during migration applying or in CICD

In last project I had it in test (that is why I actually added it), where getting green result was mandatory during PR

zhiltsov-max: That's what I'm talking about - it should just be a check before migration application attempt or a separate check on CI. Having it in the tests is a hacky solution.

Rwwwrl: > a separate check on CI

But we already have running tests in CI, so it is almost identical, except you can get that problem even through just running tests, not through any separate script

Rwwwrl: updated: created dev script, added to ci

zhiltsov-max: This version works, but I'd prefer it to work similarly to the `manage.py migrate` command, which checks for naming conflicts before applying migrations. While the CI check works in general, it supposes that the validation script is executed manually. If the check is performed before application, it protects the user automatically, without extra actions from their side. We call it the backend startup script, so it will just work. Doing so will also help to keep the validation logic in the same app.

Rwwwrl: Updated:
- created command checkredismigrations.py
- added executing checkredismigrations.py before migration running in cvat/apps/redis_handler/management/commands/migrateredis.py through call_command

cvat/apps/redis_handler/management/commands/checkredismigrations.py
```diff
@@ -0,0 +1,56 @@
...
+# SPDX-License-Identifier: MIT
+
+import re
+from collections import Counter
+
+from django.core.management.base import BaseCommand, CommandError
+
+from cvat.apps.redis_handler.migration_loader import MigrationLoader
+
+_MIGRATION_FILENAME_PATTERN = re.compile(r"^\d{3}_[a-z0-9_]+$")
```
zhiltsov-max: Why the pattern is limited by only 3 digits? I don't see any references to this number in the migration loader.

zhiltsov-max (approved):

zhiltsov-max:
@codex review

zhiltsov-max:
@Rwwwrl, actually, codex seems to be right. However, it looks like a part of a bigger issue - it the uploading fails, it will not be retried later either. So I think it makes sense to try to handle these issues in a separate PR.

PR #10594, opened 2026-05-13
Allow drop-in replacement for strings with enums

Grigorii777:
### Motivation and context

This change makes generated string enum models in the lower-level SDK compatible with plain strings.

Previously, SDK enum values were generated as `ModelSimple` instances. Because of that, replacing string fields in serializers with enum/choice fields could break SDK users: enum values were not directly comparable with strings and required accessing `.value`.

The generated string enum models now inherit from `str`, so they can be used as drop-in replacements for string values:

- `models.ShapeType("rectangle") == "rectangle"`
- `isinstance(models.ShapeType("rectangle"), str)`
- enum values can be used as dict keys compatible with string keys
- enum values serialize to JSON as plain strings

This allows cleaner serializer definitions while preserving compatibility in the SDK API.

### How has this been tested?

Tested locally with the SDK API client tests.

Also regenerated the SDK locally from the updated OpenAPI generator templates and verified that generated string enum models inherit from `StringEnumModel`.

### Checklist

- ~[ ] I have updated the documentation accordingly~
- ~[ ] I have linked related issues~

### License


cvat-sdk/gen/templates/openapi-generator/model_utils.mustache
```diff
@@ -371,6 +371,54 @@ class ModelSimple(OpenApiModel):
...
+        if isinstance(value, ModelSimple):
+            value = value.value
+
+        return str.__new__(cls, value)
+
+    def __setattr__(self, attr, value):
+        if attr == 'value' and isinstance(value, ModelSimple):
+            value = value.value
+
+        super().__setattr__(attr, value)
```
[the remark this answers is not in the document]
Grigorii777: str. __setattr__ <- object. __setattr__

MRO = ModelSimple, object

cvat-sdk/gen/templates/openapi-generator/model.mustache
```diff
@@ -13,6 +13,7 @@ from {{packageName}}.model_utils import (  # noqa: F401
     ModelComposed,
     ModelNormal,
     ModelSimple,
+    StringEnumModel,
```
SpecLad: All the other model classes start with `Model`, so maybe this should too?

cvat-sdk/gen/templates/openapi-generator/model_utils.mustache
```diff
@@ -371,6 +371,54 @@ class ModelSimple(OpenApiModel):
...
+
+    def __new__(cls, *args, **kwargs):
+        value = kwargs.get('value')
+        if args:
+            value = args[0]
+        elif 'value' not in kwargs:
+            value = getattr(cls, '_default_value', '')
+
+        if value is None and is_type_nullable(cls):
+            return None
```
SpecLad, line 389: What is this meant to do?

cvat-sdk/gen/templates/openapi-generator/model_utils.mustache
```diff
@@ -371,6 +371,54 @@ class ModelSimple(OpenApiModel):
...
+
+        if isinstance(value, ModelSimple):
+            value = value.value
+
+        return str.__new__(cls, value)
+
+    def __setattr__(self, attr, value):
```
SpecLad: Could you comment the purpose of this and the following methods? It's rather unclear.

SpecLad:
It seems that this implementation exhibits strange behavior if an object's `value` is reassigned:

```
>>> st = models.ShapeType("rectangle")
>>> st.value = "ellipse"
>>> st
'rectangle'
>>> st.value
'ellipse'
```

I doubt this could be fixed if we keep `str` as the base class of `StringEnumModel`. Do you suppose there's any way to still achieve what we want without adding `str` as a base class?

SpecLad:
By the way, we have a bunch of places in the SDK where we explicitly access `.value` - could you remove those accesses? It would clean up the code and serve as additional tests for the change.

cvat-sdk/gen/templates/openapi-generator/model_utils.mustache
Grigorii777: This preserves nullable string enum values as Python None. Since ModelStringEnum inherits from str, passing None into str.__new__ would create an inconsistent string value instead of preserving the OpenAPI nullable semantics.

cvat-sdk/gen/templates/openapi-generator/model_utils.mustache
```diff
@@ -371,6 +371,85 @@ class ModelSimple(OpenApiModel):
...
+
+    def __getitem__(self, name):
+        # Support model-style access such as enum["value"], while preserving
+        # normal string indexing/slicing for integer and slice keys.
+        if isinstance(name, str):
+            if name in self.required_properties and name in self.__dict__:
+                return self.__dict__[name]
+
+            if name in self.__dict__.get('_data_store', {}):
+                return self.__dict__['_data_store'][name]
```
SpecLad, line 434: Any reason this isn't `ModelSimple.__getitem__(self, name)`?

cvat-sdk/gen/templates/openapi-generator/model_utils.mustache
```diff
@@ -371,6 +371,85 @@ class ModelSimple(OpenApiModel):
...
+
+    def set_attribute(self, name, value):
+        if name == 'value':
+            value = self._prepare_value_assignment(value)
+
+        super().set_attribute(name, value)
+
+    def __getattr__(self, attr):
+        # str appears before ModelSimple in the MRO, so restore OpenAPI model
+        # attribute lookup for required properties and _data_store values.
```
SpecLad: I don't think I get it. What would break if this method wasn't here? Since `str.__getattr__` doesn't exist, wouldn't `ModelSimple.__getattr__` still be invoked?

PR #10595, opened 2026-05-13
Added ability to change username on the profile page

bsekachev:
### Motivation and context

This change enables username updates through the existing `PATCH /api/users/{id}` API and exposes the capability on the profile page.

The profile form now allows editing the username, sends only changed user fields, and refreshes the current user state from the PATCH response so the header and profile page show the updated username without ending the session. UI username validation now keeps the existing minimum length of 5 characters while matching the server character and max-length restrictions.

### How has this been tested?

Added REST API coverage for:
- updating the current user's username via `PATCH /api/users/{id}`
- rejecting invalid username values
- preventing a non-admin from updating another user's username

Updated Cypress profile-page coverage to verify:
- username is included in the profile PATCH request
- the response contains the updated username
- the profile greeting and header user dropdown show the new username

Not run locally: targeted pytest/Cypress verification was blocked in the current workspace.

### Checklist

### License


bsekachev:
@codex review

PR #10596, opened 2026-05-14
Implemented attribute deletion

bsekachev:
### Motivation and context

This PR adds end-to-end support for deleting label attributes.

Previously CVAT allowed creating and updating label attributes, but saved attributes could not be removed cleanly after task/project creation. This affected regular label attributes and skeleton element attributes. Removing an attribute now also removes corresponding annotation attribute values through the existing database cascade.

Supported attribute deletion flows:
- Delete a label attribute with `PATCH /api/labels/<label_id>` and `attributes: [{ ..., deleted: true }]`.
- Delete a label attribute through task/project label update paths with `PATCH /api/tasks/<task_id>` or `PATCH /api/projects/<project_id>` and `labels: [{ id: <label_id>, attributes: [{ ..., deleted: true }] }]`.
- Delete a skeleton element attribute with `PATCH /api/labels/<parent_skeleton_label_id>` and `sublabels: [{ id: <sublabel_id>, attributes: [{ ..., deleted: true }] }]`.
- Modifying skeleton elements individually is not supported (old behavior), e.g. `PATCH /api/labels/<sublabel_id>` returns `400`.

Main changes:
- Added write-only `deleted` support for label attributes in the REST API.
- Implemented server-side deletion of `AttributeSpec`.
- Applied nested skeleton sublabel attribute updates through parent skeleton label PATCH.
- Added nested scope validation so sublabel updates must belong to the patched parent skeleton.
- Updated the label editor to preserve saved attribute deletions in PATCH payloads.
- Added UI confirmation/warning text before deleting saved attributes because annotation values will be removed.
- Added recursive raw label editor deletion warnings for saved attributes, including skeleton element attributes.
- Allowed adding, editing, and deleting skeleton element attributes after task/project creation while keeping saved skeleton structure editing disabled.
- Updated client label saving to use `Task.save({ labels })` / `Project.save({ labels })` instead of mutating `task.labels` / `project.labels`.
- Made task/project labels read-only on the client.

### How has this been tested?

Added REST API coverage for:
- Deleting a regular label attribute via `PATCH /api/labels/<id>`.
- Deleting regular label attributes through task/project label update paths.
- Deleting a root skeleton label attribute via `PATCH /api/labels/<id>`.
- Deleting a skeleton element attribute through parent skeleton label PATCH.
- Rejecting direct skeleton element updates via `PATCH /api/labels/<sublabel_id>`.
- Rejecting attribute deletion without an ID.
- Rejecting unknown attribute deletion.
- Rejecting deletion of an attribute from another label.

### Checklist

### License

image
image
image


bsekachev:
@codex review

bsekachev:
@codex review

cvat/apps/engine/tests/test_backup.py
zhiltsov-max, line 1: Hos this file is related to the PR?

bsekachev: Hmm, probably not

cvat/apps/engine/serializers.py
```diff
@@ -596,15 +620,36 @@ def update_label(
...
+
+        def get_db_attr(attr_id: int) -> models.AttributeSpec:
+            try:
+                return models.AttributeSpec.objects.get(id=attr_id, label=db_label)
+            except models.AttributeSpec.DoesNotExist as ex:
+                raise exceptions.NotFound(
+                    f'Attribute with id #{attr_id} does not exist'
+                ) from ex
+
+        # Apply deletions before creates/updates. This keeps an atomic request valid
```
zhiltsov-max, line 637: This code has several race conditions possible, could you add a comment on why we're not handling them? Examples: if 2 concurrent requests try to add an attribute with the same name, the result is undetermined.

cvat/apps/engine/serializers.py
```diff
@@ -351,21 +351,32 @@ def to_internal_value(self, data):
...
+        return attrs
 
 
 class SublabelSerializer(serializers.ModelSerializer):
     id = serializers.IntegerField(required=False)
     attributes = AttributeSerializer(many=True, source='attributespec_set', default=[],
         help_text="The list of attributes. "
-        "If you want to remove an attribute, you need to recreate the label "
-        "and specify the remaining attributes.")
+        "To remove an attribute, pass the full attribute body with deleted=true. "
```
zhiltsov-max, line 378: It seems the description deviates from the implementation. The description says that the full attribute body is required, but only the `id` field is required by the implementation.

bsekachev: > but only the id field is required by the implementation 

This is not true. The implementation requires full body. 

I initially made other fields in AttributeSerializer optional, but it led to multiple issues even with annotation requests and sdk tests. So, I decided to left it as was.

cvat/apps/engine/serializers.py
bsekachev: There is probably no reason like "we do not fix race condition because of ..."
But were these race conditions introduced with this PR? Like, what if in the previous implementation we have 2 concurrent requests to add an attribute with the same name?
Or has this patch removed any comments related to race conditions?

bsekachev: I mean that if you see race condition, we want to fix it, instead of commenting. 
But if the condition was not introduced in this patch, it does not have to be fixed in this patch

zhiltsov-max: These issues weren't exactly introduced by this PR, I just suggest adding a comment about the reason (like "we don't think it's a significant issue") or a TODO comment.

cvat/apps/engine/serializers.py
zhiltsov-max, line 1: Swapping names of 2 existing attributes in a single request results in a 500 error.

cvat/apps/engine/serializers.py
bsekachev: Discovered that AttributeSpec has unique_together = ('label', 'name')
So, the second request will have IntegrityError

cvat/apps/engine/serializers.py
zhiltsov-max, line 1: When trying to swap attribute names in the UI, after you rename the second attribute and resolve the name conflict, the UI still displays the warning:

Image

1. rename old a1 -> a2 (get the warning)
2. rename old a2 -> a1 (the warning stays despite no conflict)

Image

bsekachev: This is exactly because of db uniquness for attr name. This cannot be easily solved, and again, not something new in this patch.

bsekachev: Truly not related to the PR, but I wanted to fix it.
However if users do this, they will later see code 500 from the server, you mentioned above. So, I actually not sure we need to fix it

bsekachev: But on the other hand we may workaround it from client, updating swapped attributes one by one

bsekachev: Never mind, we can't workaround it normally. Without like "temporary" attribute name

zhiltsov-max (approved):

PR #10597, opened 2026-05-14
Update the server Docker base image to Ubuntu 24.04

SpecLad:
### Motivation and context

The EOL of 22.04 is approaching, and it would be nice to use a newer Python in server code.

Some notes:

* This Ubuntu has Python 3.12 as the default, in which setuptools is no longer installed into venvs by default. So we don't need to delete it anymore.

* Remove the pip upgrade, since a) In this Ubuntu the default pip is already 24.0, and b) upgrading system packages is now kind of problematic (see <https://packaging.python.org/en/latest/specifications/externally-managed-environments/>).

* In this version of Ubuntu, there's already a user (`ubuntu`) with ID 1000. To preserve the UID of the Django user, delete that user.

* A few Python packages don't have wheels for Python 3.12, so bump them to versions that do.

* There was a bug in clickhouse-connect where it would unintentionally return TZ-aware datetime objects in some cases (<https://github.com/ClickHouse/clickhouse-connect/issues/629>). Now it's fixed, but we actually want aware datetimes, so change the connection settings to restore the old behavior.

### Checklist

### License


nmanovic:
@SpecLad , probably it makes sense to update the document https://docs.cvat.ai/docs/contributing/development-environment/

PR #10598, opened 2026-05-14
Retry downloading chunks after error

klakhov:
## Summary

This PR adds resumable/retryable annotation chunk downloads to improve annotation page reliability during transient network interruptions.

## Changes

- Added HTTP Range support for task/job data chunk endpoints:
  - `Range` request header
  - `206 Partial Content`
  - `416 Range Not Satisfiable`
  - `Accept-Ranges`, `Content-Range`, and normalized `Content-Length` headers
- Updated API schema for chunk range download responses.
- Reworked the download worker to use streaming `fetch` for chunk downloads.
- Added retry logic for transient download failures and network interruptions.
- Added resume support using `Range: bytes=<received>-` when a download fails after partial data was received.

## Tests

- Added REST API test:
  - `tests/python/rest_api/test_task_data.py`
- Added E2E test:
  - `tests/cypress/e2e/actions_tasks/chunk_download_retry.js`

### Checklist

### License


bsekachev (commented):
We have retry logic in `server-proxy.ts` which now does not seem relevant

cvat-core/src/download.worker.ts
```diff
@@ -1,14 +1,191 @@
...
+        nextReceivedBytes += value.byteLength;
+    }
+}
+
+async function fetchData(url: string, requestConfig): Promise<{
+    data: ArrayBuffer;
+    headers: Record<string, string>;
+}> {
+    const requestUrl = appendParams(url, requestConfig.params);
+    const requestHeaders = new Headers(requestConfig.headers || {});
```
bsekachev:
```suggestion
    const requestHeaders = new Headers(requestConfig.headers ?? {});
```

cvat-core/src/download.worker.ts
```diff
@@ -1,14 +1,191 @@
...
+    const retryAfter = response?.headers.get('retry-after');
+    if (retryAfter && /^\d+$/.test(retryAfter)) {
+        return +retryAfter * 1000;
+    }
+
+    return Math.min(BASE_RETRY_DELAY_MS * 2 ** retry, MAX_RETRY_DELAY_MS);
+}
+
+function shouldRetry(response: Response | null): boolean {
+    return !response || [408, 429, 502, 503, 504].includes(response.status);
```
bsekachev: Would you mind to add a short explanation as a code comment why each code is here? 
With 429  it is obvious, with others, less obvious

cvat-core/src/download.worker.ts
```diff
@@ -1,14 +1,191 @@
...
+}
+
+function shouldRetryError(response: Response | null): boolean {
+    return !response || response.ok || shouldRetry(response);
```
bsekachev: Why should we retry if response.ok?

cvat-core/src/download.worker.ts
bsekachev: Do we need requestHeaders?
Below you just use it to create headers. 
But headers also may be created from requestConfig.headers

cvat-core/src/download.worker.ts
```diff
@@ -1,14 +1,191 @@
...
+                    const contentRange = parseContentRange(response.headers.get('content-range'));
+                    if (!contentRange || contentRange.start !== receivedBytes) {
+                        throw new Error('Unexpected Content-Range header');
+                    }
+
+                    expectedSize = contentRange.total;
+                } else if (response.status === 200) {
+                    chunks.splice(0, chunks.length);
+                    receivedBytes = 0;
+                    expectedSize = null;
```
bsekachev: how server can answer 200 if it programmed to answer 206 when range header is set?

bsekachev: And what is logic behind this branch? We just remove everything
And then we have `if (expectedSize === null)` doing `mergeChunks(chunks, receivedBytes)`, with empty `chunks`

cvat-core/src/download.worker.ts
```diff
@@ -1,14 +1,191 @@
...
+                } else {
+                    throw new Error(`Unexpected response status: ${response.status}`);
+                }
+            } else {
+                expectedSize = +(response.headers.get('content-length') || 0) || null;
```
bsekachev: I am not sure that fallback to 0 in `content-length` is correct
How will it work further?

With such condition (receivedBytes (0) >= expectedSize (0)) it returns 

```{ 'content-length': 0 }```, but what it means?

**Review by bsekachev — COMMENTED** [review]
Please comment each code path in this pull request (asking specifically about frontend code).

File: cvat-core/src/download.worker.ts [hunk 6]
```diff
@@ -1,14 +1,191 @@
...
+            }
+
+            receivedBytes = await readResponse(response, chunks, receivedBytes);
+
+            if (expectedSize === null || receivedBytes >= expectedSize) {
```
bsekachev: I am not sure `>` should be here
Probably it means that something is broken.

cvat-core/src/download.worker.ts
```diff
@@ -1,14 +1,191 @@
...
+
+    while (true) {
+        const { done, value } = await reader.read();
+        if (done) {
+            return nextReceivedBytes;
+        }
+
+        chunks.push(value);
+        nextReceivedBytes += value.byteLength;
+    }
```
bsekachev: Do we need to support both cases when payload is already in body and streamed?

cvat-core/src/download.worker.ts
klakhov: Removed

cvat-core/src/download.worker.ts
klakhov: Removed

cvat-core/src/download.worker.ts
klakhov: This condition will return `null` if its 0. So there is no really case with 0. Re-written the confition to be more readable.

cvat-core/src/download.worker.ts
klakhov: Removed first case.

cvat-core/src/download.worker.ts
klakhov: Removed.

cvat-core/src/download.worker.ts
klakhov: We need that logic for resume case:

1. Server starts the chunk response successfully `206 Partial Content`.
2. Browser receives some bytes.
3. Network breaks while `reader.read()` is still reading the body.
4. `fetch()` already returned a successful `Response`, so `response.ok === true`.
5. The failure happens later, inside body streaming.

bsekachev: I understand the case, just having `response.ok` in function called `shouldRetryError` itself is not obvious

Maybe just replace it:

```if (retry < MAX_RETRIES && shouldRetryError(error, response)) {```

by

```if (retry < MAX_RETRIES && (error instanceof DownloadReadError || shouldRetryError(error, response))) {```

**Comment by bsekachev** [comment]
One more thought. 
I am not sure we want to consider downloading of a truncated body as a retry.
E.g. imagine a user has very unstable connection and data chunk 100 Mb

During downloading, the connection may be interrupted more times (15, 20, 30). 
But each time it will receive more and more bytes. 

If we consider truncated as a retry, the user hardly ever get the chunk on client.
If we do not consider, earlier or sooner, the user receives it. 

What do you think?

File: cvat/apps/engine/views.py [hunk 8]
```diff
@@ -638,20 +645,83 @@ def __init__(self, data_type: str, data_num: str | int | None, data_quality: str
...
+        range_unit, separator, ranges = range_header.partition("=")
+        if range_unit.strip().lower() != "bytes" or separator != "=" or "," in ranges:
+            raise ValidationError("Invalid Range header")
+
+        start_text, separator, end_text = ranges.strip().partition("-")
+        if separator != "-":
+            raise ValidationError("Invalid Range header")
+
+        if not start_text and not end_text:
+            raise ValidationError("Invalid Range header")
```
SpecLad: You ought to use a regex instead of all of this manual parsing. It would be much simpler.

cvat-core/src/download.worker.ts
klakhov: Removed this case from the function.

cvat/apps/engine/views.py
```diff
@@ -638,20 +645,83 @@ def __init__(self, data_type: str, data_num: str | int | None, data_quality: str
...
         )
+        self._request = request
 
     @abstractmethod
     def _get_frame_provider(self) -> IFrameProvider: ...
 
+    @staticmethod
+    def _parse_range(range_header: str, content_size: int) -> tuple[int, int]:
+        range_unit, separator, ranges = range_header.partition("=")
+        if range_unit.strip().lower() != "bytes" or separator != "=" or "," in ranges:
```
SpecLad: RFC 9110 says:

> An origin server MUST ignore a `Range` header field that contains a range unit it does not understand.

klakhov:
Makes sence. Added logic with `MAX_NO_PROGRESS_RETRIES`. If we have some progress, we try to continue.

cvat/apps/engine/views.py
```diff
@@ -638,20 +645,83 @@ def __init__(self, data_type: str, data_num: str | int | None, data_quality: str
...
+
+        range_header = self._request.headers.get("Range")
+        if not range_header:
+            return HttpResponse(data, content_type=chunk_data.mime, headers=chunk_headers)
+
+        try:
+            start, end = self._parse_range(range_header, content_size)
+        except ValidationError:
+            return HttpResponse(
+                status=status.HTTP_416_REQUESTED_RANGE_NOT_SATISFIABLE,
```
SpecLad, line 727: I don't think this response code is appropriate for all validation errors. It should be returned when the range is syntactically valid, but unsatisfiable (basically, when start >= length). If the range header has syntax errors, I think we should just return a regular 400.

cvat/apps/engine/views.py
```diff
@@ -1397,25 +1468,35 @@ def _handle_upload_backup(request: ExtendedRequest):
                 type=OpenApiTypes.INT,
                 description="A unique number value identifying chunk or frame",
             ),
+            OpenApiParameter(
+                "Range",
```
SpecLad: IMO, this should not be listed here, since it's just a standard HTTP header.

cvat/apps/engine/views.py
```diff
@@ -621,7 +622,13 @@ def _get_rq_response(queue, job_id):
 
 
 class _DataGetter(metaclass=ABCMeta):
-    def __init__(self, data_type: str, data_num: str | int | None, data_quality: str) -> None:
+    def __init__(
+        self,
+        request: ExtendedRequest,
```
SpecLad: This is subjective, but IMO, the view should parse the range header and pass just a range tuple (start, end) to the data getter. Otherwise we're passing duplicate data here (since `data_type`, `data_num`, and `data_quality` are all contained in `request`).

bsekachev:
@codex review

bsekachev (approved):
lgtm

cvat/apps/engine/views.py
klakhov: Added regex

cvat/apps/engine/views.py
klakhov: Adjusted.

Now _parse_range separates the cases:

- unsupported unit, like Range: items=0-99 -> ignored, full 200
- malformed byte range, like Range: bytes=abc -> 400 Bad Request
- syntactically valid but unsatisfiable byte range, like Range: bytes=<len>- -> 416

cvat/apps/engine/views.py
klakhov: Added case with 400 error

cvat/apps/engine/views.py
klakhov: Removed

cvat/apps/engine/views.py
klakhov: Applied.

SpecLad (approved):

PR #10600, opened 2026-05-14
Cancellable started export jobs

bsekachev:
### Motivation and context

This PR adds the ability to cancel export requests that are already in progress. Previously, CVAT could cancel only requests that were still queued or deferred, so a user had to wait for a long-running annotations, dataset, or backup export to finish even if the result was no longer needed.

Core ideas and design solutions:

- Preserve existing cancellation behavior for queued and deferred requests.
- Add started-request cancellation only for export jobs, because exports are non-mutating background operations.
- Keep started cancellation unsupported for mutating jobs such as imports, task creation, quality report calculation, consensus merge, and auto-annotation.
- Use RQ’s stop-job command for started export jobs so the worker stops the running job and RQ releases dependent deferred jobs normally.
- Preserve one-running-job-per-user export sequencing: when a started export is canceled, the next dependent export can continue.
- Keep debug `SimpleWorker` same-process for debugger ergonomics. It marks itself with `cvat_can_stop_started_jobs = 0`, and the API rejects started export cancellation for that worker type because there is no separate worker process to stop safely.
- Treat missing `cvat_can_stop_started_jobs` as “can stop” for backward compatibility with already running/default workers.
- Allow started export cancellation from the Requests page UI.

### How has this been tested?

Manually verified with lightweight checks:
- Python compile check for touched backend files.
- Reviewed request cancellation flow for queued/deferred, started export, started non-export, and SimpleWorker paths.

No automated tests were added for this change.

### Checklist

### License


cvat/apps/redis_handler/views.py
```diff
@@ -242,14 +286,41 @@
...
+                "Cancellation of started requests is supported only for export requests "
+                "executed by production workers"
+            )
+
+        # RQ will stop the forked work horse and enqueue dependent jobs from its stopped-job path.
+        queue = django_rq.get_queue(rq_job.origin)
+        try:
+            send_stop_job_command(queue.connection, rq_job.id, serializer=queue.serializer)
+        except InvalidJobOperation as ex:
+            return HttpResponseBadRequest(str(ex))
```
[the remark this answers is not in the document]
bsekachev: Fixed

cvat/apps/redis_handler/views.py
zhiltsov-max, line 1: Having at least 1 test for this functionality would be really nice.

cvat/apps/redis_handler/views.py
```diff
@@ -242,14 +282,42 @@ def cancel(self, request: ExtendedRequest, pk: str):
...
+        # Jobs that have not started yet are safe to cancel directly. RQ will also enqueue
+        # dependents when ONE_RUNNING_JOB_IN_QUEUE_PER_USER is enabled.
+        if rq_job_status in {RQJobStatus.QUEUED, RQJobStatus.DEFERRED}:
+            # FUTURE-TODO: race condition is possible here
+            rq_job.cancel(enqueue_dependents=settings.ONE_RUNNING_JOB_IN_QUEUE_PER_USER)
+            rq_job.delete()
+            return Response(status=status.HTTP_200_OK)
 
-        if rq_job.get_status(refresh=False) not in {RQJobStatus.QUEUED, RQJobStatus.DEFERRED}:
+        # Finished, failed, scheduled, and other non-started states cannot be interrupted.
```
zhiltsov-max, line 298: It may be mistaken, but cancelling scheduled jobs can actually be the second main use case, since we chain jobs per user. I know that canceling them is quite challenging though, so I don't insist on supporting cancellation for them.

cvat/apps/redis_handler/views.py
```diff
@@ -242,14 +282,42 @@ def cancel(self, request: ExtendedRequest, pk: str):
             return HttpResponseNotFound("There is no request with specified id")
 
         self.check_object_permissions(request, rq_job)
+        rq_job_status = rq_job.get_status(refresh=False)
+
+        # Terminal canceled jobs are not useful for users in the requests API. Remove them when
+        # they are observed after RQ has already finished cancellation handling.
+        if rq_job_status in {RQJobStatus.CANCELED, RQJobStatus.STOPPED}:
+            rq_job.delete()
```
zhiltsov-max: I think this delete can probably be removed, as it may create a race with task creation.

cvat/apps/redis_handler/views.py
bsekachev: > since we chain jobs per user 

It uses "deferred" mechanism. Such jobs already handled normally. 
As far as I know, "scheduled" status is not used in CVAT

bsekachev: gpt:

```
Yes, scheduled exists in django-rq / RQ.

In RQ, scheduled is a valid job status. It applies when a job has been scheduled to run in the future, for example through enqueue_in() or enqueue_at(), because apparently “run this later” needed its own little bureaucratic holding pen.

Example:

from datetime import timedelta
import django_rq

queue = django_rq.get_queue("default")

job = queue.enqueue_in(timedelta(minutes=10), my_func)

job.get_status(refresh=True)
# "scheduled"
```

cvat/apps/redis_handler/views.py
bsekachev: task creation? 
I do not think it can be currently cancelled. I believe task creation maybe queued, deferred, started, finished or failed.

zhiltsov-max: I meant new rq job creation for exports

cvat/apps/redis_handler/views.py
zhiltsov-max: Ok, my bad.

cvat/apps/redis_handler/views.py
bsekachev: okay, removed here and above

cvat/apps/redis_handler/views.py
bsekachev: Added several unit tests 

test_cancel_started_export_request_from_simple_worker_returns_bad_request (cvat.apps.redis_handler.tests.test_views.TestRequestViewSetCancel.test_cancel_started_export_request_from_simple_worker_returns_bad_request) ... ok
test_cancel_started_export_request_sends_stop_command (cvat.apps.redis_handler.tests.test_views.TestRequestViewSetCancel.test_cancel_started_export_request_sends_stop_command) ... ok
test_cancel_started_non_export_request_returns_bad_request (cvat.apps.redis_handler.tests.test_views.TestRequestViewSetCancel.test_cancel_started_non_export_request_returns_bad_request) ... ok

cvat/rqworker.py
```diff
@@ -30,6 +30,13 @@ class SimpleWorker(Worker):
 
     death_penalty_class = BaseDeathPenalty
 
+    def register_birth(self):
+        super().register_birth()
+        # The server process uses this flag to reject started-job cancellation in debug mode:
+        # SimpleWorker runs jobs in-process, so there is no forked work horse to stop.
+        # As we do not want just kill debug process
+        self.connection.hset(self.key, "cvat_can_stop_started_jobs", 0)
```
zhiltsov-max, line 38: This call should probably use the `CVAT_CAN_STOP_STARTED_JOBS_KEY` variable.

zhiltsov-max (approved):

PR #10602, opened 2026-05-14
Refactor redis_handler migration tests

zhiltsov-max:
## Summary

Redis migration tests previously wrote real `.py` files into installed apps' on-disk `redis_migrations/` dirs and relied on cleanup callbacks; a crash between create and cleanup left stray files behind.

- Updated migration tests to avoid writing to the real migration dirs
- For the bad-class case, also patch `get_class_from_module` so the loader never tries to import a non-existent module.
- Split the suite into one class per Django management command (`checkredismigrations`, `migrateredis`).

## Test plan
- [x] `python -m django test cvat.apps.redis_handler.tests.test_redis_migrations` — 5/5 pass
- [x] Verify `cvat/apps/*/redis_migrations/` dirs untouched after run

### DIFF
--- a/cvat/apps/redis_handler/tests/test_redis_migrations.py
+++ b/cvat/apps/redis_handler/tests/test_redis_migrations.py
@@ -2,82 +2,66 @@
 #
 # SPDX-License-Identifier: MIT
 
+import contextlib
 import io
-import os
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


zhiltsov-max:
@codex review

PR #10605, opened 2026-05-14
Make test DB dump output more stable

zhiltsov-max:
## Summary

Make the test DB fixture (`tests/python/shared/assets/cvat_db/data.json`) easy to update without manual cleanup of spurious diffs.

* Add `tests/python/shared/utils/dump_test_db.py`, a wrapper around `manage.py dumpdata` that post-processes the output for stable diffs:
  * records sorted alphabetically by `(model, pk)`;
  * field keys sorted alphabetically per record;
  * known unordered many-to-many fields (e.g. `auth.user.groups`) sorted;
  * trailing `.000` milliseconds (a `dumpdata` → `loaddata` → `dumpdata` round-trip artifact in `DjangoJSONEncoder`) stripped so the round-trip becomes idempotent.
* By default, preserve volatile timestamp fields (`last_login`, `*_updated_date`, `last_used_date`, ...) from the existing `data.json` so re-dumping after a no-op test run produces zero diff. Use `--refresh-volatile` to take new values instead.
* Add `loaddata_sorted` management command (`cvat/apps/engine/management/commands/loaddata_sorted.py`) that topologically reorders the JSON fixture by FK dependencies via `django.core.serializers.sort_dependencies` before calling `loaddata`. This lets us commit `data.json` in any order (alphabetical for stable diffs across schema changes) and still load it correctly when natural-foreign keys point at not-yet-loaded NOT NULL columns. Supports reading the fixture from stdin (`-`, default) or a file path.
* Wire `dump_db()` in `tests/python/shared/fixtures/init.py` to call `dump_test_db.py`; wire the test bootstrap (`local_start` / `kube_start`) to use `loaddata_sorted` instead of plain `loaddata`.
* Update `tests/python/README.md` backup and restore snippets accordingly.
* Re-dump `tests/python/shared/assets/cvat_db/data.json` with the new layout (alphabetical, idempotent).

## Test plan

- [ ] `python tests/python/shared/utils/dump_test_db.py --help` shows new options.
- [ ] `pytest tests/python --start-services --rebuild` brings up the test stack and loads `data.json` via `loaddata_sorted` without integrity errors.
- [ ] On a freshly-restored test DB, `python tests/python/shared/utils/dump_test_db.py` produces zero diff vs the committed `data.json`.
- [ ] `pytest tests/python --dumpdb` writes a `data.json` that round-trips (re-running produces no further diff).
- [ ] `cat tests/python/shared/assets/cvat_db/data.json | docker exec -i test_cvat_server_1 sh -c "./manage.py flush --no-input && ./manage.py loaddata_sorted"` reports `Installed 1786 object(s) from 1 fixture(s)`.


zhiltsov-max:
@codex review

nmanovic:
@zhiltsov-max , can we try to explore custom serializer? Basically, we can try to redefine different methods of https://docs.djangoproject.com/en/5.2/topics/serialization/#custom-serialization-formats. One way to do that is to inherit https://github.com/django/django/blob/stable/5.2.x/django/core/serializers/json.py and modify order and values of serialized values. It looks like a recommended way to modify an output format of dumpdata command.

nmanovic:
@zhiltsov-max , another idea is to use jsonl format. It will be easy to sort.

zhiltsov-max:
@nmanovic, I think the same sorting for fields and filtering for volatile fields can be implemented via custom serializers. It's easy to sort the output, but the real issue is on the import (`loaddata`) side though - it requires the referenced to models to be loaded before the referencing models. Otherwise, there will be errors about missing objects. It can potentially be solved by using a custom loader that sorts the data, I just thought that we don't want to maintain such a solution. I'll check, maybe it can be a simple command or script.

Django docs do say about a similar situation - https://docs.djangoproject.com/en/5.2/topics/serialization/#natural-keys-and-forward-references, but the solution requires each FK relationship to be nullable.

Using jsonl doesn't look a very good idea to me. It will be hell to review and merge manually. Sorting isn't really a difficult task here.

zhiltsov-max:
Ok, I added a customized version of loaddata that sorts the inputs before importing. The script is quite simple, so I think we can keep it.

cvat/apps/engine/management/commands/loaddata_sorted.py
```diff
@@ -0,0 +1,79 @@
...
+
+        ordered_models = sort_dependencies(list(models_by_app.items()), allow_cycles=True)
+
+        reordered = [
+            record
+            for model in ordered_models
+            for record in models_by_name[f"{model._meta.app_label}.{model._meta.model_name}"]
+        ]
+
+        tmp_path = Path(tempfile.gettempdir()) / f"loaddata_sorted_{input_filename}.json"
```
SpecLad: You didn't delete this temporary directory (also, it seems wasteful to create a directory for just one file).

zhiltsov-max: Agree, updated.

tests/python/shared/utils/dump_test_db.py
```diff
@@ -0,0 +1,285 @@
...
+)
+
+# Fields whose value changes on every dump even when no test-relevant state
+# has changed (auto_now timestamps, last-login markers, etc.). When
+# preservation is enabled the script copies these values from the existing
+# data.json so that re-dumping an unchanged DB produces no diff.
+VOLATILE_FIELDS: dict[str, frozenset[str]] = {
+    "auth.user": frozenset({"last_login"}),
+    "engine.profile": frozenset({"last_activity_date"}),
+    "engine.project": frozenset({"updated_date", "assignee_updated_date"}),
```
SpecLad: Can we just put a blanket ignore on `updated_date` and `*_updated_date` for all models?

tests/python/shared/utils/dump_test_db.py
```diff
@@ -0,0 +1,285 @@
...
+    source_group = parser.add_mutually_exclusive_group()
+    source_group.add_argument(
+        "--from-file",
+        type=Path,
+        help="Read pre-existing dumpdata JSON from this file instead of running docker exec.",
+    )
+    source_group.add_argument(
+        "--from-stdin",
+        action="store_true",
+        help="Read pre-existing dumpdata JSON from stdin.",
```
SpecLad: Seems like for consistency this should be replaced with `--from-file -`.

tests/python/README.md
```diff
@@ -83,15 +83,17 @@ To backup DB and data volume, please use commands below.
...
 
-> Note: if you won't be use --indent options or will be use with other value
-> it potentially will lead to problems with merging of this file with other branch.
+`dump_test_db.py` runs `manage.py dumpdata` inside the test container and
+post-processes the output for stable diffs: records are sorted by
+`(model, pk)`, field keys are sorted alphabetically, and volatile
+timestamp fields (`last_login`, `*_updated_date`, ...) are preserved from
+the existing `data.json` so a re-dump after a no-op test run produces zero
+diff. Pass `--refresh-volatile` to take new timestamps from the dump
+instead, or `--help` for other options.
```
SpecLad: IMO, this comment is redundant with what is already documented in the script itself.

tests/python/shared/utils/dump_test_db.py
zhiltsov-max: I think we could, but it's quite a weak heuristic.

tests/python/shared/utils/dump_test_db.py
zhiltsov-max: Updated

tests/python/README.md
zhiltsov-max: Updated

tests/python/shared/utils/dump_test_db.py
zhiltsov-max: Updated

SpecLad (approved):

tests/python/shared/utils/dump_test_db.py
```diff
@@ -134,6 +134,8 @@
 # round-trip is idempotent.
 ZERO_MS_PATTERN = r"^(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2})\.000(Z|[+-]\d{2}:\d{2})$"
 
+_DEFAULT = object()
```
SpecLad, line 137: FYI, [sentinels](https://peps.python.org/pep-0661/) have been approved for Python 3.15, and you can already use them via typing-extensions.

PR #10611, opened 2026-05-14
Add server opt-in for client-side preview placeholders

zhiltsov-max:
## Summary

Server side
- Preview endpoints (`/api/{projects,tasks,jobs}/{id}/preview`) accept `Prefer: handling=empty` (RFC 7240). When the entity has no media-derived preview, the server responds `204 No Content` with `Preference-Applied: handling=empty` and `Vary: Prefer`. Without the header, the response is `200` + default PNG.
- "No media-derived preview" covers point-cloud tasks and audio tasks without an embedded cover image.

Client side (cvat-core)
- `serverProxy.*.getPreview` sends `Prefer: handling=empty` automatically and yields `PREVIEW_DEFAULT` on 204; `resolvePreviewResponse(response, mediaType)` returns a placeholder URL (or `''`) for that case and a data URL for real previews.
- Placeholder URL map is owned by the consumer through the existing config mechanism: `cvat.config.previewPlaceholders: Record<MediaType, string>`. cvat-core has no asset paths baked in.
- New `mediaType` accessor on `Task` / `Job`. `SerializedTask.media_type | mode | dimension` are typed as `?: T`.

Client side (cvat-ui)
- Registers placeholder URLs at init: `point_cloud` → `/assets/point_cloud_preview.png`, `audio` → `/assets/audio_preview.png`.
- `preview.tsx` consumes the URL returned by cvat-core.

The split lets the UI decide how to present placeholders and distinguish them from real previews, and avoids transferring the default PNG per entity for media types without a derived preview.

## Test plan

- [x] Parametrized REST test: `pytest tests/python/rest_api/test_tasks.py::TestPreviewPreferHeader`
- [ ] Manual: `curl -i -H 'Prefer: handling=empty' /api/tasks/<3D-task>/preview` → `204`; without the header → `200` + PNG.
- [ ] Manual: same for an audio task without a cover image.
- [ ] Manual UI: 3D and audio (no cover) task/job tiles render bundled placeholders; 2D and audio-with-cover tiles render real previews.


cvat-core/src/server-response-types.ts
```diff
@@ -111,6 +111,7 @@ export interface SerializedTask {
     data_original_chunk_type: ChunkType;
     data_cloud_storage_id: number | null;
     dimension: DimensionType;
+    media_type: MediaType | '';
```
bsekachev: Do we need a default value for media_type? Are there any cases when Job/Task do not return media type?

cvat-core/src/frames.ts
```diff
@@ -880,6 +891,27 @@ export function decodePreview(preview: Blob): Promise<string> {
...
+): Promise<string> {
+    if (response === null) {
+        return '';
+    }
+    if (response === PREVIEW_DEFAULT) {
+        const kind = placeholderKindFor(mediaType);
+        return kind ? `${PREVIEW_PLACEHOLDER_PREFIX}${kind}` : '';
+    }
+    return decodePreview(response);
+}
```
bsekachev: For better readability, would propose

if (response === PREVIEW_DEFAULT and mediaType) {
    ...
} else if (response) {
    return decodePreview();
}

return '';

cvat-core/src/session.ts
```diff
@@ -696,6 +699,10 @@ export class Job extends Session {
         return this.#data.dimension;
     }
 
+    public get mediaType(): MediaType | '' {
+        return this.#data.media_type;
```
bsekachev: You declare `MediaType | ''` but return `MediaType` or `undefined`. 

Anyway, I do not see any reason to have empty string as a default value for media type in this implementation 
It may be normal value or undefined.

cvat-core/src/session.ts
```diff
@@ -1096,6 +1106,9 @@ export class Task extends Session {
                 dimension: {
                     get: () => data.dimension,
                 },
+                mediaType: {
+                    get: () => data.media_type,
```
bsekachev, line 1138: The same comment

cvat-ui/src/components/common/preview.tsx
```diff
@@ -9,10 +9,16 @@ import { PictureOutlined } from '@ant-design/icons';
...
-import { Job, Task, Project } from 'cvat-core-wrapper';
+import {
+    Job, Task, Project, isPreviewPlaceholder, previewPlaceholderKind,
+} from 'cvat-core-wrapper';
 import MLModel from 'cvat-core/src/ml-model';
 import { previewQueue, getRequestId } from 'utils/preview-queue';
 
+const PLACEHOLDER_ASSETS: Record<string, string> = {
+    point_cloud: '/assets/point_cloud_preview.png',
+};
```
bsekachev: The small idea to simplify all code is to pass placeholder urls to `cvat-core`, by using its configuration mechanism: cvat_enterprise/cvat/cvat-core/src/config.ts

You need to add config setter in cvat_enterprise/cvat/cvat-core/src/api.ts

From cvat-ui:

```
core.config.previewPlaceholders = {
    {mediaType}: 'url',
},
```

Thus cvat-core will return urls for supported media types.

cvat-core/src/server-response-types.ts
zhiltsov-max: Tasks without data have empty media type, as well as dimension, and mode. Chunk types are not included in the server responses. All of this is only for tasks, jobs are guaranteed to have values, as they are created after data uploading.

bsekachev: Below I wrote that another default value `''` seems unnecessary. 
We already have `undefined`.

zhiltsov-max: Just to confirm, so your suggestion is to convert `''` to `undefined` after receiving the response?

bsekachev: Yes, I believe semantic of these values in current implementation is the same (literally - media type not defined).
Instead of `mediaType: MediaType | '' | undefined` we may use just `mediaType?: MediaType`

zhiltsov-max:
@codex review

cvat-core/src/server-response-types.ts
zhiltsov-max: Ok, updated to handle just after at the server response. Dimension and mode aligned as well.

cvat-core/src/frames.ts
zhiltsov-max: Updated.

cvat-ui/src/components/common/preview.tsx
zhiltsov-max: Ok, done.

bsekachev (approved):

PR #10612, opened 2026-05-15
Add Dependabot updates for Python requirements

azhavoro:
## Summary

- Add Dependabot version updates for GitHub Actions, root Dockerfiles, and Python requirement source files.
- Configure Python Dependabot updates to ignore generated `*.txt` lock files and `all.in`.
- Add a workflow that regenerates Python requirement lock files on Dependabot PRs after `*.in` files change.
- Set the requirements Python version to 3.10 so Dependabot resolves the same Python dependency graph as the project tooling.
- Update Shapely to a version that can be resolved in the Dependabot environment.

## Notes

Python requirements are managed by `pip-compile-multi`, so Dependabot should only update source `*.in` files. The generated `*.txt` files are refreshed by the workflow using `regenerate.sh --no-upgrade`.

## Testing

- Validated `.github/dependabot.yml` as YAML.
- Tested the regeneration flow in the fork with Dependabot-created PRs.


cvat/requirements/.python-version
```diff
@@ -0,0 +1 @@
+3.10
```
nmanovic: 3.12?

.github/workflows/update-python-requirements.yml
```diff
@@ -0,0 +1,66 @@
...
+          if git diff --quiet -- cvat/requirements/*.txt utils/dataset_manifest/requirements.txt; then
+            echo "No generated requirements changes"
+            exit 0
+          fi
+
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git add cvat/requirements/*.txt utils/dataset_manifest/requirements.txt
+          git commit -m "Regenerate Python requirements"
+          git push
```
SpecLad: As I recall, using the default GitHub token to perform actions will not trigger workflows. Isn't this going to be a problem here? It means that CI won't run for the updated version of the pull request.

.github/workflows/update-python-requirements.yml
```diff
@@ -0,0 +1,66 @@
...
+      - "utils/dataset_manifest/requirements.in"
+
+permissions:
+  contents: write
+
+jobs:
+  update:
+    # Dependabot PRs can be synchronized by maintainers, so github.actor is not
+    # always dependabot[bot]. Limit this write-enabled job to Dependabot branches.
+    if: startsWith(github.event.pull_request.head.ref, 'dependabot/')
```
SpecLad: We should probably make sure the request is coming from our own repository, not a fork. I'm fairly certain that executing pip-compile on untrusted requirements leads to arbitrary code execution, and I'm not comfortable doing that in a workflow with write permissions.

.github/workflows/update-python-requirements.yml
```diff
@@ -0,0 +1,92 @@
...
+          python-version: "3.12"
+
+      - name: Install system dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install --no-install-recommends -y \
+            cargo \
+            g++ \
+            gcc \
+            libgeos-dev \
```
SpecLad: I think libgeos is no longer required, I've updated Shapely to a version that has binary wheels.

SpecLad (approved):
LGTM, just one nitpick.

PR #10613, opened 2026-05-15
Enhanced layering management

bsekachev:
### Motivation and context

This PR improves z-order/layer workflows in the annotation UI.

Changes include:
- Added layer stack support in the objects sidebar when sorting by z-order.
- Added drag-and-drop actions to move objects between layers, create new layers, and merge/move existing layers.
- Added layer compaction to re-enumerate z-order layers.
- Added an “Open layer stack” canvas button with the current layer number.
- Added z-order filtering for annotations.
- Added canvas object text support for displaying the object layer.
- Added a `Layer` option to workspace object text settings.

### How has this been tested?

Ran:
- `yarn --cwd cvat_enterprise/cvat/cvat-core type-check`
- `yarn --cwd cvat_enterprise/cvat/cvat-ui lint --quiet src/components/annotation-page/top-bar/filters-modal.tsx`
- `yarn --cwd cvat_enterprise/cvat/cvat-ui lint --quiet src/components/header/settings-modal/workspace-settings.tsx`
- `yarn --cwd cvat_enterprise/cvat/cvat-ui lint --quiet src/components/annotation-page/canvas/views/canvas2d/canvas-wrapper.tsx`
- `git diff --check`

Also ran `yarn --cwd cvat_enterprise/cvat/cvat-canvas build`; webpack completed, but TypeScript reported existing unrelated `Uint8ClampedArray` generic errors in `cvat-canvas/src/typescript/shared.ts`.

### Checklist

### License


cvat-ui/src/containers/annotation-page/standard-workspace/objects-side-bar/objects-list.tsx
```diff
@@ -75,6 +75,8 @@ interface DispatchToProps {
     changeHideEditedState(value: boolean): void;
 }
 
+type LayerDragMode = 'move' | 'merge';
```
bsekachev: This extra type declaration is unnecessary. We only use it in one place.

cvat-ui/src/containers/annotation-page/standard-workspace/objects-side-bar/objects-list.tsx
```diff
@@ -342,6 +344,20 @@ function sortAndMap(objectStates: ObjectState[], ordering: StatesOrdering): numb
 
 type Props = StateToProps & DispatchToProps;
 
+function isLayerState(state: ObjectState): boolean {
+    return state.objectType !== ObjectType.TAG;
```
bsekachev: It would be better to use "white list" logic here

cvat-ui/src/containers/annotation-page/standard-workspace/objects-side-bar/objects-list.tsx
```diff
@@ -425,6 +441,114 @@ class ObjectsListContainer extends React.PureComponent<Props, State> {
         changeShowGroundTruth(!showGroundTruth);
     };
 
+    private moveObjectToLayer = (clientID: number, targetZOrder: number): void => {
+        const { updateAnnotations } = this.props;
+        const { filteredStates } = this.state;
+        const objectState = filteredStates.find((state: ObjectState): boolean => state.clientID === clientID);
+
+        if (!objectState || objectState.objectType === ObjectType.TAG || objectState.zOrder === targetZOrder) {
```
bsekachev: instead objectState.objectType === ObjectType.TAG we probably shall use isLayerState

cvat-ui/src/containers/annotation-page/standard-workspace/objects-side-bar/objects-list.tsx
```diff
@@ -425,6 +441,114 @@ class ObjectsListContainer extends React.PureComponent<Props, State> {
...
+    };
+
+    private moveObjectToNewLayer = (clientID: number, targetZOrder: number): void => {
+        const { updateAnnotations } = this.props;
+        const { filteredStates } = this.state;
+        const objectState = filteredStates.find((state: ObjectState): boolean => state.clientID === clientID);
+
+        if (!objectState || !isLayerState(objectState) || objectState.zOrder === targetZOrder) {
+            return;
+        }
```
bsekachev: This code seems to be repeated in several methods

cvat-ui/src/containers/annotation-page/standard-workspace/objects-side-bar/objects-list.tsx
```diff
@@ -425,6 +441,114 @@ class ObjectsListContainer extends React.PureComponent<Props, State> {
...
+
+        const sourceZOrder = objectState.zOrder;
+        const targetOccupied = filteredStates.some((state: ObjectState): boolean => (
```
bsekachev: This is actually not guaranteed to discover that layer is occupied. UI knows only about filtered states. It does not know about the all states from cvat-core collection. Thus, this logic probably have to implemented somewhere in cvat-core. 

But updateAnnotationsAsync saves objects one by one. And this is problem 
For example we move layer 3 to space between layer 1 and 2. 
Thus layer 3 must became layer 2. Layer 2 must be shifted. But if layer 3 have 2 objects, the first object receives layer 3. The second object does not know anything about the first object. It also inserts to layer 2, and shift both 2 and 3. Thus we get 1,2,3,4. 

Perhaps we need to implement function in annotationCollection
collection.moveOnLayer(frame, { exact?: number, before?: number, after?: number }, states)
only one of exact, before, after applicable. And one must be presented. 
The implementation must preserve batched undo/redo features. 
And shift objects from upper layers (if necessary). 

Because of the same reasons, compactLayers also must be implemented on collection.

So, the path is:
Session.moveObjectsToLayer (just wrapper) -> collection
Session.compactFrameLayers() -> collection

compactFrameLayers also must preserve batched undo/redo

for optimization purpose, both should return updated ObjectState instances for all touched objects.

PR #10616, opened 2026-05-17
fix: handle degenerate bboxes in QC skeleton/keypoint matching

### LINKED ISSUE #9957: Quality control: can't match skeletons or point groups in corner cases

### Steps to Reproduce

1. Create a task with 1 skeleton label (with 1 point) a GT job
2. Create 1 annotation that you'd expect to match in the GT job and a regular job
3. Run quality check, find that the skeletons were not matched

The reason is that the skeleton boxes are checked for intersection first. Only if there's an intersection, the points are compared between each other. This works in "normal" cases, where there are 5+ visible skeleton points and they cover some area, but in some corner cases this check produces incorrect results.

Related cases:
- a pair of skeletons / points that are quasi-parallel lines without bbox intersection
- a pair of skeletons / points, with one or two of them having just 1 visible point

Code references:
https://github.com/cvat-ai/cvat/blob/650f2b204ce5fd52630cfdb44dafa3936bfc2de3/cvat/apps/quality_control/quality_reports.py#L1078-L1081

Known workarounds:
Append a grouped bounding box to each skeleton. These boxes can be used by the existing implementation. It solves the problem, but it's not always convenient to do and increases the amount of annotation work. In some annotation scenarios there is no simple way to add more annotation types in the specification.

### Expected Behavior

The skeleton and points should match when they're close to each other.

### Possible Solution

It's possible to disable this check, but doing so can reduce performance of the algorithm significantly and can result in some false-positive matches because of the OKS scaling to the common bbox.

### Context

_No response_

### Environment

```Markdown

```

Abubakarnaeem20:
## Summary

Closes #9957 — Quality control cannot match skeletons with degenerate bounding boxes (single-point or collinear skeletons).

### Root cause

Single-point skeletons produce a bbox of `(x, y, 0, 0)`. `bbox_iou()` returns 0 for any zero-area box (0÷0), so the early-exit guard blocks even *identical* annotations from matching.

### Fix (per maintainer suggestion — option 2)

When **both** bboxes have zero area, skip the IoU guard and scale OKS by the **image area** (`scale = img_h × img_w`) instead of the bbox area. This is the same approach already used for point annotations in `match_points._distance`.

Why image size works:
- Provides a meaningful reference scale regardless of object size
- Correctly handles slightly-offset points (not just exact matches)
- Mirrors existing, battle-tested logic in the codebase

Why the initial 1×1 minimum-bbox approach was wrong:
- Only worked for d=0 (exact matches); any slight offset → OKS ≈ 0 due to tiny scale

**Changes:**
- `KeypointsMatcher`: add `img_size: tuple[int, int] | None` field; when both bbox areas are 0 use `scale = img_h * img_w`
- `match_skeletons`: read image size from `item_a` and pass it to `KeypointsMatcher`, mirroring `match_points._distance`

**What did not change:** Normal skeletons with non-overlapping, non-degenerate bboxes are still rejected by the existing IoU guard — no performance regression for the common case.

## Test plan

New tests in `cvat/apps/quality_control/tests/test_keypoints_matcher.py`:

- [ ] `test_single_point_identical_returns_one` — OKS = 1.0 for identical single-point skeletons
- [ ] `test_single_point_slightly_off_returns_high_oks` — OKS > 0.9 for points 5 px apart on 640×480
- [ ] `test_single_point_far_apart_returns_low_oks` — OKS < 0.5 for points far apart
- [ ] `test_collinear_two_point_skeleton_identical_returns_one` — zero-height bbox skeletons match
- [ ] `test_degenerate_without_image_size_returns_zero` — safe fallback when img_size is None
- [ ] `test_normal_non_overlapping_skeletons_return_zero` — regression: non-overlapping normal skeletons still blocked
- [ ] `test_normal_overlapping_skeletons_return_nonzero` — regression: overlapping normal skeletons still compute OKS

### DIFF
--- a/cvat/apps/quality_control/tests/test_keypoints_matcher.py
+++ b/cvat/apps/quality_control/tests/test_keypoints_matcher.py
@@ -1,0 +1,58 @@
+# Copyright (C) CVAT.ai Corporation
+#
+# SPDX-License-Identifier: MIT
+
+import unittest
+
+import datumaro as dm
+
+from cvat.apps.quality_control.quality_reports import KeypointsMatcher
+
+
+class TestKeypointsMatcher(unittest.TestCase):
+    """Regression tests for issue #9957.
+
+    Single-point and collinear skeletons produce zero-area bboxes.
+    bbox_iou() returns 0 for zero-area boxes, which previously caused
+    the early-exit guard to block even identical annotations from matching.
+    """
+
+    def _matcher(self, pt_a, bbox_a, pt_b, bbox_b):
+        return KeypointsMatcher(
+            sigma=0.1,
+            instance_map={id(pt_a): [None, bbox_a], id(pt_b): [None, bbox_b]},
+        )
+
+    def test_single_point_identical_returns_one(self):
+        """Identical single-point skeletons must match with OKS=1.0."""
+        pt_a = dm.Points([10, 20], visibility=[dm.Points.Visibility.visible])
+        pt_b = dm.Points([10, 20], visibility=[dm.Points.Visibility.visible])
+        matcher = self._matcher(pt_a, (10, 20, 0, 0), pt_b, (10, 20, 0, 0))
+        self.assertAlmostEqual(matcher.distance(pt_a, pt_b), 1.0)
+
+    def test_collinear_two_point_skeleton_identical_returns_one(self):
+        """Identical horizontal-line skeletons (zero-height bbox) must match."""
+        pt_a = dm.Points([0, 0, 100, 0], visibility=[dm.Points.Visibility.visible] * 2)
+        pt_b = dm.Points([0, 0, 100, 0], visibility=[dm.Points.Visibility.visible] * 2)
+        matcher = self._matcher(pt_a, (0, 0, 100, 0), pt_b, (0, 0, 100, 0))
+        self.assertAlmostEqual(matcher.distance(pt_a, pt_b), 1.0)
+
+    def test_normal_non_overlapping_skeletons_return_zero(self):
+        """Non-overlapping normal (non-degenerate) skeletons must not match."""
+        pt_a = dm.Points([10, 10, 50, 50], visibility=[dm.Points.Visibility.visible] * 2)
+        pt_b = dm.Points([200, 200, 250, 250], visibility=[dm.Points.Visibility.visible] * 2)
+        matcher = self._matcher(pt_a, (10, 10, 40, 40), pt_b, (200, 200, 50, 50))
+        self.assertEqual(matcher.distance(pt_a, pt_b), 0.0)
+
+    def test_normal_overlapping_skeletons_return_nonzero(self):
+        """Overlapping normal skeletons must still compute OKS correctly."""
+        pt_a = dm.Points([10, 10, 50, 50], visibility=[dm.Points.Visibility.visible] * 2)
+        pt_b = dm.Points([12, 12, 52, 52], visibility=[dm.Points.Visibility.visible] * 2)
+        matcher = self._matcher(pt_a, (10, 10, 40, 40), pt_b, (12, 12, 40, 40))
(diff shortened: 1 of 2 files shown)


cvat/apps/quality_control/tests/test_keypoints_matcher.py
```diff
@@ -0,0 +1,58 @@
...
+            sigma=0.1,
+            instance_map={id(pt_a): [None, bbox_a], id(pt_b): [None, bbox_b]},
+        )
+
+    def test_single_point_identical_returns_one(self):
+        """Identical single-point skeletons must match with OKS=1.0."""
+        pt_a = dm.Points([10, 20], visibility=[dm.Points.Visibility.visible])
+        pt_b = dm.Points([10, 20], visibility=[dm.Points.Visibility.visible])
+        matcher = self._matcher(pt_a, (10, 20, 0, 0), pt_b, (10, 20, 0, 0))
+        self.assertAlmostEqual(matcher.distance(pt_a, pt_b), 1.0)
```
zhiltsov-max: The solution certainly works for identical points, but having such points in real annotations is unlikely. If the points are slightly off, this solution breaks. It's probably acceptable as a workaround, but it's not really helpful in practice.

I can suggest doing one of the following.
1. Boxes in OKS serve as object size guides. 
1.a. We can potentially try to pick up such a guide from other skeletons in the task, but I don't really like this solution, as it yields unpredictable results. 
1.b. Another possible option is to just pick a bigger minimal bbox, e.g. 50x50. The problem is that we can't know what is the expected object size in the task is.
2. Change the approach. Instead of using boxes, allow using the image size. This is already implemented for `points` annotations (check `match_points - _distance - (else branch)`. This option works quite good in my experience, but it has the downsize of requiring uniform images. I'd recommend choosing this option, because it's practical and quite simple to implement.
3. Implement a more complex approach, like PDJ, PCKh-likes with  configurable pivot ([read more](https://arxiv.org/pdf/1812.08008)). This is likely to work and be practical, but I don't recommend going this way without trying simpler solutions first.

zhiltsov-max (commented):
Hi, thank you for the PR! Please check the comment.

Abubakarnaeem20:
Thanks for the detailed feedback, @zhiltsov-max!

You're right — the 1x1 minimum-bbox workaround only holds for exact matches (d=0) and breaks as soon as annotations are slightly off, making it impractical for real-world use.

I've updated the PR to implement **option 2** as you recommended. The changes mirror the logic already in :

- Added an  field to 
- When **both** bboxes have zero area, the IoU guard is skipped and OKS is scaled by the image area () instead of the bbox area
-  now reads the image size from  and passes it to , exactly as  does
- Added a safe fallback: if  is , return  rather than NaN

Updated tests cover:
- Identical single-point skeletons → OKS = 1.0
- Slightly-offset points (5 px on a 640x480 image) → OKS > 0.9
- Far-apart points → OKS < 0.5
- Collinear (zero-height bbox) identical skeletons → OKS = 1.0
- Safe fallback when  → returns 0
- Regression guards for normal non-degenerate bboxes (unchanged behaviour)

Please let me know if anything needs adjusting!

zhiltsov-max:
Thank you for the updated version. It partially fixes the error, but I don't think it does it correctly now. This version silently changes how the OKS sigma values are interpreted (object bbox to image size), which can lead to matches far beyond the expected radius. For simple point comparison it's open for interpretation, so the approach can be considered working. It's not the same for skeletons. I think the proper solution should work this way:

1. Allow using `point_size_base` for skeleton matching - this will provide an option to compare skeletons without explicit bbox dependency, when it's suitable for the dataset.
2. If `point_size_base == GROUP_BBOX_SIZE`, matching should result in an error as before (because skeletons must have a bbox to be compared correctly).

It's possible to extend these 2 options with more advanced variants, e.g. compute scale based on the skeleton edges (for datasets with non-uniform images), but I'd start just with the options above.

PR #10625, opened 2026-05-19
Validate label attribute name conflicts before updating

bsekachev:
### Motivation and context

When updating label attributes, renaming attributes to names already used by the same label can hit the database unique constraint on `(label, name)`. This includes swapping names between attributes and renaming one attribute to another existing attribute name.

This PR moves that failure to serializer-level validation and returns a normal validation error before attribute updates are applied. It also preserves the valid case where an attribute is deleted and another attribute is renamed to the deleted attribute’s name in the same request.

### How has this been tested?

- Added REST API coverage for attribute name swaps.
- Added REST API coverage for renaming to an existing attribute name.
- Added REST API coverage for renaming to a deleted attribute name in the same request.
- Ran:
  - `python -m py_compile cvat/apps/engine/serializers.py cvat/apps/engine/tests/test_rest_api.py`
  - `git diff --check -- cvat/apps/engine/serializers.py cvat/apps/engine/tests/test_rest_api.py changelog.d/20260519_103538_sekachev.bs.md`

### Checklist

### License

### DIFF
--- a/cvat/apps/engine/serializers.py
+++ b/cvat/apps/engine/serializers.py
@@ -542,9 +542,7 @@ class LabelSerializer(SublabelSerializer):
             return
 
         current_names_by_id = dict(
-            db_label.attributespec_set.filter(id__in=request_names_by_id).values_list(
-                "id", "name"
-            )
+            db_label.attributespec_set.filter(id__in=request_names_by_id).values_list("id", "name")
         )
 
         # A normal rename to a new name is allowed. The unsafe case is swapping
--- a/cvat/apps/engine/serializers.py
+++ b/cvat/apps/engine/serializers.py
@@ -529,39 +529,43 @@ class LabelSerializer(SublabelSerializer):
                 encountered_names.add(attr_name)
 
     @staticmethod
-    def check_no_attr_names_swap(db_label: models.Label, attrs: list[dict[str, Any]]) -> None:
-        request_names_by_id = {
+    def check_attribute_names_available(
+        db_attributes: dict[int, str], attrs: list[dict[str, Any]]
+    ) -> None:
+        requested_attribute_names = {
             attr["id"]: attr["name"]
             for attr in attrs
-            if not attr.get("deleted")
-            and attr.get("id") is not None
-            and attr.get("name") is not None
+            if attr.get("id") is not None and attr.get("name") is not None
         }
 
-        if len(request_names_by_id) < 2:
+        if not requested_attribute_names:
             return
 
-        current_names_by_id = dict(
-            db_label.attributespec_set.filter(id__in=request_names_by_id).values_list("id", "name")
-        )
-
-        # A normal rename to a new name is allowed. The unsafe case is swapping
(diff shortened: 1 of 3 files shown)


bsekachev:
@codex review

bsekachev:
@codex review

bsekachev:
@zhiltsov-max covered couple of topics we recently discussed

cvat/apps/engine/serializers.py
zhiltsov-max, line 753: Let's say we use the PATCH method for tasks to update labels. 

1. If there is an existing label attribute, and the request contains a new one with the same name, the existing label will be updated, even if `id` wasn't specified:

```json
{
  "labels": [
    {
      "name": "test",
      "id": 93,
      "attributes": [
        {
          "name": "a",
          "input_type": "checkbox"
        }
      ]
    }
  ]
}
```
-> 
```
[
  "Attribute field \"input_type\" cannot be changed"
]
```

2. Now this simply looks confusing, but if (1) is fixed somehow - `check_attribute_names_available()` checks for conflicts on already existing attributes. If there is a rename on existing attribute and a new attribute with the previous name is created, the result depends on the declaration order in the request:

```json
{
  "labels": [
    {
      "name": "test",
      "id": 93,
      "color": "#fb117d",
      "type": "any",
      "attributes": [
        {
          "name": "a",
          "input_type": "checkbox"
        },
        {
          "id": 19,
          "name": "b"
        }
      ]
    }
  ]
}
```

Both issues can't be reproduced in the UI without extra efforts, but the first one looks like a bug. The second one is questionable, you can always say that this is not supported, but it can potentially be solved by updating the existing attributes first or by pre-computing all the results before saving. `bulk_create` ([example](https://github.com/cvat-ai/cvat/blob/8b95ed69b0c91af848381a30126007068228f571/cvat/apps/engine/serializers.py#L2079)) can also help, but with `update_conflicts` it can lead to deadlocks from concurrent runs.

cvat/apps/engine/serializers.py
```diff
@@ -528,6 +528,45 @@ def check_attribute_names_unique(attrs):
...
+            for attr in attrs
+            if attr.get("id") is not None and attr.get("name") is not None
+        }
+
+        if not requested_attribute_names:
+            return
+
+        current_attribute_ids = {name: attr_id for attr_id, name in db_attributes.items()}
+        swapped_attr_names = set()
+        busy_attr_names = set()
```
zhiltsov-max: `occupied` would be a better word, busy is more about having some ongoing activity.

cvat/apps/engine/serializers.py
bsekachev: I believe this behavior comes from the existing get_or_create(label=db_label, name=attr["name"], defaults=attr) logic for attributes without an id. I didn’t change it because that would change the endpoint semantics. This patch is limited to adding serializer-level validation for updates of existing attributes, so cases that previously hit the (label, name) DB constraint return a validation error instead.

zhiltsov-max (approved):

PR #10638, opened 2026-05-20
Use None instead of the none_type alias in generated SDK type annotations

### LINKED ISSUE #10531: Replace `none_type = type(None)` indirection usage for better static analyzer support

The current use of `none_type = type(None)` for type annotations in the OpenAPI generated code causes significant issues for static analyzers like Pylance/Pyright and Mypy. While technically valid under PEP 484, it introduces indirection that breaks type narrowing.

Replace, for example:
```python
# mode_utils.py
`none_type = type(None)`

# data_meta_read.py
class IDataMetaRead(IModelData):
    # ...
    frames: typing.Union[list[FrameMeta], none_type]
```

With standard PEP 484 notation:
```python
class IDataMetaRead(IModelData):
    # ...
    frames: list[FrameMeta] | None # (or `Optional[list[FrameMeta]]` for broader compatibility)
```

Unfortunately, I haven't found a magic comment (e.g. `# pyright: ignore`, `# type: ignore`, etc.) that disables the type parsing failure. The inability for `none_type = type(None)` to be correctly resolved to the `None` singleton means type `Unknown` gets introduced and flows throughout usage of the cvat_sdk package. And introducing additional runtime code just to make static anayzers happy shouldn't be necessary, and it mostly doesn't work anyway. That `Unknown` is rather persistent once it's introduced.

### Additional context

I've opened a [related issue](https://github.com/microsoft/pyright/issues/11407) in the Pyright repo about how this technically-correct indirection isn't handled correctly. Rather than wait for them to accept it as a bug and fix it, I propose the cvat openapi-generator templates are updated to more modern syntax that is more static analyzer friendly.

officialasishkumar:
### Motivation and context

Fixes #10531.

The Python SDK is generated with openapi-generator. For nullable fields and
return values, the templates emit type annotations that reference the
`none_type` alias, e.g. in `cvat_sdk/api_client/model/data_meta_read.py`:

```python
frames: typing.Union[list[FrameMeta], none_type]
```

`none_type` is a module-level alias defined as `none_type = type(None)` and
imported from `cvat_sdk.api_client.model_utils`. The indirection is valid at
runtime, but static type checkers cannot resolve a plain module-level variable
used in a type position. Pyright/Pylance report `Variable not allowed in type
expression (reportInvalidTypeForm)` and degrade every nullable field to
`Unknown` (e.g. `list[FrameMeta] | Unknown`), and that `Unknown` then
propagates through any code that consumes the SDK. This was deliberately left
untouched when the hand-written SDK code was modernized in #10064 ("kept the
old syntax in generator templates"); this PR addresses the remaining generated
annotations.

This PR changes the generator post-processing step (`make_type_annotation`) to
emit the `None` singleton instead of the `none_type` alias, so the generated
annotations become:

```python
frames: typing.Union[list[FrameMeta], None]
```

`None` is special-cased by the typing system and resolves correctly in all type
checkers, while remaining equivalent to the previous annotation. The change is
intentionally limited to the public type annotations: the `none_type` alias is
still used by the runtime type metadata (`openapi_types`, the `ApiClient`
parameter tables, etc.), so its import remains required and unchanged.

### How has this been tested?

- Regenerated the SDK with the updated post-processing
  (`./cvat-sdk/gen/generate.sh`) and confirmed:
  - no `none_type` remains in any type-annotation position (only in runtime
    type metadata and docstrings, as before);
  - all generated modules byte-compile and import successfully.
- Reproduced the original problem and verified the fix with Pyright on a
  faithful minimal example (alias imported from another module, mirroring the
  generated layout):
  - **before** — `error: Variable not allowed in type expression`; the field
    resolves to `list[int] | Unknown`;
  - **after** — no error; the field resolves to `list[int]`.
- Ran Pyright over the whole regenerated `cvat_sdk/api_client/model/` tree:
  no `none_type`-related diagnostics remain.

### Checklist

### License

### DIFF
--- a/cvat-sdk/gen/postprocess.py
+++ b/cvat-sdk/gen/postprocess.py
@@ -67,6 +67,14 @@ class Replacer:
         # single optional arg pattern
         type_repr = re.sub(r"^(.+, none_type)$", r"typing.Union[\1]", type_repr)
 
+        # Emit the None singleton instead of the none_type alias (none_type = type(None)).
+        # The indirection is valid at runtime, but static type checkers (Pyright/Pylance,
+        # mypy) can't resolve it to None, so every nullable field degrades to Unknown/Any
+        # in the public type annotations. None resolves correctly and is equivalent here.
+        # The none_type alias is still used by the runtime type metadata (openapi_types,
+        # etc.), so its import remains required. See cvat-ai/cvat#10531.
+        type_repr = re.sub(r"\bnone_type\b", "None", type_repr)
+
         return type_repr
 
     allowed_actions = {
--- a/changelog.d/20260520_120000_asishkumar_sdk_none_type_annotations.md
+++ b/changelog.d/20260520_120000_asishkumar_sdk_none_type_annotations.md
@@ -1,0 +1,6 @@
+### Fixed
+
+- \[SDK\] Generated type annotations for optional fields and return values no
+  longer use the `none_type` alias, which prevented static type checkers
+  (Pyright/Pylance) from resolving the annotated types
+  (<https://github.com/cvat-ai/cvat/pull/10638>)


cvat-sdk/gen/postprocess.py
```diff
@@ -67,6 +67,14 @@ def make_type_annotation(self, type_repr: str) -> str:
         # single optional arg pattern
         type_repr = re.sub(r"^(.+, none_type)$", r"typing.Union[\1]", type_repr)
 
+        # Emit the None singleton instead of the none_type alias (none_type = type(None)).
+        # The indirection is valid at runtime, but static type checkers (Pyright/Pylance,
+        # mypy) can't resolve it to None, so every nullable field degrades to Unknown/Any
+        # in the public type annotations. None resolves correctly and is equivalent here.
+        # The none_type alias is still used by the runtime type metadata (openapi_types,
+        # etc.), so its import remains required. See cvat-ai/cvat#10531.
+        type_repr = re.sub(r"\bnone_type\b", "None", type_repr)
```
SpecLad, line 76: It seems that the `none_type` annotation only ever occurs inside unions, so perhaps you can combine this substitution with the previous one?

PR #10647, opened 2026-05-21
Cache cloud storage clients + faster test polling

zhiltsov-max:
## Summary

- Added server-side caching for S3 CS clients, resulting in cold S3 CS client construction: ~150 ms → ~25 ms (cache miss) or ~0 ms (cache hit).
- Removed a multi-second IMDS credential timeout that hit on every cold `S3CloudStorage.__init__` running on non-EC2 hosts (~6-8 s per client × multiple clients per task creation).
- Together, this also improves performance of CS-heavy REST-API tests: `test_task_data.py` 24:23 → 13:00 (-47%), `test_resource_import_export.py` 50 s → 35 s (-31%), individual CS-backed tests up to -94% (21 s → 1.3 s).
- Improved performance of REST API tests via faster task-creation polling and skipping the throttle-prone anonymous version check on every SDK client setup. 

## Description

- Added caching for `db_storage_to_storage_instance()` results per process. 
  - Default size 2, can be configured via the `CVAT_CLOUD_STORAGE_INSTANCE_CACHE_SIZE` env. variable, 
  - Default cache size is 8 for server and chunks worker in docker-compose
  - Default TTL is 300 s, configurable via the `CVAT_CLOUD_STORAGE_INSTANCE_CACHE_TTL` env. variable
- Optimized creation performance of server-side CS clients by reusing `botocore.loaders.Loader`, so that the service-model JSON is parsed once, and drop the `object_pairs_hook=OrderedDict` from the loader to recover ~70% of per-instance RSS.
- Optimized and improved security of the credential uses in server-side CS clients disallowing the env/file/IMDS credential sources. CVAT CS clients should only use the provided credentials and nothing else. With an empty `CredentialResolver` and use of `Config(signature_version=UNSIGNED)` for anonymous CS, neither path stalls on `169.254.169.254` (used for EC2/EKS IMDS credential lookups).
- Improved potential connection pool reuse in server-side S3 CS clients by forcing `tcp_keepalive=True` on both the data-plane and status-check S3 clients, so the OS detects dead peer sockets while a client is parked in the cache. If a client hits the cache, it can start communicating faster.
- Cut `wait_until_task_is_created` poll interval from 1 s to 0.1 s. Matches the existing `resource_import_export._wait_request` cadence. The existing 1 s poll time was too pessimistic and was slowing the tests down significantly, especially after the CS changes.
- Disabled `check_server_version` in `make_sdk_client` in REST API tests. The default version check issues an anonymous `GET /api/server/about` before `client.login(...)`, which can trip the `AnonRateThrottle (100/min)` on bursts and trigger urllib3's `Retry-After` sleeps.

## Motivation

Profiling a single REST-API test on a non-EC2 host showed `Data.get_cloud_storage_instance()` -> `S3CloudStorage.__init__` spending ~94% of its wall-clock in `urllib3.create_connection` to IMDS at `169.254.169.254`, then timing out before falling through. Each cold boto3 client paid ~6-8 s, two clients per CS, two CS-build call sites per task creation - quickly accumulating to 12-20 s of pure timeout cost per task creation request. The cache + `UNSIGNED` + empty resolver each remove a separate slice of this cost; together they take the cold path from ~150 ms back to ~25 ms and the warm path to ~0.

This CS caching not only improves test performance, but it also speeds up CS-heavy endpoints, such as `/api/cloudstorages/{id}/content-v2` and chunk building. It can have some benefits for import and export workers as well, but it's not the main focus.

## Measured impact

| metric | develop | this branch |
| --- | --- | --- |
| `db_storage_to_storage_instance` (cs id 1, cs-only) | 156 ms | 0.0015 ms warm |
| `tests/python/rest_api/test_resource_import_export.py` | 50.27 s | 34.80 s (-31%) |
| `tests/python/rest_api/test_task_data.py` | 24:23 | 13:00 (-47%) |
| `tests/python/rest_api/test_tasks.py + test_jobs.py` | 4:42 | 3:46 (-19%) |
| `test_work_with_task_containing_non_stable_cloud_storage_files` | 22.04 s | 6.15 s (-72%) |
| `test_can_get_task_related_file_chunks[fxt_cloud_pcd_task_with_related_images-1]` | 21.33 s | 1.34 s (-94%) |

## Test plan

- [x] Existing `pytest tests/python/rest_api/test_tasks.py tests/python/rest_api/test_jobs.py` - 399 passed, 19 skipped, 1 xfailed on both develop and this branch (same 2 pre-existing TestTaskBackups failures unrelated to this work).
- [x] New `tests/python/rest_api/test_cloud_provider_caching.py` (13 tests) covering: shared botocore Loader installation, no-providers credential resolver, ambient `AWS_*` env vars are ignored, signed credentials survive the resolver swap, anonymous client uses UNSIGNED, session event emitter rejects register/unregister after freeze while per-client emitters stay mutable, build lock is exercised, cache returns same instance, field/TTL changes invalidate.
- [x] Verified optimizations are present in the rebuilt image (`cachetools`, `_SHARED_BOTOCORE_LOADER`, `tcp_keepalive`, `register_component("credential_provider", ...)` all grep-positive in the container).
- [ ] CI


zhiltsov-max:
@codex review

cvat/apps/engine/cloud_provider.py
```diff
@@ -588,6 +593,133 @@ def get_cloud_storage_instance(
...
+    # No code in botocore relies on OrderedDict-specific methods on loaded
+    # service-model data (verified by grep for move_to_end / popitem(last=...)).
+    if not os.path.isfile(full_path):
+        return
+    with open_method(full_path, "rb") as fp:
+        payload = fp.read().decode("utf-8")
+    return json.loads(payload)
+
+
+botocore.loaders.JSONFileLoader._load_file = _botocore_load_file_plaindict
```
SpecLad, line 615: Maybe you could submit a patch upstream instead? This is really hacky.

zhiltsov-max: I can try, but I think it can take a while to get into CVAT after this.

cvat/apps/engine/cloud_provider.py
```diff
@@ -588,6 +593,133 @@ def get_cloud_storage_instance(
...
+    session = boto3.Session(**kwargs)
+
+    # Inject the process-shared loader so every Session reuses the parsed
+    # service-2.json/resources-1.json/endpoint-ruleset instead of re-parsing.
+    session._session.register_component("data_loader", _SHARED_BOTOCORE_LOADER)
+
+    # Block env/file/IMDS credential discovery. CloudStorage credentials must
+    # come from the DB row; anonymous access uses Config(signature_version=UNSIGNED).
+    # Replacing the resolver with an empty one keeps `session.get_credentials()`
+    # from ever stalling on the metadata service.
```
SpecLad: We do not actually use `Config(signature_version=UNSIGNED)` anywhere in the code.

However, seems like if we _did_, then the `get_credentials` call would be skipped entirely, which would eliminate the IMDS lookup:

https://github.com/boto/botocore/blob/960f7bb17da8e3da5638667b5f49f6c97b2ad698/botocore/session.py#L959-L960

So let's just do that instead.

cvat/apps/engine/cloud_provider.py
```diff
@@ -588,6 +593,133 @@ def get_cloud_storage_instance(
...
+
+    register_first = register
+    register_last = register
+    unregister = register
+
+
+# boto3 Session objects (and the resources/clients they build) are NOT
+# thread-safe to construct; only low-level Clients are thread-safe to USE once
+# built. See:
+#   https://boto3.amazonaws.com/v1/documentation/api/latest/guide/session.html#multithreading-and-multiprocessing
```
SpecLad: I'm not sure where you're getting the information that sessions are not thread-safe to _construct_. In fact, the first reference here specifically shows an example of constructing a session in a thread, without any locks.

cvat/apps/engine/cloud_provider.py
SpecLad: Oh, never mind. I see the `if is_anonymous:` block now. But then it seems like this block here should be unnecessary?

zhiltsov-max: I think I decided to keep it to exclude any possibility of unexpected credential lookups, because this is how our CS access is designed. Basically, it's also covered by a unit test now as well, so maybe it can be removed.

tests/python/rest_api/utils.py
```diff
@@ -407,13 +407,19 @@ def build_exclude_paths_expr(ignore_fields: Iterator[str]) -> list[str]:
...
 
-def wait_until_task_is_created(api: apis.RequestsApi, rq_id: str) -> models.Request:
-    for _ in range(100):
+def wait_until_task_is_created(
+    api: apis.RequestsApi, rq_id: str, *, sleep_interval: float = 0.1, max_wait: float = 100.0
+) -> models.Request:
+    # Poll at 100 ms intervals (matches the RQ status pollers elsewhere in the
+    # test utils). At the previous 1-second interval each test paid up to
+    # ~1 s of polling overshoot waiting for an already-finished RQ job, which
+    # added up to many seconds across files like test_resource_import_export.py.
```
zhiltsov-max: Clean up this

cvat/apps/engine/cloud_provider.py
```diff
@@ -1270,20 +1422,54 @@ def values(self):
...
 
 
-def db_storage_to_storage_instance(db_storage: CloudStorage) -> AbstractCloudStorage:
+def _build_storage_instance(
+    cloud_provider: str,
+    resource: str,
+    credentials_type: str,
+    credentials_value: str,
+    specific_attributes_str: str,
+    updated_date_iso: str,  # noqa: ARG001  invalidates cache when storage row changes
```
SpecLad: Why do we need to do that? If the row changes, then wouldn't one of the other parameters change too? (and if none of them do, why can't we reuse the cache?)

cvat/apps/engine/cloud_provider.py
zhiltsov-max: I think the comment might be a little bit outdated, but the point is that now we have a `loader` object shared between `Session`s. And this lock protects the first initialization of this loader (it parses the API models and caches them, the actual heavy part of the session creation). The loader should be initialized during the `Client` object creation.

cvat/apps/engine/cloud_provider.py
zhiltsov-max: Updated date was included here for extra protection, but I think we can change it to a check that ensures there are session-affecting changes to the CS model in the unit tests.

cvat/apps/engine/cloud_provider.py
SpecLad: IIUC, this only reduces memory requirements for storage of the data files, which there's only a single copy of, so it's not really a high priority.

zhiltsov-max: Please check the commit description for this step - https://github.com/cvat-ai/cvat/pull/10647/changes/554a3992de4f137503ed1796edd44eb05c5b7b36. It's certainly optional, but it allows us to comfortably launch multiple processes without hitting the container memory limits.

cvat/apps/engine/cloud_provider.py
zhiltsov-max: Added more unit tests for this.

cvat/apps/engine/cloud_provider.py
zhiltsov-max: Updated the description and added more unit tests.

cvat/apps/engine/cloud_provider.py
zhiltsov-max: Removed updated date, added unit tests for fields in the model.

zhiltsov-max:
@codex review

cvat/apps/engine/cloud_provider.py
```diff
@@ -1270,20 +1416,55 @@ def values(self):
...
+
+def db_storage_to_storage_instance(db_storage: CloudStorage) -> AbstractCloudStorage:
+    # The kwargs passed here IS the cache key for the cached build. Pass every
+    # CloudStorage field that affects S3/Azure/GCS session construction; if
+    # none of them change, reusing the cached client is correct (cs clients
+    # are expensive to build, ~25-150 ms each). Two guard tests anchor this:
+    # `test_cloud_storage_field_set_is_stable` (catches new model fields) and
+    # `test_build_storage_instance_signature_is_stable` (catches new kwargs);
+    # both fail loudly so a reviewer classifies the change.
+    return _build_storage_instance_cached()(
```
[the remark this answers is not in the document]
zhiltsov-max: S3 client updated to only cache thread-safe objects.

zhiltsov-max:
@codex review

PR #10650, opened 2026-05-21
As/support multithreading worker2

Rwwwrl:
### Motivation and context

CVAT currently scales RQ throughput by running multiple `rq.Worker` processes per pod. For I/O-bound queues (queues whose jobs spend most of their time blocked on remote HTTP / Redis / Postgres), this is wasteful — each fork carries a full Python interpreter and Django app just to wait on a socket.

This PR introduces a thread-pool RQ worker, opt-in per queue, that runs N jobs concurrently inside one process. Operators turn it on by passing `--worker-class cvat.apps.django_rq_ext.worker.DjangoThreadPoolWorker --jobs-pool-size N` to `manage.py rqworker`. Stock `rq.Worker` remains the default for every queue; nothing changes unless a queue is explicitly migrated.

What's in the change:

- **New top-level package `cvat_libs/`**, sibling to `cvat-sdk` and `cvat-cli`, for CVAT code with no Django dependency. First module is `cvat_libs.rq_ext`, containing `ThreadPoolWorker` — a thread-pool RQ worker built on `rq.worker.BaseWorker`. Bundled into the production Docker image as an editable install and added to `cvat/requirements/base.in` so pip-compile picks it up.
- **`cvat.apps.django_rq_ext`** — Django-side wiring. Ships:
  - A shadowed `manage.py rqworker` that adds `--jobs-pool-size` and `--job-execution-time-threshold`, while falling through to django-rq's default behavior when neither flag is passed. Must be registered before `django_rq` in `INSTALLED_APPS` so Django's `get_commands()` resolves the shadow first.
  - `DjangoThreadPoolWorker`, a thin subclass of `ThreadPoolWorker` that wraps each `_perform_job` with `django.db.close_old_connections()` at the job boundary. Recommended for any CVAT queue touching the ORM; the bare `cvat_libs.rq_ext.worker.ThreadPoolWorker` is exported for non-Django consumers.
- **CI** — REST API / SDK test jobs install `-e ./cvat_libs` alongside cvat-sdk and cvat-cli so the new test subtree resolves `cvat_libs` without a sys.path shim.

Design notes worth flagging for review:

- **`--jobs-pool-size N` configures exactly N concurrent job slots.** Internally the executor is sized to `N + 1` because the worker's own heartbeat (`heartbeat()` in `_heartbeat_loop`) runs as a `ThreadPoolExecutor` future and permanently holds one slot. `_is_threadpool_full` counts the heartbeat future correctly so dequeue back-pressure kicks in at exactly N jobs — anything queued inside the executor's internal `_work_queue` would otherwise be invisible to Redis and lost on crash. The `+1` is an implementation detail; operators budget for N. Minimum is `--jobs-pool-size 1`.

- **Threads cannot be killed.** `--job-execution-time-threshold` (default 60s) is **not** a kill timeout. It drives (a) a post-completion WARNING when a job exceeds it, and (b) the drain cap on shutdown — `executor.shutdown(wait=False)` then `wait(in_flight, timeout=threshold)`. Kubernetes `terminationGracePeriodSeconds` MUST be strictly greater than the threshold so `register_death` and the final log flush can run before SIGKILL. `Job(..., timeout=N)` and `send_stop_job_command` are silently no-ops; per-job watchdogs belong in the job function (e.g. `requests.post(..., timeout=N)`). The abandonment timer is separate — it's the `StartedJobRegistry` heartbeat TTL (`job_monitoring_interval + 15s`) refreshed continuously by `_heartbeat_loop`, so a long-running job whose thread is still alive will never look abandoned to another worker's `clean_registries` pass.

- **DB / RQ connection lifecycle — two boundary fixes the fork model gave us for free.**
  - **Django DB connections.** `ThreadPoolWorker` does not fork, so a connection opened by thread T for job N stays in T's `asgiref.local.Local` slot across jobs N+1, N+2… If T idles between bursts, Postgres / pgbouncer / RDS Proxy will eventually close the TCP connection from their side; the next ORM call then raises `OperationalError: server closed the connection unexpectedly`. `DjangoThreadPoolWorker._perform_job` brackets the call with `close_old_connections()` so each job starts and ends with a fresh connection slot — reproducing what fork did implicitly.
  - **rq's current-connection contract.** `rq._connection_stack` is a `LocalStack` keyed by `threading.get_ident()`. Upstream `Worker.perform_job` does `push_connection(self.connection)` / `pop_connection()` inside the work-horse so user code can construct `rq.Queue()` without an explicit `connection=`. `ThreadPoolWorker._perform_job` does the same push/pop pair in the executor thread; without it, any job that uses an implicit rq connection raises `NoRedisConnectionException`. The regression test `test_implicit_rq_connection_is_available_inside_job_thread` pins this.

- **Feature parity with `rq.worker.Worker`.** `Retry(max=N)`, `depends_on=`, `on_success` / `on_failure` callbacks, and SIGKILL/segfault recovery via `clean_registries` all work with the same semantics as upstream. Only `Job.timeout`, callback timeouts, and the pubsub stop-job/kill-horse/shutdown commands are deliberately unsupported — all three would require killing a thread mid-blocking-IO.

- **Code split.** `cvat_libs/rq_ext/mixins.py` holds methods copy-pasted verbatim from `rq.worker.Worker @ 1.16.0` (they morally belong on `BaseWorker` upstream). `cvat_libs/rq_ext/worker.py` holds everything thread-specific plus methods adapted from upstream with documented deviations. A version-pin sentinel in the test suite asserts `rq.VERSION == "1.16.0"` with a multi-line failure message pointing the rq-bumper at the mixin/worker re-sync checklist. Full rationale is in `cvat_libs/cvat_libs/rq_ext/SKILL.md`.

No existing queue is migrated to the new worker in this PR — it is opt-in only. Follow-up PRs will move specific I/O-bound queues over once we have soak data.

### How has this been tested?

- **New pytest suite under `tests/python/libs/rq_ext/`** against real Redis — 29 tests covering thread-id isolation, pool back-pressure, heartbeat refresh on long-running jobs, drain on teardown (success and timeout), upstream-Worker feature parity ports, the implicit-rq-connection regression, and an rq-version-pin sentinel.
- **End-to-end benchmark on the `webhooks` queue** with `perform_webhook_request` stubbed to `time.sleep(2.5)`, 100 jobs:
  - Default fork-based `rq.Worker`: drained in **265s** (theoretical floor 250s — 6% overhead).
  - `DjangoThreadPoolWorker` with `--jobs-pool-size 5`: same workload in **55s** (theoretical floor 50s — 10% overhead).
  - Observed speedup **~4.82×**. All 100 jobs in both legs reached `FinishedJobRegistry` cleanly; no stale DB connections, no `NoRedisConnectionException`, no orphan heartbeats.

### Checklist

### License

- [x] I submit _my code changes_ under the same [MIT License](https://github.com/cvat-ai/cvat/blob/develop/LICENSE) that covers the project. Feel free to contact the maintainers if that's a concern.

### DIFF
--- a/changelog.d/20260521_122228_alekseyysosov_add_threadpool_rq_worker.md
+++ b/changelog.d/20260521_122228_alekseyysosov_add_threadpool_rq_worker.md
@@ -1,0 +1,9 @@
+### Added
+
+- Thread-pool variant of the RQ worker that runs multiple I/O-bound jobs
+  concurrently within one process. Operators opt in by passing
+  `--worker-class cvat_libs.rq_ext.worker.ThreadPoolWorker`
+  and `--jobs-pool-size N` to `manage.py rqworker`. Intended for queues whose
+  jobs spend most of their time blocked on remote I/O, where adding
+  worker processes per pod is wasteful
+  (<https://github.com/cvat-ai/cvat/pull/10650>)
--- a/cvat_libs/cvat_libs/rq_ext/worker.py
+++ b/cvat_libs/cvat_libs/rq_ext/worker.py
@@ -1,0 +1,436 @@
+import signal
+import sys
+import threading
+import time
+import traceback
+from concurrent.futures import Future, ThreadPoolExecutor, wait
+from contextlib import suppress
+from types import FrameType
+from typing import Optional
+
+import redis.exceptions
+from rq.connections import pop_connection, push_connection
+from rq.defaults import DEFAULT_LOGGING_DATE_FORMAT, DEFAULT_LOGGING_FORMAT
+from rq.job import Job, JobStatus
+from rq.queue import Queue
+from rq.registry import StartedJobRegistry
+from rq.utils import utcnow
+from rq.worker import BaseWorker, DequeueStrategy, StopRequested, WorkerStatus
+
+from cvat_libs.rq_ext import consts
+from cvat_libs.rq_ext.mixins import RqWorkerPortMixin
+from cvat_libs.rq_ext.utils import NoOpDeathPenalty
+
+
+class ThreadPoolWorker(RqWorkerPortMixin, BaseWorker):
+    def __init__(
+        self,
+        *args,
+        jobs_pool_size: int,
+        job_execution_time_threshold: int = 60,
+        **kwargs,
+    ) -> None:
+        super().__init__(*args, **kwargs)
+        # NOTE @sosov: +1 reserves the heartbeat slot; jobs_pool_size is the
+        # user-facing job-slot count.
+        self.threadpool_size = jobs_pool_size + 1
+
+        self._job_execution_time_threshold_sec = job_execution_time_threshold
+
+        self._executor = ThreadPoolExecutor(
+            max_workers=self.threadpool_size,
+            thread_name_prefix="rq_threadpool_",
+        )
+        self._active_futures: list[Future] = []
+        self._active_jobs: dict[Future, Job] = {}
+        self._lock = threading.Lock()
+
+        self._heartbeat_future: Optional[Future] = None
+        self._heartbeat_stop_event = threading.Event()
+
+    def subscribe(self) -> None:
+        # NOTE @sosov: BaseWorker.subscribe opens a pubsub channel for
+        # stop-job / kill-horse / shutdown commands — all horse-bound and
+        # inapplicable to threads. No-op; unsubscribe() is then also a no-op
+        # because self.pubsub_thread stays None.
+        pass
+
+    @property
+    def _is_threadpool_full(self) -> bool:
+        with self._lock:
+            return len(self._active_futures) >= self.threadpool_size
+
+    def _on_future_performed(self, future: Future) -> None:
+        with self._lock:
+            try:
+                self._active_futures.remove(future)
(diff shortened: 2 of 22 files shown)


changelog.d/20260521_122228_alekseyysosov_add_threadpool_rq_worker.md
```diff
@@ -0,0 +1,9 @@
+### Added
+
+- Thread-pool variant of the RQ worker that runs multiple I/O-bound jobs
+  concurrently within one process. Operators opt in by passing
+  `--worker-class cvat_libs.rq_ext.worker.ThreadPoolWorker`
+  and `--jobs-pool-size N` to `manage.py rqworker`. Intended for queues whose
+  jobs spend most of their time blocked on remote I/O, where adding
+  worker processes per pod is wasteful
```
zhiltsov-max, line 8: Please write from the end-user perspective. This change is likely to be interesting to administrators only, so the note should be for them.

cvat_libs/cvat_libs/rq_ext/worker.py
zhiltsov-max, line 1: Please make sure all original source files have a license header.

PR #10654, opened 2026-05-21
Add transformation for instance segmentations models in torchvision library

Pr1ncee:
### How has this been tested?
I added pytest for automation testing and also I was testing against custom script with `pytorch` testing and `cvat-sdk[pytorch]`.

### Checklist

### License

### DIFF
--- a/cvat-sdk/cvat_sdk/pytorch/_utils.py
+++ b/cvat-sdk/cvat_sdk/pytorch/_utils.py
@@ -1,3 +1,7 @@
+# Copyright (C) CVAT.ai Corporation
+#
+# SPDX-License-Identifier: MIT
+
 import numpy as np
 import PIL.Image
 import PIL.ImageDraw
@@ -6,7 +10,7 @@ from cvat_sdk import models
 from cvat_sdk.masks import decode_mask
 
 
-def _shape_bbox(shape: models.LabeledShape) -> tuple[float, float, float, float]:
+def shape_bbox(shape: models.LabeledShape) -> tuple[float, float, float, float]:
     """
     Return the unrotated bounding box of a CVAT shape as (x_min, y_min, x_max, y_max).
 
@@ -28,7 +32,7 @@ def _shape_bbox(shape: models.LabeledShape) -> tuple[float, float, float, float]
     return min(x_coords), min(y_coords), max(x_coords), max(y_coords)
 
 
-def _draw_mask(shape: models.LabeledShape, image_size: tuple[int, int]) -> np.ndarray:
+def draw_mask(shape: models.LabeledShape, image_size: tuple[int, int]) -> np.ndarray:
     """
     Rasterize a CVAT shape into a dense full-image boolean mask.
 
--- a/cvat-sdk/cvat_sdk/pytorch/_utils.py
+++ b/cvat-sdk/cvat_sdk/pytorch/_utils.py
@@ -54,4 +54,4 @@ def draw_mask(shape: models.LabeledShape, image_size: tuple[int, int]) -> np.nda
     else:
         raise AssertionError(f"unsupported shape type: {shape_type!r}")
 
-    return np.array(image)
+    return np.asarray(image)
--- a/cvat-sdk/cvat_sdk/pytorch/common.py
+++ b/cvat-sdk/cvat_sdk/pytorch/common.py
@@ -23,3 +23,8 @@ class Target:
     A mapping from label_id values in `LabeledImage` and `LabeledShape` objects
     to an integer index. This mapping is consistent across all samples for a given task.
     """
+
+    image_size: tuple[int, int] | None = None
+    """
+    The size of the corresponding image as a (width, height) tuple, if known.
+    """
--- a/cvat-sdk/cvat_sdk/pytorch/common.py
+++ b/cvat-sdk/cvat_sdk/pytorch/common.py
@@ -24,7 +24,7 @@ class Target:
     to an integer index. This mapping is consistent across all samples for a given task.
     """
 
-    image_size: tuple[int, int] | None = None
+    image_size: tuple[int, int]
     """
-    The size of the corresponding image as a (width, height) tuple, if known.
+    The size of the corresponding image as a (width, height) tuple.
     """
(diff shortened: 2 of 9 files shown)


cvat-sdk/cvat_sdk/pytorch/common.py
```diff
@@ -23,3 +23,8 @@ class Target:
     A mapping from label_id values in `LabeledImage` and `LabeledShape` objects
     to an integer index. This mapping is consistent across all samples for a given task.
     """
+
+    image_size: tuple[int, int] | None = None
+    """
+    The size of the corresponding image as a (width, height) tuple, if known.
```
SpecLad: The "if known" part seems a bit unclear - when is it not known?

cvat-sdk/cvat_sdk/pytorch/transforms.py
```diff
@@ -2,15 +2,32 @@
...
 
 import attrs
 import attrs.validators
 import torch
 import torch.utils.data
+from torch import Tensor
 
 from cvat_sdk.datasets.common import UnsupportedDatasetError
 from cvat_sdk.pytorch.common import Target
+from cvat_sdk.pytorch.utils import _draw_mask, _shape_bbox
```
SpecLad: If these functions are exported from a module, they should not begin with an underscore. Maybe the module itself should have a leading underscore, though, to discourage people from using it directly.

cvat-sdk/cvat_sdk/pytorch/transforms.py
```diff
@@ -2,15 +2,32 @@
...
+
+_SUPPORTED_SHAPE_TYPES = frozenset(["rectangle", "polygon", "polyline", "points", "ellipse"])
```
SpecLad: This one should probably be renamed to `_SUPPORTED_BBOX_SHAPE_TYPES` to distinguish it from the other. Or you could keep the name and move it inside `ExtractBoundingBoxes`, since I don't think it's used outside of it anyway. Or just inline it. 🙂

cvat-sdk/cvat_sdk/pytorch/transforms.py
```diff
@@ -81,13 +90,74 @@ def __call__(self, target: Target) -> LabeledBoxes:
...
+                torch.as_tensor(
+                    _draw_mask(shape=shape, image_size=target.image_size), dtype=torch.uint8
+                )
+            )
+
+        return LabeledMasks(
+            boxes=(
+                torch.tensor(boxes, dtype=torch.float)
+                if boxes
+                else torch.empty((0, 4), dtype=torch.float)
```
SpecLad, line 157: I see you've taken care to set the correct dimensions for empty tensors, which is good, but could you also update `ExtractBoundingBoxes` to do the same, to avoid inconsistency? Make sure to note it in the changelog, as well.

cvat-sdk/cvat_sdk/pytorch/_utils.py
```diff
@@ -0,0 +1,53 @@
+import numpy as np
```
SpecLad, line 5: Please add a copyright banner.

cvat-sdk/cvat_sdk/pytorch/_utils.py
```diff
@@ -0,0 +1,53 @@
...
+
+    image = PIL.Image.new("1", image_size)
+    draw = PIL.ImageDraw.Draw(image)
+
+    if shape_type == "polygon":
+        draw.polygon(list(zip(shape.points[0::2], shape.points[1::2])), fill=1)
+    else:
+        raise AssertionError(f"unsupported shape type: {shape_type!r}")
+
+    return np.array(image)
```
SpecLad:
```suggestion
    return np.asarray(image)
```
 This should avoid making an extra copy.

tests/python/sdk/test_pytorch.py
```diff
@@ -211,6 +212,98 @@ def test_extract_bounding_boxes(self):
         assert torch.equal(dataset[7][1]["boxes"], torch.tensor([]))
         assert torch.equal(dataset[7][1]["labels"], torch.tensor([]))
 
+    def test_extract_instance_masks(self):
```
SpecLad, line 253: Could you also add a more integrated test, where `ExtractInstanceMasks` is used as a transform (a la `test_extract_bounding_boxes`)?

cvat-sdk/cvat_sdk/pytorch/common.py
Pr1ncee: because we are using the same object `Target` for all models: classification, object detection and instance segmentation. I can propose creating a new object for instance segmentation then

SpecLad: I still don't understand. Is there anything preventing us from initializing this field in all cases?

Pr1ncee: yes, because we don't use image_size when transforming objects torchvision classification and object detection models. This field is exclusively used by instance segmentation models

SpecLad: Sure, but the fact that we don't _use_ it doesn't stop us from setting it. So what's the problem?

cvat-sdk/cvat_sdk/pytorch/transforms.py
```diff
@@ -81,13 +89,77 @@ def __call__(self, target: Target) -> LabeledBoxes:
...
+    "labels": a tensor with shape [N] containing corresponding label indices.
+    "masks": a uint8 tensor with shape [N, H, W] containing binary instance masks.
+
+    Limitations:
+
+    * Only the following shape types are supported: polygon, mask.
+    * Rotated shapes are not supported.
+    """
+
+    SUPPORTED_MASK_SHAPE_TYPES = frozenset(["polygon", "mask"])
```
SpecLad:
```suggestion
    SUPPORTED_SHAPE_TYPES = frozenset(["polygon", "mask"])
```

This would be more consistent with `ExtractBoundingBoxes`.

SpecLad (approved):

zhiltsov-max:
I guess this PR was merged without squashing. Please make sure you're merging with squashing - this is our default.

PR #10659, opened 2026-05-22
Preserve GT annotation source

### LINKED ISSUE #8874: Editing / redo-ing annotation edit in the GT job changes the shape source from "ground truth" to "manual"

### Steps to Reproduce

1. Create a task with a gt job
2. Create an annotation in the gt job
3. Edit the annotation
4. Annotation source changes from `Ground Truth` to `Manual`
5. Undo
6. Redo
7. Annotation source changes from `Ground Truth` to `Manual`

### Expected Behavior

Annotation source is not changed.

### Possible Solution

_No response_

### Context

_No response_

### Environment

_No response_

prathamesh-patil-5090:
## Summary

Fix GT annotation source handling so annotations opened inside a ground truth job keep their original source instead of being shown as `Ground truth`.

`Ground truth` is now treated as a review-mode presentation state only. GT annotations are presented with `Source.GT` when they are loaded as reference overlays for another job in Review workspace, but they are not rewritten when working directly inside the GT job.

Closes #8874


prathamesh-patil-5090:
@klakhov @zhiltsov-max I updated the annotation source handling so GT annotations keep their GT source internally during edit, undo, and redo. The GT-job UI behavior stays the same, so GT badges are still hidden in GT jobs and the objects are treated like regular annotations there.

zhiltsov-max:
> Annotations created in Ground Truth jobs should permanently retain their GT source

This is incorrect. What is actually needed is that the UI conforms to the server API when saving annotations. "Ground Truth" should only be used for presentation in the review mode for GT job annotations.

Only these values for the `source` field should come to the server: https://github.com/cvat-ai/cvat/blob/d5ddce049bf15468c36686693f7156a1fad88f3f/cvat/apps/engine/models.py#L1545

After the fix, we should be able to merge this one: https://github.com/cvat-ai/cvat/pull/10521

prathamesh-patil-5090:
> > Annotations created in Ground Truth jobs should permanently retain their GT source
> 
> This is incorrect. What is actually needed is that the UI conforms to the server API when saving annotations. "Ground Truth" should only be used for presentation in the review mode for GT job annotations.
> 
> Only these values for the `source` field should come to the server:
> 
> https://github.com/cvat-ai/cvat/blob/d5ddce049bf15468c36686693f7156a1fad88f3f/cvat/apps/engine/models.py#L1545
> 
> After the fix, we should be able to merge this one: #10521

Thanks for the clarification. I’ll keep Ground Truth as presentation-only in review mode and make sure the save path always sends one of the server-accepted source values. I’ll update the patch accordingly and re-check that it unblocks #10521.

klakhov (approved):
Thanks for contribution!

PR #10660, opened 2026-05-25
Add a thread-pool RQ worker for I/O-bound queues

Rwwwrl:
### Motivation and context

CVAT scales RQ throughput by running multiple worker processes per pod. For I/O-bound queues — those whose jobs spend almost all their time blocked on remote HTTP / Redis / Postgres — this is wasteful: every fork carries a full Python interpreter and Django app just to wait on a socket.

This adds a thread-pool worker that runs N jobs concurrently inside one process, opt-in per queue. Administrators enable it by starting a queue's worker with `--worker-class cvat.apps.redis_handler.worker.ThreadPoolWorker` and setting `CVAT_RQ_THREAD_POOL_SIZE`. It refreshes DB connections at each job boundary, since threads (unlike forks) don't get a clean connection per job. Stock `rq.Worker` stays the default — nothing changes unless a queue is explicitly switched, and no queue is migrated in this PR.

Caveats, because a running thread can't be interrupted mid-blocking-I/O:
- `Job.timeout`, callback timeouts, and the pubsub stop-job/kill commands are no-ops on this worker — per-job watchdogs belong inside the job function (e.g. `requests.post(..., timeout=...)`).
- The worker marks itself non-cancellable, so the requests API rejects cancellation of an already-started job on it instead of reporting a success that wouldn't take effect.

### Checklist

### License

### DIFF
--- a/changelog.d/20260521_122228_alekseyysosov_add_threadpool_rq_worker.md
+++ b/changelog.d/20260521_122228_alekseyysosov_add_threadpool_rq_worker.md
@@ -1,0 +1,11 @@
+### Added
+
+- A new thread-pool background worker that lets a single worker process handle
+  many jobs at once. For queues whose jobs mostly wait on network I/O (such as
+  outgoing webhooks), one such worker can now do the work that previously needed
+  several separate worker processes, lowering the memory footprint of a CVAT
+  deployment. It is off by default; administrators can enable it for a specific
+  queue by starting that worker with `--worker-class
+  cvat.apps.redis_handler.worker.ThreadPoolWorker` and setting the number
+  of concurrent jobs through the `CVAT_RQ_THREAD_POOL_SIZE` environment variable
+  (<https://github.com/cvat-ai/cvat/pull/10650>)
--- a/cvat/apps/redis_handler/mixins.py
+++ b/cvat/apps/redis_handler/mixins.py
@@ -1,0 +1,186 @@
+# Copyright (C) CVAT.ai Corporation
+#
+# SPDX-License-Identifier: MIT
+
+import traceback
+from datetime import timedelta
+from random import shuffle
+from typing import TYPE_CHECKING, Optional
+
+from rq import worker_registration
+from rq.exceptions import DeserializationError
+from rq.utils import utcformat, utcnow
+from rq.worker import DequeueStrategy
+
+if TYPE_CHECKING:
+    from redis.client import Pipeline
+
+try:
+    from setproctitle import setproctitle as setprocname
+except ImportError:
+
+    def setprocname(*args, **kwargs):  # noqa
+        pass
+
+
+class RqWorkerPortMixin:
+    """Methods copy-pasted verbatim from `rq.worker.Worker` @ 1.16.0.
+
+    Each method below is referenced by `BaseWorker` (or by us through the BaseWorker
+    contract) but lives only on `Worker`. Nothing in this mixin diverges from
+    upstream — modifications, if any, belong in the subclass that uses the mixin.
+
+    rq-version re-sync checklist (run on every rq bump; `test_rq_version_pin`
+    pins the version and points here):
+      1. Diff each method below against the new upstream `rq.worker.Worker` and
+         re-apply any changes verbatim.
+      2. Re-validate the upstream-adapted methods in `worker.py`
+         (`handle_job_failure`, `get_heartbeat_ttl`, `prepare_job_execution`,
+         `_perform_job`, `handle_job_success`) against upstream and re-apply
+         documented divergences.
+    """
+
+    @property
+    def connection_timeout(self) -> int:
+        return self.dequeue_timeout + 10
+
+    def set_state(self, state: str, pipeline=None) -> None:
+        self._state = state
+        connection = pipeline if pipeline is not None else self.connection
+        connection.hset(self.key, "state", state)
+
+    def get_state(self) -> str:
+        return self._state
+
+    def push_exc_handler(self, handler_func) -> None:
+        self._exc_handlers.append(handler_func)
+
+    def pop_exc_handler(self):
+        return self._exc_handlers.pop()
+
+    def procline(self, message: str) -> None:
+        setprocname(f"rq:worker:{self.name}: {message}")
+
+    def reorder_queues(self, reference_queue) -> None:
+        if self._dequeue_strategy is None:
+            self._dequeue_strategy = DequeueStrategy.DEFAULT
+
+        if self._dequeue_strategy not in ("default", "random", "round_robin"):
+            raise ValueError(
+                f"Dequeue strategy {self._dequeue_strategy} is not allowed. "
+                "Use `default`, `random` or `round_robin`."
+            )
+        if self._dequeue_strategy == DequeueStrategy.DEFAULT:
+            return
+        if self._dequeue_strategy == DequeueStrategy.ROUND_ROBIN:
+            pos = self._ordered_queues.index(reference_queue)
+            self._ordered_queues = self._ordered_queues[pos + 1 :] + self._ordered_queues[: pos + 1]
+            return
+        if self._dequeue_strategy == DequeueStrategy.RANDOM:
+            shuffle(self._ordered_queues)
+            return
+
+    def register_birth(self) -> None:
+        self.log.debug("Registering birth of worker %s", self.name)
+        if self.connection.exists(self.key) and not self.connection.hexists(self.key, "death"):
+            raise ValueError(f"There exists an active worker named {self.name!r} already")
+        with self.connection.pipeline() as p:
+            p.delete(self.key)
+            now = utcnow()
+            now_in_string = utcformat(now)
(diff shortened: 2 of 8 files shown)


cvat/apps/redis_handler/worker.py
zhiltsov-max, line 1: I suggest renaming the file to `threaded_worker.py` or something like this.

cvat/apps/redis_handler/mixins.py
```diff
@@ -0,0 +1,186 @@
...
+
+try:
+    from setproctitle import setproctitle as setprocname
+except ImportError:
+
+    def setprocname(*args, **kwargs):  # noqa
+        pass
+
+
+class RqWorkerPortMixin:
```
zhiltsov-max, line 26: Do I understand correctly that the only purpose of this class is to protect against unknown upstream implementation changes?

Rwwwrl: Not exactly. Methods inside that class are exactly copied from the Worker that is not the BaseWorker - that is the whole problem in rq design. Some base methods are completely fork logic independent and they are required for any class: Multithreading / Gevent / Asyncio / whateverelse - they all will need those found methods. 

See the picture 
image

zhiltsov-max: Ok, so these are reusable bits of (Fork)Worker that should belong to the base class. What do you think about just making it a base class instead? So the inheritance chain will be `BaseWorker` -> new extended `BaseWorker` -> `ThreadedWorker`? Unless you plan adding more mixins, it seems like a straightforward solution.

Rwwwrl: I also thought about that, but that would miss clarity. Here we have like
```
class RqWorkerPortMixin:
    """Methods copy-pasted verbatim from `rq.worker.Worker` @ 1.16.0.

    Each method below is referenced by `BaseWorker` (or by us through the BaseWorker
    contract) but lives only on `Worker`. Nothing in this mixin diverges from
    upstream — modifications, if any, belong in the subclass that uses the mixin.

    rq-version re-sync checklist (run on every rq bump; `test_rq_version_pin`
    pins the version and points here):
      1. Diff each method below against the new upstream `rq.worker.Worker` and
         re-apply any changes verbatim.
      2. Re-validate the upstream-adapted methods in `worker.py`
         (`handle_job_failure`, `get_heartbeat_ttl`, `prepare_job_execution`,
         `_perform_job`, `handle_job_success`) against upstream and re-apply
         documented divergences.
    """
```
that is clear, that we just copied methods from (Fork)Worker and nothing more. But we absolutely can create BaseWorker(lib_BaseWorker, RqWorkerPortMixin) - that one is cleaniest approach I believe

Rwwwrl: updated, see the https://github.com/cvat-ai/cvat/pull/10660/changes/122f76dc07a1469dfaa18bd7f55fd9751518e4c2

cvat/apps/redis_handler/worker.py
Rwwwrl: I would go with worker.py, because:
1. we already have rq_worker (that I can important to that file, for now I left it as out of scope)
2. it follows the rq library naming

at the end I can rename it to "rq_worker.py", since redis_handler is mixed up with all redis stuff

changelog.d/20260521_122228_alekseyysosov_add_threadpool_rq_worker.md
```diff
@@ -0,0 +1,11 @@
+### Added
+
+- A new thread-pool background worker that lets a single worker process handle
+  many jobs at once. For queues whose jobs mostly wait on network I/O (such as
+  outgoing webhooks), one such worker can now do the work that previously needed
+  several separate worker processes, lowering the memory footprint of a CVAT
+  deployment. It is off by default; administrators can enable it for a specific
+  queue by starting that worker with `--worker-class
+  cvat.apps.redis_handler.worker.ThreadPoolWorker` and setting the number
+  of concurrent jobs through the `CVAT_RQ_THREAD_POOL_SIZE` environment variable
```
zhiltsov-max, line 10: Please contract this description to 1-3 lines? The full description can stay in the PR or in the docs.

cvat/apps/redis_handler/worker.py
```diff
@@ -0,0 +1,486 @@
...
+
+        self._heartbeat_future: Optional[Future] = None
+        self._heartbeat_stop_event = threading.Event()
+
+    def subscribe(self) -> None:
+        # NOTE @sosov: BaseWorker.subscribe opens a pubsub channel for
+        # stop-job / kill-horse / shutdown commands — all horse-bound and
+        # inapplicable to threads. No-op; unsubscribe() is then also a no-op
+        # because self.pubsub_thread stays None.
+        pass
```
zhiltsov-max, line 69: Why `shutdown` is also ignored? It seems to be a whole worker signal, not specific to the workhorse. The default Worker's handler sends `SIGINT` to self, which results in a `request_stop()` call.

cvat/apps/redis_handler/worker.py
```diff
@@ -0,0 +1,486 @@
...
+                else settings.RQ_THREAD_POOL_JOB_EXECUTION_TIME_THRESHOLD
+            ),
+            **kwargs,
+        )
+
+    def register_birth(self) -> None:
+        super().register_birth()
+        # NOTE @sosov: a running thread can't be interrupted, so RequestViewSet
+        # must reject started-job cancellation for this worker. Mirrors SimpleWorker.
+        self.connection.hset(self.key, "cvat_can_stop_started_jobs", 0)
```
zhiltsov-max, line 479: Please use the variable from `cvat/apps/redis_handler/views.py:41`

cvat/apps/redis_handler/worker.py
```diff
@@ -0,0 +1,486 @@
...
+
+_JOB_HEARTBEAT_REFRESH_BUFFER_SEC: int = 15
+
+
+class BaseWorker(_RqBaseWorker, RqWorkerPortMixin):
+    pass
+
+
+class RqThreadPoolWorker(BaseWorker):
+    def __init__(
```
zhiltsov-max, line 39: Please add a warning for the cases when the worker is attached to more than 1 queue - it may lose task execution guarantees. The reason is that the `dequeue_job_and_maintain_ttl() -> dequeue_any()` call will use the less safe call to `lpop()`, which may result in the dequeued job lost if it wasn't added to started job queue.

cvat/apps/redis_handler/worker.py
```diff
@@ -0,0 +1,486 @@
...
+
+    def _shutdown(self) -> None:
+        self._stop_requested = True
+
+        self.set_shutdown_requested_date()
+
+        if self.scheduler:
+            self.stop_scheduler()
+
+        raise StopRequested()
```
zhiltsov-max, line 409: Raising exceptions from signal handlers is not recommended (https://docs.python.org/3/library/signal.html#note-on-signal-handlers-and-exceptions), but should be fine for webhhoks, provided there are retries and 1 queue. Just FYI.

cvat/apps/redis_handler/worker.py
```diff
@@ -0,0 +1,486 @@
...
+        self._shutdown()
+
+    def _stop_heartbeat(self) -> None:
+        self._heartbeat_stop_event.set()
+
+    def teardown(self) -> None:
+        if self.scheduler:
+            self.stop_scheduler()
+
+        with self._lock:
```
zhiltsov-max, line 432: This line can potentially hang for ~`worker_ttl`, if Redis connection hangs for some reason, e.g. during `set_state()` call. It's unlikely to be a big issue on practice, but consider adding some timeout.

cvat/apps/redis_handler/worker.py
```diff
@@ -0,0 +1,486 @@
...
+            except ValueError:
+                pass
+
+            self._active_jobs.pop(future, None)
+
+            if future is self._heartbeat_future:
+                return
+
+            if all(f is self._heartbeat_future for f in self._active_futures):
+                self.set_state(WorkerStatus.IDLE)
```
zhiltsov-max, line 89: Calling `set_state()` while holding the lock is fine if we have few threads, but if we scale it up to 100, there are several potential issues:
- it's possible that a single Redis connection issue will result in prolonged waiting for all the threads
- there's a risk of thread contention on the lock

I think it makes sense to add a `SOCKET_TIMEOUT` for the Redis connection in `RQ_QUEUES` (it also addresses the potential lock hang in `teardown()`). The second issue is unlikely to be significant, given our load and GIL presence.

An extended solution is to move these status notifications to the heartbeat update. Given that they are needed just for monitoring purposes, some delay on the updates should be fine. I'm not sure we really will scale up to the numbers where it's important soon, so just adding a comment for this is fine.

I can also see there is a little race on setting the state from `dequeue_job_and_maintain_ttl()` in the main thread and worker threads that start processing a new job.

Upd: testing showed no contention issues for 50 threads, if the Redis connection is fine.

PR #10665, opened 2026-05-25
Fix SDK methods IAM issues

### LINKED ISSUE #8914: SDK methods that list child objects don't always work when the parent object is in an organization

### Steps to Reproduce

1. Create two accounts, `user1` and `user2`.
1. Create an organization as `user1`, and add `user2` as a maintainer.
1. Create a project in the organization as `user1`.
1. Create a task in the project as `user1`.
1. Create the following script:

   ```python
   import os
   import sys

   from cvat_sdk import make_client

   def main():
       proj_id = int(sys.argv[1])

       with make_client(
           host="http://localhost:8080",
           credentials=(os.environ["CVAT_AI_USER"], os.environ["CVAT_AI_PASS"])
       ) as client:
           project = client.projects.retrieve(proj_id)

           for task in project.get_tasks():
               print(task.id)

   main()
```

Run the script with `user2`'s credentials, passing the ID of the project. It prints nothing.

### Expected Behavior

The script should print the ID of the task.

By comparison, if you run the script with `user1`'s credentials, this does happen.

### Possible Solution

We should probably make the SDK override the current organization while making the "list tasks" request, if the project is in an organization.

Alternatively, we could alter the server's response to list all tasks that the user has access to, regardless of what organization they are in.

### Context

I believe this affects all SDK methods that list child objects of some other objects, although I haven't tested them all. These methods are:

```
Issue.get_comments
Job.get_labels
Job.get_issues
Project.get_tasks
Project.get_labels
Task.get_jobs
Task.get_labels
```

### Environment

```Markdown
Reproducible with commit b594b1cbc7118d64621a78bb4f937ebd44c451f1.
```

Pr1ncee:
<!-- Raise an issue to propose your change (https://github.com/cvat-ai/cvat/issues).
It helps to avoid duplication of efforts from multiple independent contributors.
Discuss your ideas with maintainers to be sure that changes will be approved and merged.
Read the [Contribution guide](https://docs.cvat.ai/docs/contributing/). -->

### Motivation and context
Fixes #8914 

**Issue**: Several high-level SDK methods — `Project.get_tasks()`, `Project.get_labels()`, `Task.get_jobs()`, `Task.get_labels()`, `Job.get_issues()`, `Job.get_labels()`, and `Issue.get_comments()` — failed with IAM        
  permission errors when called on resources belonging to an organization, unless the client had already been configured with the correct organization context via `client.organization_context()`.                
   
**Root cause**: These methods issued API requests without propagating the organization scope inferred from the resource itself, so CVAT's IAM layer rejected them as if they were personal workspace requests.     
                                                                                                                                                                                                               
**Fix**: Added a `organization_context_for(client, organization_id)` context manager in `model_proxy.py`. It looks up the organization from the resource's organization field and temporarily sets the matching org slug on the client (skipping the switch if the slug is already correct). All affected proxy methods now wrap their API calls with this context manager, so the organization scope is automatically applied regardless of how the client was initialized.                                                                                                                                                                  
                                                                                                                                                                                                               
  Tests covering all affected methods with org-owned resources were added to verify the fix. 

### How has this been tested?
The code was tested against custom scripts (including one in the issue) as well as automated tests

### Checklist
<!-- Go over all the following points, and put an `x` in all the boxes that apply.
If an item isn't applicable for some reason, then ~~explicitly strikethrough~~ the whole
line. If you don't do that, GitHub will show incorrect progress for the pull request.
- [x] I have created a changelog fragment <!-- see top comment in CHANGELOG.md -->

### License

### DIFF
--- a/changelog.d/20260526_000000_andrei_sdk_iam_fix.md
+++ b/changelog.d/20260526_000000_andrei_sdk_iam_fix.md
@@ -9,7 +9,8 @@
 
 ### Changed
 
-- \[SDK\] Changed requests that explicitly specify an organization by ID being
-  overridden by the client-level `X-Organization` header, causing them to
+- \[SDK\] The `x_organization` parameter passed to individual API calls
+  now takes precedence over the client-level `organization_slug`. Previously,
+  `organization_slug` would silently override it, causing those calls to
   operate in the wrong organization context
   (<https://github.com/cvat-ai/cvat/pull/10665>)
--- a/changelog.d/20260526_000000_andrei_sdk_iam_fix.md
+++ b/changelog.d/20260526_000000_andrei_sdk_iam_fix.md
@@ -1,6 +1,7 @@
 ### Fixed
 
-- \[SDK\] Fixed `get_tasks()`, `get_labels()`, `get_jobs()`, `get_issues()`, and `get_comments()`
+- \[SDK\] Fixed `Project.get_tasks()`, `Project.get_labels()`, `Task.get_jobs()`,
+  `Task.get_labels()`, `Job.get_issues()`, `Job.get_labels()`, and `Issue.get_comments()`
   methods failing with IAM permission errors when called on organization resources without
   an explicit organization context set on the client
   (<https://github.com/cvat-ai/cvat/pull/10665>)
--- a/changelog.d/20260526_000000_andrei_sdk_iam_fix.md
+++ b/changelog.d/20260526_000000_andrei_sdk_iam_fix.md
@@ -5,3 +5,11 @@
   methods returning an empty list when called on organization resources without
   an explicit organization context set on the client
   (<https://github.com/cvat-ai/cvat/pull/10665>)
+
+- \[SDK\] Fixed the generated API client in
+  `cvat-sdk/gen/templates/openapi-generator/api_client.mustache` so per-request
+  header arguments override default headers, and header arguments with `None`
+  values are omitted instead of being serialized. This allows organization-aware
+  SDK helpers to suppress the default `X-Organization` header while sending
+  requests with `org_id`
+  (<https://github.com/cvat-ai/cvat/pull/10665>)
--- a/changelog.d/20260526_000000_andrei_sdk_iam_fix.md
+++ b/changelog.d/20260526_000000_andrei_sdk_iam_fix.md
@@ -2,6 +2,6 @@
 
 - \[SDK\] Fixed `Project.get_tasks()`, `Project.get_labels()`, `Task.get_jobs()`,
   `Task.get_labels()`, `Job.get_issues()`, `Job.get_labels()`, and `Issue.get_comments()`
-  methods failing with IAM permission errors when called on organization resources without
+  methods returning an empty list when called on organization resources without
   an explicit organization context set on the client
   (<https://github.com/cvat-ai/cvat/pull/10665>)
--- a/changelog.d/20260526_000000_andrei_sdk_iam_fix.md
+++ b/changelog.d/20260526_000000_andrei_sdk_iam_fix.md
@@ -3,13 +3,13 @@
 - \[SDK\] Fixed `Project.get_tasks()`, `Project.get_labels()`, `Task.get_jobs()`,
   `Task.get_labels()`, `Job.get_issues()`, `Job.get_labels()`, and `Issue.get_comments()`
   methods returning an empty list when called on organization resources without
-  an explicit organization context set on the client
+  an explicit organization context set on the client, or with a different
+  organization context set
   (<https://github.com/cvat-ai/cvat/pull/10665>)
 
-- \[SDK\] Fixed the generated API client in
-  `cvat-sdk/gen/templates/openapi-generator/api_client.mustache` so per-request
-  header arguments override default headers, and header arguments with `None`
-  values are omitted instead of being serialized. This allows organization-aware
-  SDK helpers to suppress the default `X-Organization` header while sending
-  requests with `org_id`
+### Changed
+
+- \[SDK\] Changed requests that explicitly specify an organization by ID being
+  overridden by the client-level `X-Organization` header, causing them to
+  operate in the wrong organization context
   (<https://github.com/cvat-ai/cvat/pull/10665>)
--- a/changelog.d/20260526_000000_andrei_sdk_iam_fix.md
+++ b/changelog.d/20260526_000000_andrei_sdk_iam_fix.md
@@ -1,0 +1,6 @@
+### Fixed
+
+- \[SDK\] Fixed `get_tasks()`, `get_labels()`, `get_jobs()`, `get_issues()`, and `get_comments()`
+  methods failing with IAM permission errors when called on organization resources without
+  an explicit organization context set on the client
+  (<https://github.com/cvat-ai/cvat/pull/10665>)
--- a/cvat-sdk/cvat_sdk/core/client.py
+++ b/cvat-sdk/cvat_sdk/core/client.py
@@ -171,9 +171,7 @@ class Client:
     _ORG_SLUG_HEADER = _ORG_SLUG_HEADER
 
     @contextmanager
-    def _scoped_organization_slug(
-        self, org_slug: str | None
-    ) -> Generator[None, None, None]:
+    def _scoped_organization_slug(self, org_slug: str | None) -> Generator[None, None, None]:
         """Temporarily override the X-Organization header for the calling
         thread only. Safe to use concurrently across threads, unlike
         :py:attr:`organization_slug` which mutates shared client state.
--- a/cvat-sdk/cvat_sdk/core/client.py
+++ b/cvat-sdk/cvat_sdk/core/client.py
@@ -6,6 +6,7 @@
 from __future__ import annotations
 
 import logging
+import threading
 import urllib.parse
 from abc import ABCMeta
 from collections.abc import Generator, Sequence
@@ -37,6 +38,8 @@ from cvat_sdk.core.proxies.users import UsersRepo
 from cvat_sdk.version import VERSION
 
 _DEFAULT_CACHE_DIR = platformdirs.user_cache_path("cvat-sdk", "CVAT.ai")
+_ORG_SLUG_HEADER = "X-Organization"
+_THREAD_LOCAL_UNSET = object()
 
 _RepoType = TypeVar("_RepoType", bound=Repo)
 
@@ -90,6 +93,40 @@ class AccessTokenCredentials(Credentials):
 _VERSION_OBJ = pv.Version(VERSION)
 
 
+class _ThreadOrgAwareApiClient(ApiClient):
+    """ApiClient that allows the X-Organization header to be overridden
+    per-thread, layered on top of ``default_headers``.
+
+    Mutating ``default_headers["X-Organization"]`` to scope a single request
+    is not thread-safe because ``default_headers`` is shared across all
+    requests issued by every thread that uses this client. This subclass
+    adds a ``threading.local()`` override that ``get_common_headers``
+    consults, so a thread can temporarily change its own effective
+    X-Organization for a single request without affecting concurrent
+    requests from other threads.
+
+    Override values:
+        - ``UNSET``  → use the header from ``default_headers``.
+        - ``None``   → suppress the header for this thread's requests.
+        - ``str``    → use this exact value for this thread's requests.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._org_slug_local = threading.local()
+
+    def get_common_headers(self) -> dict[str, str]:
+        headers = super().get_common_headers()
+        override = getattr(self._org_slug_local, "value", _THREAD_LOCAL_UNSET)
+        if override is _THREAD_LOCAL_UNSET:
+            return headers
+        if override is None:
+            headers.pop(_ORG_SLUG_HEADER, None)
+        else:
+            headers[_ORG_SLUG_HEADER] = override
+        return headers
+
+
 class Client:
     """
     Provides session management, implements authentication operations
@@ -120,7 +157,7 @@ class Client:
         self.api_map = CVAT_API_V2(url)
         """Handles server API URL interaction logic"""
 
-        self.api_client = ApiClient(
+        self.api_client = _ThreadOrgAwareApiClient(
             Configuration(host=self.api_map.host, verify_ssl=self.config.verify_ssl)
         )
         """Provides low-level access to the CVAT server"""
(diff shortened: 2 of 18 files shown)


cvat-sdk/cvat_sdk/core/proxies/model_proxy.py
```diff
@@ -86,6 +87,20 @@ class Repo(ModelProxy[ModelType, ApiType]):
     _entity_type: type[Entity[ModelType, ApiType]]
 
 
+@contextmanager
+def organization_context_for(client: Client, organization_id: int | None):
+    if not organization_id:
+        yield
+        return
```
SpecLad: I'm not sure this is correct; suppose `client.organization_slug` is already set, and the user calls `project.get_tasks()` on a project in their personal workspace. Seems like we'd have to switch to the personal workspace for the listing to work.

cvat-sdk/cvat_sdk/core/proxies/issues.py
```diff
@@ -52,12 +53,15 @@ class Issue(
     _model_partial_update_arg = "patched_issue_write_request"
 
     def get_comments(self) -> list[Comment]:
-        return [
-            Comment(self._client, m)
-            for m in get_paginated_collection(
-                self._client.api_client.comments_api.list_endpoint, issue_id=self.id
-            )
-        ]
+        with organization_context_for(
```
SpecLad: There's a big problem with this approach: it's not thread safe. If a user works with different objects in separate threads, this will create a race condition. Can you think of a way to make this thread-safe?

cvat-sdk/cvat_sdk/core/proxies/model_proxy.py
```diff
@@ -86,6 +87,20 @@ class Repo(ModelProxy[ModelType, ApiType]):
...
+
+    org = client.organizations.retrieve(organization_id)
```
SpecLad: It's a bit wasteful to query the organization each time. As I recall, the server supports specifying the organization by ID if you use the `org_id` query parameter. Could you see if you can make use of that?

cvat-sdk/cvat_sdk/core/proxies/model_proxy.py
Pr1ncee: I fixed so the new code explicitly sends `{"org": ""}` when `organization_id is None`, which forces the server to use the personal workspace. If `client.organization_slug` is set, it creates a temp_api_client without the `X-Organization` header, then appends org= as a query param

cvat-sdk/cvat_sdk/core/proxies/model_proxy.py
Pr1ncee: Now the new code passes `org_id=organization_id` directly to the server as a query parameter, eliminating that retrieval entirely. All callers already have the organization int field on the model

cvat-sdk/cvat_sdk/core/proxies/issues.py
Pr1ncee: The applied fix:
1. _ThreadOrgAwareApiClient (subclass of ApiClient) adds a threading.local() slot and overrides get_common_headers() so each request reads its own thread's override first, falling back to default_headers if
  none is set. Three states: UNSET (use default), None (suppress header), str (use that value).
2. Client._scoped_organization_slug(...) is a contextmanager that sets/restores the override on threading.local() for the calling thread only — no shared state touched.
3. get_paginated_collection_with_organization now wraps its request in with client._scoped_organization_slug(temporary_org_slug): instead of assigning to client.organization_slug.

  Result: thread A scoping its request to "" (personal) and thread B with client.organization_slug = "org-x" no longer interfere — A's override lives in its own threading.local, B keeps reading the unchanged
  default_headers.

cvat/apps/engine/views.py
```diff
@@ -789,6 +789,12 @@ def __call__(self):
                 else "\n".join([str(d) for d in ex.detail])
             )
             return Response(data=msg, status=ex.status_code)
+        except FileNotFoundError:
+            slogger.glob.exception("Requested media file was not found")
```
SpecLad: How can this happen? The client can only identify media files using indexes, and the server maps those indexes to known media files, so it should never try to open a media file that doesn't exist. Maybe you found a bug in the server logic?

changelog.d/20260526_000000_andrei_sdk_iam_fix.md
```diff
@@ -0,0 +1,7 @@
+### Fixed
+
+- \[SDK\] Fixed `Project.get_tasks()`, `Project.get_labels()`, `Task.get_jobs()`,
+  `Task.get_labels()`, `Job.get_issues()`, `Job.get_labels()`, and `Issue.get_comments()`
+  methods failing with IAM permission errors when called on organization resources without
```
SpecLad: Did they actually fail with permission errors? I thought they just returned an empty list.

cvat-sdk/cvat_sdk/core/proxies/model_proxy.py
```diff
@@ -86,6 +87,27 @@ class Repo(ModelProxy[ModelType, ApiType]):
...
+    if organization_id is None:
+        # Empty org slug is the SDK/server convention for the personal workspace.
+        params = {"org": ""}
+        temporary_org_slug = ""
+    else:
+        # org_id cannot be combined with X-Organization, so clear any active slug.
+        params = {"org_id": organization_id}
+        temporary_org_slug = None
+
+    with client._scoped_organization_slug(temporary_org_slug):
```
SpecLad: I'm a bit wary of the complexity introduced by the thread-local variables involved in this solution. Consider a more banal solution: change the `ApiClient` logic so that a) header values passed as parameters override default header values, and b) headers with `None` values are not added to the request at all. Then you should be able to just set `params = {"org_id": organization_id, "X-Organization": None}` here and pass it to `get_paginated_collection`.

Both of these changes would technically be breaking, but...
* Making per-request headers override default headers seems like common sense, so I would argue the old behavior is broken;
* The old behavior with `None` would just serialize the header value as "None", which I don't think is a useful value for any headers CVAT supports.
* This way, users of the low-level API (and not just of the high-level API) could also override the default organization.

cvat/apps/engine/views.py
Pr1ncee: As I investigated it's still reachable even when the client only sends valid indexes. The index is validated first in frame_provider.py (line 212) and frame_provider.py (line 221), but after that the server still has to open the backing chunk/media file on disk in frame_provider.py (line 95) and frame_provider.py (line 567). So the failure mode is “valid logical index, missing physical file,” not “bad client index.” That can happen if cached chunks were removed, share/local files were moved or deleted after task creation, or backing storage drifted. But I'm not sure whether it's a bug or not

SpecLad: Yes, if the backing storage gets altered without the server's knowledge, the server can fail opening the data files, however:

1. This would not be a client error, so returning a 4xx response status would be inappropriate.
2. What does this have to do with the current PR? I don't think you've touched any functionality that involves media data.

cvat-sdk/gen/templates/openapi-generator/api_client.mustache
```diff
@@ -183,8 +183,11 @@ class ApiClient(object):
         config = self.configuration
 
         # header parameters
-        header_params = header_params or {}
-        header_params.update(self.get_common_headers())
+        header_params = {
+            **self.get_common_headers(),
+            **(header_params or {}),
+        }
+        header_params = {k: v for k, v in header_params.items() if v is not None}
```
SpecLad, line 190: Make sure to describe the changes made here in the changelog.

cvat-sdk/cvat_sdk/core/proxies/model_proxy.py
```diff
@@ -86,6 +87,31 @@ class Repo(ModelProxy[ModelType, ApiType]):
...
+    if organization_id is None:
+        # Empty org slug is the SDK/server convention for the personal workspace.
+        params = {"org": ""}
+    else:
+        params = {"org_id": organization_id}
+
+    return get_paginated_collection(
+        endpoint,
+        return_json=return_json,
+        _headers=header_overrides,
```
SpecLad: It's odd that you've had to add a custom parameter for this; since `X-Organization` is a parameter that's declared in the schema, you should be able to just pass it as part of the `params` dictionary. Does that not work for some reason?

cvat/apps/engine/views.py
Pr1ncee: Just thought about it - agree about it, I touched what I shouldn't do

cvat-sdk/cvat_sdk/core/client.py
```diff
@@ -131,7 +132,7 @@ def __init__(
         self._repos: dict[str, Repo] = {}
         """A cache for created Repository instances"""
 
-    _ORG_SLUG_HEADER = "X-Organization"
+    _ORG_SLUG_HEADER = _ORG_SLUG_HEADER
```
SpecLad: What's this change about? It doesn't seem to have any impact.

cvat-sdk/cvat_sdk/core/proxies/model_proxy.py
```diff
@@ -86,6 +87,30 @@ class Repo(ModelProxy[ModelType, ApiType]):
...
+    return_json: bool = False,
+    **kwargs,
+):
+    params = {"x_organization": None}
+
+    if organization_id is None:
+        # Empty org slug is the SDK/server convention for the personal workspace.
+        params["org"] = ""
+    else:
+        params["org_id"] = organization_id
```
SpecLad: Why not integrate this functionality directly in `get_paginated_collection`? Seems like it could be useful to all clients.

tests/python/sdk/util.py
```diff
@@ -96,3 +102,79 @@ def restricted_request(self, method, url, *args, **kwargs):
...
+    maintainer_username: str
+    org_slug: str
+    project_id: int
+    task_id: int
+    job_id: int
+    issue_id: int | None = None
+    comment_id: int | None = None
+
+
+def create_org_resource_hierarchy(
```
SpecLad: This sort of thing is usually done using fixtures; any reason you can't do that here?

tests/python/sdk/util.py
```diff
@@ -96,3 +102,79 @@ def restricted_request(self, method, url, *args, **kwargs):
...
+    image_path: Path, *, include_issue: bool = False, include_comment: bool = False
+) -> OrgResourceHierarchy:
+    suffix = uuid4().hex[:6]
```
SpecLad: Why do we need a random suffix? (Also, it seems rather silly to generate an entire UUID when you only need 6 characters...)

changelog.d/20260526_000000_andrei_sdk_iam_fix.md
```diff
@@ -0,0 +1,15 @@
+### Fixed
+
+- \[SDK\] Fixed `Project.get_tasks()`, `Project.get_labels()`, `Task.get_jobs()`,
+  `Task.get_labels()`, `Job.get_issues()`, `Job.get_labels()`, and `Issue.get_comments()`
+  methods returning an empty list when called on organization resources without
+  an explicit organization context set on the client
```
SpecLad: FWIW, the bug would also occur if an explicit organization context _was_ set (to a different organization).

changelog.d/20260526_000000_andrei_sdk_iam_fix.md
```diff
@@ -0,0 +1,15 @@
...
+  (<https://github.com/cvat-ai/cvat/pull/10665>)
+
+- \[SDK\] Fixed the generated API client in
```
SpecLad: Not really a fix - this is a behavior change and should be in the `Changed` section.

changelog.d/20260526_000000_andrei_sdk_iam_fix.md
```diff
@@ -0,0 +1,15 @@
...
+  `cvat-sdk/gen/templates/openapi-generator/api_client.mustache` so per-request
```
SpecLad: This file name doesn't tell anything to the user. Describe the change in behavior, not implementation details.

changelog.d/20260526_000000_andrei_sdk_iam_fix.md
```diff
@@ -0,0 +1,15 @@
...
+  header arguments override default headers, and header arguments with `None`
+  values are omitted instead of being serialized. This allows organization-aware
+  SDK helpers to suppress the default `X-Organization` header while sending
+  requests with `org_id`
```
SpecLad: We don't need to list the impact of the change on ourselves; that's what commit messages are for. The changelog is for the user.

cvat-sdk/cvat_sdk/core/client.py
Pr1ncee: I'm sorry, forgot to remove it after refactoring

SpecLad: Okay, but now you've added even more redundant changes instead of removing the old ones...

cvat-sdk/cvat_sdk/core/helpers.py
```diff
@@ -17,14 +17,28 @@
...
+# Sentinel distinguishing "not passed" from None, which means personal workspace.
+_UNSET = object()
+
 
 def get_paginated_collection(
-    endpoint: Endpoint, *, return_json: bool = False, **kwargs
+    endpoint: Endpoint,
+    *,
+    return_json: bool = False,
+    organization_id: int | None = _UNSET,
```
SpecLad: I would suggest to call this parameter `org_id` (since, when it's a number, it actually becomes the `org_id` query parameter).

changelog.d/20260526_000000_andrei_sdk_iam_fix.md
```diff
@@ -0,0 +1,15 @@
...
+  methods returning an empty list when called on organization resources without
+  an explicit organization context set on the client, or with a different
+  organization context set
+  (<https://github.com/cvat-ai/cvat/pull/10665>)
+
+### Changed
+
+- \[SDK\] Changed requests that explicitly specify an organization by ID being
+  overridden by the client-level `X-Organization` header, causing them to
+  operate in the wrong organization context
```
SpecLad: Changed them _how_?

Also, could this actually cause them to operate in the wrong context? IIRC, we have logic on the server that rejects requests with multiple conflicting organization IDs.

cvat-sdk/cvat_sdk/core/proxies/issues.py
```diff
@@ -52,10 +52,16 @@ class Issue(
     _model_partial_update_arg = "patched_issue_write_request"
 
     def get_comments(self) -> list[Comment]:
+        # IssueRead does not expose `organization`, so we have to look it up
+        # via the parent job. Without an explicit org context the comments
```
zhiltsov-max, line 56: I think the actual problem here is that we have several ways to specify an organization, and they do not cover all the use cases. It should work like this now: 
`?org=<name>` - returns all results from a specific org
`?org=` - returns sandbox results
`?<no org>` - the default, equal to `?org=`

It deviates from the typical filtering logic, when `?<no org>` would return all results without extra filtering, and there's no designed way to return all results. I don't think it should be addressed in this PR, but it certainly deserves a separate server issue, because the implementation here is not the logic our clients should be doing.

site/content/en/docs/api_sdk/sdk/highlevel-api.md
```diff
@@ -234,6 +234,13 @@ with client.organization_context('myorg'):
 # the slug is now reset to its previous value
 ```
 
+Entity methods that retrieve related resources (such as `task.get_jobs()`,
+`task.get_labels()`, `project.get_tasks()`, `project.get_labels()`,
+`job.get_issues()`, `job.get_labels()`, and `issue.get_comments()`)
+automatically apply the correct organization context based on the entity's
+organization. You do not need to set `organization_slug` manually before
+calling these methods.
```
zhiltsov-max: I think this comment is extra, because it just says that the functions behave as the user would expect anyway.

cvat-sdk/cvat_sdk/core/proxies/issues.py
SpecLad: Would you care to create this issue? :slightly_smiling_face: 

I agree in principle, although even if we resolve this on the server, there will still need to be some logic on the client to suppress the global organization context for these child object list requests.

zhiltsov-max: Created https://github.com/cvat-ai/cvat/issues/10776

SpecLad (approved):

PR #10667, opened 2026-05-26
Clearup helm rest api logs

archibald1418:
### Motivation and context

We are having pretty fat debug log from helm rest api runs
https://github.com/cvat-ai/cvat/actions/runs/26415563822/job/77852787260?pr=10664

This change saves debug logs into a file rather than print it in the github runner's stdout

### Checklist

### License

### DIFF
--- a/.github/workflows/comment.yml
+++ b/.github/workflows/comment.yml
@@ -55,7 +55,7 @@ jobs:
 
   run-full:
     needs: verify_author
-    uses: ./.github/workflows/main.yml
+    uses: ./.github/workflows/full.yml
     with:
       ref: ${{ needs.verify_author.outputs.ref }}
 
--- a/.github/workflows/comment.yml
+++ b/.github/workflows/comment.yml
@@ -55,7 +55,7 @@ jobs:
 
   run-full:
     needs: verify_author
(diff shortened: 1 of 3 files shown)


SpecLad:
Is there anything useful in those logs? Maybe we could just remove them entirely?

archibald1418:
Most of the debug logs here are `pytest fixture` logs with `kubectl` cmdlines and db dropping and restoration. Db operations are trivial and any related exceptions are visible from server logs, can be removed. What about `kubectl` lines, do we need to log this?

yaizkazani:
If this is the same case, we've discussed with @archibald1418, there were something like 35000 lines of logs ))) @SpecLad  I think we have to remove them

SpecLad:
Then let's remove them.

.github/workflows/comment.yml
```diff
@@ -55,7 +55,7 @@ jobs:
 
   run-full:
     needs: verify_author
-    uses: ./.github/workflows/full.yml
+    uses: ./.github/workflows/main.yml
```
SpecLad: ?

archibald1418: reverted

SpecLad (approved):

PR #10670, opened 2026-05-26
Load the secret key in a way that doesn't involve code execution

SpecLad:
### Motivation and context

This is a defense-in-depth measure aimed to reduce the damage in case a vulnerability grants an attacker the ability to overwrite arbitrary files. It prevents the attacker from easily executing arbitrary code by overwriting `secret_key.py`.

In addition, this avoids adding `/home/django` to `sys.path`, which is writable by the `django` user, and thus could potentially also be used for ACE attacks. Note that even with this patch, this directory will still be in `sys.path`, as we currently install the CVAT sources in there directory too. I will try to fix that with a followup patch.

This could break instances where `secret_key.py` was manually modified with custom code; however, such modifications were never supported, and better alternatives exist (`/etc/cvat/init.d`, custom settings files).

### Checklist

### License

### DIFF
--- a/cvat/settings/base.py
+++ b/cvat/settings/base.py
@@ -15,8 +15,8 @@ For the full list of settings and their values, see
 https://docs.djangoproject.com/en/2.0/ref/settings/
 """
 
+import ast
 import os
-import sys
 import tempfile
 import urllib
 from datetime import timedelta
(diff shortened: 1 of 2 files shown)


SpecLad:
@zhiltsov-max Could you take a look?

cvat/settings/base.py
```diff
@@ -69,14 +69,43 @@ def generate_secret_key():
...
+    secret_key_path = BASE_DIR / "keys/secret_key.py"
+    module_node = ast.parse(secret_key_path.read_text(), secret_key_path)
+
+    secret_key = None
 
+    for statement_node in module_node.body:
+        error_prefix = f"{secret_key_path}:{statement_node.lineno}: "
+        match statement_node:
+            case ast.Assign(targets=[ast.Name("SECRET_KEY")]):
+                secret_key = ast.literal_eval(statement_node.value)
```
zhiltsov-max, line 89: Why we should allow such a key file to contain anything but the key characters? I understand that there can be backward compatibility questions, but aside of this, here we still allow to execute some expressions. It there a practical need for allowing more than just the string symbols? Or the idea was to avoid writing parsing logic, yet allow different quotes and escape characters?

SpecLad: I'm not sure what you mean; what expressions could this execute?

> Or the idea was to avoid writing parsing logic, yet allow different quotes and escape characters?

It's both really; it should be safe and also allow any possible literal.

zhiltsov-max: > The string or node provided may only consist of the following Python literal structures: strings, bytes, numbers, tuples, lists, dicts, sets, booleans, None and Ellipsis.

https://docs.python.org/3/library/ast.html#ast.literal_eval

The docs also say it can crash the interpreter with malicious inputs.

I just think if we can leave just the key inside the file or simply drop `\s*SECRET_KEY\s*=\s*` and quotes instead.

SpecLad: > > The string or node provided may only consist of the following Python literal structures: strings, bytes, numbers, tuples, lists, dicts, sets, booleans, None and Ellipsis.
> 
> https://docs.python.org/3/library/ast.html#ast.literal_eval

Sure, but these can't cause side effects.

> The docs also say it can crash the interpreter with malicious inputs.

Yeah, but it will also crash the interpreter if we don't find a valid secret key and raise `ImproperlyConfigured`. :slightly_smiling_face: So I don't think it really makes a difference. 

Whereas on the other hand, using `literal_eval` means we don't have to worry about matching all of the varieties of Python string literals; Python just handles it for us.

zhiltsov-max (approved):

PR #10675, opened 2026-05-27
Native function agents now download and process task image chunks incrementally

Pr1ncee:
### Motivation and context
**Problem Statement:**
Before these changes, when the agent receives a task annotation request, it first downloads all the images (in chunks), then iterates through them and runs the model on each one. But it's not the best behavior for the agent: it means the agent needs enough disk space to store all the images (even though they're not all in use at the same time); moreover, the progress bar in the UI hangs at zero for a long time, which isn't very informative.

Now it works like this: the agent downloads the first chunk, then immediately processes it. While it's processing, the agent downloads the next chunk. After processing, the chunk is deleted.

### How has this been tested?
Manual and automated tests.

### Checklist

### License

### DIFF
--- a/cvat-cli/src/cvat_cli/_internal/agent.py
+++ b/cvat-cli/src/cvat_cli/_internal/agent.py
@@ -867,7 +867,7 @@ class _Agent:
         sample_index = 0
 
         with ds.iter_samples(delete_finished_chunks=True) as samples:
-            for sample in samples:
+            for sample_index, sample in enumerate(samples):
                 context = self._create_detection_function_context(ar_params, sample.frame_name)
                 annotations = self._executor.result(
                     self._executor.submit(_worker_job_detect, context, sample.media.load_image())
--- a/cvat-cli/src/cvat_cli/_internal/agent.py
+++ b/cvat-cli/src/cvat_cli/_internal/agent.py
@@ -866,7 +866,7 @@ class _Agent:
 
         sample_index = 0
 
-        with ds.iter_samples(delete_finished_chunks=True) as samples:
+        with ds.iter_samples(temporary_chunks=True) as samples:
             for sample_index, sample in enumerate(samples):
                 context = self._create_detection_function_context(ar_params, sample.frame_name)
                 annotations = self._executor.result(
--- a/cvat-cli/src/cvat_cli/_internal/agent.py
+++ b/cvat-cli/src/cvat_cli/_internal/agent.py
@@ -864,8 +864,6 @@ class _Agent:
 
         all_annotations = models.PatchedLabeledDataRequest(tags=[], shapes=[])
 
-        sample_index = 0
-
         with ds.iter_samples(temporary_chunks=True) as samples:
             for sample_index, sample in enumerate(samples):
                 context = self._create_detection_function_context(ar_params, sample.frame_name)
@@ -877,11 +875,10 @@ class _Agent:
                 all_annotations.tags.extend(tags)
                 all_annotations.shapes.extend(shapes)
 
-                sample_index += 1
                 current_timestamp = datetime.now(tz=timezone.utc)
 
                 if current_timestamp >= last_update_timestamp + _UPDATE_INTERVAL:
-                    self._update_ar(ar_id, sample_index / len(ds.samples))
+                    self._update_ar(ar_id, (sample_index + 1) / len(ds.samples))
                     last_update_timestamp = current_timestamp
 
                 # Interactive requests are time sensitive, so if there are any,
--- a/cvat-cli/src/cvat_cli/_internal/agent.py
+++ b/cvat-cli/src/cvat_cli/_internal/agent.py
@@ -866,8 +866,8 @@ class _Agent:
 
         sample_index = 0
 
-        for chunk_samples in ds.iter_sample_chunks(delete_finished_chunks=True):
-            for sample in chunk_samples:
+        with ds.iter_samples(delete_finished_chunks=True) as samples:
+            for sample in samples:
                 context = self._create_detection_function_context(ar_params, sample.frame_name)
                 annotations = self._executor.result(
                     self._executor.submit(_worker_job_detect, context, sample.media.load_image())
--- a/cvat-cli/src/cvat_cli/_internal/agent.py
+++ b/cvat-cli/src/cvat_cli/_internal/agent.py
@@ -157,7 +157,7 @@ class _TrackingStateContainer:
             raise _BadArError(f"Tracking state {state_id!r} is not for task #{task_id}")
 
         if image_dims != ext_state.original_image_dims:
-            raise _BadArError(f"Image sizes of the start frame and the current frame are different")
+            raise _BadArError("Image sizes of the start frame and the current frame are different")
 
         ext_state.last_accessed_at = datetime.now(tz=timezone.utc)
         self._id_to_ext_state.move_to_end(state_id)
@@ -848,7 +848,12 @@ class _Agent:
         )
 
     def _calculate_result_for_annotate_task_ar(self, ar_id: str, ar_params) -> dict[str, Any]:
-        ds = cvatds.TaskDataset(self._client, ar_params["task"], load_annotations=False)
+        ds = cvatds.TaskDataset(
+            self._client,
+            ar_params["task"],
+            load_annotations=False,
+            media_download_policy=cvatds.MediaDownloadPolicy.FETCH_CHUNKS_ON_DEMAND,
+        )
 
         # Fetching the dataset might take a while, so do a progress update to let the server
         # know we're still alive.
@@ -859,25 +864,29 @@ class _Agent:
 
         all_annotations = models.PatchedLabeledDataRequest(tags=[], shapes=[])
 
-        for sample_index, sample in enumerate(ds.samples):
-            context = self._create_detection_function_context(ar_params, sample.frame_name)
-            annotations = self._executor.result(
-                self._executor.submit(_worker_job_detect, context, sample.media.load_image())
-            )
+        sample_index = 0
 
-            tags, shapes = mapper.validate_and_remap(annotations, sample.frame_index)
-            all_annotations.tags.extend(tags)
-            all_annotations.shapes.extend(shapes)
+        for chunk_samples in ds.iter_sample_chunks(delete_finished_chunks=True):
+            for sample in chunk_samples:
+                context = self._create_detection_function_context(ar_params, sample.frame_name)
+                annotations = self._executor.result(
+                    self._executor.submit(_worker_job_detect, context, sample.media.load_image())
+                )
 
-            current_timestamp = datetime.now(tz=timezone.utc)
+                tags, shapes = mapper.validate_and_remap(annotations, sample.frame_index)
+                all_annotations.tags.extend(tags)
+                all_annotations.shapes.extend(shapes)
 
-            if current_timestamp >= last_update_timestamp + _UPDATE_INTERVAL:
-                self._update_ar(ar_id, (sample_index + 1) / len(ds.samples))
-                last_update_timestamp = current_timestamp
+                sample_index += 1
+                current_timestamp = datetime.now(tz=timezone.utc)
 
-            # Interactive requests are time sensitive, so if there are any,
-            # we have to put the current AR on hold and process them ASAP.
-            self._process_available_ars(REQUEST_CATEGORY_INTERACTIVE)
+                if current_timestamp >= last_update_timestamp + _UPDATE_INTERVAL:
+                    self._update_ar(ar_id, sample_index / len(ds.samples))
+                    last_update_timestamp = current_timestamp
+
+                # Interactive requests are time sensitive, so if there are any,
+                # we have to put the current AR on hold and process them ASAP.
+                self._process_available_ars(REQUEST_CATEGORY_INTERACTIVE)
 
         return {"annotations": all_annotations}
 
--- a/cvat-sdk/cvat_sdk/datasets/common.py
+++ b/cvat-sdk/cvat_sdk/datasets/common.py
@@ -69,5 +69,12 @@ class MediaDownloadPolicy(Enum):
     PRELOAD_ALL = auto()
     """Download and cache all media data when the dataset object is created."""
 
+    FETCH_CHUNKS_ON_DEMAND = auto()
+    """
+    Download image chunks lazily as frames are accessed and cache them locally.
+
+    This policy is only supported for image-set tasks.
+    """
+
     FETCH_FRAMES_ON_DEMAND = auto()
     """Download the media element for each frame whenever MediaElement.load_* is invoked."""
--- a/cvat-sdk/cvat_sdk/datasets/common.py
+++ b/cvat-sdk/cvat_sdk/datasets/common.py
@@ -67,13 +67,19 @@ class MediaDownloadPolicy(Enum):
     """Defines policies controlling when media data is downloaded."""
 
     PRELOAD_ALL = auto()
-    """Download and cache all media data when the dataset object is created."""
+    """
+    Download and cache all media data when the dataset object is created.
+
+    This requires the task's original chunks to be image sets, meaning that each chunk
+    contains individual images. Tasks created from video may still satisfy this requirement.
+    """
 
     FETCH_CHUNKS_ON_DEMAND = auto()
     """
-    Download image chunks lazily as frames are accessed and cache them locally.
+    Download image-set chunks lazily as they are needed and cache them locally.
 
-    This policy is only supported for image-set tasks.
+    This requires the task's original chunks to be image sets, meaning that each chunk
+    contains individual images. Tasks created from video may still satisfy this requirement.
     """
 
     FETCH_FRAMES_ON_DEMAND = auto()
--- a/cvat-sdk/cvat_sdk/datasets/task_dataset.py
+++ b/cvat-sdk/cvat_sdk/datasets/task_dataset.py
@@ -5,7 +5,7 @@
 from __future__ import annotations
 
 import contextlib
-import shutil
+import itertools
 import tempfile
 import zipfile
 from collections.abc import Iterable, Sequence
@@ -13,6 +13,7 @@ from concurrent.futures import ThreadPoolExecutor
 from pathlib import Path
 from typing import Any, Generator, Iterator
 
+import attrs
 import PIL.Image
 
 import cvat_sdk.core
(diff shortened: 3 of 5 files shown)


cvat-sdk/cvat_sdk/datasets/common.py
```diff
@@ -69,5 +69,12 @@ class MediaDownloadPolicy(Enum):
     PRELOAD_ALL = auto()
     """Download and cache all media data when the dataset object is created."""
 
+    FETCH_CHUNKS_ON_DEMAND = auto()
+    """
+    Download image chunks lazily as frames are accessed and cache them locally.
+
+    This policy is only supported for image-set tasks.
```
SpecLad: Seems like this note should also be added to `PRELOAD_ALL`.

cvat-sdk/cvat_sdk/datasets/task_dataset.py
```diff
@@ -189,6 +196,77 @@ def samples(self) -> Sequence[Sample]:
         """
         return self._samples
 
+    def iter_sample_chunks(
```
SpecLad: I don't think it really makes sense to expose a function to iterate over _chunks_. The user can't really do anything with chunks other than iterate over them, so it seems better to directly return an iterator over samples. A sample iterator would also be compatible with `FETCH_FRAMES_ON_DEMAND`.

cvat-sdk/cvat_sdk/datasets/task_dataset.py
```diff
@@ -189,6 +196,77 @@ def samples(self) -> Sequence[Sample]:
...
+        self, *, delete_finished_chunks: bool = False
+    ) -> Iterator[Sequence[Sample]]:
```
SpecLad: This function yields while the threadpool is active, so:

1. The caller should make sure to call `.close()` on the resulting generator if an exception occurs (probably via the use of `contextlib.closing`). Otherwise, the worker thread will be leaked.
2. The return type should use `Generator` instead of `Iterator`, so that the `close` call can typecheck.

cvat-sdk/cvat_sdk/datasets/task_dataset.py
```diff
@@ -189,6 +196,77 @@ def samples(self) -> Sequence[Sample]:
...
+                    next_chunk_future = pool.submit(
+                        self._cache_manager.ensure_chunk,
+                        self._task,
+                        sample_chunks[chunk_offset + 1][0],
+                    )
+
+                yield chunk_samples
+
+                if delete_finished_chunks:
+                    (self._chunk_dir / f"{chunk_index}.zip").unlink(missing_ok=True)
```
SpecLad: This is problematic, because another process may be accessing the same task cache at the same time. If you delete a chunk that the other process expects to be present (e.g. if it used `PRELOAD_ALL`), then it will crash.

IMO, if the user wants the `delete_finished_chunks` behavior, we have to download the chunks into a temporary directory, completely separate from the regular cache.

cvat-sdk/cvat_sdk/datasets/common.py
SpecLad: Also, probably worth clarifying what an "image-set task" is. Note that tasks created from a video can still have image set chunks, so they can be used with this.

cvat-sdk/cvat_sdk/datasets/task_dataset.py
SpecLad: Alternatively, consider returning an `AbstractContextManager[Iterator[Sample]]`, to make sure the client uses `with`.

cvat-cli/src/cvat_cli/_internal/agent.py
```diff
@@ -859,25 +864,29 @@ def _calculate_result_for_annotate_task_ar(self, ar_id: str, ar_params) -> dict[
...
 
-        for sample_index, sample in enumerate(ds.samples):
-            context = self._create_detection_function_context(ar_params, sample.frame_name)
-            annotations = self._executor.result(
-                self._executor.submit(_worker_job_detect, context, sample.media.load_image())
-            )
+        sample_index = 0
+
+        with ds.iter_samples(delete_finished_chunks=True) as samples:
+            for sample in samples:
```
SpecLad:
```suggestion
            for sample_index, sample in enumerate(samples):
```

Now that we're back to iterating over samples, we can use `enumerate` again.

cvat-sdk/cvat_sdk/datasets/task_dataset.py
```diff
@@ -126,22 +148,44 @@ def __init__(
...
+
+        if chunk_dir == self._chunk_dir:
+            self._cache_manager.ensure_chunk(self._task, chunk_index)
+            return
+
+        cache_chunk_path = self._chunk_dir / f"{chunk_index}.zip"
+        if cache_chunk_path.exists():
+            shutil.copyfile(cache_chunk_path, chunk_path)
+            return
+
```
SpecLad: I don't think this logic here is necessary. An agent is always going to fetch chunks on-demand; so it will never have an existing cache of chunks available to copy from. Let's just keep it simple and always download the chunk when it's missing.

cvat-sdk/cvat_sdk/datasets/task_dataset.py
```diff
@@ -189,19 +233,154 @@ def samples(self) -> Sequence[Sample]:
...
+        self, *, delete_finished_chunks: bool = False
+    ) -> Generator[Iterator[Sample], Any, None]:
+        """
+        Returns a context manager yielding an iterator over the task samples.
+
+        When `media_download_policy` is `FETCH_CHUNKS_ON_DEMAND`, the iterator downloads the first
+        image-set chunk, then starts prefetching the next chunk while the caller processes samples
+        from the current chunk.
+
+        If `delete_finished_chunks` is true, chunk files are materialized in a temporary directory
```
SpecLad: I don't think `delete_finished_chunks` fully describes the semantics of this parameter anymore. Maybe something like `temporary_chunks` would be better?

cvat-sdk/cvat_sdk/datasets/task_dataset.py
```diff
@@ -189,19 +233,154 @@ def samples(self) -> Sequence[Sample]:
...
+        This applies only to chunk-based media download policies.
+        """
+
+        if self._media_download_policy == MediaDownloadPolicy.FETCH_FRAMES_ON_DEMAND:
+            yield iter(self._samples)
+            return
+
+        sample_chunks = self._group_samples_by_chunk()
+
+        if delete_finished_chunks:
```
SpecLad: `delete_finished_chunks` doesn't really make sense when combined with `PRELOAD_ALL`, IMO (why would someone preload chunks and then not use them?). So I would suggest to put this code in the `FETCH_CHUNKS_ON_DEMAND` branch, while in the `PRELOAD_ALL` branch you should assert that `delete_finished_chunks` is false.

cvat-sdk/cvat_sdk/datasets/task_dataset.py
```diff
@@ -189,19 +233,154 @@ def samples(self) -> Sequence[Sample]:
...
+
+        with ThreadPoolExecutor(max_workers=1) as pool:
+            yield self._iter_samples_from_chunks(
+                sample_chunks,
+                chunk_dir=self._chunk_dir,
+                delete_finished_chunks=False,
+                download_pool=pool,
+            )
+
+    def _group_samples_by_chunk(self) -> list[tuple[int, list[Sample]]]:
```
SpecLad, line 289: Seems like this function could be simplified a lot with `itertools.groupby`.

cvat-sdk/cvat_sdk/datasets/task_dataset.py
```diff
@@ -189,19 +233,154 @@ def samples(self) -> Sequence[Sample]:
...
+                    sample_chunks[chunk_offset + 1][0],
+                )
+            else:
+                next_chunk_future = None
+
+            for sample in chunk_samples:
+                if chunk_dir == self._chunk_dir:
+                    yield sample
+                else:
+                    yield Sample(
```
SpecLad: You should use `attrs.evolve` to create a modified `Sample`.

cvat-sdk/cvat_sdk/datasets/task_dataset.py
```diff
@@ -189,19 +233,154 @@ def samples(self) -> Sequence[Sample]:
...
+                        frame_index=sample.frame_index,
+                        frame_name=sample.frame_name,
+                        annotations=sample.annotations,
+                        media=self._TaskChunkDirMediaElement(self, sample.frame_index, chunk_dir),
+                    )
+
+            if delete_finished_chunks:
+                (chunk_dir / f"{chunk_index}.zip").unlink(missing_ok=True)
```
SpecLad: The logic for building the chunk file path occurs multiple times in both new and existing code; could you factor it out into a function?

cvat-sdk/cvat_sdk/datasets/task_dataset.py
```diff
@@ -126,22 +148,44 @@ def __init__(
...
-        self._chunk_dir = cache_manager.chunk_dir(task_id)
+        self._chunk_dir = self._cache_manager.chunk_dir(self._task.id)
         self._chunk_dir.mkdir(exist_ok=True, parents=True)
 
+    def _ensure_chunk_in_dir(self, chunk_dir: Path, chunk_index: int) -> None:
+        chunk_path = chunk_dir / f"{chunk_index}.zip"
+        if chunk_path.exists():
+            return
+
+        chunk_dir.mkdir(exist_ok=True, parents=True)
```
SpecLad: This seems redundant; looks like the chunk directory always exists by the time this function is called.

cvat-sdk/cvat_sdk/datasets/task_dataset.py
```diff
@@ -126,22 +148,44 @@ def __init__(
...
+
+        if chunk_dir == self._chunk_dir:
+            self._cache_manager.ensure_chunk(self._task, chunk_index)
+            return
```
SpecLad, line 166: This should probably be moved to the top of the function, since otherwise we'll be doing a duplicate existence check (there is one in this function and one in `CacheManager.ensure_chunk`).

cvat-sdk/cvat_sdk/datasets/task_dataset.py
```diff
@@ -189,19 +233,154 @@ def samples(self) -> Sequence[Sample]:
...
 
-        with zipfile.ZipFile(self._chunk_dir / f"{chunk_index}.zip", "r") as chunk_zip:
+        with zipfile.ZipFile(chunk_dir / f"{chunk_index}.zip", "r") as chunk_zip:
             with chunk_zip.open(chunk_zip.infolist()[member_index]) as chunk_member:
                 image = PIL.Image.open(chunk_member)
                 image.load()
 
         return image
 
+    def _load_frame_image_from_chunk_cache(self, frame_index: int) -> PIL.Image:
```
SpecLad: I find it hard to tell apart `_load_frame_image_from_chunk_cache` and `_load_frame_image_from_cache`, since AFAICS they load the images from the exact same place. Could you give them names that make the difference clearer?

Or actually, do we even need two separate functions? It seems like you can just add the `ensure_chunk` call to `_load_frame_image_from_cache`.

tests/python/sdk/test_datasets.py
```diff
@@ -229,3 +232,151 @@ def test_no_annotations(self):
...
+    def test_iter_samples_works_with_fetch_frames_on_demand(self):
+        dataset = cvatds.TaskDataset(
+            self.client,
+            self.task.id,
+            load_annotations=False,
+            media_download_policy=cvatds.MediaDownloadPolicy.FETCH_FRAMES_ON_DEMAND,
+        )
+
+        with dataset.iter_samples() as samples:
+            actual_frame_indexes = [sample.frame_index for sample in samples]
```
SpecLad: You should probably check that the returned samples have usable media data, as well.

tests/python/sdk/test_datasets.py
```diff
@@ -229,3 +232,151 @@ def test_no_annotations(self):
...
+
+        with pytest.raises(
+            cvatds.UnsupportedDatasetError,
+            match="tasks whose original chunks are image sets",
+        ):
+            cvatds.TaskDataset(
+                self.client,
+                video_task.id,
+                load_annotations=False,
+                media_download_policy=cvatds.MediaDownloadPolicy.FETCH_CHUNKS_ON_DEMAND,
```
SpecLad, line 387: You should check both this policy and `PRELOAD_ALL`.

tests/python/sdk/test_datasets.py
```diff
@@ -229,3 +232,151 @@ def test_no_annotations(self):
...
+
+        chunk_dir = dataset._cache_manager.chunk_dir(self.task.id)
+
+        with dataset.iter_samples(delete_finished_chunks=False) as samples:
+            actual_frame_indexes = [sample.frame_index for sample in samples]
+
+        assert actual_frame_indexes == list(range(self.task.size))
+        assert all((chunk_dir / f"{chunk_index}.zip").exists() for chunk_index in range(4))
+
+    def test_iter_samples_yields_single_chunk(self):
```
SpecLad: Why is this test needed? It doesn't seem like there's any logic specific to single-chunk tasks.

cvat-sdk/cvat_sdk/datasets/task_dataset.py
```diff
@@ -189,19 +233,154 @@ def samples(self) -> Sequence[Sample]:
...
 
+    @contextlib.contextmanager
+    def iter_samples(
+        self, *, delete_finished_chunks: bool = False
+    ) -> Generator[Iterator[Sample], Any, None]:
+        """
+        Returns a context manager yielding an iterator over the task samples.
+
+        When `media_download_policy` is `FETCH_CHUNKS_ON_DEMAND`, the iterator downloads the first
+        image-set chunk, then starts prefetching the next chunk while the caller processes samples
```
SpecLad:
```suggestion
        chunk, then starts prefetching the next chunk while the caller processes samples
```

This is an unnecessary detail here. If we happen to add support for other chunk types later, it'll be one less thing to update.

tests/python/sdk/test_datasets.py
```diff
@@ -229,3 +232,151 @@ def test_no_annotations(self):
             assert actual_image == expected_image
 
             assert sample.annotations is None
+
+    def test_iter_samples_prefetches_and_deletes_finished_chunks(
+        self, monkeypatch: pytest.MonkeyPatch
+    ):
+        cvatds.TaskDataset(
```
SpecLad, line 266: Could you add some comments to explain how this test works? It's pretty complex. For example, it's unclear why this dataset is created.

tests/python/sdk/test_datasets.py
```diff
@@ -229,3 +232,151 @@ def test_no_annotations(self):
...
+        observed_temp_chunk_dirs = set()
+        original_ensure_chunk_in_dir = dataset._ensure_chunk_in_dir
+
+        def wrapped_ensure_chunk_in_dir(chunk_dir, chunk_index):
+            if chunk_dir != dataset._chunk_dir:
+                observed_temp_chunk_dirs.add(chunk_dir)
+
+            if chunk_dir != dataset._chunk_dir and chunk_index == 1:
+                second_chunk_download_started.set()
+                assert allow_second_chunk_download.wait(timeout=5)
```
SpecLad: I would suggest to remove the timeouts; they are liable to break the test if the test runner happens to be slow.

tests/python/sdk/test_datasets.py
```diff
@@ -229,3 +232,151 @@ def test_no_annotations(self):
...
+        monkeypatch.setattr(dataset, "_ensure_chunk_in_dir", wrapped_ensure_chunk_in_dir)
+
+        chunk_dir = dataset._cache_manager.chunk_dir(self.task.id)
+        assert (chunk_dir / "0.zip").exists()
+        assert (chunk_dir / "1.zip").exists()
+
+        with dataset.iter_samples(delete_finished_chunks=True) as samples:
+            first_chunk_samples = [next(samples) for _ in range(3)]
+
+            assert [sample.frame_index for sample in first_chunk_samples] == [0, 1, 2]
```
SpecLad: Could you also add a test to ensure `iter_samples` works properly in the presence of deleted frames?

tests/python/sdk/test_datasets.py
```diff
@@ -150,6 +153,31 @@ def test_deleted_frame(self, media_download_policy: cvatds.MediaDownloadPolicy):
...
+
+        dataset = cvatds.TaskDataset(
+            self.client,
+            self.task.id,
+            load_annotations=False,
+            media_download_policy=media_download_policy,
+        )
+
+        with dataset.iter_samples() as samples:
+            actual_samples = list(samples)
```
SpecLad: I think we should refrain from collecting all samples in a list, because the contract of `iter_samples` implies that only the current sample is valid. Even though it might be safe with the current implementation, I can see potential future changes that might make it not so (e.g. instead of reopening the ZIP file every time, we might hold an open file handle to the current chunk, so accessing frames from any other chunk will not work).

So to ensure the tests will work in the future, we should only test one sample at a time.

cvat-sdk/cvat_sdk/datasets/task_dataset.py
```diff
@@ -189,13 +227,130 @@ def samples(self) -> Sequence[Sample]:
         """
         return self._samples
 
+    @contextlib.contextmanager
+    def iter_samples(
+        self, *, temporary_chunks: bool = False
+    ) -> Generator[Iterator[Sample], Any, None]:
+        """
+        Returns a context manager yielding an iterator over the task samples.
```
SpecLad, line 235: Let's actually add an explicit warning to let users know that they should not keep any references to the previous sample after advancing to the next one (see the comment I left in the tests for reasons why).

cvat-cli/src/cvat_cli/_internal/agent.py
```diff
@@ -859,25 +864,29 @@ def _calculate_result_for_annotate_task_ar(self, ar_id: str, ar_params) -> dict[
...
 
-            tags, shapes = mapper.validate_and_remap(annotations, sample.frame_index)
-            all_annotations.tags.extend(tags)
-            all_annotations.shapes.extend(shapes)
+                tags, shapes = mapper.validate_and_remap(annotations, sample.frame_index)
+                all_annotations.tags.extend(tags)
+                all_annotations.shapes.extend(shapes)
 
-            current_timestamp = datetime.now(tz=timezone.utc)
+                sample_index += 1
```
SpecLad: Since we're now using `enumerate` again, you need to remove this and put the +1 back in the `_update_ar` call.

SpecLad (approved):

PR #10684, opened 2026-05-29
Support throttling for annotations endpoints

bsekachev:
### Checklist

### License

### DIFF
--- a/cvat/apps/engine/tests/test_rest_api.py
+++ b/cvat/apps/engine/tests/test_rest_api.py
@@ -6430,6 +6430,33 @@ class JobAnnotationAPITestCase(ApiTestBase):
     def test_api_v2_jobs_id_annotations_no_auth(self):
         self._run_api_v2_jobs_id_annotations(self.user, self.user, None)
 
+    @override_settings(
+        REST_FRAMEWORK={
+            **settings.REST_FRAMEWORK,
+            "DEFAULT_THROTTLE_RATES": {
+                **settings.REST_FRAMEWORK["DEFAULT_THROTTLE_RATES"],
+                "job_annotations_get": "10/minute",
+            },
+        }
+    )
+    def test_api_v2_jobs_id_annotations_get_is_throttled(self):
+        _, jobs = self._create_task(self.user, self.user)
+        job = jobs[0]
+
+        for _ in range(10):
+            response = self._get_api_v2_jobs_id_data(job["id"], self.user)
+            self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        response = self._get_api_v2_jobs_id_data(job["id"], self.user)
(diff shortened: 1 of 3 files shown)


cvat/apps/engine/views.py
```diff
@@ -173,6 +174,22 @@
...
+    def get_throttles(self):
+        throttles = super().get_throttles()
+        throttle_scope = self.annotations_get_throttle_scope
+        if (
+            self.action == "annotations"
+            and self.request.method == "GET"
+            and throttle_scope in api_settings.DEFAULT_THROTTLE_RATES
+        ):
+            self.throttle_scope = throttle_scope
+            throttles.append(ScopedRateThrottle())
```
SpecLad: Is there any reason why we shouldn't just add `ScopedRateThrottle` to the default list of throttles? It should do nothing when `throttle_scope` is not set.

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -6430,6 +6432,36 @@ def test_api_v2_jobs_id_annotations_somebody(self):
...
+
+        with mock.patch.object(
+            ScopedRateThrottle, "THROTTLE_RATES", api_settings.DEFAULT_THROTTLE_RATES
+        ):
+            for _ in range(10):
+                response = self._get_api_v2_jobs_id_data(job["id"], self.user)
+                self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+            response = self._get_api_v2_jobs_id_data(job["id"], self.user)
+            self.assertEqual(response.status_code, status.HTTP_429_TOO_MANY_REQUESTS)
```
SpecLad: This seems like it would give unstable results based on minor timing variations. I'm frankly not convinced we need to test this at all; maybe it would be enough to test that `throttle_scope` is correctly set on the appropriate view?

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -6430,6 +6432,36 @@ def test_api_v2_jobs_id_annotations_somebody(self):
     def test_api_v2_jobs_id_annotations_no_auth(self):
         self._run_api_v2_jobs_id_annotations(self.user, self.user, None)
 
+    @override_settings(
```
SpecLad: It seems kind of pointless to override the settings if you have to override them again later on the `ScopedRateThrottle` object.

cvat/apps/engine/tests/test_rest_api.py
bsekachev: Not sure it will be so unstable, as minute is quite a lot. 
But generally we may remove them as for me they are mostly useful to make sure throttling works

cvat/apps/engine/views.py
bsekachev: > It should do nothing when throttle_scope is not set.

If true, yes, that would be better

SpecLad (approved):
LGTM.

PR #10688, opened 2026-05-29
Add an ability to customize the minimum user role needed to create organizations

SpecLad:
### Motivation and context

This has been requested by a few customers.

The implementation is straightforward: add a Django setting (initialized from an environment variable), and pass it to OPA as an extra parameter when evaluating the policy.

### How has this been tested?

Manual testing + unit tests.

### Checklist

### License

### DIFF
--- a/cvat/apps/iam/permissions.py
+++ b/cvat/apps/iam/permissions.py
@@ -182,6 +182,7 @@ class OpenPolicyAgentPermission(metaclass=ABCMeta):
                 "scope": self.scope,
                 "auth": self.get_opa_auth_payload(),
                 "resource": self.get_resource(),
+                "settings": self.get_opa_settings_payload(),
             }
         }
 
@@ -206,6 +207,9 @@ class OpenPolicyAgentPermission(metaclass=ABCMeta):
             ),
             "organization_specified": self.org_specified,
         }
+
+    def get_opa_settings_payload(self):
+        return {}
 
     @abstractmethod
     def get_resource(self):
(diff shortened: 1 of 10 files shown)


cvat/apps/iam/permissions.py
```diff
@@ -180,37 +180,36 @@ def get_opa_payload(self):
         return {
             "input": {
                 "scope": self.scope,
-                **self.get_opa_auth_payload(),
-                **self.get_opa_resource_payload(),
+                "auth": self.get_opa_auth_payload(),
+                "resource": self.get_resource(),
+                "settings": self.get_opa_settings_payload(),
```
zhiltsov-max, line 185: I agree with passing settings in the input as a temporary solution, but I think the proper implementation should modify the rules bundle instead, because this information is static and isn't related to the request specifically. I could suggest the following design:

- we add an optional "settings.rego" plugin per app, which can be populated by the server
- when OPA asks the rules bundle, CVAT embeds its configuration to the settings plugins and returns the modified ruleset
- OPA rules check these settings during execution (syntactically, like in this PR now, just without `input.`)

zhiltsov-max (approved):

cvat/apps/iam/permissions.py
SpecLad: I tried something like this in my initial solution. The problem is that since OPA requests the rules bundle from the server, you need to restart the server container to make any changes to the permission settings, which IMO makes testing too complicated to be practical.

PR #10693, opened 2026-06-01
Add attempt and request_duration fields to webhook deliveries

Rwwwrl:
### Title
Run webhook delivery workers through an RQ worker pool

### Motivation and context

An upcoming partner integration drives webhook delivery to ~60 tasks/sec at peak (500 annotators, ~0.12 webhook tasks/sec each). Scaling that with independent `DefaultWorker` processes is expensive — ~300 MB RAM each. A fork-based RQ worker pool shares the post-`django.setup()` memory: ~230 MB resident + ~35 MB per forked worker, so ~20 workers fit in one pod instead of one process each.

This PR lands the backend pieces:

- `cvat_rqworker_pool` management command — subclasses django-rq's `rqworker-pool` and pins the multiprocessing start method via `settings.CVAT_RQ_POOL_MULTIPROCESSING_START_METHOD` (default `fork`). Pinning is a no-op on Python 3.10 where `fork` is already the default, but makes the intent explicit and survives the 3.14 default switch to `forkserver`/`spawn`, which breaks `django.setup()` in workers. Fork safety was audited — `django.setup()` opens no threads/connections pre-fork; all DB/Redis connections are lazy and post-fork.
- `WebhookDelivery` gains `attempt`, `outcome` (`success`/`failure`) and `request_duration` (ms) for delivery observability, exposed via the read serializer and schema. `attempt` is derived from the RQ job's retry config.

Devops wiring (pod command override, `terminationGracePeriodSeconds`, KEDA autoscaling) is intentionally deferred — `cvat_rqworker_pool` is not yet the entrypoint for any worker. The new delivery fields are nullable for existing rows but always populated on new deliveries.

### Checklist

### License


cvat/apps/redis_handler/default_settings.py
```diff
@@ -0,0 +1,12 @@
+# Copyright (C) CVAT.ai Corporation
+#
+# SPDX-License-Identifier: MIT
+import os
+
+CVAT_RQ_POOL_MULTIPROCESSING_START_METHOD = os.environ.get(
+    "CVAT_RQ_POOL_MULTIPROCESSING_START_METHOD",
+    "fork",
```
nmanovic: Do we really need that? It looks like RQ "fork" by default itself.

https://github.com/rq/rq/blob/298f09e68da972a6faa12921bdc63b73a2f7baed/rq/worker_pool.py#L30

cvat/apps/webhooks/models.py
```diff
@@ -91,6 +103,10 @@ class WebhookDelivery(TimestampedModel):
     status_code = models.PositiveIntegerField(null=True, default=None)
     redelivery = models.BooleanField(default=False)
 
+    attempt = models.PositiveIntegerField(null=True)
+    outcome = models.CharField(choices=WebhookDeliveryOutcomeChoice.choices(), null=True)
```
zhiltsov-max: Why do we need this field? It's fully derivable from `status_code`, provides less information, and isn't used anywhere.

cvat/apps/redis_handler/default_settings.py
Rwwwrl: Nice catch

Actually, I was reading only source of pinned 1.16.0 version

As we can see, starts from 2.7 version, it explicitly uses "fork" logic regardless of the set-process method

My custom override is meaningless, once we move to python3.14+ AND upgrade rq to 2.7+ version

Why I still like my solution: that variable adds clarity: we are using fork explicitly, that is not a default method in python3.14+

But agree with removing that and starting using django-rq`s command

cvat/apps/webhooks/models.py
Rwwwrl: Pretty much sugar, to have easier metrics parsing and so on

like:
```
SELECT * FROM ...
WHERE outcome = success / fail
```

vs
```
SELECT * FROM ...
WHERE status_code BETWEEN 200 AND 299
```

Originally I also was trying to add "retry" option, but it led to a lot of code changes, at the end I decided that it did not worth it

Would you like me to remove it?

cvat/apps/redis_handler/utils.py
```diff
@@ -5,8 +5,24 @@
...
+import rq
+
 
 def get_class_from_module(module_path: str | Path, class_name: str) -> type | None:
     module = importlib.import_module(module_path)
     klass = getattr(module, class_name, None)
     return klass
+
+
+def get_current_job_current_attempt() -> int:
```
zhiltsov-max:
```suggestion
def get_current_job_attempt() -> int:
```

Consider renaming

Rwwwrl: updated

cvat/apps/redis_handler/default_settings.py
Rwwwrl: updated. Removed that custom command

cvat/apps/webhooks/models.py
zhiltsov-max: I don't see why we would need this field in this PR. If we had, for instance, some different status after getting a connection error, I'd agree with adding a field, but even then it would probably be some status with an explanation, not a simple boolean.

Rwwwrl: updated: removed that field back

changelog.d/20260601_160615_aleksei_webhook_worker_pool.md
```diff
@@ -0,0 +1,4 @@
+### Added
+
+- Added `attempt`, `outcome` and `request_duration` fields to webhook deliveries
```
zhiltsov-max:
```suggestion
- Added `attempt`, and `request_duration` fields to webhook deliveries
```

cvat/apps/redis_handler/utils.py
```diff
@@ -5,8 +5,24 @@
...
     module = importlib.import_module(module_path)
     klass = getattr(module, class_name, None)
     return klass
+
+
+def get_current_job_attempt() -> int:
+    job = rq.get_current_job()
+
+    if job is None or job.retry_intervals is None or job.retries_left is None:
+        raise ValueError("Current RQ job has no retry configuration")
```
zhiltsov-max: It seems like it would be more logical to raise an error if there's no job and return 1 if there are no retries configured. If you want to support only retries, consider reflecting it in the function name.

cvat/apps/webhooks/models.py
zhiltsov-max, line 1: Please regenerate the test DB assets after the model changes. The instruction is in `tests/python/README.md`, the `dump_test_db.py` part is needed. Clean up unrelated changes, if any.

cvat/apps/redis_handler/utils.py
Rwwwrl: updated

changelog.d/20260601_160615_aleksei_webhook_worker_pool.md
Rwwwrl: updated

cvat/apps/webhooks/models.py
Rwwwrl: updated

zhiltsov-max (approved):

PR #10695, opened 2026-06-01
Fix flaky TestGetAuditEvents test

zhiltsov-max:
### Motivation and context

`TestGetAuditEvents::test_filter_by_project` intermittently failed with:

```
assert event_count["create:job"] == 4
E   assert 2 == 4
```

Audit events are pushed to ClickHouse **asynchronously**. The `setup` fixture's wait loop (`_wait_for_request_ids`) only checked that at least one event matched each filter — `_filter_events` returns a list, used as a truthy bool. The `create:job` filter had neither a request-id nor a count, so the loop broke as soon as the **first** of the 4 job events landed, while the rest were still in flight. The subsequent assertions then saw fewer than 4 `create:job` events.

### How has this been tested?

Carry an expected count per filter and wait until each scope has at least that many events present. `create:job` is now scoped per task (`task_id`) with an expected count of 2 each → 4 total enforced before the tests run.

### Checklist

- [x] I have linked related issues (if any)
- [x] I have updated/created tests where applicable

### DIFF
--- a/tests/python/rest_api/test_analytics.py
+++ b/tests/python/rest_api/test_analytics.py
@@ -117,20 +117,38 @@ class TestGetAuditEvents:
         assert project_request_id is not None
         assert all(t[1] is not None for t in task_ids)
 
+        # Events are pushed to ClickHouse asynchronously,
+        # so we must wait until the expected number of events for every scope is present.
+        # Items: (filters, expected_count)
         event_filters = [
             (
-                (lambda e: json.loads(e["payload"])["request"]["id"], [project_request_id]),
-                ("scope", ["create:project"]),
+                (
+                    (lambda e: json.loads(e["payload"])["request"]["id"], [project_request_id]),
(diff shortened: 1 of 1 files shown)


zhiltsov-max:
@archibald1418, FYI.

zhiltsov-max:
@codex review

tests/python/rest_api/test_analytics.py
```diff
@@ -272,20 +293,30 @@ def test_delete_project(self, api_version: int):
         response = delete_method("admin1", f"projects/{self.project_id}")
         assert response.status_code == HTTPStatus.NO_CONTENT
 
+        request_id = response.headers.get("X-Request-Id")
+        # Items: (filters, expected_count). The cascade delete emits one event per
```
archibald1418, line 297: This should be a typehint
```suggestion
event_filters: list[EventFilter]
```

PR #10700, opened 2026-06-01
Accept .tsv files in the import dataset UI

zhiltsov-max:
Depends on #10510

### Motivation and context

- The import-dataset modal derives the file-picker `accept` attribute from the selected loader's format, so the `Generic TSV` importer accepts `.tsv`; cvat-core `importDataset` allows the `.tsv` extension.
- Added tests for the case with uploading audio annotations from the annotation workspace

Note: `.tsv` import works via annotation upload into a 1D audio task; project dataset import stays ZIP-only, because there is no audio media import support on the server side.

### How has this been tested?

New Cypress test `case_audio_29_import_tsv_annotations` uploads a `.tsv` transcription file into an audio job and asserts the picker accepts `.tsv`, the import succeeds, and the intervals render — passing on the audio stack.

### Checklist

- ~~I submit my changes into the `develop` branch~~ (targets `ap/audio-client`)

### License


cvat-ui/src/components/import-dataset/import-dataset-modal.tsx
```diff
@@ -392,7 +392,13 @@ function ImportDatasetModal(props: StateToProps): JSX.Element {
...
                 listType='text'
                 fileList={file ? [file] : ([] as any[])}
-                accept='.zip,.json,.xml'
+                accept={
+                    selectedLoader?.format
+                        .toLowerCase()
+                        .split(',')
+                        .map((v: string) => `.${v.trim()}`)
+                        .join(',')
+                }
```
zhiltsov-max, line 401: Currently, the browser will display either the single selected extension or "Custom files" if there are more than 1. Would be great to display the full extension list somehow, but it doesn't seem to be possible to do in a compatible way. Maybe the uploading dialog (instead of the OS file selector) should be updated to display the list of supported extensions.

cvat-ui/src/actions/import-actions.ts
```diff
@@ -140,7 +142,12 @@ export const importDatasetAsync = (
                 const relevantInstance = getState().annotation.job.instance;
                 if (relevantInstance && relevantInstance.id === instance.id) {
                     setTimeout(() => {
-                        dispatch(fetchAnnotationsAsync());
+                        if (relevantInstance.mediaType === MediaType.AUDIO) {
+                            // TODO: unify annotation loading for different workspaces
+                            dispatch(loadAudioAnnotationsAsync());
```
bsekachev: `loadAudioAnnotationsAsync` does not exist anymore. Use just `fetchAnnotationsAsync`

zhiltsov-max: I can see it in the `develop` branch: https://github.com/cvat-ai/cvat/blob/51c8c6a9a411171e91beeeb3035f11d4f3c7a9d7/cvat-ui/src/actions/audio-actions.ts#L143. `fetchAnnotationsAsync` doesn't seem to handle audio annotations. Job loading also uses `loadAudioAnnotationsAsync`: https://github.com/cvat-ai/cvat/blob/51c8c6a9a411171e91beeeb3035f11d4f3c7a9d7/cvat-ui/src/actions/annotation-actions.ts#L1082.

bsekachev: There is major update: https://github.com/cvat-ai/cvat/pull/10720

zhiltsov-max: Rebased

bsekachev (approved):

bsekachev:
I believe we have similar issue when exporting (.tsv exported as .zip if custom name has specified)
image

I'll fix it

PR #10701, opened 2026-06-01
Add request timeout in SDK client

zhiltsov-max:
### Motivation

A client request could hang indefinitely when a keep-alive connection was silently dropped (e.g. by a load balancer or NAT): urllib3 reuses pooled connections but sets no read timeout, so the reused socket blocked on `recv()` forever, ending only when an external killer (e.g. `pytest-timeout`) fired. 

### Changes

**SDK**
- The generated `api_client` `Configuration` now enables TCP keepalive (portable `SO_KEEPALIVE`, OS-default timing) and applies a finite default request timeout (`connect=60`, `read=120`). A broken connection is now detected, evicted from the pool, and retried (for idempotent requests) instead of blocking forever.
- New `Config.request_timeout` knob, wired through `Client` to `Configuration.timeout`.
- `Client.wait_for_completion()` gains an optional `timeout` that raises a clear `TimeoutError` instead of polling unbounded. Default `None` preserves existing behaviour.

**Tests**
- Shared `TEST_REQUEST_TIMEOUT` and `DEFAULT_INTERVAL` constants in `shared/utils/config.py`, applied across every test HTTP path: SDK fixtures, the CLI-built client (`build_client`), `make_api_client`, `make_sdk_client`, the raw `requests` helpers, and the startup health poll. A hang now surfaces fast with a clear error, well before the pytest-timeout.
- Added SDK unit tests for the default timeout, `request_timeout`, and the `wait_for_completion` deadline.

### DIFF
--- a/cvat-sdk/cvat_sdk/core/client.py
+++ b/cvat-sdk/cvat_sdk/core/client.py
@@ -11,7 +11,7 @@ from abc import ABCMeta
 from collections.abc import Generator, Sequence
 from contextlib import contextmanager, suppress
 from pathlib import Path
-from time import sleep
+from time import monotonic, sleep
 from typing import Any, TypeVar
 from urllib.parse import urlsplit
 
@@ -49,6 +49,11 @@ class Config:
 
     status_check_period: float = 5
     """Operation status check period, in seconds"""
+
+    request_timeout: float | tuple[float, float] | None = None
+    """Timeout for individual HTTP requests, in seconds. A single value sets the
+    total timeout; a (connect, read) tuple sets them separately. If None, the
+    default of (60, 120) is used."""
 
     allow_unsupported_server: bool = True
     """Allow to use SDK with an unsupported server version. If disabled, raise an exception"""
@@ -120,9 +125,15 @@ class Client:
         self.api_map = CVAT_API_V2(url)
         """Handles server API URL interaction logic"""
 
-        self.api_client = ApiClient(
-            Configuration(host=self.api_map.host, verify_ssl=self.config.verify_ssl)
-        )
+        configuration = Configuration(host=self.api_map.host, verify_ssl=self.config.verify_ssl)
+        if self.config.request_timeout is not None:
+            if isinstance(self.config.request_timeout, tuple):
+                connect, read = self.config.request_timeout
+                configuration.timeout = urllib3.Timeout(connect=connect, read=read)
+            else:
+                configuration.timeout = urllib3.Timeout(total=self.config.request_timeout)
+
+        self.api_client = ApiClient(configuration)
         """Provides low-level access to the CVAT server"""
 
         if check_server_version:
@@ -277,10 +288,13 @@ class Client:
         rq_id: str,
         *,
         status_check_period: int | None = None,
+        timeout: float | None = None,
         log_prefix: str | None = None,
     ) -> tuple[models.Request, urllib3.HTTPResponse]:
         if status_check_period is None:
             status_check_period = self.config.status_check_period
+
+        deadline = None if timeout is None else monotonic() + timeout
 
         while True:
             request, response = self.api_client.requests_api.retrieve(rq_id)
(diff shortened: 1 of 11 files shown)


cvat-sdk/gen/templates/openapi-generator/configuration.mustache
```diff
@@ -330,11 +332,28 @@ class Configuration:
         )
         """Adding retries to override urllib3 default value 3"""
 
+        self.timeout = urllib3.Timeout(connect=60.0, read=120.0)
+        """Default timeout for HTTP requests. A finite read timeout ensures a
+           request blocked on a silently-dropped (e.g. by a load balancer or
+           NAT) keep-alive connection eventually fails - and is retried for
```
SpecLad, line 338: It seems odd to mention these scenarios, and not the obvious one - where the server fails to respond to the request in time.

tests/python/cli/util.py
```diff
@@ -103,16 +104,28 @@ def setup(
...
     ):
         self.tmp_path = tmp_path
         self.stdout = fxt_stdout
         self.host, self.port = BASE_URL.rsplit(":", maxsplit=1)
         self.user = admin_user
         self.password = USER_PASS
+
+        # The CLI builds its own Client (inside build_client), so inject a short
+        # HTTP request timeout via the Config it constructs. This makes a request
+        # blocked on a dead connection fail fast instead of stalling until the
```
SpecLad, line 117: When would a dead connection occur during testing?

tests/python/cli/util.py
```diff
@@ -103,16 +104,28 @@ def setup(
...
+        monkeypatch: pytest.MonkeyPatch,
     ):
         self.tmp_path = tmp_path
         self.stdout = fxt_stdout
         self.host, self.port = BASE_URL.rsplit(":", maxsplit=1)
         self.user = admin_user
         self.password = USER_PASS
+
+        # The CLI builds its own Client (inside build_client), so inject a short
+        # HTTP request timeout via the Config it constructs. This makes a request
```
SpecLad, line 116: Maybe you could add an actual CLI option for this? Seems potentially useful.

cvat-sdk/cvat_sdk/core/client.py
```diff
@@ -50,6 +50,11 @@ class Config:
     status_check_period: float = 5
     """Operation status check period, in seconds"""
 
+    request_timeout: float | tuple[float, float] | None = None
+    """Timeout for individual HTTP requests, in seconds. A single value sets the
+    total timeout; a (connect, read) tuple sets them separately. If None, the
+    default of (60, 120) is used."""
```
SpecLad, line 56: Does this mean there's no way to disable the timeout? It seems like a big omission.

tests/python/cli/util.py
zhiltsov-max: Apparently, we can get them on a DB reset in tests, when our restore SQL script forcefully disconnects all clients. If a keep-alive connection from the SDK pool is reused, we may never get a server response. In tests, this typically results in a test cancelled by a pytest timeout.

cvat-sdk/cvat_sdk/core/client.py
zhiltsov-max: I think it's doable by specifying a big number, but what is the purpose for this?

SpecLad: Well, let's say you have a slow server and you don't care how long your requests take as long as they succeed in the end.

Plus, it just seems weird to go from "no timeout by default" to "timeout by default and no way to go back to the old behavior".

Plus plus, it seems mildly confusing that `None` means different things in `Config.request_timeout` and `Configuration.timeout`.

zhiltsov-max: Ok, I think we can replace `None` with a sentinel to allow disabling the timeout for those who want it. 

I'm not totally sure about the slow server argument - not only it has to be slow, it also has to be customized (directly in nginx) to have the default 60 second timeouts changed, otherwise the slow requests will be cancelled.

PR #10704, opened 2026-06-02
Generalized quality UI

klakhov:
### Motivation and context

This PR adds the community client APIs and UI required to configure quality requirements.

Key changes:

- Add dedicated `cvat-core` quality modules for settings, requirements, reports, and conflicts.
- Add quality requirement CRUD, filtering, inheritance, and effective-value support.
- Add forms for base and child requirements, annotation filters, comparison settings, and attribute rules.
- Improve quality tables and conflict rendering, including tag/shape conflicts with matching object IDs.
- Preserve staged requirement values when opening the edit form.

### Checklist

### License


cvat-core/src/api-implementation.ts
```diff
@@ -32,13 +32,14 @@ import Organization, { Invitation } from './organization';
 import Webhook from './webhook';
 import { ArgumentError } from './exceptions';
 import {
-    AnalyticsEventsFilter, QualityConflictsFilter,
+    AnalyticsEventsFilter, QualityConflictsFilter, QualityRequirementsFilter,
```
bsekachev: I am not sure that generally storing filters in server-response-types is correct as it is about requests, not about responses

cvat-core/src/api.ts
```diff
@@ -432,6 +433,23 @@ function build(): CVATCore {
...
+                requirements: {
+                    async get(filter = {}, aggregate = false) {
+                        const result = await PluginRegistry.apiWrapper(
+                            cvat.analytics.quality.requirements.get,
+                            filter,
+                            aggregate,
+                        );
+                        return result;
+                    },
+                    async create(data = {}) {
```
bsekachev: Shall we have default value here? What is requirement without data?

cvat-core/src/api.ts
```diff
@@ -432,6 +433,23 @@ function build(): CVATCore {
                         return result;
                     },
                 },
+                requirements: {
+                    async get(filter = {}, aggregate = false) {
```
bsekachev: if filter marked as required in index.ts‎, shall we have a default value here?

cvat-core/src/index.ts
```diff
@@ -170,6 +172,15 @@ export default interface CVATCore {
                     aggregate?: boolean,
                 ) => Promise<PaginatedResource<QualitySettings>>;
             };
+            requirements: {
+                get: (
+                    filter: QualityRequirementsFilter,
+                    aggregate?: boolean,
+                ) => Promise<PaginatedResource<QualityRequirement>>;
+                create: (
+                    data: QualityRequirementSaveFields | SerializedQualityRequirementData,
```
bsekachev: Server is probably expects data in a certain format, why we have union here?

cvat-core/src/server-response-types.ts
```diff
@@ -267,6 +267,159 @@ export interface APIQualitySettingsFilter extends APICommonFilterParams {
...
+    'polyline' |
+    'mask' |
+    'polygon' |
+    'ellipse';
+
+export type QualityRequirementMetric = 'accuracy' | 'precision' | 'recall';
+
+export type QualityRequirementPointSizeBase = 'image_size' | 'group_bbox_size';
+
+export type QualityRequirementAttributeComparator = 'exact' | 'levenshtein';
```
bsekachev: Usually we use enums 
Is there a way to use string consts?

cvat-core/src/server-response-types.ts
```diff
@@ -267,6 +267,159 @@ export interface APIQualitySettingsFilter extends APICommonFilterParams {
 
 export type QualitySettingsFilter = Camelized<APIQualitySettingsFilter>;
 
+export type QualityRequirementAnnotationType =
```
bsekachev: imho, file is huge enough. 
Would be better to move quality related types to a dedicated file. 
And dedicated folder for quality features, like

```
src/
   quality/
       index.ts
       quality-report.ts
       quality-settings.ts
       ...
```
This is optional refactoring, but cvat-core is increasing and initial flat structure is not really convenient anymore

cvat-core/src/server-response-types.ts
```diff
@@ -267,6 +267,159 @@ export interface APIQualitySettingsFilter extends APICommonFilterParams {
...
+export interface APIQualityRequirementsFilter extends APICommonFilterParams {
+    id?: number;
+    settings_id?: number;
+    task_id?: number;
+    project_id?: number;
+    annotation_type?: QualityRequirementAnnotationType;
+    enabled?: boolean;
+    created_date?: string;
+    updated_date?: string;
+}
```
bsekachev: All these fields, marked with `?`  in serialized interfaces are actually optional?
Many of them also are marked as nullable

cvat-core/src/quality-requirement.ts
```diff
@@ -0,0 +1,247 @@
...
+        enumerable: false,
+        value: async function implementation(
+            fields: Parameters<typeof QualityRequirement.prototype.save>[0],
+        ): Promise<QualityRequirement> {
+            const data = fieldsToSnakeCase(fields);
+            const result = typeof this.id === 'number' ?
+                await serverProxy.analytics.quality.requirements.update(this.id, data) :
+                await serverProxy.analytics.quality.requirements.create(data);
+
+            return new QualityRequirement({ ...result });
```
bsekachev: This way is aligned with existing approach of saving resources (if we have id, we do update, if we do not, we do create). 

But I believe this approach was initially wrong by design as it creates difficulties in both classes and interfaces APIs.

Thus many fields in classes, constructors, interfaces shall be marked as undefined or null just because in create path we do not specify them. And then in code where these are used, we need additionally check for exact types (not null, not undefined, etc).

Thus I realized that for creation we probably need a dedicated path. A good option I see here is instead of

```
const req = new QualityRequirement(...)
req.save()
```

We need to define a static method, e.g.

```
const req = await QualityRequirement.create(data)
```

This interfaces supports partially initialized SerializedBody, creates object on the server, and returns to already normally initialized from server response QualityRequirement.

What do you think?

cvat-core/src/quality-report.ts
```diff
@@ -41,11 +40,24 @@ export interface QualitySummary {
...
+    requirements: {
+        total: number;
+        enabled: number;
+        completed: number;
+        items: {
+            requirementId: number | null;
+            name: string;
+            metric: string;
+            score: number | null;
+            threshold: number;
```
bsekachev: Should it be `CamelizedV2<SerializedQualityRequirementReportSummaryItem>?`

cvat-core/src/quality-report.ts
```diff
@@ -140,6 +156,18 @@ export default class QualityReport {
...
+                total: this.#summary.requirements.total,
+                enabled: this.#summary.requirements.enabled,
+                completed: this.#summary.requirements.completed,
+                items: this.#summary.requirements.items.map((item) => ({
+                    requirementId: item.requirement_id,
+                    name: item.name,
+                    metric: item.metric,
+                    score: item.score,
+                    threshold: item.threshold,
+                })),
```
bsekachev: Shall we re-use fieldsToCamelCase?

cvat-ui/src/components/quality-control/quality-control-page.tsx
```diff
@@ -206,6 +209,9 @@ function QualityControlPage(): JSX.Element {
     const requestedInstanceID = useInstanceId(requestedInstanceType);
 
     const { instance } = state;
+    const pluginTabs = usePlugins((combinedState: CombinedState) => (
+        combinedState.plugins.components.qualityControlPage.tabs.items
+    ), { instance });
```
bsekachev: `usePlugins` accepts props as first argument and state as the second argument. 
Strictly speaking, instance is a state, not prop. And we should pass whole state to this hook

cvat-ui/src/components/quality-control/quality-control-page.tsx
```diff
@@ -351,12 +361,23 @@ function QualityControlPage(): JSX.Element {
         }
     }, [state.gtJobInstance]);
 
+    const onJobUpdate = useCallback((job: Job, fields: Parameters<Job['save']>[0]): void => {
+        reduxDispatch(updateJobAsync(job, fields))
+            .then(() => receiveInstance(requestedInstanceType, requestedInstanceID))
+            .catch((error: unknown) => {
+                notification.error({
```
bsekachev, line 404: We probably do not need two notifications when receiveInstance is failed:

`Could not receive requested ${type}` + 'Could not update job'

cvat-ui/src/components/quality-control/quality-control-page.tsx
```diff
@@ -428,61 +449,108 @@ function QualityControlPage(): JSX.Element {
...
                         setQualitySettings={onSaveQualitySettings}
+                        refreshQualitySettings={refreshQualitySettings}
+                    />
+                );
+            }
+
+            return null;
+        };
+
+        if (pluginTabs.length) {
```
bsekachev: Perhaps the condition is useless. 
The code below works normally even for empty array

cvat-ui/src/components/quality-control/quality-control-page.tsx
```diff
@@ -428,61 +449,108 @@ function QualityControlPage(): JSX.Element {
...
+            const sortedPluginTabs = [...pluginTabs].sort((first, second) => first.weight - second.weight);
+            for (const { component: Component } of sortedPluginTabs) {
+                const key = Component.tabKey;
+                const label = Component.tabLabel;
+                if (typeof key === 'string' && typeof label === 'string') {
```
bsekachev: Looks a little weird
As a component in current design is just a function (which can be a component code, or return just menu item), can we do the same and return tab item instead of assignment implicit `tabKey`, `tabLabel`?

cvat-ui/src/components/quality-control/quality-control-page.tsx
```diff
@@ -428,61 +449,108 @@ function QualityControlPage(): JSX.Element {
...
         const isTaskWithGT = instance instanceof Task && gtJobInstance && gtJobMeta;
         const isProject = instance instanceof Project;
 
-        if (isTaskWithGT && validationLayout && qualitySettings) {
-            tabsItems.push({
-                key: 'management',
-                label: 'Management',
-                children: (
+        const renderDefaultManagementTab = (): JSX.Element | null => {
+            if (isTaskWithGT && validationLayout && qualitySettings) {
```
bsekachev: Just saw that `isTaskWithGT` in runtime is initialized as `FramesMetaData`

cvat-ui/src/components/quality-control/quality-control-page.tsx
```diff
@@ -428,61 +449,108 @@ function QualityControlPage(): JSX.Element {
...
+            });
+        }
+
+        if (!pluginTabs.length && (isTaskWithGT || isProject)) {
+            tabsItems.push({
+                key: 'settings',
+                label: 'Settings',
+                children: renderDefaultSettingsTab(),
+            });
+        }
```
bsekachev: Because we pass `renderDefaultManagementTab` to plugin component the logic now seems somehow complicated to understand. 
It would be better to have it implemented via fallback logic, e.g.:

const SettingsTab = pluginTabs.find((tab) => tab.key === 'settings') ?? defaultSettingsTab; 

etc.

Then:

```
if (isTaskWithGT || isProject) {
   // push item
}

Similar refactoring for Requirements / Management. 

Additionally, I do not really understand why we expect `Settings` tab from the plugin as it is the same for community/enterprise versions.

File: cvat-core/src/api-implementation.ts (hunk 1 above)
Comment (klakhov) [reply cvat-core/src/api-implementation.ts]: Refactored

File: cvat-core/src/server-response-types.ts (hunk 5 above)
Comment (klakhov) [reply cvat-core/src/server-response-types.ts]: Converted to enums

File: cvat-core/src/server-response-types.ts (hunk 6 above)
Comment (klakhov) [reply cvat-core/src/server-response-types.ts]: Applied refactoring

File: cvat-core/src/server-response-types.ts (hunk 7 above)
Comment (klakhov) [reply cvat-core/src/server-response-types.ts]: Updated those structures with way less `?`

File: cvat-core/src/quality-requirement.ts (hunk 8 above)
Comment (klakhov) [reply cvat-core/src/quality-requirement.ts]: Good thought. Implemented this approach.
Probably makes sense to refactor other entities, but out of scope of this PR.

File: cvat-ui/src/components/quality-control/quality-control-page.tsx (hunk 12 above)
Comment (klakhov) [reply cvat-ui/src/components/quality-control/quality-control-page.tsx:404]: Removed second error in that case

File: cvat-ui/src/components/quality-control/quality-control-page.tsx (hunk 14 above)
Comment (klakhov) [reply cvat-ui/src/components/quality-control/quality-control-page.tsx]: Removed this.

File: cvat-ui/src/components/quality-control/quality-control-page.tsx (hunk 15 above)
Comment (klakhov) [reply cvat-ui/src/components/quality-control/quality-control-page.tsx]: Updated, now its boolean

File: cvat-ui/src/components/quality-control/quality-control-page.tsx (hunk 16 above)
Comment (klakhov) [reply cvat-ui/src/components/quality-control/quality-control-page.tsx]: Removed this logic. We dont really have plugin settings and plugin management tabs(in theory we could have, and we will need to override, but this is overcomplication)

**Comment by nmanovic** [comment]
@codex review

**Comment by nmanovic** [comment]
@claude review

File: cvat/apps/quality_control/migrations/0012_qualityrequirement_and_more.py [hunk 17]
```diff
@@ -0,0 +1,192 @@
...
+            ],
+        ),
+        migrations.RemoveConstraint(
+            model_name="qualitysettings",
+            name="quality_settings_task_or_project",
+        ),
+        migrations.RemoveField(
+            model_name="qualitysettings",
+            name="check_covered_annotations",
+        ),
```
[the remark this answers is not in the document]
klakhov: Will resolve as its not relevant to UI changes. Same comment is in server PR.

cvat/apps/quality_control/quality_handlers.py
```diff
@@ -0,0 +1,1252 @@
...
+                        "outside", element_visibility == dm.Points.Visibility.absent
+                    )
+                    element_attrs.setdefault(
+                        "occluded", element_visibility == dm.Points.Visibility.hidden
+                    )
+
+                element_attrs[RequirementJsonLogicFilter.PARENT_SKELETON_CONTEXT_KEY] = (
+                    parent_skeleton_context
+                )
+                flattened_annotations.append(element.wrap(attributes=element_attrs))
```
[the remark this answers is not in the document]
klakhov: Will resolve as its not relevant to UI changes. Same comment is in server PR.

bsekachev (approved):
UI lgtm

cvat-ui/src/components/header/header.tsx
```diff
@@ -399,6 +340,40 @@ function HeaderComponent(props: Props): JSX.Element {
...
+                <a href={GUIDE_URL} target='_blank' rel='noopener noreferrer'>
+                    <FileTextOutlined />
+                    <span>Documentation</span>
+                    <ExportOutlined />
+                </a>
+                <a href={GITHUB_URL} target='_blank' rel='noopener noreferrer'>
+                    <GithubOutlined />
+                    <span>GitHub</span>
+                    <ExportOutlined />
+                </a>
```
bsekachev: I do not think these changes shall be in this pull request

cvat-core/src/quality/quality-settings.ts
zhiltsov-max, line 1: I can see 2 section headers for requirements (apparently, for the section and for the table):

Image

It looks redundant, do we really need both?

zhiltsov-max, line 1: If I try to rename a base requirement for rectangles, I'm getting the error:

```
Uncaught (in promise) f: **point_size**: Root requirements cannot inherit.
```

I think the UI could clean up the irrelevant fields from the request, doing so should fix the error. But the server could also implement such filtering.

zhiltsov-max, line 1: The filter field and control buttons can overlap:

Image

When the cursor is over the filter or a button. But it doesn't work the same way for the filter cleanup button and the requirement name field. Probably, they should behave uniformly, if possible.

zhiltsov-max, line 1: Shouldn't the Leventshtein comparator have a configurable threshold?

Image


Upd: found it under the "+" button on the left side in the row. I think we should use a more recognizable button, because currently it also associates with adding a new row (even more confusing to have a similar button to add a row and expand a row). Maybe I'd propose using a `>` icon instead (like we have for annotation attributes in the annotation editing) or `v` (like we have in the requirements tree).

zhiltsov-max, line 1: Image

Seems to always show toggled off, despite the server responses.

Also, the "Match groups" checkbox nearby also is either shown as toggled off, or as overridden and toggled off, even when it should be turned on.

Also, the option is applicable to tags as well, but the UI doesn't show it for tag requirements.

cvat-ui/src/components/quality-control/shared/settings/quality-settings-form.tsx
zhiltsov-max, line 1: Image

Would it be possible to:
- move each toggle before the corresponding tweaks (e.g. match groups before group match threshold, check covered before visibility threshold)
- extract each such group into a subsection

? I think it would make the configuration more clear to the users. Currently, it's not that obvious what is affected by an individual tweak, e.g. visibility threshold). It can be found in the description tooltips, but having some visual organization for them would make it obvious without reading.

Like this:

Image

cvat-ui/src/cvat-core-wrapper.ts
zhiltsov-max, line 1: Image

It seems that the UI handles annotation ids partially incorrectly when drawing conflict markers. In this example, the tag and the box have the same object id and job id, and the tag conflict has its marker drawn next to the bbox. The tag and the marker are correctly highlighted when the mouse is over the conflict in the list, but the box is also highlighted (incorrectly).

cvat-core/src/quality/quality-settings.ts
klakhov: Removed, lets keep only table title

klakhov: I think by design control buttons should always be on top. Applied.

klakhov: There was a bug for checkbox setting. Fixed it.

cvat-ui/src/components/quality-control/shared/settings/quality-settings-form.tsx
klakhov: Implemented, added dividers to the setting groups

cvat-ui/src/cvat-core-wrapper.ts
klakhov: Yes, there was a bug with id handling without checking object type. Fixed

cvat-core/src/quality/quality-settings.ts
klakhov: Excessive fields were sent to the server, added filtering for fields. Base req should be updated normally now

cvat-ui/src/components/quality-control/shared/requirements/quality-requirement-form.tsx
zhiltsov-max, line 1: If you switch "enabled" for a requirement on the tree view and then click the edit button for the requirement, the edit page will continue showing the "enabled" as it was before switching. Seems like a bug.

cvat-ui/src/components/annotation-page/review/issues-aggregator.tsx
zhiltsov-max, line 1: I got the conflicts list under the Appearance section header:

Image

zhiltsov-max, line 1: I guess we'll need to update the conflict list design and improve their presentation anyway, but if it doesn't require big changes, I'd ask to check these:
1. Now we can have skeleton keypoint-level conflicts (currently reported with `shape_type=points`, which looks a little bit confusing, @Grigorii777), they are not displayed. It would be nice to support them.
2. Now we can get more detailed errors, e.g. related to specific attributes, a single shape can have several conflicts, and the conflicts are related to different shape types and requirements. I think it would be nice to allow filtering the list somehow (maybe, with the same frame filter, as usual), but this is likely to be a larger feature. What I'd like to see now is a shape type tag in the list of conflicts, to be able to find conflicts for a specific shape type.

zhiltsov-max, line 1: `Items` is likely only counts the issues, but when there are conflicts, it looks a little bit confusing.

Image

cvat-ui/src/components/quality-control/shared/requirements/quality-requirement-form.tsx
klakhov: Fixed

cvat-ui/src/components/annotation-page/review/issues-aggregator.tsx
klakhov: This sound like a feature that needs to be designed.

Grigorii777: shape_type=points is expected here because a skeleton keypoint is represented as a child points shape in CVAT
skeleton_keypoint is a quality requirement annotation type, not a shape type
 The UI should probably look through skeleton elements when resolving the conflict annotation ID
  If the list needs to distinguish standalone points from skeleton keypoints, we would need separate requirement/parent metadata rather than changing shape_type.

Grigorii777: You mean that you can’t tell the difference between a normal points shape and a skeletal keypoint based on the shape_type, don’t you?

zhiltsov-max: So far my point is that we can have conflicts in the list that are not displayed on the canvas. But I agree that displaying them can require more designing than desired now.

PR #10729, opened 2026-06-09
feat(coco): accept COCO imports without iscrowd field

Pr1ncee:
Inject a default iscrowd=0 into COCO annotation files that omit the field, so datasets exported by third-party tools (which often skip iscrowd) can be imported without errors.

### Motivation and context
Currently, CVAT only accepts COCO annotations if they contain the "iscrowd" parameter; annotations downloaded from Azure do not have this parameter. We need to modify the application to upload annotations to CVAT without modifying the annotation file.

### How has this been tested?
Manual and automated tests

### Checklist

### License


cvat/apps/dataset_manager/formats/utils.py
```diff
@@ -3,13 +3,66 @@
...
+        return
+
+    annotations = annotation_data.get("annotations")
+    if not isinstance(annotations, list):
+        return
+
+    updated = False
+    for annotation in annotations:
+        if isinstance(annotation, dict) and "iscrowd" not in annotation:
+            annotation["iscrowd"] = 0
```
zhiltsov-max: This can result in incorrect parsing, if there is a non-empty `segmentation` field.

zhiltsov-max:
I think it might be better to add an import option on the Datumaro side instead - these files are going to be read in other places, not only in CVAT import.

cvat/apps/dataset_manager/formats/utils.py
```diff
@@ -2,13 +2,35 @@
...
+
+
+def import_coco_dataset(path: str, format_name: str, *, env, default_iscrowd: int):
+    if not supports_default_iscrowd():
+        logger.warning(
+            "The installed Datumaro version does not support the 'default_iscrowd' "
+            "COCO import option. COCO annotations without the 'iscrowd' field will "
+            "fail until Datumaro is upgraded."
+        )
+        return StreamDataset.import_from(path, format_name, env=env)
```
zhiltsov-max: What is the purpose of this check?

Pr1ncee: If Datumaro is updated, CVAT calls with default_iscrowd=0 so COCO annotations missing iscrowd are accepted. If Datumaro is still old pinned version passing default_iscrowd will crash so CVAT falls back to the old import call without that option

zhiltsov-max: Why Datumaro can be old?

Pr1ncee: As I investigated CI and server uses Datumaro version pinned in `cvat/requirements/base.in` and `base.txt` not sibling Datumaro checkout. I just thought I can merge https://github.com/cvat-ai/datumaro/pull/138 and update the datumaro commit in this PR, therefore that logic won't be needed anymore

cvat/requirements/base.txt
zhiltsov-max, line 1: Updating the in file typically results in a changed hash. How did you update the `.txt` file?

cvat/apps/engine/tests/test_rest_api.py
zhiltsov-max, line 1: The added tests duplicate each other. Why do we need all of them?

changelog.d/coco-import-without-iscrowd.md
```diff
@@ -0,0 +1,4 @@
+### Fixed
```
zhiltsov-max: It can't be considered a fix, because it wasn't broken to start with. The COCO [specification](https://cocodataset.org/#format-data) doesn't say anything about that the field can be missing.

```suggestion
### Changed
```

cvat/apps/dataset_manager/tests/test_formats.py
```diff
@@ -1054,3 +1055,55 @@ def test_can_import_mots_annotations_with_split_masks(self):
...
+                    json.dumps(self._make_coco_annotation_without_iscrowd()),
+                )
+
+            dm.task.import_task_annotations(zip_path, task["id"], "COCO 1.0", True)
+
+            task_ann = TaskAnnotation(task["id"])
+            task_ann.init_from_db()
+            self.assertEqual(1, len(task_ann.ir_data.shapes))
+
+    def test_can_import_coco_instances_without_iscrowd_json(self):
```
zhiltsov-max: I think these tests can be deduplicated via parametrization: `['with iscrowd', 'without iscrowd'] x ['coco_instances', 'coco_keypoints']`, if you want to cover all these cases. Supporting both `zip` and `json`-focused tests also doesn't seem reasonable to me, as the functionality is orthogonal. But if needed, it can also be a parametrization axis.

cvat/apps/dataset_manager/tests/test_formats.py
```diff
@@ -1054,3 +1055,55 @@ def test_can_import_mots_annotations_with_split_masks(self):
...
+            "images": [{"id": 1, "file_name": "image_0.jpg", "height": 100, "width": 100}],
+            "annotations": [
+                {
+                    "id": 1,
+                    "image_id": 1,
+                    "category_id": 1,
+                    "segmentation": [[10.0, 10.0, 20.0, 10.0, 20.0, 20.0, 10.0, 20.0]],
+                    "bbox": [10.0, 10.0, 10.0, 10.0],
+                    "area": 100.0,
+                    # No "iscrowd" field — simulates Azure-sourced annotations
```
zhiltsov-max: Is the problem with Azure import only in the `iscrowd` field? If the field is missing, I guess `segmentation` can be missing too.

Upd: Datumaro supports missing segmentation, I checked.

cvat/apps/dataset_manager/tests/test_formats.py
```diff
@@ -1054,3 +1056,62 @@ def test_can_import_mots_annotations_with_split_masks(self):
...
+        return {
+            "info": {},
+            "licenses": [],
+            "images": [{"id": 1, "file_name": "image_0.jpg", "height": 100, "width": 100}],
+            "annotations": [annotation],
+            "categories": [category],
+        }
+
+    @pytest.mark.parametrize("has_segmentation", [True, False])
+    @pytest.mark.parametrize(
```
zhiltsov-max: We don't have pytest in these tests. Unittest has `self.subTest` for nested tests.

cvat/apps/dataset_manager/tests/test_formats.py
```diff
@@ -1054,3 +1056,62 @@ def test_can_import_mots_annotations_with_split_masks(self):
 
             dm.task.import_task_annotations(dataset_path, task["id"], format_name, True)
             self._test_can_import_annotations(task, format_name)
+
+    def _make_coco_annotation_without_iscrowd(self, format_name="COCO 1.0", has_segmentation=True):
```
zhiltsov-max: Please use type annotations in the new/updated code.

cvat/apps/dataset_manager/tests/test_formats.py
```diff
@@ -1054,3 +1055,67 @@ def test_can_import_mots_annotations_with_split_masks(self):
 
             dm.task.import_task_annotations(dataset_path, task["id"], format_name, True)
             self._test_can_import_annotations(task, format_name)
+
+    def _make_coco_annotation_without_iscrowd(
+        self, format_name: str = "COCO 1.0", has_segmentation: bool = True
+    ) -> dict:
+        annotation: dict = {
```
zhiltsov-max: We typically only add annotations to regular variables if the type inference produces too shallow, confusing, or invalid annotations. By default, we annotate only functions and classes. These `dict` annotations for variables don't add anything to what the IDE type inference would do automatically, so I suggest keeping only necessary annotations.

cvat/apps/dataset_manager/tests/test_formats.py
```diff
@@ -1054,3 +1093,84 @@ def test_can_import_mots_annotations_with_split_masks(self):
...
+
+    def test_can_import_coco_without_iscrowd(self) -> None:
+        cases: list[tuple[bool, str, str | None]] = [
+            (True, "COCO 1.0", "annotations/instances_default.json"),
+            (False, "COCO 1.0", None),
+            (True, "COCO Keypoints 1.0", "annotations/person_keypoints_default.json"),
+            (False, "COCO Keypoints 1.0", None),
+        ]
+        for use_zip, format_name, zip_annotation_filename in cases:
+            for has_segmentation in (True, False):
```
zhiltsov-max: Why these cases are not included in the loop above?

cvat/apps/dataset_manager/formats/utils.py
andreistrongin-3yhealth: That fallback logic was removed — `require_iscrowd=False` is now passed directly to Datumaro in all code paths.

cvat/requirements/base.txt
andreistrongin-3yhealth: Regenerated it following `cvat/requirements/README.txt` after updating the Datumaro commit hash in `base.in`.

cvat/apps/engine/tests/test_rest_api.py
andreistrongin-3yhealth: Done — the duplicated tests were removed.

cvat/apps/dataset_manager/tests/test_formats.py
andreistrongin-3yhealth: Updated — the test now uses a `cases` list × inner `has_segmentation` loop with `self.subTest`, covering all combinations.

cvat/apps/dataset_manager/tests/test_formats.py
andreistrongin-3yhealth: `has_segmentation=False` cases are covered in the test as well.

cvat/apps/dataset_manager/tests/test_formats.py
andreistrongin-3yhealth: Fixed — switched to `self.subTest`.

cvat/apps/dataset_manager/tests/test_formats.py
andreistrongin-3yhealth: Done.

cvat/apps/dataset_manager/tests/test_formats.py
andreistrongin-3yhealth: Done — removed variable annotations.

cvat/apps/dataset_manager/tests/test_formats.py
Pr1ncee: has_segmentation is orthogonal to the other three parameters — every (use_zip, format_name, zip_annotation_filename) combination should be tested both with and without it. A nested loop produces the full cross-product (4 × 2 = 8 subtests) without duplicating entries in cases.

zhiltsov-max: I think this loop with cases is supposed to include all the cases, so please move it there. You can use `itertools.product` to make it clear. The same is for `use_zip`, which is also orthogonal.

zhiltsov-max (approved):

PR #10734, opened 2026-06-10
Use RQ callbacks for export and backup webhooks

Rwwwrl:
### Title
Use RQ callbacks for export and backup webhooks

### Motivation and context

Export and backup webhooks were emitted from custom worker-side success/failure
signal handling, duplicating RQ lifecycle behavior. Route these notifications
through RQ `on_success` / `on_failure` callbacks instead: export and backup jobs
now enqueue a small webhook fan-out task on the webhooks queue, and that task
builds the existing `create:export` / `create:backup` payload from the request
id. Lock-acquisition retries still do not emit terminal failure webhooks.

### Checklist

### License


cvat/apps/engine/utils.py
zhiltsov-max, line 1: Could you reduce code duplication please? Functions here have the same implemetation.

Rwwwrl: updated

cvat/apps/engine/tests/test_utils.py
zhiltsov-max, line 1: I don't think we need any extra tests in this file except those in the `TestRequestCompletionWebhookCallbacks` class. The rest is already checked in `TestWebhookExportEvents` and `TestWebhookBackupEvents`. The file should also be renamed, because it only tests webhooks, which are not `engine` "utils" exactly.

cvat/apps/engine/utils.py
zhiltsov-max, line 1: I think this new logic should be moved either to the webhooks app or at least into a separate file. In general, I'd consider adding 2 signals ("request_completed", "request_failed") and registering the completion callbacks just to send the signals as a more general solution.

Rwwwrl: updated. Moved to django signals approach:
1. AbstractExporter sends either request_succeeded or request_failed signals
2. handler for those signals are in cvat/apps/webhooks/event_handlers.py - request_completed_event_handler

cvat/apps/engine/tests/test_utils.py
Rwwwrl: removed the tests. Sticked with existing classes - thank you

cvat/apps/webhooks/event_handlers.py
```diff
@@ -1,80 +1,61 @@
-# Copyright (C) CVAT.ai Corporation
-#
-# SPDX-License-Identifier: MIT
```
zhiltsov-max, line 3: Please keep the license header.

cvat/apps/webhooks/apps.py
```diff
@@ -21,4 +21,5 @@ def ready(self):
 
         load_app_iam_rules(self)
 
-        from . import event_handlers, signals  # pylint: disable=unused-import
+        import cvat.apps.webhooks.event_handlers  # pylint: disable=unused-import
+        import cvat.apps.webhooks.signals  # pylint: disable=unused-import
```
zhiltsov-max: Why do we need this change?

cvat/apps/webhooks/event_handlers.py
```diff
@@ -1,80 +1,61 @@
...
+@receiver(request_failed)
+def request_completed_event_handler(
     sender: Any,
-    target: Project | Task | Job,
-    dst_format: str,
-    status: ExportStatus,
-    message: str = "",
-    **kwargs: Any,
+    request_id: str,
+    status: Literal["succeeded", "failed"],
```
zhiltsov-max: Please use a enum

cvat/apps/webhooks/event_handlers.py
Rwwwrl: I wrote the file manually - that is why I forgot that. Nice to have linter for those cases actually

cvat/apps/webhooks/apps.py
Rwwwrl: nvm. I did it manually, thought there is a profit in how linter classify that. Double checked that - no profit actually, returned back

cvat/apps/engine/signals.py
```diff
@@ -23,6 +23,9 @@
 # analytics feature is available. Now the log system can write information
 # into a file inside removed directory.
 
+request_succeeded = Signal()
+request_failed = Signal()
```
zhiltsov-max, line 27: I think these definitions should belong to the `redis_handler` app.

cvat/apps/engine/utils.py
```diff
@@ -30,15 +30,71 @@
...
+        return
+
+    _ = signals.request_failed.send_robust(
+        sender=AbstractExporter,
+        request_id=rq_job.id,
+        status="failed",
+        message=parse_exception_message(
+            "".join(traceback.format_exception_only(exc_type, exc_value))
+        ),
+    )
```
zhiltsov-max, line 103: Please move to `redis_handler`.

cvat/utils/enums.py
```diff
@@ -0,0 +1,9 @@
+from enum import Enum
+
+
+class StrEnum(str, Enum):
```
zhiltsov-max: Why not just use the standard library `StrEnum`? It's added in 3.11 and we are at 3.12 now, so we can just use it. We also use Django's `TextChoices` for the same purposes.

https://docs.python.org/3/library/enum.html#enum.StrEnum

cvat/apps/engine/background.py
```diff
@@ -68,11 +90,193 @@
...
+    def _init_callback_with_params(self):
+        """
+        Private method that should initialize callback function with its args/kwargs
+        like the init_callback_with_params method in the parent class.
+        """
+
+    def init_callback_with_params(self):
+        """
+        Method should not be overridden
+        """
```
zhiltsov-max, line 224:
```suggestion
        """
        Method should not be overridden, override 
        `_init_callback_with_params()` instead.
        """
```

Rwwwrl: updated

cvat/utils/enums.py
Rwwwrl: Double checked that we do not use StrEnum in non-server context (so we do not have it in python3.10 context) -> removed StrEnum and updated all imports to use stdin enum instead

tests/python/rest_api/test_webhooks_sender.py
```diff
@@ -824,6 +824,6 @@ def test_webhook_create_backup_for_task(self, tasks: Container) -> None:
 
         _, payload = get_deliveries(webhook_id)
         assert payload["event"] == "create:backup"
-        assert payload["status"] == "completed"
+        assert payload["status"] == "succeeded"
```
zhiltsov-max, line 827: Why do we need to change the enum value? I think it can break some users for no reason.

zhiltsov-max (approved):

PR #10770, opened 2026-06-12
Fix 10503

### LINKED ISSUE #10503: `saveAnnotationsAsync` can create duplicate shapes when the save pipeline throws after the server PATCH succeeds

prathamesh-patil-5090:
### Motivation and context

Fixes https://github.com/cvat-ai/cvat/issues/10503

When `action=create` succeeds on the server but the client loses the HTTP response (typical proxy/ASGI timeouts on large saves), local shapes keep no server IDs. The next save re-submits them as creates and produces coordinate-identical duplicates.

Reconciliation-by-geometry (already on `develop`, and previously extended in this PR) is only a workaround: it guesses whether create finished and matches objects heuristically. It does not make create safe to retry.

This PR fixes the root cause by making annotation creates **idempotent**:

1. Annotation objects (`shapes`, `tracks`, `tags`, `intervals`) get an optional client-generated `uuid`
2. The server stores `uuid` with a unique constraint per `(job, uuid)`
3. On `action=create`, if the same `uuid` already exists for the job, the existing object ID is returned instead of inserting a duplicate
4. `cvat-core` generates a UUID for each new object, sends it on save, and regenerates it on copy/paste (`clearServerId`)

Geometry-based proxy-timeout reconciliation remains as defense-in-depth for older clients that do not send `uuid`.

### How has this been tested?

**Unit tests:**

```bash
python manage.py test --settings cvat.settings.testing \
  cvat.apps.engine.tests.test_annotation_uuid_idempotency -v 2
```

Covers:
- Creating twice with the same `uuid` returns the same ID and does not duplicate rows
- Creating without `uuid` keeps legacy insert behavior

**Manual / regression checks:**
- Draw many shapes, save, simulate lost create response / timeout, then save again — no duplicates
- Normal save still assigns server IDs and works as before
- Copy/paste creates a new `uuid` (does not reuse the source object's key)
- Audio interval / track / tag creates with `uuid` behave the same way

### Checklist

### License


bsekachev:
Hello,

How the PR solves the issue? 
Reconsilation was already implemented, but it is only workaround. It does not fix the issue. And as I see this PR does not solve it as well.

prathamesh-patil-5090:
> Hello,
> 
> How the PR solves the issue? Reconsilation was already implemented, but it is only workaround. It does not fix the issue. And as I see this PR does not solve it as well.

Ohh yeah — the earlier changes were still just hardening the existing reconciliation path.

I've reworked this so creates are idempotent: each annotation gets a client-generated uuid, and action=create reuses the existing row if (job, uuid) already exists. That should stop the duplicate inserts when the response is lost after the server has already saved. The geometry matching is still there as a fallback for clients that don't send uuid yet.

bsekachev:
Hello, thanks for the contribution.
However this problem is quite difficult and requires design stage before implementation.
Now we are not ready accept this pull request.

PR #10775, opened 2026-06-15
fix(engine): match manifest entries by filename not by index (#10104)

### LINKED ISSUE #10104: Task creation fails if a manifest is present but frame order does not match it

adhavan18:
## Summary

Fixes #10104 — task creation raised `Incorrect file mapping to manifest content` whenever the extractor's sort order differed from the order of entries in an already-existing `manifest.jsonl`.

**Root cause**

`_collect_image_dataset_descriptors` looked up each manifest entry with `manifest[frame_id]` — an index-based access that silently assumed the manifest and the extractor iterate files in the same order. When they diverged (e.g. user uploaded images sorted lexicographically but provided a manifest in a different order), the index mismatch triggered the validation error.

**Fix**

Before the frame loop, build a `{full_name: entry}` dict from the manifest. Inside the loop, look up by relative path (falling back to bare filename for nested paths). This validates that every frame has a corresponding manifest entry without requiring matching order.

```python
# before
image_info = manifest[frame_id]           # index-based — order-sensitive
manifest_image_path = f"{image_info['name']}{image_info['extension']}"
if image_path != manifest_image_path ...:
    raise ValidationError("Incorrect file mapping to manifest content")

# after
image_info = manifest_by_name.get(image_path) or manifest_by_name.get(image_path.rsplit("/", 1)[-1])
if image_info is None:
    raise ValidationError(f"File '{image_path}' was not found in the manifest")
```

## Test plan

- [ ] Create a task with images uploaded in lexicographic order and a `manifest.jsonl` whose entries are in a different order — task creation should succeed.
- [ ] Create a task with a manifest that references a file not present in the upload — should raise `ValidationError` with the new message.
- [ ] Existing task creation flows without a manifest are unaffected.


adhavan18:
gentle ping. this has been open since mid-june with a reviewer requested but no review yet.

**correction to the original version of this comment:** i wrote "CI is green" here, and that was wrong. CI has never run on this PR at all — there are zero checks on the head commit. fork PRs need a maintainer to approve the workflow run first, so the suite has never executed. i should have checked rather than assumed, and i'd rather correct it than leave it standing.

it is a small fix: manifest entries were matched by index rather than by filename, so a manifest whose order differs from the file listing associated the wrong metadata with a frame. happy to rebase, add tests, or close it if it is not wanted.

adhavan18:
Bumping this in case it fell off the radar — still green, no conflicts. Happy to address anything if it needs another look.

adhavan18:
the filename-based manifest lookup and missing-entry validation are in the latest head. the remaining validation items are the two order/missing-file cases listed in the test plan; please let me know if you want those added as repository tests before review.

PR #10777, opened 2026-06-15
Fix 10470: preserver query params for empty filters

lleelleeoo:
### Motivation and context

Broken navigation for Projects and Tasks
https://github.com/cvat-ai/cvat/issues/10470

#### What this changes

This refactors `ResourceFilterHOC` so that the applied filter is derived from the value prop instead of being mirrored in local state and synced back through an effect.

__Before__, the component kept `appliedFilter` in `useState` and ran a `useEffect` over `[appliedFilter, value]` that translated state mutations into `onApplyFilter(...)` calls. UI controls only mutated `appliedFilter`; the effect decided what to actually apply. There was also an `isMounted` flag to suppress the effect on first render, plus several if `(value === ...)` return guards to avoid redundant calls.

__After__:

`value` is the single source of truth. `appliedFilter` is computed from it via `useMemo` (predefined / built parsed from `value`, recent derived as `recentFilters[value] ?? null`).
UI handlers call `onApplyFilter(...)` directly, so the intent of each control is explicit at the call site.
The syncing `useEffect`, the `isMounted` flag, and the dedup guards are removed.

#### Why

_Single source of truth_. The old code had two representations of the same thing — the `value` prop (the real state, owned by the parent/URL) and the local `appliedFilter` mirror. Keeping them in sync via an effect is exactly the pattern that produced subtle bugs: the mount guard, the redundant-call checks, and stale state that could diverge from value.

_No stale state_. Deriving recent from `value` means it can never get out of sync with the actually applied filter. The previous attempt with a dedicated `appliedRecentFilter` state only reset itself inside the recent menu, so it went stale after Clear / Apply / predefined changes; the derived version is self-correcting.

_Explicit data flow_. Each control now says what it does (`handleApplyFilter`, `handleClearFilter`, …) instead of nudging shared state and letting an effect infer intent one render later.
Fewer effects / re-renders, easier to follow, and it removes the indirection that made the filter interact badly with pagination/back-navigation.
This is the change that fixes the recent-filter toggle and keeps the filter consistent with value across reloads and back-navigation.

#### Issue-specific test introduced

`tests/cypress/e2e/issues_prs/issue_10470_back_navigation_pagination.js`

#### Related tests checked — all green

- `tests/cypress/e2e/issues_prs/issue_5566_filter_url.js`
- `tests/cypress/e2e/actions_tasks/registration_involved/case_69_filters_sorting_jobs.js`
- `tests/cypress/e2e/features/case_113_new_organization_pipeline.js`
- `tests/cypress/e2e/features2/consensus.js`

### Checklist

- ~~I have updated the documentation accordingly~~

### License

### DIFF
--- a/changelog.d/20260615_234209_vsbrute_10470_back_navigation_pagination.md
+++ b/changelog.d/20260615_234209_vsbrute_10470_back_navigation_pagination.md
@@ -1,6 +1,4 @@
 ### Fixed
 
-- Pagination state on the projects and project pages is no longer reset to the
-  first page when a filter is cleared, so the current page is preserved on back
-  navigation
+- Current page on projects/tasks/jobs pages is no longer reset after backward navigation from a resource page
   (<https://github.com/cvat-ai/cvat/pull/10777>)
--- a/changelog.d/20260615_234209_vsbrute_10470_back_navigation_pagination.md
+++ b/changelog.d/20260615_234209_vsbrute_10470_back_navigation_pagination.md
@@ -1,0 +1,6 @@
+### Fixed
+
+- Pagination state on the projects and project pages is no longer reset to the
+  first page when a filter is cleared, so the current page is preserved on back
+  navigation
+  (<https://github.com/cvat-ai/cvat/pull/10777>)
--- a/cvat-ui/src/components/project-page/project-page.tsx
+++ b/cvat-ui/src/components/project-page/project-page.tsx
@@ -62,6 +62,7 @@ export default function ProjectPageComponent(): JSX.Element {
     const history = useHistory();
     const [projectInstance, setProjectInstance] = useState<Project | null>(null);
     const [fechingProject, setFetchingProject] = useState(true);
+    const [isMounted, setIsMounted] = useState(false);
     const mounted = useRef(false);
 
     const {
@@ -98,6 +99,7 @@ export default function ProjectPageComponent(): JSX.Element {
                     if (project && mounted.current) {
                         dispatch(getProjectTasksAsync({ ...updatedQuery, projectId: id }));
                         setProjectInstance(project);
+                        setIsMounted(true);
                     }
                 }).catch((error: Error) => {
                     if (mounted.current) {
@@ -126,9 +128,11 @@ export default function ProjectPageComponent(): JSX.Element {
     }, []);
 
     useEffect(() => {
-        history.replace({
-            search: updateHistoryFromQuery(tasksQuery),
-        });
+        if (isMounted) {
+            history.replace({
+                search: updateHistoryFromQuery(tasksQuery),
+            });
+        }
     }, [tasksQuery]);
 
     useEffect(() => {
--- a/cvat-ui/src/components/resource-sorting-filtering/filtering.tsx
+++ b/cvat-ui/src/components/resource-sorting-filtering/filtering.tsx
@@ -165,7 +165,7 @@ export default function ResourceFilterHOC(
                 setSearchTree(tree);
                 return {
                     recent: recentFilters[value] ?? null,
-                    predefined: splitFilterIntoPredefined(Object.values(predefinedFilters), value),
+                    predefined: splitFilterIntoPredefined(Object.values(predefinedFilters ?? {}), value),
                     built: JSON.stringify(QbUtils.jsonLogicFormat(tree, config).logic),
                 };
             } catch {
--- a/cvat-ui/src/components/resource-sorting-filtering/filtering.tsx
+++ b/cvat-ui/src/components/resource-sorting-filtering/filtering.tsx
@@ -199,7 +199,7 @@ export default function ResourceFilterHOC(
             }
 
             if (appliedFilter.predefined?.length) {
-                const predefinedFilterToApply = unite(appliedFilter.predefined)
+                const predefinedFilterToApply = unite(appliedFilter.predefined);
                 if (value === predefinedFilterToApply) {
                     return;
                 }
@@ -229,7 +229,7 @@ export default function ResourceFilterHOC(
                 onApplyFilter(null);
                 setState(defaultTree);
             }
-        }, [appliedFilter, value]);
+        }, [appliedFilter, value, isMounted]);
 
         const renderBuilder = (builderProps: any): JSX.Element => (
             <div className='query-builder-container'>
--- a/cvat-ui/src/components/resource-sorting-filtering/filtering.tsx
+++ b/cvat-ui/src/components/resource-sorting-filtering/filtering.tsx
@@ -130,6 +130,16 @@ export default function ResourceFilterHOC(
         return result;
     }
 
+    function getSearchTreeFromFilterValue(filterValue: string | null): ImmutableTree {
+        if (!filterValue || filterValue === '{}') return defaultTree;
+        try {
+            const tree = QbUtils.loadFromJsonLogic(JSON.parse(filterValue), config);
+            return (tree && isValidTree(tree)) ? tree : defaultTree;
+        } catch {
+            return defaultTree;
+        }
+    }
+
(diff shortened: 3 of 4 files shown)


cvat-ui/src/components/resource-sorting-filtering/filtering.tsx
```diff
@@ -199,22 +199,37 @@ export default function ResourceFilterHOC(
             }
 
             if (appliedFilter.predefined?.length) {
-                onApplyFilter(unite(appliedFilter.predefined));
+                const predefinedFilterToApply = unite(appliedFilter.predefined)
+                if (value === predefinedFilterToApply) {
+                    return;
+                }
+
+                onApplyFilter(predefinedFilterToApply);
```
bsekachev: May we just add a memoized function wrapper, e.g. `onApplyFilterCond` that checks new filter with `value` and calls only if new is different?

cvat-ui/src/components/resource-sorting-filtering/filtering.tsx
```diff
@@ -199,22 +199,37 @@ export default function ResourceFilterHOC(
...
             } else {
+                if (value === null) {
+                    return;
+                }
+
                 onApplyFilter(null);
                 setState(defaultTree);
             }
-        }, [appliedFilter]);
+        }, [appliedFilter, value]);
```
bsekachev: not related to the PR, but it seems `isMounted` is not presented in dependencies, but used

changelog.d/20260615_234209_vsbrute_10470_back_navigation_pagination.md
```diff
@@ -0,0 +1,6 @@
+### Fixed
+
+- Pagination state on the projects and project pages is no longer reset to the
```
bsekachev: I would rephrase somehow simpler, e.g. Current page on projects/tasks/jobs pages is no longer reset after backward navigation from a resource page

cvat-ui/src/components/project-page/project-page.tsx
```diff
@@ -126,9 +128,11 @@ export default function ProjectPageComponent(): JSX.Element {
     }, []);
 
     useEffect(() => {
-        history.replace({
-            search: updateHistoryFromQuery(tasksQuery),
-        });
+        if (isMounted) {
```
bsekachev, line 131: Could you please clarify how is it related to the problem?

tests/cypress/e2e/issues_prs/issue_10470_back_navigation_pagination.js
```diff
@@ -0,0 +1,94 @@
...
+        });
+
+        it('Page 2 is preserved after page reload', () => {
+            cy.reload();
+            cy.get('.cvat-spinner').should('not.exist');
+            cy.get('.ant-pagination-item-active').should('have.text', '2');
+        });
+    });
+
+    describe('Single project page (tasks)', () => {
```
bsekachev: The task process creation is generally rather long. 
Actually, as we fixed the common code, I believe just one test with projects would be enough.

tests/cypress/e2e/issues_prs/issue_10470_back_navigation_pagination.js
```diff
@@ -0,0 +1,94 @@
...
+
+import { defaultTaskSpec } from '../../support/default-specs';
+
+const issueId = '10470';
+const PROJECTS_COUNT = 15; // projects pageSize=12 — page 2 exists
+const TASKS_COUNT = 11; // single-project tasks pageSize=10 — page 2 exists
+const projectPrefix = `issue-${issueId}-project`;
+const taskPrefix = `issue-${issueId}-task`;
+
+context(`Issue ${issueId} - Pagination state is preserved on page reload`, () => {
```
bsekachev: Initial issue was not a reload.
Initial issue was: from resources page open a resource, then do backward navigation. So, I would propose to test this case. I do not mind to have one more test for reload.

cvat-ui/src/components/project-page/project-page.tsx
lleelleeoo: This causes the URL to flicker in the task list. It shouldn’t break anything, but it feels a bit distracting.
It might be better to move this to a separate PR or leave it unchanged for now.

https://github.com/user-attachments/assets/9e3d0edd-bfdb-48a7-b7f0-d4db216e486b

cvat-ui/src/components/resource-sorting-filtering/filtering.tsx
lleelleeoo: Hm. I'll take a little more time for this comment 🤔 
I don't like sync flow via this useEffect. Maybe I can make something different

cvat-ui/src/components/resource-sorting-filtering/filtering.tsx
lleelleeoo: fixed

changelog.d/20260615_234209_vsbrute_10470_back_navigation_pagination.md
lleelleeoo: Fixed

tests/cypress/e2e/issues_prs/issue_10470_back_navigation_pagination.js
lleelleeoo: Oh, it's fair
Removed

tests/cypress/e2e/issues_prs/issue_10470_back_navigation_pagination.js
lleelleeoo: Ok, now there is two similar tests: with backward navigation and page refresh

cvat-ui/src/components/resource-sorting-filtering/filtering.tsx
lleelleeoo: @bsekachev

I'd like to propose a refactor that uses value as the single source of truth instead of keeping internal state. Happy to discuss!

bsekachev:
@codex review

cvat-ui/src/components/resource-sorting-filtering/filtering.tsx
```diff
@@ -138,40 +138,41 @@ export default function ResourceFilterHOC(
...
+                if (!tree || !isValidTree(tree)) {
+                    return innerAppliedFilter;
                 }
-            } catch (_: any) {
-                // nothing to do
+
+                setSearchTree(tree);
+                return {
+                    recent: recentFilters[value] ?? null,
+                    predefined: splitFilterIntoPredefined(Object.values(predefinedFilters), value),
```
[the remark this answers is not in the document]
lleelleeoo: Hm. Valid, fixed
this code didn't change, but useEffect suppressed this sideeffect

cvat-ui/src/components/resource-sorting-filtering/filtering.tsx
```diff
@@ -138,40 +138,41 @@ export default function ResourceFilterHOC(
...
         useEffect(() => {
             setRecentFilters(receiveRecentFilters());
-            setIsMounted(true);
+        }, []);
+
+        const appliedFilter = useMemo(() => {
+            const innerAppliedFilter = { ...defaultAppliedFilter };
+
+            if (!value || value === '{}') {
+                setSearchTree(defaultTree);
```
bsekachev: It does not seem natural to call any setState from inside of useMemo (actually, does not seem a good practice from what I just read)

lleelleeoo: Yes, same as calling setState inside useEffect — which is exactly what this code was doing before it was moved into useMemo.

Either way, I've moved the call into the render function following the official React recommendation:
https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

PR #10790, opened 2026-06-17
Prepare the UI for agent-based interactors

SpecLad:
### Motivation and context

This involves two changes:

* Pass an annotation request type when invoking lambda functions.

* Move certain AR response postprocessing logic to `ToolsControlComponent`, so that it still works when `LambdaManager.call` is overridden by a plugin. In particular, the conversion of points to an `Int32Array` and the dropping of unsupported shape types.

Also, to simplify the code, drop the keys from `InteractorShape` that the code doesn't actually do anything with.

### How has this been tested?

I have manually tested with with IOG, SAM, and SAM2.

### Checklist

### License

### DIFF
--- a/cvat-core/src/lambda-manager.ts
+++ b/cvat-core/src/lambda-manager.ts
@@ -6,15 +6,18 @@
 import serverProxy from './server-proxy';
 import { ArgumentError } from './exceptions';
 import MLModel from './ml-model';
-import {
-    ModelKind, RQStatus, ShapeType, Source,
-} from './enums';
-import { SerializedCollection, SerializedShape, SerializedFunctionRequest } from './server-response-types';
+import { ModelKind, RQStatus, ShapeType } from './enums';
+import { SerializedCollection, SerializedFunctionRequest } from './server-response-types';
 import { mask2Rle } from './rle-utils';
 
-type InteractorShape = Pick<SerializedShape, 'group' | 'source' | 'attributes' | 'occluded' | 'rotation' | 'type'> & {
-    points: Int32Array;
-};
+export interface MinimalShape {
+    type: ShapeType;
+    points: number[];
+}
+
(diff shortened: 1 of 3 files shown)


bsekachev:
Could you please explain motivation of this?

> Move certain AR response postprocessing logic to ToolsControlComponent, so that it still works when LambdaManager.call is overridden by a plugin. In particular, the conversion of points to an Int32Array and the dropping of unsupported shape types.

SpecLad:
If I don't do this, I'll have to reimplement these operations in the functions UI plugin, which would lead to code duplication.

cvat-core/src/lambda-manager.ts
```diff
@@ -102,39 +100,20 @@ class LambdaManager {
...
+                        points: item.points,
+                        attributes: Array.isArray(item.attributes) && item.attributes.every(
+                            (attr) => typeof attr === 'object' &&
                                 typeof attr.spec_id === 'number' &&
-                                typeof attr.value === 'string';
-                        }) ? item.attributes : [],
-                        occluded: typeof item.occluded === 'boolean' ? item.occluded : false,
-                        rotation: typeof item.rotation === 'number' ? item.rotation : 0,
+                                typeof attr.value === 'string',
+                        ) ? item.attributes : undefined,
```
bsekachev: Let's keep empty array as a standard value for attributes.
It is more convenient to always array from this API, rather than two different types

bsekachev:
> If I don't do this, I'll have to reimplement these operations in the functions UI plugin, which would lead to code duplication.

I see, in this case we need one more related patch for sam_decoders

cvat-core/src/lambda-manager.ts
SpecLad: Alright, updated.

bsekachev (approved):

PR #10794, opened 2026-06-18
Introduce region of interest for detectors and interactors

### LINKED ISSUE #7696: Improve automatic detection by sending extra data (a region of interest)

I am working on a specific segmentation problem.

I have "raw" images of various dimensions, and my goal is to make a semantic segmentation ML model that expects a cropped part of my raw image.

In CVAT, I uploaded my "raw" images and I did my annotations manually. Now that I have a working model, I transformed it into a Nuclio function in order to make automatic annotations.

The problem is, when executing an automatic annotation (at least for type = "detector), CVAT only sends the "raw" image, making the prediction unusable.

Long story short, i have to upload pre-processed images into CVAT if i want to use my automatic annotation model, which means:
- working on pre-processed images
- extra work before annotating
- extra work after annotating
- need to maintain correlation between raw images and cvat jobs/tasks/frames to be able to undo the "pre-processing"

I "need" my annotations on my full raw image to data augmentation.

I want CVAT to send special annotations to nuclio function.

Before starting my automatic annotation on a task/frame, i would need to do one specific annotation (in my case, a rectangle shape that would represent my region of interest)

When I execute the model detection, this annotation details would be sent to my nuclio function. My function would take care of the cropping/resize/prediction/resize/uncrop and send back the result to CVAT.

I looked in the source code, and i believe the feature would belong here `cvat/apps/lambda_manager/views.py`

``` python
# L 394
if self.kind == LambdaType.DETECTOR:
            payload.update({
                "image": self._get_image(db_task, mandatory_arg("frame"), quality),
                # add here the current frame annotation data
            })
```

- Do a pre-processing of my image => store my ROI in a file, and the operations done on each image
- Upload my pre-processed images to CVAT, manually or automatically do my annotations
- Export my annotations
- Do a reverse-processing to have my annotations on my "raw" image

### Additional context

If someone can explain to me how to get a specific annotation from the current frame, i can work myself out for my special use case.

I tried adding `"shapes": data.get("shapes", []),` but it didn't work out. The field is sent, but is empty.

bsekachev:
### Motivation and context

Resolved #7696

This PR adds region of interest support for automatic annotation detector and interactor workflows.

Users can now restrict automatic annotation to a selected image area. The UI supports entering ROI coordinates and drawing an ROI in the annotation canvas. The server crops the image before invoking the function and remaps prompts/results between ROI-local coordinates and full-image coordinates, so existing Nuclio functions can continue receiving only an image.

Main changes:
- Added reusable ROI input controls for detector/interactor UI flows.
- Added canvas interaction support for ROI drawing and ROI-aware prompt placement.
- Added ROI request support for lambda detector and interactor calls.
- Added server-side ROI parsing, image cropping, and result coordinate translation.
- Added SAM plugin ROI handling for prompt/result coordinate mapping.

### How has this been tested?

Manually (human pose, IOG, SAM, SAM2/3, YOLO)

### Checklist

### License


bsekachev:
@codex review

cvat/apps/lambda_manager/utils.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+class ROIHelper:
+    """
+    Utilities for running automatic annotations functions on a cropped region of an image.
+
+    Public API requests pass ROI as image coordinates: [xtl, ytl, xbr, ybr].
+    Lambda functions still receive only an image, so the server crops the image
+    before invocation and then maps prompts/results between these coordinate
+    spaces:
```
SpecLad: The text is incomplete.

cvat/apps/lambda_manager/utils.py
```diff
@@ -0,0 +1,155 @@
...
+    """
+
+    MIN_ROI_SIZE = 128
+
+    @classmethod
+    def parse_roi(cls, image_width: int, image_height: int, roi: list | None) -> dict | None:
+        """
+        Validate request ROI and normalize it to a named internal structure.
```
SpecLad: It's a dict. It's neither named, nor a structure.

cvat/apps/lambda_manager/utils.py
```diff
@@ -0,0 +1,155 @@
...
+        code explicit at call sites and avoids relying on positional indexes
+        after validation.
+        """
+        if roi is None:
+            return None
+
+        try:
+            assert isinstance(roi, list) and len(roi) == 4
+            xtl, ytl, xbr, ybr = (int(coordinate) for coordinate in roi)
+        except (TypeError, ValueError, AssertionError) as ex:
```
SpecLad: How could `AssertionError` possibly occur here?

cvat/apps/lambda_manager/utils.py
```diff
@@ -0,0 +1,155 @@
+# Copyright (C) CVAT.ai Corporation
+#
+# SPDX-License-Identifier: MIT
+
+from __future__ import annotations
+
+import numpy as np
+from django.core.exceptions import ValidationError
```
SpecLad: You should use `ValidationError` from DRF instead, because DRF automatically translates it to an HTTP 400 response.

cvat/apps/lambda_manager/utils.py
```diff
@@ -0,0 +1,155 @@
...
+        """
+        if roi is None:
+            return None
+
+        try:
+            assert isinstance(roi, list) and len(roi) == 4
+            xtl, ytl, xbr, ybr = (int(coordinate) for coordinate in roi)
+        except (TypeError, ValueError, AssertionError) as ex:
+            raise ValidationError(
+                "ROI must contain four integer coordinates: [xtl, ytl, xbr, ybr]",
```
SpecLad, line 33: The code in the try block doesn't actually check if the coordinates are integers.

cvat/apps/lambda_manager/utils.py
```diff
@@ -0,0 +1,155 @@
...
+            ) from ex
+
+        if xtl < 0 or ytl < 0 or xbr > image_width or ybr > image_height:
+            raise ValidationError("ROI is outside the image", code=status.HTTP_400_BAD_REQUEST)
+
+        if xbr - xtl < cls.MIN_ROI_SIZE or ybr - ytl < cls.MIN_ROI_SIZE:
+            raise ValidationError(
+                f"ROI size must be at least {cls.MIN_ROI_SIZE}x{cls.MIN_ROI_SIZE}px",
+                code=status.HTTP_400_BAD_REQUEST,
+            )
```
SpecLad: What's the reason for this restriction? It seems kind of artificial.

cvat/apps/lambda_manager/utils.py
```diff
@@ -0,0 +1,155 @@
...
+            assert isinstance(roi, list) and len(roi) == 4
+            xtl, ytl, xbr, ybr = (int(coordinate) for coordinate in roi)
+        except (TypeError, ValueError, AssertionError) as ex:
+            raise ValidationError(
+                "ROI must contain four integer coordinates: [xtl, ytl, xbr, ybr]",
+                code=status.HTTP_400_BAD_REQUEST,
+            ) from ex
+
+        if xtl < 0 or ytl < 0 or xbr > image_width or ybr > image_height:
+            raise ValidationError("ROI is outside the image", code=status.HTTP_400_BAD_REQUEST)
```
SpecLad: This check seems like it ought to be in `crop_image`.

cvat/apps/lambda_manager/utils.py
```diff
@@ -0,0 +1,155 @@
...
+
+from __future__ import annotations
+
+import numpy as np
+from django.core.exceptions import ValidationError
+from PIL import Image
+from rest_framework import status
+
+
+class ROIHelper:
```
SpecLad, line 13: IMO, there's a missed opportunity here for a more object-oriented design: have `parse_roi` return a `ROI` object, and turn `crop_image` and `translate_xxx` methods on that object. It should make for cleaner client code.

cvat/apps/lambda_manager/utils.py
bsekachev: Yes, it is something artificial but I believe we need some kind of that with a reasonable value to avoid corner cases in target agents. 
E.g. I do not have any idea what result will be if we send a tiny image to our nuclio functions / hf/ roboflow / ai agents, etc.

SpecLad: But it's already possible to send them a tiny image (if the original frame is already tiny). Seems like we should handle this uniformly regardless of whether a ROI is specified.

bsekachev: alright, in this case I removed this check from everywhere

cvat/apps/lambda_manager/utils.py
```diff
@@ -0,0 +1,200 @@
...
+        from cvat.apps.engine.models import MediaType, Task
+
+        db_task = Task.objects.select_related("data", "data__video").get(pk=task_id)
+
+        db_data = db_task.data
+        frames = [frame for frame in range(db_data.size) if frame not in db_data.deleted_frames]
+        if not frames:
+            return parsed_roi
+
+        if db_task.media_type == MediaType.VIDEO:
```
SpecLad: There is no `MediaType.VIDEO`.

SpecLad (approved):

PR #10815, opened 2026-06-23
Better UX around create_thread db_lock

Rwwwrl:
### Title
Reduce task creation lock contention

### Motivation and context

Task data creation can hold database locks for a long time and block updates or deletions of related resources. Lock-sensitive operations now use a bounded PostgreSQL lock timeout and return `409 Conflict` when the resource is busy instead of waiting indefinitely or surfacing a server error. Task data initialization also locks only the task/data rows it needs, shares the referenced cloud storage row, delays job-creation row locking until after cheap validations, and rejects annotation/auto-annotation operations while task media data is still being initialized.

### Checklist

### License


cvat/utils/django_database/utils.py
```diff
@@ -0,0 +1,182 @@
+import logging
```
zhiltsov-max, line 5: Please add a license header.

cvat/utils/django_database.py
```diff
@@ -0,0 +1,182 @@
...
+        if isinstance(current, PsycopgError):
+            return current
+
+        return None
+
+    return find_deepest(exc)
+
+
+def set_local_lock_timeout(timeout_seconds: int | None = None) -> None:
+    if connection.vendor != "postgresql":
```
zhiltsov-max: I think it will be better to add a branch just for sqlite and fail with an assertion error otherwise. We know why there can be postgres and sqlite, but we don't know why there could be anything else. And if another DB backed is used, the function can silently hide an error in the code. The same is for other backend-dependent code, e.g. `get_object_by_id_for_share`.

cvat/utils/django_database.py
```diff
@@ -0,0 +1,182 @@
...
+                    timeout_seconds
+                    if timeout_seconds is not None
+                    else settings.CVAT_POSTGRES_TRANSACTION_LOCK_TIMEOUT_SECONDS
+                )
+                * 1000
+            ],
+        )
+
+
+def set_local_lock_timeout_decorator(
```
zhiltsov-max: What do you think about extracting decorators into a separate file and renaming to something like `decorators.with_lock_timeout`? It would be more expressive and clear.

cvat/utils/django_database/utils.py
```diff
@@ -0,0 +1,182 @@
...
+    def decorator(func: Callable[P, R]) -> Callable[P, R]:
+        @wraps(func)
+        def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
+            set_local_lock_timeout(timeout_seconds=timeout_seconds)
+            return func(*args, **kwargs)
+
+        return wrapper
+
+    return decorator
+
```
zhiltsov-max, line 91: Please move https://github.com/cvat-ai/cvat/blob/57cc6434c5660d58264c1bc73903610ffd223a11/cvat/apps/engine/utils.py#L529-L534 into this file as well.

cvat/settings/base.py
```diff
@@ -123,6 +123,13 @@ def generate_secret_key():
 SITE_ID = 1
 
 
+CVAT_POSTGRES_TRANSACTION_LOCK_TIMEOUT_SECONDS = int(
+    os.getenv("CVAT_POSTGRES_TRANSACTION_LOCK_TIMEOUT_SECONDS", 30)
+)
+
```
zhiltsov-max: I think it should be renamed to something backend-agnostic, like `DB_LOCK_TIMEOUT` or even `VIEW_LOCK_TIMEOUT`. The need for a timeout is not related to a specific DB, it's a CVAT UX requirement.

cvat/apps/events/handlers.py
```diff
@@ -676,12 +680,26 @@ def handle_rq_exception(rq_job, exc_type, exc_value, tb):
     return False
 
 
-def handle_viewset_exception(exc, context):
+def handle_viewset_exception(exc: Exception, context):
+    if isinstance(exc, DatabaseError):
+        if isinstance(find_psycopg_cause(exc=exc), LockNotAvailable):
```
zhiltsov-max: Could you try to hide these backend details behind a function to check that the error is a lock error? It would simplify the function usage.

cvat/apps/engine/views.py
```diff
@@ -1294,6 +1305,11 @@ def perform_create(self, serializer, **kwargs):
         # Required for the extra summary information added in the queryset
         serializer.instance = self.get_queryset().get(pk=serializer.instance.pk)
 
+    @transaction.atomic
+    @db_utils.set_local_lock_timeout_decorator()
+    def perform_destroy(self, instance: Task) -> None:
```
zhiltsov-max: Could you check if we can set a global lock acquisition timeout? It feels like we just need to set it for all endpoints, potentially using a longer value in the background tasks. I assume it can make resource management errors in endpoints more obvious, but they are still errors. Having a global policy will simplify management for this and make system behavior consistent.

cvat/utils/django_database/utils.py
Rwwwrl: updated

cvat/utils/django_database.py
Rwwwrl: Agree, updated - https://github.com/cvat-ai/cvat/pull/10815/changes/9bf80d760ae4376d40fc32f58f342d3ecefee7eb

cvat/utils/django_database.py
Rwwwrl: Agree. Splitted single django_database into decorators.py / utils.py / ... modules dc791af48e96cc470c220b3353e5175913cbb836

but I did not change the name to "with_lock_timeout". Why? - `with` is about something that __enter__ and __exit__, that decorator does not have __exit__. I thought about that though

nevertheless, I have added explicit requirement: it must be run in transaction:
```
    if not connection.in_atomic_block:
        raise RuntimeError("set_local_lock_timeout must be called inside a transaction")
```

cvat/utils/django_database/utils.py
Rwwwrl: updated d97f363ba0dd67b14ac673ccaa5623ccb3695980

cvat/apps/events/handlers.py
Rwwwrl: updated c473c3d488605fec9ab25993fc5ee98c54245ecd

I looked at how drf.exception_handler works: it knows about django exceptions, such as NotFound and PermissionError and converts those into ApiException. I just extended it in exactly the same way: it converts DatabaseError in particular case into ApiException - into ResourceIsBusyApiException

cvat/settings/base.py
Rwwwrl: Kept it database related, but made provider agnostic - CVAT_DATABASE_TRANSACTION_LOCK_TIMEOUT_SECONDS

cvat/utils/django_database/__init__.py
```diff
@@ -0,0 +1,3 @@
+# Copyright (C) 2023 Intel Corporation
+#
+# SPDX-License-Identifier: MIT
```
zhiltsov-max:
```suggestion
```

You can add a CVAT.ai header, alternatively.

cvat/utils/django_database/utils.py
zhiltsov-max: I still don't see a header in this file.

cvat/utils/django_database/decorators.py
```diff
@@ -0,0 +1,24 @@
+# Copyright (C) 2023 Intel Corporation
```
zhiltsov-max:
```suggestion
# Copyright (C) CVAT.ai Corporation
```

cvat/utils/django_database/contextmanagers.py
```diff
@@ -0,0 +1,24 @@
+# Copyright (C) 2023 Intel Corporation
```
zhiltsov-max:
```suggestion
# Copyright (C) CVAT.ai Corporation
```

cvat/apps/events/handlers.py
zhiltsov-max: Yes, but the problem is that now this code depends on the DB backend details. Django mostly encapsulates this, and my idea is to maintain this encapsulation here as well.

cvat/settings/base.py
zhiltsov-max: Could you keep the naming aligned with other variables then? We would typically drop `_seconds` and contract `database` to `db` for internal names.

cvat/utils/django_database.py
zhiltsov-max: Actually, the suggestion was only about `decorators`. I'm ok with splitting the other files, but I'd like to have them importable in a single line like this:

```python
import cvat.utils.django_database as db_utils

...

    @db_utils.decorators.with_lock_timeout
    def perform_update(...):
...

class Data:
    task: db_utils.MaybeUndefined[Task]
```

I'm not sure calling it `with_lock_timeout` would be confusing with a context manager - saying `with with_lock_timeout():` looks quite odd to me. If I wanted a context manager, I'd go with a single `with` in a line.  I don't like having a `set_` in the name, because decorators express a declarative-style definition, while `set` is imperative.

cvat/utils/django_database/utils.py
Rwwwrl: updated

cvat/utils/django_database.py
Rwwwrl: decorator is gone

I updated init.py file in order to provide public interface to that package internals. Update usages to use db_utils.<thing> as well

cvat/settings/base.py
Rwwwrl: it is gone. New variable CVAT_POSTGRES_LOCK_TIMEOUT does follow it

cvat/utils/django_database/__init__.py
Rwwwrl: updated

cvat/utils/django_database/decorators.py
Rwwwrl: updated

cvat/utils/django_database/contextmanagers.py
Rwwwrl: updated

cvat/apps/engine/views.py
Rwwwrl: Final solution: we stick with the global default lock timeout, see 5cdbd856897f8d4f6abbc97794436353639eb9bf

cvat/apps/events/handlers.py
Rwwwrl: 2f3cde4a349401f28bedece767e07510d931e075

cvat/settings/base.py
zhiltsov-max: I think it returns us back to the previous iteration. Could you rename it to `CVAT_DB_LOCK_TIMEOUT`?

Rwwwrl: updated to CVAT_DB_LOCK_TIMEOUT

zhiltsov-max (approved):

PR #10816, opened 2026-06-23
Remove the pyunpack dependency

SpecLad:
### Motivation and context

We're pulling in three packages for what amounts to a thin wrapper around `patool` and `zipfile`. Just use them directly instead (well, for convenience use `shutil` rather than `zipfile`). We don't even have to implement logic for selecting between them, because everywhere we use `pyunpack`, the archive is either guaranteed to be ZIP or guaranteed to be non-ZIP.

### How has this been tested?

I manually tested the exception handling in `media_extractors.py`.

### Checklist

### License


cvat/apps/engine/media_extractors.py
```diff
@@ -424,7 +424,24 @@ def __init__(
...
+                    self._archive_source,
+                ],
+                check=True,
+                stdout=subprocess.DEVNULL,
+                stderr=subprocess.PIPE,
+                text=True,
+                errors="replace",
+            )
+        except subprocess.CalledProcessError as ex:
+            raise RuntimeError("unable to extract archive:\n" + ex.stderr) from ex
```
zhiltsov-max: Consider extracting a function instead of duplicating the code here and in tests. It can even be potentially registered in `shutil` then.

SpecLad: > Consider extracting a function instead of duplicating the code here and in tests.

Yeah, good point. Done.

> It can even be potentially registered in `shutil` then.

I don't really see the need, TBH.

zhiltsov-max (approved):

PR #10818, opened 2026-06-23
SDK: Add persistent auth store foundation

Pr1ncee:
### Motivation and context
Adds the initial SDK persistent-auth storage layer for the planned auth.json profile store.
#### Changes
* Adds cvat_sdk.core.auth
* Defines DEFAULT_SERVER = "https://app.cvat.ai"
* Adds ProfileEntry for saved server/token profile data
* Adds get_auth_store_path() using platformdirs.user_config_path("cvat-sdk", "CVAT.ai") / "auth.json"
* Adds AuthStore load/save internals with:
  * fresh empty document when missing
  * schema version validation
  * corrupt JSON handling
  * atomic writes
  * 0700 parent directory and 0600 file permissions on POSIX
  * refusal to read/write group/other-accessible store paths

### How has this been tested?
Adds focused SDK tests for the new storage behavior

### Checklist

### License


changelog.d/10818.added.md
```diff
@@ -0,0 +1,4 @@
+### Added
+
+- Added SDK persistent authentication storage primitives for future saved profile support.
```
zhiltsov-max: Please focus on the user-facing changes. If there are none, no changelog message is needed.

cvat-sdk/cvat_sdk/core/auth.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+class AuthStoreError(Exception):
+    """Raised when the auth store cannot be safely read or written."""
+
+
+class AuthStore:
+    """Reads/writes the persistent auth.json, enforcing 0600/0700 permissions."""
+
+    def __init__(self, path: Path | None = None) -> None:
```
zhiltsov-max: Could you be more specific about what the expected path is supposed to be? Maybe call it `config_file_path` or similarly.

cvat-sdk/cvat_sdk/core/auth.py
```diff
@@ -0,0 +1,122 @@
...
+        for p in (self._path.parent, self._path):
+            if not p.exists():
+                continue
+
+            mode = stat.S_IMODE(p.stat().st_mode)
+            if mode & 0o077:
+                raise AuthStoreError(
+                    f"Refusing to use {self._path}: '{p}' has insecure permissions "
+                    f"{oct(mode)} (group/other access). "
+                    f"Run: chmod {'700' if p.is_dir() else '600'} '{p}'"
```
zhiltsov-max: The error message will be incorrect if `self._path` is a directory. I think we should check that the passed path is a file, if it exists.

cvat-sdk/cvat_sdk/core/auth.py
```diff
@@ -0,0 +1,122 @@
...
+    def _check_secure_permissions(self) -> None:
+        if self._is_windows():
+            return
+
+        for p in (self._path.parent, self._path):
+            if not p.exists():
+                continue
+
+            mode = stat.S_IMODE(p.stat().st_mode)
+            if mode & 0o077:
```
zhiltsov-max: This check incorrectly allows any leading number, despite what the error message and specification say.

cvat-sdk/cvat_sdk/core/auth.py
```diff
@@ -0,0 +1,122 @@
...
+    created_date: str
+    """ISO-8601 UTC timestamp of when the profile was saved."""
+
+
+def get_auth_store_path() -> Path:
+    """Return the path to the persistent auth.json store."""
+    return platformdirs.user_config_path(_APP_NAME, _APP_AUTHOR) / "auth.json"
+
+
+class AuthStoreError(Exception):
```
zhiltsov-max: SDK-related exceptions are stored in `core/exceptions.py`, and should be inherited from the base type.

cvat-sdk/cvat_sdk/core/auth.py
```diff
@@ -0,0 +1,122 @@
...
+    """Raised when the auth store cannot be safely read or written."""
```
zhiltsov-max: The docstring focuses on the security aspects, but the class is used for any errors in this component.

changelog.d/10818.added.md
Pr1ncee: I'll remove it then, there are no user-facing changes made in this PR

cvat-sdk/cvat_sdk/core/auth.py
```diff
@@ -0,0 +1,124 @@
...
+import stat
+import tempfile
+from pathlib import Path
+
+import attrs
+import platformdirs
+
+from cvat_sdk.core.exceptions import AuthStoreError
+
+DEFAULT_SERVER = "https://app.cvat.ai"
```
zhiltsov-max: Didn't we decide to keep `localhost` default so far?

tests/python/sdk/test_auth_store.py
```diff
@@ -0,0 +1,113 @@
...
+    assert doc == {"version": 1, "profiles": {}}
+
+
+def test_save_then_load_roundtrips(tmp_path):
+    store = _store(tmp_path)
+    store._save({"version": 1, "profiles": {"a": {"server": "https://x", "token": "t"}}})
+    assert store._load()["profiles"]["a"]["token"] == "t"
+
+
+@pytest.mark.skipif(sys.platform == "win32", reason="POSIX permission semantics")
```
zhiltsov-max: Why the OS checks here are different from those in `_is_windows()`?

cvat-sdk/cvat_sdk/core/auth.py
Pr1ncee: as I remember we were discussing using CVAT_HOST instead of introducing a new one. From my perspective it doesn't much matter which is the default, isn't it?

tests/python/sdk/test_auth_store.py
```diff
@@ -0,0 +1,113 @@
...
+from cvat_sdk.core.auth import (
+    DEFAULT_SERVER,
+    AuthStore,
+    AuthStoreError,
+    ProfileEntry,
+    get_auth_store_path,
+)
+
+
+def test_profile_entry_holds_fields():
```
zhiltsov-max: What is the point of this test? Why it's needed?

Pr1ncee: It only verifies that ProfileEntry echoes back constructor values. I can probably delete it

cvat-sdk/cvat_sdk/core/auth.py
```diff
@@ -0,0 +1,124 @@
...
+    created_date: str
+    """ISO-8601 UTC timestamp of when the profile was saved."""
+
+
+def get_auth_store_path() -> Path:
+    """Return the path to the persistent auth.json store."""
+    return platformdirs.user_config_path(_APP_NAME, _APP_AUTHOR) / "auth.json"
+
+
+class AuthStore:
```
zhiltsov-max, line 46: I can see the class has only 1 public method (actually a property) - `path`, besides the builtins. It's quite hard to say if it's correct or not at the moment. Would you consider making it more useful to clients, e.g. make it a writer/reader that works with a profile container, config, etc.? The tests now can only check private methods, and are fragile because of this.

Pr1ncee: I've created more public methods in this PR: https://github.com/cvat-ai/cvat/pull/10819

tests/python/sdk/test_auth_store.py
```diff
@@ -0,0 +1,113 @@
...
+    with pytest.raises(AuthStoreError, match="permission"):
+        store._load()
+
+
+@pytest.mark.skipif(sys.platform == "win32", reason="POSIX permission semantics")
+def test_load_refuses_file_with_special_permission_bits(tmp_path):
+    store = _store(tmp_path)
+    store._save({"version": 1, "profiles": {}})
+    path = tmp_path / "cvat" / "auth.json"
+    os.chmod(path, 0o1600)
```
zhiltsov-max, line 65: Could you comment what is 1600 and why it should be declined?

tests/python/sdk/test_auth_store.py
```diff
@@ -0,0 +1,113 @@
...
+@pytest.mark.skipif(sys.platform == "win32", reason="POSIX permission semantics")
+def test_save_creates_0600_file_in_0700_dir(tmp_path):
+    store = _store(tmp_path)
+    store._save({"version": 1, "profiles": {}})
+    path = tmp_path / "cvat" / "auth.json"
+    assert (path.stat().st_mode & 0o777) == 0o600
+    assert (path.parent.stat().st_mode & 0o777) == 0o700
+
+
+@pytest.mark.skipif(sys.platform == "win32", reason="POSIX permission semantics")
```
zhiltsov-max: I think this check isn't really correct wrt. the result value options in https://docs.python.org/3/library/sys.html#sys.platform. And at this point, this should probably be extracted into a separate function with a clear name (e.g. `is_posix`) that will be used in checks everywhere.

Pr1ncee: Please, refresh the file, I already adjusted this check

cvat-sdk/cvat_sdk/core/auth.py
zhiltsov-max: What is `CVAT_HOST`? I don't think there's such a name related to CLI or SDK.

Pr1ncee: Got it, I meant the other thing. I'll leave localhost as the default

tests/python/sdk/test_auth_store.py
zhiltsov-max: The comment is still valid though. There are more than 2 values even in `os.name`. https://docs.python.org/3/library/os.html#os.name

tests/python/sdk/test_auth_store.py
```diff
@@ -0,0 +1,98 @@
...
+    with pytest.raises(AuthStoreError, match="permission"):
+        store._load()
+
+
+@pytest.mark.skipif(os.name == "nt", reason="POSIX permission semantics")
+def test_load_refuses_file_with_special_permission_bits(tmp_path):
+    store = _store(tmp_path)
+    store._save({"version": 1, "profiles": {}})
+    path = tmp_path / "cvat" / "auth.json"
+    # 0o1600 is 0o600 plus the POSIX sticky bit, so exact 0o600 validation must reject it.
```
zhiltsov-max, line 64: What is the problem with the sticky bit and why it should not be allowed? Does SSH enforce this as well?

tests/python/sdk/test_auth_store.py
Pr1ncee: I inversed the check to detect "posix" os platforms

tests/python/sdk/test_auth_store.py
Pr1ncee: The sticky bit itself is not really the security issue here. It only matters if we intentionally require the mode to be exactly 0600. That's the reason I've added it

zhiltsov-max: I'd say there's no point for us to limit more than necessary for security, so I'd propose to limit only the 3 base numbers and to check only them in tests.

cvat-sdk/cvat_sdk/core/auth.py
zhiltsov-max: Commented there, but the suggestion above stands.

tests/python/sdk/test_auth_store.py
Pr1ncee: ok, will do

cvat-sdk/cvat_sdk/core/auth.py
Pr1ncee: I'd like to leave it as it is if it's not necessary at this moment. I'll decide later about it

PR #10823, opened 2026-06-25
Add TwelveLabs Pegasus video auto-annotation helper to the SDK

mohit-twelvelabs:
Hi! I'm Mohit, I work at TwelveLabs (@mohit-twelvelabs).

This PR adds an opt-in auto-annotation helper to the CVAT SDK that uses the [TwelveLabs](https://twelvelabs.io) Pegasus video-understanding model to pre-label video tasks.

### Motivation and context

CVAT's predefined auto-annotation functions (the torchvision ones in `cvat_sdk/auto_annotation/functions/`) operate per-frame, which is a great fit for image models. Video understanding is a different problem: a model that reasons about a whole clip can describe *what happens and when*, which is useful for bootstrapping temporal/event annotations on video tasks.

This adds `cvat_sdk.auto_annotation.functions.twelvelabs_pegasus`, which sends a video task's source video to Pegasus, asks it to enumerate the distinct events with their start times, and writes each event as a tag on the matching frame. Event labels are matched to task labels by name, reusing the same name-based matching the existing functions rely on.

Because Pegasus is a video model rather than a per-frame one, it is intentionally *not* wired into the `annotate_task` driver / CLI `auto-annotate` command (those apply a `DetectionFunction` image-by-image and explicitly support image tasks only). Instead the helper exposes its own task-level `annotate_task` method, mirroring the SDK's annotation-upload conventions.

It is fully opt-in and non-breaking: it lives behind a new `cvat-sdk[twelvelabs]` extra, nothing imports `twelvelabs` unless you call the helper, and no existing defaults or behavior change.

### How has this been tested?

- No-network unit tests (`tests/python/sdk/test_twelvelabs_pegasus.py`) covering event parsing, label-by-name matching with fallback, timestamp→frame mapping via `fps`, frame clamping to task size, and the `clear_existing` path. They fake both the TwelveLabs and CVAT clients, so they need no server or network. All 7 pass locally.
- The TwelveLabs side of the contract was verified end-to-end against the live API with `twelvelabs==1.2.8`: `analyze(model_name="pegasus1.5", video=..., response_format=json_schema, ...)` returns structured events for a public sample video.

### A note on process

The contributing guide suggests opening an issue first for larger changes — I went straight to a PR since this is small, additive, and behind an optional extra, but I'm very happy to move the discussion to an issue if you'd prefer that.

You can grab a free API key at https://twelvelabs.io — there's a generous free tier.

### Checklist
- [ ] ~~I have linked related issues~~ (no related issue)

### License
- [x] I submit my code changes under the same MIT License that covers the project.

### DIFF
--- a/changelog.d/20260625_000000_mohit_twelvelabs_pegasus.md
+++ b/changelog.d/20260625_000000_mohit_twelvelabs_pegasus.md
@@ -1,0 +1,7 @@
+### Added
+
+- \[SDK\] A new opt-in auto-annotation helper,
+  `cvat_sdk.auto_annotation.functions.twelvelabs_pegasus`,
+  that uses the TwelveLabs Pegasus video-understanding model to label a video
+  task's frames with the events detected in the video
+  (<https://github.com/cvat-ai/cvat/pull/XXXX>)
--- a/cvat-sdk/cvat_sdk/auto_annotation/functions/twelvelabs_pegasus.py
+++ b/cvat-sdk/cvat_sdk/auto_annotation/functions/twelvelabs_pegasus.py
@@ -1,0 +1,206 @@
+# Copyright (C) CVAT.ai Corporation
+#
+# SPDX-License-Identifier: MIT
+
+"""
+Video auto-annotation using the TwelveLabs Pegasus video-understanding model.
+
+Unlike the torchvision detection functions in this package, Pegasus is a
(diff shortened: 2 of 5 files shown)


bsekachev:
Hello, 

If you are interested in a partnership, please contact us using the corresponding [form](https://www.cvat.ai/general).

PR #10824, opened 2026-06-25
SDK resolution, arg parser, client builders, exports

Pr1ncee:
### Checklist

### License


cvat-sdk/cvat_sdk/core/auth.py
```diff
@@ -43,13 +55,150 @@ def get_auth_store_path() -> Path:
...
+    """Build and authenticate a Client from parsed CLI-style arguments."""
+    store = store or AuthStore()
+
+    profile = getattr(parsed_args, "profile", None)
+    auth = getattr(parsed_args, "auth", None)
+    server_host = getattr(parsed_args, "server_host", None)
+    server_port = getattr(parsed_args, "server_port", None)
+    insecure = getattr(parsed_args, "insecure", False)
+    organization = getattr(parsed_args, "organization", None)
+
```
zhiltsov-max: I think you'll need to provide a function to define the parameters as well, if you want to provide `make_client_from_cli` in SDK. Otherwise the clients will always have to define the parameters on their own, possibly introducing deviations from the CVAT CLI in their scripts. It also makes sense to make the parameter set declared somewhere, e.g. as a dataclass, attrs class, or typed dict.

cvat-sdk/cvat_sdk/core/auth.py
```diff
@@ -43,13 +55,150 @@ def get_auth_store_path() -> Path:
...
+    if profile is not None and (
+        server_host is not None or server_port is not None or auth is not None
+    ):
```
zhiltsov-max, line 206:
```suggestion
    if profile is not None and any(v is not None for v in (server_host, server_port, auth)):
```

cvat-sdk/cvat_sdk/core/auth.py
```diff
@@ -43,13 +55,150 @@ def get_auth_store_path() -> Path:
...
+
+    default = store.get_default_profile()
+    if default is not None:
+        return default[1].server
+
+    return store.get_default_server() or DEFAULT_SERVER
+
+
+def make_client_from_profile(name: str | None = None, *, store: AuthStore | None = None) -> Client:
+    """Build and authenticate a Client from a saved profile."""
```
zhiltsov-max: Why not expect a profile object as the parameter? Currently, the logic for default profile handling is duplicated between `make_client_from_cli` and `make_client_from_profile`, and there are 2 `make_client_from_profile` calls in `make_client_from_cli`. If you pass a profile object, you'll be able to deduplicate both.

cvat-sdk/cvat_sdk/core/auth.py
```diff
@@ -43,13 +55,150 @@ def get_auth_store_path() -> Path:
...
+
+    if profile is not None and (
+        server_host is not None or server_port is not None or auth is not None
+    ):
+        raise AuthStoreError(
+            "--profile is mutually exclusive with --server-host/--server-port/--auth."
+        )
+
+    if profile is not None:
+        client = make_client_from_profile(profile, store=store)
```
zhiltsov-max: It looks like this function should be able to receive config overrides as well. For instance, this implementation will ignore `--insecure` at the moment. Logger and version check are not controllable as well.

cvat-sdk/cvat_sdk/core/auth.py
```diff
@@ -43,13 +55,150 @@ def get_auth_store_path() -> Path:
...
-    def __init__(self, config_file_path: Path | None = None) -> None:
-        self._path = (
-            Path(config_file_path) if config_file_path is not None else get_auth_store_path()
-        )
+    def __init__(
+        self,
+        path: Path | None = None,
+        *,
+        config_file_path: Path | None = None,
+    ) -> None:
```
zhiltsov-max: Why we need this parameter?

tests/python/sdk/test_auth_helpers.py
```diff
@@ -0,0 +1,220 @@
...
+def test_default_auth_factory_reads_env_token(monkeypatch):
+    monkeypatch.setenv(CVAT_ACCESS_TOKEN_ENV_VAR, "pat-123")
+    cred = default_auth_factory()("https://app.cvat.ai")
+    assert isinstance(cred, AccessTokenCredentials)
+    assert cred.token == "pat-123"
+
+
+def test_resolve_server_host_prefers_explicit(tmp_path):
+    store = AuthStore(path=tmp_path / "auth.json")
+    assert resolve_server_host("https://explicit", store) == "https://explicit"
```
zhiltsov-max: Please avoid using any names that can be real servers.

cvat-sdk/cvat_sdk/core/auth.py
```diff
@@ -43,13 +55,150 @@ def get_auth_store_path() -> Path:
...
+    """
+
+    token = os.getenv(CVAT_ACCESS_TOKEN_ENV_VAR)
+    if token is not None:
+        return lambda _: AccessTokenCredentials(token)
+
+    return get_auth_factory(getpass.getuser())
+
+
+def resolve_server_host(cli_value: str | None, store: AuthStore) -> str:
```
zhiltsov-max: I can't see this function is used anywhere but tests, can we remove or use it?

cvat-cli/src/cvat_cli/_internal/common.py
```diff
@@ -35,66 +33,30 @@ class CriticalError(Exception):
...
             User and password to use for authentication;
-            defaults to the current user and supports the PASS
-            environment variable or password prompt.
-            A Personal Access Token (PAT) can be generated on the server
-            and specified in the {} environment variable instead.
-            (default user: {}).
-        """).format(CVAT_ACCESS_TOKEN_ENV_VAR, getpass.getuser()),
+            supports the PASS environment variable or a password prompt.
+            A Personal Access Token (PAT) can be supplied via the {} environment
+            variable, or saved as a profile (see 'cvat-cli profile').
```
zhiltsov-max: `cvat-cli profile` is not defined in this PR.

Pr1ncee: it'll be delivered in the next PR, should I remove it until it?

zhiltsov-max: Yes.

cvat-cli/src/cvat_cli/_internal/common.py
```diff
@@ -135,24 +58,21 @@ def configure_logger(logger: logging.Logger, parsed_args: argparse.Namespace) ->
...
-        config=config,
-        check_server_version=False,  # version is checked after auth to support versions < 2.3
+    auth_args = ClientAuthParameters(
+        profile=popattr(parsed_args, "profile"),
+        insecure=popattr(parsed_args, "insecure"),
+        server_host=popattr(parsed_args, "server_host"),
+        server_port=popattr(parsed_args, "server_port"),
+        auth=popattr(parsed_args, "auth"),
+        organization=popattr(parsed_args, "organization"),
     )
```
zhiltsov-max: I can see `ClientAuthParameters` already has `from_namespace`, but it's not used here. Why it's not used?

cvat-sdk/cvat_sdk/core/auth.py
```diff
@@ -38,18 +51,244 @@ class ProfileEntry:
     """ISO-8601 UTC timestamp of when the profile was saved."""
 
 
+@attrs.frozen
```
zhiltsov-max: Why it has to be `frozen`?

Pr1ncee: It doesn’t have to be frozen; attrs.frozen just makes ProfileEntry a safer immutable snapshot of persisted auth data rather than something callers can mutate and mistake for an auth-store update.

cvat-sdk/cvat_sdk/core/auth.py
```diff
@@ -38,18 +51,244 @@ class ProfileEntry:
...
+    if not explicit_host and not explicit_cred:
+        default = store.get_default_profile()
+        if default is not None:
+            client = make_client_from_profile(
+                default[1],
+                logger=logger,
+                config=client_config,
+                check_server_version=check_server_version,
+            )
+            return _set_organization(client, params.organization)
```
zhiltsov-max: Could you refactor this to avoid duplicating logic?

cvat-sdk/cvat_sdk/core/auth.py
zhiltsov-max: I doubt this is a real concern. This object doesn't have any `save()` method. Making this class frozen can complicate creation of objects for those, who don't use the factory function.

cvat-sdk/cvat_sdk/core/auth.py
```diff
@@ -38,18 +51,244 @@ class ProfileEntry:
...
+        return attrs.evolve(config, verify_ssl=False)
+
+    return config
+
+
+def _get_profile_or_raise(store: AuthStore, name: str) -> ProfileEntry:
+    profile = store.get_profile(name)
+    if profile is None:
+        raise AuthStoreError(
+            f"Unknown profile {name!r}. Run 'cvat-cli profile list' to see saved profiles."
```
zhiltsov-max: I don't think we have `cvat-cli profile list` in the CLI now.

cvat-sdk/cvat_sdk/core/auth.py
```diff
@@ -38,18 +51,261 @@ class ProfileEntry:
...
+    if not explicit_host and not explicit_cred:
+        default = store.get_default_profile()
+        if default is not None:
+            return _make_client_from_profile(
+                default[1],
+                logger=logger,
+                config=client_config,
+                check_server_version=check_server_version,
+                organization=params.organization,
+            )
```
zhiltsov-max: I can see the `_make_client_from_profile` function is still used twice in this code fragment. Please change the code to use it only once.

cvat-sdk/cvat_sdk/core/auth.py
```diff
@@ -38,18 +51,261 @@ class ProfileEntry:
...
 
-    def __init__(self, config_file_path: Path | None = None) -> None:
-        self._path = (
-            Path(config_file_path) if config_file_path is not None else get_auth_store_path()
-        )
+    def __init__(
+        self,
+        path: Path | None = None,
+        *,
+        config_file_path: Path | None = None,
```
zhiltsov-max: I can see the parameter is still here and my question wasn't answered, yet the thread is marked resolved twice.

cvat-sdk/cvat_sdk/core/auth.py
Pr1ncee: What's the reason to do it? I pass 2 different arguments each time. I don't think it'll bring more readability to the code to be honest

cvat-sdk/cvat_sdk/core/auth.py
Pr1ncee: I apologize. I thought you were asking about typehint parameter

Pr1ncee: for now it lets tests or callers point AuthStore at a non-default auth.json location instead of the real user config path from platformdirs, so tests can use tmp_path and CLI code can support an override without touching the user’s actual credentials.

cvat-sdk/cvat_sdk/core/auth.py
zhiltsov-max: As I said, currently the invocation logic is duplicated for no reason. The duplication can easily be removed by extracting a `profile` variable produced by an if-else block before the function call.

cvat-sdk/cvat_sdk/core/auth.py
zhiltsov-max: I'm not sure I understood the answer. Both arguments are handled the same way inside the constructor, and `config_file_path` is not used anywhere else in this PR. You can point to a custom file location by using any of the arguments.

Pr1ncee: ok, I removed it

cvat-sdk/cvat_sdk/core/auth.py
```diff
@@ -38,18 +51,247 @@ class ProfileEntry:
...
+    check_server_version: bool,
+    organization: str | None,
+) -> Client:
+    client = make_client_from_profile(
+        profile,
+        logger=logger,
+        config=config,
+        check_server_version=check_server_version,
+    )
+    return _set_organization(client, organization)
```
zhiltsov-max: Can we remove this function? It's used only once and it just forwards the arguments.

Pr1ncee: done

tests/python/sdk/test_auth_helpers.py
```diff
@@ -0,0 +1,272 @@
...
+    store = AuthStore(path=tmp_path / "auth.json")
+    store.add_profile(
+        "p",
+        ProfileEntry(server=PROFILE_URL, token="pat", created_date="2026-01-01T00:00:00+00:00"),
+    )
+    monkeypatch.setattr(auth_mod, "Client", _FakeClient)
+    client = make_client_from_cli(_ns(profile="p"), store=store)
+    assert client.url == PROFILE_URL
+    assert client.logged_in_with.token == "pat"
+
```
zhiltsov-max: Could you clarify what this test is testing? I assume it's supposed to test that an existing profile can be used. If so, please name the test this way.

tests/python/sdk/test_auth_helpers.py
```diff
@@ -0,0 +1,272 @@
...
+
+def test_make_client_from_cli_profile_conflicts(tmp_path):
+    store = AuthStore(path=tmp_path / "auth.json")
+    with pytest.raises(AuthStoreError, match="mutually exclusive"):
+        make_client_from_cli(_ns(profile="p", server_host=EXPLICIT_URL), store=store)
+    with pytest.raises(AuthStoreError, match="mutually exclusive"):
+        make_client_from_cli(_ns(profile="p", auth=get_auth_factory("u:p")), store=store)
+
```
zhiltsov-max: Please make the test name more clear about what is supposed to be in a conflict.

tests/python/sdk/test_auth_helpers.py
```diff
@@ -0,0 +1,272 @@
...
+        _ns(profile="p", insecure=True),
+        logger=logger,
+        config=config,
+        check_server_version=True,
+        store=store,
+    )
+
+    assert client.url == PROFILE_URL
+    assert client.config.status_check_period == 0.1
+    assert client.config.verify_ssl is False
```
zhiltsov-max, line 206: I'd say it's not really clear now what is the priority order between the `parsed_args` and kw args, e.g. `config`. I think it should be at least specified in the function docstring. Avoiding this conflict would even be better, if you see a way to do this.

zhiltsov-max (approved):

PR #10843, opened 2026-06-26
Update test packages to enable running tests with Python 3.14

SpecLad:
### Motivation and context

Our Helm tests currently use the default Python version in `ubuntu-latest`, which will become 3.14 once `ubuntu-latest` becomes Ubuntu 26.04. We should make sure the tests are ready for that.

Some notes:

* The `numba` pin should not be necessary anymore, since the problem it works around was fixed in coverage 7.6.1, and I've already updated coverage to 7.14.0 in 76e15a01e0.

* There is no version of NumPy that supports both 3.10 and 3.14. I've elected to use the lowest version that supports each, since it helps ensure that we didn't make the SDK dependent on a newer version by accident.

* pytest 9.0.3 is not the latest version, but the newest one is currently incompatible with pytest-cases.

### Checklist

### License

### DIFF
--- a/tests/python/requirements.txt
+++ b/tests/python/requirements.txt
@@ -1,21 +1,16 @@
 allure-pytest==2.14.2
 av==16.0.1
-pytest==6.2.5
-pytest-cases==3.8.6
-pytest-timeout==2.1.0
-pytest-cov==4.1.0
-pytest-xdist==3.5.0
+pytest==9.0.3
+pytest-cases==3.10.1
+pytest-timeout==2.4.0
+pytest-cov==7.1.0
+pytest-xdist==3.8.0
 requests==2.33.0
 deepdiff==8.6.2
 boto3==1.35.95
 Pillow==12.2.0
 python-dateutil==2.8.2
-pyyaml==6.0.2
-numpy==2.0.0
+pyyaml==6.0.3
+numpy==2.3.2; python_version >= "3.14"
+numpy==2.0.0; python_version < "3.14"
 librosa==0.11.0
-
-# a librosa dep, Higher versions conflict with coverage,
-# https://github.com/numba/numba/issues/10239
-numba==0.61
-
-# TODO: update pytest to 7.0.0 and pytest-timeout to 2.3.1 (better debug in vscode)
--- a/tests/python/requirements.txt
+++ b/tests/python/requirements.txt
@@ -1,6 +1,6 @@
 allure-pytest==2.14.2
 av==16.0.1
-pytest==9.0.3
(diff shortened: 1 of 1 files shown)


tests/python/requirements.txt
```diff
@@ -1,21 +1,16 @@
 allure-pytest==2.14.2
 av==16.0.1
-pytest==6.2.5
-pytest-cases==3.8.6
-pytest-timeout==2.1.0
-pytest-cov==4.1.0
-pytest-xdist==3.5.0
+pytest==9.0.3
+pytest-cases==3.10.1
```
zhiltsov-max, line 4:
```suggestion
pytest-cases==3.10.1  # pytest 9.x is supported from 3.10
```

tests/python/requirements.txt
```diff
@@ -1,21 +1,16 @@
 allure-pytest==2.14.2
 av==16.0.1
-pytest==6.2.5
-pytest-cases==3.8.6
-pytest-timeout==2.1.0
-pytest-cov==4.1.0
-pytest-xdist==3.5.0
+pytest==9.0.3
```
zhiltsov-max:
```suggestion
pytest==9.0.3  # https://github.com/smarie/python-pytest-cases/issues/385
```

SpecLad: Yeah, fair enough.

tests/python/requirements.txt
SpecLad: I don't think this information really matters in the long run. I doubt anyone would want to downgrade it.

zhiltsov-max:
Running with 26.04 is tested in https://github.com/cvat-ai/cvat/pull/10809

tests/python/requirements.txt
```diff
@@ -1,21 +1,16 @@
...
 av==16.0.1
-pytest==6.2.5
-pytest-cases==3.8.6
-pytest-timeout==2.1.0
-pytest-cov==4.1.0
-pytest-xdist==3.5.0
+pytest==9.0.3  # https://github.com/smarie/python-pytest-cases/issues/385
+pytest-cases==3.10.1
+pytest-timeout==2.4.0
+pytest-cov==7.1.0
```
zhiltsov-max, line 6: We might need to update the excludes per 7.x [migration notes](https://coverage.readthedocs.io/en/latest/migrating.html#migrating-to-coverage-py-7-x) later. They [added](https://coverage.readthedocs.io/en/latest/config.html#config-report-exclude-also) `exclude_also` to extend the [defaults](https://coverage.readthedocs.io/en/latest/excluding.html#default-exclusions).

zhiltsov-max (approved):

PR #10849, opened 2026-06-30
Fix `Cannot read properties of null (reading 'toLowerCase')` on requests page

klakhov:
### Motivation and context

The Requests API (GET /api/requests/{id} and GET /api/requests) could return a request with status: null. On the client this crashed the UI with:

`Cannot read properties of null (reading 'toLowerCase')`
at Request.status in cvat-core/src/request.ts, because the getter does this.#status.toLowerCase().

Root cause: RequestSerializer.status resolved its value via source="get_status", which calls RQ's Job.get_status() with the default refresh=True — a fresh hget(key, 'status') against Redis at serialization time. The job is fetched once in the view (so its status is already loaded), but the serializer then re-reads it. If the job's Redis hash loses its status field in that window — most plausibly a TTL-expiry race for a finished/failed job near the end of its result_ttl/failure_ttl, or a partially-evicted hash — the read returns None, which serializes to null.

Fix (server, primary):

Added a RequestStatusField(ChoiceField) that overrides get_attribute to call get_status(refresh=False), reusing the status already loaded when the job was fetched instead of issuing a second, racy Redis read. It remains a ChoiceField, so the generated OpenAPI/SDK schema is unchanged.
get_message likewise switched to get_status(refresh=False).

### Checklist

### License

### DIFF
--- a/changelog.d/20260630_000000_klakhov_fix_request_null_status.md
+++ b/changelog.d/20260630_000000_klakhov_fix_request_null_status.md
@@ -1,4 +1,4 @@
 ### Fixed
 
 - Requests API no longer returns a `null` status when a job's status cannot be
-  read from Redis (e.g. the job expired)
+  read from Redis (e.g. the job expired) (<https://github.com/cvat-ai/cvat/pull/10849>)
--- a/changelog.d/20260630_000000_klakhov_fix_request_null_status.md
+++ b/changelog.d/20260630_000000_klakhov_fix_request_null_status.md
@@ -1,0 +1,4 @@
+### Fixed
+
+- Requests API no longer returns a `null` status when a job's status cannot be
+  read from Redis (e.g. the job expired)
--- a/cvat/apps/redis_handler/serializers.py
+++ b/cvat/apps/redis_handler/serializers.py
@@ -34,6 +34,16 @@ class RequestStatus(TextChoices):
     FINISHED = "finished"
 
 
+class RequestStatusField(serializers.ChoiceField):
+    def get_attribute(self, instance: CustomRQJob) -> str | None:
+        # Reuse the status loaded when the job was fetched instead of re-reading it from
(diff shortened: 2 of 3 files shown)


cvat/apps/redis_handler/serializers.py
```diff
@@ -83,7 +93,7 @@ class RequestSerializer(serializers.Serializer):
     # Marking them as read_only leads to generating type as allOf with one reference to RequestStatus component.
     # The client generated using openapi-generator from such a schema contains wrong type like:
     # status (bool, date, datetime, dict, float, int, list, str, none_type): [optional]
-    status = serializers.ChoiceField(source="get_status", choices=RequestStatus.choices)
+    status = RequestStatusField(choices=RequestStatus.choices)
```
SpecLad, line 95: Could you use a `SerializerMethodField` instead, like with `message`? It would be more consistent.

SpecLad: Wait, never mind, I just read the comment above. :facepalm:

klakhov: The approach with SerializerMethodField will also work, but it will change schema a bit. 
I dont really know what is better here :thinking:

SpecLad: Let's just go back to your original version. I assume the comment on the field is still correct, so it's better not to mess up the SDK codegen.

changelog.d/20260630_000000_klakhov_fix_request_null_status.md
```diff
@@ -0,0 +1,4 @@
+### Fixed
+
+- Requests API no longer returns a `null` status when a job's status cannot be
+  read from Redis (e.g. the job expired)
```
SpecLad: Please add a link to the PR.

SpecLad (approved):

PR #10853, opened 2026-07-01
Fix `Cannot read properties of undefined (reading 'occluded')`

klakhov:
### Motivation and context

Fixes a crash when placing a pasted mask if the brush toolset is reconfigured while the paste is still in progress:

```
Cannot read properties of undefined (reading 'occluded')
```

**Reproduction:**
1. Create a mask and finish it.
2. Copy it (start a paste).
3. Press the brush tool in the toolbar before clicking to place the mask.
4. Click on the canvas to place it.

**Root cause:** pasting a mask starts the insertion pipeline in `MasksHandlerImpl.draw()`, which sets `isInsertion = true` and stores `this.drawData` (containing `initialState`). While that insertion is still live, the brush toolset legitimately re-issues `canvasInstance.draw({ enabled: true, shapeType: 'mask', ... })` on brush size/tool changes  but *without* `initialState`. That second call fell through to the drawing branch and unconditionally ran `this.drawData = drawData`, wiping out `initialState` while `isInsertion` stayed `true`. The next `mouse:down` then read `this.drawData.initialState.occluded` on `undefined` and threw.

This only affects masks because they are the only shape whose toolset reconfigures the active `draw()` mid-insertion; insertion also reports as `Mode.DRAW`, so only the canvas (which owns the private `isInsertion` flag) can distinguish a paste from a fresh draw. The fix therefore lives in `cvat-canvas` and makes `draw()` robust to reconfiguration calls during an active insertion:
- capture whether an insertion was already running before the call,
- skip the brush/drawing pipeline init while inserting,
- preserve the insertion's `drawData` (and its `initialState`) until the insertion actually ends.

### Checklist

### License


bsekachev (approved):
lgtm

PR #10855, opened 2026-07-01
fix(engine): return 400 when task metadata requested before media_type is set

Pr1ncee:
#### Motivation                                                                                                                                                    

  Production is throwing 500s on GET /api/tasks/<id>/data/meta:
```python
UnboundLocalError: cannot access local variable 'media' where it is not associated with a value
  File "cvat/apps/engine/views.py", line 1935, in metadata
    for item in media
```
 #### Root cause
  
  TaskViewSet.metadata() dispatches on db_task.media_type through an if / elif / elif chain:
```python
if db_task.media_type == AUDIO and mode == INTERPOLATION:
    media = [db_data.audio]; chapters = None; def serialize_media_item(...)
elif db_task.media_type in (IMAGE, POINT_CLOUD):
    media = ...; chapters = ...
elif db_task.media_type:            # ← truthiness guard
    assert False, f"Unknown media type '{db_task.media_type}'"
# no else
```
  Task.media_type is declared CharField(..., default="", blank=True), so a freshly-created Task whose Data row exists but whose media has not been populated yet has media_type == "". That value is falsy, so none of the branches execute and media, chapters, and serialize_media_item are never bound. The subsequent for item in media raises UnboundLocalError and DRF returns a 500.

The empty-media state was already known at the top of the same method — the prefetch() match block explicitly handles ("", "") as a no-op — but that handling was never carried through to the body of the view. The regression was introduced when the audio media type was added https://github.com/cvat-ai/cvat/pull/10551, which restructured the previous if/else into the current chain and put a truthy guard on the final branch.

 #### Fix

  Raise ValidationError early, mirroring the neighbouring db_data is None check just a few lines above:
```python
  db_data = db_task.data
  if db_data is None:
      raise ValidationError("Data is not uploaded for the task yet")

  if not db_task.media_type:
      raise ValidationError("Task has no media data yet")
  ```
  This turns the silent 500 into an actionable 400 with a clear message, is consistent with the existing "data not uploaded" pattern, and requires no changes to the dispatch chain below.

### Checklist

### License

### CODE REVIEW

File: cvat/apps/engine/views.py [hunk 1]
```diff
@@ -1967,6 +1967,9 @@ def prefetch():
         if db_data is None:
             raise ValidationError("Data is not uploaded for the task yet")
 
+        if not db_task.media_type:
```
zhiltsov-max: I think it will be better to merge this check with the `if` block above, they check for the same problem.

cvat/apps/engine/tests/test_rest_api.py
```diff
@@ -2776,6 +2776,52 @@ def test_api_v1_tasks_data_meta_updated_date(self):
             self.assertLess(res.data["updated_date"], res2.data["updated_date"])
 
 
+class TaskDataMetaGetAPITestCase(ApiTestBase):
```
zhiltsov-max: The tests should be in `tests/python/rest_api/test_tasks.py`.

changelog.d/20260701_000000_andrei_task_meta_unbound_media.md
```diff
@@ -0,0 +1,7 @@
+### Fixed
+
+- `GET /api/tasks/<id>/data/meta` returned a 500 `UnboundLocalError`
+  for tasks whose `Data` row existed but whose `media_type` was still empty
+  (task created but media not yet attached); the endpoint now returns a
+  400 with a clear message.
+  (<https://github.com/cvat-ai/cvat/pull/10855>)
```
zhiltsov-max: Please make the message less verbose about the implementation details. Speak from the user perpective.

nmanovic:
@codex review

nmanovic:
@claude review

tests/python/rest_api/test_tasks.py
```diff
@@ -571,6 +571,46 @@ def test_cannot_create_data_with_non_http_remote_files(self, admin_user):
...
+                    "./manage.py",
+                    "shell",
+                    "-c",
+                    (
+                        "from cvat.apps.engine.models import Data, Task; "
+                        "db_data = Data.objects.create(image_quality=70); "
+                        f"Task.objects.filter(pk={task.id}).update(data_id=db_data.id, media_type='')"
+                    ),
+                ],
+            )
```
zhiltsov-max: Why can't we just call `create_data`? I think you were using it earlier. Current implementation doesn't check the issue reproducible by the users.

Pr1ncee: yes, I tried to do it, but the test was failing, I decided to move to another appoarch since it's already used in the tests

zhiltsov-max: Can you show how it was failing and where the current approach is used?

Pr1ncee: here it is: https://github.com/cvat-ai/cvat/actions/runs/28526919592/job/84567022802

zhiltsov-max: I see the run fails, but the new test passes. Did you investigate the reason?

Pr1ncee: Yes, here it is: The Task.DoesNotExist in the RQ worker's message points to state leakage from a test that started an upload session but the DB was restored before the upload could complete. TestGetTaskDataMeta only uses restore_db_per_function — no restore_cvat_data_per_function and no restore_redis_ondisk_per_function — so a real create_data(upload_start=True) call leaves disk directories (Data.make_dirs()) and (potentially) Redis upload session state behind that pollute the cloud-storage tests that run after it.

zhiltsov-max (approved):

PR #10857, opened 2026-07-02
Fix `Cannot read properties of undefined (reading 'status')`

klakhov:
### Motivation and context

When Axios fails without an HTTP response — network error, DNS failure, timeout, or request cancellation — `errorData.response` is undefined, so reading `.status` throws "Cannot read properties of undefined". Using ?. short-circuits to undefined, so it correctly falls through to throw generateError(errorData) and surfaces the real error.

### Checklist

### License

### DIFF
--- a/cvat-ui/src/components/global-error-boundary/global-error-boundary.tsx
+++ b/cvat-ui/src/components/global-error-boundary/global-error-boundary.tsx
@@ -47,7 +47,7 @@ function mapStateToProps(state: CombinedState): StateToProps {
 
     return {
         job,
-        serverVersion: (server?.version as string) ?? 'unknown',
+        serverVersion: server.version as string,
         uiVersion: packageVersion.ui,
     };
 }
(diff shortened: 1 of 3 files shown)


cvat-ui/src/components/global-error-boundary/global-error-boundary.tsx
```diff
@@ -47,7 +47,7 @@ function mapStateToProps(state: CombinedState): StateToProps {
 
     return {
         job,
-        serverVersion: server.version as string,
+        serverVersion: (server?.version as string) ?? 'unknown',
```
bsekachev: If this change is related to the pull request, then I have a comment about it. 
If not, let's remove this change. 

If request was not finished, we set `about.initialized` as `true` anyway. 
And then we read properties of nullable `about` (e.g. logo, server name, or description, links in plugins, etc) in different places. 

So, I believe we should fix it somehow else. For example, run request again, or provide placeholder values, I do not know exactly, need to think of pros and cons.

bsekachev (approved):

PR #10860, opened 2026-07-02
Fix AI tools tracking for multiple drawn rectangles

bsekachev:
### Motivation and context

AI tools tracking had two related issues. First, tracker-only setups could fail because the shared canvas interaction listener returned early when no interactor was selected, even though tracking does not require an interactor. Second, tracking initialization only used the first drawn rectangle, so drawing multiple rectangles before finishing setup created only one tracked object.

This change scopes the `activeInteractor` guard to interaction mode, restoring tracking in tracker-only configurations. It also adds support for initializing tracking from multiple drawn rectangles by creating one track per rectangle and registering all returned client IDs in `trackedShapes`, allowing the existing batched tracker flow to continue tracking them together.

### How has this been tested?

Manually

### Checklist

### License

### DIFF
--- a/changelog.d/20260702_000000_codex_track_multiple_rectangles.md
+++ b/changelog.d/20260702_000000_codex_track_multiple_rectangles.md
@@ -1,4 +1,4 @@
 ### Fixed
 
 - Fixed AI tools tracking when starting tracking from multiple drawn rectangles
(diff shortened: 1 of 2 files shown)


bsekachev:
@claude review

PR #10862, opened 2026-07-02
Fix request card layout

### LINKED ISSUE #10158: The description of the request does not fit in the block when the page size is changed.

### Steps to Reproduce

- Open the requests tab
- Change the page size

### Expected Behavior

Correct view

### Possible Solution

_No response_

### Context

Image
Image
Image

### Environment

```Markdown
- MacOS, Chrome
```

lexiv0re:
### Motivation and context

Resolves https://github.com/cvat-ai/cvat/issues/10158

On the smallest responsive breakpoint the request title wraps on the next line pushing other details beyond the card boundaries. I assume that possibly it could affect other sizes if it fails to process the resize events and detects wrong responsive breakpoint in certain scenario as shown in the related issue.

This change updates the layout so the request type is never wrapped but rather displayed with an ellipsis if there's not enough space. The reasoning behind this fix is that the card height is fixed and we never want the request type text to occupy more height as it leads to the content overflow. Also, we cannot fix this by just increasing the width because the request type text can be of different width and even if we adjust it for all the existing options longer ones might be eventually introduced causing the same issue.

Before:
requests_layout_before

After:
requests_layout_after

### How has this been tested?

Tested manually

### Checklist

- [X] I have created a changelog fragment 

### License

### DIFF
--- a/changelog.d/20260702_171804_aleksey.zinovyev_request_card_layout.md
+++ b/changelog.d/20260702_171804_aleksey.zinovyev_request_card_layout.md
@@ -1,0 +1,4 @@
+### Fixed
+
+- Prevent requests cards content overflow
+  (<https://github.com/cvat-ai/cvat/pull/10862>)
--- a/cvat-ui/src/components/requests-page/request-card.tsx
+++ b/cvat-ui/src/components/requests-page/request-card.tsx
@@ -65,6 +65,10 @@ function constructName(operation: Request['operation']): string | null {
     return null;
 }
 
+function constructTypeText(type: string): string {
+    return type.split(':').map((word) => word.charAt(0).toUpperCase() + word.slice(1)).join(' ');
+}
+
 function constructTimestamps(request: Request): JSX.Element {
     const started = dayjs(request.startedDate).format('MMM Do YY, H:mm');
     const finished = dayjs(request.finishedDate).format('MMM Do YY, H:mm');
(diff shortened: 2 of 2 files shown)


bsekachev:
@lexiv0re Do you think there is a reason to keep `height` of `.cvat-requests-card` fixed?

lexiv0re:
@bsekachev I think it's just a UI/UX design preference rather than a strict requirement. It makes all the cards looking consistent when there are multiple on the screen. Also, when resizing nothing is "jumping" vertically as heights are fixed. If you think the dynamic height makes more sense on this page I can work on the updated layout. Clearly it would prevent all issues of such kind. WDYT?

bsekachev:
It might make a little more sense, because not only description in theory can be long. 

For example if we have long username:

image

bsekachev:
@klakhov do you have any opinion?

klakhov:
I would prefer to keep the fixed card height. It looks consistent with other CVAT cards, where variable text is usually truncated with `ellipsis` and shown in a tooltip instead of changing card height.

That said, the current change only handles the request type. The long username example shows that timestamp text can still overflow, so maybe we should additionally apply ellipsis/tooltip to the timestamp rows or owner username. This would keep the card height stable while covering the same class of overflow issues.

lexiv0re:
Thanks, decided to keep the fixed height of the card and wrap all texts with dynamic width into an ellipsis. Will implement the remaining changes shortly.

lexiv0re:
Wrapped other dynamic width texts on the request card with ellipsis.
E.g. with long username:
long_username

bsekachev (approved):

PR #10871, opened 2026-07-06
Audio workspace: consistent duration display, context menu for intervals on the track

lexiv0re:
### Motivation and context

1. Selected interval duration and current position on the track do not display milliseconds and not consistent with the duration in the list of annotations.

These changes update the formatting so durations are displayed consistently and include milliseconds across the audio workspace interface.

AudioDurationFormat

2. No custom context menu is opened when right clicking on an interval on the track.

These changes replace the standard browser context menu with the interval-specific menu consistent with the menu accessible from the list of annotations.

audio-interval-context-menu

### How has this been tested?

Tested manually

### Checklist

### License


changelog.d/20260706_171148_aleksey.zinovyev_audio_workspace_minors.md
```diff
@@ -0,0 +1,5 @@
+### Fixed
+
+- Made duration display consistent across the audio annotation editor.
+- Right-clicking an audio interval opens the custom context menu instead of the standard browser context menu.
+  (<https://github.com/cvat-ai/cvat/pull/10871>)
```
bsekachev: I believe the PR link shall be added to each item

cvat-ui/src/audio/components/annotation-page/audio-workspace/styles/_context-menu.scss
```diff
@@ -0,0 +1,31 @@
...
+    overflow-y: auto;
+    border-radius: 8px;
+    box-shadow: $box-shadow-base;
+
+    .ant-menu {
+        padding: 4px;
+    }
+
+    .ant-menu-item {
+        padding: 5px 12px;
```
bsekachev: usually we use `$grid-unit-size` (8px) with multipliers: 0.5, 1,2,10, etc. for spacing and sizes

cvat-ui/src/components/annotation-page/standard-workspace/objects-side-bar/color-picker.tsx
```diff
@@ -131,6 +132,7 @@ function ColorPicker(props: Props, ref: React.Ref<any>): JSX.Element {
             trigger='click'
             open={typeof visible === 'boolean' ? visible : pickerVisible}
             onOpenChange={changeVisible}
+            onPopupAlign={onPopupAlign}
```
bsekachev: I can't really find this prop in public antd API for Popover (5.*). Shall we use it?

cvat-ui/src/audio/containers/annotation-page/audio-workspace/audio-context-menu.tsx
```diff
@@ -0,0 +1,110 @@
...
+        },
+        onChangeIntervalColor(clientID: number, color: string): void {
+            dispatch(updateAudioIntervalAsync(clientID, { color }));
+        },
+    };
+}
+
+type Props = StateToProps & DispatchToProps;
+
+function AudioContextMenuContainer(props: Props): JSX.Element | null {
```
bsekachev: Container/Component logic is a kind of legacy React. Sometimes in complex components we might need it as it makes things easier, but in the most of cases - we do not

bsekachev: Like this container is simple wrapper which can be removed

cvat-ui/src/audio/components/annotation-page/audio-workspace/audio-region-details.tsx
```diff
@@ -201,7 +193,7 @@ function AudioRegionDetails(props: AudioRegionDetailsProps): JSX.Element {
     const isReadonly = !!interval.lock;
     const start = interval.start / 1000;
     const end = (interval.stop ?? interval.start) / 1000;
-    const duration = Math.max(0, end - start);
+    const durationMs = Math.max(0, (interval.stop ?? interval.start) - interval.start);
```
bsekachev: The logic is not exactly correct here (it was not correct before also).

When interval does not have `stop`, it means that interval ends at the end of timeline.
But we can't just take `interval.start` as a fallback.

cvat-ui/src/audio/components/annotation-page/audio-workspace/utils/audio-interval.ts
```diff
@@ -28,3 +29,10 @@ export function intervalEndSeconds(interval: AudioIntervalState): number {
 export function intervalDurationSeconds(interval: AudioIntervalState): number {
     return Math.max(0, intervalEndSeconds(interval) - intervalStartSeconds(interval));
 }
+
+export function copyAudioIntervalURL(serverID: number | null | undefined): void {
```
bsekachev:
```suggestion
export function copyAudioIntervalURL(serverID?: number | null): void {
```

cvat-ui/src/audio/components/annotation-page/audio-workspace/audio-canvas-wrapper.tsx
```diff
@@ -126,6 +126,13 @@ function AudioCanvasWrapper(): JSX.Element {
             stop: Math.round(stop * 1000),
         }));
     }, [dispatch]);
+    const onIntervalContextMenu =
```
bsekachev: onIntervalContextMenuOpen? from the current name it is not clear what action triggers the callback

changelog.d/20260706_171148_aleksey.zinovyev_audio_workspace_minors.md
lexiv0re: Thanks, updated

cvat-ui/src/audio/components/annotation-page/audio-workspace/styles/_context-menu.scss
lexiv0re: Thanks, updated

cvat-ui/src/audio/components/annotation-page/audio-workspace/utils/audio-interval.ts
lexiv0re: Thanks, simplified

cvat-ui/src/audio/components/annotation-page/audio-workspace/audio-canvas-wrapper.tsx
lexiv0re: This name was inspired by the standard event name "contextmenu" we're subscribing to in the end. But I agree it's a bit unclear so updated it, thanks.

cvat-ui/src/audio/components/annotation-page/audio-workspace/audio-region-details.tsx
lexiv0re: Updated the logic so intervals without explicit stop value are considered to end at the end of the audio track.
Just for my education, how can we create an interval without an explicit stop?

cvat-ui/src/audio/containers/annotation-page/audio-workspace/audio-context-menu.tsx
lexiv0re: I considered merging them into a single component but ended up not to. What I found useful is that here we can return from render function early with
```
if (!interval) {
    return null;
}
```
and `interval` here is being found by `state.audio.player.contextMenu.clientID`. So essentially if no interval has its context menu open we won't be doing any render work further. Besides this it narrows the type of `interval` from `AudioIntervalState || null` to `AudioIntervalState` which simplifies the code in the component as we don't need to handle null-ish case explicitly everywhere we use `interval`. And we cannot return early there because hooks cannot be called conditionally.

So I'd prefer keeping them separate in this case. Or maybe you could suggest how we can avoid the issues I mentioned in the merged version?

cvat-ui/src/reducers/audio-reducer.ts
```diff
@@ -122,6 +128,24 @@ export default function audioReducer(state: AudioState = defaultState, action: A
...
+
+            return {
+                ...state,
+                player: {
+                    ...state.player,
+                    contextMenu: {
+                        visible,
+                        left,
+                        top,
+                        clientID,
```
bsekachev: Do we need both `visible` and `clientID`? Maybe `clientID === null` might replace any `visible` logic?

cvat-ui/src/components/annotation-page/standard-workspace/objects-side-bar/color-picker.tsx
lexiv0re: Good catch, thanks. Reworked using the exposed public API.

cvat-ui/src/audio/components/annotation-page/audio-workspace/audio-context-menu-component.tsx
```diff
@@ -0,0 +1,139 @@
...
+    } = props;
+
+    const menuRef = useRef<HTMLDivElement>(null);
+    const colorPickerPopoverRef = useRef<HTMLElement | null>(null);
+    // track local position adjustments e.g. when the menu is outside of the viewport
+    const [position, setPosition] = useState({ top, left });
+    const [colorPickerVisible, setColorPickerVisible] = useState(false);
+
+    useEffect(() => {
+        setColorPickerVisible(false);
```
bsekachev, line 49: May you please add some clarifying comments to each useEffect? 
E.g. this one as  understand needed to close color picker each time when context menu for another interval has opened

cvat-ui/src/audio/components/annotation-page/audio-workspace/audio-context-menu-component.tsx
```diff
@@ -0,0 +1,139 @@
...
+                colorPickerPopoverRef.current?.contains(target)
+            );
+        };
+
+        const onMouseDown = (event: MouseEvent): void => {
+            if (!isEventInsideMenu(event.target)) {
+                onCloseContextMenu();
+            }
+        };
+        const onContextMenu = (event: MouseEvent): void => {
```
bsekachev:
```suggestion

        const onContextMenu = (event: MouseEvent): void => {
```

cvat-ui/src/audio/components/annotation-page/audio-workspace/audio-region-details.tsx
bsekachev: On UI we currently can't. 
But this rule comes from the server where corresponding IntervalSerializer allows `None` for `end`.

cvat-ui/src/audio/components/annotation-page/audio-workspace/audio-context-menu-component.tsx
lexiv0re: Thanks, updated

cvat-ui/src/audio/containers/annotation-page/audio-workspace/audio-context-menu.tsx
bsekachev: Having an extra component to simplify the logic is up to you. 
Maybe we should not consider it as a container? As containers are legacy React concept. 
This may be just a simple context menu wrapper component. 
And I would suggest to use `useSelector` + `useDispatch` instead mapping functions

cvat-ui/src/reducers/audio-reducer.ts
lexiv0re: Good catch! Not needed in this final version, simplified, thanks.

cvat-ui/src/audio/components/annotation-page/audio-workspace/audio-context-menu-component.tsx
lexiv0re: Sure, added clarifying comments

cvat-ui/src/audio/containers/annotation-page/audio-workspace/audio-context-menu.tsx
lexiv0re: Sure, reworked with `useSelector` + `useDispatch`, renamed to wrapper.

bsekachev (approved):

PR #10874, opened 2026-07-06
Implement agent support for interactors

SpecLad:
### Motivation and context

Closing the feature gap with Nuclio functions.

### How has this been tested?

Tests are in the private repository.

### Checklist

### License


cvat-sdk/cvat_sdk/auto_annotation/interface.py
```diff
@@ -241,6 +242,182 @@ def detect(
...
+    The confidence level. The corresponding attribute value must be a real number between 0 and 1.
+    """
+
+
+@attrs.frozen(kw_only=True)
+class InteractionResultShape:
+    """
+    A shape that can be returned by an auto-annotation interaction function.
+
+    All attributes have the same meaning as attributes of the same name in the `LabeledShape` model.
```
zhiltsov-max:
```suggestion
    All members have the same meaning as members of the same name in the `LabeledShape` model.
```

cvat-cli/src/cvat_cli/_internal/agent_driver_interaction.py
```diff
@@ -0,0 +1,218 @@
...
+        self._contents[key] = _PpImageCacheEntry(pp_image=pp_image)
+
+    def prune(self) -> None:
+        cutoff = datetime.now(tz=timezone.utc) - _MAX_AGE_OF_PP_IMAGE
+
+        while self._contents and next(iter(self._contents.values())).last_accessed_at < cutoff:
+            self._contents.popitem(last=False)
+
+
+_pp_image_cache: _PpImageCache
```
zhiltsov-max, line 69: Not a blocker, but I think we might need to add some ways to control the cache in the future, as it currently can grow without restrictions. Eviction also only happens at insertion, so the allocated memory stays allocated.

site/content/en/docs/api_sdk/sdk/auto-annotation.md
```diff
@@ -280,14 +283,119 @@ cvataa.rectangle(
 )
 ```
 
+### Interaction function protocol
+
+An interaction function is a type of AA function that accepts an image
+and a set of prompts describing an object (or objects) in that image,
+and returns the locations of those objects.
```
zhiltsov-max:
```suggestion
and returns annotations for those objects.
```

cvat-cli/src/cvat_cli/_internal/agent_driver_interaction.py
SpecLad: > Not a blocker, but I think we might need to add some ways to control the cache in the future, as it currently can grow without restrictions.

Yeah, I agree.

> Eviction also only happens at insertion, so the allocated memory stays allocated.

I figured it'd be okay to keep memory allocated for longer than the specified lifetime, as long as we're not allocating any additional memory. It could improve the cache hit rate, and the downside doesn't seem very significant. I suppose we could make this configurable in the future as well.

cvat-sdk/cvat_sdk/auto_annotation/interface.py
SpecLad: What's the difference?

site/content/en/docs/api_sdk/sdk/auto-annotation.md
SpecLad: I feel like that's a bit too broad; interaction functions can't return arbitrary annotations. On the other hand, maybe "locations" is too narrow. What if I say "shapes of those objects" instead? I think it describes it pretty well (and is consistent with the name of the return type).

zhiltsov-max: I think saying "shapes" would work for me. But the fact that we can only return shapes looks like a design oversight to me. It basically means we can't run interactors to preannotate attributes, e.g. text transcriptions.

cvat-sdk/cvat_sdk/auto_annotation/interface.py
zhiltsov-max: I think speaking about attributes is a little bit confusing, given that there are annotation attributes in `LabeledShape` and these attributes are only usable to report confidence scores at the moment. So my suggestion is to use a more established term for class members.

zhiltsov-max (approved):

cvat-sdk/cvat_sdk/auto_annotation/interface.py
SpecLad: Yeah, alright. Updated.

zhiltsov-max (approved):

PR #10879, opened 2026-07-08
Change skeleton control points to be visually distinguishable from annotation

lexiv0re:
### Motivation and context

Skeleton control points look very similar to the points of the annotation itself. It sometimes causes confusion why the control points are not part of the annotation.

1. This change visually differentiates skeleton control points from the annotation points so it's not confusing anymore.

Before:
old_skeleton_cp

After:
new_skeleton_cp

Note: this change affects only skeleton control points display. It doesn't affect the behavior. Also, rect annotations are unaffected as well.

2. Rotation control point offset calculation is changed for all shapes where applicable. Instead of static number it now:
- Depends on the control point size setting
- Compensates current scale factor 

It prevents the rotation control point from overlapping with other control points of a shape.

Before:
ellipse_rot_old

After:
ellipse_rot_new

Note: change of the offset affects the rotation speed in a freeform rotation hence expected rotation values had to be adjusted in tests.

### How has this been tested?

Tested manually

### Checklist

### License

### DIFF
--- a/cvat-canvas/src/scss/canvas.scss
+++ b/cvat-canvas/src/scss/canvas.scss
@@ -328,14 +328,7 @@ g.cvat_canvas_shape_occluded {
     // wrapping rect must not apply transform attribute from selectize.js
     // otherwise it rotated twice, because we apply the same rotation value to parent element (skeleton itself)
     transform: none !important;
-}
-
-.cvat_canvas_shape > .cvat_canvas_skeleton_wrapping_rect {
     visibility: hidden;
-}
-
-.cvat_canvas_shape.cvat_canvas_shape_activated > .cvat_canvas_skeleton_wrapping_rect {
-    visibility: initial;
 }
 
 .cvat_canvas_pixelized {
--- a/cvat-canvas/src/typescript/canvasView.ts
+++ b/cvat-canvas/src/typescript/canvasView.ts
@@ -1290,13 +1290,16 @@ export class CanvasViewImpl implements CanvasView, Listener {
     }
 
     private refreshSkeletonControlPointDecorators(): void {
-        // there can be only one selection on canvas so we handle first element only
-        const [element] = window.document.getElementsByClassName('cvat_canvas_skeleton_select_wrapper');
-        if (!element) {
+        const { clientID } = this.activeElement;
+        if (clientID === null || this.drawnStates[clientID]?.shapeType !== 'skeleton') {
             return;
         }
 
-        const selectContainer = (element as SVG.LinkedHTMLElement).instance as SVG.Container;
+        const skeleton = this.svgShapes[clientID] as SVG.G;
+        const wrappingRect = skeleton.children().find((child: SVG.Element): boolean => (
+            child.hasClass('cvat_canvas_skeleton_wrapping_rect')
+        ));
+        const selectContainer = wrappingRect?.remember('_selectHandler')?.nested as SVG.Container | undefined;
         if (!selectContainer) {
             return;
         }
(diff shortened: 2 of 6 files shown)


bsekachev:
@codex review

cvat-canvas/src/typescript/canvasView.ts
```diff
@@ -1106,7 +1110,239 @@ export class CanvasViewImpl implements CanvasView, Listener {
         pathElement.dmove(-pathElement.width() / 2, -pathElement.height() / 2);
     }
 
+    private getSkeletonResizeHandle(point: SVG.Element): string | null {
```
bsekachev: The function does not use `this`. Shall it be a class method?

cvat-canvas/src/scss/canvas.scss
```diff
@@ -328,16 +328,9 @@ g.cvat_canvas_shape_occluded {
     // wrapping rect must not apply transform attribute from selectize.js
     // otherwise it rotated twice, because we apply the same rotation value to parent element (skeleton itself)
     transform: none !important;
-}
-
-.cvat_canvas_shape > .cvat_canvas_skeleton_wrapping_rect {
     visibility: hidden;
```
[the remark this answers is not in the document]
lexiv0re: `pointer-events: all` on the wrapping group keeps the rect the target for pointer events despite being hidden

bsekachev (commented):
I would suggest to rename overall:

skeleton control point decorator -> skeleton resizer

It is simpler and reflects better the function element implements.

bsekachev (commented):
Did you look at selectize.js? 

It allows to replace default circle handlers

https://github.com/svgdotjs/svg.select.js

cvat-canvas/src/typescript/canvasView.ts
```diff
@@ -1106,7 +1110,239 @@ export class CanvasViewImpl implements CanvasView, Listener {
...
+                });
+
+            this.updateSkeletonControlPointDecoratorView(outerDecorator.node, isHovered);
+            this.updateSkeletonControlPointDecoratorView(innerDecorator.node, isHovered);
+        }
+    }
+
+    private refreshSkeletonControlPointDecorators(): void {
+        // there can be only one selection on canvas so we handle first element only
+        const [element] = window.document.getElementsByClassName('cvat_canvas_skeleton_select_wrapper');
```
bsekachev: Can we refer to this.activeElement + this.svgShapes rather then introducing anchor class `cvat_canvas_skeleton_select_wrapper`?

lexiv0re:
@bsekachev yes I looked at svg.select.js. In v3.0.1 we're currently using the only option to control the shape of the control points is `pointType` and they update the points positions using their hardcoded logic which uses specific methods of the shapes. I had an attempt of passing different shapes (decorators) directly through this but it ended up being too ugly. I also left a comment here: https://github.com/cvat-ai/cvat/pull/10879/changes#diff-6fd88f64dcd8d9f75ba14894ce4a9b90bfced230aba4803ff105694e2a70c3e5R1234-R1236.

I did not consider upgrading to v4 because I thought it would be too invasive as it affects all selection logic in the canvas. But from what I see they now expose `updateHandle` which I believe should help defining decent updates for custom shapes based on their update semantic. So with v4 we'd most likely be able to pass decorators directly.

cvat-canvas/src/typescript/canvasView.ts
lexiv0re: As it's tightly coupled with specific skeleton control logic and it's very speicifc I'd keep it close. Changed it to static method to highlight it doesn't use `this`.

cvat-canvas/src/typescript/canvasView.ts
lexiv0re: Sure, fixed.

lexiv0re:
Applied `skeleton control point decorator` -> `skeleton resizer` renaming across the changes.

bsekachev (approved):

PR #10897, opened 2026-07-14
As/add webhook notification for the rest requests

### LINKED ISSUE #9613: Need a webhook for requests

Rwwwrl:
### Title
Fix #9613: Add webhooks for completed requests

### Motivation and context

Long-running operations currently require clients to poll `/api/requests/{rq_id}`, increasing server and network load and delaying completion handling by up to the polling interval. Add webhook completion events for dataset and backup exports, task data creation, consensus merges, and quality report creation, allowing integrations to react immediately to successful or failed requests. The webhook payload embeds the request in the same format as the requests API response.

Entity webhook payloads are also aligned with the corresponding retrieve endpoints. Create and update events contain the latest persisted resource state, while `before_update` contains the previous values of only the fields changed by the triggering operation. This prevents concurrent partial updates from causing consumers to overwrite unrelated fields with stale values, while preserving operation-specific change information.

`GET /api/webhooks/events` now returns structured metadata for grouping and displaying events.

This is a breaking API change: `create:export` and `create:backup` are replaced by the corresponding completion events, entity webhook payload semantics have changed, and `GET /api/webhooks/events` now returns event objects instead of strings.

### Checklist

- [ ] I have linked related issues (see [[GitHub docs](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)](

### License


changelog.d/20260714_145730_aleksei_sosov_change_webhook_event_responses.md
zhiltsov-max, line 1: Please be brief and keep at most 2 lines per change. The details can be explained in the PR description.

cvat/utils/utils.py
```diff
@@ -0,0 +1,60 @@
...
+from typing import TYPE_CHECKING
+
+import rq
+from crum import get_current_request, get_current_user
+
+if TYPE_CHECKING:
+    from cvat.apps.engine.models import User
+
+
+def get_user(instance=None) -> User | dict | None:
```
zhiltsov-max: Why these functions were moved into this file, outside of the apps?

cvat/apps/webhooks/utils.py
```diff
@@ -6,22 +6,162 @@
...
+        EventGroupDTO(display_name="Consensus merge"),
+    ),
+    (
+        QualityRequestId(target=RequestTarget.TASK, target_id=1).type,
+        EventGroupDTO(display_name="Quality report creation"),
+    ),
+)
+
+
+def retrieve_instance(model: type[ModelT], pk: int) -> ModelT:
```
zhiltsov-max, line 95: Can we avoid code duplication with the owning viewsets somehow?

cvat/apps/webhooks/signals.py
```diff
@@ -32,22 +36,23 @@
...
+def pre_save_resource_event(sender: type[ModelT], instance, **kwargs):
     instance._webhooks_selected_webhooks = []
+    instance._webhooks_old_data = None
 
     if instance.pk is None:
         created = True
     else:
         try:
-            old_instance = sender.objects.get(pk=instance.pk)
+            old_instance = utils.retrieve_instance(model=sender, pk=instance.pk)
```
zhiltsov-max: For this call to be correct, it has to be done within a repeatable read transaction or a more strong one. This implementation is subject to data races. Additionally, for viewsets, it repeats the requests already made at the calling site.

cvat/apps/webhooks/signals.py
```diff
@@ -119,21 +129,26 @@ def post_save_resource_event(sender, instance, created: bool, raw: bool, **kwarg
     old_data = instance._webhooks_old_data
     del instance._webhooks_old_data
 
+    retrieved_instance = utils.retrieve_instance(
+        model=sender,
+        pk=instance.pk,
+    )
+
```
zhiltsov-max: Similar comment as above. I think we should have a way to reuse the cached data, if it's already loaded. Maybe doable via `django_database.is_field_cached()`, but need to test it.

cvat/apps/webhooks/services.py
```diff
@@ -15,31 +17,35 @@
...
     select_for_project: bool = True,
 ) -> list[Webhook]:
     selected_webhooks = []
 
+    queryset = Webhook.objects.filter(
+        Q(events=event_key)
+        | Q(events__startswith=f"{event_key},")
+        | Q(events__endswith=f",{event_key}")
+        | Q(events__contains=f",{event_key},"),
+        is_active=True,
```
zhiltsov-max, line 33: Could you check the performance of this and compare with regex search or other possible implementations? Also, if it has `or`, there must be `distinct()` or `union()` of per-term calls.

cvat/apps/webhooks/serializers.py
```diff
@@ -23,33 +23,45 @@ def get_webhook_type(self, attrs, serializer):
...
+    action = serializers.CharField(read_only=True)
+    resource = serializers.CharField(read_only=True)
+    key = serializers.CharField(read_only=True)
+    group = EventGroupDTOSerializer(read_only=True)
+
+
 class EventsSerializer(serializers.Serializer):
     webhook_type = serializers.ChoiceField(choices=WebhookTypeChoice.choices())
-    events = EventTypesSerializer()
+    events = EventDTOSerializer(many=True, read_only=True)
```
zhiltsov-max: Please try to choose a more problem-oriented name for the class.

Rwwwrl: updated df5a4da29e5a924b19527720f53c21b05639f34f

cvat/apps/webhooks/services.py
Rwwwrl: Double checked how postgres will perform that query

query:
```
BEGIN;

SET LOCAL enable_seqscan = off;

EXPLAIN ANALYZE
SELECT *
FROM webhooks_webhook
WHERE is_active = TRUE
  AND type = 'project'
  AND project_id = 2
  AND (
      events = 'completed:export:annotations'
      OR events LIKE 'completed:export:annotations,%'
      OR events LIKE '%,completed:export:annotations'
      OR events LIKE '%,completed:export:annotations,%'
  );

ROLLBACK;
```

output:
```
Bitmap Heap Scan on webhooks_webhook  (cost=4.18..6.30 rows=3 width=200) (actual time=0.019..0.022 rows=5 loops=1)
  Recheck Cond: (project_id = 2)
"  Filter: (is_active AND ((type)::text = 'project'::text) AND (((events)::text = 'completed:export:annotations'::text) OR ((events)::text ~~ 'completed:export:annotations,%'::text) OR ((events)::text ~~ '%,completed:export:annotations'::text) OR ((events)::text ~~ '%,completed:export:annotations,%'::text)))"
  Heap Blocks: exact=2
  ->  Bitmap Index Scan on webhooks_webhook_project_id_76846b5e  (cost=0.00..4.18 rows=5 width=0) (actual time=0.013..0.013 rows=5 loops=1)
        Index Cond: (project_id = 2)
Planning Time: 0.069 ms
Execution Time: 0.054 ms
```

So as we can see it first hits index on project_id (`Bitmap Index Scan on webhooks_webhook_project_id_76846b5e  (cost=0.00..4.18 rows=5 width=0)`) and then does Bitmap Heap Scan over those 5 rows - `Bitmap Heap Scan on webhooks_webhook  (cost=4.18..6.30 rows=3 width=200) (actual time=0.019..0.022 rows=5 loops=1)`

Since number of webhooks inside a project / organization is not high there are not problems with performance

Rwwwrl: Considering adding `distinct`:
1. I was wrong yesterday, we do have "validation" on uniquiness (we are using set in WebhookWriteSerializer) - https://github.com/cvat-ai/cvat/blob/2b28a45570d2ca8a701af7f202345a3a14fc86eb/cvat/apps/webhooks/serializers.py#L26
2. but even if we have duplicates, it still will return single row, not row per `OR condition`

cvat/apps/webhooks/utils.py
Rwwwrl: If I were to work with fastapi stack I would have something like that:
```
from django.db import models


class Job(models.Model):
    name = models.CharField(max_length=255)


class Task(models.Model):
    name = models.CharField(max_length=255)
    job = models.ForeignKey(Job, on_delete=models.CASCADE, related_name="tasks")


class TaskRepository:
    _retrieve_queryset = Task.objects.select_related("job")

    @classmethod
    def get_by_id(cls, _id: int) -> Task:
        return cls._retrieve_queryset.get(id=_id)

    @classmethod
    def get_all(cls) -> list[Task]:
        return list(cls._retrieve_queryset.all())

    @classmethod
    def update_by_id(cls, _id: int, data: dict) -> Task:
        task = cls.get_by_id(_id)

        for field, value in data.items():
            setattr(task, field, value)

        task.save(update_fields=list(data.keys()))

        return task
```

but since we have DRF, whose viewsets work with querysets (through defining `queryset` or `get_queryset`) I came up with those two ideas:
1. `Task.retrieve_objects.<method>`
2. `Task.objects.with_retrieve_details().<method>`

(Subjectively I like the first option more, whereas second one is a project pattern: `with_job_summary`, `with_issue_counts`)

In future, we might widely use it to remove retrieving instance one more time inside signals (as we discussed before)

For the time being, I would not invent it and keep `retrieve_instance` as it is

changelog.d/20260714_145730_aleksei_sosov_change_webhook_event_responses.md
Rwwwrl: updated 6cd32b37b2a263d998fc90963fcf5067a52ff61f

cvat/utils/utils.py
Rwwwrl: Those function are being used in apps/webhooks. It seems they are completely app agnostic and ideal candidates to be utility functions. Is there a better place where I can put them?

cvat/apps/webhooks/utils.py
zhiltsov-max: In Django, the model's manager is responsible for that by default (`objects` is the default one). Alternatively, we can also extract a static method in the viewset, but I don't like this option very much because of the dependency order.

- Docs: https://docs.djangoproject.com/en/6.0/topics/db/queries/#retrieving-objects
- Example: https://github.com/cvat-ai/cvat/blob/e69df2854a93ccafc5b67a0daead38acc20222a1/cvat/apps/engine/models.py#L878

cvat/schema.yml
```diff
@@ -7922,6 +7922,12 @@ components:
       required:
       - scope
       - timestamp
+    EventGroupDTO:
```
zhiltsov-max: Please try to rename this type as well

cvat/schema.yml
```diff
@@ -7994,15 +8000,20 @@ components:
...
-            $ref: '#/components/schemas/EventsEnum'
+            $ref: '#/components/schemas/WebhooksEvent'
+          readOnly: true
       required:
-      - events
       - webhook_type
     EventsEnum:
       enum:
-      - create:backup
+      - completed:calculate:quality
```
zhiltsov-max: If you want to follow `:` uses in other events, this one should be renamed and split into `completed:quality_check:project`, `completed:quality_check:task`.

cvat/utils/utils.py
zhiltsov-max: I'd keep them as they were or move into webhooks, where they are used.

cvat/schema.yml
nmanovic: Guys, let's review Events enum and vote for naming. All previous events were verb:resource. I would prefer to keep it in the same format. Now it is something new and I'm not sure that it is right.

cvat/utils/utils.py
Rwwwrl: Moved them back

cvat/schema.yml
Rwwwrl: updated: EventGroupDTOSerializer -> EventGroupSerializer

cvat/schema.yml
Rwwwrl: Our convention is to use `request.type` and to not provide them manually in any particular way

That said, quality request returns calculate:quality:
```
_id = QualityRequestId(
    target=RequestTarget.TASK,
    target_id=1,
)
print(_id.render()) # action=calculate&target=task&target_id=1&subresource=quality
print(_id.type) # calculate:quality
```

Rwwwrl: a thread where we discussed that https://docs.google.com/document/d/1VdIzIQ9at64M32hPFUhrpLKHuXttKDpAd39hJhf-2q4/edit?disco=AAAB_n7cbjY

nmanovic: I replied in the thread. I really don't want to influence naming, but it is a part of API and I don't understand the logic behind naming. For me it looks wrong and over completed. Think about requests not like a resource (it is indeed), but about a wrapper to create a resource asynchronously. The same resource can be created in different way synchronously and asynchronously, the name of the event should be the same. Something like "create:quality_report" or something like that. Nobody cares how a resource is created actually. But I'm happy to discuss and accept an alternative point of view.

cvat/apps/webhooks/signals.py
Rwwwrl: Final logic:

image

example of update:task payload:
```
    "payload": {
      "event": "update:task",
      "webhook_id": 48,
      "task": {
        "url": "http://localhost:3000/api/tasks/8",
        "id": 8,
        "name": "task3",
        "project_id": 11,
        "project_name": "project1",
        "owner": {
          "url": "http://localhost:3000/api/users/1",
          "id": 1,
          "username": "aleksei.sosov@cvat.ai",
          "first_name": "Aleksei",
          "last_name": "Sosov"
        },
        "assignee": null,
        "bug_tracker": "",
        "created_date": "2026-07-19T08:01:59.219093Z",
        "updated_date": "2026-07-19T08:01:59.585566Z",
        "overlap": null,
        "segment_size": 0,
        "status": "annotation",
        "data_cloud_storage_id": 1,
        "guide_id": null,
        "size": 0,
        "data": 8,
        "dimension": "",
        "mode": "",
        "media_type": "",
        "subset": "Train",
        "organization_id": null,
        "organization": null,
        "target_storage": {
          "id": 38,
          "location": "local",
          "cloud_storage_id": null
        },
        "source_storage": {
          "id": 37,
          "location": "local",
          "cloud_storage_id": null
        },
        "jobs": {
          "count": 0,
          "completed": 0,
          "validation": 0,
          "url": "http://localhost:3000/api/jobs?task_id=8"
        },
        "labels": {
          "url": "http://localhost:3000/api/labels?task_id=8"
        },
        "assignee_updated_date": null,
        "validation_mode": null,
        "consensus_enabled": false
      },
      "sender": {
        "url": "http://localhost:3000/api/users/1",
        "id": 1,
        "username": "aleksei.sosov@cvat.ai",
        "first_name": "Aleksei",
        "last_name": "Sosov"
      },
      "before_update": {
        "updated_date": "2026-07-19T08:01:59.419386Z"
      },
      "changes": {
        "updated_date": {
          "from": "2026-07-19T08:01:59.419386Z",
          "to": "2026-07-19T08:01:59.585566Z"
        }
      }
    }
```

1. in "task" we return D + B state
2. in "before_update": changes in old format (D - A changes), kept it only for backward compatability
3. in "changes": cnanges in new format (D - A changes)

in `create:_entity_` event we do not send "before_update" and "changes"

Expected logic on consumer side:
if consumer mirrors task state on his side then they should update its state with the state in "task"
if consumer have logic like "once task status has moved from "A" to "B"' , they should check for "changes"

Expected logic on consumer side before those changes:
If consumer mirrors task state on his side then they should update its state with values from "task" payload for fields from "before_update"
if consumer have logic like "once task status has moved from "A" to "B"' , they should check for "before" state in "before_update" and "after" state in "task"

Q: Do those changes broke that setup?
A: No, it does not. If a consumer continues with its previous logic then the only change they will have - instead of getting D state at "_if consumer mirrors task state on his side then they should update its state with the state in "task"_" logic they will get D + B state, that they will get anyway once a webhook with B`s changes appears.

Q: previous logic of "before_updates" returns diff of serializer(D_state) - serializer(A_state), did not we break that?
A:
1. previous logic did not return D - A state, instead it returned D - A` state where A` is a state that object had at "pre_save" moment, not at a moment when the object was fetched. It leads to major bug: "before_changes" should have had changes inside a operation.
2. Unfortunately, there is no buildin/libs solutions for having serialization of a A state, where the A state is not just a state of a particular entity, that is state of all entities related to the entity that we serialize. Said that, the most of what we can do: check dirty changes ONLY of an entity we are going to serialize and put only them into "before_update" and "changes"


Choosing library for getting dirty changes:
1. https://pypi.org/project/django-model-utils/ - abandoned library, does not support django 5.2, python > 3.12
2. https://pypi.org/project/django-dirtyfields/ - supports python 3.14, django6.0. Last release - January 2026. That library I actually picked
3. https://pypi.org/project/django-lifecycle/ - creates their own "save" signals, that, unfortunately, also does not give us full A state
4. https://pypi.org/project/django-tracking-model/ - fork of https://pypi.org/project/django-model-utils/, does not a have major version - 0.1.8

cvat/schema.yml
Rwwwrl: The idea is to split: 
1. you can subscribe to resources updates (create/update/delete:<entity>)
2. you can subscribe to requests completions (important: not to result that it produces, but to "that operation is completed")

Lets imagine we go with "create:quality_report" as a result of create quality report request:
1. What if later we want to have webhook over "orm object QualityReport was created" - that is exactly where I would go with `create:quality_report`, even more that continues 1th idea
2. What if `create quality report` request is failed? We should send any particular event in that case, since user basically wants to know when operation is completed (either failed or successed)
3. Semantically, if we have "I have operation X that produces Y object, I want to know when that operation is completed, I like much more "X is completed, X.payload.result_id is Y.id" than "Y is created (that is a signal that X is finished)"
4. What if we have request that does not produce any object, but just performs some work and we want to know when that work is finished? It also will not land into resource-first idea

having said that, I had to chose a naming for those operations. To not introduce new words, I am reusing `request.type`

nmanovic: @Rwwwrl, after thinking about it more, I agree with the main distinction you are making: if the goal of this feature is to replace polling of `/api/requests/{rq_id}`, then the webhook should represent the terminal state of the request, not only the resource that may be produced by it.

A request rejected synchronously because of invalid parameters is a different case: no asynchronous request has been accepted, so no webhook is needed. However, once CVAT accepts an operation and returns an `rq_id`, the consumer needs to know whether it eventually finished or failed. Otherwise, the webhook does not fully replace polling.

I also agree that the resource-first model does not cover all cases. A quality report can produce a persistent resource, but a consensus merge or task data creation mainly modifies existing state, and an export may only produce a temporary downloadable artifact.

My remaining concern is the public event taxonomy. Names such as `completed:create:task` or `completed:calculate:quality` do not clearly communicate that these are request lifecycle events and can be confused with domain resource events such as `create:task`.

At the same time, I agree that we need granular subscriptions. A consumer may want to receive only task creation request completions and not every completed request.

What about representing this as a single event key composed of the base request lifecycle event and the request operation type:

```text
complete:request[create:task]
complete:request[export:dataset]
complete:request[calculate:quality]
complete:request[merge:task]
```

This is not intended to be a separate filtering language. Each value is simply one selectable event key, similar to the granular events in the current proposal.

This approach would give us both:

1. Clear semantics: it is explicitly a request completion event.
2. Granular subscriptions: consumers can select only the request operation types they need.

The value inside brackets would correspond to `request.operation.type`, while `request.status` in the payload would indicate whether the request `finished` or `failed`.

Resource lifecycle events would remain separate, for example:

```text
create:quality_report
create:task
update:task
```

A successful operation may therefore emit both events:

```text
complete:request[calculate:quality]
create:quality_report
```

The first event says that the asynchronous request has completed. The second says that a new domain resource has been created. They serve different integration use cases.

Rwwwrl: Agree, updated the namings in 49e9b9c013dae5b9b3b1c075f0749dd0b5b3909e

cvat/schema.yml
```diff
@@ -8026,9 +8036,14 @@ components:
       - update:task
       type: string
       description: |-
-        * `create:backup` - CREATE:BACKUP
+        * `completed:requests[calculate:quality]` - COMPLETED:REQUESTS[CALCULATE:QUALITY]
```
nmanovic:
```suggestion
        * `complete:request[calculate:quality]`
```

I'm not sure why complete[d]:request[s] and duplicate it using upper case.

nmanovic: Also, Maxim proposed another approach. Let's think if we can follow it.

cvat/apps/webhooks/signals.py
zhiltsov-max: So far, settled on returning D+B and adding a new payload item with `after_update` state. Together, this results in:

1. The webhook receiver can rely on the `before_update` and `after update` fields to sync their own state with CVAT state on the modified fields.
2. The webhook receiver can take CVAT state from the whole payload, regardless of the parallel transactions at the CVAT side. Note that the payload state fields may be different from what is in the `after_update` because of the parallel transactions. The `updated_date` field in the payload can typically be used to invalidate the receiver state.

cvat/schema.yml
```diff
@@ -7994,15 +8000,20 @@ components:
...
     EventsEnum:
       enum:
-      - create:backup
+      - completed:request[calculate:quality]
+      - completed:request[create:task]
+      - completed:request[export:annotations]
+      - completed:request[export:backup]
+      - completed:request[export:dataset]
+      - completed:request[merge:job]
+      - completed:request[merge:task]
```
zhiltsov-max, line 8015: Why we still have:
- `calculate:quality` instead of `quality_check:project` and `quality_check:task`
- `export:annotations` together with `export:dataset`
?

cvat/utils/schemas.py
```diff
@@ -0,0 +1,6 @@
+import attrs
+
+
+@attrs.define(frozen=True)
+class DTO:
```
zhiltsov-max: Why do we need this class?

cvat/schema.yml
zhiltsov-max: I don't think we have any other established convention for webhooks except the existing event naming in the `/api/webhooks/events` endpoint. What you're showing are some implementation details. They are observable in the API responses and ids for requests indeed, but they are not supposed to interacted with directly by users - the ids are returned by the server at request creation and supposed to be used as is to retrieve request statuses or results.

cvat/utils/schemas.py
Rwwwrl: Deleted

cvat/apps/webhooks/event_handlers.py
```diff
@@ -3,64 +3,101 @@
...
+@receiver(request_succeeded, sender=BackupExporter)
+@receiver(request_failed, sender=BackupExporter)
+@receiver(request_succeeded, sender=TaskCreator)
+@receiver(request_failed, sender=TaskCreator)
+@receiver(request_succeeded, sender=QualityReportRQJobManager)
+@receiver(request_failed, sender=QualityReportRQJobManager)
+@receiver(request_succeeded, sender=MergingManager)
+@receiver(request_failed, sender=MergingManager)
+def enqueue_request_completion_webhooks(
     sender: Any,
```
zhiltsov-max, line 41: Probably it makes sense to add a comment about the excluded requests, otherwise it looks like they are forgotten.

changelog.d/20260714_145730_aleksei_sosov_change_webhook_event_responses.md
```diff
@@ -0,0 +1,21 @@
...
+  (<https://github.com/cvat-ai/cvat/pull/10897>)
+
+- \[Server API\] Added a `changes` field to update webhook payloads
+  (<https://github.com/cvat-ai/cvat/pull/10897>)
+
+### Changed
+
+- \[Server API\] `GET /api/webhooks/events` now returns objects with `key` and
+  `group.display_name` instead of plain event key strings
+  (<https://github.com/cvat-ai/cvat/pull/10897>)
```
zhiltsov-max, line 11:
```suggestion
  (<https://github.com/cvat-ai/cvat/pull/10897>)

- \[Server API\] `GET /api/webhooks/events` now returns objects with `key` and
  `group.display_name` instead of plain event key strings
  (<https://github.com/cvat-ai/cvat/pull/10897>)
```

changelog.d/20260714_145730_aleksei_sosov_change_webhook_event_responses.md
```diff
@@ -0,0 +1,21 @@
+### Added
+
+- \[Server API\] Added request-completion webhooks with requests API payloads
+  (<https://github.com/cvat-ai/cvat/pull/10897>)
+
```
zhiltsov-max:
```suggestion
- Added webhooks for async requests
  (<https://github.com/cvat-ai/cvat/pull/10897>)

```

I don't think it's just about the server, the UI is also updated in this PR.

changelog.d/20260714_145730_aleksei_sosov_change_webhook_event_responses.md
```diff
@@ -0,0 +1,21 @@
...
+  `group.display_name` instead of plain event key strings
+  (<https://github.com/cvat-ai/cvat/pull/10897>)
+
+### Removed
+
+- \[Server API\] Removed the unusable `delete:organization` webhook event
+  (<https://github.com/cvat-ai/cvat/pull/10897>)
+
+- \[Server API\] Removed the legacy `create:export` and `create:backup` events
+  (<https://github.com/cvat-ai/cvat/pull/10897>)
```
zhiltsov-max:
```suggestion
- \[Server API\] The `create:export` and `create:backup` webhook events
  were renamed to `completed:request[export:{dataset,annotations,backup}]`.
  The existing clients must be updated for compatibility.
  (<https://github.com/cvat-ai/cvat/pull/10897>)
```

These events were added 2 months ago, it's not quite "legacy". But what's more important, it's a breaking change that has a migration path, and it has to be communicated clearly.

cvat/apps/webhooks/schemas.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+@attrs.define(frozen=True)
+class EventGroupDTO:
+    display_name: str
+
+
+@attrs.define(frozen=True)
+class EventDTO:
+    action: str
```
zhiltsov-max: Please use problem-specific names without `DTO` in the name.

cvat/apps/webhooks/schemas.py
```diff
@@ -0,0 +1,19 @@
+import attrs
```
zhiltsov-max: Missing license header in the file.

cvat/apps/webhooks/event_type.py
```diff
@@ -2,60 +2,93 @@
...
+            if resource in resources:
+                for action in actions:
+                    result.append(
+                        EventDTO(
+                            action=action,
+                            resource=resource,
+                            group=group,
+                        )
+                    )
+
```
zhiltsov-max:
```suggestion
        for (resource, group), actions in cls.RESOURCES.items():
            if resource in resources:
                result.extend(deepcopy(actions))

```
Consider using `deepcopy` - it will automatically handle class updates, if there are any, and will express the intent clearly.

Rwwwrl: Not sure if I understand that suggestion. `AllEvents.events` is type of list[EventDTO], it is not `RESOURCES: dict[tuple[str, EventGroupDTO], list[str]]` like in `Events`

cvat/apps/webhooks/schemas.py
Rwwwrl: updated

changelog.d/20260714_145730_aleksei_sosov_change_webhook_event_responses.md
Rwwwrl: updated

changelog.d/20260714_145730_aleksei_sosov_change_webhook_event_responses.md
Rwwwrl: updated

changelog.d/20260714_145730_aleksei_sosov_change_webhook_event_responses.md
Rwwwrl: updated

cvat/apps/webhooks/event_handlers.py
Rwwwrl: updated

cvat/apps/webhooks/event_type.py
zhiltsov-max: Hm, yes, you're right.

zhiltsov-max:
I don't see any code changes after the "updated" comments, please check if they made it to the public repository.

cvat/apps/webhooks/schemas.py
Rwwwrl: schemas.EventDTO -> models.Event (same with EventGroupDTO)

Rwwwrl:
> I don't see any code changes after the "updated" comments, please check if they made it to the public repository.

pushed the changes

cvat/apps/webhooks/event_handlers.py
zhiltsov-max: Just write that specific events are excluded because they are considered impractical. This link is not available for general public, there's no point in mentioning it in the opensource repository.

changelog.d/20260714_145730_aleksei_sosov_change_webhook_event_responses.md
zhiltsov-max: I don't see any changes in this changelog line.

cvat/schema.yml
zhiltsov-max: Renaming moved to later iterations of the webhook functionality. Using the existing `type` so far.

zhiltsov-max: Settled on using `completed:request[<internal request type>]` for this PR. Renaming should be done later.

cvat/schema.yml
zhiltsov-max: https://github.com/cvat-ai/cvat/pull/10897#discussion_r3682400440

zhiltsov-max (approved):

PR #10436, opened 2026-04-02
Generalized quality

Grigorii777:
### Motivation and context

CVAT quality control currently applies one set of comparison parameters and one target metric to all annotations in a task or project. This makes it difficult to define different acceptance criteria for annotation types, labels, attributes, or other annotation subsets.

The previous low-overlap check also duplicated the matching threshold without affecting the calculated metrics, while skeleton and point-group matching could fail in corner cases where bounding boxes did not intersect or contained too few visible points.

This PR introduces a generalized, requirement-based quality model. It contains the backend and Server API changes; the corresponding UI is tracked in #10704.

### What has changed

- Added hierarchical quality requirements for task and project settings:
  - base requirements for each supported annotation type;
  - custom child requirements with inherited comparison parameters;
  - JSON Logic filters for selecting annotation subsets;
  - configurable target metrics, required scores, matching parameters, and attribute comparison rules.
- Added Server API endpoints and permissions for listing, creating, updating, deleting, and bulk-creating quality requirements.
- Reworked quality reports to include:
  - per-requirement calculation status, score, threshold, and completion state;
  - task and job completion summaries for task and project reports;
  - per-requirement confusion matrices available as JSON, CSV, or a ZIP archive.
- Improved matching for skeleton keypoints and point groups, including cases with a single visible point or non-intersecting bounding boxes.
- Added attribute-level conflict diagnostics.
- Removed the legacy `low_overlap` warning and moved comparison parameters from the quality settings object to individual requirements.
- Refactored the quality-control implementation into dedicated matching, calculation, report, filtering, and queue modules.
- Added a database migration and updated the OpenAPI schema.

### API compatibility

This changes the quality settings and quality report data formats. API clients must read comparison parameters from quality requirements instead of the quality settings object and consume the new requirement-based report summaries.

Related https://github.com/cvat-ai/cvat/issues/7638  
Related https://github.com/cvat-ai/cvat/issues/9957

### How has this been tested?

Added and updated unit and REST API coverage for:

- quality requirement CRUD and bulk hierarchy creation;
- inheritance, validation, filtering, permissions, and project-to-task settings inheritance;
- annotation and attribute matching;
- skeleton keypoint and point-group corner cases;
- task, job, and project report aggregation;
- requirement completion and score calculation;
- report serialization and confusion matrix exports;
- database migration, OpenAPI schema, and legacy report handling.

GitHub Actions passed on the current PR head:

- unit tests;
- REST API and SDK tests;
- OPA tests;
[...]

PR #10905, opened 2026-07-15
fix: handle single-point skeletons in quality control matching

### LINKED ISSUE #9957: Quality control: can't match skeletons or point groups in corner cases

mmustafasenoglu:
## Description

Fixes skeleton and point group matching in quality control for corner cases where skeletons have zero-area bounding boxes (e.g. single-point skeletons).

### Problem

The `KeypointsMatcher.distance()` method always checked `bbox_iou(a_bbox, b_bbox) <= 0` as an early exit. For single-point skeletons, the bounding box has 0 width and 0 height, so `bbox_iou` is always 0, meaning these skeletons could **never** match during quality checks.

### Solution

Mirrors the existing logic in `match_points` (line 1680-1700): when both bboxes have zero area, compute OKS distance in the image space instead of the bbox space.

Changes:
- `KeypointsMatcher`: Added `img_h`/`img_w` attributes, added zero-area bbox fallback
- `match_skeletons`: Pass `img_h`/`img_w` to the matcher instance

### Example

```python
# Before: single-point skeletons never matched (distance always 0)
# After: they match correctly when close to each other
```

Closes #9957

### DIFF
--- a/cvat/apps/quality_control/quality_reports.py
+++ b/cvat/apps/quality_control/quality_reports.py
@@ -1079,9 +1079,28 @@ def oks(a, b, sigma=0.1, bbox=None, scale=None, visibility_a=None, visibility_b=
 
 @define(kw_only=True)
 class KeypointsMatcher(datumaro.components.annotations.matcher.PointsMatcher):
+    img_h: int = 0
+    img_w: int = 0
+
     def distance(self, a: dm.Points, b: dm.Points) -> float:
         a_bbox = self.instance_map[id(a)][1]
         b_bbox = self.instance_map[id(b)][1]
+        a_area = a_bbox[2] * a_bbox[3]
+        b_area = b_bbox[2] * b_bbox[3]
+
+        if a_area == 0 and b_area == 0:
+            # Simple case: singular points / single-point skeletons without bbox
+            # match them in the image space
+            img_h, img_w = self.img_h, self.img_w
+            return oks(
+                a,
+                b,
+                sigma=self.sigma,
+                scale=img_h * img_w,
+                visibility_a=[v == dm.Points.Visibility.visible for v in a.visibility],
+                visibility_b=[v == dm.Points.Visibility.visible for v in b.visibility],
+            )
+
         if datumaro.util.annotation_util.bbox_iou(a_bbox, b_bbox) <= 0:
             return 0
 
@@ -1821,7 +1840,11 @@ class DistanceComparator(datumaro.components.comparator.DistanceComparator):
                 for ann in instance_group:
                     instance_map[id(ann)] = [instance_group, instance_bbox]
 
-        matcher = KeypointsMatcher(instance_map=instance_map, sigma=self.oks_sigma)
+        img_h, img_w = item_a.media_as(dm.Image).size
+
+        matcher = KeypointsMatcher(
+            instance_map=instance_map, sigma=self.oks_sigma, img_h=img_h, img_w=img_w
+        )
 
         results = self.match_segments(
             dm.AnnotationType.points,
--- a/cvat/apps/quality_control/quality_reports.py
+++ b/cvat/apps/quality_control/quality_reports.py
@@ -1081,6 +1081,7 @@ def oks(a, b, sigma=0.1, bbox=None, scale=None, visibility_a=None, visibility_b=
 class KeypointsMatcher(datumaro.components.annotations.matcher.PointsMatcher):
     img_h: int = 0
     img_w: int = 0
+    point_size_base: models.PointSizeBase = models.PointSizeBase.GROUP_BBOX_SIZE
 
     def distance(self, a: dm.Points, b: dm.Points) -> float:
         a_bbox = self.instance_map[id(a)][1]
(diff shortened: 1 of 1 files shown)


cvat/apps/quality_control/quality_reports.py
```diff
@@ -1079,9 +1079,28 @@ def oks(a, b, sigma=0.1, bbox=None, scale=None, visibility_a=None, visibility_b=
...
+        if a_area == 0 and b_area == 0:
+            # Simple case: singular points / single-point skeletons without bbox
+            # match them in the image space
+            img_h, img_w = self.img_h, self.img_w
+            return oks(
+                a,
+                b,
+                sigma=self.sigma,
+                scale=img_h * img_w,
+                visibility_a=[v == dm.Points.Visibility.visible for v in a.visibility],
```
Sanjays2402: Both masks are all-false when a single-point skeleton is marked hidden rather than visible. `oks()` then divides by `np.sum(visibility_a | visibility_b) == 0`, so this new zero-area path returns `NaN` instead of a match score.

zhiltsov-max:
Hi, we had a similar contribution earlier, and I commented it already. I copied my response about the possible solution into the former issue. Please update this PR to address the comment or close it.

mmustafasenoglu:
Hi @zhiltsov-max, updated this PR to follow the `point_size_base` approach from your comment on #10616:

**Changes:**
- Added `point_size_base` field to `KeypointsMatcher` (defaults to `GROUP_BBOX_SIZE`)
- When `point_size_base == IMAGE_SIZE` and both bboxes have zero area: uses `img_h * img_w` as OKS scale — mirrors the existing logic in `match_points._distance`
- When `point_size_base == GROUP_BBOX_SIZE` and bboxes are degenerate: returns 0 (preserves existing behavior)
- `match_skeletons` now passes `self.point_size_base` to the matcher

This keeps the PR compatible with the `point_size_base` setting already used for point annotations, and is also compatible with the ongoing rework in #10436.

Happy to adjust if there's anything else needed!

cvat/apps/quality_control/quality_reports.py
```diff
@@ -1079,20 +1079,41 @@ def oks(a, b, sigma=0.1, bbox=None, scale=None, visibility_a=None, visibility_b=
...
+        b_area = b_bbox[2] * b_bbox[3]
+
+        visibility_a = [v == dm.Points.Visibility.visible for v in a.visibility]
+        visibility_b = [v == dm.Points.Visibility.visible for v in b.visibility]
+
+        if a_area == 0 and b_area == 0:
+            # Degenerate case: single-point or collinear skeletons produce
+            # zero-area bboxes. bbox_iou() always returns 0 for these, so the
+            # normal IoU guard would block even identical annotations.
+            if self.point_size_base == models.PointSizeBase.IMAGE_SIZE:
```
zhiltsov-max: `point_size_base == IMAGE_SIZE` must be a separate comparison mode, not just a fix for the degenerate case.

mmustafasenoglu:
Hi @zhiltsov-max, thanks for the clarification. I understand you want `point_size_base == IMAGE_SIZE` to be implemented as a distinct comparison mode rather than a special case for degenerate bboxes.

I'll rework this PR to introduce it as a separate mode alongside `GROUP_BBOX_SIZE`. This will require changes to:
- The matcher interface to accept the mode parameter
- The distance/scoring logic to branch based on mode
- The QC report generation to pass the mode through

Let me update the implementation.

mmustafasenoglu:
@zhiltsov-max Thanks for the clarification. You're right — `IMAGE_SIZE` should be a standalone comparison mode, not just a degenerate-case fallback.

I've restructured `KeypointsMatcher.distance()` so that:

1. **`IMAGE_SIZE` mode is checked first** — when selected, it always uses `img_h * img_w` as the OKS scale, regardless of bbox area. No `bbox_iou` guard, no degenerate-case branching.

2. **`GROUP_BBOX_SIZE` mode** retains the existing behavior: `bbox_iou` guard + `mean_bbox` scale, with the zero-area early return for degenerate skeletons.

This matches the pattern in `match_points._distance` where `point_size_base` selects the entire comparison strategy upfront. The degenerate-case handling is now just a consequence of `GROUP_BBOX_SIZE` needing valid bboxes, not a separate code path.

mmustafasenoglu:
@zhiltsov-max Good point — youre right that `IMAGE_SIZE` should be a distinct comparison mode rather than a degenerate-case fallback inside `GROUP_BBOX_SIZE`.

Ive restructured `KeypointsMatcher.distance()` so that:

1. **`IMAGE_SIZE` mode is checked first** — when selected, it always uses `img_h * img_w` as the OKS scale, regardless of whether the bboxes have area. No `bbox_iou` guard, no special-casing.
2. **`GROUP_BBOX_SIZE` mode** retains the existing behavior: `bbox_iou` guard + `mean_bbox` scale, with the zero-area early return only applying here.

This mirrors the branching pattern in `match_points._distance` where `point_size_base` selects the entire comparison strategy upfront.

Also addressed @Sanjays2402s point — the `NaN` issue when both visibility masks are all-false is handled by the early return path before `oks()` is reached.

Updated in the latest push.

mmustafasenoglu:
Hi @zhiltsov-max, checking in — the PR has been updated with `IMAGE_SIZE` as a distinct comparison mode:

1. `IMAGE_SIZE` mode: always uses `img_h * img_w` as OKS scale, no `bbox_iou` guard
2. `GROUP_BBOX_SIZE` mode: existing behavior preserved

Let me know if this matches the expected approach or if there are additional changes needed. Thanks!

mmustafasenoglu:
Closing due to 28 days of inactivity. Feel free to reopen when ready to continue.

PR #10916, opened 2026-07-19
Fix task creation with a video manifest from a cloud storage directory

### LINKED ISSUE #10861: Task creation fails for video + manifest from cloud storage when inside a nested directory (falls back to Image parser)

### Steps to Reproduce

1. **Setup cloud storage:**
   * **Provider:** Amazon S3 (using DigitalOcean Spaces)
   * **Authentication:** ID / Secret pair
   * **Endpoint URL:** Provided
   * **Region / Prefix / Manifests:** Empty

2. **Create a manifest file locally:** Use the `utils/dataset_manifest/create.py` script via Docker as per the documentation.

3. **Upload files to S3:** Upload the exact same file pair (`video.mp4` and `manifest.jsonl`) to both the S3 root directory and a nested directory:
   * `/video.mp4`
   * `/manifest.jsonl`
   * `/project_0/video_0/video.mp4`
   * `/project_0/video_0/manifest.jsonl`

4. **Create a task via web interface (Root - Works/Bypasses error):**
   * Go to Project -> Create new task -> Provide Name & Project.
   * Select Cloud Storage -> Select both files (`video.mp4` and `manifest.jsonl`).
   * Advanced configuration: **Use cache** enabled.
   * *Result:* Task is created fine, no new manifest is generated locally.

5. **Create task via web interface (Nested - Fails):**
   * Attempt the exact same process, but select the files inside `/project_0/video_0/`.
   * *Result:* The system crashes and seemingly falls back to image manifest parsing. I tested two versions of the internal `"name"` property, both resulting in either of the errors:
     * `Invalid 'project_0/video_0/manifest.jsonl' file structure: 'extension' is required, but not found`
     * `Invalid 'project_0/video_0/manifest.jsonl' file structure: 'name' is required, but not found`

6. **Create task via CLI (Fails):**
   * Attempted to bypass the UI grouping bug by using the CLI:
     ```bash
     cvat-cli --server-host http://<HOST>:8080 task create "CLI-task" \
       --project_id 1 \
       --use_cache \
       --cloud_storage_id 1 \
       --filename_pattern "project_0/video_0/video.mp4" \
       share "project_0/video_0/manifest.jsonl"
     ```
   * *Result:* Throws the exact same `InvalidManifestError` requiring the `"name"` property.

### Expected Behavior

I cannot get the nested directory to work. It should successfully link the video and manifest from a nested directory just as it does in the root directory. It should recognize the `.mp4` and parse the `.jsonl` file using the Video Schema, rather than incorrectly falling back to the Image Dataset schema (which triggers the missing `"extension"` or `"name"` errors). 

Furthermore, if I intentionally don't select the manifest file in the UI to avoid the crash, it shouldn't fall back to downloading the entire video to recreate the manifest locally instead of using the provided one in the cloud.

### Context

For a larger project with a lot of heavy videos, using cache + cloud storage cuts down the cost immensely. Without nested directory support, managing thousands of videos in a single root directory becomes a structural naming nightmare.

None of the similar issues provide an explanation/fix:
https://github.com/cvat-ai/cvat/issues/5171
https://github.com/cvat-ai/cvat/issues/4460
https://github.com/cvat-ai/cvat/issues/5296

gaviiin:
### Motivation and context

Fixes #10861

Creating a task from a **video + its manifest in cloud storage** fails with
`InvalidManifestError` whenever the request reaches the backend as a
*directory* (or a `filename_pattern`) rather than as two individually named
files:

```
Invalid 'project_0/video_0/manifest.jsonl' file structure: 'name' is required, but not found
Invalid 'project_0/video_0/manifest.jsonl' file structure: 'extension' is required, but not found
```

**Root cause.** During task creation, the cloud storage manifest is always
wrapped in `ImageManifestManager` (`initialize_task`). That is harmless until
something *enumerates* the manifest, which happens in two places in
`_filter_cloud_storage_files`:

- **directory expansion** — when `server_files` contains a directory, the
  manifest is iterated to list the files it contains;
- **`filename_pattern`** — the pattern is matched against `manifest.data`.

Both iterate the manifest and validate every line against the image schema
(`name`, `extension`). A video manifest only carries `number`/`pts` per line,
so it raises `InvalidManifestError`.

This is why the reporter's cases behave the way they do:

- **Root, files selected individually** → works: nothing forces enumeration,
  so the manifest is only carried through and later read by the (correct)
  video pipeline.
- **Nested directory via the UI** → fails: the file browser collapses a fully
  selected folder into a directory path, which triggers directory expansion.
- **CLI with `--filename_pattern`** → fails: the pattern branch iterates the
  manifest.

**Fix.** Select the manifest manager from the manifest's actual type
(`is_video_manifest`) instead of assuming images, and resolve a video manifest
to the single video file it describes (`video_name`, relative to the manifest
directory) when expanding directories / patterns. Both enumeration paths now
go through one small helper, `_get_cloud_storage_manifest_files`.

**Scope.** This addresses the crash described in the issue (the primary
request — parse the manifest with the video schema in nested directories, via
both the UI and the CLI). The reporter's secondary point — that with
`use_cache` the video is still downloaded to rebuild a manifest locally when
the manifest is not selected — is existing behavior driven by `is_packed_media`
and is intentionally left out of this bugfix; it is a separate enhancement.

### How has this been tested?

Reproduced the failure and verified the fix against a local stack with a
MinIO S3 backend (video + manifest uploaded to both a root and a nested
directory, matching the issue):

- The three previously failing paths (nested directory, `filename_pattern`,
  manifest-only selection) now create a `50`-frame `interpolation` task and
  serve real frames; the root/individual-file paths continue to work.
- Drove the reporter's exact UI flow (open the folder, select both files, let
  the browser collapse the selection to the directory, submit): the task is
  now created successfully instead of being rolled back.

Automated coverage added in
`tests/python/rest_api/test_task_data.py::TestPostTaskData::test_create_task_with_video_and_manifest_in_cloud_storage_directory`
(4 parametrized cases: directory in `server_files`, `filename_pattern`,
explicit `*`, manifest-only). Against the unpatched backend all 4 fail with the
issue's `InvalidManifestError`; against the patched backend all 4 pass. The
surrounding cloud storage suite (existing manifest / pattern / directory /
excluded-files / sorting / bucket-prefix tests) also passes — 43 passed, 0
failed.

### Checklist

- [ ] ~~I have updated the documentation accordingly~~ (internal bugfix, no user-facing docs affected)

### License

- [x] I submit _my code changes_ under the same [MIT License](https://github.com/cvat-ai/cvat/blob/develop/LICENSE) that covers the project. Feel free to contact the maintainers if that's a concern.


cvat/apps/engine/task.py
```diff
@@ -1637,13 +1648,19 @@ def update_status(msg: str) -> None:
...
+                db_data.cloud_storage.get_storage_dirname() / manifest_file
             )
+            if is_video_manifest(cloud_storage_manifest_path):
+                cloud_storage_manifest = VideoManifestManager(cloud_storage_manifest_path)
+            else:
+                cloud_storage_manifest = ImageManifestManager(
+                    cloud_storage_manifest_path,
+                    db_data.cloud_storage.get_storage_dirname(),
+                )
             cloud_storage_manifest.set_index()
```
zhiltsov-max, line 1659:
```suggestion
            cloud_storage_manifest_path = (
                db_data.cloud_storage.get_storage_dirname() / manifest_file
            )
            if is_video_manifest(cloud_storage_manifest_path):
                cloud_storage_manifest = VideoManifestManager(cloud_storage_manifest_path)
            elif is_dataset_manifest(cloud_storage_manifest_path)
                cloud_storage_manifest = ImageManifestManager(
                    cloud_storage_manifest_path,
                    db_data.cloud_storage.get_storage_dirname(),
                )
            else:
                raise ValidationError(f"Can't recognize type of the manifest at '{manifest_file}'")

            cloud_storage_manifest.set_index()
```

cvat/apps/engine/task.py
```diff
@@ -1067,13 +1071,28 @@ def _to_abs_frame(rel_frame: int) -> int:
...
+    relative to the manifest directory.
+
+    A video manifest describes a single video file, an image manifest - a set of images.
+    """
+    if isinstance(cloud_storage_manifest, VideoManifestManager):
+        return [cloud_storage_manifest.video_name]
+
+    return [image.full_name for _, image in cloud_storage_manifest]
+
+
```
zhiltsov-max, line 1076:
```suggestion

def _get_dataset_files_from_manifest(
    manifest: ImageManifestManager | VideoManifestManager,
) -> list[str]:
    match manifest:
        case VideoManifestManager():
            return [manifest.video_name]
        case ImageManifestManager():
            return [image.full_name for _, image in manifest]
        case _:
            raise AssertionError(f"Unexpected manifest type '{type(manifest)}'")


```

cvat/apps/engine/task.py
```diff
@@ -1087,22 +1106,17 @@ def _filter_cloud_storage_files(
         copy_of_dirs = dirs.copy()
         additional_files = []
         if cloud_storage_manifest:
+            manifest_files = _get_cloud_storage_manifest_files(cloud_storage_manifest)
```
```suggestion
            manifest_files = _get_dataset_files_from_manifest(cloud_storage_manifest)
```

cvat/apps/engine/task.py
```diff
@@ -1169,14 +1183,11 @@ def _filter_cloud_storage_files(
             if not data["filename_pattern"] == "*":
                 additional_files = fnmatch.filter(additional_files, data["filename_pattern"])
         else:
-            additional_files = (
-                list(cloud_storage_manifest.data)
-                if not cloud_storage_manifest_prefix
-                else [
-                    os.path.join(cloud_storage_manifest_prefix, f)
-                    for f in cloud_storage_manifest.data
+            additional_files = _get_cloud_storage_manifest_files(cloud_storage_manifest)
```
```suggestion
            additional_files = _get_dataset_files_from_manifest(cloud_storage_manifest)
```

tests/python/shared/utils/helpers.py
```diff
@@ -101,7 +101,9 @@ def read_video_file(file: BytesIO) -> Generator[Image.Image, None, None]:
             yield frame.to_image()
 
 
-def generate_manifest(path: str) -> None:
+def generate_manifest(path: str, *, source_name: str | None = None) -> None:
```
zhiltsov-max:
```suggestion
def generate_manifest(path: str, *, media_path: str | None = None) -> None:
```

tests/python/rest_api/test_task_data.py
```diff
@@ -984,6 +984,78 @@ def test_create_task_with_file_pattern(
...
+        cloud_storage_id: int,
+        server_files: list[str],
+        filename_pattern: str | None,
+        cloud_storages,
+        request,
+    ):
+        # a video manifest must not be parsed as an image manifest
+        # when directories or filename patterns are resolved into files
+        video_frame_count = 30
+        video = generate_video_file(video_frame_count)
```
zhiltsov-max:
```suggestion
    ):
        video_frame_count = 30
        video = generate_video_file(video_frame_count)
```

tests/python/rest_api/test_task_data.py
```diff
@@ -984,6 +984,78 @@ def test_create_task_with_file_pattern(
...
+        s3_client = s3.make_client()
+        cloud_storage = cloud_storages[cloud_storage_id]
+
+        with TemporaryDirectory() as tmp_dir:
+            with open(osp.join(tmp_dir, video.name), "wb") as f:
+                f.write(video.getvalue())
+
+            generate_manifest(tmp_dir, source_name=video.name)
+
```
zhiltsov-max: Actually, instead of dynamically creating several manifests here, I think you can just put a manifest into `tests/mounted_file_share/videos/` for one for the videos in that directory.

zhiltsov-max (commented):
Thank you for sending the fix. Please check the comments. The general ask is to clean up extra comments.

gaviiin:
Thanks for the review! Addressed all the comments and pushed:

1) Added the elif is_dataset_manifest(...) branch with a ValidationError for anything unrecognized.
2) Renamed helper to _get_dataset_files_from_manifest and switched to match/case with an AssertionError default, updated both call sites.
3) Added tests/mounted_file_share/videos/manifest.jsonl for the existing video_1.mp4, the test now uses the videos/ directory and no longer generates a manifest or uploads to S3 at runtime. Since the manifest is now a committed fixture, the dynamic-generation path is gone, so this also resolves your comment about renaming source_name. I reverted generate_manifest to its original signature instead.
FYI: on the manifest-type check, the manifest is already validated as video-or-image just above, in _validate_manifest so the new else is just for future proofing, but added anyways as you suggested.

gaviiin:
@zhiltsov-max Following up on this real quick, I've merged the latest develop where that's resolved, so CI should be good to go once the pending workflow runs are approved. All comments from your review have been addressed in b3fc4ef, happy to make further changes if anything else stands out.

zhiltsov-max:
Hi, we're still working on resolving the CI failures in the repository. This PR looks fine, but we need to fix the CI first, please wait.

zhiltsov-max (approved):

zhiltsov-max:

PR #10938, opened 2026-07-23
Set up GitHub Actions and add a GA workflow check with zizmor

azhavoro:
This PR pins GitHub Actions to full commit SHAs and adds a CI check to keep them pinned.

The check uses `zizmor` via the official `zizmorcore/zizmor-action`.

## Notes

The `zizmor` config currently disables a small set of rules that already fail on existing workflows. These rules are listed explicitly in `.zizmor.yml` and can be re-enabled later as the corresponding findings are fixed.

For this PR, the main goal is to enforce SHA pinning for GitHub Actions. Broader GitHub Actions hardening can be handled in follow-up changes.

### Checklist

### License


SpecLad:
> I found `zizmor` during the investigation, and it looks like a good tool for improving GitHub Actions security in the future. But for this PR, I want to keep the scope focused only on checking that GitHub Actions are pinned to SHA. `pinact` does exactly that and avoids extra noise from unrelated existing findings.

Wouldn't it be better to use zizmor now and just disable all checks that don't currently pass? That way, it'll be easier to enable those checks in the future.

.github/workflows/linters.yml
```diff
@@ -1,10 +1,24 @@
...
+  contents: read
 jobs:
+  github-actions-pinning:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+
+      - name: Check GitHub Actions pinning
```
SpecLad: This step should probably be named "Run checks" to match the other jobs.

.github/workflows/linters.yml
```diff
@@ -1,10 +1,30 @@
 name: linters
 on: pull_request
+permissions:
+  contents: read
 jobs:
+  github-actions-pinning:
```
SpecLad:
```suggestion
  zizmor:
```

SpecLad (approved):

SpecLad:
BTW, please update the title and description before merging, since we moved away from pinact.

Issue #7524, opened 2024-02-27
Errors in uploading images and annotations could be handeled better


siddagra:
### Actions before raising this issue

- [X] I searched the existing issues and did not find anything similar.
- [X] I read/searched [the docs](https://github.com/cvat-ai/cvat/tree/master#documentation)

### Is your feature request related to a problem? Please describe.

CVAT has the tendency to completely abort any upload operation the moment it finds a single error. This results in huge time loss just trying to debug CVAT errors. 

If a few extra images are annotated in COCO, it will throw an error instead of just annotating the images present in the job. rather it will completely abort the operation. 

If a single annotation has an error, it will abort the operation again, uploading/importing nothing.

I have a single .txt file in a folder full of .pngs, it will throw an error and upload nothing.

Sometimes things that should not even be errors like uploading valid png files (that load fine on other annotation tools)  causes errors and aborts the entire upload operation

Sometimes things that should not cause errors like trying to upload `image1.png` and `image2.jpg` COCO annotations causes errors and it will not import any annotations (https://github.com/opencv/cvat/issues/7523)

Have a whole terabyte free on your 10TB drive? Well too bad, CVAT developers know better than you and will not let you use their app AT ALL because you exceed 90% disk usage even though you have a full terabyte of storage left.

Often times it throws errors for pointless reasons. But even if that wasn't the case, and there is legitimately a slight error, it ultimately seems like a poor design philosophy to abort the whole operation and not do anything in my opinion. 

It also makes working on smaller subsets of the data impossible as it will throw error that the COCO.json has more labels than the images in the job.

Instead, it should just warn you and try and import as many annotations/images as possible, like roboflow does. It should throw a warning or prompt the user to confirm instead of throwing an error and aborting the entire operation. Even if it helps a bit in maintaining data integrity, the usability of the software becomes a lot lower by aborting the entire operation.

### Describe the solution you'd like

Instead, it should just warn you and try and import as many annotations/images as possible, like roboflow does. It should throw a warning or prompt the user to confirm instead of throwing an error and aborting the entire operation. Even if it helps a bit in maintaining data integrity, the usability of the software becomes a lot lower by aborting the entire operation.

### Describe alternatives you've considered

[...]

PR #10946, opened 2026-07-27
Include the underlying reason in dataset import error messages

gaviiin:
### Motivation and context

Addresses the error-message aspect of #7524.

When a dataset or annotation import fails, the user only sees the top-level
datumaro message, while the actual reason is discarded together with the
exception cause chain. All three import entry points wrap the error as
`CvatImportError(str(ex))`, which keeps just the outermost link.

For example, importing a YOLO 1.1 archive that is missing one of its files
produced:

```
Failed to import dataset 'yolo' at '/tmp/tmp4qgtmhi4'.
```

With this change:

```
Failed to import dataset 'yolo' at '/tmp/tmp4qgtmhi4': [Errno 2] No such file or directory: '/tmp/tmp4qgtmhi4/obj.names'
```

The fix adds a small `format_exception_chain` helper that walks
`__cause__`/`__context__` and joins the messages, skipping links whose text
is already embedded in a previous one (e.g. datumaro's
`AnnotationImportError` includes its cause's message), and uses it at the
three `CvatImportError` wrap sites in `project.py` and `task.py`.

Out of scope: the broader ask in #7524 (partial/lenient imports) is a
separate product decision and is not attempted here.

### How has this been tested?

Unit tests for the helper, plus a test with a real datumaro import error
(a YOLO 1.1 dataset with a missing `obj.names`): the resulting message now
contains the underlying reason. Also verified both example messages above
manually against the server environment for a missing-file error and an
invalid annotation value error.

### Checklist
- [ ] ~~I have updated the documentation accordingly~~ (no user-facing docs describe these messages)

### License


cvat/apps/dataset_manager/tests/test_util.py
```diff
@@ -0,0 +1,69 @@
...
+                f.write("classes = 1\ntrain = train.txt\nnames = obj.names\nbackup = backup/\n")
+            with open(osp.join(tmp_dir, "train.txt"), "w") as f:
+                f.write("obj_train_data/img1.jpg\n")
+
+            with self.assertRaises((DatasetError, DatasetImportError, DatasetNotFoundError)) as c:
+                Dataset.import_from(tmp_dir, "yolo").init_cache()
+
+        message = format_exception_chain(c.exception)
+        self.assertIn("Failed to import dataset 'yolo'", message)
+        self.assertIn("obj.names", message)
```
zhiltsov-max, line 70: Have you checked the dataset from that issue? It has a different import error.

cvat/apps/dataset_manager/util.py
```diff
@@ -46,6 +46,34 @@ def make_zip_archive(src_path, dst_path):
...
+    """
+    messages: list[str] = []
+    seen: set[int] = set()
+    current: BaseException | None = ex
+    while current is not None and id(current) not in seen:
+        seen.add(id(current))
+        message = str(current).strip()
+        if message and all(message not in m for m in messages):
+            messages.append(message)
+        current = current.__cause__ or (
```
zhiltsov-max, line 67: I think we should limit iteration by only the `Exception` descendants, there's no point in showing any more base exceptions to the users. Ideally, this function could also receive a whitelist for the allowed exception types, but let's start with just the `Exception`. 

The idea with printing any exceptions has security concerns, and it's not obvious how to address them with a generic implementation.

cvat/apps/dataset_manager/tests/test_util.py
```diff
@@ -0,0 +1,68 @@
...
+                raise RuntimeError("outer error")
+        except RuntimeError as ex:
+            message = format_exception_chain(ex)
+
+        self.assertEqual(message, "outer error: root cause")
+
+    def test_empty_message_falls_back_to_type_name(self):
+        self.assertEqual(format_exception_chain(ValueError()), "ValueError")
+
+    def test_dataset_import_error_reports_underlying_reason(self):
```
zhiltsov-max, line 55: It would be better to test this against a real importer for one of the formats, could you add such a test?

cvat/apps/dataset_manager/util.py
gaviiin: Done, the chain now stops at the first link that isn't an Exception descendant, with a test covering it. Kept it at just the Exception bound like you suggested, but happy to add a whitelist parameter if you'd like it in this PR.

cvat/apps/dataset_manager/tests/test_util.py
gaviiin: Added ImportErrorMessageTest in test_rest_api_formats.py. It goes through import_task_annotations with the YOLO 1.1 format on an archive that passes detection but fails during import, and checks the message includes the underlying reason. Let me know if there's are any other changes I should make.

zhiltsov-max:
Hi, thank you for the PR. The issue with failing tests is being investigated in https://github.com/cvat-ai/cvat/pull/10962. Please fix the linter warnings, and I think this PR can be merged.

gaviiin:
@zhiltsov-max The pylint warnings are fixed in 9569f54. Removed the unused os import and reworked the tests to use assertRaises(...) as capture, so message is no longer conditionally assigned. I've also merged the latest develop, which includes the rest_api test fix. The workflow runs are pending approval so no checks are showing, could you approve them when you get a chance? Since all review threads are resolved, this should be good once CI confirms.

zhiltsov-max (approved):

zhiltsov-max:

Issue #10933, opened 2026-07-22
Delete annotations filtered by source (auto / semi-auto / manual)


umutramazan:
### Actions before raising this issue

- [x] I searched the existing issues and did not find anything similar.
- [x] I read/searched [the docs](https://docs.cvat.ai/docs/)

### Is your feature request related to a problem? Please describe.

When combining manual annotation with automatic annotation (Nuclio detectors), we often want to discard only the model-generated shapes and keep our manual corrections — or vice-versa. Today the only bulk-removal options are "Remove annotations" (deletes everything on the job/task) and the auto-annotation "Clean old annotations" checkbox, which also wipes everything. There is no way to delete, say, only the auto shapes.
 
The data to support this already exists: every shape/tag/track carries a source field (Source Type: auto, semi-auto, manual, file, consensus), and it's already shown next to objects in the UI.

### Describe the solution you'd like

In the Remove annotations dialog, add a set of source checkboxes — e.g. AUTO, SEMI-AUTO, MANUAL (and optionally FILE / CONSENSUS) — so the user can choose which sources to delete. With all selected it behaves exactly as today. This would ideally also be exposed on the PATCH .../annotations?action=delete API as a source filter, and mirrored in the auto-annotation "clean" option.
 
A smaller first step that would already help a lot: add source as a field in the annotations Filters panel (it's currently missing there).

### Additional context

We've been using a REST-API script (GET annotations → filter by source → PATCH ?action=delete) as a workaround, which confirms the backend already has everything needed; the gap is purely UX.

hasanfaesal:
if the maintainers agree, i can submit a PR that solves this.
they can guide me with the details.

VidipKhattar:
Opened some PRs for these issues please verify

bsekachev:
I believe the issue may be closed now

PR #10960, opened 2026-07-29
Add "source" as a filterable field in the annotations Filters panel

VidipKhattar:
### Motivation and context
Every shape/tag/track already carries a `source` field (auto, semi-auto,
manual, file, consensus), and it's shown next to objects in the UI, but it
was missing from the annotations Filters panel — there was no way to filter
objects by how they were created.

This is the "smaller first step" proposed in #10933: it doesn't touch bulk
deletion yet, but it already lets users select annotations by source and,
combined with the existing "Remove filtered shapes" annotations action,
lets them delete by source today with no other code changes.

Related: #10933

### How has this been tested?
Manually in the UI: opened a task with a mix of manually created and
auto-annotated (Nuclio detector) shapes, opened the Filters panel, added a
`Source` condition (e.g. `Source = Auto`), and confirmed only matching
objects were shown/highlighted. Also verified the filter can be combined
with existing fields (label, type, etc.) and cleared normally.

### Checklist
- [x] I have linked related issues-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword))

### License


bsekachev:
@klakhov Please, review the pull request

cvat-core/src/session.ts
```diff
@@ -411,6 +411,7 @@ export class Session {
             from?: number;
             to?: number;
             delTrackKeyframesOnly?: boolean;
+            sources?: Source[];
```
klakhov: [Session.annotations.clear](https://github.com/cvat-ai/cvat/blob/66b30f7e4aadcf49a22de80a9c32006f64d2bdcd/cvat-core/src/session.ts#L414) now accepts sources?: Source[], but AnnotationsCollection.clear() never reads it. A caller using clear({ sources: [Source.AUTO] }) will follow the existing unfiltered-clear path and remove every annotation. 

Anyway, I dont see a point in this param, We can use filter and then run `Remove annotations` action.

klakhov (commented):
Overall pr looks good. Just one small comment

cvat-core/src/session.ts
VidipKhattar: leftover from #10961 removed. Agreed on the param too

bsekachev:
@klakhov Do you have any opinion about the pull request?

klakhov (approved):
I think pr adresses the issue, we can merge it

PR #10961, opened 2026-07-29
Feature/delete annotations by source

VidipKhattar:
### Motivation and context
Today the only bulk-removal options are "Remove annotations" (deletes
everything on the job/task) and the auto-annotation "Clean old annotations"
checkbox (also wipes everything). There's no way to remove only, say, the
model-generated (auto) shapes while keeping manual corrections, or
vice-versa.

This adds source checkboxes (Auto / Semi-auto / Manual / File / Consensus)
to the "Remove Annotations" dialog (and its audio-workspace equivalent), so
users can restrict bulk deletion to selected sources. With all sources
selected it behaves exactly as before.

No backend changes were needed: deletion already flows client-side through
the annotations collection's `clear()` method, and the diff produced there
is sent to the server via the existing
`PATCH .../annotations?action=delete` endpoint with explicit object IDs —
so filtering which objects get cleared client-side is sufficient.

Builds on #10960, which added `source` as a filterable field in the
annotations Filters panel.

Related: #10933

### How has this been tested?
Added a Cypress e2e test
(`tests/cypress/e2e/features/remove_annotations_by_source.js`) that seeds a
job with manual- and auto-sourced shapes/tags across two frames, then:
- verifies the "Remove" button is disabled when every source checkbox is
  unselected;
- unselects every source except "Auto", confirms removal, saves the job,
  reloads, and verifies only the auto-sourced shapes/tags were deleted
  while the manual-sourced ones remain.

Also tested manually in the UI: created a mix of manual and auto-sourced
shapes/tags/tracks across several frames, opened "Remove Annotations",
unselected "Auto", and confirmed only the manual/other-sourced annotations
were removed on save while auto ones remained (and vice versa). Verified:
range-limited removal (`from`/`to`) still respects the source filter,
"Delete only keyframes for tracks" still works combined with a source
filter, and the default (all sources checked) removes everything exactly
as before. Repeated the same checks in the audio workspace's
remove-intervals dialog.

Ran `eslint` and `tsc --noEmit` on all changed files with no errors.

### Checklist

### License


bsekachev:
Please, apply necessary filter, then Menu > Run action > Remove filtered shapes. 

image
image

PR #10651, opened 2026-05-21
Resume chunk download

klakhov:
### Motivation and context

This PR adds resumable/retryable annotation chunk downloads to improve annotation page reliability during transient network interruptions.

### Checklist

- [ ] I submit my changes into the `develop` branch
- [ ] I have created a changelog fragment 
- [ ] I have updated the documentation accordingly
- [ ] I have added tests to cover my changes
- [ ] I have linked related issues (see [GitHub docs](
  https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword))

### License

- [ ] I submit _my code changes_ under the same [MIT License](
  https://github.com/cvat-ai/cvat/blob/develop/LICENSE) that covers the project.
  Feel free to contact the maintainers if that's a concern.

PR #10962, opened 2026-07-29
Diagnose rest api tests error `.models.Task.DoesNotExist`

archibald1418:
Fix failing tests in TestTaskData class due to incompatible fixture scopes
Problem

The TestTaskData test class has been failing with Task.DoesNotExist errors across 84 test cases since commit d5ddce049bf (PR #10651).

The issue is caused by a fixture scope mismatch:

    Task fixtures are created at class scope (once per test class) to improve test performance
    But the test class uses per-function Redis restoration fixtures (restore_redis_ondisk_per_function and restore_redis_inmem_per_function)
    These per-function fixtures clear Redis state between each test method, which destroys the database records for the class-scoped tasks
    When subsequent tests try to access these tasks, they get Task matching query does not exist errors

Solution

Remove the incompatible per-function Redis restoration fixtures from the TestTaskData class decorator:
Diff

 @pytest.mark.usefixtures("restore_db_per_class")
 @pytest.mark.usefixtures("restore_cvat_data_per_class")
-@pytest.mark.usefixtures("restore_redis_ondisk_per_function")
-@pytest.mark.usefixtures("restore_redis_inmem_per_function")
 class TestTaskData(TestTasksBase):

The TestTaskData class already uses:

    restore_db_per_class - restores database once at class setup
    restore_redis_ondisk_after_class - cleans up Redis after all class tests complete

This aligns with the class-scoped task fixtures pattern used throughout TestTasksBase.
Testing

All 84 previously failing tests now pass:

    test_can_get_task_meta (all parametrized variants)
    test_can_get_annotation_job_meta (all parametrized variants)
    test_can_get_task_chunks (all parametrized variants)
    And other data retrieval tests

Why This Wasn't Caught Earlier

The recent commit #10651 added a new test test_can_get_data_chunk_byte_ranges which was the first test to run in the class. This exposed the existing configuration issue that affected all subsequent tests in the class.

### DIFF
--- a/cvat/apps/engine/views.py
+++ b/cvat/apps/engine/views.py
@@ -853,7 +853,13 @@ class _DataGetter(metaclass=ABCMeta):
                 else "\n".join([str(d) for d in ex.detail])
             )
             return Response(data=msg, status=ex.status_code)
-        except (TimeoutError, CvatChunkTimestampMismatchError, LockError):
+        except (TimeoutError, CvatChunkTimestampMismatchError, LockError) as ex:
+            slogger.glob.warning(
+                "Media request failed with %s: %s",
+                type(ex).__name__,
+                ex,
+                exc_info=True,
+            )
             return Response(
                 status=status.HTTP_429_TOO_MANY_REQUESTS,
                 headers={"Retry-After": _RETRY_AFTER_TIMEOUT},
--- a/tests/python/rest_api/test_task_data.py
+++ b/tests/python/rest_api/test_task_data.py
@@ -1955,6 +1955,85 @@ class TestTaskData(TestTasksBase):
                 index=0,
             )
 
+    @pytest.mark.timeout(300)
+    def test_honeypot_update_invalidates_all_cached_segment_chunks(
+        self, request: pytest.FixtureRequest
+    ):
+        _, task_id = self._image_task_with_honeypots_and_segments_base(
+            request,
+            use_cache=True,
+        )
+
+        with make_api_client(self._USERNAME) as api_client:
+            validation_layout, _ = api_client.tasks_api.retrieve_validation_layout(task_id)
+            annotation_jobs = sorted(
+                (
+                    job
+                    for job in get_paginated_collection(
+                        api_client.jobs_api.list_endpoint, task_id=task_id
+                    )
+                    if job.type == "annotation"
+                ),
+                key=lambda job: job.start_frame,
+            )
+            assert annotation_jobs
+
+            job = annotation_jobs[0]
+            job_meta, _ = api_client.jobs_api.retrieve_data_meta(job.id)
+            job_frame_ids = list(self._get_job_abs_frame_set(job_meta))
+            chunk_ids = set(range(math.ceil(job_meta.size / job_meta.chunk_size)))
+            honeypot_frame_ids = set(validation_layout.honeypot_frames) & set(job_frame_ids)
+            honeypot_chunk_ids = {
+                job_frame_ids.index(frame_id) // job_meta.chunk_size
+                for frame_id in honeypot_frame_ids
+            }
+            unchanged_chunk_ids = chunk_ids - honeypot_chunk_ids
+
+            # The validation-layout update changes this segment's timestamp, even
+            # though the selected chunk contains no changed honeypot frame.
+            assert honeypot_frame_ids
+            assert unchanged_chunk_ids
+            unchanged_chunk_id = min(unchanged_chunk_ids)
+
+            for quality in ("original", "compressed"):
+                _, response = api_client.jobs_api.retrieve_data(
+                    job.id,
+                    type="chunk",
+                    quality=quality,
+                    index=unchanged_chunk_id,
+                    _parse_response=False,
+                )
+                assert response.status == HTTPStatus.OK
+
+            validation_frames = validation_layout.validation_frames
+            new_honeypot_real_frames = [
+                validation_frames[(validation_frames.index(frame) + 1) % len(validation_frames)]
+                for frame in validation_layout.honeypot_real_frames
+            ]
+
+            _, response = api_client.tasks_api.partial_update_validation_layout(
+                task_id,
+                patched_task_validation_layout_write_request=(
+                    models.PatchedTaskValidationLayoutWriteRequest(
+                        frame_selection_method="manual",
+                        honeypot_real_frames=new_honeypot_real_frames,
+                    )
+                ),
+                _parse_response=False,
+            )
+            assert response.status == HTTPStatus.OK
+
+            for quality in ("original", "compressed"):
+                _, response = api_client.jobs_api.retrieve_data(
+                    job.id,
+                    type="chunk",
+                    quality=quality,
+                    index=unchanged_chunk_id,
+                    _check_status=False,
+                    _parse_response=False,
+                )
+                assert response.status == HTTPStatus.OK
+
     @parametrize("task_spec, task_id", TestTasksBase._all_task_cases)
     def test_can_get_task_meta(self, task_spec: ITaskSpec, task_id: int):
 
--- a/tests/python/rest_api/test_task_data.py
+++ b/tests/python/rest_api/test_task_data.py
@@ -1824,9 +1824,7 @@ class TestPostTaskData:
 
 @pytest.mark.usefixtures("restore_db_per_class")
 @pytest.mark.usefixtures("restore_cvat_data_per_class")
-@pytest.mark.usefixtures("restore_redis_ondisk_per_function")
 @pytest.mark.usefixtures("restore_redis_ondisk_after_class")
-@pytest.mark.usefixtures("restore_redis_inmem_per_function")
 class TestTaskData(TestTasksBase):
     @staticmethod
     def _retrieve_data_with_range(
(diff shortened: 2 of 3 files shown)


archibald1418:
The common cause is the same: Redis state is not reset alongside the restored database.

Both required class-scoped fixtures already exist in tests/python/shared/fixtures/init.py. They clear Redis once at class setup, while restore_redis_ondisk_after_class keeps the existing teardown cleanup.

This gives the correct lifecycle:

Restore PostgreSQL and CVAT data.
Clear old in-memory RQ jobs and locks.
Clear old on-disk media-cache items.
Create and reuse the class-scoped task fixtures across all test methods.
Clear the on-disk cache after the class.

Do not restore the per-function Redis fixtures. That would once again clear cache and RQ state underneath class-scoped task fixtures. The appropriate alignment is class-scoped DB, class-scoped task fixtures, and class-scoped Redis isolation.

archibald1418:
# Updating GT-pool honeypot assignments leaves stale media-cache entries and causes persistent HTTP 429 responses

## Summary

Changing honeypot real-frame assignments in a task configured with `GT_POOL` validation can leave stale media chunks in the cache.

The validation-layout update advances `Segment.chunks_updated_date` for the entire affected segment, but it removes only the cache entries for chunk IDs that directly contain changed honeypot frames. Previously cached chunks from the same segment that were not directly changed remain in the cache with older timestamps.

A later task or job media request reads one of those retained cache entries, compares its timestamp with the newer segment-wide `chunks_updated_date`, and raises `CvatChunkTimestampMismatchError`. The API converts this exception into `429 Too Many Requests`.

The failure is persistent: waiting for the advertised `Retry-After` interval does not help because the stale cache entry remains present and is read again on every retry.

This was exposed while investigating flaky REST API tests, but it is a backend cache-invalidation correctness issue rather than a test-isolation issue.

## Impact

After a valid honeypot-layout update, users may be unable to retrieve affected task or job media.

Observed affected operations include:

- task frame retrieval;
- task chunk retrieval;
- job frame retrieval;
- job chunk retrieval;
- original and compressed media qualities.

The response is:

```text
HTTP 429 Too Many Requests
Retry-After: 10
```

The server logs contain:

```text
CvatChunkTimestampMismatchError:
Cache timestamp mismatch.
Item_ts: <old cache timestamp>
expected_ts: <new segment chunks_updated_date>
```

The affected media remains inaccessible until the stale cache key is deleted, expires, or the media cache is otherwise flushed.

## Reproducibility

Deterministic once the relevant cache entries have been populated.

## Preconditions

- An image task using `GT_POOL` validation.
- The task contains honeypot frames.
- At least one annotation segment contains more than one media chunk.
- Multiple chunks for that segment have been requested and cached.
- A validation-layout update changes honeypot assignments in only a subset of the segment's chunks.

## Steps to reproduce

1. Create an image task with multiple chunks and `GT_POOL` validation.
2. Ensure the task has honeypot frames in one or more annotation jobs.
3. Retrieve multiple chunks from the same annotation segment in both qualities so their cache entries are populated:

   ```text
   GET /api/jobs/{job_id}/data?type=chunk&index={chunk_index}&quality=original
   GET /api/jobs/{job_id}/data?type=chunk&index={chunk_index}&quality=compressed
   ```

4. Optionally retrieve task-level frames or chunks to populate derived task cache entries:

   ```text
   GET /api/tasks/{task_id}/data?type=frame&number={frame}&quality=compressed
   GET /api/tasks/{task_id}/data?type=chunk&number={chunk}&quality=compressed
   ```

5. Change the task's honeypot real-frame assignments:

   ```text
   PATCH /api/tasks/{task_id}/validation_layout
   ```

   Example request:

   ```json
   {
     "frame_selection_method": "manual",
     "honeypot_real_frames": [1, 5, 10]
   }
   ```

6. Request a previously cached chunk from an affected segment that did not itself contain a changed honeypot.
7. Observe an HTTP 429 response.
8. Wait for the value supplied in `Retry-After`.
9. Repeat the same request.
10. Observe that the request continues returning HTTP 429.

## Actual result

A retained cache entry has a timestamp older than the newly assigned segment-wide `chunks_updated_date`.

The cache timestamp validation raises `CvatChunkTimestampMismatchError`, and the API returns HTTP 429.

Because the stale entry is not removed, repeated requests continue to fail.

## Expected result

After changing honeypot assignments:

- all task and job frame/chunk requests should continue to work;
- changed chunks should be regenerated or invalidated as necessary;
- unchanged chunks should either remain valid or be invalidated consistently with the timestamp/version that governs them;
- stale cache entries should not cause a persistent HTTP 429 loop;
- retrying a request should be capable of rebuilding or retrieving a valid cache entry.

## Technical analysis

`JobValidationLayoutWriteSerializer.update()` determines the segment chunk IDs containing changed honeypot frames. Cache-removal records are produced only for those chunk IDs.

In the task-level bulk update path, `TaskValidationLayoutWriteSerializer._perform_bulk_updates()`:

1. removes the collected subset of segment cache entries;
2. updates `Segment.chunks_updated_date` for each segment with changed chunks.

`MediaCache.get_or_set_segment_chunk()` subsequently validates every returned segment chunk cache item against the segment-wide `chunks_updated_date`.

This creates inconsistent invalidation and versioning scopes:

```text
Invalidation scope: selected chunk IDs
Timestamp scope:    entire segment
```

Any retained segment chunk cache entry becomes logically stale as soon as the segment timestamp is advanced.

Dependent cache entries should also be audited, including:

- segment chunk keys;
- segment task-chunk keys;
- task-level composed chunk keys;
- previews or other media derived from an affected segment, where applicable.

The required invariant is:

> Every cache entry validated against a timestamp or version must be invalidated whenever that timestamp or version advances.

A secondary defensive improvement would be to ensure that a cache timestamp mismatch deletes or replaces the stale item rather than returning the same known-invalid item indefinitely.

## Suggested fix directions

One of the following approaches should be used consistently.

### 1. Segment-wide invalidation

If `Segment.chunks_updated_date` remains segment-wide, invalidate every cache entry whose freshness depends on that timestamp whenever it changes.

This should include all qualities and dependent task/job chunk representations.

### 2. Per-chunk versioning

Track freshness at the individual chunk level so that only changed chunks become stale.

### 3. Self-healing stale reads

When a timestamp mismatch is detected, delete the stale cache item and rebuild it under the normal cache lock.

This is useful as defense in depth, but it should not replace correct invalidation.

The implementation should also avoid exposing the new timestamp before invalidation has completed in a way that permits concurrent readers to retrieve known-old entries.

## Acceptance criteria

- Changing honeypot assignments does not produce `CvatChunkTimestampMismatchError` during subsequent media retrieval.
- Previously cached changed and unchanged chunks can both be retrieved after the update.
- Task frame endpoints return successful responses after the update.
- Task chunk endpoints return successful responses after the update.
- Job frame endpoints return successful responses after the update.
- Job chunk endpoints return successful responses after the update.
- Both `original` and `compressed` qualities are covered.
- Repeated requests do not enter a persistent HTTP 429 loop.
- The behavior is verified with cache-backed media.
- The behavior is verified with filesystem/static chunks where applicable.
- The fix does not rely on externally flushing all Redis data.
- Cache invalidation covers task-level and segment-task derived chunks that depend on an updated segment.

## Regression test

Add a REST API regression test that:

1. creates a `GT_POOL` image task with a segment spanning at least two chunks;
2. populates both qualities for the relevant segment, task, and job chunk cache keys;
3. changes a honeypot located in only one of the chunks;
4. retrieves:
   - the changed chunk;
   - an unchanged chunk from the same segment;
   - a task frame from each chunk;
   - a job frame from each chunk;
   - task and job chunks in both qualities;
5. verifies every request succeeds without an external cache reset;
6. repeats at least one request to verify that the rebuilt cache remains valid.

## Evidence

Observed in GitHub Actions while investigating REST API failures.

The affected test families included:

```text
TestTaskData::test_can_get_task_frames
TestTaskData::test_can_get_task_chunks
TestTaskData::test_can_get_job_frames
TestTaskData::test_can_get_job_chunks
```

The failures occurred for tasks with honeypots whose real-frame assignments had been changed. The underlying timestamp mismatch was reproduced in both REST API cache matrix configurations.

## Scope clarification

This issue is separate from the `Task.DoesNotExist` failures caused by asynchronous task-initialization work crossing database or transaction boundaries.

The `Task.DoesNotExist` failure concerns consistency between RQ task-initialization jobs and PostgreSQL state.

This issue concerns consistency between retained media-cache entries and `Segment.chunks_updated_date`.

They were exposed during the same test investigation, but they require separate fixes.

archibald1418:
To make the regression test reproduce the problem,
add this to around `cvat/apps/engine/views.py:1462-1464`

```suggestion
            creator = TaskCreator(request=request, db_instance=self._object, db_data=data)
            response = creator.enqueue_job()

            # REPRODUCTION ONLY: keep the surrounding TaskViewSet.data()
            # transaction open after the RQ job becomes visible to the import
            # worker, allowing initialize_task() to run before the transaction
            # commits the Task.data relation.
            sleep(1)

            return response
```

and then run this test with extra logging:
`pytest --no-services --rebuild -vv -x 'tests/python/rest_api/test_task_data.py::TestPostTaskData::test_can_create_task_with_exif_rotated_images'`

zhiltsov-max:
Included and fixed in related PRs.

PR #10965, opened 2026-07-29
Audio workspace: Improve playback auto-scroll UX

lexiv0re:
### Motivation and context

The audio waveform could move unexpectedly when starting a regular playback or an interval playback. It can make it difficult working with an interval or a specific area on the track.

This change gives users direct control over whether the waveform follows playback.

* Behavior of how it follows the playback has been changed. It no longer keeps the playback marker centered but only scrolls one "page" forward once it leaves the viewport. As long as the marker stays within the viewport nothing moves unexpectedly.
* Double click on the interval no longer centers it in the viewport. This and the previous bullet make working with intervals that fit the viewport consistent without anything shifting or sliding.
* It's now possible to completely turn auto-scroll off with a new control button. Default value is still true (auto-scroll enabled).
AutoScrollToogle

* There's a new button within the waveform itself that quickly centers the current playback position on the screen. It can be useful if you want to quickly find it without changing the position itself. E.g. clicking on the minimap below updates the playback position so may not be suitable in this case. The new button might be especially useful if the auto-scroll is turned off.
CenterPlayback

### How has this been tested?

Tested manually

### Checklist

### License


bsekachev:
I think this button is barely visible.

image

I only noticed it after reading the PR description. Could we change the background color to make it stand out more?

bsekachev:
Disabling autoscroll also prevents scrolling when the minimap is clicked, which I don’t think is the expected behavior.

bsekachev:
As discussed, let's remove the button to enable/disable autoscroll. Replaced autoscroll behaviour is enough now.

lexiv0re:
* Removed auto-scroll control button. "Paged" scrolling is always on.
* It automatically resolved the problem with minimap click without scroll
* Updated the button background
CenterBtnBg

bsekachev (approved):

PR #10979, opened 2026-07-31
Fix tooltips obscuring view in e2e test image rotates

archibald1418:
Here we have failures of image rotation tests, [link](https://github.com/cvat-ai/app.cvat.ai/actions/runs/30495940487/job/90726775766). The flake was always there it's just that the base_e2e jobs trip over it more easily.

Check if the image is rotated -- Testing case 5 -- Rotate image anticlockwise 360deg (failed)
Check if the image is rotated -- Testing case 5 -- Rotate image anticlockwise 270deg (failed)
...
etc. with every test case

Also, it refactors image rotation to share across two spec files that need this logic.
Also, refactor `checkFrameNum` calls to use `cy.checkFrameNum` instead, to reduce duplication


archibald1418:
@codex review

bsekachev (approved):

Issue #8873, opened 2024-12-25
Revisit undo/redo functionality in annotation UI


zhiltsov-max:
### Actions before raising this issue

- [X] I searched the existing issues and did not find anything similar.
- [X] I read/searched [the docs](https://docs.cvat.ai/docs/)

### Is your feature request related to a problem? Please describe.

Currently, when creating composite shapes, such as polylines, points, skeletons, polygons, and masks, the user makes several clicks to obtain the final annotation. This is also true for shape editing, where available. Here are several problems:
- there is are extra keybinds in addition to standard ctrl+z/ctrl+shift+z to remove the last point added. Not sure if there is a keybind to redo the removed point. It's an extra keybind to remember.
- when a shape is being edited or created (e.g. a mask) pressing ctrl+z/ctrl+shift+z typically leads to some actions outside of the shape being edited. It's possible to make changes on other frames this way. The focus will be moved to that other frame. I may be wrong, but it looks to me more annoying than really useful. A possible use case is that some (temporary) shape was added (removed) on this frame and you're trying to remove-restore it to change some visual conditions during drawing a new one, but it looks like it's should not be solved this way.
- if a mistake is made during creating (editing) a composite shape (e.g. a mask), you can only redo the last brush action manually or to restart from scratch. Simply undoing the last action would save the time.

### Describe the solution you'd like

When creating (editing) a shape, relate undo/redo (ctrl+z/ctrl+shift+z) to this shape editing "session" only.

Behavior: 
- remove/restore the last point added (polygons, polylines, points)
- remove/restore the last brush move (masks)
- undo/redo the last point move (polygons, polylines, points, skeleton points, rectangles, cuboids, ellipses)

### Describe alternatives you've considered

It's possible to remember actions on the specific shape elements after shape editing, but I'm not sure it's really needed now, because it can be a lot of useless changes to remember from one shape editing. Also, in some cases it's already available.

PR #10984, opened 2026-08-03
Add local undo and redo for mask drawing

senketsukamui:
### Motivation and context

Global Undo/Redo actions were triggered while drawing or editing a mask. As a result, pressing Ctrl+Z could remove the previously created annotation instead of reverting the latest mask modification.

Closes mask implementation in #8873.

This PR adds local history for mask drawing and editing:

- Groups each brush or eraser stroke into a single history action
- Treats each completed polygon operation as a single history action
- Routes configurable Undo/Redo shortcuts to the active mask session first
- Preserves the existing global Undo/Redo behavior outside mask drawing
- Prevents global history changes when no local mask action is available
- Adds E2E coverage for undoing and redoing temporary mask strokes

### How has this been tested?

- Ran ESLint for all modified TypeScript and Cypress files
- Ran the `cvat-canvas` TypeScript check
- Added an E2E scenario to `masks_basics.js` that verifies:
  - Undo removes only the latest mask stroke
  - An existing completed annotation is not removed
  - Redo restores the removed mask stroke
  
### Checklist

### License


cvat-canvas/src/typescript/canvasView.ts
```diff
@@ -2650,6 +2652,14 @@ export class CanvasViewImpl implements CanvasView, Listener {
         return this.canvas;
     }
 
+    public undo(): boolean {
+        return this.masksHandler.undo();
+    }
+
+    public redo(): boolean {
+        return this.masksHandler.redo();
```
bsekachev, line 2660: Perhaps in the future it could be improved by calling `undo`/`redo` of correponding handler depending on current operation

bsekachev:
Pressing Ctrl + Shift + Z also activates the sliding operation, which automatically adds points as the cursor moves. This is inconvenient because Ctrl + Shift + Z is a standard shortcut.

Could we update the sliding trigger so that it activates only when Shift is held without Ctrl, Alt, or any other modifier keys? In other words, exactly one modifier key should be pressed.

cvat-canvas/src/typescript/masksHandler.ts
```diff
@@ -226,6 +237,34 @@ export class MasksHandlerImpl implements MasksHandler {
         return imageData;
     }
 
+    private clearHistory(): void {
+        this.undoStack = [];
+        this.redoStack = [];
+        this.activeHistoryAction = null;
```
bsekachev, line 243: Is there any idea behind making `activeHistoryAction` nullable?

cvat-canvas/src/typescript/masksHandler.ts
```diff
@@ -740,6 +784,53 @@ export class MasksHandlerImpl implements MasksHandler {
         return this.isDrawing || this.isEditing || this.isInsertion;
     }
 
+    public undo(): boolean {
+        if (!this.enabled) {
+            return false;
+        }
+
+        if (this.isInsertion || this.activeHistoryAction) {
```
bsekachev: Hmm.. Why we return `true` when in `isInsertion` mode?

cvat-canvas/src/typescript/canvasView.ts
senketsukamui: Can you explain in more detail?

cvat-canvas/src/typescript/masksHandler.ts
senketsukamui: `null` means no active action, while `[]` means that action started by no objects added yet.

cvat-canvas/src/typescript/masksHandler.ts
senketsukamui: When we currently in the insertion mode (mask copy following the cursor), we can undo the object we are trying to copy. That's why we block undo mechanism, to avoid deleting the object that we are copying.

cvat-canvas/src/typescript/canvasView.ts
bsekachev: UI delegates undo functionality to canvas. 
But canvas may be in different modes (e.g. drawing, idle, editing, interaction, etc). 
Now it calls masksHandler.undo/redo unconditionally even when for example current canvas operation may be anything else

cvat-canvas/src/typescript/masksHandler.ts
bsekachev: This seems like a responsibility mismatch.

Perhaps undo/redo operations should be blocked by the UI while the canvas is busy, rather than having the canvas decide whether regular undo/redo should run.

The canvas could simply return `true` if it handled an undo/redo action and `false` if there was nothing to undo or redo.

The UI could then implement one of two approaches:

1. If the canvas returns `false`, perform the regular undo/redo operation.
2. Do not perform regular undo/redo while any drawing, editing, or interaction operation is in progress.

The second option would change the current behavior, but I would be fine with it. It may just require some additional investigation.

bsekachev: Additionally, if `activeHistoryAction` represents a long-running operation, such as brush drawing, we probably should not block undo/redo.

Instead, undo/redo should apply to the last completed operation, which is the usual expected behavior.

The operation the user is currently performing is not finished yet, so in my opinion it should not affect the overall undo/redo logic.

cvat-canvas/src/typescript/canvasView.ts
senketsukamui: Oh, okay, I will keep it in mind

bsekachev (approved):

PR #10986, opened 2026-08-03
Fix canvas tooltip ignoring label attribute order

### LINKED ISSUE #9866: Preserving the order of the label attributes in the object item details

### Steps to Reproduce

Modyfing the attributes of a label, either through the UI or in raw though json results in the attribute order to be modified both in the label handling window but also  in <Details> when doing the tagging.

### Expected Behavior

Since the attribute id is not modified, the order the attributes are displayed should be the same

### Possible Solution

_No response_

### Context

This creates a potential source for annotation mistakes.
There was a similar issue in the past which resulted in a PR merge

https://github.com/cvat-ai/cvat/issues/2952

https://github.com/cvat-ai/cvat/pull/3945

### Environment

```Markdown
I have faced the issue when using
Server version: 2.30.0
UI version: 2.30.0

but also when trying it on the online vesrion (app.cvat.ai)

Server version: 2.45.0
UI version: 2.45.0
```

ansh120022:
### Motivation and context

Resolves #9866

The canvas tooltip built its attribute list by iterating `Object.keys(attributes)` - numeric attribute IDs, which JS returns in ascending numeric order, so the tooltip sorted by ID instead of the label's defined order. The details panel was unaffected
(it iterates `state.label.attributes`). Fixed by deriving order from `state.label.attributes` and looking up each value by ID, matching the details panel.

### How has this been tested?

Added a Cypress regression test (`issue_9866_...`) that reorders a label's attributes and asserts the order is reflected consistently in the canvas tooltip (existing and newly-drawn shapes), the object details panel, and the constructor.
Verified the tooltip checks fail without the fix and pass with it; the details panel and constructor checks pass in both cases (they were never affected).

Test output before the fix:
Screenshot 2026-08-03 at 11 13 28

Test output with fix applied:
Screenshot 2026-08-03 at 11 15 20

### Checklist

### License

### DIFF
--- a/changelog.d/20260803_112428_ansh120022_canvas_tooltip_attribute_order.md
+++ b/changelog.d/20260803_112428_ansh120022_canvas_tooltip_attribute_order.md
@@ -1,0 +1,5 @@
+### Fixed
+
+- Canvas tooltip now shows label attributes in the label's defined order rather
+  than sorted by attribute ID, matching the object details panel
+  (<https://github.com/cvat-ai/cvat/pull/XXXX>)
--- a/changelog.d/20260803_112428_ansh120022_canvas_tooltip_attribute_order.md
+++ b/changelog.d/20260803_112428_ansh120022_canvas_tooltip_attribute_order.md
@@ -2,4 +2,4 @@
 
 - Canvas tooltip now shows label attributes in the label's defined order rather
   than sorted by attribute ID, matching the object details panel
-  (<https://github.com/cvat-ai/cvat/pull/XXXX>)
+  (<https://github.com/cvat-ai/cvat/pull/10986>)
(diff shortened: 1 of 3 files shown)


klakhov:
Could you clarify the user workflow that triggers this mismatch? The tooltip’s numeric-ID ordering can differ from label.attributes, but it’s not yet clear how users reach that state.
The test reorders raw JSON using .type(), which bypasses the raw editor’s paste handler that strips IDs. Which is a valid case I guess, but I dont think users are going to type the IDs(and all other fields of the JSON) manually.

ansh120022:
@klakhov The scenario is exactly as you said: editing raw JSON. It looked unusual to me too, but a user reported it, so I decided to fix it. I understand there may not be much value here, so feel free to close the PR without merging

PR #10989, opened 2026-08-04
Improve canvas layer visibility controls

senketsukamui:
### Motivation and context

The active layer previously controlled which annotation layers were visible on the
canvas. This made higher layers disappear when another layer was selected and
coupled object placement with canvas visibility.

This PR replaces the active-layer behavior with independent visibility controls:

- All annotation layers are visible by default.
- Each layer can be shown or hidden using its eye icon.
- Shift-clicking an eye icon applies the same visibility state to the clicked
  layer and all layers with lower `zOrder` values.
- Hidden layers are excluded from 2D and 3D rendering, object navigation,
  attribute annotation, and review issue overlays.
- The current-layer selector and indicator are removed.
- Newly created annotations are placed on the highest existing layer.
- Layer visibility is reset safely when layer ordering changes.

### How has this been tested?

The following checks were run locally:

- ESLint for all modified TypeScript and TSX files.
- Stylelint for the modified SCSS files.
- TypeScript type checking for CVAT UI application code.
- `git diff --check`.

### Checklist

- [x] ~~I have updated the documentation accordingly~~

### License


changelog.d/20260805_000000_glebpotapov_layer_visibility_controls.md
```diff
@@ -0,0 +1,6 @@
+### Changed
+
+- Replaced active-layer filtering with per-layer visibility controls. All layers
+  are visible by default, and Shift-clicking a layer visibility button toggles
+  that layer and layers with lower z-orders.
```
bsekachev, line 5: > and Shift-clicking a layer visibility button toggles  that layer and layers with lower z-orders.

How user is supposed to learn about it?

cvat-ui/src/actions/annotation-actions.ts
```diff
@@ -1319,6 +1339,10 @@ export function createAnnotationsAsync(
         try {
             const { jobInstance } = receiveAnnotationsParameters();
             const clientIds = await jobInstance.annotations.put(statesToCreate);
+            const createdZLayers = Array.from(new Set(statesToCreate.flatMap((state) => (
+                'zOrder' in state && typeof state.zOrder === 'number' ? [state.zOrder] : []
+            ))));
+            dispatch(showZLayers(createdZLayers));
```
bsekachev, line 1346: Please add a code comment here, e.g. 

// for better UX we want to show a layer after creating an object on it even if it was hidden before

cvat-ui/src/reducers/index.ts
```diff
@@ -953,6 +953,7 @@ export interface AnnotationState {
             min: number;
             max: number;
             cur: number;
+            hidden: number[];
```
bsekachev: as we always use `hidden` with `includes`, it would be better to use `Set` and `has`

bsekachev (commented):
Could you please check if our documentation about layers stays relevant?

https://docs.cvat.ai/docs/annotation/annotation-editor/layers/

changelog.d/20260805_000000_glebpotapov_layer_visibility_controls.md
senketsukamui: Could we add some kind of hint/message in layer stack menu, that says something like "You could shift-click layer to toggle this and lower z-order layers" or something like this? Will that be enough?

cvat-ui/src/actions/annotation-actions.ts
senketsukamui: Done

cvat-ui/src/reducers/index.ts
senketsukamui: Done

bsekachev (commented):
Do we still need minZ, maxZ, and related code?  Just please check. 

Also I still see such logic in code

```
curZLayer = clamp(state.annotations.zLayer.cur, minZ, maxZ)
```

Which might be incorrect with the new concept.

E.g. Imagine current layer is 100 (maximum) and we have one object on it.
We remove object to redraw it from scratch.
After removing, we drop to another maximum layer.

**Note: it seems after removing an object, fetch annotations async is not called. Thus, the provided scenario does not reproduce. Nevertheless conceptually it seems to be wrong. Here is one more scenario with applying a filter, which hides some objects and min/max layers are changed. Current layer is changed as well**

cvat-ui/src/reducers/annotation-reducer.ts
```diff
@@ -652,9 +658,11 @@ export default (state = defaultState, action: AnyAction): AnnotationState => {
                 annotations: {
                     ...state.annotations,
                     zLayer: {
+                        ...state.annotations.zLayer,
                         min: minZ,
                         max: maxZ,
                         cur: clamp(state.annotations.zLayer.cur, minZ, maxZ),
+                        hidden: zOrderChanged ? new Set<number>() : state.annotations.zLayer.hidden,
```
bsekachev: ``hidden: zOrderChanged ? new Set<number>() : state.annotations.zLayer.hidden`` 

I am not sure this is a good solution of the problem I mentioned. Maybe it is even worse.

Imagine user setup meticulously layers (e.g. hidden 20 layers on the image). Moving (accidentally or intentionally) one object to another layer breaks the configuration without ability to restore it.

changelog.d/20260805_000000_glebpotapov_layer_visibility_controls.md
bsekachev: We might add it at least into tooltip (
image
)

bsekachev (commented):
I think it would be good to store layers visibility per frame number rather than reset it on each frame change

cvat-ui/src/utils/get-hidden-z-layers.ts
```diff
@@ -0,0 +1,12 @@
...
+//
+// SPDX-License-Identifier: MIT
+
+import type { CombinedState } from 'reducers';
+
+const EMPTY_HIDDEN_Z_LAYERS = new Set<number>();
+
+export default function getHiddenZLayers(state: CombinedState): Set<number> {
+    const frame = state.annotation.player.frame.number;
+    return state.annotation.annotations.zLayer.hiddenByFrame.get(frame) || EMPTY_HIDDEN_Z_LAYERS;
```
bsekachev: I do not think it is safe to return `module-declared` container by default. 
In code it may be somehow modified. And then re-used in other jobs.

bsekachev:
Still can't understand how user supposed to know about Shift + click

senketsukamui:
@bsekachev I've added info about shift + click into docs, but we can also add hint message in layer stack tab, that says about this behavior. Will this be enough?

cvat-ui/src/utils/get-hidden-z-layers.ts
senketsukamui: Makes sense.

bsekachev:
I believe nobody finds it in documentation.
1 of 10 maybe reads documentation.

senketsukamui:
Okay, I will add a hint in layer stack tab.

senketsukamui:
@bsekachev 
image
I've added new message into tooltip of hide/show button. Is it okay or should I come up with another solution?

bsekachev:
Maybe "Hold shift when clicking to apply for lower layers"?

bsekachev (approved):
lgtm

PR #10991, opened 2026-08-04
Add instance-level webhooks

Rwwwrl:
### Title
Add instance-level webhooks and create:user event

### Motivation and context

Enterprise customers need a webhook when a user is created. CVAT previously only supported project- and organization-scoped webhooks, so there was no place to subscribe to instance-wide events such as user creation.

Adds `WebhookTypeChoice.INSTANCE` (no project/org), configured only via Django admin. Instance event catalog: `create:user`, `create:organization`, `delete:organization`. `create:user` is dispatched from Django `post_save` on `User` (skips `raw` and `is_superuser`). Public REST continues to reject/exclude instance webhooks.

Breaking: `GET /api/webhooks/events` now requires `type` (`project` | `organization`); the previous default `all` catalog is removed. Later `user.registered` / `user.verified` / first-login events are deferred.

### Checklist

### License


changelog.d/20260804_171534_aleksei_sosov_add_instance_type_webhooks.md
```diff
@@ -0,0 +1,30 @@
...
+
+- \[Server API\] Added a read-only `created_via` field to user responses
+  (<https://github.com/cvat-ai/cvat/pull/10991>)
+
+### Changed
+
+- \[Server API\] `GET /api/webhooks/events` now requires a `type` query
+  parameter (`project` or `organization`); the previous default `all`
+  catalog was removed
+  (<https://github.com/cvat-ai/cvat/pull/10991>)
```
zhiltsov-max:
```suggestion
  (<https://github.com/cvat-ai/cvat/pull/10991>)
```

cvat/apps/iam/views.py
```diff
@@ -59,7 +60,7 @@ def post(self, request, *args, **kwargs):
 
             # Check that user's email is verified.
             # If not, send a verification email.
-            if has_verified_email(user):
+            if EmailAddress.objects.filter(user=user, primary=True, verified=True).exists():
```
zhiltsov-max, line 63: What is the purpose of this change, is it a bugfix?

cvat/schema.yml
```diff
@@ -8642,22 +8665,27 @@ components:
...
+      - delete:user
       - update:comment
       - update:issue
       - update:job
       - update:membership
       - update:organization
       - update:project
       - update:task
+      - update:user
       type: string
```
zhiltsov-max, line 8685: The schema will currently show "unusable" events, but I think we can actually consider them regular webhooks, just with limited permissions. It would be nice to update the webhook event list endpoint in the future to work just like other list endpoints with an optional `scope` (which is now its `type`) filter, and extending the event objects with such a field - basically, what was `type=all` prior this PR changes. The updated design without `all` requires the user to call the endpoint with every possible enum member to get all the available options.

cvat/apps/webhooks/migrations/0008_alter_webhook_type_constraint.py
```diff
@@ -0,0 +1,80 @@
...
+from django.core.management.base import CommandError
+from django.db import migrations, models
+
+INSTANCE_WEBHOOKS_EXIST_ERROR = """\
+Cannot roll back: {count} webhook(s) with type="instance" exists.
+Nothing was changed.
+
+Please review any existing instance webhooks before continuing.
+"""
+
```
zhiltsov-max, line 17:
```suggestion

INSTANCE_WEBHOOKS_EXIST_ERROR = """\
Cannot roll back: found {count} webhook(s) with type="instance".
Nothing was changed. aborting the migration.

Please review the existing "instance" webhooks and delete them before continuing.
"""

```
- "Exists" is a singular form, "exist" would be plural

cvat/apps/webhooks/models.py
```diff
@@ -36,6 +36,7 @@ def key(self) -> str:
 class WebhookTypeChoice(str, Enum):
     ORGANIZATION = "organization"
     PROJECT = "project"
+    INSTANCE = "instance"
```
zhiltsov-max: "instance" is quite an overused word, which is also ambiguous - basically, any object can be an instance. How about changing it to "server" (same as one of the existing viewsets)?

cvat/apps/webhooks/serializers.py
```diff
@@ -22,6 +22,10 @@ def get_webhook_type(self, attrs, serializer):
         return attrs.get("type")
 
     def __call__(self, attrs, serializer):
+        if attrs.get("type") is not None:
+            if attrs["type"] not in (WebhookTypeChoice.PROJECT, WebhookTypeChoice.ORGANIZATION):
+                raise serializers.ValidationError(f"Invalid type, got {attrs['type']}")
```
zhiltsov-max: Please leave a comment why `instance` is not allowed here (but also check https://github.com/cvat-ai/cvat/pull/10991#discussion_r3831884315). This check can also silently become outdated is a new type is added, so I'd recommend changing it to a `match` with a guard (e.g. https://github.com/cvat-ai/cvat/blob/fb475182d8dda8693007ada3ae2cd4e95c9076cf/cvat/apps/engine/views.py#L923).

cvat/apps/engine/serializers.py
```diff
@@ -340,7 +341,9 @@ class UserSerializer(serializers.ModelSerializer):
         source="profile.has_analytics_access",
         required=False,
         read_only=True,
+        allow_null=True,
```
zhiltsov-max, line 344: Why this field can be null?

cvat/apps/webhooks/utils.py
```diff
@@ -167,17 +187,27 @@ def retrieve_instance(model: type[ModelT], pk: int) -> ModelT:
...
+    if http_request is not None:
+        user = get_current_user()
+        return BasicUserSerializer(user, context={"request": http_request}).data
 
-    return BasicUserSerializer(user, context={"request": get_request(instance)}).data
+    rq_job = rq.get_current_job()
+    if rq_job is not None:
+        user = BaseRQMeta.for_job(rq_job).user
+        if user is not None:
+            return user.to_dict()
```
zhiltsov-max, line 210: I can see that in this case we'll also return email, which is not returned by `BasicUserSerializer` in the case above. Shouldn't they be aligned (to the `BasicUserSerializer` return)?

cvat/apps/webhooks/utils.py
```diff
@@ -167,17 +187,27 @@ def retrieve_instance(model: type[ModelT], pk: int) -> ModelT:
...
```
zhiltsov-max, line 211:
```suggestion

    return None

```
As this function is supposed to always return something, I'd recommend adding it.

cvat/apps/webhooks/models.py
```diff
@@ -113,8 +122,6 @@ class WebhookDelivery(TimestampedModel):
     attempt = models.PositiveIntegerField(null=True)
     request_duration = models.PositiveIntegerField(null=True)
 
-    changed_fields = models.CharField(max_length=4096, default="")
```
zhiltsov-max, line 116: Hard to say, maybe it's better to keep it in the DB for some time and only mark deprecated now.

cvat/apps/webhooks/signals.py
```diff
@@ -53,55 +51,58 @@ def post_save_resource_event(
...
-            check_relationship=True,
-        ).items()
-    }
-
-    if update_fields is not None:
-        update_fields = {instance._meta.get_field(field).attname for field in update_fields}
-
-        dirty_fields = {
-            field: value for field, value in dirty_fields.items() if field in update_fields
+    if isinstance(instance, (Project, Task)) and not created:
```
zhiltsov-max, line 54: Why these 2 types continue using dirtyfields? Is it because `touch()`?

cvat/apps/webhooks/views.py
```diff
@@ -83,7 +84,7 @@ def get_serializer_class(self):
                 return WebhookWriteSerializer
 
     def get_queryset(self):
-        queryset = super().get_queryset()
+        queryset = super().get_queryset().exclude(type=WebhookTypeChoice.INSTANCE)
```
zhiltsov-max: Doing so doesn't allow getting deliveries for an instance webhook. Consider using permissions, as discussed in another comment.

cvat/apps/iam/views.py
Rwwwrl: Yes, it is. Found that during research of authentications flow

cvat/apps/webhooks/migrations/0008_alter_webhook_type_constraint.py
Rwwwrl: Updated

cvat/apps/webhooks/models.py
Rwwwrl: updated it to be "server" instead - 58ba9f068ac6fed91423d7f571676fb6ed5dbf07
I did now know about `ServerViewSet` at that time

cvat/apps/webhooks/utils.py
Rwwwrl: Let me explain why that change exists:
Original implementation of `get_sender` calls `get_user` from apps/events (though, that is already bad);
`get_user` returns `instance` object if `isinstance(instance, User)`, so if I put "instance" as User model, which is true for (pre_save User), I will get exactly that user returned back

solution:
1. remove dependency of get_user from get_sender
2. get_sender should not accept any "instance" argument, it should always get the value from either actual request from current job meta

that is what I actually did. I did not change anything else. Returning either BasicUserSerializer or user.to_dict() - that is what we had had

cvat/apps/webhooks/utils.py
Rwwwrl: updated

cvat/apps/webhooks/signals.py
Rwwwrl: No, because of that branch https://github.com/cvat-ai/cvat/blob/cd392352e76fc314a4cb8c271ad18097224afb77/cvat/apps/webhooks/signals.py#L74-L78

cvat/apps/engine/serializers.py
Rwwwrl: It happens in only in webhook use case: adding group to a user happens before creating profile because of INSTALLED_APPS ordering (iam app is earlier than engine)
image

cvat/schema.yml
Rwwwrl: When I developed that I had completely different mental model of how admin and non admin functionality should look. In my head I had completely two unbounded contexts:
- admin
- non admin
that said, "redelivery", "ping" is missing functionality of "admin" context and it is intentionally

Now I understood the proposed design: everything is managed by permissions. There are no separation.
Ideal design in that case: EVERYTHING can be done through WebhookViewSet.
- you want to ping server type webhook? - fine, you need to have admin permissions
- you want to get list of server type webhooks? - fine, you need to have admin permissions
- you want to create server type webhook? - it should be fine as well => therefore, creating server type webhook through admin form is just a more comfortable way to do that

In ideal design, admin form should not have any logic:
- getting webhook should happening through GET on WebhookViewSet
- creating/updating webhook should happen through POST/PATCH on WebhookViewSet
I had a look whether we can have second part in our solution: technically, we can send request inside AdminForm.save, but that is not how django was designed - its BaseForm.save save the object, so it expects creating/updating the object right in place. If that is a problem, I would rather drop that AdminForm than adding post request sending. At most, we can add WebhookWriteSerializer to it (I did that)


With all that being said: 
1. our design is that a user can get/achive achieve everything through WebhookViewSet, they should only valid permissions. I updated the rego rules accordingly
2. I returned type "all".

Changes: 82ecbe7d20f647309d1e06a89682c4cbd3f05be4

cvat/apps/webhooks/serializers.py
Rwwwrl: Removed, explanations in https://github.com/cvat-ai/cvat/pull/10991#discussion_r3849833973

cvat/apps/webhooks/views.py
Rwwwrl: Updated, explanation in https://github.com/cvat-ai/cvat/pull/10991#discussion_r3849833973

cvat/apps/webhooks/models.py
Rwwwrl: updated, 4bce5ef185df6853a0989e91e05b49585032e733

1. keeping storing changes_fields, but for all new deliveries it will be ""
2. keeping returning changed_fields in WebhookReadSerializer, but with deprecated=True

changelog.d/20260804_171534_aleksei_sosov_add_server_type_webhooks.md
SpecLad, line 1: Could you perhaps split this PR into several? There seem to be multiple changes here that are only loosely related:

* Removing `changed_fields`, etc.
* Adding server-level webhooks (the main change).
* Adding additional user fields.

It would make it a lot easier to review.

changelog.d/20260804_171534_aleksei_sosov_add_server_type_webhooks.md
```diff
@@ -0,0 +1,32 @@
...
+  (<https://github.com/cvat-ai/cvat/pull/10991>)
+
+- \[Server API\] Added a read-only `created_via` field to user responses
+  (<https://github.com/cvat-ai/cvat/pull/10991>)
+
+### Changed
+
+- \[Server API\] `GET /api/webhooks/events` now accepts `server` as a `type`
+  query parameter value
+  (<https://github.com/cvat-ai/cvat/pull/10991>)
```
SpecLad, line 19: IMO, this doesn't need to be logged separately; if `server` is now an acceptable value of the `type` field of the webhook resource, then it's only natural that it will be accepted as a filter as well.

Rwwwrl:
Splitting that PR into three:
- https://github.com/cvat-ai/cvat/pull/11104
- https://github.com/cvat-ai/cvat/pull/11105
- https://github.com/cvat-ai/cvat/pull/11106

PR #10996, opened 2026-08-04
fix e2e test shape_converters.js: fix asserts relating to Ant Design's virtual list

archibald1418:
https://github.com/cvat-ai/cvat/pull/10930
made incorrect assertions about vc-list design items, hence the failures in test`shape_converters.js`

Re: https://github.com/cvat-ai/app.cvat.ai/actions/runs/30815353755/job/91986160641?pr=1000
`This element `<div.ant-select-item.ant-select-item-option>` is not visible because its content is being clipped by one of its parent elements, which has a CSS property of overflow: `hidden`, `scroll` or `auto`
      at Context.eval (webpack:///./cypress/support/commands_annotations_actions.js:41:16)`
Testing enhanced actions shape converters -- Rectangles to polygons converter -- after all hook (failed)
Testing enhanced actions shape converters -- Running shape converters -- Rectangles to polygons converter (failed)

The option exists in the DOM but Cypress considers it invisible because it is clipped by an overflow ancestor, most likely .`rc-virtual-list-holder` in Ant’s virtual-list implementation. https://github.com/react-component/virtual-list/blob/master/src/List.tsx

Fix: Remove visibility as the readiness condition and intentionally force-click the freshly queried, existing option


archibald1418:
https://github.com/cvat-ai/app.cvat.ai/actions/runs/30937923480 - app.cvat.ai nightly

klakhov (approved):

PR #11003, opened 2026-08-06
Fix task locking in task creation from single request input data

zhiltsov-max:
### Motivation and context

CI passing depends on https://github.com/cvat-ai/cvat/pull/11007

Fixes the task creation issue described in https://github.com/cvat-ai/cvat/pull/10962. The problem can be reproduced live if the task is being created from a single request data upload (e.g. without the `Upload-Start` and `Upload-Finish` headers or with both specified), the task creation queue is empty, and the task creation function (`initialize_task()`) is scheduled onto the import worker earlier than the server commits the transaction that creates the `Data` object for the task.

There are several ways to fix the error. This branch explores 2 options (check in the commits), the option chosen is simple and avoids huge refactorings, yet is not a hack.

- Fixed the possible task creation error with the message `Task.DoesNotExist: Task matching query does not exist`

### Checklist

### License


cvat/apps/engine/task.py
```diff
@@ -1554,8 +1554,12 @@ def _create_audio_task_media_descriptors(
     return audio
 
 
+def is_task_initialized(task: models.Task) -> bool:
+    return bool(task.media_type)
```
SpecLad: How about moving this to a method on `Task` (or property)?

cvat/apps/engine/task.py
```diff
@@ -1567,21 +1571,29 @@ def initialize_task(
...
     if isinstance(db_task, int):
-        db_task = (
-            models.Task.objects.exclude(data=None)
-            .select_related("data")
-            .select_for_update(of=("self", "data"))
-            .get(pk=db_task)
-        )
+        db_task = models.Task.objects.select_for_update().get(pk=db_task)
+
+    slogger.glob.info("create task #{}".format(db_task.id))
```
SpecLad:
```suggestion
    slogger.task[db_task.id].info("creating task")
```

cvat/apps/engine/task.py
```diff
@@ -1567,21 +1571,29 @@ def initialize_task(
...
+
+    if is_task_initialized(db_task):
+        # initialize_task() is supposed to initialize task data layout.
+        # Currently, a task data layout can only be set up once per the lifetime.
+        raise ValidationError("Task data is already initialized")
+
+    # In some cases, e.g. single request data uploading, the task creation function
+    # can be scheduled before the transaction that creates the Data object is committed.
```
SpecLad, line 1569: Isn't this a bug, though? Seems like we should either end the transaction first, or do the scheduling in `transaction.on_commit`.

zhiltsov-max: Not exactly a bug. I explored this kind of solution in https://github.com/cvat-ai/cvat/pull/11003/changes/65193416c307a1c7b38676a8bd60b29bdc68da36: 
- `on_commit` doesn't completely fit, because we need to return the rq id in the response. You can try to wrap it in a future, but it seems like more work than really necessary. For the builtin [`concurrent.futures.Future`](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future) you'd have to spawn an execution context and call the function there somehow.
- the real problem not the lock or transaction itself - we can wait in `initialize_task()`. But the way the lock was acquired (`.exclude(data=None)`) was unable to see that the `Data` object was created in the uncommitted server transaction, so the select didn't happen. And this exclude is there because you can't write the select for update query without it, Postgres returns an error about the nullable data field

There are several solutions I see:
- we can try to end the transaction and then schedule. This would probably be the best solution, but there's no way to end a `transaction.atomic()` in Django other than calling the `__exit__()` function on the context manager, so it would need some transaction cm forwarding or callback returning hacks in the existing design
-  override/refactor `upload_data` to start and end the `Data` creation transaction inside `_handle_upload_data()`. Seems like a big refactoring, but probably the best option long-term
- we can try to create the `Data` object in a separate transaction in the same request, the commit above does this version. It works, but looks quite clumsy this way
- change the way the lock is acquired, it's the chosen variant. Can potentially be extended with several locking attempts

The current task creation design still has flaws and potential data races. For example, you can call the uploading twice and potentially get an overridden rq job, before the first one is started. Or get the `Data` model different from the task creation parameters used in the first rq job, before it locks.

cvat/apps/engine/task.py
zhiltsov-max: Agree, no issues with this.

cvat/apps/engine/task.py
SpecLad: Are you sure that the call to `self.upload_data` in `TaskViewSet.data` needs to be inside the transaction? Seems like if we moved it out, that would resolve the problem.

zhiltsov-max: It's hard to say for sure, there are 4 different branches in `upload_data`. The `init_tus_upload` looks potentially vulnerable to this. I suspect we'd need to at least keep the locks for the non `upload_finished` cases.

SpecLad: > The `init_tus_upload` looks potentially vulnerable to this.

Vulnerable to what?

zhiltsov-max: To having data races if called concurrently without a lock.

zhiltsov-max:
@codex review

PR #11012, opened 2026-08-07
e2e test ground_truth_jobs.js: fix racing in annotation mode dropdown

archibald1418:
`ground_truth_jobs.js` always fails with same problem: annotation mode dropdown gets clicked too early until the page is ready. Diagnostics was hampered by uninformative screenshots and choppy videos where no error is shown, and the logs saying the element disappared from page whilst the server logs showed successful requests all round

Ground truth jobs -- Regression tests -- GT jobs from images -- Check GT annotation source in GT job and regular job review overlay (failed)

 No retry-based fixes worked for this case, so I use `defaultWorkspace` UI param to circumvent the dropdown. The dropdown is proven to work correctly in the dedicated tests, so we don't need to commit to clicking on it every time.

If the fix proves to last, will refactor similar places to use the parameter in the next pr and where the annotation mode dropdown is not the main element under test

### DIFF
--- a/tests/cypress/e2e/features2/ground_truth_jobs.js
+++ b/tests/cypress/e2e/features2/ground_truth_jobs.js
@@ -484,7 +484,7 @@ context('Ground truth jobs', () => {
                 });
             });
 
-            it('Check GT annotation source in GT job and regular job review overlay', { scrollBehavior: false }, () => {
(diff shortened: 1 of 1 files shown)


archibald1418:
@codex review

klakhov (approved):

PR #11020, opened 2026-08-10
harden org menu for new_organinzation_pipeline

archibald1418:
#10985 didn't account for dropdown flake in organization menu, so the spinner assertions were not enough.
On close inspection, the `openOrganizationsMenu` does not assert on the filtered elements, added assertions after filtering. Also, added re-querying of the header menu to avoid DOM issues


archibald1418:
@codex review

bsekachev (approved):

PR #11022, opened 2026-08-10
Fix dropdown flake in organization pipeline

archibald1418:
### Root Cause

The flake in `tests/cypress/e2e/features/case_113_new_organization_pipeline.js`
was caused by remaining timing assumptions around organization UI state.

Two unstable paths were still present:

1. `openOrganizationsMenu()` clicked the header dropdown from a chained subject:

   ```js
   cy.get('.cvat-header-menu-user-dropdown')
       .should('exist').and('be.visible').click();
   ```

   After headless login, organization activation, or navigation, React can
   re-render the header while Cypress is waiting for the dropdown to become
   actionable. When that happens, Cypress reports that the page updated while
   the command was executing because the original DOM node was detached.

2. `inviteMembersToOrganization()` clicked `OK` and returned immediately. The
   UI sends invitation requests asynchronously and then refreshes members. The
   next test assertion could run before the organization members list reflected
   the completed invitations, which made later stateful steps fail as a cascade.

### Why Earlier Fixes Were Not Enough

- PR #9659 waited for the logged-in username after `cy.headlessLogin()`. That
  stabilized one login/header transition, but the shared organization menu
  helper still held a dropdown DOM subject across later React re-renders.

- PR #10985 waited for `.cvat-spinner` after `cy.createOrganization()` and moved
  the switcher regression setup into the suite setup hook. That improved setup
  ordering, but it did not remove the fixed animation wait in
  `openOrganizationsMenu()`, re-query dropdown elements before clicking, or wait
  for invitation requests plus a fresh members fetch.

### Fix

This is a test-only synchronization fix in
`tests/cypress/support/commands_organizations.js`.

- `openOrganizationsMenu()` now waits for the stable header username label,
  checks the dropdown trigger separately, and re-queries before clicking. It
  also waits for the `Organization` submenu item to be visible and enabled
  before re-querying and clicking it.

- `createOrganization()`, `activateOrganization()`, and
  `deactivateOrganization()` now wait for the header/spinner state that callers
  depend on before returning.

- `inviteMembersToOrganization()` now waits for every
  `POST /api/invitations` request, verifies the submitted emails, waits for the
  modal to close, and then triggers the existing organization members search to
  force a fresh `GET /api/memberships` without reloading the page.

### Notes

This avoids adding sleeps. A page reload was tried during investigation, but it
lost the active organization context and produced `You are not in an
organization`, which caused false cascading failures.

The final approach follows Cypress best practices:

- split action chains;
- re-query before clicks;
- wait on network requests that represent the state transition;
- avoid chained `clear().type()` commands.

### Validation

```bash
cd tests
yarn cypress-repeat run -n 10 --browser chrome --spec cypress/e2e/features/case_113_new_organization_pipeline.js
```

Result:

```text
***** finished 10 run(s) successfully *****
```


archibald1418:
@codex review

tests/cypress/support/commands_organizations.js
```diff
@@ -154,9 +163,28 @@ Cypress.Commands.add('inviteMembersToOrganization', (members) => {
...
+    cy.wrap(invitedEmails).should('have.members', members.map((el) => el.email));
+    cy.get('.cvat-organization-invitation-modal').should('not.exist');
+    cy.intercept('GET', '/api/memberships**').as('getOrganizationMembersAfterInvite');
+    cy.get('.cvat-organization-page-search-bar').find('input').as('organizationMembersSearch');
+    cy.get('@organizationMembersSearch').clear();
+    cy.get('@organizationMembersSearch').type('{enter}');
+    cy.wait('@getOrganizationMembersAfterInvite')
+        .its('response.statusCode')
+        .should('equal', 200);
+    cy.get('.cvat-organization-page').should('be.visible');
```
klakhov: What is the point of using empty search here?
The core app  automatically sends `GET /memberships` after all invite requests are done and then renders the members list

archibald1418: agreed, removed the empty search

klakhov:
Hmm I see in the last ran the test failed. https://github.com/cvat-ai/cvat/actions/runs/31590847703/job/94096624084?pr=11022

Can it seems getting wrong username. Can you invistigate?

archibald1418:
So I made the checks order-less as they were before. This is the one case where looping with
`for (const member of members)` vs `for (let i =0; i < members.length; i++)` matters

klakhov (approved):

PR #11023, opened 2026-08-11
As/add new user model

Rwwwrl:
### Title
Switch to a custom iam.User model

### Motivation and context

Django's stock `auth.User` cannot be extended later without a painful data migration. CVAT now owns `iam.User` (`AbstractUser`) with `AUTH_USER_MODEL = "iam.User"`, still backed by the existing `auth_user` table and M2M tables, and call sites use `get_user_model()` / `settings.AUTH_USER_MODEL` instead of importing the swapped-out model. The `auth|user` content type is re-labelled to `iam|user` in place so permissions and GenericForeignKeys keep their row ids; existing users and grants are unchanged.

### Checklist

### License


cvat/apps/iam/migrations/0003_move_user_content_type.py
```diff
@@ -0,0 +1,59 @@
...
+        # Drop that empty row so `old` can reclaim the (app_label, model) unique key.
+        rows.filter(pk=stale.pk).delete()
+
+    rows.filter(pk=old.pk).update(app_label=NEW_APP_LABEL)
+    ContentType.objects.clear_cache()
+
+
+def restore_user_content_type(apps, schema_editor):
+    ContentType = apps.get_model("contenttypes", "ContentType")
+    db_alias = schema_editor.connection.alias
```
zhiltsov-max: Why `db_alias` is used explicitly in this migration? Do we really need it?

cvat/apps/iam/migrations/0003_move_user_content_type.py
```diff
@@ -0,0 +1,59 @@
...
+
+    ContentType.objects.using(db_alias).filter(app_label=NEW_APP_LABEL, model="user").update(
+        app_label=OLD_APP_LABEL
+    )
+
```
zhiltsov-max: Shouldn't the ContentType cache be cleared after this call?

cvat/apps/iam/migrations/0003_move_user_content_type.py
```diff
@@ -0,0 +1,59 @@
...
+        # auth_permission points at. Deleting it here would cascade away every grant.
+        return
+
+    stale = rows.filter(app_label=NEW_APP_LABEL, model="user").first()
+    if stale is not None:
+        # Left behind by a previous rollback: restore_user_content_type re-labels the
+        # real row back to auth|user, then post_migrate immediately recreates iam|user
+        # (iam.User is still an installed model) with a fresh, ungranted permission set.
+        # Drop that empty row so `old` can reclaim the (app_label, model) unique key.
+        rows.filter(pk=stale.pk).delete()
```
zhiltsov-max: I think the migration has to check that the row is really unused before deleting anything. I could find the [`remove_stale_contenttypes`](https://github.com/django/django/blob/stable/6.0.x/django/contrib/contenttypes/management/commands/remove_stale_contenttypes.py) command, maybe it makes sense to ask the user to run it instead.

Rwwwrl: Updated. The only models that are using the ContentType: auth_permissions and admin.log_entry
Cases that me performed during downgrade -> upgrade window:
1. admin created new permission that refs to iam.User content_type - it is handled
2. the admin added to a user a permission that refers to iam.User content_type -it is handled
3. the same as (2), but with groups
4. admin created (not directly) new LogEntry that refs to iam.User content_type - it is handled

cvat/apps/iam/migrations/0003_move_user_content_type.py
Rwwwrl: cleared out

cvat/apps/iam/migrations/0003_move_user_content_type.py
Rwwwrl: Added

cvat/apps/iam/migrations/0003_move_user_content_type.py
Rwwwrl: Regarding the "remove_stale_contenttypes":
If that migration and the code is rollbacked, then in contentypes db table we would have iam.User and auth.User, while in the code there would be only auth.User model. In that case remove_state_contenttypes would remove the iam.User from contentypes table

Q: Is remove_state_contenttypes required in that case?
A: No, having iam.User in the db does not break a code

Q: would the application work on the db rollback and the code rollback?
A: Yes, without any manual repairing

Q: Would a new try of the release without having called the remove_state_contentypes success?
A: Yes

zhiltsov-max: Ok, so with the current implementation, my question is how it can be proven for correctness? I mean, the first time you implemented it, you didn't find the LogEntry and permission dependencies. This time you could find them, but it doesn't mean this list is complete and it will be complete in the future or in repos with more components. Why do you think it's better to have this implementation over the simple one that checks the row already exists and just suggests calling a manage.py command?

Rwwwrl: Updated to raise an exception, as we agreed - 328a381edd4ced133065bebda409b1031a89878f. Thank you

cvat/apps/iam/migrations/0003_move_user_content_type.py
```diff
@@ -0,0 +1,86 @@
...
+with a ForeignKey to ContentType or a GenericForeignKey - still points at id={new_id},
+and only you can decide which of those records should survive.
+
+To continue:
+
+  1. Re-point everything that references content type id={new_id} at id={old_id}.
+  2. Delete the "{new_label}|user" row once nothing references it.
+  3. Re-run this migration.
+"""
+
```
zhiltsov-max, line 43:
```suggestion

CONFLICTING_USER_CONTENT_TYPE_ERROR = """\
Cannot re-label the user content type: the "{db_table}" table holds both
"{old_label}|user" (id={old_id}) and "{new_label}|user" (id={new_id}) rows.
Only one of them may exist. The migration stopped without changing anything.

This is likely because this migration was applied previously and then was rolled back:

- upgrade:
   The "{new_label}|user" row is created or re-labelled from the old "{old_label}|user" in place.

- downgrade:
   The row is re-labelled back to "{old_label}|user". Django ContentTypes' post_migrate
   handler automatically creates a new "{new_label}|user" row after all migrations,
   because "{new_label}.User" is still an installed model after the migration.

- upgrade (the current call):
   Both rows now exist and conflict between each other.

This migration deliberately does not merge or delete the duplicate for you. Anything
created while "{new_label}|user" existed - permission grants, admin log entries, or any row
with a ForeignKey to ContentType or a GenericForeignKey - still points at id={new_id},
and only you can decide which of those records should survive.

To continue:

1. Re-point everything that references the "{new_label}|user" (id={new_id}) content type at
   "{old_label}|user" (id={old_id}).
2. Delete the "{new_label}|user" row once nothing references it. This can be done manually or
   via the "django manage.py remove_stale_contenttypes" command after switching to the previous
   version of the repository.
3. Re-run this migration.
"""

```

cvat/apps/iam/migrations/0003_move_user_content_type.py
```diff
@@ -0,0 +1,86 @@
...
+def move_user_content_type(apps, schema_editor):
+    ContentType = apps.get_model("contenttypes", "ContentType")
+    rows = ContentType.objects.filter(model="user")
+
+    old_content_type_row = rows.filter(app_label=OLD_APP_LABEL).first()
+    if not old_content_type_row:
+        # Fresh installation - auth.User is swapped out, so auth|user is never created and
+        # there is nothing to move. Any iam|user row here is the real one (post_migrate
+        # created it and auth_permission points at it), so it must not be deleted.
+        return
```
zhiltsov-max, line 53:
```suggestion
    old_content_type_row = rows.filter(app_label=OLD_APP_LABEL).first()
    if not old_content_type_row:
        # Fresh installation - auth.User is swapped out, so auth|user is never created and
        # there is nothing to migrate.
        return
```

cvat/apps/iam/migrations/0003_move_user_content_type.py
```diff
@@ -0,0 +1,86 @@
...
+    contenttype_post_migration_created_row = rows.filter(app_label=NEW_APP_LABEL).first()
+    if contenttype_post_migration_created_row:
+        raise CommandError(
+            STALE_USER_CONTENT_TYPE_ERROR.format(
+                old_label=OLD_APP_LABEL,
+                new_label=NEW_APP_LABEL,
+                old_id=old_content_type_row.pk,
+                new_id=contenttype_post_migration_created_row.pk,
+            )
+        )
```
zhiltsov-max:
```suggestion
    new_content_type_row = rows.filter(app_label=NEW_APP_LABEL).first()
    if new_content_type_row:
        raise CommandError(
            CONFLICTING_USER_CONTENT_TYPE_ERROR.format(
                db_table=ContentType._meta.db_table,
                old_label=OLD_APP_LABEL,
                new_label=NEW_APP_LABEL,
                old_id=old_content_type_row.pk,
                new_id=new_content_type_row.pk,
            )
        )
```

cvat/apps/iam/migrations/0003_move_user_content_type.py
```diff
@@ -0,0 +1,93 @@
...
+    ContentType = apps.get_model("contenttypes", "ContentType")
+    rows = ContentType.objects.filter(model="user")
+
+    old_content_type_row = rows.filter(app_label=OLD_APP_LABEL).first()
+    if not old_content_type_row:
+        # Fresh installation - auth.User is swapped out, so auth|user is never created and
+        # there is nothing to migrate.
+        return
+
+    contenttype_post_migration_created_row = rows.filter(app_label=NEW_APP_LABEL).first()
```
zhiltsov-max:
```suggestion
```

zhiltsov-max (approved):
LGTM, but I recommend to defer merging before all dependent PRs are ready.

PR #11024, opened 2026-08-11
Audio workspace: region bulk boundaries editing

lexiv0re:
### Motivation and context

It's not convenient to edit boundaries for adjacent intervals.

The change introduces new function to bulk move region boundaries.
* In _Cursor_ mode press and hold _Shift_ modifier key
* Hover over the position where close boundaries are located. All the targeted boundaries are highlighted in blue.
* Drag and drop to the desired position
* All targeted boundaries are moved to the new position

It helps with editing regions that semantically share a boundary.

bulk-boundaries

### How has this been tested?

Tested manually

### Checklist

### License


cvat-core/src/annotations-collection.ts
```diff
@@ -1664,6 +1664,42 @@ export default class Collection {
...
+            if (Math.abs(position - state.start) <= delta) {
+                boundaries.push({ state, side: 'start' });
+            }
+            if (Math.abs(position - (state.stop ?? this.stopFrame + 1)) <= delta) {
+                boundaries.push({ state, side: 'end' });
+            }
+        }
+
+        return boundaries;
+    }
```
bsekachev: Actually, having this method in `cvat-core` seems like an overcomplication.

It requires adding quite a bit of code to `implementTask` and `implementJob`, as well as making the interface async to stay consistent with the rest of `cvat-core`.

Perhaps this was done for consistency with `select`, but `select` relies on internal `distance` methods implemented by different objects. In this case, the logic seems much simpler and doesn't really require that level of abstraction.

So, I would suggest keeping this functionality in the UI layer instead.

cvat-core/src/annotations-collection.ts
```diff
@@ -1664,6 +1664,42 @@ export default class Collection {
...
+        try {
+            states.forEach((state) => {
+                const interval = state.clientID === null ? null : this.objects[state.clientID];
+                if (!(interval instanceof AudioInterval)) return;
+                interval.save(state);
+            });
+        } finally {
+            this.history.endTransaction();
+        }
+    }
```
bsekachev: Could we use a more generic name for this method, such as `bulkSave`?

That could make it reusable for other object types in the future. For now, we can keep the implementation limited to audio intervals only.

cvat-core/src/annotations-history.ts
```diff
@@ -10,18 +10,54 @@ const MAX_HISTORY_LENGTH = 32;
...
+            await action.redo();
+        }
+    }
+}
+
 export default class AnnotationHistory {
     private frozen: boolean;
     private _undo: ActionItem[];
     private _redo: ActionItem[];
+    private transaction: HistoryTransaction | null;
```
bsekachev, line 61: Before I added `frozen` flag to achieve similar idea? Can we re-use or just leave a better option?

cvat-core/src/annotations-collection.ts
lexiv0re: Right, it was done in consistency but with `selectInterval` rather than `select`. It uses `AudioInterval.distance` but it's the only place using it. So I feel like we should either keep both here or move both to the UI layer. 
It would probably make more sense if it weren't using `AudioIntervalState` in params but some generic `IntervalState` that'd make sense not only for audio but for any 1D interval annotation. But probably the model is generic, just the name is too specific. 
What do you think?

cvat-core/src/annotations-collection.ts
lexiv0re: Done

cvat-core/src/annotations-collection.ts
bsekachev: As I mentioned, keeping it here forces us having extra code and async interface.
Obviously - drawbacks of such solution.

If you want to move `selectInterval` to UI level, so, for consistency we probably need to remove `select` as well. 

Additionally, about consistency, it was my guess why you done this. But actually there is no correlation between object selection and finding interval borders.

bsekachev: And naming of these methods potentially may be more abstract, like

`select1D` - accepts objects + position
`select2D` - accepts objects + x,y
`select3D` - accepts objects + x,y,z

However for 3D selection uses raycast algorithm and currently implemented on canvas3d level just because this was simpler

lexiv0re: Ok, I agree `selectIntervalBoundaries` should be kept to UI without a good reason to move it here. Updated.
I kept the select methods here for now and didn't touch their naming.

cvat-core/src/annotations-history.ts
lexiv0re: I applied the new transaction approach to zOrder updates and skeleton save and it looks simpler now. It helps to collect bulk operations into one history item without constructing undo/redo manually.
But I had to keep `frozen` which is used in polygon simplification in preview mode. So these updates should actually be transparent rather than collected into a transaction while the preview is active. And then only the last preview state is actually committed as the history item.

bsekachev (approved):

PR #11031, opened 2026-08-13
Audio workspace: single unified Cursor mode for selection and editing

lexiv0re:
### Motivation and context

UX problems like: In Edit mode, hotkey H hides the wrong interval (hides last selected in Cursor/Esc mode).
The root cause is that Edit mode doesn't allow interval selection. It makes a number of workflows confusing or not convenient requiring switching between Edit/Cursor mode for simple operations. 

This change removes Edit mode completely making all the functions available in Cursor mode resolving all issues of such kind.

The primary drawback of this solution is that it increases risk of accidental moves/resizes of the regions during playback/attributes editing. It is mitigated by:
* Introducing new _Pin_ option for audio intervals similar to the existing option for image annotations. **Essentially "Pin all" button or shortcut "T P" provides an experience that is very close to the cursor mode before the changes.**
* Undo command also stays available to rollback accidental changes as well. 

With the implemented approach all visible unlocked unpinned intervals are movable/resizable in Cursor mode right away without prior selection need. Resize handles are only visible for selected/hovered/interacted intervals. We may think about requiring the selection for drag/resize later if accidental geometry changes of unrelated intervals become a problem.

Overall, it also becomes more consistent with the image annotation geometry editing process.

NOTE: this PR is based on the reworked resize handling logic from https://github.com/cvat-ai/cvat/pull/11002 because there's was an issue using the standard WS logic in the new combined Cursor mode.

### How has this been tested?

Tested manually.

### Checklist

### License


cvat-core/src/annotations-objects/audio-interval.ts
```diff
@@ -115,6 +117,10 @@ export class AudioInterval extends ScoredMixin(AnnotationBase) {
             delete updated[readOnlyField];
         }
 
+        if (this.pinned && data.pinned) {
```
bsekachev, line 120: I think we only need to check `pinned` in `data`. It can be both set or unset together with changing position.

bsekachev (approved):
lgtm

cvat-core/src/annotations-objects/audio-interval.ts
lexiv0re: Currently:
| this.pinned | data.pinned | Position updated? |
| --- | --- | --- |
| true | true | No |
| true | false | Yes |
| false | true | Yes |
| false | false | Yes |

If only checking `data.pinned`:
| this.pinned | data.pinned | Position updated? |
| --- | --- | --- |
| true | true | No |
| true | false | Yes |
| false | true | No |
| false | false | Yes |

So it already can be both set or unset together with changing position, the position update is only dropped if it was already pinned and remains pinned. 
If we check only `data.pinned` it won't be possible to update both position and pin in a single update of an unpinned interval. 
So looks like already works as expected?

bsekachev: makes sense

PR #11047, opened 2026-08-14
Fix validation of numerical attributes in the raw label editor

### LINKED ISSUE #11034: Incorrect validation for numerical attributes in the raw label editor

### Steps to Reproduce

1. Create an empty project
2. Go to the label editor, create a label with a `number` typed attribute, the values are `1;4;1` (so, 1, 2, 3, 4 should be allowed)
3. Save the labels
4. Switch to the raw label editor
5. Copy the text, paste
6. Get the `Attribute: "attr": attribute values must be unique` error. 

### Expected Behavior

No error.

### Possible Solution

Probably, the value validation is incorrect for numerical attributes in the raw editor.

### Context

_No response_

### Environment

```Markdown
CVAT v2.73.1
```

MannXo:
### Motivation and context

A number attribute could not be pasted back into the raw label editor. A label
carrying an attribute with the range `1;4;1` was rejected with
`Attribute: "attr": attribute values must be unique`.

`validateParsedAttribute` applied a single uniqueness rule to `values` for every
input type. Number attributes do not keep a list of options in that field, they
keep a `[minimum, maximum, step]` triple, so `1;4;1` is stored as
`["1", "4", "1"]` and the repeated `1` looked like a duplicate. Every range
where two of the three numbers coincide was affected.

Number attributes are now validated as a range, using the rules
`renderNumberRangeInput` already applies in the constructor editor. This closes
the opposite gap as well. A range such as `["5", "1", "0"]` used to pass the raw
editor and reach the server, and is now rejected the way the constructor editor
rejects it.

Fixes #11034

### How has this been tested?

`validateParsedLabel` was driven directly through 18 cases covering the reported
range, boundary ranges, malformed ranges, and the `select`, `radio`, `text` and
`checkbox` paths that must keep the uniqueness check. On `develop` 6 of the 18
pass. With this change all 18 pass.

The new Cypress spec asserts that `["1", "4", "1"]` raises no validation error
and leaves Save enabled, that five malformed ranges each raise their own
message, and that duplicated values on a `select` attribute are still rejected.

I was not able to run that spec. The Cypress binary fails its own smoke test on
my machine at both the pinned 13.10.0 and at 15.20.1, rejecting its launch flags
before any spec is loaded. The spec was written against a running server and UI,
but it has not been executed, so it needs the CI run to confirm.

Linters were run locally and are clean. ESLint on both changed files,
`yarn run type-check`, remark on the changelog fragment, and typos 1.40.0.

### Checklist

### License


cvat-ui/src/components/labels-editor/common.ts
```diff
@@ -18,6 +18,34 @@ export type LabelOptColor = SerializedLabel;
...
 let id = 0;
 
+// Number attributes store a [minimum, maximum, step] triple in "values"
+// instead of a list of options, so they need their own checks.
+function validateNumberAttributeRange(attrName: string, values: string[]): void {
+    if (values.length !== 3) {
+        throw new Error(`Attribute: "${attrName}": three numbers are expected`);
+    }
+
+    const numbers = values.map((value: string): number => (value.length ? Number(value) : NaN));
```
bsekachev: We already have similar code in `renderNumberRangeInput`. 
Can we reduce code duplication?

bsekachev (commented):
I do not think we want to have dedicated e2e test for this issue. 
These tests are too heavy to add them for each minor issue. Would it be possible just update existing one?

MannXo:
Done. Dropped the dedicated spec and added the two cases to `issue_1498_message_ui_raw_labels_wrong.js`, which already covers raw label editor validation.

One note on the diff. `taskRaw` in that file has no label `type`, which is a validation error in its own right. The existing cases never reach it because each of them raises an attribute error first, but a case that has to produce no error at all needs the type set, so the two new ones set it.

cvat-ui/src/components/labels-editor/common.ts
MannXo: Done. Dropped the dedicated spec

bsekachev: What about code duplication?

MannXo: reflected now

cvat-ui/src/components/labels-editor/label-form.tsx
```diff
@@ -337,32 +337,8 @@ export default class LabelForm extends React.Component<Props> {
...
-                return Promise.reject(new Error('Step must be less than minmax difference'));
-            }
-
-            if (step <= 0) {
-                return Promise.reject(new Error('Step must be a positive number'));
-            }
-
-            return Promise.resolve();
+            const error = getNumberRangeError(strNumbers.split(';'));
+            return error ? Promise.reject(new Error(error)) : Promise.resolve();
```
bsekachev: Returning error does not seem to be natural. Errors should be raised instead. 
And the function probably shall be called just 'validateNumberAttributeValues'

Also, when we return rejected promise, is there any reason to create a new `error` object?

MannXo: thanks for catching this, addressed now

bsekachev (approved):

bsekachev:
Thank you for the contribution!

PR #11048, opened 2026-08-15
Prevent deleted Lambda auto-annotation requests from overwriting a newer same-task request

### LINKED ISSUE #11016: Running Lambda auto-annotation can overwrite or replay a newer same-task request after deletion
### Summary

`DELETE /api/lambda/requests/{id}` deletes the RQ job record even when the job is
already running (`STARTED`). The work horse can keep executing after the delete.
A new request for the same task reuses the same deterministic RQ job ID
(`action=autoannotate&target=task&target_id=<task>`), so the stale job can later
write its metadata or serialized payload over the new job, and the queue can
replay the stale job's kwargs instead of the new request's.

### Affected code (`develop` @ `d215a005ae1038a786b803cbcd90e4ff953e07bc`)

- `LambdaQueue.enqueue()` derives the RQ ID only from the task
  (cvat/apps/lambda_manager/views.py)
- `RequestViewSet.destroy()` calls `rq_job.delete()` unconditionally, with no
  status-specific cancel/stop lifecycle (cvat/apps/lambda_manager/views.py)
- detector progress, cleanup, and annotation submit paths have no run-generation
  fence

### Deterministic reproduction

No real Nuclio worker is needed to observe the unsafe API contract. Mocking
`Queue.enqueue_job` to save the created job with `STARTED` status (simulating a
job whose work-horse is still running) and then exercising the public API:

1. `POST /api/lambda/requests` for a task → `200`, job saved as `STARTED`.
2. `DELETE /api/lambda/requests/{id}` while the job is `STARTED`.
3. `POST /api/lambda/requests` again for the same task, immediately after.

### Current result

- Step 2 returns `204` and deletes the Redis job hash even though the job may
  still be running.
- Step 3 is accepted with `200` and reuses the same deterministic RQ id,
  because the existing "only one running request per task" check in
  `enqueue()` only fires when a job hash is still present — which DELETE just
  removed.
- Once two live jobs can occupy the same RQ id, the older job's late
  progress/terminal write (still pending in a real deployment, e.g. a slow or
  hung Nuclio call) can overwrite the newer job's status/meta/payload in
  Redis, and the newer job's queued entry can dequeue and re-execute the
  older job's kwargs instead.

### Expected result

- A running job's identity is not deleted or reused before some form of
  terminal acknowledgement.
- A cannot update B's metadata, status, payload, cleanup flag, or annotations.
- An accepted new request executes exactly once, with its own kwargs.

### Proposed minimum policy

I'd like to confirm which initial policy is preferred before opening a PR:

1. **Reject cancellation of a running (started) auto-annotation job.** DELETE
   on a `STARTED` job returns `400`/`409` and leaves the Redis hash untouched;
   the job keeps running to completion. QUEUED/DEFERRED/SCHEDULED jobs can
   still be canceled and their hash removed immediately, freeing the task
   slot. This is the smallest change: it only touches the DELETE lifecycle
   and reuses the existing "one running request per task" enqueue check once
   the hash is no longer removed prematurely.
2. **Support real cancellation of a running job** (stop-and-ack): send RQ's
   `send_stop_job_command()`, keep the hash until the work-horse
   acknowledges the stop, and reject new same-task requests until then. This
   gives users an actual "cancel and restart" UX but requires more surface
   area (progress/annotation-write guards, partial-result handling for
   already-committed chunks, idempotent stop-failure handling).

I have a local, focused implementation and a deterministic regression test for
option 1, and can open a PR once the direction is confirmed. Option 1 does not
address two separate, already-known duplicate-annotation sources (repeated
`cleanup=false` runs appending on purpose, and partial chunks from a canceled
run persisting into the next run) — those look like they need their own
follow-up policy and are out of scope for this first fix.

ksm463:
### Motivation and context

Fixes #11016.

`DELETE /api/lambda/requests/{id}` previously deleted the RQ job hash even
while the job was `STARTED`. Lambda auto-annotation uses a deterministic RQ job
ID derived from the target task. Deleting the hash therefore made the ID
available for a new request while the previous work horse could still be
running and submitting annotations.

This allowed the previous execution to overwrite the newer request's
progress/result or to submit annotations into the same task after the new run
had started.

This PR changes deletion of a running Lambda request to stop the work horse and
wait for the stop to be acknowledged before releasing the deterministic job ID.

### What has changed?

- `LambdaQueue.delete_job()` takes the same per-job RQ lock as `enqueue()`, so
  DELETE and POST cannot race over the same deterministic job ID.
- The object permission check is performed inside that lock on the exact RQ job
  instance that will be stopped and deleted. This prevents an ID-reuse race from
  applying permissions from an older job to its replacement.
- For a `STARTED` job, DELETE sends `send_stop_job_command` and waits until the
  job leaves the `STARTED` state before deleting its Redis hash and returning
  `204 No Content`.
- If the worker cannot stop started jobs, or the stop is not acknowledged within
  10 seconds, DELETE returns `409 Conflict` and keeps the job hash. The ID then
  remains unavailable to a new request while the previous execution may still
  be running.
- `QUEUED`, `DEFERRED`, and `SCHEDULED` jobs are canceled with
  `enqueue_dependents=settings.ONE_RUNNING_JOB_IN_QUEUE_PER_USER` before their
  hashes are deleted.
- The worker capability check used by export cancellation was extracted into a
  shared helper and reused by the Lambda cancellation path.
- The OpenAPI response descriptions and changelog fragment were updated for the
  new cancellation behavior.

### How has this been tested?

The tests cover:

- successful cancellation of a started request and safe reuse of its
  deterministic ID;
- a worker that cannot stop started jobs, including the `409` response and job
  hash retention;
- a stop command that is not acknowledged before the timeout, including the
  `409` response and job hash retention;
- permission checking under the per-job lock when the deterministic ID is
  replaced by a request owned by another user;
- the existing queued and terminal-state deletion behavior.

Because `cvat.settings.testing` runs RQ queues synchronously, the started-job
tests simulate the RQ lifecycle and stop acknowledgement without requiring a
live Nuclio function.

Tests run after rebasing onto `upstream/develop`:

```bash
python manage.py test --settings cvat.settings.testing \
  cvat.apps.lambda_manager.tests.test_lambda
python manage.py test --settings cvat.settings.testing \
  cvat.apps.redis_handler.tests.test_views
```

Results:

- Lambda manager: 71 tests run, 3 skipped, OK
- Redis handler cancellation tests: 3 tests run, OK
- Django system checks: no issues
- OpenAPI generation: 0 errors; the generated Lambda DELETE response block
  matches `cvat/schema.yml`
- `git diff --check upstream/develop...HEAD`: passed

### Checklist

### License

- [x] I submit _my code changes_ under the same
  [MIT License](https://github.com/cvat-ai/cvat/blob/develop/LICENSE) that covers the project.

### DIFF
--- a/changelog.d/20260809_031454_ksm46351_lambda_aba_fix.md
+++ b/changelog.d/20260809_031454_ksm46351_lambda_aba_fix.md
@@ -1,8 +1,10 @@
 ### Fixed
 
-- \[Server API\] Deleting a still-running Lambda auto-annotation request no
-  longer frees up its deterministic RQ job ID for immediate reuse, which
-  could let a later request for the same task overwrite or replay the
-  in-flight one. `DELETE` on a running request now returns `409 Conflict`
-  instead of `204`.
+- \[Server API\] Canceling a running Lambda auto-annotation request now stops the
+  work horse process instead of only deleting the RQ job hash. Previously the
+  deleted request kept running and submitting annotations while its
+  deterministic RQ job ID was already free for reuse, so a later request for the
+  same task could be overwritten or replayed by the old execution. If the work
+  horse cannot be stopped, the API returns `409 Conflict` and keeps the RQ job
+  hash to prevent the request ID from being reused
   (<https://github.com/cvat-ai/cvat/pull/11048>)
--- a/changelog.d/20260809_031454_ksm46351_lambda_aba_fix.md
+++ b/changelog.d/20260809_031454_ksm46351_lambda_aba_fix.md
@@ -1,0 +1,8 @@
+### Fixed
+
+- \[Server API\] Deleting a still-running Lambda auto-annotation request no
+  longer frees up its deterministic RQ job ID for immediate reuse, which
+  could let a later request for the same task overwrite or replay the
+  in-flight one. `DELETE` on a running request now returns `409 Conflict`
+  instead of `204`.
+  (<https://github.com/cvat-ai/cvat/pull/XXXX>)
--- a/changelog.d/20260809_031454_ksm46351_lambda_aba_fix.md
+++ b/changelog.d/20260809_031454_ksm46351_lambda_aba_fix.md
@@ -5,4 +5,4 @@
   could let a later request for the same task overwrite or replay the
   in-flight one. `DELETE` on a running request now returns `409 Conflict`
   instead of `204`.
-  (<https://github.com/cvat-ai/cvat/pull/XXXX>)
+  (<https://github.com/cvat-ai/cvat/pull/11048>)
--- a/cvat/apps/lambda_manager/tests/test_lambda.py
+++ b/cvat/apps/lambda_manager/tests/test_lambda.py
@@ -9,6 +9,7 @@ import json
 import os
 from collections import Counter
 from contextlib import contextmanager
+from datetime import timedelta
 from itertools import groupby
 from unittest import mock, skip
 
@@ -18,7 +19,9 @@ from django.http import HttpResponseNotFound, HttpResponseServerError
 from PIL import Image
 from rest_framework import status
 from rq.job import JobStatus
+from rq.worker import Worker as RQWorker
 
+from cvat.apps.engine.rq import RQJobMetaField
 from cvat.apps.engine.tests.utils import (
     ApiTestBase,
     ForceLogin,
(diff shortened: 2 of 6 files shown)


bsekachev:
Hello,

That would be more correct to stop work horse process with `send_stop_job_command`.
We earlier implemented it for export functionality: https://github.com/cvat-ai/cvat/pull/10600

ksm463:
Thanks for the suggestion. I reworked the cancellation flow to stop the work
horse with `send_stop_job_command`.

The DELETE endpoint now waits for the RQ job to leave the `STARTED` state before
deleting its Redis hash. If the worker cannot stop the job or does not acknowledge
the stop within the 10-second timeout, the endpoint returns `409 Conflict` and
keeps the hash. This prevents the deterministic request ID from being reused
while the previous execution may still be running.

The permission check is now performed under the same per-job lock and against
the exact job instance that is stopped and deleted. This also closes the race
where the deterministic ID could be reused by a request owned by another user.

I also added regression coverage for successful cancellation, workers that
cannot stop started jobs, stop timeouts, deterministic ID reuse, and the
permission race.


Could a maintainer approve the CI workflows for the updated commits? They are currently in `action_required`.

PR #11049, opened 2026-08-15
feat(sdk-examples): add domain-organized SDK examples

Pr1ncee:
### Motivation and context

  The CVAT Python SDK currently ships with only a handful of ad-hoc snippets scattered across the docs. New users have no single, runnable, end-to-end reference for the common workflows (authenticate, create/inspect/export projects and tasks, manage jobs, register cloud storage).

  This branch adds a domain-organized set of copy-and-run recipes under `cvat-sdk/examples/`, a matching "Examples" section in the SDK docs site, and a pytest harness that executes every recipe as a subprocess against a live CVAT instance. The result:

  - users get an authoritative, self-contained starting point for each domain (auth, projects, tasks, jobs, cloud storage), with a consistent `argparse` interface (`--host`, `--token`, `--cleanup`, plus recipe-specific flags);
  - docs and recipes stay in lockstep — the docs pages mirror the scripts one-to-one;
  - regressions in recipe behavior (or in the SDK surface the recipes lean on) are caught by CI instead of by users.

  ### How has this been tested?

  - Ran the new pytest suite `tests/python/sdk/test_examples.py` under the standard CVAT REST-API test setup (docker-compose stack with CVAT server + MinIO), which invokes each recipe as a real subprocess via `python <recipe>.py --host <BASE_URL> --token <PAT> ...` and asserts on both stdout and side effects (created/renamed/deleted projects, job assignments and CSV report, project status CSV, backup zip round-trip via `project_backup.py` → `project_restore.py`, dataset export zip size, cloud-storage register/update/delete).
  - External-service recipes (`cloud_storage_register.py`, `task_create_from_cloud.py`, `tasks_bulk_from_cloud.py`, `task_inspect_and_export.py`, `project_export_dataset.py`) are gated behind `@pytest.mark.with_external_services` and were exercised against the MinIO container.
  - Auth-profile isolation was verified by pointing the subprocess `AuthStore` at a private tmp path via `XDG_CONFIG_HOME` + `HOME`, seeding a profile with `AuthStore.put_profile`, and running `auth_profiles.py` with and without `--profile`, plus a negative test (`test_auth_profiles_missing_default_fails`) that asserts the "No default profile configured" error path.
  - Also ran each recipe manually with `--help` to confirm the argparse surface matches the README table, and spot-checked `auth_cli.py` against a real CVAT deployment using the `--auth USER:PASS` shape shared with `cvat-cli`.

### Checklist

### License


changelog.d/20260814_211211_sdk_domain_examples.md
```diff
@@ -0,0 +1,3 @@
+### Added
+
+- SDK: New CVAT SDK usage examples under `cvat-sdk/examples/`.
```
zhiltsov-max, line 3: Missing PR link

cvat-sdk/examples/cloud_storage_register.py
```diff
@@ -0,0 +1,96 @@
...
+    return value
+
+
+HOST = require_env("CVAT_HOST", "your CVAT server URL, e.g. https://app.cvat.ai")
+TOKEN = require_env("CVAT_ACCESS_TOKEN", "create one in the CVAT UI: Profile -> Security")
+BUCKET = require_env("S3_BUCKET", "the bucket name, e.g. my-bucket")
+ACCESS_KEY = require_env("S3_ACCESS_KEY", "the bucket's access key id")
+SECRET_KEY = require_env("S3_SECRET_KEY", "the bucket's secret key")
+ENDPOINT_URL = require_env("S3_ENDPOINT_URL", "e.g. https://s3.amazonaws.com or http://minio:9000")
+CLEANUP = os.environ.get("CVAT_EXAMPLES_CLEANUP") == "1"
```
zhiltsov-max: Why did you choose to use env vars instead of CLI args? I think with argparse and CLI args we'd get automatic launch documentation via `--help`. Also, the examples currently assume a Unix-like shell, and switching to CLI args should make the examples applicable for windows as well.

cvat-sdk/examples/cloud_storage_register.py
```diff
@@ -0,0 +1,96 @@
...
+retrieve, and update it. Any S3-compatible service works (AWS S3, minio, ...)
+via the AWS_S3_BUCKET provider and a custom endpoint URL.
+
+There is no high-level proxy for cloud storages yet, so this recipe uses the
+low-level API (client.api_client.cloudstorages_api).
+
+Steps:
+  1. Register the bucket with key/secret credentials.
+  2. List all registered storages.
+  3. Retrieve the new one (the server never returns credentials).
```
zhiltsov-max:
```suggestion
  3. Retrieve the new one.
```
What is the point of saying it in the list of steps? Probably, a line comment next to the call is enough.

cvat-sdk/examples/cloud_storage_register.py
```diff
@@ -0,0 +1,96 @@
...
+  5. Optionally detach it (CVAT_EXAMPLES_CLEANUP=1) — the bucket's contents
+     are never touched.
+
+Usage:
+  export CVAT_HOST=https://app.cvat.ai
+  export CVAT_ACCESS_TOKEN=...    # CVAT UI: Profile -> Security
+  export S3_BUCKET=my-bucket
+  export S3_ACCESS_KEY=...
+  export S3_SECRET_KEY=...
+  export S3_ENDPOINT_URL=https://s3.amazonaws.com
```
zhiltsov-max: I think we should add single quotes around the env variable values, otherwise some values can be treated differently by the shell.

cvat-sdk/examples/cloud_storage_register.py
```diff
@@ -0,0 +1,96 @@
...
+                credentials_type="KEY_SECRET_KEY_PAIR",
+                key=ACCESS_KEY,
+                secret_key=SECRET_KEY,
+                specific_attributes=f"endpoint_url={ENDPOINT_URL}",
+            )
+        )
+        print(f"Registered cloud storage {storage.id} -> {BUCKET}")
+
+        # 2. List
+        page, _ = api.list()
```
zhiltsov-max: I think it makes sense to include an example for `get_paginated_collection()` here, as it's supposed to be used together with `list()` methods in the low-level API. Otherwise the users will tend to reinvent the wheel for getting the paginated list.

cvat-sdk/examples/auth_profiles.py
```diff
@@ -0,0 +1,79 @@
...
+Create a profile once on the command line, then any script can use it:
+
+  cvat-cli --server-host https://app.cvat.ai profile create --name app --set-default
+
+Steps:
+  1. If CVAT_PROFILE is set, use that profile; otherwise try the default profile.
+  2. If no profile exists and CVAT_USERNAME/CVAT_PASSWORD are set, fall back to
+     the DEPRECATED password sign-in (kept for local/dev servers only).
+  3. Print who you are authenticated as.
+
```
zhiltsov-max: I think, it would be better to keep only the profile auth and add another example on how to use the `make_client_from_cli` function to make a CLI-compatible script instead. The problem with the current variant is that it's nether a recommended, nor a full, nor a go-to client creation routine, that would be replaced either by token auth or by the CLI params in practice. If we change it as suggested, we'll get examples for the 2 or 3 recommended options - a general one for token clients, one for custom CLI-integrated clients (using profiles), and one for custom CLI-like clients, all practical.

cvat-sdk/examples/cloud_storage_register.py
```diff
@@ -0,0 +1,96 @@
...
+There is no high-level proxy for cloud storages yet, so this recipe uses the
+low-level API (client.api_client.cloudstorages_api).
+
+Steps:
+  1. Register the bucket with key/secret credentials.
+  2. List all registered storages.
+  3. Retrieve the new one.
+  4. Update its display name.
+  5. Optionally detach it (CVAT_EXAMPLES_CLEANUP=1) — the bucket's contents
+     are never touched.
```
zhiltsov-max:
```suggestion
  5. Optionally, detach it from CVAT.
```
This part also doesn't seem relevant to the list here.

cvat-sdk/examples/cloud_storage_register.py
```diff
@@ -0,0 +1,96 @@
...
+
+"""Attach an S3-compatible bucket to CVAT as a cloud storage, then list,
+retrieve, and update it. Any S3-compatible service works (AWS S3, minio, ...)
+via the AWS_S3_BUCKET provider and a custom endpoint URL.
+
+There is no high-level proxy for cloud storages yet, so this recipe uses the
+low-level API (client.api_client.cloudstorages_api).
+
+Steps:
+  1. Register the bucket with key/secret credentials.
```
zhiltsov-max:
```suggestion
  1. Attach the bucket with key/secret credentials to CVAT.
```

cvat-sdk/examples/job_list_and_assign.py
```diff
@@ -0,0 +1,65 @@
...
+
+"""List an existing task's jobs, find the unassigned ones, and distribute them
+across annotators round-robin (CVAT has no built-in auto-assignment).
+
+Steps:
+  1. List all jobs of the task with their stage/state/assignee.
+  2. Filter the jobs that have no assignee yet.
+  3. Assign them round-robin across CVAT_ASSIGNEE_IDS — or all to you,
+     if CVAT_ASSIGNEE_IDS is not set.
+
```
zhiltsov-max: I think it would be better to represent self-assignment with `me` instead. Having the variable unset looks like an error (in the context of these examples) and having it empty looks like an unassignment request.

cvat-sdk/examples/job_list_and_assign.py
```diff
@@ -0,0 +1,65 @@
...
+#
+# SPDX-License-Identifier: MIT
+
+"""List an existing task's jobs, find the unassigned ones, and distribute them
+across annotators round-robin (CVAT has no built-in auto-assignment).
+
+Steps:
+  1. List all jobs of the task with their stage/state/assignee.
+  2. Filter the jobs that have no assignee yet.
+  3. Assign them round-robin across CVAT_ASSIGNEE_IDS — or all to you,
```
zhiltsov-max: Currently, the example is not really practical. You need to get a list of users somehow, and when you have them because you maintain them, they typically do not contain ids. Consider providing a way to retrieve a list of such users. I think it would work fine, if there was an example on getting a list of users satisfying a search query. So you could get a list of users in the first call and then use the output to assign users in the second call.

cvat-sdk/examples/job_list_and_assign.py
```diff
@@ -0,0 +1,65 @@
...
+        print(f"Unassigned jobs: {[job.id for job in unassigned]}")
+
+        # 3. Round-robin assignment. To pull a team automatically instead of
+        # passing ids, use client.users.list(...).
+        assignees = ASSIGNEE_IDS or [client.users.retrieve_current_user().id]
+        for i, job in enumerate(unassigned):
+            user_id = assignees[i % len(assignees)]
+            job.update(models.PatchedJobWriteRequest(assignee=user_id))
+            print(f"Assigned job {job.id} -> user {user_id}")
+
```
zhiltsov-max: We had a really practical example with returning a csv table in the result, consider doing it here. Such a table can be easily imported into other tools, e.g. for analysis or bookkeeping.

cvat-sdk/examples/job_list_and_assign.py
```diff
@@ -0,0 +1,65 @@
...
+def main() -> None:
+    with make_client(HOST, access_token=TOKEN) as client:
+        # 1. List all jobs of the task
+        jobs = client.jobs.list(filter=F.task_id == TASK_ID)
+        print(f"Task {TASK_ID} has {len(jobs)} jobs")
+        for job in jobs:
+            assignee = job.assignee.username if job.assignee else "-"
+            print(f"  job {job.id}: stage={job.stage}, state={job.state}, assignee={assignee}")
+
```
zhiltsov-max: This part can be a standalone example of job listing in a task. I don't think it really makes sense in this example though.

cvat-sdk/examples/job_list_and_assign.py
```diff
@@ -0,0 +1,65 @@
...
+        # 2. Only the unassigned ones
+        unassigned = client.jobs.list(filter=all_(F.task_id == TASK_ID, not_(F.assignee.is_set())))
+        print(f"Unassigned jobs: {[job.id for job in unassigned]}")
```
zhiltsov-max: Consider adding a filter on the job stage and state (basically, `annotation` - `new` should work fine).

cvat-sdk/examples/job_list_and_assign.py
```diff
@@ -0,0 +1,65 @@
...
+
```
zhiltsov-max:
```suggestion
        task = client.tasks.retrieve(TASK_ID)
        unassigned = client.jobs.list(filter=all_(F.task_id == TASK_ID, not_(F.assignee.is_set())))
        print(f"Unassigned jobs: {[job.id for job in unassigned]} out of {task.jobs.count}")

```
Consider doing something like this instead of (1).

cvat-sdk/examples/job_workflow.py
```diff
@@ -0,0 +1,72 @@
...
+"""Drive a job through its workflow: pick the most recently updated job of a
+task, import annotations into it, and move it to the validation stage.
+
+Steps:
+  1. List the task's jobs, most recently updated first (server-side ordering;
+     the same endpoint also accepts free-text search, e.g. search="alice").
+  2. Import annotations from a file into the first job. The file's format must
+     match ANNOTATIONS_FORMAT (an importer name, e.g. "COCO 1.0").
+  3. Verify the shapes arrived, then move the job to the validation stage.
+
```
zhiltsov-max: This workflow is quite meaningless. Basically, if you have a recently updated job (likely, annotated), why would you need to override the annotations? And why would you need to annotate the job manually, if you had annotations?

I'd suggest making a script that finds all `annotation` - `completed` jobs and moves them to the review stage, or finds all `review` - `completed` jobs to move them to the `acceptance` stage. Then prints the list of modified jobs.

cvat-sdk/examples/job_workflow.py
```diff
@@ -0,0 +1,72 @@
...
+        job = jobs[0]
+        print(f"Working with job {job.id} (stage={job.stage}, state={job.state})")
+
+        # 2. Import annotations
+        job.import_annotations(ANNOTATIONS_FORMAT, ANNOTATIONS_PATH)
+        print(f"Imported {ANNOTATIONS_FORMAT} annotations from {ANNOTATIONS_PATH}")
+        shapes = job.get_annotations().shapes
+        print(f"Job {job.id} now has {len(shapes)} shapes")
+
+        # 3. Advance the workflow stage: annotation -> validation -> acceptance
```
zhiltsov-max: The comment is invalid. There's nothing about the `acceptance` stage in the code.

cvat-sdk/examples/project_backup_restore.py
```diff
@@ -0,0 +1,63 @@
+# Copyright (C) CVAT.ai Corporation
+#
+# SPDX-License-Identifier: MIT
+
+"""Back up an existing project to a zip and restore it as a new copy.
+
+A backup contains the project's tasks, jobs, annotations, and settings, so
+this doubles as a copy-a-project recipe.
+
```
zhiltsov-max: I think you rarely need to make both actions together in a single call. It's a valid for the project transfer case, but this example does not do a transfer.

cvat-sdk/examples/task_create_from_images.py
```diff
@@ -0,0 +1,102 @@
...
+
+Steps:
+  1. Collect *.jpg / *.jpeg / *.png files from IMAGE_DIR (sorted).
+  2. Create the task and upload the images. With CVAT_PROJECT_ID the task is
+     created inside that project and inherits its labels; without it, the task
+     gets its own labels from CVAT_LABELS.
+  3. List tasks currently in the "annotation" status.
+  4. Retrieve the new task by id and rename it.
+  5. Optionally delete it (CVAT_EXAMPLES_CLEANUP=1).
+
```
zhiltsov-max: I think we have enough coverage for single task creation from images already. We have a hello world example in the high-level SDK docs and we have a CLI command for this. I'd suggest changing the example to show bulk task creation from cloud images or videos. Putting a set of tasks into a project would make even more sense.

cvat-sdk/examples/project_status_report.py
```diff
@@ -0,0 +1,85 @@
...
+                "job_id",
+                "stage",
+                "state",
+                "assignee",
+                "frames",
+            ]
+        )
+        for task in project.get_tasks():
+            for job in task.get_jobs():
+                assignee = job.assignee.username if job.assignee else ""
```
zhiltsov-max: I'd call this pair an antipattern, please replace it with `client.jobs.list(<project_id filter>)`

cvat-sdk/examples/cloud_storage_register.py
Pr1ncee: I chose env vars because it was the easiest way to implement working examples, CLI args required more efforts and maintance

cvat-sdk/examples/project_status_report.py
Pr1ncee: thanks, replaced

cvat-sdk/examples/job_list_and_assign.py
Pr1ncee: made it flexible for user to select what state and stage he wants to list

cvat-sdk/examples/job_list_and_assign.py
Pr1ncee: thanks for the advice, added a csv table with rows `job_id`, `previous_assignee`, `new_assignee`, `new_assignee_id`

cvat-sdk/examples/README.md
```diff
@@ -0,0 +1,39 @@
...
+Conventions: recipes that create resources keep them (pass `--cleanup` to
+delete them at the end); recipes that inspect or export take an existing id via
+`--project-id` / `--task-id`; list-valued options accept multiple values
+(e.g. `--labels car person`).
+
+| Recipe | What it does | Recipe-specific flags |
+| --- | --- | --- |
+| `auth_connect.py` | Connect with a Personal Access Token, whoami | — |
+| `auth_profiles.py` | Authenticate from a saved profile | `--profile` (omit for the default profile) |
+| `auth_cli.py` | Build a CLI-compatible script via `make_client_from_cli` | reuses cvat-cli's flags: `--server-host`, `--auth`, `--profile`, ... |
```
zhiltsov-max:
```suggestion
# CVAT SDK Examples

Complete, copy-and-run examples for the CVAT Python SDK. Each example is a command-line script.
Invoke a script with the `--help` parameter to see all its options. Example:

\```sh
python tasks_bulk_from_cloud.py --help
python tasks_bulk_from_cloud.py \
    --host 'https://app.cvat.ai' \
    --token '<your token>' \
    --cloud-storage-id 7 --project-id 42 \
    --task 'videos/clip_01.mp4' \
    --task 'videos/clip_02.mp4'
\```

Most of the examples use Personal Access Tokens for user authentication. You'll need to create one
to run a script, the instructions are available [here](https://docs.cvat.ai/docs/api_sdk/access_tokens/).

Conventions:
- examples that create resources keep them by default. Pass `--cleanup` to delete them at the end
- examples that operate with existing server objects take an id
  as the parameter (e.g. `--project-id`)
- list-valued options accept multiple values (e.g. `--labels car person`)

| Example | What it does | Parameters |
| --- | --- | --- |
| `auth_connect.py` | Authenticate with a Personal Access Token, get current user | — |
| `auth_profiles.py` | Authenticate from a saved profile | `--profile` (omit for the default profile) |
| `auth_cli.py` | Build a CLI-compatible script via `make_client_from_cli` | reuses cvat-cli's [flags](https://docs.cvat.ai/docs/api_sdk/cli/#authentication): `--server-host`, `--auth`, `--profile`, ... |
```

Please clean the escaping slashes for the backquotes, I added them for github to parse this correctly.

cvat-sdk/examples/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+Conventions: recipes that create resources keep them (pass `--cleanup` to
+delete them at the end); recipes that inspect or export take an existing id via
+`--project-id` / `--task-id`; list-valued options accept multiple values
+(e.g. `--labels car person`).
+
+| Recipe | What it does | Recipe-specific flags |
+| --- | --- | --- |
+| `auth_connect.py` | Connect with a Personal Access Token, whoami | — |
+| `auth_profiles.py` | Authenticate from a saved profile | `--profile` (omit for the default profile) |
```
zhiltsov-max: Consider renaming to `auth_token.py` and `auth_profile.py`. I think it will be more clear to the users.

Pr1ncee: but the script is about how to auth using profiles, I believe it's quite clear now, isn't it?

cvat-sdk/examples/auth_cli.py
```diff
@@ -0,0 +1,47 @@
...
+  2. Add your own script-specific arguments on top.
+  3. Hand the parsed namespace to make_client_from_cli — it picks the right
+     Client factory (profile / PAT / password) from the arguments.
+
+Usage (run ``python auth_cli.py --help`` for the full list of options):
+  python auth_cli.py --profile app
+  python auth_cli.py --server-host 'https://app.cvat.ai'   # uses CVAT_ACCESS_TOKEN env
+  python auth_cli.py --server-host 'https://app.cvat.ai' --auth me:secret
+"""
+
```
zhiltsov-max, line 27:
```suggestion

Steps:
  1. Register the shared auth flags with configure_client_auth_arguments().
  2. Add your own script-specific arguments on top.
  3. Hand the parsed namespace to make_client_from_cli() to create a server API client object.

Usage (run ``python auth_cli.py --help`` for the full list of options):
  python auth_cli.py --profile app

  export CVAT_ACCESS_TOKEN='<token>'
  python auth_cli.py --server-host 'https://app.cvat.ai'

  python auth_cli.py --server-host 'https://app.cvat.ai' --auth me:secret
"""

```
- It would be better to make the functions more visible in the text
- The SDK implementation details about factories are not needed here
- It would be nice to show how to set an environment variable in the example

cvat-sdk/examples/README.md
zhiltsov-max: From what I can see, it allows passing a profile or using the default profile. There's no other work with profiles. It's mostly fine now, but when you're looking for an auth example, you'd see either `auth_token` or `auth_profile` - and it's clear what to choose. With `auth_profiles` I'd expect some more profile work, but it's not intended for the example.

cvat-sdk/examples/auth_cli.py
Pr1ncee: > It would be better to make the functions more visible in the text

I apologize, but I'm not sure how to highlight them in the dostring inside python script

zhiltsov-max: My suggestion would be to enclose the name in quotes or add braces after, just as in the comment above. It's the standard way of highlighting something in the plain text, and it works fine here.

Pr1ncee: > It would be better to make the functions more visible in the text

Got it, done for both: arguments and function names since it's also important to distinguish arguments

--- 
> It would be nice to show how to set an environment variable in the example

Added example how to export env variable in PowerShell in addition to MacOS/Linux

cvat-sdk/examples/README.md
Pr1ncee: ok, replaced to `auth_token.py`

cvat-sdk/examples/job_assign.py
```diff
@@ -0,0 +1,116 @@
...
+import sys
+from pathlib import Path
+
+from cvat_sdk import make_client, models
+from cvat_sdk.core.filters import F, all_, not_
+from cvat_sdk.core.proxies.users import User
+
+
+def parse_args() -> argparse.Namespace:
+    parser = argparse.ArgumentParser(description=__doc__.splitlines()[0])
```
zhiltsov-max: This line produces a truncated usage line ending with "and". But do we really need to split the descriptions, why not just output the whole text?

cvat-sdk/examples/README.md
Pr1ncee: renamed `auth_connect.py` -> `auth_token.py` and `auth_profiles.py` -> `auth_profile.py`

cvat-sdk/examples/job_assign.py
```diff
@@ -0,0 +1,116 @@
...
+  4. Write assignments.csv into the current directory.
+
+Usage (run ``python job_assign.py --help`` for the full list of options):
+  python job_assign.py --host 'https://app.cvat.ai' --token '<your token>' \\
+      --task-id 42                              # self-assign
+  python job_assign.py --host 'https://app.cvat.ai' --token '<your token>' \\
+      --task-id 42 --assignees alice bob
+  python job_assign.py --host 'https://app.cvat.ai' --token '<your token>' \\
+      --task-id 42 --search 'annotator-team'    # pool = every match of the search
+"""
```
zhiltsov-max, line 31: I think this example is hard to use on practice. The search field will check matches in the `username`, `first_name`, and `last_name` fields, so it can only work for a team of uniformly named accounts (e.g. `worker1`, `worker2`, ...) or a team of people with same names, which is unlikely. The idea is good though, I think it would better be represented via supplying an org slug/org id and searching people there (maybe with some _role_), and will actually be practical in this case.

cvat-sdk/examples/job_assign.py
Pr1ncee: I trimmed the description to 1 line because I think it's redundant here. The full script is described in 2 places (in the script itself and documentation) so in my opinion description here should display only brief explanation about the script itself.
It was truncated to the first row because I forgot updating it since it's changed. Thanks, I made it to 2 first rows

cvat-sdk/examples/job_list.py
```diff
@@ -0,0 +1,59 @@
...
+        required=True,
+        help="Personal Access Token (CVAT UI: Profile -> Security)",
+    )
+    parser.add_argument(
+        "--task-id", type=int, required=True, help="id of an existing task, e.g. 42"
+    )
+    parser.add_argument("--stage", help="only jobs at this stage, e.g. 'annotation'")
+    parser.add_argument("--state", help="only jobs in this state, e.g. 'new'")
+    return parser.parse_args()
+
```
zhiltsov-max, line 54: Could you extend the script with options to pass a project id and an option to save the results as csv? It would make it really useful. This will basically merge it with `project_status_report.py`, so we'll need only 1 of them.

cvat-sdk/examples/project_export_dataset.py
```diff
@@ -0,0 +1,85 @@
...
+    parser.add_argument(
+        "--cloud-storage-id",
+        type=int,
+        required=True,
+        help="a registered cloud storage id (see cloud_storage_register.py)",
+    )
+    parser.add_argument(
+        "--export-format",
+        default="COCO 1.0",
+        help="exporter name, e.g. 'COCO 1.0' (default: 'COCO 1.0')",
```
zhiltsov-max:
```suggestion
        help="exporter name, e.g. 'COCO 1.0' (default: '%(default)s')",
```
Please update how other defaults are reported as well.

cvat-sdk/examples/project_export_dataset.py
```diff
@@ -0,0 +1,85 @@
...
+            sys.exit(
+                f"Unknown export format {args.export_format!r}. Choose one of: {', '.join(names)}"
+            )
+
+        project = client.projects.retrieve(args.project_id)
+
+        # 2. Export to a local zip
+        local_path = Path(f"project_{project.id}_dataset.zip")
+        project.export_dataset(
+            args.export_format, local_path, include_images=True, location=Location.LOCAL
```
zhiltsov-max: I think it's better to export without images by default, the same as in the CLI command. It will make the example available to all users, including both self-hosted, enterprise, and SaaS (free and paid). Export with images is a paid option on the SaaS.

cvat-sdk/examples/project_export_dataset.py
```diff
@@ -0,0 +1,85 @@
...
+import argparse
+import sys
+from pathlib import Path
+
+from cvat_sdk import make_client
+from cvat_sdk.core.proxies.types import Location
+
+
+def parse_args() -> argparse.Namespace:
+    parser = argparse.ArgumentParser(description=__doc__.splitlines()[0])
```
zhiltsov-max: Truncated description.

cvat-sdk/examples/project_status_report.py
```diff
@@ -0,0 +1,90 @@
...
+from pathlib import Path
+
+from cvat_sdk import make_client
+from cvat_sdk.core.filters import F
+from cvat_sdk.core.proxies.jobs import Job
+from cvat_sdk.core.proxies.projects import Project
+
+
+def parse_args() -> argparse.Namespace:
+    parser = argparse.ArgumentParser(description=__doc__.splitlines()[0])
```
zhiltsov-max: Truncated description.

cvat-sdk/examples/task_create_from_cloud.py
```diff
@@ -0,0 +1,82 @@
...
+"""
+
+import argparse
+
+from cvat_sdk import make_client, models
+from cvat_sdk.core.proxies.tasks import ResourceType
+
+
+def parse_args() -> argparse.Namespace:
+    parser = argparse.ArgumentParser(description=__doc__.splitlines()[0])
```
zhiltsov-max: Truncated description.

cvat-sdk/examples/task_inspect_and_export.py
```diff
@@ -0,0 +1,91 @@
...
+import argparse
+import sys
+from pathlib import Path
+
+from cvat_sdk import make_client
+from cvat_sdk.core.proxies.types import Location
+
+
+def parse_args() -> argparse.Namespace:
+    parser = argparse.ArgumentParser(description=__doc__.splitlines()[0])
```
zhiltsov-max: Truncated description.

cvat-sdk/examples/tasks_bulk_from_cloud.py
```diff
@@ -0,0 +1,116 @@
...
+"""
+
+import argparse
+
+from cvat_sdk import make_client, models
+from cvat_sdk.core.proxies.tasks import ResourceType
+
+
+def parse_args() -> argparse.Namespace:
+    parser = argparse.ArgumentParser(description=__doc__.splitlines()[0])
```
zhiltsov-max: Truncated description.

cvat-sdk/examples/cloud_storage_register.py
```diff
@@ -0,0 +1,102 @@
...
+"""
+
+import argparse
+
+from cvat_sdk import make_client, models
+from cvat_sdk.core.helpers import get_paginated_collection
+
+
+def parse_args() -> argparse.Namespace:
+    parser = argparse.ArgumentParser(description=__doc__.splitlines()[0])
```
zhiltsov-max: Truncated description.

cvat-sdk/examples/auth_cli.py
```diff
@@ -0,0 +1,50 @@
...
+import argparse
+
+from cvat_sdk import make_client_from_cli
+from cvat_sdk.core.auth import configure_client_auth_arguments
+
+
+def parse_args() -> argparse.Namespace:
+    parser = argparse.ArgumentParser(description=__doc__.splitlines()[0])
```
zhiltsov-max: Truncated description.

Pr1ncee: I'll fix it in every script

cvat-sdk/examples/project_export_dataset.py
```diff
@@ -0,0 +1,85 @@
+# Copyright (C) CVAT.ai Corporation
+#
+# SPDX-License-Identifier: MIT
+
+"""Export an existing project's dataset (all tasks) to a local zip AND to a
```
zhiltsov-max: This example is basically the same as `task_inspect_and_export.py` and it does what's already available in cvat-cli project (except that adds cloud storage export). What would be useful is making a script to export a set of tasks within the project individually or just by a list of ids (this should also become a CLI command at some point). Consider replacing one or both of the scripts with this.

cvat-sdk/examples/task_inspect_and_export.py
```diff
@@ -0,0 +1,91 @@
...
+        # Low-level API: there is no high-level proxy for the format list yet.
+        formats, _ = client.api_client.server_api.retrieve_annotation_formats()
+        names = [f.name for f in formats.exporters]
+        if args.export_format not in names:
+            sys.exit(
+                f"Unknown export format {args.export_format!r}. Choose one of: {', '.join(names)}"
+            )
+
+        # 3. Export to a local zip
+        local_path = Path(f"task_{task.id}_dataset.zip")
```
zhiltsov-max, line 87: I think it would be nice to change the repeated annotation export example to an analytics export instead. You could export the logs and show some statistics - e.g. the number of people assigned or job reworks made.

cvat-sdk/examples/tasks_bulk_from_cloud.py
```diff
@@ -0,0 +1,116 @@
+# Copyright (C) CVAT.ai Corporation
+#
+# SPDX-License-Identifier: MIT
+
+"""Bulk-create tasks inside a project, each task's data read from a registered
+cloud storage. Nothing is uploaded from your machine.
```
zhiltsov-max: > Nothing is uploaded from your machine.

Do we really need to say this? It's repeated in several examples.

cvat-sdk/examples/tasks_bulk_from_cloud.py
```diff
@@ -0,0 +1,116 @@
...
+  python tasks_bulk_from_cloud.py --host 'https://app.cvat.ai' --token '<your token>' \\
+      --cloud-storage-id 7 --project-id 42 \\
+      --task 'videos/clip_01.mp4' --task 'videos/clip_02.mp4' --task 'videos/clip_03.mp4'
+
+  # two image-batch tasks in project 42
+  python tasks_bulk_from_cloud.py --host 'https://app.cvat.ai' --token '<your token>' \\
+      --cloud-storage-id 7 --project-id 42 \\
+      --task 'batch_a/img_1.jpg,batch_a/img_2.jpg' \\
+      --task 'batch_b/img_1.jpg,batch_b/img_2.jpg'
+"""
```
zhiltsov-max, line 45: I think this command could be a good place to demonstrate support of wildcards (`data_params = {"filename_pattern": ...}`).

cvat-sdk/examples/tasks_bulk_from_cloud.py
Pr1ncee: Why is it bad? In total it's said in two examples related which both related to cloud storage

cvat-sdk/examples/cloud_storage_register.py
```diff
@@ -0,0 +1,102 @@
...
+        print(f"Storage {fetched.id}: {fetched.display_name!r} ({fetched.provider_type})")
+
+        # 4. Update the display name (PATCH — only the passed fields change)
+        updated, _ = api.partial_update(
+            storage.id,
+            patched_cloud_storage_write_request=models.PatchedCloudStorageWriteRequest(
+                display_name=f"{args.bucket} (updated)"
+            ),
+        )
+        print(f"Renamed storage {updated.id} to {updated.display_name!r}")
```
zhiltsov-max, line 118: Consider extending the example with listing the storage contents as well, https://github.com/cvat-ai/cvat/blob/c6a827cb3cc2a47d36c2c75197f2266ea5202d79/tests/python/rest_api/test_cloud_storages.py#L430

cvat-sdk/examples/tasks_bulk_from_cloud.py
zhiltsov-max: In this paragraph, it's already said in the previous sentence:

> each task's data read from a registered cloud storage

so it's unlikely that you'd otherwise think that the data could be taken from the local machine. It also changes the reader's focus from the purpose of the example, which is that it automates manual actions for creating of N tasks in a row.

cvat-sdk/examples/project_export_dataset.py
Pr1ncee: didin't know it, thanks

cvat-sdk/examples/job_assign.py
Pr1ncee: Updated

cvat-sdk/examples/job_list.py
Pr1ncee: Yes, got it, updated and remove `project_status_report.py`

cvat-sdk/examples/project_export_dataset.py
Pr1ncee: Updated in all places occurred

cvat-sdk/examples/auth_cli.py
Pr1ncee: adjusted, used `__doc__.split("\n\n")[0]` in every script not to manually type index/slice as it was before

cvat-sdk/examples/cloud_storage_register.py
Pr1ncee: Added, used `retrieve_content_v2` together with `next_token` to fetch contents. I also update the correspoding documentation

site/content/en/docs/api_sdk/sdk/examples/_index.md
```diff
@@ -0,0 +1,36 @@
...
+- Recipes that create resources keep them and print their ids and UI links. Pass
+  `--cleanup` to delete what the script created (never the sources it read).
+- Recipes that inspect or export take an **existing** resource id
+  (`--project-id`, `--task-id`), so they work directly against your data.
+- List-valued options accept multiple values, e.g. `--labels car person`.
+- Missing arguments exit with a friendly message; SDK errors surface as normal
+  Python tracebacks.
+
+All recipes are executed end-to-end in CI (`tests/python/sdk/test_examples.py`), so
+they stay in sync with the SDK.
```
bsekachev: I do not think this is meaningful information for a user that we test them on CI. 
But it would be meaningful just say that all examples are tested in the latest SDK version.

bsekachev (commented):
I would propose to drop a line to Usage section
Here:

Image

/docs/api_sdk/sdk/#usage

bsekachev:
Currently we duplicate python code in markdown files.
It actually does not prove these examples to be sync and working. 
Somebody may easily fix example, but forget about documentation.

I would propose consider inserting python code immediately via hugo.

Add a shortcode file to `site/layouts/shortcodes`

E.g.:

```
{{ $path := .Get 0 }}
{{ $code := os.ReadFile $path }}
{{ highlight $code "python" "" }}
```

Then in markdown file use it, instead of 



```
```python
 .... python code
...

```

write

```

### The script

{{< include-code "sdk_examples/auth_token.py" >}}
```

The problem that hugo only sees files below its working directory (project root)
It cannot see files in cvat/cvat-sdk/examples

But we may update `README.md` and `build_docs.py` (used on CI) to copy examples to hugo project directory before documentation building


Thus we will get rid of duplication and will actually make sure that examples are relevant.

site/content/en/docs/api_sdk/sdk/examples/_index.md
Pr1ncee: I dont mind rephrasing it

Pr1ncee:
> Currently we duplicate python code in markdown files. It actually does not prove these examples to be sync and working. Somebody may easily fix example, but forget about documentation.
> 
> I would propose consider inserting python code immediately via hugo.
> 
> Add a shortcode file to `site/layouts/shortcodes`
> 
> E.g.:
> 
> ```
> {{ $path := .Get 0 }}
> {{ $code := os.ReadFile $path }}
> {{ highlight $code "python" "" }}
> ```
> 
> Then in markdown file use it, instead of
> 
> ```
> ```python
>  .... python code
> ...
> ```
> 
> write
> 
> ```
> 
> ### The script
> 
> {{< include-code "sdk_examples/auth_token.py" >}}
> ```
> 
> The problem that hugo only sees files below its working directory (project root) It cannot see files in cvat/cvat-sdk/examples
> 
> But we may update `README.md` and `build_docs.py` (used on CI) to copy examples to hugo project directory before documentation building
> 
> Thus we will get rid of duplication and will actually make sure that examples are relevant.

@bsekachev I had similar idea to create some script that will detect differences between the code and the documentation. But I didn't like the prototype created, I'll use hugo instead I guess

site/build_docs.py
```diff
@@ -5,17 +5,15 @@
 #
 # SPDX-License-Identifier: MIT
 
+import argparse
```
bsekachev: Please update README.md file (section resposnible for hugo dev server), mention that we need to link or copy sdk examples before `hugo server`

site/build_docs.py
```diff
@@ -157,8 +190,30 @@ def validate_env():
...
+        ),
+    )
+    return parser.parse_args(argv)
+
+
 if __name__ == "__main__":
+    args = parse_args()
     repo_root = Path(__file__).resolve().parents[1]
+
+    if args.copy_sdk_examples_only:
```
bsekachev: How do you use this option?

site/build_docs.py
```diff
@@ -84,19 +111,24 @@ def git_checkout(ref: str, temp_repo: git.Repo, temp_dir: Path):
...
     data = toml.load(filename)
     data["params"]["version_menu"] = version
 
     with open(filename, "w") as f:
         toml.dump(data, f)
 
 
-def generate_docs(repo: git.Repo, output_dir: os.PathLike, tags):
+def generate_docs(repo, output_dir: os.PathLike, tags):
+    import git
```
bsekachev: Why did you move these imports?

site/build_docs.py
Pr1ncee: I've added it for local development: to be able to automatically sync examples with the documentation without running the whole full build of `build_docs.py`. The reason I wanted to add - see how the documentation looks like locally, without deploying anything

site/build_docs.py
Pr1ncee: I wanted to make lazy imports since I introduced a new functionality that doesn't use them

site/build_docs.py
Pr1ncee: I've updated it

site/build_docs.py
bsekachev: This script is intended for CI. For local development you can just do `ln -s ../cvat-sdk/examples assets`

site/build_docs.py
bsekachev: As I wrote, do not think it is a good idea to use this script just to copy assets once during local development

site/build_docs.py
Pr1ncee: okay, it's deleted now

bsekachev (approved):
lgtm

PR #11055, opened 2026-08-17
e2e `audio_workspace`: use generic annotation saving logic

archibald1418:
This fixes audio_workspace test flake related to pressing the Save button
ex failure: https://github.com/cvat-ai/app.cvat.ai/actions/runs/31975396306/job/95235253836

Old command relied on text content change when pressing the button, which is not reliable and leads to races like this:

- `cy.click` 'Save'
- `'Saving...'` disappears too quickly
- 'Save' text is back
- `cy.click` not finding `'Saving...'`

actual failure:
image
image

expected behavior:
image
image

I believe this is the exact reason our `cy.saveJob` is the way it is, it already
- knows which component to click (`.cvat-annotation-header-save-button`)
- waits for `/jobs/**/annotations PATCH` request and checks it terminated correctly

There was no need in a special audio_workspace-related command in the first place


tests/cypress/support/utils.cy.js
```diff
@@ -186,3 +186,15 @@ export function imageRotate(direction = 'anticlockwise', deg) {
     }
     cy.get('body').click();
 }
+
+export function removeAudioAnnotations(save = false) {
```
lexiv0re: Do we need both this utility and `cy.removeAnnotations();`/`cy.clearAnnotationsAndSave();`? If yes, what's the idea about when to use which one?

klakhov (commented):
I don’t think this extraction improves reuse: `removeAudioAnnotations()` is only an implementation detail of the existing audio commands, while `clearAnnotationsAndSave()` has one call site. 
Could we keep the audio cleanup logic local and use cy.removeAnnotations() + cy.saveJob('PUT') where needed? 
Also, removeAudioAnnotations({ save: true }) doesn’t match its boolean save parameter. I would vote for simplification of this code.

tests/cypress/support/utils.cy.js
archibald1418: Two different testing scenarios: delete annotation or leave them, refactored into two different cypress commands, that re-use this functinon

archibald1418:
@klakhov I would argue that in this case it's necessary because these are two completely different testing scenarios, and can be abstracted away using two different cypress commands with different names. Its much easier to read, the command's goal is obvious, without having to look into the definition

The invocation from a spec would be simplier as well, since both calls require no args

tests/cypress/support/utils.cy.js
lexiv0re: So we need audio-specific wrappers for these scenarios after all?
I see in `case_audio_29...` you use generic `cy.clearAnnotationsAndSave();` which is not audio specific - why then?
I think we need to be consistent. Either use generic helpers to clear/clearAndSave everywhere, or use audio-specific helpers, just don't mix them.

archibald1418: `cy.removeAnnotations` is a step in the other command, sometimes needed alone for more control in the spec. We could rename the command to `cy.removeAnnotationsAndSave`

lexiv0re: I think we got confused already.
There are 2 scenarios:
1. Clear all annotations without saving
Here now we have the following `cy` commands:
* `cy.removeAnnotations()` - generic
* `cy.audioClearAnnotations()` - audio specific
2. Clear all annotations and save the clean state
* `cy.clearAnnotationsAndSave()` - generic
* `cy.audioClearAnnotationsAndSave()` - audio-specific

Also, there's this **public** utility now `removeAudioAnnotations(save: boolean)` which can again be found by somebody and used instead of `cy` helpers.

So the questions are:
1. Do we really need audio-specific equivalents or can we use generic helpers to clear audio intervals?
2. If we need audio-specific helpers in audio specs - why do you use this generic helper in audio test https://github.com/cvat-ai/cvat/pull/11055/changes#diff-42424b3ade03fe1a7a13e099f2adc6ee7aefac801afdad7fd5eaf4416fcda75aR541 instead of audio-specific helper?
3. Do we need this utility that only helps to deduplicate the implementation of audio-specific helpers to be public and available for test writers? Feels it should be private to the place where `cy` helpers are defined.

archibald1418: > I think we need to be consistent. Either use generic helpers to clear/clearAndSave everywhere, or use audio-specific helpers, just don't mix them.

Let's use audio specific ones yes

archibald1418: > So the questions are:

>Do we really need audio-specific equivalents or can we use generic helpers to clear audio intervals?
>If we need audio-specific helpers in audio specs - why do you use this generic helper in audio test https://github.com/cvat-ai/cvat/pull/11055/>changes#diff-42424b3ade03fe1a7a13e099f2adc6ee7aefac801afdad7fd5eaf4416fcda75aR541 instead of audio-specific helper?
>Do we need this utility that only helps to deduplicate the implementation of audio-specific helpers to be public and available for test writers? Feels it should be private to the place where cy helpers are defined.

1. We  do need audio-specific helpers because they interact with a different element `.cvat-audio-regions-list-wrapper` and, more generally, will never be used in non-audio specs
2. Fixed to use audio-specific helper
3. Moved the removeAudioAnnotations declaration into audio commands, without export

tests/cypress/support/commands_audio.js
```diff
@@ -17,6 +17,18 @@ import { defaultTaskSpec } from './default-specs';
 
 const WAVEFORM_TIMEOUT = 30000;
 
+function removeAudioAnnotations({ save = false } = {}) {
+    cy.get('.cvat-audio-regions-list-wrapper').then(($list) => {
```
lexiv0re, line 21: Might give wrong result if `Labels` tab is open in the right-hand panel instead of `Objects` when it's called.

archibald1418: image

This elem `.cvat-audio-regions-list-wrapper` is not visible there when I switch to Labels.
I can add `.should('be.visible')`.

lexiv0re: If you add `.should('be.visibile')` condition to wait it may never fulfil. 
What need to be verified is that even if `Labels` tab is open is still can find `cvat-audio-region-item` despite it being hidden and correctly detect whether there are existing intervals that need to be cleared.

archibald1418: Right, reverted visibility checks
I believe, no other changes should be made: expression `(...).find('.cvat-audio-region-item').length)` checks length despite visibility 

A small demo shows desired behaviour

https://github.com/user-attachments/assets/f0082fb1-0369-4b19-a753-71e4f5abc477

Spec source:
```js
/// <reference types="cypress" />

import { taskName, firstLabelName } from '../../support/const_audio';

before(() => {
    cy.prepareUserSession();
    cy.openAudioJob(taskName);
});

afterEach(() => {
    cy.wait(1000);
});

it("removes audio annos and doesn't if no annos", () => {
    cy.audioCreateRegionViaButton(firstLabelName, 100, 250);
    cy.get('.cvat-audio-region-item').should('have.length', 1);
    cy.audioClearAnnotationsAndSave();
    cy.audioClearAnnotations();

    cy.audioCreateRegionViaButton(firstLabelName, 100, 250);
    cy.get('.cvat-audio-region-item').should('have.length', 1);
    cy.get('.ant-tabs-tab[data-node-key=\'labels\']').click();
    cy.audioClearAnnotationsAndSave();
    cy.audioClearAnnotations();

    cy.get('.ant-tabs-tab[data-node-key=\'objects\']').click();
});
```

lexiv0re (approved):

PR #11076, opened 2026-08-24
Audio workspace: add new interval-related functions

lexiv0re:
### Motivation and context

Working with intervals currently provides not many options around intervals specifically, some existing functions are not very explicit.

These changes revisit the layout of the right-hand list item and selected interval details header to introduce the following new functions:
1. **Set caret to interval start/end**. It sets the playback position to the interval's start/end correspondingly and scrolls the viewport to it. Available as quick actions. Default shortcuts: **[** and **]** respectively.
2. **Fit interval** - helps to quickly focus on the interval by adjusting the zoom level and scrolling the waveform so the selected interval is fully displayed in the center of the viewport with some margins around. Available as a menu action. Default shortcut: **I**.
AfterFit

3. **Play interval once** - besides double click on the waveform interval/right-hand panel interval item now explicitly available as a quick action button in the active interval details header and the right-hand panel interval item. Available as a quick action. Default shortcut: **\\**.

Layout before:
IntervalActionsBefore

Layout after:
NewIntervalActions

P.S. also adds support for native horizontal scrolling of the waveform with touch mouse/touchpads.

### How has this been tested?

Tested manually.
Added e2e tests for the new functions.

### Checklist

### License


cvat-ui/src/audio/components/annotation-page/audio-workspace/audio-interval-play-icon.tsx
```diff
@@ -0,0 +1,16 @@
...
+import React from 'react';
+import classNames from 'classnames';
+
+/** Distinguishes creating an interval playback range from ordinary playback. */
+export default function AudioIntervalPlayIcon(props: React.SVGProps<SVGSVGElement>): JSX.Element {
+    return (
+        <svg {...props} className={classNames('cvat-audio-interval-play-icon', props.className)} viewBox='0 0 24 16' fill='none' aria-hidden>
+            <path d='M4 2V14M4 2H8M4 14H8M20 2V14M20 2H16M20 14H16' stroke='currentColor' strokeWidth='1.7' strokeLinecap='round' />
+            <path d='M9 4.2L16.5 8L9 11.8V4.2Z' fill='currentColor' />
+        </svg>
```
bsekachev: For icons we have `cvat-ui/assets/*.svg` path

bsekachev:
Visually this group of button seems a little bit messy

image

- Three buttons are responsible for navigation
- Three buttons to switch properties
- One button to open menu

I think we need somehow split them (at least by margin), or on different lines as for 2D workspaces.

bsekachev:
I understand that tooltip is about "]"
But I believe it also will confuse users, because it seems just like a typos.

And why did you choose these shortcuts?
IMO these are not really convenient. 

image

bsekachev:
Not related to this PR, but I think label select should be like in 2D workspace. It supports text input there (very useful when multiple labels are in the task)

image
image

Just write to your TODO list :)

lexiv0re:
@bsekachev regarding **[** and **]** shortcuts I was inspired by two things:
1. Audacity uses **[** / **]** to mark selection start/end and **Ctrl+[** / **Ctrl+]** to scroll selection start/end into view, basically functions that work with interval boundaries in some way.
2. **[** and **]** visually designate interval boundaries so it felt they give somewhat correct association

Some tools use **\\** to start playback of a certain part/item and it's close to **[** and **]** on the keyboard so it felt like an okay addition to form the interval playback/navigation shortcuts group.

But I agree that there are cons: 
1. Tooltips end up being confusing
2. They are not convenient to be used by left hand while having mouse in the right hand as they are naturally in the right hand keyboard area

If you have a better suggestion about what set of default shortcuts to use for these three functions I'd be more than happy to consider it :)

cvat-ui/src/audio/components/annotation-page/audio-workspace/audio-interval-play-icon.tsx
lexiv0re: Extracted as .svg icons.

lexiv0re:
Added margin between quick action groups:
IntervalActionsMargin

bsekachev:
Maybe a key with modificator? 

E.g. Shift + Space to play/pause. 
Shift + D to navigate to beginning (usually we use D to navigate back)
Shift + F to navigate to the end (usually we use F to navigate forward)

lexiv0re:
Actually, `Shift+Space` was my second choice for the interval playback. Combining it with `Shift+D/F` may work, although `D/F` are used for a short step backward/forward so not exactly equivalent. Better analogy might have been `Jump to start/end` of the track (and in our case with shift modifier - start/end of interval) but there are no shortcuts for these actions at all currently. But yes, I think I like it better, will update to these defaults.

lexiv0re:
Updated to use `Shift + Space`, `Shift + D` and `Shift + F`

bsekachev (approved):
Added icons shall have external width and height '1em` for consistency.

lexiv0re:
Thanks, updated.

PR #11077, opened 2026-08-24
Fix raw validation for number attributes

### LINKED ISSUE #11034: Incorrect validation for numerical attributes in the raw label editor

auster-vn:
### Motivation and context

Fixes #11034.

The raw label editor validated all attribute values as unique.
For number attributes, `values` represents `[min, max, step]`, so values such as `["1", "4", "1"]` are valid even though `min` and `step` are equal.

This change skips the uniqueness check for number attributes while preserving it for other input types.

### How has this been tested?

- Verified a number attribute with min=1, max=4, step=1 can be pasted in the raw label editor without validation errors.
- Verified duplicate values for select/radio attributes are still rejected.
- Ran CVAT UI lint/build.

### DIFF
--- a/cvat-ui/src/components/labels-editor/common.ts
+++ b/cvat-ui/src/components/labels-editor/common.ts
@@ -54,7 +54,10 @@ function validateParsedAttribute(attr: SerializedAttribute): void {
     }
 
     const attrValues = attr.values.map((value: string) => value.trim());
-    if (new Set(attrValues).size !== attrValues.length) {
+    if (
+        attr.input_type.toLowerCase() !== 'number' &&
+        new Set(attrValues).size !== attrValues.length
+    ) {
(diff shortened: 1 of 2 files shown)


bsekachev:
Hello, 

Thank you, but we already have related contribution in this PR #11047

PR #11083, opened 2026-08-26
Prevent corrupted chunks during resumed downloads

bsekachev:
### Motivation and context

The chunk download worker preserves data received before a connection interruption. It stores successfully read byte arrays and tracks their total size. After a stream error, it requests the remaining data using:

```http
Range: bytes=<receivedBytes>-
```

The worker then appends the resumed response to the preserved prefix, avoiding a complete re-download of a potentially large chunk.

This behavior had two integrity gaps:

- The initial response did not provide a reliable expected size. If it ended cleanly but prematurely, the worker could accept an incomplete chunk.
- A chunk could be regenerated between requests. The worker could then combine a prefix from the old chunk with a suffix from the new chunk, producing a corrupted archive.

The worker now sends an open-ended range on every attempt, including the initial request:

```http
Range: bytes=0-
```

CVAT returns `206 Partial Content` and includes the complete decoded chunk size in `Content-Range`. The same response validation is therefore used for initial and resumed requests, independently of HTTP content encoding.

The worker also preserves the chunk checksum and update version from the initial response. Before appending resumed data, it verifies that both values remain unchanged. If the chunk has changed, the worker cancels the response, discards the preserved data, and immediately restarts from `bytes=0-`.

The required range and identity headers are also allowed and exposed through CORS.

### How has this been tested?

- Manually

### Checklist

### License

### DIFF
--- a/tests/cypress/e2e/actions_tasks/chunk_download_retry.js
+++ b/tests/cypress/e2e/actions_tasks/chunk_download_retry.js
@@ -17,6 +17,11 @@ context('Retry chunk downloads on annotation page', () => {
         cy.visit('/auth/login');
         cy.login();
         cy.url().should('contain', '/tasks');
+    });
+
+    beforeEach(() => {
+        taskId = null;
+        jobId = null;
         cy.headlessCreateTask({
             labels: [{ name: labelName, attributes: [], type: 'any' }],
             name: taskName,
@@ -35,7 +40,7 @@ context('Retry chunk downloads on annotation page', () => {
         });
     });
 
-    after(() => {
+    afterEach(() => {
         if (taskId) {
             cy.headlessDeleteTask(taskId);
         }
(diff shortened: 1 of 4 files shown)


tests/cypress/e2e/actions_tasks/chunk_download_retry.js
```diff
@@ -77,4 +83,97 @@ context('Retry chunk downloads on annotation page', () => {
...
+            pathname: `/api/jobs/${jobId}/data`,
+            query: {
+                type: 'chunk',
+            },
+        }, (req) => {
+            chunkRequests++;
+
+            if (chunkRequests === 1) {
+                expect(req.headers.range).to.equal('bytes=0-');
+                req.continue((res) => {
```
klakhov, line 102:
```
req.continue((res) => {
    const responseSize = res.body.byteLength ?? res.body.length;
    truncatedSize = Math.floor(responseSize / 2);

    res.send(
        res.statusCode,
        res.body.slice(0, truncatedSize),
        {
            ...res.headers,
            'content-length': `${truncatedSize}`,
        },
    );
});
```

Will something like this work? It will prevent us from using eslint supression

bsekachev: applied

PR #11084, opened 2026-08-26
Add multi-object canvas selection

### LINKED ISSUE #2340: Selecting more than one object - Moving/Copying/Pasting or Assigning Label  - Enhancement
I saw a related issue, but not completely similar.
[Issue 1115](https://github.com/openvinotoolkit/cvat/issues/1115)

### Expected Behaviour  - Enhancement 
Being able to select more than one object for copying, pasting, moving, or changing labels
### Current Behaviour
At the moment, it is possible to select only one object for moving and copy and pasting.

### Possible Solution
by explicit selecting while holding a bottom or using a Free/Rectangle Select tool similar to Gimp in which any objects which falls into the selected region completely is selected.

I am using the origin/develop branch which was pulled and built on 21st October 2020.

senketsukamui:
### Motivation and context

This PR introduces multi-object selection and batch operations in the 2D annotation workspace.

It addresses one of the most requested CVAT features:

Closes #2340

Users can now:

- Select multiple objects using a selection bounding box.
- Add or remove individual objects with `Ctrl + click`.
- Select an object range in the Objects sidebar with `Shift + click`.
- Select all visible objects with `Ctrl + A`.
- Move selected movable objects together using the selection bounding box.
- Keep locked and pinned objects selected but stationary during movement.
- Change a common applicable label for all selected objects.
- Copy, paste, and delete a selection.
- Lock, unlock, pin, or unpin selected objects.
- Use label shortcuts with the active selection.
- Restore selection changes through Undo and Redo.
- Restore a selection after navigating away from its frame and returning.
- Clear the selection by clicking outside it, pressing Escape, or changing frames.

The selection is represented by a visible bounding box and a `SELECTION (N)` label. Its context menu exposes batch actions and disables operations that cannot be applied to every selected object.

Batch updates, deletion, creation, and movement are recorded as single history actions where applicable. Ground-truth and unsupported objects are validated before a batch operation begins.

The implementation also adds immutable selection clipboard snapshots, transformed bounding-box checks for rotated objects, and shared helpers for selection state and batch lock/pin behavior.

### How has this been tested?

The following checks were run locally:

- TypeScript checks for `cvat-core`, `cvat-canvas`, and `cvat-ui`.
- ESLint for all changed TypeScript and React files.
- `git diff --check`.
- Existing label shortcut E2E scenarios were extended for multi-selection.

The following workflows were also tested manually in the 2D annotation workspace:

- Bounding-box selection and individual `Ctrl + click` selection.
- `Shift + click` range selection in the Objects sidebar.
- Selecting all visible objects with `Ctrl + A`.
- Adding and removing objects from an existing selection.
- Moving selections containing movable, locked, and pinned objects.
- Batch copy, paste, deletion, label change, lock, unlock, pin, and unpin.
- Undo and Redo for selection changes and batch operations.
- Restoring a selection after frame navigation.
- Selection behavior for rectangles, polygons, polylines, ellipses, masks, tracks, and tags.

### Checklist

  https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue-using-a-keyword))

### License


senketsukamui:
Proposed testing scenarios:

Creating and refining a selection
- Shift-drag creates a selection only from objects fully contained by the bbox.
- Partially intersecting objects are excluded.
- Rotated objects use their transformed bounds.
- Hidden, filtered, outside, hidden-layer, and tag objects are excluded.
- A one-object selection still shows SELECTION (1) and its bbox.
- An empty bbox clears the existing selection.
- Command-click adds and removes individual objects on canvas and in the Objects tab.
- Command-click works through the selection bbox.
- Command combined with Shift or Alt does not trigger object selection.
- A Points object is selected and moved as a complete object; individual points are not added separately.

Selection mode
- Clicking the Select tool enables persistent bbox selection.
- Shift-drag from cursor mode activates the Select tool.
- Releasing Shift does not leave selection mode.
- The toolbar button is highlighted while selection mode is active, not merely because a selection exists.
- Clicking the highlighted Select tool returns to cursor mode.

Clearing and replacing selection
- Clicking outside the bbox clears the selection.
- Escape clears the selection.
- Changing frame clears selected frame-local shapes.
- Selected tracks remain selected when changing frames if they are available and visible on the destination frame.
- Outside, hidden, filtered, and hidden-layer tracks are removed from selection on the destination frame.
- A preserved track selection rebuilds its bbox from geometry on the destination frame.
- Clicking an outside object clears the selection and activates that object.
- Hovering outside objects does not activate them while selection is active.
- Hovering Objects-tab entries does not activate them while selection is active.
- Clicking an Objects-tab entry without a modifier clears selection and activates it.

Selection presentation
- Selected shapes retain their label colors.
- Selected objects use increased selected opacity.
- The combined bbox is black and includes padding around the objects.
- Masks have a visible selection bbox that is recalculated after movement.
- Points, polylines, polygons, ellipses, skeletons, and masks have an unambiguous selected state.
- Point numbering and polyline direction indicators remain visible without enabling individual point editing.
- SELECTION (N) updates as membership changes.
- SELECTION (N) keeps a constant screen-space size while zooming.
- The actions menu opens using both the ellipsis and right-click.
- A menu opened from the ellipsis remains attached to it while zooming.
- Menu actions, label selection, layer selection, and Details remain interactive while the menu is attached.
- Starting a selection drag closes the actions menu.

Moving selected objects
- Dragging the combined bbox moves all movable members by the same delta.
- Dragging the SELECTION (N) header moves the selection.
- Dragging a selected movable object moves the remaining movable members.
- Relative geometry and z-order remain unchanged.
- Selection remains active after movement.
- Locked, pinned, and ground-truth members remain selected but stationary.
- A selection containing only stationary members cannot be moved and does not show a drag cursor.
- Mixed selections move only movable members without removing stationary members from selection.
- Masks move using integer translation and retain valid RLE data.
- A completely out-of-frame selected mask follows the same core behavior as a single mask.
- Undo reverts the complete movement in one action; redo reapplies it.

Deleting a selection
- Delete and Backspace remove an unlocked selection.
- A locked member prevents regular deletion atomically.
- Shift+Delete and Shift+Backspace force-delete locked members.
- Ground-truth members prevent the complete deletion.
- Undo restores all deleted objects and their selection.
- Redo deletes all objects again and clears the selection.
- Batch deletion does not leave an active or nested history transaction.

Copying and pasting
- Ctrl+C copies the complete selection.
- Ctrl+V starts interactive placement with transparent ghost objects and black outlines.
- Clicking places every copied object while preserving relative geometry.
- Repeated placement works without blocking later canvas interaction.
- Relative positions, labels, attributes, descriptions, object types, rotations, groups, lock state, pin state, and instance colors are preserved.
- Pasted shapes use the current z-layer.
- Tracks remain tracks.
- Pasted objects become the active selection.
- Changing source objects after copying does not mutate clipboard data.
- Undo and redo treat selection paste as one history action.
- Single-object Make a copy retains its established placement and activation flow.
- Selection copying does not change 3D cuboid copy or audio behavior.

Batch properties
- Context-menu lock applies to every selected object; the next action unlocks all.
- Context-menu pin applies to every compatible object; the next action unpins all.
- Mixed lock/pin state follows the same aggregate behavior and icon state as the Objects sidebar header.
- S L toggles lock/unlock for the selection.
- S P toggles pin/unpin for the selection.
- Pin is rejected for selections containing locked, ground-truth, tag, or skeleton-element objects.
- Standalone Points objects can be pinned; pinned Points cannot be moved as part of a selection.
- Failed batch operations leave every member unchanged.

Batch label changes
- The menu shows the current label when all members share it.
- Mixed selections display Multiple labels.
- Only labels applicable to every member are available.
- No common label disables the selector.
- Locked, ground-truth, and skeleton members disable label changes.
- Ctrl+number changes the label for a compatible selection.
- An incompatible shortcut displays a warning and changes nothing.

Batch attribute changes
- Details uses the same controls and layout as a single-object attribute editor.
- Only attributes applicable to every selected object are shown.
- A shared value is displayed when all selected objects have it.
- Mixed values are indicated without overwriting them until the user chooses a value.
- One attribute change updates every compatible selected object in one batch action.
- Locked, ground-truth, mixed-label, and unsupported selections disable attribute editing with an explanation.

Grouping and ungrouping
- Group selection groups all selected compatible objects in one operation.
- Group selection is disabled for fewer than two objects or objects already in one group.
- Ungroup selection removes group membership from every grouped selected object.
- Selection remains active after grouping and ungrouping.
- Undo and redo treat each group or ungroup command as one action.

Annotation and layer actions
- Run annotation action opens with the selected objects as its explicit target set.
- An annotation action receives every selected compatible object and does not process unselected objects.
- To background and To foreground update all movable selected objects together.
- Move one layer backward and forward preserve relative z-order.
- Move to layer applies the chosen layer to all movable selected objects.
- Layer actions are disabled when no selected member can be updated.
- Layer changes remain one undoable batch action.

Selection history
- Undo and redo restore each add/remove selection change.
- Undo restores a selection after clicking outside or pressing Escape.
- A new selection after Undo clears the redo stack.
- Select objects on frame A, move to frame B, return to A, then Undo restores the frame-A selection.
- Track-only selections persist across frame navigation without adding unnecessary selection-history entries.
- Mixed shape/track selections preserve applicable tracks and record only the removed membership.
- Undo/redo frame navigation does not create extra selection-history entries.
- Move, delete, paste, lock, pin, label, attribute, grouping, and layer updates each use one history action.

Objects, Labels, and Layer sidebar views
- Shift-click selects a continuous object range using the visible sort order.
- Range selection works with ID, label, updated-time, and layer ordering.
- Collapsed layer members are excluded from the Shift range.
- Command+A selects all visible shapes and tracks and prevents native page text selection.
- Command+A excludes tags, filtered, hidden, outside, and hidden-layer objects.
- Command+A retains normal text-selection behavior in inputs, textareas, selectors, and editable fields.
- Command-clicking a label selects or removes all selectable objects with that label.
- A label receives a selection outline when all its selectable objects are selected.
- Command-clicking a layer header selects or removes all selectable objects on that layer.
- Command-clicking individual draggable object cards works when Sort by Layer is active and does not start layer dragging.
- A layer receives a selection outline when all its selectable objects are selected.
- Moving layers, moving objects between layers, compacting layers, or hiding layers resets selection when required.

Regression and compatibility checks
- Color By Instance assigns generated instance colors to newly created objects instead of black.
- Selection copies preserve explicit instance colors without changing colors for ordinary object creation.
- Ctrl/right-click on a point still opens the Delete point menu.
- Existing single-object hover, activation, context menu, movement, copy, simplify, and delete flows remain unchanged when no selection is active.
- Existing 3D cuboid Make a copy scenarios continue to pass.
- Audio annotation selection and history behavior remain unchanged.

julled:
Great to see another multi object selection  implementation!

I have another one here, maybe that can also give some inspirations?

https://github.com/cvat-ai/cvat/pull/10937

cvat-canvas/src/typescript/objectSelector.ts
lexiv0re, line 1: 1. Click `Select objects` control on the left panel
2. Ctrl+select any annotation
3. Click `Undo`
The selection is removed but `Redo` is not appears to be available.
Similar issue is with shift+selection in the `Select objects` mode

cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/select-control.tsx
lexiv0re, line 1: Confusing behavior of the new mode.
Sometimes it is selected on its own. Sometimes both cursor and selection are enabled.
E.g. in `Cursor` mode `Ctrl+click` an object:

Image

lexiv0re, line 1: `Shift+draw a selection box` switches to `Select objects` mode even if no objects were actually selected.

lexiv0re, line 1: 1. Click `Select objects` mode
2. Click any object without any modifier -> multi selection group is created with this element
3. Click any other object without any modifier -> group selection is dropped and mode switches to `Cursor`

At the same time drawing a selection box in this mode without modifiers work as expected.

cvat-canvas/src/typescript/objectSelector.ts
lexiv0re, line 1: 1. Select multiple objects
2. When hovering over the selection with `Shift` or `Ctrl` modifiers pressed `Drag` cursor is displayed
If you try to drag - another action is performed not selection drag which is misleading.

lexiv0re, line 1: 1. Draw one rectangle
2. Draw another rectangle that fully encloses the first one
`Ctrl+click` can never target the first one.
Regular activation uses semantic hover resolution and allows to work with both.

lexiv0re, line 1: 1. Select only objects that are not movable
Hovering over the selection still shows `Drag` cursor but it starts bbox selection if you try to move

lexiv0re, line 1: 1. Select multiple items with `Shift+drag`
2. Drag the selection

Selection is dropped after drag is finished.
If the same items are selected with `Ctrl+click` instead and then dragged - the selection remains. Although, feels like not always also.
It should be consistent regardless of how the selection was made. And I think it should keep the selection after drag.

lexiv0re, line 1: Unlocked and unpinned `skeleton` is not moved with the multi-selection while it can be moved in regular mode.

cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/select-control.tsx
senketsukamui: We discussed this behavior with @bsekachev, so user can keep drawing the selection box even if he unpresses the Shift button

lexiv0re: I agree that the selection should keep going if the user releases `Shift` mid-drawing. But it doesn't explain why user ends up in the selection mode if nothing were selected after all. 
Also, the selection mode becomes enabled only when mouse button is released and its not enabled during the selection itself which is also a bit confusing.

cvat-canvas/src/typescript/objectSelector.ts
lexiv0re, line 1: How does one see and work with a selection containing only `Tag` objects?

lexiv0re, line 1: `Points`, `Skeleton` and `Polyline` do not change their look when included in a selection. It's confusing as all other shapes now do change their visual state.

cvat-ui/src/components/annotation-page/standard-workspace/objects-side-bar/objects-list.tsx
lexiv0re, line 1: Multi-selection in the list doesn't work when items are grouped by layers.

lexiv0re, line 1: Now that we have multi-selection it might be convenient to allow multi selection in the `Labels` list along with `Objects` list. `Ctrl+click` adds/removes all objects of the label to/from the selection. `Shift+click` operates with multiple labels in a similar way it operates with objects in the`Objects` list.

Image

lexiv0re, line 1: Selection is dropped when clicking on an action in the list of items. It makes the following scenario not possible:
1. Select multiple objects to move them all
2. Realize that one is locked/pinned
3. Unlock/unpin it
4. Move the whole selection
At step 3 selection is dropped. `Undo` helps to restore the selection but it's not ideal.

cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/select-control.tsx
lexiv0re, line 1: 1. `Ctrl+click` on a shape
2. Click on `...` - nothing happens
The selection menu is only opened on the second click on `...`. Also, it's never closed on subsequent clicks.
Note: if I make selection with `Shift+drag` it opens the selection menu on the first click on `...`

cvat-ui/src/components/annotation-page/canvas/views/canvas2d/canvas-wrapper.tsx
```diff
@@ -1245,6 +1566,171 @@ class CanvasWrapperComponent extends React.PureComponent<Props> {
...
+                            items: [
+                                {
+                                    key: 'change-label',
+                                    label: (
+                                        <div
+                                            className='cvat-canvas-selected-objects-label-selector'
+                                            title={labelSelectorDisabled ? labelSelectorDisabledReason : undefined}
+                                        >
+                                            {applicableLabels.length ? (
+                                                <LabelSelector
```
lexiv0re: After selecting a new label or just opening/closing labels selector the expander icon becomes invisible on hover:
Before:

Image

After:

Image

cvat-ui/src/components/annotation-page/canvas/views/canvas2d/canvas-wrapper.tsx
```diff
@@ -1245,6 +1566,171 @@ class CanvasWrapperComponent extends React.PureComponent<Props> {
                     }}
                 />
 
+                {selectionMenuPosition && (
```
lexiv0re, line 1639: All selection actions but `Copy` drop the selection. I'm not sure it's convenient. E.g. what if you want to change both "Pinned" and "Label" for the same selection?

lexiv0re, line 1639: `Copy` copies the selection but the only way to actually `Paste` is is via the `Ctrl+v` shortcut. It's a bit confusing and not consistent with how `Make a copy` works for one shape.

lexiv0re, line 1639: When all selected items share the same label and some are locked the tooltip about why changing label is disable is only displayed when hovering outside of the selected value.
Hover over the selected value:
Image
Hover over the frame:
Image

It is displayed when labels are different though:
Hover over the selected value:
Image

cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/select-control.tsx
lexiv0re, line 1: 1. Click `Select objects` mode
2. Click any shape so 1-shape multi-selection is displayed
3. Right click on the selection bbox
Selection is dropped instead of showing the selection menu

cvat-canvas/src/typescript/objectSelector.ts
lexiv0re, line 1: It's a bit confusing that it's not possible to drag the selection by its `SELECTION(N)` header only the bbox.

Image

lexiv0re, line 1: When multi-selecting tracks the selection is dropped when going to the next frame.
Regular active track remains active when moving to the next frame.

cvat-ui/src/components/annotation-page/canvas/views/canvas2d/selection-attributes-editor.tsx
lexiv0re, line 1: Not sure why it should be a modal for multi-selection?
Overall, why can't we make the selection controls look the same as the control displayed when a single shape is right clicked instead of the context menu? 
It will be consistent and familiar to the user.

lexiv0re, line 1: 1. Select multiple shapes with the same label
2. Edit attributes, click apply
3. Click **Undo** -> first it brings back the selection
4. Click **Undo** again -> it reverts the attributes change
5. **Redo** is not available at this point for some reason
`3+4` - should be single undo operation
`5` - **Redo** should be availble

cvat-ui/src/containers/annotation-page/standard-workspace/objects-side-bar/objects-list.tsx
```diff
@@ -105,6 +120,12 @@ const componentShortcuts = {
         sequences: ['l'],
         scope: ShortcutScope.OBJECTS_SIDEBAR,
     },
+    SWITCH_SELECTION_LOCK: {
+        name: 'Lock/unlock selected objects',
+        description: 'Change locked state for all selected objects',
+        sequences: ['s l'],
```
lexiv0re: Do we need separate sequences for multi-selection actions? Can we reuse the active component sequences? Like simply `l` here, `p` for pinned etc.

senketsukamui: Good idea actually, I thought to make it separate like `t l` and `t p` for all objects, but I guess using `l` and `p` is better way

senketsukamui:
Also, how do we handle points for selection? I proposed allowing adding points to selection to move them all at once. What do you think about this behavior? I can also make so we can add individual points to the selection, but it's kinda complicated.

lexiv0re:
I think we definitely don't want to add individual points to selection, at least not with the way they are organized now. 
As a user I'd expect the points shape to be movable as a part of the selection. So I'd suggest to add `Pin` property to the `Points` shape. In regular mode it would control whether individual points are movable. In multi-selection it would allow/disallow moving the whole set.

bsekachev:
Hello 

1. Select a group
2. Selection actions button never works from the first time (only from the second one)  [bug]
3. Double click on this button fits the object  [bug]
4. Finally open menu 
5. Zoom, drag image. Menu moves away from "..." [bug]
6. Start dragging objects. Menu stays opened [bug]
7. Select two masks. Lock them via "...". Objects cannot be moved - OK
8. Lock all with T L or with button in side menu - all locked. OK. 
9. Unlock all with T L or with button in side menu - all unlocked. OK
10. Start dragging masks - they can be dragged but after dragging return to initial place [bug] 

UPD: probably latest related with that one of masks in the group slightly crossed the image border. 
1. This is not UX friendly solution.
2. If we drag a single mask or bounding box, production behaviour is broken.  [bug]

Production: mask/bounding box cut when partially moved outside of frame.
This branch: they returned to the initial position when partially moved outside of frame.

So, these are found for 5 minutes of using the feature. 
Please, test your patch meticulously first.

bsekachev:
image

Icons swapped. Black lock icon means object is locked now. But in menu you implemented such icon means unlocked now.

senketsukamui:
Can you explain more? I don't think I get it. You mean that for Lock selection option is context menu we need to add unlocked icon?

bsekachev:
https://github.com/user-attachments/assets/19572342-1939-4422-8b6a-4c28a6de2aa7

One object get selected when switching lock/unlock.

bsekachev:
> Can you explain more? I don't think I get it. You mean that for Lock selection option is context menu we need to add unlocked icon?



image

One object is selected, right? 
The object is not locked. Icon outlined (in sidebar).
But in selection menu icon is filled.

senketsukamui:
That makes sense, but in the sidebar icon is both button and the indicator of lock/unlock state of an object while inside selection context menu it's just a button that makes certain action. I'm not sure if it's a mismatch.

bsekachev:
Снимок экрана — 2026-09-02 в 11 56 21 AM


Similar issue when one object is locked, and one unlocked (both in selection). 
We should keep behaviour like for all lock / unlock button.

If at least one unlocked, common button state is unlocked and switching it will lock objects in the group.

Please, check that logic is consistent for left part of functionality

bsekachev:
> it's just a button that makes certain action. I'm not sure if it's a mismatch

It is not just a button that makes certain actions. This icon is switched after clicking the button. So, it is a button + indicator of current state.

senketsukamui:
> Production: mask/bounding box cut when partially moved outside of frame.
> This branch: they returned to the initial position when partially moved outside of frame.

Is it expected behavior? I fixed it manually because it seemed like a bug to me

bsekachev:
>  Is it expected behavior? I fixed it manually because it seemed like a bug to me

It is expected and very important. Without this, how could you drag a shape to an object on frame edge?

senketsukamui:
> > Is it expected behavior? I fixed it manually because it seemed like a bug to me
> 
> It is expected and very important. Without this, how could you drag a shape to an object on frame edge?

Yeah, I missed this. I will revert this

bsekachev:
Something is wrong when I create new object with "Color By" instance, all created objects have black color. The issue can not be reproduced in develop branch.

Selection does not work in sidebar when sort by "Layer":

image


Have you considered using `command` instead of `ctrl` on MacOS? I believe it would be more consistent if consider also this change: https://github.com/cvat-ai/cvat/pull/11155 and take into account default mac behaviour (ctrl+left click = right click)

bsekachev:
Menu should not work like this when scrolling 

https://github.com/user-attachments/assets/39deddc5-2cc3-4586-b8e3-e39a18a44f6c

bsekachev:
As I remember we discussed that for tracks selection should stay if just moving between frames. Now I see it resets after changing frame (actually we have a real use-case of that).

senketsukamui:
Hello, @bsekachev @lexiv0re! I fixed all the of comments and failed tests, I'm going through all the comments one more to double-check. 
There are few new things I've discovered and added:
1. I copied all layer actions for selection (to foreground, to background etc.)
2. Layer stack menu now supports adding whole layer to selection + layer interval selection using shift
3. Fixed mask out of frame case using @bsekachev PR https://github.com/cvat-ai/cvat/pull/11145
4. Implemented annotation action on selection
I'm adding new test cases and starting writing e2e scenarios. In the meantime, you can rereview this PR.

bsekachev:
> I'm adding new test cases

Please, try to fix all existing tests first

klakhov:
Something is wrong with hide during seleciton. Select two objects. Hide one of them, selection rect jumps. 
Maybe we should exclude hidden objects from selection?
Kooha-2026-09-09-11-44-16

klakhov:
1. Select two objects
2. Copy paste them somewhere
3. Undo
4. Object activation on hover is not working

klakhov:
Escape → Undo behaves differently depending on how objects were un-selected.
Ctrl-click A, Ctrl-click B, press Escape, then Undo. Instead of restoring A+B, it restores A only. 
Ctrl-click A, Ctrl-click B, press Escape, click on empty space, then undo it restores A+B

klakhov:
Ctrl-click cannot deselect a layer containing hidden objects.
Put visible A and hidden B on the same layer. Ctrl-click the layer header: A becomes selected. Ctrl-click again: A stays selected.

klakhov:
Dragging rotated shapes displays incorrect preview:
Kooha-2026-09-09-13-04-54

klakhov:
The persistent Select tool exits before you can draw a second box.
Activate Select, draw a box around some objects, then start another box on empty canvas without holding Shift. On mouse-down, the wrapper immediately disables Select and clears the selection—before it can distinguish a click from a drag. The second selection gesture therefore cannot complete.

cvat-canvas/src/typescript/canvasView.ts
lexiv0re, line 1: Something is wrong with rotation command history:
1. Rotate a rectangle
2. It writes two commands: 
* Object rotated - actual rotation
* Changed points - I don't see any visual difference when making undo/redo of this

There should be a single command in history

cvat-canvas/src/typescript/objectSelector.ts
lexiv0re, line 1: `Ctrl+click` no longer adds anything to multi-selection for me in any mode (Cursor/Selection) - is it expected? It just does nothing for non-active shape or opens the context menu for an active shape.

senketsukamui: Are you on Mac? I changed shortcuts to Command + click

lexiv0re: Oh, ok. Yes, I'm on Mac.

senketsukamui: Done. Ctrl/Command-click and Shift-based selection changes are now recorded as `Changed selection` history actions, so Undo and Redo restore the previous and next selected object IDs correctly.

cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/select-control.tsx
senketsukamui: Done. Toolbar highlighting now depends only on the active control: Ctrl/Command-click can modify a selection while Cursor remains active, and **Select objects** is highlighted only when persistent selection-box mode is explicitly enabled.

senketsukamui: Done. Shift-drag now enters persistent Select objects mode only when the box selects at least one object; an empty Shift-drag keeps Cursor active. Explicitly enabled Select objects mode remains persistent.

senketsukamui: Done. While Select objects mode is active, unmodified clicks now toggle object membership through the selector and keep the mode active. The regular click-to-clear behavior only runs outside Select mode.

cvat-canvas/src/typescript/objectSelector.ts
senketsukamui: Done. The selection box and label now use the default cursor while Shift or Ctrl/Command is pressed, and selection dragging is blocked under those modifiers so the cursor no longer misleadingly indicates a move action.

senketsukamui: Done. Ctrl/Command-click now uses the same semantic annotation hit-testing as regular hover and activation instead of relying on the topmost SVG element, allowing enclosed or overlapping objects to be targeted correctly.

senketsukamui: Done. Selections with no movable members are now marked as non-draggable and display the default cursor. The drag handler also rejects movement when there are no movable selected objects, preventing an unintended bbox-selection action.

senketsukamui: Done. Group dragging now saves movable members as one batch update and then reapplies the original selected object IDs, so the selection remains active regardless of how it was created. Stationary members also remain selected.

senketsukamui: Done. Skeletons are no longer excluded from movable selection members; their points are translated by the same group delta and saved with the rest of the selection.

senketsukamui: Done. Tags are excluded from multi-object selection because they have no canvas geometry for displaying or manipulating the selection overlay; selection inputs are sanitized to shapes and tracks only, while tags remain individually editable.

senketsukamui: Done. Selected polylines and skeleton edges now use thicker, fully opaque strokes, while Points objects receive emphasized point borders, making their selected state consistent with other shapes.

cvat-ui/src/components/annotation-page/standard-workspace/objects-side-bar/objects-list.tsx
senketsukamui: Done. In Layer ordering, the sidebar now builds selection ranges from the flattened visible object order across expanded layers, so Ctrl/Command-click and Shift-click work correctly. Objects inside collapsed layers are excluded from range selection.

senketsukamui: Done. Ctrl/Command-clicking a label now adds or removes all selectable objects with that label, while Shift-click selects a continuous range of labels from the previous anchor. Label rows also reflect when all their selectable objects are selected.

senketsukamui: Done. Clicking interactive controls on selected sidebar items no longer triggers selection-clearing logic, and annotation updates preserve the current selected IDs. Unlocking or unpinning an item therefore keeps the selection active for subsequent movement.

cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/select-control.tsx
senketsukamui: Done. The ellipsis now stops canvas event propagation and sends an explicit toggle request, so the menu opens on the first click and closes on the next regardless of how the selection was created.

cvat-ui/src/components/annotation-page/canvas/views/canvas2d/canvas-wrapper.tsx
senketsukamui: Done.

cvat-ui/src/components/annotation-page/canvas/views/canvas2d/canvas-wrapper.tsx
senketsukamui: Done. Batch updates and annotation refreshes now preserve the selected object IDs, so label, attribute, lock/pin, layer, and grouping actions can be applied consecutively. Delete clears the selection, while paste replaces it with the newly created objects.

senketsukamui: Done. **Make a copy** now copies the selection and immediately starts interactive placement, matching the single-object workflow. Ctrl/Command+C remains clipboard-only, with Ctrl/Command+V starting placement.

senketsukamui: Done. The disabled label selector is now wrapped by a tooltip-enabled parent container, so the reason is displayed when hovering anywhere over the selector, including the selected value.

cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/select-control.tsx
senketsukamui: Done. The selection bounding box now handles `contextmenu` directly, stops propagation, and opens the selection menu at the click position. Selection-clearing logic is limited to left clicks, so a one-object selection remains active.

cvat-canvas/src/typescript/objectSelector.ts
senketsukamui: Done. Dragging the **SELECTION (N)** header now forwards the pointer interaction to the bounding-box drag handler, so it moves the selection identically. Menu-button clicks and non-draggable selections remain excluded.

senketsukamui: Done. Frame navigation now carries selected track IDs forward and keeps those that remain visible and available on the destination frame. Frame-local shapes and unavailable, outside, hidden, filtered, or hidden-layer tracks are removed from the selection.

cvat-ui/src/components/annotation-page/canvas/views/canvas2d/selection-attributes-editor.tsx
senketsukamui: Done. The separate attributes modal was removed, and multi-selection now uses the same `cvat-object-item-menu` style as single-object controls. Shared attributes are edited inline through the familiar expandable **Details** section.

senketsukamui: Done. Attribute edits now apply inline through one batched `Changed attributes` history transaction while preserving the active selection. A single Undo reverts the complete change, and Redo remains available.

cvat-ui/src/containers/annotation-page/standard-workspace/objects-side-bar/objects-list.tsx
senketsukamui: Done. The existing **L** and **P** shortcuts are now reused: they apply to the active multi-selection first and fall back to the active object when no selection exists. The separate multi-selection shortcut sequences were removed.

cvat-ui/src/components/annotation-page/standard-workspace/controls-side-bar/select-control.tsx
lexiv0re: `Shift+drag` resulting in no selected objects still switches the workspace to `Select objects` mode.

cvat-ui/src/components/annotation-page/standard-workspace/objects-side-bar/objects-list.tsx
lexiv0re: `Shift+click` still doesn't work for me in Layer ordering.

cvat-canvas/src/typescript/canvasView.ts
lexiv0re, line 1: 1. Select multiple objects
2. Hide one of the selected objects
3. Undo
The object is displayed again but no longer a part of the selection. I'd expect `hide` to record `deselect + hide` in a single command and undo to restore the selection as well. If one just clicks `hide`, `unhide` it's ok that it's no longer selected.

lexiv0re, line 1: 1. Cmd+click to select a Skeleton
2. Zoom in/out
Selection point borders change their width while they should not.

The same problem with Points and Polyline.

Image

lexiv0re, line 1: `Shift+drag` switches to `Select objects` mode.
Selecting with `Cmd+click` doesn’t switch.

lexiv0re, line 1: Not sure how I got into this state but at some point the cursor was broken.
When selecting with `Cmd+click` it turned the cursor into grab and it was grab for all the canvas regardless of where I move or click.
For `Shift+drag` selection it was pointer and was never turning to grab even when I hovered over the draggable selection.
Maybe you can find it in the code where it can possibly get stuck with the grab cursor. I believe it might be related to dragging by selection header as I was testing this scenario when I ended up in this state. But I was not able to reproduce this for the second time.

lexiv0re, line 1: Select mode behavior is still confusing.
1. Click `Select objects` mode in the control bar
2. Without any `Shift/Cmd` modifiers start clicking on the objects
3. It only adds objects to the selection on click, second click doesn’t exclude object from selection
4. Not all objects can be selected this way: e.g. if you click on a point in a `Skeleton` or `Points` shape it’s not selected. 
5. It doesn’t allow to select object behind the selection bbox
Maybe it could work the same way `Cmd+click` works but without `Cmd` needed? I don’t know if you discussed the behavior in this mode?

lexiv0re, line 1: 1. Create multi-selection
2. From its context menu select `Make a copy`
3. Try to position it that part of it is outside the canvas
It doesn’t allow you such placement.
For regular shape's `Make a copy` it allows such placement and cuts the object geometry to fit the canvas.

lexiv0re, line 1: 1. Multi-select 3 objects
2. Delete one of them from the context menu of the right-hand objects list
The object is deleted but the selection still says `SELECTION(3)` although there’re only 2 objects left in the selection.

cvat-ui/src/components/annotation-page/canvas/views/canvas2d/canvas-wrapper.tsx
lexiv0re: I can still see this problem. `v` icon on the right is missing on hover and when opening the `Select` you can see it briefly replaced with magnifier glass icon (search) but it quickly disappears so the opened `Select` also misses the icon there. Please compare with similar label selection control for regular single shape active mode.

cvat-canvas/src/typescript/canvasView.ts
lexiv0re, line 1: It's possible to have both popups opened at the same time:
1. Open for single selection
2. `Cmd+click` on the shape
3. Open for multi-selection

Image

lexiv0re, line 1: `Details` label looks disabled even though it's not and the section can be expanded.

Image

lexiv0re, line 1: The menu moves on zoom in/out and can end up being cropped:

Image

lexiv0re, line 1: I think the current layout for the menu is not the best. It's very tall because of all these actions are displayed one per line and then the details section at the bottom barely fits my monitor (27-inch (2560 × 1440), DPR=1). On a laptop screen (13,3-inch (2560 × 1600), DPR=2) its just cropped right away without it being possible to access the bottom part completely:

Image

senketsukamui: Done. Hiding a selected object now records the visibility and selection changes as one history action, so Undo restores both. Manually showing the object again does not reselect it.

senketsukamui: Done. Selection point borders for Skeletons, Points, and Polylines are now normalized by the canvas scale, so their visual width remains constant while zooming.

senketsukamui: Done. Shift+drag keeps **Select objects** active only after selecting at least one object; an empty drag returns to **Cursor**. Cmd/Ctrl+click remains a one-off selection toggle and intentionally does not change the active tool.

senketsukamui: Done. The selection box and header now consume their complete left-button gesture, preventing rejected or modifier-assisted drags from starting canvas panning and leaving the grab cursor stuck. Valid header and selection-box drags still move the selection normally.

senketsukamui: Done. Unmodified clicks in **Select objects** mode now use the same semantic toggle behavior as Cmd/Ctrl+click, including removing an object on the second click and selecting Points or Skeletons. Zero-movement clicks intercepted by the selection bounding box are also forwarded to object hit-testing.

senketsukamui: Done. Multi-selection paste now uses the same shape-specific geometry fitting, cropping, and validation as single-object **Make a copy**. Valid objects are created as one batch, while objects rejected by the single-object paste rules are omitted.

senketsukamui: Done. Individually deleting a selected object now removes it from the active selection, so the selection counter and bounding box immediately reflect the remaining objects.

senketsukamui: Done. Opening the multi-selection actions menu now closes any open single-object menu, preventing both popups from being displayed simultaneously.

senketsukamui: Done. The Details header now keeps its normal enabled appearance while only the unavailable attribute controls are shown as disabled.

senketsukamui: image

I have an idea to move layer actions to a submenu. What do you think?
Also, we can replace Lock/Pin selection with icons under label select, which will lower the height of context menu

lexiv0re: I would reuse the existing layout for one shape adjusting it for the selection group:
1. It's consistent and familiar for the user
2. No need to invent anything

image

senketsukamui: Done.

senketsukamui: Done. The selection context menu remains anchored to its opening point during zoom and can be reopened at another right-click position. It also flips or shifts automatically to stay within the viewport.

changelog.d/20260826_000000_glebpotapov_multi_object_selection.md
```diff
@@ -0,0 +1,5 @@
+### Added
+
+- Added multi-object selection in the 2D annotation workspace, including
+  bounding-box selection, batch operations, and undo and redo support
```
bsekachev, line 4: you probably mean selection with a bounding box, but it sounds like selection of bounding boxes

Also, I believe the idea might be simple, e./g/

- Multiple objects selection and batch operations on 2D annotation workspace (or something like this)

Ways of selections, support undo / redo, etc - details

bsekachev:
@codex review

cvat-core/src/annotations-collection.ts
```diff
@@ -199,6 +199,126 @@ export default class Collection {
         return updatedStates;
     }
 
+    public updateBatch(objectStates: ObjectState[]): ObjectState[] {
```
bsekachev, line 202: How is it compatible with https://github.com/cvat-ai/cvat/pull/11103 where you use another batch method?

cvat-core/src/annotations-collection.ts
```diff
@@ -199,6 +199,126 @@ export default class Collection {
...
+        return updatedStates;
+    }
+
+    public removeBatch(objectStates: ObjectState[], force: boolean): number[] {
+        checkObjectType('objectStates', objectStates, null, { cls: Array, name: 'Array' });
+
+        const clientIDs = new Set<number>();
+        const objects = objectStates.map((state) => {
+            const object = this.objects[state.clientID];
+            if (!(object instanceof Shape || object instanceof Track || object instanceof Tag) || object.removed) {
```
bsekachev, line 285: if object already removed, why can't we consider this action as successful?

cvat-core/src/annotations-collection.ts
```diff
@@ -199,6 +199,126 @@ export default class Collection {
...
+                throw new ArgumentError(`Object with client ID ${state.clientID} cannot be removed`);
+            }
+            if (object.lock && !force) {
+                throw new ArgumentError(`Object with client ID ${state.clientID} is locked`);
```
bsekachev, line 289: Should not we just skip locked object instead of failing the whole operation?

cvat-core/src/annotations-collection.ts
```diff
@@ -199,6 +199,126 @@ export default class Collection {
...
+            }
+            if (state.isGroundTruth) {
+                throw new ArgumentError(`Ground truth object with client ID ${state.clientID} cannot be removed`);
```
bsekachev, line 292: Basically same policies. Ignoring them instead of failing

cvat-core/src/annotations-collection.ts
```diff
@@ -199,6 +199,126 @@ export default class Collection {
...
+            }
+            if (clientIDs.has(state.clientID)) {
+                throw new ArgumentError(`Object with client ID ${state.clientID} is duplicated`);
+            }
```
bsekachev, line 296: Basically I do not think it is possible
But even if it is - what is the problem with that?

cvat-core/src/annotations-collection.ts
```diff
@@ -199,6 +199,126 @@ export default class Collection {
...
+        });
+
+        const removedObjects: (Shape | Track | Tag)[] = [];
+        this.history.beginTransaction(HistoryActions.REMOVED_SELECTION);
+        try {
+            for (let index = 0; index < objectStates.length; index++) {
+                const state = objectStates[index];
+                const object = objects[index];
+                if (!object.delete(state.frame, force)) {
+                    throw new ArgumentError(`Object with client ID ${state.clientID} could not be removed`);
```
bsekachev, line 308: See comments above

cvat-core/src/annotations-collection.ts
```diff
@@ -1275,8 +1395,12 @@ export default class Collection {
                         attributes,
                         descriptions: state.descriptions,
                         frame: state.frame,
-                        group: 0,
+                        group: state.group?.id || 0,
```
bsekachev:
```suggestion
                        group: state.group?.id ?? 0,
```

cvat-core/src/annotations-collection.ts
```diff
@@ -1305,7 +1433,11 @@ export default class Collection {
                         attributes: attributes.filter((attr) => !labelAttributes[attr.spec_id].mutable),
                         descriptions: state.descriptions,
                         frame: state.frame,
-                        group: 0,
+                        group: state.group?.id || 0,
```
bsekachev, line 1436:
```suggestion
                        group: state.group?.id ?? 0,
```

cvat-core/src/server-response-types.ts
```diff
@@ -283,6 +283,11 @@ export interface SerializedShape {
     rotation: number;
     z_order: number;
     type: ShapeType;
+    // Client-side state used when cloning; absent from server responses.
+    pinned?: boolean;
+    lock?: boolean;
+    hidden?: boolean;
+    color?: string;
```
bsekachev, line 290: What is the reason for changing the existing copy/paste behavior to preserve client-side properties such as lock, hidden, pinned, and color? Multi-object copy does not appear to require this, and previously pasted objects received the normal defaults. Was this behavioral change explicitly requested?

bsekachev: I see an inconsistency in the new behavior: when a locked object is copied and pasted individually, the pasted object is unlocked, but when the same object is copied as part of a selection, the pasted object preserves its locked state.

cvat-core/src/annotations-collection.ts
senketsukamui: It's not yet compatible, I'm gonna update this, when #11103 is merged to develop

cvat-core/src/server-response-types.ts
senketsukamui: > Multi-object copy does not appear to require this, and previously pasted objects received the normal defaults. Was this behavioral change explicitly requested?

I thought that copying properties is more consistent, but if we are keeping behavior close to single objects, I will revert it then

cvat-canvas/src/typescript/canvasView.ts
senketsukamui: Done. The bug already exists in app.cvat.ai by the way

cvat-ui/src/components/annotation-page/standard-workspace/objects-side-bar/objects-list.tsx
senketsukamui: Done

cvat-ui/src/components/annotation-page/canvas/views/canvas2d/canvas-wrapper.tsx
senketsukamui: Fixed

cvat-canvas/src/typescript/canvasView.ts
lexiv0re: It does but if I undo such deletion the returned object no longer a part of the selection. I'd expect undo to restore selected state for the deleted object.

lexiv0re: The only change I see is that now it creates a scroll so you can access the content at the bottom which is good but still it looks not good as you have to scroll the whole annotation layout down:
image

Is there an issue to use layout similar to the single object popup displayed on right click where most of the actions fit in 2 lines of just icons and the remaining available through `...` menu? It's much more compact and will look familiar to the users.

lexiv0re, line 1: Selection result depends on the type of selection being used which is confusing:

1. `Shift+select` one object
2. `Shift+select` another object
Both objects are in the selection

1. `Cmd+click` to select one object
2. `Shift+select` another object
Only the latter object is selected.

It’s because now `Cmd+click` selection does not change the mode to `Select objects` while `Shift+select` does.

I'd expect both scenarios to lead to the same final selection.

senketsukamui: Oh, you've meant to use the menu completely, I didn't get you.

lexiv0re: It leads to confusing workflows. Added another comment describing them. So resolving this one.

lexiv0re: It works now for all shapes but `Skeleton`. For the latter it still doesn't restore the selection on undo.

senketsukamui: image
@lexiv0re you meant something like that?

lexiv0re: Yes, as it makes it much more compact.

senketsukamui: Done. Each Shift-drag now starts a new selection, so both scenarios finish with only the object selected by the latest Shift-drag; `Cmd`-click remains additive on its own.

senketsukamui: Done, context menu UI updated

senketsukamui: I've updated it for skeleton too

senketsukamui: Fixed

lexiv0re, line 1: I don't see any difference between the `Select objects` and `Cursor` modes now. Both require `Cmd/Shift` modifier for selection. Both select objects the same way. 
Do we still need the separate `Select objects` mode?

lexiv0re, line 1: 1. `Shift+select` an item(-s). Important is to be in `Select objects` mode.
2. Right click on the selection.
The context menu is immediately closed.
In `Cursor` mode it stays open as expected.

lexiv0re: Actually, it's not the case, DnD selection still works without `Shift`.
So I think there's just a bug/changed behavior for single click:
it no longer adds/removes to/from selection without Cmd.
Is it intentional?

lexiv0re, line 1: Selection context menu is closed after any action.
Single object context menu is not closed so you can perform multiple actions while one menu is opened. 
Lets make it consistent - working the same way it works for a single object context menu.

lexiv0re, line 1: There probably should be a separate handling for multi-selection for the scenario when one object gets completely out of the canvas when the selection is moved:

https://github.com/user-attachments/assets/0ab0160e-2eeb-4c4a-88f2-aee31a8c16b8

I get that it currently just applies moves to individual objects and for an individual object it works this way. But for an individual object it makes sense. For the part of the selection - not really IMO.
I'd prefer it to discard the object completely in this case. And only rollback the position when the whole selection ends up outside the canvas.

senketsukamui: Thanks for catching this bug, I got it once and didn't understand how I did it.

senketsukamui: I couldn't come to a general solution to handle OOF situations with different types of objects, I settled at the current solution, because it keeps individual objects "features" such cutting rects and masks with the border and allow things like circles to be completely/partially outside of frame. I'm not sure we should "discard" objects  completely only for selection, when in case of single rect, for example, it returns to initial position, so it could be confusing.

lexiv0re: Overall, I think it can be addressed later in a separate PR.

As for discarding, it's consistent with single scenario in the following sense:
Single:
1. Shape inside - OK, just move
2. Shape partially outside - crop the outstanding part, keep the remaining
3. Shape fully outside - invalid movement position, rollback to initial position

Multi-selection:
1. The whole selection inside - OK, just move
2. Selection partially outside - crop the outstanding part, keep the remaining. One shape of multi-selection fully outside falls into this category. It's just a part of multi-selection that's outside and should be cropped. In this case completely removed.
3. Selection fully outside - invalid movement position, rollback the whole selection to the initial position

But again, I'm not insisting it's the best option, maybe not. So we can discuss it separately.

senketsukamui: > Selection fully outside - invalid movement position, rollback the whole selection to the initial position

Here's the issue for selection, we can have objects that are allowed to be OOF, such as ellipses or cuboids, I will make notes to address this problem later, but I don't think that is too critical for now

senketsukamui: It was, but now I think it's really more natural to let just click on objects in select mode to add/remove them, so I returned this feature back.

senketsukamui: Fixed

senketsukamui: Good call, fixed

lexiv0re: One more subtle case which remains inconsistent:
When a single object is hidden from the context menu the menu stays visible.
When a selection is hidden from the context menu the menu is closed.

lexiv0re, line 1: 1. Duplicate rotated object as a part of the selection. 
Movement is not correctly applied to its position. See the attached recording.

https://github.com/user-attachments/assets/0f8623ce-1ffa-4883-8dd5-12eac7377f24

lexiv0re, line 1: Selection context menu is not always closed on outside click.
1. In "Select objects" mode open selection menu
2. Click outside the menu
It's not closed.
It's expected to close as would menu for a single object close.

lexiv0re, line 1: `Shift+selection` selects two objects but only one was fully in the selection frame.

https://github.com/user-attachments/assets/8b28f2c2-7242-446f-b8a4-ebe5baff50ab

senketsukamui: Fixed, I've made it stay visible in both cases, the same as for single object

senketsukamui: Fixed

senketsukamui: Fixed

senketsukamui: Fixed

lexiv0re (approved):
I didn't check the code. UX LGTM.

bsekachev:
@senketsukamui 

- Generally I think it is a great job and like selection UX now!
- We probably miss select + propagate, this is not critical, but I know this is a real use-case. And we can add it later. 
- Small bug: select an object (with ctrl) or two -> delete -> undo -> go next frame -> go previous frame -> try redo (shall remove two object but not working)
- It seems now we have two ways to copy objects: `copyShape` and `copySelection`. And apparently even for one selected object pipeline works via `copySelection`. From first look it is redundant. 

```
copyShape()
→ Redux: drawing.activeInitialState
→ pasteShapeAsync()
→ canvas.draw({ initialState })
→ DrawHandler / MasksHandler
→ canvas.drawn
→ onCanvasShapeDrawn()
→ createAnnotationsAsync()
→ job.annotations.put()


copySelection()
→ snapshotSelectionState()
→ Redux: drawing.copiedStates[]
→ pasteSelectionAsync()
→ startPastePlacementAsync()
→ canvas.draw({ initialStates, onDrawDone })
→ CanvasView.startSelectionPaste()
→ CanvasView.placeSelectionPaste()
→ onDrawDone()
→ placeCopiedStatesAsync()
→ job.annotations.put()

```

- Apparently in 3D workspace I can select multiple objects (what is useless because there is nothing we can do with selection in this workspace). I think for 3D we should not be able to use `cmd + click` for now.
image


- Did not you check why tests failing? 

```
  Multi-object selection
    ✓ Keeps selection control borders at a constant screen size while zooming (30300ms)
    ✓ Adds and removes objects with Mod-click on the canvas and in the Objects tab (956ms)
    ✓ Replaces objects selected with Mod-click when starting a Shift-drag selection (593ms)
    ✓ Selects a sidebar range with Shift and all selectable objects with Mod+A (859ms)
    ✓ Selects only objects fully contained by the selection bbox (734ms)
    ✓ Ignores the click emitted after a selection drag (741ms)
    ✓ Toggles semantic object hits with unmodified clicks in Select mode (1070ms)
    ✓ Keeps Cursor active when a Shift-drag selects no objects (661ms)
    ✓ Starts bbox selection with Shift-drag while the Cursor tool is active (524ms)
    ✓ Excludes hidden objects and tags from selection (942ms)
    ✓ Removes an object from the active selection when it is hidden (1057ms)
    ✓ Toggles a layer selection when the layer contains hidden objects (1362ms)
    ✓ Selects complete label and layer groups and works with Layer ordering (1869ms)
    ✓ Limits Layer Stack Shift-click ranges to one layer (2602ms)
    ✓ Keeps selected tracks across frames and restores the previous selection with Undo (1360ms)
    ✓ Restores the complete selection when undoing an Escape clear (790ms)
    ✓ Moves a selection as one action and restores it with Undo (1111ms)
    ✓ Keeps a rotated shape aligned with selection movement previews (2318ms)
    ✓ Moves unlocked members while locked members stay selected and stationary (1051ms)
    ✓ Locks, unlocks, pins, and unpins every selected object (1001ms)
    ✓ Occludes and hides the selection with quick actions, shortcuts, and batch history (1983ms)
    ✓ Keeps selection actions open while hiding and showing selected objects (912ms)
    1) Disables incompatible actions and normalizes a mixed lock state
    ✓ Does not pan the canvas when a modified selection-header drag is rejected (640ms)
    2) Keeps selection actions open after a Shift-drag selection
    ✓ Opens selection actions by right-clicking the selection bbox (797ms)
    ✓ Does not reopen selection actions after clearing and starting a new selection (788ms)
    ✓ Closes an object menu before opening selection actions (921ms)
    3) Runs layer and annotation actions for the complete selection
    4) Applies layer actions only to editable selection members
    5) Changes common labels and attributes and groups the selection
    ✓ Copies and pastes a selection with platform modifier shortcuts (1459ms)
    6) Uses normal client-state defaults for pasted selection objects
    ✓ Places a copied selection partially outside the frame (1246ms)
    7) Copies and deletes a selection as batch history actions
    8) Skips locked objects when deleting a selection
    9) Removes an individually deleted object from the selection
    10) "after all" hook for "Removes an individually deleted object from the selection"
```

PR #11097, opened 2026-08-27
Audio workspace: add split interval function

lexiv0re:
### Motivation and context

Some scenarios require splitting existing interval into multiple. It's only possible by manually reducing the existing interval and creating a new one.

This PR introduces new **Split interval at playback position** function.
* New button in the left control bar
* Default shortcut: **S**
SplitControl

* If no visible intervals are under the playback - does nothing
* If there's a single visible interval under the playback - splits it in two and selects the right part
* If there're multiple visible intervals under the playback - opens the picker to select which one to split
ChooseToSplit

**Note:** split does not work for locked or pinned intervals. If there are multiple intervals, the picker still lists pinned/locked intervals with the corresponding icon and they cannot be selected.

### How has this been tested?

* Tested manually
* Added e2e tests

### Checklist

### License


bsekachev:
> Default shortcut: S

I think it is similar to feature "slice" we have in CV workspace. 
And default shortcut there is Alt + J. So, maybe it would be good to use the same

bsekachev:
> If no visible intervals are under the playback - does nothing
If there's a single visible interval under the playback - splits it in two and selects the right part
If there're multiple visible intervals under the playback - opens the picker to select which one to split

I believe the things are much simpler if we may split only currently activated interval. 
Are there any reasons why you did not like this approach? I believe it would be consistent with existing features. Usually in order to change object with a hotkey (except for bulk changes, like lock all, hide all) you need to activate this object.

bsekachev:
> or pinned intervals

I also think pinned property should not block splitting. This feature does not allow moving intervals only.

lexiv0re:
I can see the following existing shortcuts:
image
image

`Alt+J` is closest behaviour-wise?

lexiv0re:
> I believe the things are much simpler if we may split only currently activated interval.
Are there any reasons why you did not like this approach? I believe it would be consistent with existing features. Usually in order to change object with a hotkey (except for bulk changes, like lock all, hide all) you need to activate this object.

I implemented it to work with active interval only initially but I found it to be not convenient during testing: too many times I first listened to the place where I want to cut and only then realized that I don't have this interval selected. Now if I select it from the waveform it seeks to the clicked position effectively changing the place where I wanted to cut. With high zoom it could be clicked close to the playback but with low zoom the difference can be substantial. Alternatively, you can select an interval from the right-hand list of intervals, but if there are a lot of them it might take a while to find the correct one. 
The final approach is more complicated but the UX felt much smoother so I preferred it over the one working with active intervals only.

lexiv0re:
> I also think pinned property should not block splitting. This feature does not allow moving intervals only.

I was on a fence about this one. 
Currently, pin prevents moving and resizing an interval. 
From my perspective a user would want to pin an interval when they are happy with its boundaries and wanna make sure they won't be accidentally changed.
On one hand, split changes the interval boundary. On another hand you won't likely accidentally split an interval.

So if you think `Pin` shouldn't prevent `Split`, I'll change it this way.

bsekachev:
> Alt+J is closest behaviour-wise?

Maybe `split` is closer, but actually it very depend on perspective. 
From naming - yes, split is even closer

bsekachev:
> Currently, pin prevents moving and resizing an interval.

I do not think pinned should prevent resizing. It does not in CV workspace.

bsekachev:
> On another hand you won't likely accidentally split an interval. 

Ctrl + Z?

lexiv0re:
> I do not think pinned should prevent resizing. It does not in CV workspace.

Oh, ok, in this case split should definitely be allowed. I'll update both resize and split to work for pinned intervals.

bsekachev:
> I implemented it to work with active interval only initially but I found it to be not convenient during testing ...

I see. I think it may works similar to what you implemented.

We have almost the same context in slice feature.
- if you have an active polygon/mask when activating slice tool, the tool will be applied to this object.
- if you do not have, the tool will suggest you to click the object. In your implementation clicking is replaced by a menu with objects - conceptually the same thing. 

The only adjustment I would suggest in this case - is to use an activated object if any.
The only question is what to do if we have single and not activated under cursor. If you find "auto selection" of such object is convenient, so, we may keep it also

lexiv0re:
Yes, to me it felt much smoother compared to requiring always selecting the interval for splitting explicitly. 
So the final logic may be as:
* If no visible intervals are under the playback - does nothing
* If there's a single visible interval under the playback - splits it in two and selects the right part
* If there're multiple visible intervals under the playback - if one of them is active then split the active one. Otherwise, display the interval picker for explicit selection.

Does it sound good?

lexiv0re:
* Changed default shortcut to `Alt+M` to be consistent with other split functions
* Enabled `Split` and `Resize` functions for pinned intervals
* Made selected interval preferred split target if under playback and there are multiple options

bsekachev (approved):

